PR #7, opened 2016-09-06
Bug Fix#233637 :  DataSource tests name is not getting appended by "(Data Row *)"

jayaranigarg:


src/Adapter/MSTestAdapter/ObjectModel/UnitTestResult.cs
```diff
@@ -37,6 +38,7 @@ internal UnitTestResult(TestFailedException testFailedException)
         {
             this.Outcome = testFailedException.Outcome;
             this.ErrorMessage = testFailedException.Message;
+            DatarowIndex = -1;
```
AbhitejJohn: nit: Just call the default constructor with a this() for this ctor and the next.

jayaranigarg: Done.

PR #16, opened 2016-10-02
RFC for the Framework extensibility for description attributes.

AbhitejJohn:


docs/001-Framework-Extensibility-Description-Attributes.md
```diff
@@ -0,0 +1,91 @@
...
+## Summary
+This details the MSTest V2 framework extensibility for attributes that describe a test.  
+
+## Motivation
+It is much more convenient to have strongly typed description attributes over a test method as opposed to just a KeyvaluePair<string,string>. Users can also filter tests based on the values of these attributes.    
+
+## Detailed Design
+
+### Requirements
+1. One should be able to have a strongly typed desciption attribute on a test method.
```
SadagopanRajaram, line 12: Source compatability with mstest v1 is also a requirement. 
In addition, we need to make sure that compat in integration scenarios with existing loggers etc. is maintained
We also have a goal to make sure that all attributes meet our extensibility design

AbhitejJohn: > Added the source compat as another requirement. 
> 
> Currently the rocksteady V1 adapter does not include the values of default attributes other than Owner and Priority in the TestCase object. Making these extend from TestProperty class would now ensure they are part of the TestCase object in V2 so loggers can consume them. However the format in which loggers should see these should be similar to mstest.exe flow. 
> 
> The attributes that are missing from V1 are Description, WorkItem, CssIteration and CssProjectStructure. Verified that all four of them fit into this model well.

AbhitejJohn: Also added a section on the requirements from the Test Platform.

Issue #8, opened 2016-09-08
TestContext.WriteLine throwing System.FormatException when json string is passed.


harshjain2:
## Description

> TestContext.WriteLine throwing System.FormatException when json string is passed.

Method Signature:
`public abstract void WriteLine(string format, params object[] args);`
## Steps to reproduce
- Create a unit test project, add MSTest.TestFramework dll and execute the below test method

```
[TestClass]
    public class UnitTest1
    {
        private TestContext testContextInstance;
        /// <summary>
        ///Gets or sets the test context which provides
        ///information about and functionality for the current test run.
        ///</summary>
        public TestContext TestContext
        {
            get
            {
                return testContextInstance;
            }
            set
            {
                testContextInstance = value;
            }
        }
        [TestMethod]
        public void TestMethod1()
        {
            var msg = "{curly braces}";
            //Debug.WriteLine(msg);
            TestContext.WriteLine(msg);
        }
    }
```
## Expected behavior

Test sholuld pass, user should be able to write the json in output.
## Actual behavior

Test is failing with System.FormatException. 

RCA :
This is happening because string.format is used for implementing this API and it expects format which placeholders {0}, {1}, etc. and corresponding arguments. 

But when json string is passed, it just throws System.FormatException as string is present inside curly braces.

Workaround:
As a workaround, user has to add one extra curly brace for every curly brace present in string. 

`string str = "{{mystring}}";`

Fix : 
Do not use string.format api when count of arguments are null in implementation of API
`public abstract void WriteLine(string format, params object[] args);`

PR #17, opened 2016-10-20
Addd new API TestContext.WriteLine(string message).

harshjain2:
Fix for issue : [https://github.com/Microsoft/testfx/issues/8](url)

Change : 
Added overload method of TestContext.WriteLine with formatted string as input.

### DIFF
--- a/src/TestFramework/UnitTestFramework.Extension.Desktop/TestContext.cs
+++ b/src/TestFramework/UnitTestFramework.Extension.Desktop/TestContext.cs
@@ -18,6 +18,12 @@ namespace Microsoft.VisualStudio.TestTools.UnitTesting
         /// <summary>
         /// Used to write trace messages while the test is running
         /// </summary>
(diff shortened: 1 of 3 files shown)


src/TestFramework/UnitTestFramework.Extension.Desktop/TestContext.cs
```diff
@@ -18,6 +18,12 @@ public abstract class TestContext
         /// <summary>
         /// Used to write trace messages while the test is running
         /// </summary>
+        /// <param name="message">formatted message string</param>
+        public abstract void WriteLine(string message);
```
AbhitejJohn, line 22: Can you check if Extension.Core needs a similar API?

harshjain2: We are not supporting TestContext.WriteLine in Extension.Core, that's why I haven't added this new overloaded method there.

PR #31, opened 2017-01-17
Modifications in scripts for CI/PR build

jayaranigarg:


scripts/Build.ps1
```diff
@@ -80,3 +85,5 @@ function Print-Help {
 Print-Help
 Perform-Restore
 Perform-Build
+
+if ($Script:ScriptFailed) { Exit 1 } else { Exit 0 }
```
AbhitejJohn: should this just return even if Perform-Restore fails without running Perform-Build?

PR #39, opened 2017-01-30
Enable localization and signing.

AbhitejJohn:
Fixed up localization and signing. While this currently relies on the lcl format for signing. We would need to move to an xlf format.

### DIFF
--- a/scripts/Build.ps1
+++ b/scripts/Build.ps1
@@ -8,7 +8,11 @@ Param(
   [Alias("c")]
   [string] $configuration = "Debug",
   
-  [switch] $clearPackageCache
+  [switch] $clearPackageCache,
+  [switch] $templates,
+  [switch] $wizards,
+  [switch] $vsmanprojects,
+  [switch] $full
 )
 
 $ErrorActionPreference = "Stop"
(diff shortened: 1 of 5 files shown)


scripts/build/TestFx.Settings.targets
```diff
@@ -2,10 +2,11 @@
 <Project ToolsVersion="4.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
   <PropertyGroup>
     <TestFxRoot Condition="$(TestFxRoot) == ''">..\..\</TestFxRoot>
+    <RepoRoot>..\..\</RepoRoot>
```
jayaranigarg, line 5: Why do we need a separate variable for RepoRoot? when TestFxRoot effectively provides same value...

scripts/build/TestFx.targets
```diff
@@ -3,17 +3,33 @@
   <Import Project="$(MSBuildExtensionsPath32)\Microsoft\Portable\$(TargetFrameworkVersion)\Microsoft.Portable.CSharp.targets" Condition="$(TargetFrameworkProfile) == 'Profile259'" />
   <Import Project="$(MSBuildToolsPath)\Microsoft.CSharp.targets" Condition="$(TargetFrameworkProfile) == '' or $(TargetFrameworkProfile) != 'Profile259'"/>
 
-  <Import Project="$(TestFxRoot)packages\MicroBuild.Core.0.2.0\build\MicroBuild.Core.targets" Condition="Exists('$(TestFxRoot)packages\MicroBuild.Core.0.2.0\build\MicroBuild.Core.targets')" />
+  <!-- Import props/targets with $(RepoRoot) since msbuild takes the relative path based on settings.targets and not with respect to the project. -->
+  <Import Project="$(RepoRoot)packages\MicroBuild.Core.0.2.0\build\MicroBuild.Core.targets" Condition="Exists('$(RepoRoot)packages\MicroBuild.Core.0.2.0\build\MicroBuild.Core.targets')" />
```
jayaranigarg, line 7: same here

scripts/build/TestFx.Settings.targets
AbhitejJohn: TestFxRoot is defined relative to the csproj. RepoRoot is relative to the targets file. :)

PR #95, opened 2017-03-06
Fixing issues with deployment item from VS IDE

AbhitejJohn:
Moved the DeploymentItem attribute over to the desktop Fx extension. This fixes two issues:

1. It ensures that the Fx extension assembly gets deployed as well so it can be used at runtime.
2. Removes confusion of DeploymentItem support in UWP and Core as its a no-op today without any warnings/notification.

### DIFF
--- a/src/TestFramework/Extension.Core/Attributes/DeploymentItemAttribute.cs
+++ b/src/TestFramework/Extension.Core/Attributes/DeploymentItemAttribute.cs
@@ -16,8 +16,10 @@ namespace Microsoft.VisualStudio.TestTools.UnitTesting
     /// [DeploymentItem("file2.xml", "DataFiles")]
     /// [DeploymentItem("bin\Debug")]
     /// </example>
+    /// <remarks>
+    /// DeploymentItemAttribute is currently not supported in .Net Core. This is just a placehodler for support in the future.
+    /// </remarks>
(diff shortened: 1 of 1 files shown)


src/TestFramework/Extension.Desktop/Attributes/DeploymentItemAttribute.cs
```diff
@@ -0,0 +1,53 @@
...
+    /// Can have multiple instances of the attribute to specify more than one item.
+    /// The item path can be absolute or relative, if relative, it is relative to RunConfig.RelativePathRoot.
+    /// </summary>
+    /// <example>
+    /// [DeploymentItem("file1.xml")]
+    /// [DeploymentItem("file2.xml", "DataFiles")]
+    /// [DeploymentItem("bin\Debug")]
+    /// </example>
+    [AttributeUsage(AttributeTargets.Method | AttributeTargets.Class, AllowMultiple = true)]
+    public sealed class DeploymentItemAttribute : Attribute
```
codito, line 20: Should we not support it for .NETCore?

AbhitejJohn: Added the attribute in .Net Core as well. This would however not be visible in UWP since it does not even make sense there.

PR #123, opened 2017-03-23
Fixing an unhandled exception in the adapter

AbhitejJohn:
We do not catch AppDomainUnload issues. This is usually hit when a test spawns of a thread and does not clean it up.


src/Adapter/PlatformServices.Desktop/Utilities/AppDomainWrapper.cs
```diff
@@ -0,0 +1,24 @@
...
+
+namespace Microsoft.VisualStudio.TestPlatform.MSTestAdapter.PlatformServices
+{
+    using System;
+    using System.Security.Policy;
+
+    /// <summary>
+    /// Abstraction over the AppDomain APIs.
+    /// </summary>
+    internal class AppDomainWrapper : IAppDomain
```
codito, line 12: ~`TestAppDomain` is a better name.~

We're using wrappers at other places in code.

src/Adapter/PlatformServices.Desktop/Services/DesktopTestSourceHost.cs
```diff
@@ -209,10 +220,33 @@ public void Dispose()
...
+                    if (EqtTrace.IsErrorEnabled)
+                    {
+                        EqtTrace.Error("The app domain running tests could not be unloaded. Exception: {0}", exception);
+                    }
+
+                    if (this.frameworkHandle != null)
+                    {
+                        // Let the test platform know that it should tear down the test host process
+                        // since we we have issues in unloading appdomain. We do so to avoid any assembly locking issues.
+                        this.frameworkHandle.EnableShutdownAfterTestRun = true;
```
codito, line 241: Not sure if this is supported in new test platform.

src/Adapter/PlatformServices.Desktop/Services/DesktopTestSourceHost.cs
```diff
@@ -209,10 +220,33 @@ public void Dispose()
...
-            if (this.appDomain != null)
+            if (this.domain != null)
             {
-                AppDomain.Unload(this.appDomain);
-                this.appDomain = null;
+                try
+                {
+                    this.appDomain.Unload(this.domain);
+                }
+                catch (CannotUnloadAppDomainException exception)
```
codito: In this particular case, why was this exception thrown? Is the test written incorrectly?

What if a regular `Exception` is thrown in `AppDomain.Unload`? Do we want to continue execution?

src/Adapter/PlatformServices.Desktop/Utilities/AppDomainWrapper.cs
```diff
@@ -0,0 +1,24 @@
...
+    {
+        public AppDomain CreateDomain(string friendlyName, Evidence securityInfo, AppDomainSetup info)
+        {
+            return AppDomain.CreateDomain(friendlyName, securityInfo, info);
+        }
+
+        public void Unload(AppDomain appDomain)
```
codito, line 19: If testability is the only reason for this interface, it is probably better to use `Func`s for setup and teardown in the TestSourceHost.
`internal TestSourceHost(string sourceFileName, IRunSettings runSettings, IFrameworkHandle frameworkHandle, Func<x, y, z> createAppDomain, Func<AppDomain> unloadAppDomain)`

AppDomainWrapper and AppDomainUtilities are now two similar concepts in the code base.

src/Adapter/PlatformServices.Desktop/Services/DesktopTestSourceHost.cs
```diff
@@ -140,11 +151,11 @@ public object CreateInstanceForType(Type type, object[] args)
             var configFile = this.GetConfigFileForTestSource(this.sourceFileName);
             AppDomainUtilities.SetConfigurationFile(appDomainSetup, configFile);
 
-            this.appDomain = AppDomain.CreateDomain("TestSourceHost: Enumering assembly", null, appDomainSetup);
+            this.domain = this.appDomain.CreateDomain("TestSourceHost: Enumering assembly", null, appDomainSetup);
```
codito, line 154: s/Enumering/Enumerating/

test/UnitTests/PlatformServices.Desktop.Unit.Tests/Services/DesktopTestSourceHostTests.cs
```diff
@@ -130,6 +132,25 @@ public void CreateInstanceForTypeShouldSetNewDomainsAppBaseToAdaptersLocationFor
...
+            // Arrange
+            var frameworkHandle = new Mock<IFrameworkHandle>();
+            var testableAppDomain = new Mock<IAppDomain>();
+
+            testableAppDomain.Setup(ad => ad.CreateDomain(It.IsAny<string>(), It.IsAny<Evidence>(), It.IsAny<AppDomainSetup>())).Returns(AppDomain.CurrentDomain);
+            testableAppDomain.Setup(ad => ad.Unload(It.IsAny<AppDomain>())).Throws(new CannotUnloadAppDomainException());
+            var sourceHost = new TestSourceHost(typeof(DesktopTestSourceHostTests).Assembly.Location, null, frameworkHandle.Object, testableAppDomain.Object);
+
+            // Act
+            sourceHost.CreateInstanceForType(typeof(DesktopTestSourceHostTests), null);
```
codito: Not required in act.

src/Adapter/PlatformServices.Desktop/Services/DesktopTestSourceHost.cs
AbhitejJohn: Would have to work with TP V1.

src/Adapter/PlatformServices.Desktop/Utilities/AppDomainWrapper.cs
AbhitejJohn: Yes, the plan is to refactor appdomanutilities - not a very testable class.

src/Adapter/PlatformServices.Desktop/Services/DesktopTestSourceHost.cs
AbhitejJohn: There were tests that spawned off a thread and did not clean it up properly. msdn isn't very clear when a regular exception is thrown. This is actually the last stage of execution. Will go ahead and put in a catch(Exception) so that we do not rely on the test platform to handle these exceptions.

PR #130, opened 2017-03-30
Update ci script for public repo

jayaranigarg:


netci.groovy
```diff
@@ -32,7 +31,7 @@ def branch = GithubBranchName
         Utilities.setMachineAffinity(newJob, 'Windows_NT', 'latest-or-auto-dev15-internal')
```
mmitche: Remove the -internal here.

mmitche:
LGTM after fix to machine affinity.

Issue #23, opened 2016-11-09
Inheritance support for base classes that resides in different assemblies.


smadala:
## Description
> Try to have derive test class in different assembly than base test class.

## Steps to reproduce
> Add a base test class TestClassBase in assembly1.dll
> Add a derived test class TestClassDerived in assembly2.dll

## Expected behavior
> Test methods in TestClassDerived class should be discovered/run

## Actual behavior
> Test methods on TestClassDerived.cs class not discovered/run

## Environment
> Package version of MSTest - preview-1.1.4

ajryan:
This looks like a relatively simple issue for a newcomer to pick up. I'd like to help out.

It looks like [this TODO](https://github.com/Microsoft/testfx/blob/abd5ea3585657914490c83d5d2b4c9e732640a89/src/Adapter/MSTest.CoreAdapter/Discovery/TypeEnumerator.cs#L100) (lline 100 of testfx/src/Adapter/MSTest.CoreAdapter/Discovery/TypeEnumerator.cs) is the place to implement it.

Is [MsTestSettings](https://github.com/Microsoft/testfx/blob/abd5ea3585657914490c83d5d2b4c9e732640a89/src/Adapter/MSTest.CoreAdapter/MSTestSettings.cs) the right place to add the new setting? Would a good name be `BaseClassTestMethodDiscovery` with possible values `DeclaringAssembly` and `AnyAssembly`?

AbhitejJohn:
@ajryan: That would be wonderful.
>It looks like this TODO (lline 100 of testfx/src/Adapter/MSTest.CoreAdapter/Discovery/TypeEnumerator.cs) is the place to implement it.

Yes, that's exactly the place.

>Is MsTestSettings the right place to add the new setting? Would a good name be BaseClassTestMethodDiscovery with possible values DeclaringAssembly and AnyAssembly?

Right again. We can name the setting `EnableDerivedTypesFromOtherAssemblies` and it would be a bool. We are only interested in the declaring assembly.

ajryan:
Thanks for the quick response. I'm don't think your suggested name is correct, a "derived type" would be an inheriting class (which already works) but we're concerned with enabling test methods found in base/parent classes.

Maybe one of these?

* `EnableTestMethodsInBaseClassesFromOtherAssemblies`
* `EnableBaseClassTestMethodsFromOtherAssemblies`

AbhitejJohn:
[...]

PR #147, opened 2017-04-09
Allow discovering test methods from base class in different assembly

ajryan:
Fixes issue #23


test/UnitTests/MSTest.CoreAdapter.Unit.Tests/Discovery/TypeEnumeratorTests.cs
```diff
@@ -101,13 +103,12 @@ public void GetTestsShouldReturnBaseTestMethodsInSameAssembly()
 
             var tests = typeEnumerator.Enumerate(out this.warnings);
 
-            var methodCount = typeof(DummyDerivedTestClass).GetMethods(BindingFlags.Instance | BindingFlags.Public).Count(m => m.DeclaringType.Assembly == typeof(DummyDerivedTestClass).Assembly);
             Assert.IsNotNull(tests);
-            Assert.AreEqual(methodCount, tests.Count);
+            Assert.IsTrue(tests.Any(t => t.TestMethod.Name == "BaseTestMethod"));
```
AbhitejJohn: >Assert.IsTrue(tests.Any(t => t.TestMethod.Name == "BaseTestMethod")); [](start = 12, length = 69)

Do add a message on assertion failure so it is easy to identify why this test failed.

test/UnitTests/MSTest.CoreAdapter.Unit.Tests/Discovery/TypeEnumeratorTests.cs
```diff
@@ -121,6 +122,58 @@ public void GetTestsShouldNotReturnBaseTestMethodsFromAnotherAssembly()
...
+            mockRunContext.Setup(dc => dc.RunSettings).Returns(mockRunSettings.Object);
+            mockRunSettings.Setup(rs => rs.SettingsXml).Returns(runSettingxml);
+
+            MSTestSettings.PopulateSettings(mockRunContext.Object);
+            this.SetupTestClassAndTestMethods(isValidTestClass: true, isValidTestMethod: true);
+            TypeEnumerator typeEnumerator = this.GetTypeEnumeratorInstance(typeof(DummyDerivedFromRemoteTestClass), Assembly.GetExecutingAssembly().FullName);
+
+            var tests = typeEnumerator.Enumerate(out this.warnings);
+
+            Assert.IsTrue(tests.All(t => t.TestMethod.Name != "BaseTestMethod"));
```
AbhitejJohn: >t.TestMethod.Name != "BaseTestMethod" [](start = 41, length = 37)

add a contains check instead.

src/Adapter/MSTest.CoreAdapter/Discovery/TypeEnumerator.cs
```diff
@@ -97,8 +97,10 @@ internal Collection<UnitTestElement> GetTests(ICollection<string> warnings)
             // Test class is already valid. Verify methods.
             foreach (var method in this.type.GetRuntimeMethods())
             {
-                // Todo: Provide settings to allow users to pick up tests from other assemblies as well.
-                if (!method.DeclaringType.GetTypeInfo().Assembly.Equals(this.type.GetTypeInfo().Assembly))
+                var isMethodDeclaredInTestTypeAssembly = method.DeclaringType.GetTypeInfo().Assembly.Equals(this.type.GetTypeInfo().Assembly);
```
AbhitejJohn: >method.DeclaringType.GetTypeInfo().Assembly.Equals(this.type.GetTypeInfo().Assembly) [](start = 57, length = 84)

This seems to be causing issues in testability. How about putting this in a separate API in reflecthelper and mocking that out in your tests?

src/Adapter/MSTest.CoreAdapter/MSTestSettings.cs
```diff
@@ -238,6 +245,16 @@ private static MSTestSettings ToSettings(XmlReader reader)
                                 break;
                             }
 
+                        case "ENABLEBASECLASSTESTMETHODSFROMOTHERASSEMBLIES":
```
AbhitejJohn, line 248: >ENABLEBASECLASSTESTMETHODSFROMOTHERASSEMBLIES [](start = 30, length = 45)

can we add tests for this too please? you would find them under MSTest.CoreAdapter.Unit.Tests -> MSTestSettingsTests.cs

AbhitejJohn:
Thanks a bunch for getting this fixed quickly @ajryan. Had a few comments. Will push this in as soon as they are resolved.

AbhitejJohn:
@pvlakshm should this be turned on by default(searching for test methods in a base class defined in another assembly) given that this is a common scenario?

test/UnitTests/MSTest.CoreAdapter.Unit.Tests/Discovery/TypeEnumeratorTests.cs
ajryan: Will do for all test methods I've changed/added.

src/Adapter/MSTest.CoreAdapter/Discovery/TypeEnumerator.cs
ajryan: Great idea, will do.

src/Adapter/MSTest.CoreAdapter/MSTestSettings.cs
ajryan: Will do.

test/UnitTests/MSTest.CoreAdapter.Unit.Tests/Discovery/TypeEnumeratorTests.cs
ajryan: How would that work? tests.Contains would expect an instance of UnitTestElement and that doesn't implement equality members, nor does TestMethod...

ajryan:
Thanks for the quick review! I'll have a revision soon. Please see my question above about the Contains thing too...

ajryan:
This should be ready except for the comment about using Contains

AbhitejJohn:
@ajryan : I meant contains for the name "BaseTestMethod". Thinking again, I think we already do this in multiple other places. This shouldn't be a problem. Merging this change.
Thanks for the contribution :)
/cc @sbaid @pvlakshm

ajryan:
Awesome, thanks for reviewing and merging so quickly! :tada:

AbhitejJohn:
Pleasure. Thanks for the contribution. This was a long standing issue with around 96 votes on [uservoice](https://visualstudio.uservoice.com/forums/330519-team-services/suggestions/6030736-support-test-inheritance-for-base-classes-in-differ). Great impact 🎉

PR #150, opened 2017-04-13
Nuspec file change for NetCore Platform Service

jayaranigarg:


src/Package/MSTest.TestFramework.nuspec
```diff
@@ -27,40 +27,6 @@
     <tags>MSTest TestFramework Unittest MSTestV2 Microsoft Test Testing TDD Framework </tags>
   </metadata>
   <files>
-    <!--DotNet-->
-    <file src="MSTest.Core\Microsoft.VisualStudio.TestPlatform.TestFramework.dll" target="lib\dotnet" />
-    <file src="MSTest.Core\Microsoft.VisualStudio.TestPlatform.TestFramework.xml" target="lib\dotnet" />
```
AbhitejJohn, line 32: This should ideally be netstandard1.0. So extension writers can build against the lowest version. Please file an issue for this.

src/Package/MSTest.TestAdapter.Enu.nuspec
```diff
@@ -23,7 +23,8 @@
 	<dependencies>
 	    <group targetFramework="netstandard1.3">
 			<dependency id="System.Diagnostics.TextWriterTraceListener" version="4.3.0" />
-		</group>      
+			<dependency id="System.Xml.ReaderWriter" version="4.3.0" />
```
AbhitejJohn: This seems to be something we just stumbled upon at runtime. Should we just add in all the packages we are dependent upon in the dependencies section and just build against them instead?

AbhitejJohn:
LGTM.

PR #168, opened 2017-04-27
CollectSourceInformation flag

Shyam-Gupta:
Added logic to handle CollectSourceInformation flag. This flag gets set to false during test discovery from TestWindow code when test window has sufficient information to obtain source information itself using Roslyn APIs.


src/Adapter/MSTest.CoreAdapter/Discovery/UnitTestDiscoverer.cs
```diff
@@ -129,5 +136,26 @@ internal void SendTestCases(string source, IEnumerable<UnitTestElement> testElem
                 PlatformServiceProvider.Instance.FileOperations.DisposeNavigationSession(navigationSession);
             }
         }
+
+        private bool CollectSourceInformation(IRunSettings runSettings)
```
AbhitejJohn: This logic should be part of the [MSTestSettings](https://github.com/Microsoft/testfx/blob/master/src/Adapter/MSTest.CoreAdapter/MSTestSettings.cs) class which parses the runsettings already.

Issue #163, opened 2017-04-23
Test methods in a base class defined in a different assembly are not navigable in Test Explorer


AbhitejJohn:
> Pulled this out of #23.

## Description
After turning on `EnableBaseClassTestMethodsFromOtherAssemblies` fixed in #23, new tests defined in the base class start appearing in the Test Explorer. These tests show up under the External node in Test Explorer instead of the derived project. This is happening because the DIA source information logic is searching for these methods in the derived assembly and cannot find it.

## Steps to reproduce
Add the following runsettings
```
<RunSettings>    
  <MSTest> 
    <EnableBaseClassTestMethodsFromOtherAssemblies>true</EnableBaseClassTestMethodsFromOtherAssemblies> 
  </MSTest> 
</RunSettings>
```
1. Create two test projects - one with a `BaseTest` class and another with a `DerivedTest` class with the `DerivedTest` extending `BaseTest`.
2. Add test methods in both the classes.
3. Discover tests from the project containing `DerivedTest`.
4. Double click the `BaseTest` tests.

## Expected behavior
The user should be taken to the tests definition.

## Actual behavior
Nothing happens.

AbhitejJohn:
From @AbhitejJohn
> @ajryan , do you want to give a shot at fixing that? It is to do with the assembly full path we pass in here. If we pass in the (base)assembly where the test method is actually defined this might be fixed.

AbhitejJohn:
From @ajryan
> sure, I'll take a look and come back with questions

AbhitejJohn:
From @AbhitejJohn
> Here is little more context on how the 2nd part works: This code gets the line number and file name where a test is defined based on the source. We use DIA internally to populate this data(DIA in-turn works on pdb files). In our case, we need to be searching for the base test methods in the base assembly and not in the derived test assembly. After we make a change to set the appropriate source on a test element as I detailed above, we would also need to change this code to pass in the source from the test element so it gets the right data. Hope that makes sense.

AbhitejJohn:
From @ajryan
>Thanks for the pointers, this makes sense. I think TestMethod should carry a DeclaringAssemblyName parallel to the way it exposes DeclaringClassFullName. I'll thread it through from the discovery end and then make sure the navigator can locate the source correctly.

AbhitejJohn:
From @AbhitejJohn
[...]

Issue #164, opened 2017-04-23
Make discovering test methods from base classes defined in another assembly the default.


AbhitejJohn:
> Pulled out from #23.

## Description
We plugged in this support in the `1.1.17` version of the packages as part of #23. This should however be the default workflow even though this is a difference in behavior from MSTest V1. This should however be documented though.

AbhitejJohn:
From @AbhitejJohn
> Here are the top level discovery numbers I got from VS for a solution with 9 projects each having 10 test classes each with 100 tests:

>With the config enabled: 4.67 s. (average of 5 runs)
Without the config enabled: 4.76 s. (average of 5 runs)

>This is on a 16 Gb, 2 core machine.

>Although the numbers should ideally be reversed based on our theory above, the point is that this does not affect performance and can be turned on by default.

AbhitejJohn:
This is currently blocked by #163.

MikeChristensen:
Just checking on this one - We'd really love to see this get done!  Any estimate on when this will be fixed?

AbhitejJohn:
@MikeChristensen : This feature is already available in version 1.1.18 of the Framework/Adapter tuple. Its just not the default. You can turn it on by passing in the following runsettings:
```
<RunSettings>    
  <MSTest> 
    <EnableBaseClassTestMethodsFromOtherAssemblies>true</EnableBaseClassTestMethodsFromOtherAssemblies> 
  </MSTest> 
</RunSettings>
```

MikeChristensen:
I'm aware of this.  I'm asking when it will be turned on by default, without needing a runsettings file.

AbhitejJohn:
That is scheduled for the next release - End of next week is when we are hoping to get this through.

PR #174, opened 2017-05-03
Fixes navigation for base class methods declared in other assemblies

ajryan:
Fix for #163


src/Adapter/MSTest.CoreAdapter/Discovery/TypeEnumerator.cs
```diff
@@ -106,11 +106,19 @@ internal UnitTestElement GetTestFromMethod(MethodInfo method, ICollection<string
             var isAsync = ReflectHelper.MatchReturnType(method, typeof(Task));
 
             var testMethod = new TestMethod(method.Name, this.type.FullName, this.assemblyName, isAsync);
+
             if (!method.DeclaringType.FullName.Equals(this.type.FullName))
             {
                 testMethod.DeclaringClassFullName = method.DeclaringType.FullName;
             }
 
+            if (!this.reflectHelper.IsMethodDeclaredInSameAssemblyAsType(method, this.type))
```
AbhitejJohn: How about passing this in from the parent caller when it calls GetTestFromMethod? Wouldn't need to query for it again then.

src/Adapter/MSTest.CoreAdapter/Discovery/UnitTestDiscoverer.cs
```diff
@@ -80,15 +80,29 @@ internal UnitTestDiscoverer()
 
         internal void SendTestCases(string source, IEnumerable<UnitTestElement> testElements, ITestCaseDiscoverySink discoverySink)
         {
-            object navigationSession = null;
+            Dictionary<string, object> navigationSessions = new Dictionary<string, object>();
```
AbhitejJohn: Am loving this. 👍 Performance is one of our key pillars. Although we haven't heard much in terms of performance issues in mstest v2, we learnt the hard way that this could hit us hard down the line.

AbhitejJohn: nit: this can be a var instead.

src/Adapter/MSTest.CoreAdapter/ObjectModel/TestMethod.cs
```diff
@@ -51,6 +56,25 @@ public TestMethod(string name, string fullClassName, string assemblyName,  bool
 
         /// <summary>
         /// Gets or sets the declaring class full name. This will be used while getting navigation data.
+        /// This will be null if AssemblyName is same as DeclaringAssemblyName.
+        /// Reason to set to null in the above case is to minimise the transfer of data across appdomains and not have a perf hit.
```
AbhitejJohn, line 60: 👍

test/UnitTests/MSTest.CoreAdapter.Unit.Tests/Discovery/UnitTestDiscovererTests.cs
```diff
@@ -196,6 +201,29 @@ public void SendTestCasesShouldSendTestCasesWithNaigationDataWhenDeclaredClassFu
...
+
+            var testElements = new List<UnitTestElement> { test };
+
+            this.SetupNavigation(source, test, test.TestMethod.DeclaringClassFullName, test.TestMethod.Name);
+
+            // Act
+            this.unitTestDiscoverer.SendTestCases(source, testElements, this.mockTestCaseDiscoverySink.Object);
+
+            // Assert
+            this.mockTestCaseDiscoverySink.Verify(ds => ds.SendTestCase(It.Is<TestCase>(tc => VerifyTestCase(tc))), Times.Once);
```
AbhitejJohn: Strange. It looks like all methods in this class seem to be validating FullyQualifiedName instead of the navigationData(line number and fileName). Changing it to that assertion would fail this test though since the mock setup is done on `DummyAssembly.dll` and we query navigation session on `DA` instead. 
Its perfectly fine if we just fix this one test and I'l take an AI to fix up the rest.

test/UnitTests/MSTest.CoreAdapter.Unit.Tests/Discovery/TypeEnumeratorTests.cs
```diff
@@ -366,18 +366,37 @@ public void GetTestFromMethodShouldSetDeploymentItems()
             CollectionAssert.AreEqual(deploymentItems, testElement.DeploymentItems.ToArray());
         }
 
+        [TestMethod]
+        public void GetTestFromMethodShouldSetDeclaringAssemblyName()
+        {
+            this.SetupTestClassAndTestMethods(isValidTestClass: true, isValidTestMethod: true, isMethodFromSameAssembly: false);
+            TypeEnumerator typeEnumerator = this.GetTypeEnumeratorInstance(typeof(DummyTestClass), "DummyAssemblyName");
+            var methodInfo = typeof(DummyTestClass).GetMethod("MethodWithVoidReturnType");
+            var deploymentItems = new[] { new KeyValuePair<string, string>("C:\\temp", string.Empty) };
```
AbhitejJohn: nit: This does not seem to be used.

src/Adapter/PlatformServices.Shared/netstandard1.0/Services/ns10FileOperations.cs
```diff
@@ -48,6 +48,20 @@ public Assembly LoadAssembly(string assemblyName, bool isReflectionOnly)
         }
 
         /// <summary>
+        /// Gets the path to the .DLL of the assembly.
+        /// </summary>
+        /// <param name="assembly">The assembly.</param>
+        /// <returns>Path to the .DLL of the assembly.</returns>
+        public string GetAssemblyPath(Assembly assembly)
+        {
+            #if NETSTANDARD1_5
```
AbhitejJohn, line 57: @jayaranigarg: Do let us know if you see any concerns with this.

AbhitejJohn:
@ajryan : Apologies this took some time. We've been heads down with a few release activities.

ajryan:
Thanks for the thorough review! I'll take some time to address the issues soon.

src/Adapter/MSTest.CoreAdapter/Discovery/UnitTestDiscoverer.cs
ajryan: :)

src/Adapter/MSTest.CoreAdapter/Discovery/TypeEnumerator.cs
ajryan: Good idea. I was thinking since it's internal the method could be used outside `GetTests` but it's only called by unit tests, so OK.

src/Adapter/MSTest.CoreAdapter/Discovery/UnitTestDiscoverer.cs
ajryan: 👍

test/UnitTests/MSTest.CoreAdapter.Unit.Tests/Discovery/UnitTestDiscovererTests.cs
ajryan: Ah yeah I thought this was weird. Actually I didn't mean to leave that little `VerifyTestCase` method but had it for edit-and-continue convenience to find a better check. Will improve this test.

ajryan: Updated test is named `SendTestCasesShouldUseNaigationSessionForDeclaredAssemblyName` which I think is the only `SendTestCases` responsibility that needs to be tested at this layer.

test/UnitTests/MSTest.CoreAdapter.Unit.Tests/Discovery/TypeEnumeratorTests.cs
ajryan: Good catch

ajryan:
Last changes should resolve your comments. Sorry for the slow turnaround, and thanks again for welcoming my work!

AbhitejJohn:
No problem. Thanks again. This should just make #164 a simple flip of a switch now.

PR #179, opened 2017-05-09
fix #178 InstanceOfType on null value

### LINKED ISSUE #178: Assert.IsInstanceOf passes on value null
{
            Assert.IsInstanceOfType(null, typeof(Dummy));
        }`

LarsCelie:
Not sure if the tests provided are desired, so i can remove them if necessary.

### DIFF
--- a/test/UnitTests/MSTest.Core.Unit.Tests/Assertions/AssertTests.cs
+++ b/test/UnitTests/MSTest.Core.Unit.Tests/Assertions/AssertTests.cs
@@ -245,5 +245,37 @@ namespace UnitTestFramework.Tests
         }
 
         #endregion
+
+        #region InstanceOfType tests
+
+        [TestMethod]
+        public void InstanceOfTypeShouldFailWhenValueIsNull()
+        {
+            Action action = () => TestFrameworkV2.Assert.IsInstanceOfType(null, typeof(AssertTests));
+            TestFrameworkV2.Assert.ThrowsException<TestFrameworkV2.AssertFailedException>(action);
+        }
+
+        [TestMethod]
+        public void InstanceOfTypeShouldFailWhenTypeIsNull()
+        {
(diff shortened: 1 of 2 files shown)


test/UnitTests/MSTest.Core.Unit.Tests/Assertions/AssertTests.cs
```diff
@@ -245,5 +245,37 @@ public void ReplaceNullCharsShouldReplaceNullCharsInAString()
...
 
         #endregion
+
+        #region InstanceOfType tests
+
+        [TestMethod]
+        public void InstanceOfTypeShouldFailWhenValueIsNull()
+        {
+            Action action = () => TestFrameworkV2.Assert.IsInstanceOfType(null, typeof(AssertTests));
+            TestFrameworkV2.Assert.ThrowsException<TestFrameworkV2.AssertFailedException>(action);
```
AbhitejJohn: Can we use [ActionUtility.ActionShouldThrowExceptionOfType](https://github.com/Microsoft/testfx/blob/5f8e5a421aa6a7f64f3aa458aebb418c2fd90ce6/test/UnitTests/MSTest.CoreAdapter.TestUtilities/ActionUtility.cs#L26) instead please. Any issue with ThrowsException would leave these tests in an indeterminate state too. Hence the ask.

test/UnitTests/MSTest.Core.Unit.Tests/Assertions/AssertTests.cs
```diff
@@ -245,5 +245,37 @@ public void ReplaceNullCharsShouldReplaceNullCharsInAString()
...
+        }
+
+        [TestMethod]
+        public void InstanceOfTypeShouldFailWhenTypeIsNull()
```
AbhitejJohn: If possible can we add a positive test too please?

PR #204, opened 2017-06-21
Making EnableBaseClassTestMethodsFromOtherAssemblies as True by default

jayaranigarg:
#164


AbhitejJohn:
This also requires a doc addition to call this out as a delta compared to mstest v1 in testfx-docs.

PR #206, opened 2017-06-22
Proposed changes that could improve perf.

harshjain2:
Proposed changes that could improve perf.


AbhitejJohn:
@harshjain2 : Is this still required?

AbhitejJohn:
Spoke to @harshjain2 . This isn't needed anymore. This would be in the test platform now.

PR #208, opened 2017-06-27
Populated RunConfiguration Settings for RunAll.

harshjain2:
Populated RunConfiguration Settings for RunAll.


src/Adapter/MSTest.CoreAdapter/MSTestExecutor.cs
```diff
@@ -76,6 +76,7 @@ public void RunTests(IEnumerable<string> sources, IRunContext runContext, IFrame
 
             // Populate the runsettings.
             MSTestSettings.PopulateSettings(runContext);
+            RunConfigurationSettings.PopulateSettings(runContext);
```
AbhitejJohn, line 79: Do you mind adding a test please?

PR #211, opened 2017-06-28
Adapter to honour CollectSourceInformation Flag and not DesignMode

jayaranigarg:


src/Adapter/MSTest.CoreAdapter/RunConfigurationSettings.cs
```diff
@@ -29,7 +29,7 @@ public class RunConfigurationSettings
         /// </summary>
         public RunConfigurationSettings()
         {
```
AbhitejJohn, line 26: Was talking to @harshjain2 yesterday. Should this settings also be exposed through MSTestSettings something like `MSTestSettings.RunConfiguration.CollectSourceInformation`. We will then have a one stop please instead of searching in multiple places.

src/Adapter/MSTest.CoreAdapter/Discovery/UnitTestDiscoverer.cs
```diff
@@ -81,12 +81,12 @@ internal UnitTestDiscoverer()
 
         internal void SendTestCases(string source, IEnumerable<UnitTestElement> testElements, ITestCaseDiscoverySink discoverySink)
         {
-            var isDesignMode = RunConfigurationSettings.ConfigurationSettings.DesignMode;
+            var collectSourceInformation = RunConfigurationSettings.ConfigurationSettings.CollectSourceInformation;
```
AbhitejJohn: nit: `ShouldCollectSourceInformation`

Issue #217, opened 2017-06-30
Add mapping for TestOutcome.None to the UnitTestOutcome enum to achieve NotExecuted behaviour in VSTS. 


irguzhav:
## Description
> There is no way to create a custom TestMethodAttribute which will return the Not Executed state for the test. With what is available today one can only get Passed/Failed/Skipped. We want to closely monitor amount of Ignored tests as they “pollute” the codebase. So if we use the Skipped(Inconclusive) outcome we won’t be able to distinguish the Ignored tests from the tests which were not supposed to run based on the custom attribute.

AbhitejJohn:
More details from an offline conversation:
To achieve a Not Executed  state however we would need to update the mapping [here](https://github.com/Microsoft/testfx/blob/a0cfbcffe9f292064bf57d784eae9e2bc3ea2103/src/Adapter/MSTest.CoreAdapter/Extensions/TestResultExtensions.cs#L49) and [here](https://github.com/Microsoft/testfx/blob/abd5ea3585657914490c83d5d2b4c9e732640a89/src/Adapter/MSTest.CoreAdapter/Helpers/UnitTestOutcomeHelper.cs#L26) to get it to eventually return TestOutcome.None which would map to Not Executed in VSTS.

PR #219, opened 2017-07-05
Test outcome none mapping

irguzhav:
Fix for https://github.com/Microsoft/testfx/issues/217


jayaranigarg:
Thanks @irguzhav for fixing this. PR looks good to me. Please resolve the merge conflicts in your branch. Will push this in as soon as they are resolved.

src/TestFramework/MSTest.Core/UnitTestOutcome.cs
```diff
@@ -46,6 +46,11 @@ public enum UnitTestOutcome : int
         Aborted,
 
         /// <summary>
+        /// Test cannot be executed.
+        /// </summary>
+        NotRunnable,
```
AbhitejJohn: please add this to the end, so we do not break compat.

Issue #92, opened 2017-03-01
Multiple copies added for same test on running multiple times in IntelliTest


jayaranigarg:
## Description
Running IntelliTest multiple times, results in multiple copies of the same test being added to the test class with an incremental suffix in the name. This is an issue with MSTestV2 and works for MSTest.

## Steps to reproduce
1. Create a Class Library. 
2. Right CLick -> IntelliTest -> Create IntelliTest with MSTestV2 framework
3. RightClick-> IntelliTest -> Run IntelliTest multiple times

## Expected behavior
*.Test.g.cs should have one copy of all possible test cases.

## Actual behavior
*.Test.g.cs have multiple copies of the same test being added to the test class with an incremental suffix in the name

## Environment
VS2017 RC4

PR #220, opened 2017-07-06
Multiple copies getting added while running tests using Intellitest

jayaranigarg:
#92

### DIFF
--- a/WizardExtensions/MSTestv2IntelliTestExtension/MSTestv2TestFramework.cs
+++ b/WizardExtensions/MSTestv2IntelliTestExtension/MSTestv2TestFramework.cs
@@ -127,7 +127,7 @@ namespace MSTestv2IntelliTestExtension
             {
                 if (this.expectedExceptionAttribute == null)
                 {
-                    this.expectedExceptionAttribute = MSTestv2TestFrameworkMetadata.AttributeName("ExpectedExceptionAttribute");
+                    this.expectedExceptionAttribute = MSTestv2TestFrameworkMetadata.AttributeName("ExpectedException");
                 }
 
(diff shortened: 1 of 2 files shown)


WizardExtensions/MSTestv2IntelliTestExtension/MSTestv2TestFramework.cs
```diff
@@ -127,7 +127,7 @@ public override TypeName ExpectedExceptionAttribute
             {
                 if (this.expectedExceptionAttribute == null)
                 {
-                    this.expectedExceptionAttribute = MSTestv2TestFrameworkMetadata.AttributeName("ExpectedExceptionAttribute");
+                    this.expectedExceptionAttribute = MSTestv2TestFrameworkMetadata.AttributeName("ExpectedException");
```
AbhitejJohn, line 130: interesting, how is this working with MSTest v1 then?

jayaranigarg: [This](https://github.com/Microsoft/testfx/blob/master/WizardExtensions/MSTestv2IntelliTestExtension/MSTestv2TestFrameworkMetadata.cs#L46) AttributeName is now appending the "Attribute" keyword. In MSTestV1, it was not appending the "Attribute" keyword in AttributeName function and we had to append it in arguments itself.

AbhitejJohn: Interesting. Do validate through the other pex scenarios before merging. @harshjain2 can help with that.

Issue #183, opened 2017-05-15
Debug.Writeline() output message is not coming for netcoreapp


Faizan2304:
## Description
> Debug.Writeline() output message is not coming for netcoreapp.

## Steps to reproduce
1) dotnet new mstest
```xml
<Project Sdk="Microsoft.NET.Sdk">

  <PropertyGroup>
    <TargetFramework>netcoreapp1.1</TargetFramework>

    <IsPackable>false</IsPackable>
  </PropertyGroup>

  <ItemGroup>
    <PackageReference Include="Microsoft.NET.Test.Sdk" Version="15.3.0-preview-20170425-07" />
    <PackageReference Include="MSTest.TestAdapter" Version="1.1.17" />
    <PackageReference Include="MSTest.TestFramework" Version="1.1.17" />
  </ItemGroup>

  <ItemGroup>
    <Service Include="{82a7f48d-3b50-4b1e-b82e-3ada8210c358}" />
  </ItemGroup>

</Project>
```
2) Add following test method 
```csharp
        [TestMethod]
        public void TestMethod1()
        {
            System.Diagnostics.Debug.WriteLine("Hello World");
            System.Console.WriteLine("hurrey!!");
        }
```
3) dotnet restore
4) dotnet test --logger "Console;Verbosity=normal"

## Expected behavior
The output should be 
```
Starting test execution, please wait...
Passed   Mstest.UnitTest1.TestMethod1
Standard Output Messages:
 Faizan
 Ahmad


Debug Trace:
Hello World


Total tests: 1. Passed: 1. Failed: 0. Skipped: 0.
```

## Actual behavior
The output coming is 
```
Starting test execution, please wait...
Passed   Mstest.UnitTest1.TestMethod1
Standard Output Messages:
 Faizan
 Ahmad


Total tests: 1. Passed: 1. Failed: 0. Skipped: 0.
```

PR #222, opened 2017-07-07
Debug.WriteLine not working for netcore.

jayaranigarg:
System.Diagnostics.Debug.Listeners property have been removed from coreCLR. So listeners will no longer get debug events. Debug.WriteLine will only write to the debugger output and not to Listeners.
Refer [this](https://github.com/dotnet/corefx/issues/16596) for more details.


jayaranigarg:
#183

Issue #223, opened 2017-07-10
Assert class has public constructor


MikeChristensen:
## Description
It's very cool that Assert is no longer a static class, which allows us to add extension methods and access the singleton "Assert" instance using Assert.That.  However, is there a reason why the Assert constructor is public?  I'd expect private, or at least Internal.

## Steps to reproduce
var a = new Assert();

## Expected behavior
I'd expect this not to work.

## Actual behavior
Works.  But is there a reason why you would ever need/want to do this?

## Environment
Latest MSTestV2 on VS2015.

AbhitejJohn:
Thanks for filing this @MikeChristensen . This should be private and sealed.

AbhitejJohn:
Fixed.

PR #224, opened 2017-07-11
Assertion fix up.

AbhitejJohn:
Fix for #223.
Making all assert ctor's private and the classes sealed. We want users to extend from Assert.That instead and not by creating instances of the Assert class or extending from it.


pvlakshm:
Looks good.

PR #240, opened 2017-08-12
Fix where Unit Tests were failing on local machine.

khilan2goel:
The Unit Tests were failing on local machine due to static field dependency.Added that field in Test Cleanup.


test/UnitTests/MSTest.CoreAdapter.Unit.Tests/Discovery/UnitTestDiscovererTests.cs
```diff
@@ -58,6 +58,7 @@ public void Cleanup()
             this.test = null;
             this.testElements = null;
             PlatformServiceProvider.Instance = null;
```
AbhitejJohn, line 60: `MSTestSettings.Reset() is what you need here.`

khilangoel: Done.

Issue #231, opened 2017-08-03
Deployment pulls in symbols from configured symbol servers


AbhitejJohn:
## Description
Tests with `[DeploymentItem]s` when run pull in symbols for dependent assemblies from configured symbols servers(if they aren't on disk). We do this to get the line number information in the stack trace for failed tests. This however is a very costly operation.

The proposal here is to just deploy symbols that are on disk and nothing else.

AbhitejJohn:
We would have to fix this in MSTest V1 as well.

codito:
Impact: this leads to hangs (and potential timeouts) test execution in CI servers.

PR #242, opened 2017-08-14
Fix for deployment Item attribute where it was fetching pdb file from symbol store.

khilan2goel:
Fix for #231


src/Adapter/PlatformServices.Desktop/Utilities/FileUtility.cs
```diff
@@ -293,9 +293,15 @@ internal virtual string[] GetDirectoriesInADirectory(string directoryPath)
...
+            Debug.Assert(!string.IsNullOrEmpty(path), "path");
+
             try
             {
-                return DiaHelper.GetSymbolsFileName(path);
+                string pdbFile = Path.ChangeExtension(path, ".pdb");
+                if (File.Exists(pdbFile))
+                {
+                    return pdbFile;
+                }
```
AbhitejJohn: can you log if you find a pdb? Or is it logged already?

src/Adapter/PlatformServices.Desktop/Utilities/FileUtility.cs
```diff
@@ -293,9 +293,15 @@ internal virtual string[] GetDirectoriesInADirectory(string directoryPath)
         [System.Diagnostics.CodeAnalysis.SuppressMessage("Microsoft.Design", "CA1031:DoNotCatchGeneralExceptionTypes", Justification = "Requirement is to handle all kinds of user exceptions and message appropriately.")]
         private string GetSymbolsFileName(string path)
         {
+            Debug.Assert(!string.IsNullOrEmpty(path), "path");
+
```
AbhitejJohn: this would be a no-op in release bits. Throw an exception.

src/Adapter/PlatformServices.Desktop/Utilities/FileUtility.cs
```diff
@@ -293,9 +293,15 @@ internal virtual string[] GetDirectoriesInADirectory(string directoryPath)
...
             try
             {
-                return DiaHelper.GetSymbolsFileName(path);
+                string pdbFile = Path.ChangeExtension(path, ".pdb");
```
AbhitejJohn: From msdn, this would bomb if path has invalid chars: https://msdn.microsoft.com/en-us/library/system.io.path.changeextension(v=vs.110).aspx
Are we checking for them some place?

src/Adapter/PlatformServices.Desktop/Utilities/FileUtility.cs
```diff
@@ -293,9 +293,15 @@ internal virtual string[] GetDirectoriesInADirectory(string directoryPath)
...
+                if (File.Exists(pdbFile))
+                {
+                    return pdbFile;
+                }
             }
             catch (Exception ex)
```
AbhitejJohn: when could the changed code throw now?

src/Adapter/PlatformServices.Desktop/Deployment/DiaHelper.cs
```diff
@@ -1,625 +0,0 @@
...
-
-    using Microsoft.VisualStudio.TestPlatform.MSTestAdapter.PlatformServices.Deployment.Dia2Lib;
-
-    using Microsoft.VisualStudio.TestPlatform.ObjectModel;
-
-    /// <summary>
-    /// This class implements various utility functions for interaction with debug
-    /// information.
-    /// </summary>
-    internal static class DiaHelper
```
AbhitejJohn, line 18: thanks for deleting.

src/Adapter/PlatformServices.Desktop/Utilities/FileUtility.cs
khilangoel: When InvalidChars are there in File Path,then exception will be thrown. Checked locally it was cached in catch block

src/Adapter/PlatformServices.Desktop/Utilities/FileUtility.cs
khilangoel: I am not checking them as it will be cached in catch block.

AbhitejJohn: relying on a catch to take care of checks isn't the right paradigm. The fxcop suppression on top of the method itself is a red flag. W can remove the catch below and pre-pone checks if you think this would not cause any unforeseen errors.

src/Adapter/PlatformServices.Desktop/Utilities/FileUtility.cs
khilangoel: Will log it.

PR #257, opened 2017-08-29
Using dotnet vstest to run .net core tests instead of vstest.console.exe

khilan2goel:
Fix for CI build failing


scripts/test.ps1
```diff
@@ -162,13 +161,8 @@ function Run-Test([string[]] $testContainers, [string[]] $netCoreTestContainers)
...
 	if($netCoreTestContainers.Count -gt 0)
 	{	
-		if(!(Test-Path $tpv2VSTestPath))
-		{
-			Write-Error "Unable to find vstest.console.exe at $tpv2VSTestPath. Test aborted."
-		}
-		
-		Write-Verbose "$tpv2VSTestPath $netCoreTestContainers /framework:$TestFramework $additionalArguments /logger:trx"
-		& $tpv2VSTestPath $netCoreTestContainers /framework:$TestFramework $additionalArguments /logger:trx 
+		Write-Verbose "dotnet vstest $netCoreTestContainers /framework:$TestFramework $additionalArguments /logger:trx"
```
jayaranigarg: Add a check for dotnet.exe. Throw error if it doesn't.

scripts/test.ps1
```diff
@@ -161,14 +160,14 @@ function Run-Test([string[]] $testContainers, [string[]] $netCoreTestContainers)
...
+	    if(!(dotnet vstest))
+ 		{
+ 			Write-Error "Unable to find dotnet vstest. Test aborted."	
 		}
-		
-		Write-Verbose "$tpv2VSTestPath $netCoreTestContainers /framework:$TestFramework $additionalArguments /logger:trx"
-		& $tpv2VSTestPath $netCoreTestContainers /framework:$TestFramework $additionalArguments /logger:trx 
+
+		Write-Verbose "dotnet vstest $netCoreTestContainers /framework:$TestFramework $additionalArguments /logger:trx"
+		& dotnet vstest $netCoreTestContainers /framework:$TestFramework $additionalArguments /logger:trx
```
AbhitejJohn, line 170: I'm hoping `dotnet vstest` can take in multiple test containers for the future.

AbhitejJohn: Does this mean we would not be able to dogfood TPV2 since we always pick up the vstest package from what is installed on the box and not what we refer to?
@jayaranigarg: is that ok?

singhsarab: dotnet vstest is not different from vstest.console in any way. 
Rather I prefer dotnet vstest because it works across platforms.

PR #258, opened 2017-08-29
Adding support for DiaNavigation in UWP test adapter

mayankbansal018:
Testing:
Verifed that for UWP test running via Tpv1, no exception is thrown(Since Dia Session is not available there)

Once we take dependency on Tpv2 Obj Model, we can enable commented testcases


src/Adapter/PlatformServices.Universal/Services/UniversalFileOperations.cs
```diff
@@ -74,8 +74,21 @@ public bool DoesFileExist(string assemblyFileName)
...
         {
-            // Navigation session for UWP is handled by Desktop Adapter
-            return null;
+            var messageFormatOnException =
+                string.Join("MSTestDiscoverer:DiaSession: Could not create diaSession for source:", source, ". Reason:{0}");
+
+            try
+            {
+                Assembly objectModelAssembly = Assembly.Load(new AssemblyName("Microsoft.VisualStudio.TestPlatform.ObjectModel"));
+                Type diaType = objectModelAssembly.GetType("Microsoft.VisualStudio.TestPlatform.ObjectModel.DiaSession");
```
AbhitejJohn: can we cache?

src/Adapter/PlatformServices.Universal/Services/UniversalFileOperations.cs
```diff
@@ -91,6 +104,21 @@ public void GetNavigationData(object navigationSession, string className, string
             // Initiate values and bail out.
```
AbhitejJohn: nit: remove.

src/Adapter/PlatformServices.Universal/Services/UniversalFileOperations.cs
```diff
@@ -91,6 +104,21 @@ public void GetNavigationData(object navigationSession, string className, string
...
             fileName = null;
             minLineNumber = -1;
+
+            MethodInfo methodInfo = navigationSession.GetType().GetMethod("GetNavigationData");
```
AbhitejJohn: caching here too maybe?

test/UnitTests/PlatformServices.Universal.Unit.Tests/Services/UniversalFileOperationsTests.cs
```diff
@@ -117,6 +117,57 @@ public void GetNavigationDataShouldReturnNullFileName()
             Assert.AreEqual(-1, minLineNumber);
         }
 
+        // Enable these tests when we take dependency on TpV2 object model
+        // In Tpv1 UWP Object model these below methods are not defined.
```
AbhitejJohn, line 121: Would we ever do this given that we want to support UWP in 14.0?

AbhitejJohn: If not, how do we run these tests?

src/Adapter/PlatformServices.Universal/Services/UniversalFileOperations.cs
mayankbansal018: Not needed, since this should be invoked only once when someone is trying to create DiaSesion, & then hold the session object.
Further the session is dependent based on for what exe/dll it is opened for, so caching would be wrong.

test/UnitTests/PlatformServices.Universal.Unit.Tests/Services/UniversalFileOperationsTests.cs
mayankbansal018: we would move to Tpv2 object model, once we have made it default. Right now I do not want to go into problem of acquiring it.

src/Adapter/PlatformServices.Universal/Services/UniversalFileOperations.cs
AbhitejJohn: It gets called for each assembly. I meant the GetType reflection call. You can still go ahead and create a new object each time.

test/UnitTests/PlatformServices.Universal.Unit.Tests/Services/UniversalFileOperationsTests.cs
AbhitejJohn: hmm.. This is a debt we are carrying forward. Please file a bug to track this.

src/Adapter/PlatformServices.Universal/Services/UniversalFileOperations.cs
mayankbansal018: yes the type object I'll cache

src/Adapter/PlatformServices.Shared/netstandard1.0/Services/ns10FileOperations.cs
```diff
@@ -16,12 +16,6 @@ namespace Microsoft.VisualStudio.TestPlatform.MSTestAdapter.PlatformServices
...
     {
-        private static MethodInfo methodGetNavigationData;
-        private static PropertyInfo propertyFileName;
-        private static PropertyInfo propertyMinLineNumber;
-        private static Type typeDiaSession;
-        private static Type typeDiaNavigationData;
-
         /// <summary>
         /// Initializes static members of the <see cref="FileOperations"/> class.
         /// </summary>
```
AbhitejJohn: Should the initialization now move into the new operations class? Doesn't make much sense for this to be aware of diasession type anymore.

test/UnitTests/PlatformServices.Shared.Unit.Tests/netstandard1.0/ns10DiaSessionOperationsTests.cs
```diff
@@ -0,0 +1,274 @@
...
+
+    using System;
+    using System.IO;
+    using System.Reflection;
+    using Microsoft.VisualStudio.TestPlatform.MSTestAdapter.PlatformServices;
+    using MSTestAdapter.TestUtilities;
+#pragma warning disable SA1649 // SA1649FileNameMustMatchTypeName
+
+    [TestClass]
+    public class DiaSessionOperationsTests
```
AbhitejJohn, line 25: Thanks for getting these enabled. Can we close this issue now: https://github.com/Microsoft/testfx/issues/261

PR #276, opened 2017-09-28
Put prefix of [MSTest][Discovery][<source>] for warning messages during discovery

sbaid:
Put prefix of [MSTest][Discovery][<source>] for warning messages during discovery

Warning message as reported by the user https://developercommunity.visualstudio.com/content/problem/114611/visual-studio-20173-unit-test-do-not-appear-random.html -- does not have right contextual info.

### DIFF
--- a/src/Adapter/MSTest.CoreAdapter/Discovery/UnitTestDiscoverer.cs
+++ b/src/Adapter/MSTest.CoreAdapter/Discovery/UnitTestDiscoverer.cs
@@ -4,6 +4,7 @@
 namespace Microsoft.VisualStudio.TestPlatform.MSTest.TestAdapter
 {
     using System.Collections.Generic;
+    using System.Globalization;
     using Microsoft.VisualStudio.TestPlatform.MSTest.TestAdapter.Discovery;
     using Microsoft.VisualStudio.TestPlatform.MSTest.TestAdapter.Helpers;
     using Microsoft.VisualStudio.TestPlatform.MSTest.TestAdapter.ObjectModel;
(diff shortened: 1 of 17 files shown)


src/Adapter/MSTest.CoreAdapter/Resources/Resource.resx
```diff
@@ -296,4 +296,7 @@
   <data name="UTA_CleanupStackTrace" xml:space="preserve">
     <value>TestCleanup Stack Trace</value>
   </data>
+  <data name="DiscoveryWarning" xml:space="preserve">
```
AbhitejJohn, line 299: Thanks Sushil. Could you run build with -uxlf so the corresponding xlf files are updated for localization.

src/Adapter/MSTest.CoreAdapter/Discovery/UnitTestDiscoverer.cs
```diff
@@ -69,7 +70,8 @@ internal UnitTestDiscoverer()
                     "MSTestDiscoverer: Warning during discovery from {0}. {1} ",
                     source,
                     warning);
-                logger.SendMessage(TestMessageLevel.Warning, warning);
+                var message = string.Format(CultureInfo.CurrentCulture, Resource.DiscoveryWarning, source, warning);
```
AbhitejJohn, line 73: Could we do the same for execution too please? :)

src/Adapter/MSTest.CoreAdapter/Resources/Resource.resx
sbaid: updated. pl check

src/Adapter/MSTest.CoreAdapter/Discovery/UnitTestDiscoverer.cs
sbaid: will consider in separate PR.

src/Adapter/MSTest.CoreAdapter/Resources/Resource.resx
AbhitejJohn: LGTM.

src/Adapter/MSTest.CoreAdapter/Discovery/UnitTestDiscoverer.cs
AbhitejJohn: sure. Please go ahead.

Issue #249, opened 2017-08-22
Async tests which use Assert.Inconclusive report as Error


Terebi42:
## Description
Async tests which use Assert.Inconclusive report as Error

## Steps to reproduce
```
   [TestMethod]
        public void TestInconclusiveSync()
        {

            Assert.Inconclusive("This will report as skipped");
        }

        [TestMethod]
        public async Task TestInconclusiveAsync()
        {

            Assert.Inconclusive("This will report as error");
        }
```

v1.1.18

AbhitejJohn:
Thanks for filing this. I can repro this. I guess this seems to be happening because the error is thrown on another thread which gets wrapped in as an internal exception we do not check.

Terebi42:
This behavior is seen in several places, VS, TFS build, Resharper, etc.  Do you think all of those would be resolved by this bug (because they are all relying on something in the nuget), or should I be filing  bugs with those products? (because they will have to do this same exception peeking on their own)?

AbhitejJohn:
@Terebi42 : Fixing this bug should take care of all the client scenarios.

abatishchev:
I bugged the VSTS team with this just to learn down the road that it's unfortunately a bug in MSTest indeed. Any plans to address, and asap? It's 100% reproducible.

AbhitejJohn:
@abatishchev: We have been busy with a few other things and this one is currently parked. I can try and help anyone from the community who is willing to take a stab at fixing this though. The code to invoke the test method is [here](https://github.com/Microsoft/testfx/blob/master/src/Adapter/MSTest.CoreAdapter/Execution/TestMethodInfo.cs#L207)

jessehouwing:
I tried debugging this, which wasn't easy with all of the different test frameworks being loaded at the same time (sort of test framework inception). I think I may have found the issue where the Exception being thrown is from the MstestV2 test framework, but the exception being compared against is from the V1 framework. I am not a 100% sure, as my knowledge of the framework is limited.

I have forked and committed my changes here: 

https://github.com/jessehouwing/testfx

AbhitejJohn:
@Jessehouwing: Thanks. The [change](https://github.com/jessehouwing/testfx/commit/be3187262daeb822f7b1fd99427be457a2ec943f#diff-25d902c24283ab8cfbac54dfa101ad31) looks good to me. Can you please submit a PR? 
[...]

PR #277, opened 2017-09-28
Improve handling of Assert.Inconclusive

jessehouwing:
To help solve issues found in #249

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/testfx/277)

### DIFF
--- a/src/Adapter/MSTest.CoreAdapter/Execution/TestMethodInfo.cs
+++ b/src/Adapter/MSTest.CoreAdapter/Execution/TestMethodInfo.cs
@@ -231,7 +231,7 @@ namespace Microsoft.VisualStudio.TestPlatform.MSTest.TestAdapter.Execution
 
                     if (result.Outcome != TestTools.UnitTesting.UnitTestOutcome.Passed)
                     {
-                        if (ex.InnerException != null && ex.InnerException is AssertInconclusiveException)
+                        if (ex is UTF.AssertInconclusiveException || ex.InnerException is UTF.AssertInconclusiveException)
                         {
                             result.Outcome = TestTools.UnitTesting.UnitTestOutcome.Inconclusive;
                         }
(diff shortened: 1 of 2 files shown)


src/Adapter/MSTest.CoreAdapter/Execution/TestMethodInfo.cs
```diff
@@ -231,7 +231,7 @@ private TestResult ExecuteInternal(object[] arguments)
 
                     if (result.Outcome != TestTools.UnitTesting.UnitTestOutcome.Passed)
                     {
-                        if (ex.InnerException != null && ex.InnerException is AssertInconclusiveException)
+                        if (ex is UTF.AssertInconclusiveException || ex.InnerException is UTF.AssertInconclusiveException)
```
AbhitejJohn, line 234: can we please add a null check for `ex.InnerException`?

AbhitejJohn:
Thanks for helping out.

test/UnitTests/MSTest.CoreAdapter.Unit.Tests/Execution/TestMethodInfoTests.cs
```diff
@@ -131,6 +131,70 @@ public void TestMethodInfoInvokeShouldWaitForAsyncTestMethodsToComplete()
...
+            var result = method.Invoke(null);
+
+            Assert.AreEqual(UTF.UnitTestOutcome.Inconclusive, result.Outcome);
+        }
+
+        [TestMethodV1]
+        public void TestMethodInfoInvokeShouldHandleThrowAssertInconclusive()
+        {
+            DummyTestClass.TestMethodBody = (d) => { throw new UTF.AssertInconclusiveException(); };
+            var asyncMethodInfo = typeof(DummyTestClass).GetMethod("DummyAsyncTestMethod");
```
jayaranigarg: 1. We would want to do GetMethod on "DummyTestMethod" here..
2. Should we just name the variable as "methodInfo" rather than "asyncMethodInfo" ?

src/Adapter/MSTest.CoreAdapter/Execution/TestMethodInfo.cs
jessehouwing: That should not be needed, as `is` will return false when compared to null. If for clarity or performance reasons, I can add it back in. The expression became harder to read and the effect is the same.

test/UnitTests/MSTest.CoreAdapter.Unit.Tests/Execution/TestMethodInfoTests.cs
jessehouwing: Yes and almost. methodInfo was a class member, so I've made it "dummyMethodInfo" instead.

pvlakshm:
Jesse, great to see this; thank you for contributing!

src/Adapter/MSTest.CoreAdapter/Execution/TestMethodInfo.cs
AbhitejJohn: That's fair. Don't think there would be much of a perf hit.

Issue #295, opened 2017-10-20
An exception occurred while invoking executor 'executor://mstestadapter/v2': Could not load file or assembly 'System.IO, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a'


northtyphoon:
## Description
> The test output includes "System.IO 4.1.2.0"  and the following binding redirect in TestAssembly.dll.config

    <assemblyBinding xmlns="urn:schemas-microsoft-com:asm.v1">
      <dependentAssembly>
        <assemblyIdentity name="System.IO" publicKeyToken="b03f5f7f11d50a3a" culture="neutral" />
        <bindingRedirect oldVersion="0.0.0.0-4.1.2.0" newVersion="4.1.2.0" />
      </dependentAssembly>
    </assemblyBinding>

The test passed in Visual Studio. However, the test failed in the build machine as the following error. 

An exception occurred while invoking executor 'executor://mstestadapter/v2': Could not load file or assembly 'System.IO, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a' or one of its dependencies

I also notice if I manually removed TestAssembly.dll.config from the output folder, the test started to pass in the build machine.

## Steps to reproduce
> In VSTS build, run the latest Visual Studio Test task.

## Expected behavior
> Test pass.

## Actual behavior
> 2017-10-19T23:47:12.3630779Z Starting test execution, please wait...
2017-10-19T23:47:19.4800462Z Error: An exception occurred while invoking executor 'executor://mstestadapter/v2': Could not load file or assembly 'System.IO, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a' or one of its dependencies. The system cannot find the file specified.
2017-10-19T23:47:19.4800462Z 

## Environment
> Operating system : WS2016 
VS2017 15.4.1
Visual Studio test runner 2.1.15

Package version of MSTest: 
"MSTest.TestAdapter" Version="1.2.0" 
"MSTest.TestFramework" Version="1.2.0"

kant2002:
If this is just some binding redirect issues, it will greatly benefit from increased logging with information which test assembly could not be loaded, since that alone greatly speedup troubleshooting. I hit similar issue in the past where binding redirects was required by the test.

Izzmo:
@kant2002 

>it will greatly benefit from increased logging with information which test assembly could not be loaded

[...]

PR #299, opened 2017-10-25
Add information about which assembly failed to discover test into

kant2002:
Having indicate location of the assembly from which discovery happens allow developer reason about possible failures, and manually troubleshoot his configuration issues which cause asembly to fail

This allow simplify resolution of the issues like #295 and similar where it is not clear what exactly happens. For example when you run tests from multiple assemblies, like in VSTS CI build

### DIFF
--- a/src/Adapter/MSTest.CoreAdapter/Discovery/AssemblyEnumeratorWrapper.cs
+++ b/src/Adapter/MSTest.CoreAdapter/Discovery/AssemblyEnumeratorWrapper.cs
@@ -144,11 +144,13 @@ namespace Microsoft.VisualStudio.TestPlatform.MSTest.TestAdapter.Discovery
                     "MSTestDiscoverer.TryGetTests: Failed to discover tests from {0}. Reason:{1}",
                     assemblyFileName,
                     ex);
-                var message = string.Format(
-                    CultureInfo.CurrentCulture,
-                    Resource.TestAssembly_AssemblyDiscoveryFailure,
-                    fullFilePath,
-                    ex.Message);
+                var message = ex is FileNotFoundException fileNotFoundEx
+                    ? fileNotFoundEx.Message
+                    : string.Format(
+                        CultureInfo.CurrentCulture,
+                        Resource.TestAssembly_AssemblyDiscoveryFailure,
+                        fullFilePath,
+                        ex.Message);
                 warnings.Add(message);
                 return null;
             }
--- a/src/Adapter/MSTest.CoreAdapter/Discovery/AssemblyEnumeratorWrapper.cs
+++ b/src/Adapter/MSTest.CoreAdapter/Discovery/AssemblyEnumeratorWrapper.cs
@@ -90,8 +90,8 @@ namespace Microsoft.VisualStudio.TestPlatform.MSTest.TestAdapter.Discovery
                 {
                     var winrtFailureMessage = string.Format(
                         CultureInfo.CurrentCulture,
-                        Resource.TestAssembly_WinRTAssemblyDiscoveryFailure,
-                        fullFilePath,
+                        Resource.TestAssembly_AssemblyDiscoveryFailure,
+                        assemblyFileName,
                         e.Message);
                     warnings.Add(winrtFailureMessage);
                     return null;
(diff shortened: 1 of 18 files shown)


src/Adapter/MSTest.CoreAdapter/Discovery/AssemblyEnumeratorWrapper.cs
```diff
@@ -88,16 +88,31 @@ internal class AssemblyEnumeratorWrapper
                 }
                 catch (Exception e)
                 {
-                    warnings.Add(e.Message);
+                    var winrtFailureMessage = string.Format(
+                        CultureInfo.CurrentCulture,
+                        Resource.TestAssembly_WinRTAssemblyDiscoveryFailure,
+                        fullFilePath,
```
AbhitejJohn: This should be `assemblyFullName`.

src/Adapter/MSTest.CoreAdapter/Resources/Resource.resx
```diff
@@ -299,4 +299,10 @@
   <data name="DiscoveryWarning" xml:space="preserve">
     <value>[MSTest][Discovery][{0}] {1}</value>
   </data>
+  <data name="TestAssembly_AssemblyDiscoveryFailure" xml:space="preserve">
+    <value>Failed to discover tests from assembly located at {0}. Reason:{1}</value>
```
AbhitejJohn: Suggestion: Can we only have one string with a 
```
Failed to discover tests from assembly {0}. Reason:{1}
```

src/Adapter/MSTest.CoreAdapter/Discovery/AssemblyEnumeratorWrapper.cs
```diff
@@ -129,6 +144,11 @@ internal class AssemblyEnumeratorWrapper
                     "MSTestDiscoverer.TryGetTests: Failed to discover tests from {0}. Reason:{1}",
                     assemblyFileName,
                     ex);
+                var message = string.Format(
+                    CultureInfo.CurrentCulture,
+                    Resource.TestAssembly_AssemblyDiscoveryFailure,
+                    fullFilePath,
+                    ex.Message);
                 warnings.Add(ex.Message);
```
jayaranigarg: This should be warnings.Add(message)

AbhitejJohn:
@kant2002 : There appears to be a unit test failing - `GetTestsShouldReturnNullIfSourceFileDoesNotExistInContext` from [here](https://ci.dot.net/job/Microsoft_testfx/job/master/job/Debug_prtest/182/console). Can you take a look please? 
The validation scripts seem to be having an issue which needs fixing.

kant2002:
Ooh, I forgot to run tests, since was sidetracked with build issues. Will take a look.

kant2002:
@AbhitejJohn I actually blocked with #300 without ability to do clean build, so when I build solution the failed test does not run, since assembly where it is probably missing. Could you take a look what could I install to make build works.

jayaranigarg:
@kant2002 : We have fixed the nuget package restore issue [here](https://github.com/Microsoft/testfx/pull/331). You can now go ahead and update this PR.

jayaranigarg:
@kant2002 : Following up on this. Can you please do the needful to complete this PR?

kant2002:
@jayaranigarg thanks for your patience, I have some issues with my VS setup after 15.4 update probably which prevent me from do clean build.
```
D:\Program Files (x86)\Microsoft Visual Studio\2017\Community\MSBuild\Microsoft\WindowsXaml\v15.0\8.2\Microsoft.Windows.UI.Xaml.Common.targets(676,5): warning : Could not find target platform winmds for the SDK specified by [Windows, 10.0, UAP, 10.0.10240.0, 10.0.14393.0] [xxxx\testfx\src\TestFramework\Extension.UWP\Extension.UWP.csproj]
```

I try to reinstall VS now and check patch again. I will report back when I finish with VS install

kant2002:
@jayaranigarg or @AbhitejJohn please take a look

jayaranigarg:
@kant2002 : Thank you for completing this. Looks good to me. Will push this in.

PR #302, opened 2017-10-26
Safely Invoke methods which are called via reflection

mayankbansal018:
The exception is thrown in UWP release with .net native compilation being enabled.
The exception is "Arg_NotImplementedException" which means that the request method was not implemented. https://github.com/dotnet/corert/blob/35f68948d5ef585b67f820efd5936091f67b5418/src/System.Private.CoreLib/src/Resources/Strings.resx#L369

### DIFF
--- a/src/Adapter/MSTest.CoreAdapter/TestMethodFilter.cs
+++ b/src/Adapter/MSTest.CoreAdapter/TestMethodFilter.cs
@@ -96,6 +96,19 @@ namespace Microsoft.VisualStudio.TestPlatform.MSTest.TestAdapter
             return null;
         }
 
+        private static object SafeInvoke<T>(Func<T> action)
+        {
+            try
+            {
+                return action.Invoke();
+            }
+            catch (Exception)
+            {
+            }
+
+            return null;
+        }
+
         /// <summary>
         /// Gets filter expression from run context.
         /// </summary>
@@ -117,7 +130,7 @@ namespace Microsoft.VisualStudio.TestPlatform.MSTest.TestAdapter
             {
                 // GetTestCaseFilter is present in DiscoveryContext but not in IDiscoveryContext interface.
                 MethodInfo methodGetTestCaseFilter = context.GetType().GetRuntimeMethod("GetTestCaseFilter", new[] { typeof(IEnumerable<string>), typeof(Func<string, TestProperty>) });
-                return (ITestCaseFilterExpression)methodGetTestCaseFilter?.Invoke(context, new object[] { this.supportedProperties.Keys, (Func<string, TestProperty>)this.PropertyProvider });
+                return SafeInvoke(() => methodGetTestCaseFilter?.Invoke(context, new object[] { this.supportedProperties.Keys, (Func<string, TestProperty>)this.PropertyProvider })) as ITestCaseFilterExpression;
             }
             catch (TargetInvocationException ex)
             {
--- a/src/Adapter/MSTest.CoreAdapter/TestMethodFilter.cs
+++ b/src/Adapter/MSTest.CoreAdapter/TestMethodFilter.cs
@@ -46,7 +46,7 @@ namespace Microsoft.VisualStudio.TestPlatform.MSTest.TestAdapter
             {
                 try
                 {
-                    filter = (context is IRunContext) ? this.GetTestCaseFilterFromRunContext(context as IRunContext) : this.GetTestCaseFilterFromDiscoveryContext(context);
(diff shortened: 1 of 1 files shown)


src/Adapter/MSTest.CoreAdapter/TestMethodFilter.cs
```diff
@@ -96,6 +96,19 @@ internal object PropertyValueProvider(TestCase currentTest, string propertyName)
...
 
+        private static object SafeInvoke<T>(Func<T> action)
+        {
+            try
+            {
+                return action.Invoke();
+            }
+            catch (Exception)
+            {
+            }
```
AbhitejJohn: Needs logging.

src/Adapter/MSTest.CoreAdapter/TestMethodFilter.cs
```diff
@@ -117,7 +130,7 @@ private ITestCaseFilterExpression GetTestCaseFilterFromDiscoveryContext(IDiscove
             {
                 // GetTestCaseFilter is present in DiscoveryContext but not in IDiscoveryContext interface.
                 MethodInfo methodGetTestCaseFilter = context.GetType().GetRuntimeMethod("GetTestCaseFilter", new[] { typeof(IEnumerable<string>), typeof(Func<string, TestProperty>) });
-                return (ITestCaseFilterExpression)methodGetTestCaseFilter?.Invoke(context, new object[] { this.supportedProperties.Keys, (Func<string, TestProperty>)this.PropertyProvider });
+                return SafeInvoke(() => methodGetTestCaseFilter?.Invoke(context, new object[] { this.supportedProperties.Keys, (Func<string, TestProperty>)this.PropertyProvider })) as ITestCaseFilterExpression;
             }
             catch (TargetInvocationException ex)
```
AbhitejJohn: Can we make this a catch(Exception) and log and bail out instead?
What is the scenario and the exception being thrown here?

codito: It is difficult to review a change without scenario. Please add tests.

mayankbansal018: The exception is thrown in UWP release with .net native compilation being enabled.
The exception is "Arg_NotImplementedException" which means that the request method was not implemented. https://github.com/dotnet/corert/blob/35f68948d5ef585b67f820efd5936091f67b5418/src/System.Private.CoreLib/src/Resources/Strings.resx#L369

I cannot log everything in exception, since that would set "filterHasError" to true & would fail the discovery.

I don't know the root cause of it yet, but I'll drop a mail to UWP team asking why this could happen. I don't even know which method it is saying it could not find. The one being invoked, or something being called from inside method being invoked.

I've described the scenario, but I really do not know how to add test for this scenario

src/Adapter/MSTest.CoreAdapter/TestMethodFilter.cs
mayankbansal018: added logger, though the warning would be strange

src/Adapter/MSTest.CoreAdapter/TestMethodFilter.cs
```diff
@@ -110,14 +124,15 @@ private ITestCaseFilterExpression GetTestCaseFilterFromRunContext(IRunContext co
...
         /// <returns>Filter expression.</returns>
-        private ITestCaseFilterExpression GetTestCaseFilterFromDiscoveryContext(IDiscoveryContext context)
+        private ITestCaseFilterExpression GetTestCaseFilterFromDiscoveryContext(IDiscoveryContext context, IMessageLogger logger)
         {
             try
             {
                 // GetTestCaseFilter is present in DiscoveryContext but not in IDiscoveryContext interface.
                 MethodInfo methodGetTestCaseFilter = context.GetType().GetRuntimeMethod("GetTestCaseFilter", new[] { typeof(IEnumerable<string>), typeof(Func<string, TestProperty>) });
-                return (ITestCaseFilterExpression)methodGetTestCaseFilter?.Invoke(context, new object[] { this.supportedProperties.Keys, (Func<string, TestProperty>)this.PropertyProvider });
+                return SafeInvoke(() => methodGetTestCaseFilter?.Invoke(context, new object[] { this.supportedProperties.Keys, (Func<string, TestProperty>)this.PropertyProvider }), logger) as ITestCaseFilterExpression;
```
codito, line 135: Suggest to add a link to the UWP issue here.

mayankbansal018:
@dotnet-bot test this please

Issue #255, opened 2017-08-29
TestContext.TestName having a value in Class/Assembly Init is confusing.


alextolp:
## Description
TestContext.TestName may report the wrong value during test execution. It appears to cache the value from the first test that runs and doesn't update the value for subsequent tests.

## Steps to reproduce
    [TestClass]
    public class Tests
    {
        private static TestContext testContext;

        private string testName;

        [TestMethod]
        public void Test1()
        {
            Assert.AreEqual(nameof(this.Test1), this.testName);
        }

        [TestMethod]
        public void Test2()
        {
            Assert.AreEqual(nameof(this.Test2), this.testName);
        }

        [TestInitialize]
        public void TestInit()
        {
            this.testName = testContext.TestName;
        }

        [ClassInitialize]
        public static void ClassInit(TestContext context)
        {
            testContext = context;
        }
    }

## Expected behavior
Tests should pass.

## Actual behavior
Test2 fails with: 

Message: Assert.AreEqual failed. Expected:Test2. Actual:Test1. 

note: this is a regression from earlier MSTest versions.

## Environment
VS version: 15.3.1
Framework version: net461
MSTest.TestAdapter version: 1.1.18
MSTest.TestFramework version: 1.1.18

alextolp:
[MSTest.zip](https://github.com/Microsoft/testfx/files/1259032/MSTest.zip)

AbhitejJohn:
Thanks for filing this @alextolp. The code does say we are re-setting the TestContext for every method. Need to re-check what's going wrong here.

AbhitejJohn:
We expect test authors to declare a `public TestContext TestContext {get; set;}` for scenarios like this, so the framework resets the value of that property before running a test. A ClassInitialize on the other hand is called only once for all methods in that class, hence the TestContext passed in there would only end up being useful for session/class level data. Here is the test code that would work for your scenario:

```
[TestClass]
public class Tests
{
    public TestContext TestContext { get; set; }

    private string testName;

    [TestMethod]
    public void Test1()
    {
        Assert.AreEqual(nameof(this.Test1), this.testName);
    }

    [TestMethod]
    public void Test2()
    {
        Assert.AreEqual(nameof(this.Test2), this.testName);
    }

    [TestInitialize]
    public void TestInit()
    {
        this.testName = TestContext.TestName;
    }

    [ClassInitialize]
    public static void ClassInit(TestContext context)
    {
[...]

PR #306, opened 2017-10-29
Making TestContext property name more generic

AbhitejJohn:
Picking up any property that is of a TestContext type instead. Reduces the confusion users run into when declaring a TestContext property.
In part stemmed out of #255 and a few off band reports.


AbhitejJohn:
Not taking this because having a generic name for TestContext could result in issues when multiple TestContext properties are defined - Having a single case sensitive name as we have today eliminates that confusion and makes the definition more concrete. While this PR make things less stricter we believe that documentation of the current behavior would be a much better place to be in.

PR #301, opened 2017-10-26
Fix paths to xlftool to make it work with VS 15.4

kant2002:
Since I don't closely monitor changes in the NuGet behavior I could not say is this a settled fix, or it will not cover all cases and both ways should be supported.
Without that change I receive errors when run `Build.cmd`

PR #310, opened 2017-11-09
Add missing Microsoft.Internal.TestPlatform.ObjectModel

kant2002:
This is part of #301 which could be accepted according to review. 
Since I still want to be able build project on my computer, I probably try to find way build project on my computer if needed and find way which will work both on your CI and on my PC.


jayaranigarg:
Thank you!

PR #322, opened 2017-11-29
LOC CHECKIN | testfx - master | 20171129

v-zbsail:


src/Adapter/MSTest.CoreAdapter/Resources/xlf/Resource.cs.xlf
```diff
@@ -5,306 +5,306 @@
...
       </trans-unit>
       <trans-unit id="UTA_CleanupStackTrace">
         <source>TestCleanup Stack Trace</source>
-        <target state="new">TestCleanup Stack Trace</target>
-        <note></note>
+        <target state="translated">Trasování zásobníku čištění testu</target>
+        <note />
       </trans-unit>
       <trans-unit id="DiscoveryWarning">
         <source>[MSTest][Discovery][{0}] {1}</source>
```
singhsarab, line 305: @jayaranigarg Any idea why is this in resx if we don't expect it to get localized ?

jayaranigarg: @v-zbsail : Can you check why this string is not translated, even though it is marked as translated?

janbaltus: Do you refer to "TestCleanup Stack Trace"?

jayaranigarg: @janbaltus : I meant "[MSTest][Discovery][{0}] {1}" string..

janbaltus: Is this string to be translated?
Can you please add more context?

jayaranigarg: Yes, I believe it should have been translated like the rest of the strings.
We plan to remove [MSTest] from this string though as we do not want MSTest to be translated but would want [Discovery] to be translated. However, the change of removing [MSTest] will be done in a separate PR.

janbaltus: Ince you update the xlf we will fix this in next round of localization.
Please remove also Discovery or add note for that string explaining this is localizable.
Any string in [] {} look like placeholder from our point of view

jayaranigarg: Thanks @janbaltus . Will do those changes later. Meanwhile I will push this PR in.

jayaranigarg:
@dotnet-bot test Windows / Debug Build  please

jayaranigarg:
@dotnet-bot test this please

Issue #311, opened 2017-11-13
Build.cmd not restoring nuget packages as expected


jayaranigarg:
## Description
Pulling this issue out of [pull request](https://github.com/Microsoft/testfx/pull/301)  where nuget packages are not restoring appropriately.

jayaranigarg:
@kant2002 : Do you have any user setting for nugget that might be impacting this?

kant2002:
@jayaranigarg Right now I have VS install in broken state, (error with WinSDK installation), so I could not verify right now what exact could causes this error on my side.
I have suspect following two thing
1. Nuget.exe (3.5) in the %USERPROFILE%, but Nuget.exe 3.6 in the local TestFx folder should take precedence I suppose.
2. I have customization to global NuGet.config, but right now I don't have any Microsoft internal feed in it right now.
I try to find time to reinstall today while I relatively free and verify what could cause this error. But if this is not happens, could you give me some ideas what else could I check on my PC which could cause incorrect package layout in the local packages folder.

jayaranigarg:
@kant2002 : To find where nuget.exe is getting picked up from and which nuget feeds are getting used to restore nuget packages, you can turn on Verbose logs by specifying "Build.cmd -Verbose"

kant2002:
@jayaranigarg build working using 3.6 version which is located in the root folder of the testfx repository, I already check that.
Here command from running build.cmd -Verbose

    XXX\testfx\nuget.exe restore -msbuildVersion 15.0 -verbosity normal -nonInteractive -configFile XXX\testfx\Nuget.config XXX\testfx\scripts\Toolset\packages.config

jayaranigarg:
@kant2002 : Above command seems perfectly ok. 
Can you try removing customizations from global Nuget.config and try once (just trying to rule out this possibility) ?
Also, did you get a chance to reinstall and check?
Meanwhile, we are following up with the Nuget team to figure out what might be causing this issue.

jayaranigarg:
@kant2002 : Nuget restore issue is now fixed. Can you please go ahead and update your [PR](https://github.com/Microsoft/testfx/pull/299) ?

jayaranigarg:
@kant2002 : Checking back on this. Can you please do the needful?

kant2002:
Build is working for me on master. I think this could be closed.

PR #331, opened 2017-12-08
Fix Nuget restore issue

jayaranigarg:
The reason why we run into scenario where we have XYZ.2.0.0 and XYZ/2.0.0 in the same folder is because both the global package folder and the repositoryPath variable point to the same location.
XYZ.2.0.0 is from the packages.config restore. 
XYZ/2.0.0 is from the global package folder item. 

Normally we should reference the bootstrapped packages in our proj, so use XYZ.2.0.0 paths instead of the XYZ/2.0.0.


jayaranigarg:
https://github.com/Microsoft/testfx/issues/311

PR #332, opened 2017-12-14
Throwing exception only if it is of type TestPlatformFormatException

mayankbansal018:


mayankbansal018:
@dotnet-bot test Windows / Debug Build

src/Adapter/MSTest.CoreAdapter/TestMethodFilter.cs
```diff
@@ -119,10 +119,17 @@ private ITestCaseFilterExpression GetTestCaseFilterFromDiscoveryContext(IDiscove
                 MethodInfo methodGetTestCaseFilter = context.GetType().GetRuntimeMethod("GetTestCaseFilter", new[] { typeof(IEnumerable<string>), typeof(Func<string, TestProperty>) });
                 return (ITestCaseFilterExpression)methodGetTestCaseFilter?.Invoke(context, new object[] { this.supportedProperties.Keys, (Func<string, TestProperty>)this.PropertyProvider });
             }
-            catch (TargetInvocationException ex)
+            catch (Exception ex)
             {
-                throw ex.InnerException;
+                // In case of UWP .Net Native Tool Chain compilation. Invoking methods via Reflection doesn't work, hence discovery always fails.
+                // Hence throwing exception only if it is of type TestPlatformFormatException
+                if (ex.InnerException is TestPlatformFormatException)
```
abhishkk: AFAIK platform throws only testplatformformatexception, but instead of relying entirely on this fact, we can skip exception specifically when it is thrown because of reflection not working. In other cases, lets throw error.
Skipping all exceptions other than testplatformformatexception might skip unexpected exceptions in addition to reflection related exception.

mayankbansal018: In case any other exception is thrown, which is very unlikely, we should continue with discovery rather than not doing at all, because platform went wrong, which is why I'm only throwing error if it platform exception

PR #333, opened 2017-12-14
Enable DonotParallelize for Assembly Level

jayaranigarg:


src/TestFramework/MSTest.Core/Attributes/DoNotParallelizeAttribute.cs
```diff
@@ -8,7 +8,7 @@ namespace Microsoft.VisualStudio.TestTools.UnitTesting
     /// <summary>
     /// Specification to disable parallelization.
     /// </summary>
-    [AttributeUsage(AttributeTargets.Class | AttributeTargets.Method, AllowMultiple = false)]
+    [AttributeUsage(AttributeTargets.Assembly | AttributeTargets.Class | AttributeTargets.Method, AllowMultiple = false)]
```
smadala, line 11: Update log message to show assembly path.

jayaranigarg: Done.

Issue #284, opened 2017-10-09
DataRow DisplayName is not generated properly in .NET Framework when using null values


Mertsch:
## Steps to reproduce
Run this test in .NET Framework 4.6.1 mode
```
[TestClass]
public class UnitTest1
{
    [TestMethod]
    [DataRow("OK", "OK Test")]
    [DataRow("OK1", "OK1 Test")]
    [DataRow(null, " Test")]
    public void TestMethod1(string input, string output)
    {
        Assert.AreEqual(output, $"{input} Test");
    }
}
```

## Expected behavior
Should output the same display name as in .NET Core

## Actual behavior
![image](https://user-images.githubusercontent.com/9402861/31330806-2d141b14-ace0-11e7-875f-0a24f08f9283.png)
And even worse with multiple null tests
![image](https://user-images.githubusercontent.com/9402861/31330908-90623aac-ace0-11e7-95ab-429fb476cb38.png)

## Environment
VS 2017 15.3.5
.NET Core 2.0 / .NET Framework 4.6.1
```
<package id="MSTest.TestAdapter" version="1.2.0-beta3" targetFramework="net461" />
<package id="MSTest.TestFramework" version="1.2.0-beta3" targetFramework="net461" />
```

jayaranigarg:
This is essentially an issue with string.join(String separator, object[] values).

If first value is null like in `[DataRow(null, " Test")]` , both .net framework and net core used to return an empty string earlier. But this issue got fixed in netcore to return (,Test) refer this[https://github.com/dotnet/coreclr/issues/4506]. However, it appears that the same fix is not yet ported to .net framework and hence we are observing a difference in behavior.

PR #335, opened 2017-12-18
DataRow DisplayNameFix in .Net framework

jayaranigarg:
https://github.com/Microsoft/testfx/issues/284


test/UnitTests/MSTest.Core.Unit.Tests/Attributes/DataRowAttributeTests.cs
```diff
@@ -55,5 +59,27 @@ public void GetDataShouldReturnDataPassed()
...
+            var data2 = new string[] { "First", null, "Second" };
+
+            var displayName = dataRowAttribute.GetDisplayName(this.testMethodInfo, data);
+            Assert.AreEqual("DataRowTestMethod (First,Second,)", displayName);
+
+            displayName = dataRowAttribute.GetDisplayName(this.testMethodInfo, data1);
+            Assert.AreEqual("DataRowTestMethod (,First,Second)", displayName);
+
+            displayName = dataRowAttribute.GetDisplayName(this.testMethodInfo, data2);
+            Assert.AreEqual("DataRowTestMethod (First,,Second)", displayName);
```
AbhitejJohn, line 82: This does look a bit odd, It does not immediately occur to me that the second parameter is a null.

AbhitejJohn: It looks like this is .Net default though and consumers are already used to this. So that should be ok - it still looks slightly odd though.

src/TestFramework/MSTest.Core/Attributes/DataRowAttribute.cs
```diff
@@ -69,7 +69,7 @@ public string GetDisplayName(MethodInfo methodInfo, object[] data)
             {
                 if (data != null)
                 {
-                    return string.Format(CultureInfo.CurrentCulture, FrameworkMessages.DataDrivenResultDisplayName, methodInfo.Name, string.Join(",", data));
+                    return string.Format(CultureInfo.CurrentCulture, FrameworkMessages.DataDrivenResultDisplayName, methodInfo.Name, string.Join(",", (IEnumerable<object>)data));
```
AbhitejJohn: would `DataSource` need something similar too?

jayaranigarg: No. DisplayName in case of DataSourceAttribute doesn't consist of any data as such. It looks like "TestMethod (Data Row 0)", which is irrespective of the data.

test/UnitTests/MSTest.Core.Unit.Tests/Attributes/DataRowAttributeTests.cs
jayaranigarg: Yes, looks a bit odd to me as well. But since we want a consistent behavior irrespective of which argument is null and hence this change.
This is fixed in .net core but fix has not yet been ported to .net framework, so we are fixing this on our end.

src/TestFramework/MSTest.Core/Attributes/DataRowAttribute.cs
codito: Minor: should we consider `AsEnumerable` instead of casting to `IEnumerable<T>`?
https://stackoverflow.com/questions/2013846/why-use-asenumerable-rather-than-casting-to-ienumerablet

jayaranigarg: This is good 👍

jayaranigarg:
@dotnet-bot Test Windows / Release Build  please

Issue #361, opened 2018-02-06
Two GenericParameterHelper instances may be equal when they are created continually using the default constructor.


walterlv:
## Description

This is the default constructor of `GenericParameterHelper`:

```csharp
public GenericParameterHelper()
{
    Random randomizer = new Random();
    this.Data = randomizer.Next();
}
```

If you `new` it twice continually, `new Random().Next` may return the same value. In this case, these two instances are equals.

```csharp
var first = new GenericParameterHelper();
var second = new GenericParameterHelper();
// Equal when running but not equal when step-by-step.
Assert.AreEqual(first, second);
```

But if you `new` them separately, they not equals.

```csharp
var first = new GenericParameterHelper();
await Task.Delay(100);
var second = new GenericParameterHelper();
// Always not equal.
Assert.AreNotEqual(first, second);
```

This is serious when I try to set a breakpoint to debug my unit test method.

- If I run my unit test directly, they are equals.
- If I run my unit step-by-step (or by setting a breakpoint), they are not equals.

So, I will get different result when running and debugging the same unit tests.

## Expected behavior

Same behavior is expected between running and debugging unit tests. This means that they should be both equal or not equal.

## Recommended Improvement

1. A static Random instance should be stored in `GenericParameterHelper` so that every time when I `new` it I get different instance. If I need to get equaled instances I can use the parameterized constructor.
1. Add test case for `GenericParameterHelperTest` to test the default constructor.

## Environment

.NET Framework 4.5/4.6/4.7

PR #362, opened 2018-02-06
Make GenericParameterHelper's behavior same between running and debugging

walterlv:
See #361 for more details.

I make a static `Random` instance so that it would return a different value every time when calling `Next`. This can fix the different behavior between running and step-by-step debugging unit tests.


jayaranigarg:
Thank you @walterlv  for the fix.

PR #365, opened 2018-02-07
Added overloads to `Assert.ThrowsException` to check the parameter name of an `ArgumentException`

dlemstra:
This pull requests allows a simpler way to check if an `ArgumentException` uses the correct `ParameterName`. With the current code you will need to do this:

```C#
var exception = Assert.ThrowsException<ArgumentNullException>(() =>
{
    new FooBar(null);
});

Assert.AreEqual("myParam", exception.ParamName);
```

This pull request adds a patch to allow the following notation:

```C#
Assert.ThrowsException<ArgumentNullException>("myParam", () =>
{
    new FooBar(null);
});
```

This patch has also been applied to `Assert.ThrowsExceptionAsync`.

I renamed the `InvalidParameterToAssert` resource because it was unused. I can revert that if the original value should be preserved.

### DIFF
--- a/src/TestFramework/MSTest.Core/Resources/FrameworkMessages.Designer.cs
+++ b/src/TestFramework/MSTest.Core/Resources/FrameworkMessages.Designer.cs
@@ -20,7 +20,7 @@ namespace Microsoft.VisualStudio.TestTools.UnitTesting {
     // class via a tool like ResGen or Visual Studio.
     // To add or remove a member, edit your .ResX file then rerun ResGen
     // with the /str option, or rebuild your VS project.
-    [global::System.CodeDom.Compiler.GeneratedCodeAttribute("System.Resources.Tools.StronglyTypedResourceBuilder", "4.0.0.0")]
+    [global::System.CodeDom.Compiler.GeneratedCodeAttribute("System.Resources.Tools.StronglyTypedResourceBuilder", "15.0.0.0")]
     [global::System.Diagnostics.DebuggerNonUserCodeAttribute()]
     [global::System.Runtime.CompilerServices.CompilerGeneratedAttribute()]
     internal class FrameworkMessages {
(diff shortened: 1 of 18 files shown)


src/TestFramework/MSTest.Core/Resources/FrameworkMessages.Designer.cs
```diff
@@ -20,7 +20,7 @@ namespace Microsoft.VisualStudio.TestTools.UnitTesting {
     // class via a tool like ResGen or Visual Studio.
     // To add or remove a member, edit your .ResX file then rerun ResGen
     // with the /str option, or rebuild your VS project.
-    [global::System.CodeDom.Compiler.GeneratedCodeAttribute("System.Resources.Tools.StronglyTypedResourceBuilder", "4.0.0.0")]
+    [global::System.CodeDom.Compiler.GeneratedCodeAttribute("System.Resources.Tools.StronglyTypedResourceBuilder", "15.0.0.0")]
```
jayaranigarg, line 23: Kindly remove this file. This is auto-generated.

jayaranigarg:
Could you run build with -uxlf switch so that the corresponding xlf files are updated for localization. Please check-in those localized files as well.

src/TestFramework/MSTest.Core/Resources/FrameworkMessages.Designer.cs
dlemstra: This file was already part of the project? Do you want me to remove it now?

jayaranigarg: I meant remove the changes you made to this file as part of this PR.

dlemstra: Can you clarify why you want to move these changes to another commit? The next time that someone will do a build this file will be changed and included in the commit that they will make.

jayaranigarg: We do not make any changes to this file.It is just a placeholder. This file is auto-generated with every build. So once this PR will be pushed in(without this file), this file will get auto-generated with the relevant code changes automatically on every build. We don't commit any changes to this file explicitly.

dlemstra: The changes have been reverted.

dlemstra: @jayaranigarg Reverting the file seems to break the build? This file is not generated during the build?

dlemstra:
@jayaranigarg Reverting the changes to `FrameworkMessages.Designer.cs` broke the build. Are you investigating this or should I submit the changes to the file again?

dlemstra:
@jayaranigarg I have restored the changes to `FrameworkMessages.Designer.cs` because reverting them broke the build. Can you have another look at this PR?

jayaranigarg:
@pvlakshm: Can you please look into this?

pvlakshm:
It will be ideal if it can be made extensible to any exception property with chaining like:
`Assert.ThrowsException<ArgumentException(() => throws()).And.AssertAreEqual("expectedParam", e.ParamName)`

Closing this old PR. Kindly let me know if you have any concern.

Issue #355, opened 2018-02-01
When a class contains only tests with [Ignore], class cleanup runs but not class init


dee-see:
## Steps to reproduce
Run the following test class, only `ClassCleanup` runs and there is a `NullReferenceException`.
```
using System;
using Microsoft.VisualStudio.TestTools.UnitTesting;

namespace UnitTestProject1
{
	[TestClass]
	public class UnitTest1
	{
		private static object Obj;

		[ClassInitialize]
		public static void ClassInit( TestContext context )
		{
			Obj = new object();
		}

		[ClassCleanup]
		public static void ClassCleanup()
		{
			Console.WriteLine( Obj.GetType() );
		}

		[TestMethod]
		[Ignore] // The only test in this class is ignored
		public void TestMethod1()
		{
			Assert.AreEqual( 1, 1 );
		}
	}
}

```

## Expected behavior
One of two scenarios

1. The class is not even instantiated and nothing runs
2. `ClassInitialize` and `ClassCleanup` run

Instantiating the class but running neither `ClassInitialize` nor `ClassCleanup` could leave some undisposed class variables... Though I'm wondering as I'm writing this if that's a bad practice and should be handled by an `IDisposable` test class.

## Actual behavior
Only the class cleanup is executed, resulting in a `NullReferenceException` in this example.

My real life scenario is that we have a class with several tests and the only tests that matches the TestCaseFilter in this build is `[Ignore]`d.

## Environment
VS 2017 15.5.5
MsTest v2 1.3.0-beta2, 1.2 and 1.1.18

jayaranigarg:
@pvlakshm : Can you take this?

jayaranigarg:
@dee-see : We acknowledge this to be bug on our end and we will be fixing this in our upcoming releases.

acesiddhu:
this is now fixed

PR #372, opened 2018-02-19
Today even if classInit wasnt called we used to call class cleanup. f…

acesiddhu:
Today even if classInit wasnt called we used to call class cleanup. fixing that and adding a test to cover the scenario

#355

### DIFF
--- a/src/Adapter/MSTest.CoreAdapter/Execution/TestClassInfo.cs
+++ b/src/Adapter/MSTest.CoreAdapter/Execution/TestClassInfo.cs
@@ -108,11 +108,6 @@ namespace Microsoft.VisualStudio.TestPlatform.MSTest.TestAdapter.Execution
         public bool IsClassInitializeExecuted { get; internal set; }
 
         /// <summary>
-        /// Gets a value indicating whether class cleanup has executed.
-        /// </summary>
-        public bool IsClassCleanupExecuted { get; private set; }
-
-        /// <summary>
         /// Gets the exception thrown during <see cref="ClassInitializeAttribute"/> method invocation.
         /// </summary>
         public Exception ClassInitializationException { get; internal set; }
@@ -313,7 +308,7 @@ namespace Microsoft.VisualStudio.TestPlatform.MSTest.TestAdapter.Execution
 
             lock (this.testClassExecuteSyncObject)
             {
-                if (this.IsClassInitializeExecuted)
+                if (this.IsClassInitializeExecuted || this.ClassInitializeMethod == null)
                 {
                     try
                     {
@@ -345,10 +340,6 @@ namespace Microsoft.VisualStudio.TestPlatform.MSTest.TestAdapter.Execution
                             this.ClassCleanupMethod.Name,
                             errorMessage,
                             StackTraceHelper.GetStackTraceInformation(realException)?.ErrorStackTrace);
-                    }
-                    finally
-                    {
-                        this.IsClassCleanupExecuted = true;
                     }
                 }
             }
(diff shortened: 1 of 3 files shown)


src/Adapter/MSTest.CoreAdapter/Execution/TestClassInfo.cs
```diff
@@ -306,40 +311,52 @@ public string RunClassCleanup()
...
-                    string errorMessage;
-
-                    // special case AssertFailedException to trim off part of the stack trace
-                    if (realException is AssertFailedException ||
-                        realException is AssertInconclusiveException)
-                    {
-                        errorMessage = realException.Message;
-                    }
-                    else
+                    if (this.IsClassInitializeExecuted)
```
mayankbansal018: >IsClassInitializeExecuted [](start = 29, length = 25)

why another check needed inside?

test/UnitTests/MSTest.CoreAdapter.Unit.Tests/Execution/TestClassInfoTests.cs
```diff
@@ -111,6 +111,18 @@ public void TestClassInfoClassCleanupMethodSetShouldThrowForMultipleClassCleanup
             ActionUtility.ActionShouldThrowExceptionOfType(action, typeof(TypeInspectionException));
         }
 
+        [TestMethod]
```
jayaranigarg, line 114: This is a negative test case. Can we add a positive test case as well?

src/Adapter/MSTest.CoreAdapter/Execution/TestClassInfo.cs
```diff
@@ -103,10 +103,15 @@ internal set
...
         /// <summary>
-        /// Gets a value indicating whether is class initialize executed.
+        /// Gets a value indicating whether class initialize has executed.
         /// </summary>
         public bool IsClassInitializeExecuted { get; internal set; }
 
+        /// <summary>
+        /// Gets a value indicating whether class cleanup has executed.
+        /// </summary>
+        public bool IsClassCleanupExecuted { get; private set; }
```
smadala: Is this required?

src/Adapter/MSTest.CoreAdapter/Execution/TestClassInfo.cs
```diff
@@ -306,40 +311,52 @@ public string RunClassCleanup()
...
+            if (this.IsClassInitializeExecuted)
             {
-                try
-                {
-                    this.ClassCleanupMethod.InvokeAsSynchronousTask(null);
-
-                    return null;
-                }
-                catch (Exception exception)
+                lock (this.testClassExecuteSyncObject)
```
smadala: lock can be removed? All clean ups done in [single thread](https://github.com/Microsoft/testfx/blob/0789f7a8c1854f406e6ce93dd0bf468a1ec5ef52/src/Adapter/MSTest.CoreAdapter/Execution/TestExecutionManager.cs#L337). /cc @AbhitejJohn

src/Adapter/MSTest.CoreAdapter/Execution/TestClassInfo.cs
```diff
@@ -306,40 +311,52 @@ public string RunClassCleanup()
                 return null;
             }
 
-            lock (this.testClassExecuteSyncObject)
+            if (this.IsClassInitializeExecuted)
```
smadala: Nit: how about `this.ClassCleanupMethod == null || this.IsClassInitializeExecuted == false`?

PR #387, opened 2018-03-25
Reduction of Resharper errors.

zippy1981:
Improvements to reduce ReSharper warnings. These were not mindlessly accepting ReShaper recommendations. I went through stuff by hand. 

If some of these could be done with stylecop, let me know I'll figure out how and make a PR to add tests there.


pvlakshm:
What is the issue corresponding to this?

zippy1981:
@pvillads There is no corresponding issue. Should I open one? It seems like its easier to do the work and then discuss it in the PR since is all formatting type work.

While working on a PR for an issue I opened I noticed many warnings generated by ReSharper. I wanted to keep that PR "clean."  

I picked a subset of clear issues to fix here:

* In xmldocs the cref="" tags were wrong for generics. and some angle brackets needed to be escaped.
* A file had the wrong namespace
* A ToString() call that ReSharper deemed unnecessary via static analysis was actually necessary because it was checking for bugs in ToString() overloads. I supresssed that warning.
* Some folders that aren't part of the namespace are now marked as such.

I'm sure some of this stuff could be fixed via Stylecop. I have never configured stylecop, but I can see that a very robust stylecop ruleset is being used in the solution. If some of these things could be enforced by stylecop, I could figure out how to do that as well.

PR #395, opened 2018-04-02
Fix to show right error message on assembly load exception during test run.

smadala:
Fix to show right error message on assembly load exception during test run.

More details: https://mseng.visualstudio.com/DefaultCollection/VSOnline/Automated%20Testing/_workitems/edit/1219450


src/Adapter/MSTest.CoreAdapter/Execution/StackTraceHelper.cs
```diff
@@ -67,9 +67,28 @@ internal static StackTraceInformation GetStackTraceInformation(Exception ex)
...
+                    {
+                        // try to get stacktrace
+                        if (e.StackTrace != null)
+                        {
+                            stackTraces.Push(e.StackTrace);
+                        }
+                    }
+                    catch (Exception)
+                    {
+                        // Give up.
```
acesiddhu: logging

PR #401, opened 2018-04-11
Signing for nuget packages

singhsarab:
Adding the sign proj for nuget package signing.


scripts/verify-sign.ps1
```diff
@@ -0,0 +1,64 @@
...
+$TPB_Configuration = $Configuration
+
+function Verify-NugetPackages
+{
+    Write-Log "Verify-NugetPackages: Start"
+
+    $toolsDirectory = Join-Path $rootDirectory "tools" 
+
+    # Move acquiring nuget.exe to external dependencies once Nuget.Commandline for 4.6.1 is available.
+    $nugetInstallDir = Join-Path $toolsDirectory "nuget"
```
jayaranigarg: No need to download nuget.exe again. Our build script already does it. Consider using [this ](https://github.com/Microsoft/testfx/blob/master/scripts/common.lib.ps1#L61)function.

PR #403, opened 2018-04-17
TestTimeout in RunSettings

singhsarab:
Making test timeout configurable via runsettings
https://github.com/Microsoft/testfx-docs/pull/40

### DIFF
--- a/src/Adapter/MSTest.CoreAdapter/Execution/TypeCache.cs
+++ b/src/Adapter/MSTest.CoreAdapter/Execution/TypeCache.cs
@@ -286,7 +286,7 @@ namespace Microsoft.VisualStudio.TestPlatform.MSTest.TestAdapter.Execution
             foreach (var methodInfo in classType.GetTypeInfo().DeclaredMethods)
             {
                 // Update test initialize/cleanup method
-                this.UpdateInfoIfTestInitializeOrCleanupMethod(classInfo, methodInfo, isBase: false, instanceMethods: instanceMethods,  testInitializeAttributeType: testInitializeAttributeType, testCleanupAttributeType: testCleanupAttributeType);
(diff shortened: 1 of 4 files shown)


test/UnitTests/MSTest.CoreAdapter.Unit.Tests/Execution/TypeCacheTests.cs
```diff
@@ -815,6 +816,57 @@ public void GetTestMethodInfoShouldThrowWhenTimeoutIsIncorrect()
             Assert.AreEqual(expectedMessage, exception.Message);
         }
 
+        [TestMethodV1]
+        public void GetTestMethodInfoWhenTimeoutAttributeNotSetShouldReturnTestMethodInfoWithGlobalTimeout()
+        {
+            string runSettingxml =
+                @"<RunSettings>
+                    <MSTestV2>
+                        <TestTimeout>4000</TestTimeout>
```
jayaranigarg, line 825: Should also add one negative test case where timeout is invalid(negative value or non-integer value)

src/Adapter/MSTest.CoreAdapter/Execution/TypeCache.cs
```diff
@@ -645,6 +646,10 @@ private int GetTestTimeout(MethodInfo methodInfo, TestMethod testMethod)
 
                 return timeoutAttribute.Timeout;
             }
+            else if (globalTimeout > 0)
```
jayaranigarg, line 649: Have we asked @pvlakshm about precedence order?

jayaranigarg: What happens when we passing runsettings arguments through commandline??

singhsarab: https://github.com/Microsoft/testfx-docs/pull/40

test/UnitTests/MSTest.CoreAdapter.Unit.Tests/Execution/TypeCacheTests.cs
singhsarab: Done.

Issue #197, opened 2017-06-09
Resource folder "en" is deleted


sheitmann:
Hi,

I have a shared output directory for my library assemblies and my test assemblies.
After adding a project using MSTestV2 I noticed that my "en" resource folder is deleted when the unit test project is created. This caused some tests to fail because the resources where not there anymore.
After some investigation I found the msbuild property to disable that behavior: EnableMSTestV2CopyResources 
If it is activated (default is true) it is cleaning up some resources and deletes the whole folder.

My question is, what is this used for at all and what does that mean for me now that I have disabled it?
Is a good idea to delete the whole resources directory because of side effects?

Br
Sven

AbhitejJohn:
@shietmann: We use this to ensure that failure/warning messages from the adapter are localized for non en languages. This msbuild task copies over Framework/Adapter resources over to the Out directory and ensures they are cleaned up on Clean. Its strange that a project creation is deleting that folder. Is this `dotnet new`? If this is a project build on the other hand, can you share the msbuild logs with us please?

P.S: You have the right workaround for now and this should not impact you as long as you are running against en.

sheitmann:
@AbhitejJohn 
thanks for your answer
the problem is running a rebuild from visual studio
each project runs a clean before build and so all resources files in the en folder which were built before the test project are deleted.

AbhitejJohn:
@sheitmann : Got it. This should be a bug.
To give you context [here](https://github.com/Microsoft/testfx/blob/master/src/Adapter/Build/Desktop/MSTest.TestAdapter.targets#L41) is the code that does it. We added that piece because build was not cleaning up empty folders. We either need to add in logic to clean conditionally or not clean at all and leave the empty folders as is.

ShreyasRmsft:
Add a conditional clean to the csproj.

jayaranigarg:
@sheitmann : Can you please provide us with a small repro for your problem?

sheitmann:
@jayaranigarg 
I have attached a small project.
The important part is that all projects share the same output directory.

If you do a "rebuild solution" you will see that the en folder with the resource assemblies is missing.
The tests will fail.

If you do a "clean solution" and a "build solution" the en folder is still there and the tests will succeed.
[TestIssueRepo.zip](https://github.com/Microsoft/testfx/files/2030992/TestIssueRepo.zip)

When not sharing the same output directory it is working fine.

jayaranigarg:
[...]

PR #424, opened 2018-05-21
Don't delete resource folder while clean/rebuild

jayaranigarg:
We had 2 ways to fix this : clean conditionally or not clean at all. We are choosing to not clean at all because in large project systems rebuilding happens parallelly and one project trying to access it while other trying to clean it might create troubles.

Doing a clean/rebuild on loc machines, deletes and re-creates "en" (culture specific folder) and hence there will be no such scenario where this "en" folder will be empty which we need to cleanup. Additionally, creating/deleting the "en" folder should be a build/clean/rebuild functionality and should not be tied to MSTestv2 in any way.

https://github.com/Microsoft/testfx/issues/197


smadala:
Update the title too.

Issue #470, opened 2018-08-03
Unseal IgnoreAttribute


dotMorten:
## Description

Please unseal the IgnoreAttribute:
https://github.com/Microsoft/testfx/blob/e13ca085694aa9a9908afb3e9a90a58841b5fe03/src/TestFramework/MSTest.Core/Attributes/VSTestAttributes.cs#L93

Unsealing it would allow me to redefine when a test should be ignored or not. For instance in my day-to-day work I just want to skip over the known issues, but occasionally I'd like to check if an issue has been resolved or not. This can be accomplished like this:

```cs
public class KnownIssueAttribute :
#if EXECUTE_IGNORE_TESTS
        TestPropertyAttribute
#else
        IgnoreAttribute
#endif
{
       public KnownIssueAttribute(string issueUri) :
#if EXECUTE_IGNORE_TESTS
        base("KnownIssue", issueUri)
#else
        base(issueUri)
#endif
    {
    }
}
```

When the compilation is run with `EXECUTE_IGNORE_TESTS` defined, I'll get a property instead that defines the issue, which can then be pulled from the test report and correlated with the issue status.

Tests would then be flagged like this:

```cs
[TestMethod]
[KnownIssue("https://github.com/Microsoft/testfx/issues/470")]
public void Test() { }
```

dotMorten:
After poking around some more, I'm realizing you can accomplish the scenario with a custom TestMethod instead:

```cs
    public class KnownIssueTestMethodAttribute : TestMethodAttribute
    {
        public KnownIssueTestMethodAttribute(string issueLink, Type exceptionType)
        {
            IssueLink = issueLink;
            KnownException = exceptionType;
        }

        public string IssueLink { get; set; }

        public Type KnownException { get; set; }

        public override TestResult[] Execute(ITestMethod testMethod)
        {
            var results = base.Execute(testMethod);
            var result = results[0];
            if (result.Outcome == UnitTestOutcome.Failed)
            {
                if (result.TestFailureException.InnerException.GetType() == KnownException)
                {
                    result.Outcome = UnitTestOutcome.Inconclusive;
                    result.TestFailureException = new Exception(IssueLink, result.TestFailureException.InnerException);
                }
            }
            else if (result.Outcome == UnitTestOutcome.Passed)
            {
                result.Outcome = UnitTestOutcome.Failed;
                result.TestFailureException = new KnownIssuePassedException("Test passed, but still marked as a known issue: " + IssueLink);
            }
[...]

PR #471, opened 2018-08-03
Unseals IgnoreAttribute

dotMorten:
Addresses #470


jayaranigarg:
@dotMorten : Thank you for the contribution. But just unsealing the attribute won't work here directly. Since this class was marked as sealed, we weren't checking for attributes derived from this class. You will have to add that check here:
https://github.com/Microsoft/testfx/blob/master/src/Adapter/MSTest.CoreAdapter/Execution/TestMethodRunner.cs#L133
You will need to add a check for attributes deriving from IgnoreAttribute.

dotMorten:
Good catch! Fixed and added unit test for it:
image

dotMorten:
I have realized that this PR might not really be necessary, as this can all be done with a custom TestMethod attribute instead. I'll update the issue...

Issue #478, opened 2018-08-17
TestAdapter grouping by Traits should not create separate groups for Description and WorkItem attributes


robdalsanto:
I think the title and the picture illustrate where this can lead. Note also that Owner should not split into different groups. I don't believe this happened prior to MSTest-V2, at least not with WorkItem and Description.

Example of Description attribute:
```
        [TestMethod, SmokeTypeTest]
        [Description("Bug fix 1506 and 1516. If Alternative linked to Outage then Alternative date has to be updated to selected Outage occurrence date")]
        public void UpdateAlternative_LinkedToOutageTest()
        {
```

![image](https://user-images.githubusercontent.com/31216152/44291509-a0d2c700-a233-11e8-805e-5a019820b2e6.png)

jayaranigarg:
@robdalsanto : The behaviour has always been like this from adapter side, in both MSTestV1 and MSTestV2.

Tagging @AbhitejJohn @genlu to look into if there has been some behavior difference from test explorer end.

robdalsanto:
@jayaranigarg @AbhitejJohn We are seeing this duplication in VS2015, which occurred immediately after we started to use the NuGet MSTestV2 instead of the MSTestV1 shipped in VS2015.

We don't have a lot of these attributes, and they don't provide much value, like allowing TFS to link to the test-methods from the work-item. So we will likely remove them. Still it is odd to see these 3 attributes as traits.

singhsarab:
@robdalsanto We identified and fixed the issue. It will be available in the next preview release of the adapter.

PR #482, opened 2018-09-04
Description, WorkItem, CssIteration, CssProjectStructure should *NOT* be treated as traits.

singhsarab:
Attributes like Description should not inherit from TestPropertyAttribute, as eventually TestPropertyAttributes are treated as traits.

Fix for https://github.com/Microsoft/testfx/issues/478

Docs: Microsoft/testfx-docs#54

### DIFF
--- a/src/TestFramework/MSTest.Core/Attributes/VSTestAttributes.cs
+++ b/src/TestFramework/MSTest.Core/Attributes/VSTestAttributes.cs
@@ -182,8 +182,6 @@ namespace Microsoft.VisualStudio.TestTools.UnitTesting
     {
     }
 
-    #region Description attributes
-
     /// <summary>
     /// Test Owner
     /// </summary>
@@ -234,14 +232,13 @@ namespace Microsoft.VisualStudio.TestTools.UnitTesting
     /// Description of the test
     /// </summary>
     [AttributeUsage(AttributeTargets.Method, AllowMultiple = false)]
-    public sealed class DescriptionAttribute : TestPropertyAttribute
+    public sealed class DescriptionAttribute : Attribute
     {
         /// <summary>
         /// Initializes a new instance of the <see cref="DescriptionAttribute"/> class to describe a test.
         /// </summary>
         /// <param name="description">The description.</param>
         public DescriptionAttribute(string description)
-            : base("Description", description)
         {
             this.Description = description;
         }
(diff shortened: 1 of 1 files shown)


jayaranigarg:
@AbhitejJohn : Any reason why these attributes were inherited from TestPropertyAttribute?

AbhitejJohn:
@jayaranigarg : [This](https://github.com/Microsoft/testfx-docs/blob/master/RFCs/001-Framework-Extensibility-Description-Attributes.md) seems to explain it. you might want to revise that doc if this changes any of those behaviors for the default attributes.
/cc : @jfleisher .

jayaranigarg:
@singhsarab : Can you please update that doc as well as part of this change?

singhsarab:
Associated changes in the docs. https://github.com/Microsoft/testfx-docs/pull/54

dotMorten:
These appears to me as breaking changes. Was that intended?

singhsarab:
@dotMorten Can you please share the concerns you have around this change ?

dotMorten:
While changing a class to inherit from a subclass of the original class isn't a breaking change, doing the opposite as what is being done here is.

Scenario:
Library 1 references Nuget1 v1.3.2
Library 2 references Nuget v1.3.3 with the inheritance change
App is referencing both, and thus needs to reference 1.3.3, but hits code in Library 1 that in turn hits this changed class. Due to a break in the binary compatibility, Library 1 is no longer working as it can't find the members it expects in the location it expected. You'd have to contact the author of Library 1 and ask them to upgrade to 1.3.3 and recompile, before you can continue t use both libraries.

Breaking changes should only be done in major versions, it you'd have to waif for v2.0.0

jayaranigarg:
@dotMorten : We don't see anything breaking in this change as of now. We certainly will validate this scenario before releasing a higher version of adapter and will revert/make appropriate changes if we see any issues.

dotMorten:
@jayaranigarg You probably aren't seeing anything because you recompiled anything and didn't worry about binary compatibility. See the guidelines here:
https://github.com/dotnet/corefx/blob/master/Documentation/coding-guidelines/breaking-change-rules.md
Specifically this one:

> Removing one or more base classes for a type, including changing struct to class and vice versa

dotMorten:
...also any class accepting the base type can no longer pass instances of the derived attribute, so it's also a breaking change that the source code level.

singhsarab:
@dotMoren I understand the from Oops perspective, it seems like a breaking change. 
But thinking from the perspective of what this code is used for, this change seemed harmless.

- These are sealed classes, so cannot be extended.
- Users use these attributes only to decorate their unit test.
- I cannot think of the scenario where the user will actually try to create an object of the attribute and cast. Maybe we are missing it, can you please share some real life scenario where this is relevant.

I am actually fine bumping up the version for the adapter if we really think it breaks the users.
Also, this change has already been rolled out the pre-release version 1.4.0-beta and we haven't heard back on it.

ahmedalejo:
> * Users use these attributes only to decorate their unit test.
i agree with @dotMorten if a class public and has been released, there no way you can presume that developers only uses that the way you expect. Especially used reflection for code analysis.

There´s no problem with change itself, only with the version update.

Change the class hierarchy of public class *is* a breaking change and should be a major version version update **2.*** and not **1.***

dotMorten:
> I cannot think of the scenario where the user will actually try to create an object of the attribute and cast.
That's the thing with an API design: You don't know every possible use out there, so you have to evolve your API very defensively. That's just the harsh (and IMHO fun) nature of building APIs.

But as an example I interrogate the attributes heavily in my own TestMethod to better drive execution and logging. There are also custom test runners (typical in Xamarin etc) that'll use the attributes for running the tests on devices.

singhsarab:
@ahmedalejo @dotMorten I agree, in future, we will be more careful with the versioning.
We haven't received any reports with respect to this change though, we release 1.4.0 with this change, half a million downloads already.

Issue #442, opened 2018-06-06
Message Cleanup


spottedmahn:
## Description
VS and Rosyln are now providing better analysis of our code!  But now there are ~42 messages in the code base that need to reviewed

![image](https://user-images.githubusercontent.com/14079228/41037328-00ceafae-6961-11e8-9297-7eb7afac48df.png)

One example:

![image](https://user-images.githubusercontent.com/14079228/41037322-fd440cda-6960-11e8-8634-7235cf661bb6.png)

## Steps to reproduce
- Open Code w/ VS 2017 (15.7.3) or later

## Expected behavior
- No messages

## Actual behavior
- ~42 messages

Would you accept a PR that addressed these messages?

singhsarab:
Although these are just suggestions, I guess it is good to clean up.

parrainc:
@singhsarab It seems that no body's currently working on it...
Taking this issue :)

PR #492, opened 2018-09-26
Cleaned up messages

parrainc:
This PR fixes issue #442 
It's just cleaning up the messages/suggestions by VS.

After these changes it'll looks like this:
after

No Messages shown!

### DIFF
--- a/src/Adapter/MSTest.CoreAdapter/Execution/TypeCache.cs
+++ b/src/Adapter/MSTest.CoreAdapter/Execution/TypeCache.cs
@@ -36,20 +36,20 @@ namespace Microsoft.VisualStudio.TestPlatform.MSTest.TestAdapter.Execution
         /// <summary>
         /// Helper for reflection API's.
         /// </summary>
-        private ReflectHelper reflectionHelper;
+        private readonly ReflectHelper reflectionHelper;
 
         /// <summary>
         /// Assembly info cache
         /// </summary>
-        private Dictionary<Assembly, TestAssemblyInfo> testAssemblyInfoCache;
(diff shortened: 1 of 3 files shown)


jayaranigarg:
Thank you for the contribution. Looks good 👍 . Will push this in.

src/Adapter/MSTest.CoreAdapter/Execution/TypeCache.cs
```diff
@@ -36,20 +36,20 @@ internal class TypeCache : MarshalByRefObject
...
         /// Helper for reflection API's.
         /// </summary>
-        private ReflectHelper reflectionHelper;
+        private readonly ReflectHelper reflectionHelper;
 
         /// <summary>
         /// Assembly info cache
         /// </summary>
-        private Dictionary<Assembly, TestAssemblyInfo> testAssemblyInfoCache;
+        private readonly Dictionary<Assembly, TestAssemblyInfo> testAssemblyInfoCache;
```
singhsarab, line 44: Why does this have to be readonly?

jayaranigarg: Latest analysis tool in VS is giving suggestions for fields to be readonly if they are instantiated only in the declaration/constructor.

parrainc: in short, what @jayaranigarg mentioned. Since `testAsseblyInformation` field is only declared in TypeCache constructor, as suggested by the VS code analysis tool, it's recommended to have it `readonly`.

Issue #503, opened 2018-10-22
Assembly Initialize/Cleanup methods not being reached out when inheriting TestClass attribute


parrainc:
## Description
When inheriting from `TestClassAttribute` class, and using the newly create attribute in a test class (somehting like `[MyTestClass]`) neither the assembly initialize/cleanup methods are being called.

i.e.:
```
public class ModifiedTestClassAttribute : TestClassAttribute { }

[ModifiedTestClass]
public class MyTestClass { ... }
```

## Steps to reproduce
In order to reproduce this issue, you should do the following:
1. Create a MSTest test project in VS.
2. Create a class that inherits from TestClassAttribute.
3. Create a test class, and use the class created in previous step as its attribute.
4. Define Assembly initialize/cleanup methods with some logic inside.
5. Create a simple test method and run it.

## Expected behavior

> Call to AssemblyInitialize
> Call to ClassInitialize
> Call to TestMethod(s)
> ...
> Call to ClassCleanup
> Call to AssemblyCleanup

_should be the same behavior when using `[TestClass]`_

## Actual behavior
> Call to ClassInitialize
> Call to TestMethod(s)
> ...
> Call to ClassCleanup

_No Assembly initialize/cleanup method was executed_

## Environment
- OS: Windows 10
- MSTest: 1.3.2 (latest as of today)

parrainc:
Just figured out that the problem is in the [IsAttributeDefined](https://github.com/Microsoft/testfx/blob/master/src/Adapter/MSTest.CoreAdapter/Helpers/ReflectHelper.cs#L32) method from the `ReflectHelper`. When inherit from `TestClass`, it is just looking for an attribute match in the `memberInfo` param with the specified `attributeType` (which is TestClassAttribute).

I'm working on the fix for this issue.

parrainc:
Closing this issue since it was fixed as part of PR #525

KitoCoding:
@parrainc 

I know this is an old ticket, but this problem seems to be happening again.

Haplois:
@KitoCoding, will you able to share a quick repro with us? If not, could you share the Test Platform version (and other NuGet packages you use with their versions) with us?

PR #505, opened 2018-10-23
Issue when custom test class attribute inherits from TestClassAttribute

parrainc:
This PR fixes issue #503 

By changing the IsAttributeDefined method in the ReflectHelper to look at the memberInfo attributes base type to verify it matches with the attributeType (TestClassAttribute).

Behavior after this fix:

> Call to AssemblyInitialize
> Call to ClassInitialize
> Call to TestMethod(s)
> ...
> Call to ClassCleanup
> Call to AssemblyCleanup

after


parrainc:
@jayaranigarg can you please see why the Debug/Release build triggers are still pending?

mayankbansal018:
@dotnet-bot test this please

mayankbansal018:
@dotnet-bot test Windows / Debug Build  please

jayaranigarg:
@parrainc : Consider using `inherit` parameter of `IsAttributeDefined()`. Setting this to `true` should work as well.

parrainc:
> @parrainc : Consider using `inherit` parameter of `IsAttributeDefined()`. Setting this to `true` should work as well.

Hi @jayaranigarg if i just set inherit to `true` in line 428, the `IsAttributeDefined()` method wouldn't be reached out with a custom test class attribute. The problem is happening in [line 393](https://github.com/Microsoft/testfx/blob/master/src/Adapter/MSTest.CoreAdapter/Execution/TypeCache.cs#L393) when checking if `IsAttributeDefined()` on the `type` (even with `inherit`set to `true`) the method is not finding a match of `TestClassAttribute` in the type's attributes, hence the test class is not being examined to look for assembly initialize/cleanup methods. 

So, my change in the method, makes sure (as a last resource) to look for a match of the desired `attributeType` in the attributes' BaseType.

parrainc:
@jayaranigarg having that additional check doesn't let the initialize/cleanup method to be discovered. Still happening same issue, around 7 tests failed if line 393 looks like 

```
if (!this.reflectionHelper.IsAttributeDefined(t, typeof(TestClassAttribute), inherit: true)
 || !this.reflectionHelper.HasAttributeDerivedFrom(t, typeof(TestClassAttribute), false))
{ continue; }
```
However, even though the `.HasAttributeDerivedFrom()` would return true in [line 393](https://github.com/Microsoft/testfx/blob/master/src/Adapter/MSTest.CoreAdapter/Execution/TypeCache.cs#L393), when reaching the `.IsAssemblyOrClassInitializeMethod()` in [line 412](https://github.com/Microsoft/testfx/blob/master/src/Adapter/MSTest.CoreAdapter/Execution/TypeCache.cs#L412) or `.IsAssemblyOrClassCleanupMethod()` in line [416](https://github.com/Microsoft/testfx/blob/master/src/Adapter/MSTest.CoreAdapter/Execution/TypeCache.cs#L416), those method would return false and the property `AssemblyInitializeMethod`/`AssemblyCleanupMethod ` wouldn't cached the method, because the `IsAttributeDefined` method in [line 32](https://github.com/Microsoft/testfx/blob/master/src/Adapter/MSTest.CoreAdapter/Helpers/ReflectHelper.cs#L32) from ReflectHelper is looking for the attributes of the test class, and won't match with `TestClassAttribute` but `DummyTestClassAttribute`.

Please, review and advise.

jayaranigarg:
@parrainc : It seems to work fine for me. Refer this PR : https://github.com/Microsoft/testfx/pull/525

parrainc:
> @parrainc : It seems to work fine for me. Refer this PR : #525

You're correct, I was using OR comparator instead of AND.
Updated myself and seems to work fine.

Should I close this PR and you go ahead and merge #525 ?

jayaranigarg:
@parrainc : Thank you for the contribution. Release build is failing on this PR for some reason. I guess we will just take the other PR.

PR #510, opened 2018-10-29
Allow Task-returning test methods without async keyword

### LINKED ISSUE #500: Asynchronous test methods rejected unless they use the async keyword
## Description
I have a test method that returns a Task. This works when the method also uses the async keyword, but removing the async keyword results in UTA007: Method does not have correct signature. This appears to be intentional based on the IsValidTestMethodShouldReturnFalseForMethodsWithNonVoidReturnType test, but I'm wondering why the caller should care if the method is implemented via async state machine or not.

If it's not intentional, I have a potential fix in [my fork](https://github.com/Microsoft/testfx/compare/master...spanglerco:non-async-task-test) that I can create a PR for.

## Steps to reproduce
```csharp
[TestClass]
public class MyTests
{
    [TestMethod]
    public Task MyTest()
    {
        return Task.CompletedTask;
    }
}
```

## Expected behavior
MyTests.MyTest is discovered as a test to run.

## Actual behavior
UTA007: Method MyTest defined in class MyTests does not have correct signature. Test method marked with the [TestMethod] attribute must be non-static, public, return-type as void  and should not take any parameter. Example: public void Test.Class1.Test(). Additionally, if you are using async-await in test method then return-type must be Task. Example: public async Task Test.Class1.Test2()

## Environment
MSTest 1.3.2

spanglerco:
Enables asynchronous tests that don't use the async state machine while
still rejecting async void methods.

Fixes #500. Discussion on the issue hasn't concluded whether this behavior is intentional or not, but it was requested that I go ahead and open a pull request for it.

### DIFF
--- a/src/Adapter/MSTest.CoreAdapter/Extensions/MethodInfoExtensions.cs
+++ b/src/Adapter/MSTest.CoreAdapter/Extensions/MethodInfoExtensions.cs
@@ -114,8 +114,8 @@ namespace Microsoft.VisualStudio.TestPlatform.MSTest.TestAdapter.Extensions
         /// <returns>True if the method has a void/task return type..</returns>
         internal static bool IsVoidOrTaskReturnType(this MethodInfo method)
         {
-            return method.GetAsyncTypeName() == null ? ReflectHelper.MatchReturnType(method, typeof(void))
-                : ReflectHelper.MatchReturnType(method, typeof(Task));
+            return ReflectHelper.MatchReturnType(method, typeof(Task))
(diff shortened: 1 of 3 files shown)


mayankbansal018:
@AbhitejJohn , @jayaranigarg  do you guys see any issue with this change?

AbhitejJohn:
@spanglerco , @mayankbansal018 : Looks good to me. Looks like this also change Assembly/Class/Test Init/Cleanup. Please make sure that any documentation related is updated as well.

PR #513, opened 2018-11-07
Apply TestCategory from derived class on inherited test methods

### LINKED ISSUE #498: TestCategory on class doesn't apply to inherited test methods
## Description
I have a test class that inherits test methods from a base test class, but I want to apply a TestCategory from the derived class in order to filter tests to run. The current implementation only looks at the TestCategory starting at the class that declares the test method, rather than the class being enumerated. But since TestCategory is inherited, I think it would make sense for it to include all categories starting at the class being enumerated.

I have a working implementation of this in [a fork](https://github.com/Microsoft/testfx/compare/master...spanglerco:category-from-derived-class) and am opening this issue to determine if the change is desired or if it's working as intended.

There's also a question about how assembly-level TestCategory attributes should work in this case. The code currently only considers the assembly containing the test method and not the assemblies containing the test class hierarchy. My fork changes it to only consider the assembly containing the test class being enumerated.

## Steps to reproduce
```csharp
[TestClass]
[TestCategory("base")]
public abstract class Base
{
    [TestMethod]
    public void TestSomeBaseInterfaceMethod() {}
}

[TestClass]
[TestCategory("derived")]
public class Derived : Base {}
```

## Expected behavior
Derived.TestSomeBaseInterfaceMethod has both base and derived test categories:
image

## Actual behavior
Derived.TestSomeBaseInterfaceMethod only has the base test category:
image

## Environment
MSTest 1.3.2

spanglerco:
Enables sharing test methods across a class hierarchy while categorizing the inherited tests based on the derived class that's actually running them. It does this by passing the class type from the `TypeEnumerator` through to the `ReflectHelper` rather than relying on `MemberInfo.DeclaringType`.

Fixes #498. Previous behavior was that inherited methods would only receive test categories starting at the class defining the method. This changes it so inherited methods receive test categories starting at the test class being discovered.

I didn't add any new tests because the updated unit tests for ReflectHelper verify the correct class type argument is passed to ReflectionUtility, and the existing tests for ReflectionUtility verify correct test categories are returned based on the class hierarchy passed in. I have manually verified expected behavior specified by @karanjitsingh in https://github.com/Microsoft/testfx/issues/498#issuecomment-436188960 which is my original use case.


src/Adapter/MSTest.CoreAdapter/Helpers/ReflectHelper.cs
```diff
@@ -365,19 +367,20 @@ internal bool IsDoNotParallelizeSet(Assembly assembly)
...
             if (categories != null)
             {
-                categories = categories.Concat(this.GetCustomAttributes(attributeProvider.DeclaringType.GetTypeInfo(), typeof(TestCategoryBaseAttribute))).ToArray();
+                categories = categories.Concat(this.GetCustomAttributes(owningType.GetTypeInfo(), typeof(TestCategoryBaseAttribute))).ToArray();
             }
 
             if (categories != null)
             {
-                categories = categories.Concat(this.GetCustomAttributeForAssembly(attributeProvider, typeof(TestCategoryBaseAttribute))).ToArray();
+                categories = categories.Concat(this.GetCustomAttributeForAssembly(owningType.GetTypeInfo().Assembly, typeof(TestCategoryBaseAttribute))).ToArray();
```
karanjitsingh: https://docs.microsoft.com/en-us/dotnet/api/system.reflection.memberinfo.declaringtype?view=netframework-4.7.2#System_Reflection_MemberInfo_DeclaringType

It seems the situation you were facing was only because the test method wasn't overridden. From the 4 scenarios in the link, one of the scenarios is this issues particular scenario. Just to be sure can you provide the behavior for the rest of the three scenarios.

I know this seems far fetched, but this could lead to a big breaking change so I just want to make sure :)

karanjitsingh:
@spanglerco, sorry for the late reply. I'll be continuously monitoring this PR from now on until its conclusion.

src/Adapter/MSTest.CoreAdapter/Helpers/ReflectHelper.cs
spanglerco: Good idea. I took the example class hierarchy from that link and added test attributes to get this: https://gist.github.com/spanglerco/fdb04b4a6bab1be2963f6063c937a587

It actually exposes an interesting corner case for classes that hide test methods (not that I can immediately think of a use case for that). Here's the result of using vstest.console to run the tests filtered by one category at a time with 1.4.0:

Category | Test Class Per Execution | Test Method Executed | Total
:--------: | ------------------------ | ----------------------- | :---:
A | <ul><li>A</li><li>B</li><li>B</li><li>C</li><li>D</li></ul> | <ul><li>A.Test</li><li>B.Test</li><li>B.Test</li><li>A.Test</li><li>D.Test</li></ul> | 5
B | <ul><li>B</li></ul> | <ul><li>B.Test</li></ul> | 1
C | -- | -- | 0
D | <ul><li>D</li></ul> | <ul><li>D.Test</li></ul> | 1

The interesting thing is that test category A gets applied to B's `new Test()`, but instead of running both `A.Test()` with an instance of B and `B.Test()`, it runs `B.Test()` twice. But test category B only applies to `B.Test()`, so it runs once for that category. Test category C isn't applied to `A.Test()`, which is my original scenario.

Here's the results with this change:

Category | Test Class Per Execution | Test Method Executed | Total
:--------: | ------------------------ | ----------------------- | :---:
A | <ul><li>A</li><li>B</li><li>B</li><li>C</li><li>D</li></ul> | <ul><li>A.Test</li><li>B.Test</li><li>B.Test</li><li>A.Test</li><li>D.Test</li></ul> | 5
B | <ul><li>B</li><li>B</li></ul> | <ul><li>B.Test</li><li>B.Test</li></ul> | 2
C | <ul><li>C</li></ul> | <ul><li>A.Test</li></ul> | 1
D | <ul><li>D</li></ul> | <ul><li>D.Test</li></ul> | 1

A and D remain the same, but now that the test category of the subclass is applied to inherited methods, test category C includes `A.Test()` on an instance of C. A side-effect is that test category B then includes a second run of `B.Test()` like what happens in test category A.

The reason the duplicate execution happens is because the discoverer only sends the `TestMethod.FullClassName` and not the `TestMethod.DeclaringClassFullName`, so when the executor reacquires the MethodInfo, it does so on B instead of A, making it ambiguous. It simply chooses the first result, which is the `B.Test()` one. Assuming that it's considered backwards compatible to add new property values to the `TestCase`, it seems this could be fixed by sending the declaring class name as an optional property value and the executor can use that to find the correct MethodInfo.

karanjitsingh: @spanglerco, this is really good stuff. The changes look good but can we come up with a solution to avoid running tests multiple times for *overloaded* methods?

Since this is a case for overloading, for test category B the expectation would be that `B.Test` is run only once and since `A.Test` is being overloaded it should not run when running tests for category `B`.

spanglerco: > since A.Test is being overloaded it should not run when running tests for category B

@karanjitsingh I should have read this part more clearly to start with. I have a working implementation that resolves the duplicate execution, but I would have thought the desired behavior would be that test category B would run both `B.Test` and `A.Test` on an instance of B since B has both of those methods defined at the same time (`B.Test` is hiding `A.Test`, but `A.Test` is still there):

Category | Test Class Per Execution | Test Method Executed | Total
:--------: | ------------------------ | ----------------------- | :---:
A | <ul><li>A</li><li>B</li><li>B</li><li>C</li><li>D</li></ul> | <ul><li>A.Test</li><li>B.Test</li><li>A.Test</li><li>A.Test</li><li>D.Test</li></ul> | 5
B | <ul><li>B</li><li>B</li></ul> | <ul><li>B.Test</li><li>A.Test</li></ul> | 2
C | <ul><li>C</li></ul> | <ul><li>A.Test</li></ul> | 1
D | <ul><li>D</li></ul> | <ul><li>D.Test</li></ul> | 1

src/Adapter/MSTest.CoreAdapter/Helpers/ReflectHelper.cs
```diff
@@ -389,18 +392,17 @@ internal IEnumerable<object> GetCustomAttributesRecursively(MemberInfo attribute
...
-        /// Gets the custom attributes on the assembly of a member info
+        /// Gets the custom attributes on an assembly
         /// NOTE: having it as separate virtual method, so that we can extend it for testing.
         /// </summary>
-        /// <param name="memberInfo">The member to inspect.</param>
+        /// <param name="assembly">The assembly to inspect.</param>
         /// <param name="type">The attribute type to find.</param>
         /// <returns>Custom attributes defined.</returns>
-        internal virtual Attribute[] GetCustomAttributeForAssembly(MemberInfo memberInfo, Type type)
+        internal virtual Attribute[] GetCustomAttributeForAssembly(Assembly assembly, Type type)
```
jayaranigarg: GetCustomerAttribute*() methods usulally takes MemberInfo() as argument. Please consider keeping it consistent.

spanglerco: I originally changed it to Assembly because the caller only has a Type. But GetTypeInfo() gives back a MemberInfo, so I put it back to MemberInfo and use memberInfo.Module.Assembly instead.

src/Adapter/MSTest.CoreAdapter/Helpers/ReflectHelper.cs
karanjitsingh: @spanglerco, in the case of this type of overload, the language specifies `A.Test` will be run if the object is of type A or the reference for object B is of type A, there is no right way here in this case but I would want to follow the convention that the language follows.

karanjitsingh: @spanglerco, any update on this?

spanglerco: @karanjitsingh, sorry for the delay, I haven't had a lot of time to look into this lately. I was able to tweak TypeEnumerator to remove the duplicate methods with these results:

Category | Test Class Per Execution | Test Method Executed | Total
:--------: | ------------------------ | ----------------------- | :---:
A | <ul><li>A</li><li>B</li><li>C</li><li>D</li></ul> | <ul><li>A.Test</li><li>B.Test</li><li>A.Test</li><li>D.Test</li></ul> | 4
B | <ul><li>B</li></ul> | <ul><li>B.Test</li></ul> | 1
C | <ul><li>C</li></ul> | <ul><li>A.Test</li></ul> | 1
D | <ul><li>D</li></ul> | <ul><li>D.Test</li></ul> | 1

karanjitsingh:
@spanglerco, some tests are failing. Can you fix?

spanglerco:
@karanjitsingh, it looks like the RunTestsForTestShouldPreferParallelSettingsFromRunSettingsOverAssemblyLevelAttributes test failed in my PR release build, which I'm guessing is intermittent. Is there a way to trigger it to try again? Also, I'm wondering if that test should use something like a ConcurrentDictionary instead of HashSet (I'm just guessing and don't know the history of why the sleeps are there).

jayaranigarg:
@dotnet-bot test Windows / Release Build please

spanglerco:
Thank you for getting around to this one @jayaranigarg

Issue #529, opened 2018-12-06
Description and other attributes not available in TestCase


singhsarab:
## Description
Test attributes like Description, CssIteration, WorkItem, CssProjectStructure are not available in the TestCase object.
There attributes are eventually required to fill in the data in the trx logger and enable the following scenario: https://github.com/Microsoft/vstest/issues/98

PR #538, opened 2018-12-10
Description in TestCase

singhsarab:
Make Description, CssIteration, WorkItem, CssProjectStructure available in properties in the TestCase object.
Fixes : #529


src/Adapter/MSTest.CoreAdapter/Helpers/ReflectHelper.cs
```diff
@@ -414,6 +414,24 @@ internal virtual Attribute[] GetCustomAttributes(MemberInfo attributeProvider, T
...
+        /// Gets the first custom attribute of the provided type on a memberInfo
+        /// </summary>
+        /// <param name="attributeProvider"> The member to reflect on. </param>
+        /// <param name="type"> The attribute type. </param>
+        /// <returns>Attribute defined.</returns>
+        internal virtual Attribute GetCustomAttribute(MemberInfo attributeProvider, Type type)
+        {
+            var attribute = GetCustomAttributes(attributeProvider, type, true);
+
+            if (attribute == null || attribute.Length != 1)
```
jayaranigarg, line 427: What if we have multiple workitem attributes? Do we want to silently ignore/not fill them in property bag?

singhsarab: That's a valid concern. I just followed the same convention that was bein followed for Priority and Owner. https://github.com/Microsoft/testfx/blob/8384164a3187cd586134bfc230c337671300f68d/src/Adapter/MSTest.CoreAdapter/Helpers/ReflectHelper.cs#L599

jayaranigarg: We cannot have multiple owner and priority attributes for a method as per framework restrictions. https://github.com/Microsoft/testfx/blob/master/src/TestFramework/MSTest.Core/Attributes/VSTestAttributes.cs#L188
But we can indeed have multiple workitem attributes. https://github.com/Microsoft/testfx/blob/master/src/TestFramework/MSTest.Core/Attributes/VSTestAttributes.cs#L297

singhsarab: Good catch. Then, we'd ideally want to make workitem property an array of workitems.

singhsarab:
@dotnet-bot test this please

Issue #443, opened 2018-06-06
Cryptic error message when DynamicData doesn't have any data


jayaranigarg:
>>Pulling in from [vstest](https://github.com/Microsoft/vstest/issues/1608)

## Description
Using DynamicDataAttribute without any data produces a cryptic error message:

> Failed   _(test name)_
> Error Message:
>  Error in executing test. No result returned by extension. If using extension of
>  TestMethodAttribute then please contact vendor.

I'm less concerned with what the correct behavior should be here and more concerned with providing developers with meaningful errors. It was very confusing to figure out why my test was failing after my IEnumerable<object[]> was processed but before the first line of code was reached. Note: my DynamicDaa was a .Where() clause on different test data; the repro in this issue is a simplified version of my original test.

I don't think that a DynamicData test without any data should be considered a "passing test", but I hesitate to call it a "failing test."

## Steps to reproduce
1. Run the command:
`dotnet new mstest`
2. Update the nuget packages:
* MSTest.TestFramework 1.2.0 -> 1.3.0
* MSTest.TestAdapter 1.2.0 -> 1.3.0
* Microsoft.Net.TestSdk 15.6.0 -> 15.7.2
3. Replace the contents of UnitTest1.cs with:
```
using Microsoft.VisualStudio.TestTools.UnitTesting;
using System.Collections.Generic;

[TestClass]
public class UnitTest1
{
    public static IEnumerable<object[]> Empty => new object[][]{};
    [TestMethod]
    [DynamicData(nameof(Empty))]
    public void Object_IsNotNull(object o) => Assert.IsNotNull(o);
}
```
4. Run the command:
`dotnet test`

## Expected behavior
Something to the effect of:

> Build started, please wait...
> Build completed.
> 
> Test run for D:\(masked)\bin\Debug\netcoreapp2.0\repro.dll(.NETCoreApp,Version
> =v2.0)
> Microsoft (R) Test Execution Command Line Tool Version 15.6.0
> Copyright (c) Microsoft Corporation.  All rights reserved.
> 
> Starting test execution, please wait...
> **Skipped  Object_IsNotNull
> Error Message:
>  DynamicDataAttribute specified an IEnumerable<object[]> to use for test parameters, but the enumerable was empty.**
> 
> 
> Total tests: 1. Passed: 0. Failed: 0. Skipped: 1.
> Test Run Successful.
> Test execution time: 1.7951 Seconds

## Actual behavior

> Build started, please wait...
> Build completed.
> 
[...]

PR #544, opened 2018-12-18
Correcting error message when dynamic data doen't have any data

jayaranigarg:
#443

### DIFF
--- a/src/TestFramework/MSTest.Core/Attributes/DynamicDataAttribute.cs
+++ b/src/TestFramework/MSTest.Core/Attributes/DynamicDataAttribute.cs
@@ -129,11 +129,19 @@ namespace Microsoft.VisualStudio.TestTools.UnitTesting
 
             IEnumerable<object[]> enumerable = obj as IEnumerable<object[]>;
 
-            if (enumerable == null || !enumerable.Any())
+            if (enumerable == null)
             {
                 throw new ArgumentNullException(
                     string.Format(
                         FrameworkMessages.DynamicDataIEnumerableNull,
+                        this.dynamicDataSourceName,
+                        this.dynamicDataDeclaringType.FullName));
+            }
+            else if (!enumerable.Any())
+            {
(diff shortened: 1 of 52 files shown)


src/TestFramework/MSTest.Core/Attributes/DynamicDataAttribute.cs
```diff
@@ -128,7 +128,8 @@ public IEnumerable<object[]> GetData(MethodInfo methodInfo)
             }
 
             IEnumerable<object[]> enumerable = obj as IEnumerable<object[]>;
-            if (enumerable == null)
+
+            if (enumerable == null || !enumerable.Any())
```
singhsarab: It should be ArgumentException and not ArgumentNullException in this new case.

test/UnitTests/MSTest.Core.Unit.Tests/Attributes/DynamicDataAttributeTests.cs
```diff
@@ -293,7 +306,18 @@ public static IEnumerable<object[]> NullProperty
...
+            get
+            {
+                return new object[][] { };
+            }
+        }
+
+        /// <summary>
+        /// Gets the wrong test data property.
         /// </summary>
         public static IEnumerable<object[]> WrongDataTypeProperty
```
singhsarab: Out of the scope of this PR maybe. Why do we WrongDataTypeProperty? How is that different from Null property.

jayaranigarg: I think the intent was different but it is not very clear. Correcting it.

jamers99:
When will this be release, or what version is this fix in?

jayaranigarg:
@jamers99 : version 2.0.0-beta4 has this fix. Checkout the release notes : https://github.com/microsoft/testfx/releases/tag/v2.0.0-beta4

PR #555, opened 2019-01-24
MissingMethodException Fix

vagisha-nidhi:
When running a test project compiled for netstandard, and running in full framework throws the following error:
 Error: System.MissingMethodException: Method not found: 'System.Collections.Generic.IDictionary`2<System.String,System.Object> Microsoft.VisualStudio.TestTools.UnitTesting.TestContext.get_Properties()'..

It is expecting IDictionary`2<System.String,System.Object> hence causing run time error.

### DIFF
--- a/src/Adapter/PlatformServices.Desktop/Services/DesktopTestContextImplementation.cs
+++ b/src/Adapter/PlatformServices.Desktop/Services/DesktopTestContextImplementation.cs
@@ -118,11 +118,11 @@ namespace Microsoft.VisualStudio.TestPlatform.MSTestAdapter.PlatformServices
         }
 
         /// <inheritdoc/>
-        public override IDictionary Properties
+        public override IDictionary<string, object> Properties
         {
             get
             {
-                return this.properties as IDictionary;
+                return this.properties as IDictionary<string, object>;
             }
         }
 
--- a/src/Adapter/PlatformServices.Desktop/Services/DesktopTestContextImplementation.cs
+++ b/src/Adapter/PlatformServices.Desktop/Services/DesktopTestContextImplementation.cs
@@ -122,7 +122,7 @@ namespace Microsoft.VisualStudio.TestPlatform.MSTestAdapter.PlatformServices
         {
             get
             {
-                return this.properties as IDictionary<string, object>;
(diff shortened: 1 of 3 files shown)


src/Adapter/PlatformServices.Desktop/Services/DesktopTestContextImplementation.cs
```diff
@@ -118,11 +118,11 @@ public override DataRow DataRow
         }
 
         /// <inheritdoc/>
-        public override IDictionary Properties
+        public override IDictionary<string, object> Properties
         {
             get
             {
-                return this.properties as IDictionary;
+                return this.properties as IDictionary<string, object>;
```
acesiddhu: >is.properties as IDictionary<string, object [](start = 25, length = 43)

just return this.properties.

src/TestFramework/Extension.Desktop/TestContext.cs
```diff
@@ -19,7 +19,7 @@ public abstract class TestContext
         /// <summary>
         /// Gets test properties for a test.
         /// </summary>
-        public abstract IDictionary Properties { get; }
+        public abstract IDictionary<string, object> Properties { get; }
```
acesiddhu, line 22: >abstract [](start = 15, length = 8)

I am afraid if this is a breaking change?

acesiddhu: if folks have taken a depedency in their code for this property and they expect IDict then build will fail

---
In reply to: [250584995](https://github.com/Microsoft/testfx/pull/555#discussion_r250584995) [](ancestors = 250584995)

src/Adapter/PlatformServices.Desktop/Services/DesktopTestContextImplementation.cs
vagisha-nidhi: Done

src/TestFramework/Extension.Desktop/TestContext.cs
vagisha-nidhi: We will release a beta package and see if get any such feedback on this change.

Issue #143, opened 2017-04-07
Class Initialize inheritance


pvlakshm:
## Description
Allow a ClassInitialize in a base class of a TestClass so that in addition to initializing static variables in the test class, I can initialize static variables in the test base class.

UV item: https://visualstudio.uservoice.com/forums/121579-visual-studio/suggestions/2311366-support-classinitialize-in-base-class

MikeChristensen:
I believe this works if the test class and base class are in the same assemblies.  I'm curious if issue #22 fixed this behavior.

AbhitejJohn:
This works for `TestInitialize` today. `ClassInitialize` however is not inherited from base classes. We only execute the `ClassInitialize` of that specific class. The fix for this should just be plugging in similar functionality for Class Initialize/Cleanup as we have for Test Initialize/Cleanup [here](https://github.com/Microsoft/testfx/blob/master/src/Adapter/MSTest.CoreAdapter/Execution/TypeCache.cs#L465). FWIW, this should be the default workflow. However, since this would be a change in behavior from v1 mstest, we would also want to add a setting to turn this off as appropriate.

MikeChristensen:
That sounds great!  Though it would be fairly easy to work-around for now, since your test class [ClassInitialize] static method could just manually call your base class [ClassInitialize] static method.  Using a static constructor would also be a work around.  Would love to see this "bug" fixed though!  I've been working on a lot of unit tests lately that could really use this pattern.

TheXby:
Hello
Its also happened for "AssemblyInitialize" attribute and mentioned problem not happened for all Unit testing frameworks (nUnit, xUnit, MbUnit....)
I will be very grateful if it will be fixed soon.
Thanks.

arnonax:
I'm very much pro this idea! 
However, there are some details that must be flushed out first:
First, there's the backward compatibility issue that @AbhitejJohn mentioned. But in addition, I can think of at least 2 possible outcomes  of inheriting ClassInitialize:
1. ClassInitialize will be called only once before *any* derived class
2. ClassInitialize will be called before *each* of the derived classes.
I believe that each of these outcomes may be desired in different situations.

Therefore I suggest to add an optional enum parameter to the ClassInitializeAttribute with the following possible values:
* `[ClassInitialize]` or `[ClassInitialize(InheritanceBehavior=ClassInitializeInheritance.None)]` - ClassInitialize is called only for tests defined in the same class, as in V1.
[...]

PR #577, opened 2019-03-10
Class Initialize inheritance #143

parrainc:
Implemented feature requested in #143 

- Create enum to handle the inheritance option in ClassInitialize attribute
- Set None option to default in ClassInitialize attribute
- Added properties on TestClassInfo class
- Updated RunClassInitialize and RunClassCleanup method in TestClassInfo class
- Updated CreateClassInfo method on TypeCache class to check for classinitialize inheritance option
- Added a few unit tests for TestClassInfo and TypeCache

example output:
**===>** Init from **ParentTestClass**
**==>** Init from **BaseTestClass**
**=>** Init from **DerivedTestClass**

**=>** Cleaning up from **DerivedTestClass**
**==>** Cleaning up from **BaseTestClass**
**===>** Cleaning up from **ParentTestClass**


parrainc:
cc: @AbhitejJohn @jayaranigarg @singhsarab
Please, review...

parrainc:
Pushed new changes to fix that base cleanup were being called without calling base initialize

src/Adapter/MSTest.CoreAdapter/Execution/TestClassInfo.cs
```diff
@@ -107,6 +111,11 @@ internal set
         /// </summary>
         public bool IsClassInitializeExecuted { get; internal set; }
 
+        /// <summary>
+        /// Gets a dictionary of test initialize methods indicating if they have been executed.
```
jayaranigarg: Nit: dictionary of **class** initialize methods

src/Adapter/MSTest.CoreAdapter/Execution/TestClassInfo.cs
```diff
@@ -157,6 +166,16 @@ public bool HasExecutableCleanupMethod
             }
         }
 
+        /// <summary>
+        /// Gets a queue of test initialize methods to call for this type.
+        /// </summary>
+        public Queue<MethodInfo> BaseClassInitializeMethodsQueue { get; private set; }
```
jayaranigarg: Nit: queue of **class** initialize

src/Adapter/MSTest.CoreAdapter/Execution/TestClassInfo.cs
```diff
@@ -219,19 +238,47 @@ internal set
         [SuppressMessage("Microsoft.Design", "CA1031:DoNotCatchGeneralExceptionTypes", Justification = "Requirement is to handle all kinds of user exceptions and message appropriately.")]
         public void RunClassInitialize(TestContext testContext)
         {
-            // If no class initialize return
-            if (this.ClassInitializeMethod == null)
+            // If no class initialize and no base class initialize, return
+            if (this.ClassInitializeMethod is null)
             {
-                return;
+                if (this.BaseClassInitializeMethodsQueue.Count == 0)
```
AbhitejJohn: use `this.BaseClassInitializeMethodsQueue.Any()` instead to make this check faster.

AbhitejJohn: nit: Can we || this with the if above?

src/Adapter/MSTest.CoreAdapter/Execution/TypeCache.cs
```diff
@@ -308,6 +308,24 @@ private TestClassInfo CreateClassInfo(Type classType, TestMethod testMethod)
...
                         this.UpdateInfoIfTestInitializeOrCleanupMethod(classInfo, methodInfo, true, instanceMethods, testInitializeAttributeType, testCleanupAttributeType);
                     }
+
+                    if (this.IsAssemblyOrClassInitializeMethod(methodInfo, classInitializeAttributeType))
+                    {
+                        if (methodInfo.GetCustomAttribute<ClassInitializeAttribute>().InheritanceBehavior == ClassInitializeInheritance.None)
+                        {
+                            // if the ClassInitializeInheritance is None, means
+                            // it will pick only the derived class initalize method
+                            break;
```
AbhitejJohn: hmm, we probably might not want to break here. This would effect the TestInitialize methodinfo gathering above.

src/Adapter/MSTest.CoreAdapter/Execution/TypeCache.cs
```diff
@@ -308,6 +308,24 @@ private TestClassInfo CreateClassInfo(Type classType, TestMethod testMethod)
                         // Update test initialize/cleanup method from base type.
                         this.UpdateInfoIfTestInitializeOrCleanupMethod(classInfo, methodInfo, true, instanceMethods, testInitializeAttributeType, testCleanupAttributeType);
                     }
+
+                    if (this.IsAssemblyOrClassInitializeMethod(methodInfo, classInitializeAttributeType))
```
AbhitejJohn: similar to the above check, maybe we could limit this only to public and static methods.

src/TestFramework/MSTest.Core/Attributes/VSTestAttributes.cs
```diff
@@ -25,6 +23,24 @@ public enum TestTimeout
         Infinite = int.MaxValue
     }
 
+    /// <summary>
+    /// Enumeration for class initialize, that can be used with the <see cref="ClassInitializeAttribute"/> class.
+    /// Defines the behavior of the ClassInitialize methods of base classes.
+    /// The type of the enumeration must match
+    /// </summary>
+    public enum ClassInitializeInheritance
```
AbhitejJohn: I'm wondering if we should have this on Cleanup as well? Might be easy to miss this specification on initialize if they aren't near by and the naming seems to be specific for initialize although in code we implicitly stop performing cleanup too.

AbhitejJohn: @singhsarab , @jayaranigarg  : thoughts?

src/Adapter/MSTest.CoreAdapter/Execution/TestClassInfo.cs
```diff
@@ -219,19 +238,47 @@ internal set
...
+            MethodInfo initializeMethod = null;
+            try
+            {
+                // ClassInitialize methods for base classes are called in reverse order of discovery
+                // Base -> Child TestClass
+                var baseClassInitializeStack = new Stack<MethodInfo>(this.BaseClassInitializeMethodsQueue);
+                while (baseClassInitializeStack.Count > 0)
+                {
+                    initializeMethod = baseClassInitializeStack.Pop();
+                    if (initializeMethod.GetCustomAttribute<ClassInitializeAttribute>().InheritanceBehavior
```
AbhitejJohn: looks like we already have this check when we add methodInfo's to the queue. We could skip this check altogether.

src/Adapter/MSTest.CoreAdapter/Execution/TestClassInfo.cs
```diff
@@ -219,19 +238,47 @@ internal set
...
+                    }
+                }
+            }
+            catch (Exception ex)
+            {
+                this.ClassInitializationException = ex;
+            }
+
+            // If class initialization is not done and class initialize method is not null, then do it.
+            if (!this.IsClassInitializeExecuted && this.classInitializeMethod != null)
```
AbhitejJohn: So shouldn't the logic above also be guarded by the `IsClassInitializeExecuted` so it executes only once?

AbhitejJohn: and maybe we do not want to run the current classes initialize if any of the ones above failed. `ClassInitializationException` is non null.

src/Adapter/MSTest.CoreAdapter/Execution/TestClassInfo.cs
```diff
@@ -271,15 +318,13 @@ public void RunClassInitialize(TestContext testContext)
...
-            StackTraceInformation exceptionStackTraceInfo = null;
-            if (!realException.TryGetUnitTestAssertException(out outcome, out errorMessage, out exceptionStackTraceInfo))
+            if (!realException.TryGetUnitTestAssertException(out outcome, out string errorMessage, out StackTraceInformation exceptionStackTraceInfo))
             {
                 errorMessage = string.Format(
                     CultureInfo.CurrentCulture,
                     Resource.UTA_ClassInitMethodThrows,
                     this.ClassType.FullName,
-                    this.ClassInitializeMethod.Name,
+                    this.ClassInitializeMethod?.Name ?? initializeMethod.Name,
```
AbhitejJohn: hmm, we might want to store the method that failed and report the name here. This would report the childs method failed even if the actual failure happened in one of the base classes.

src/Adapter/MSTest.CoreAdapter/Execution/TestClassInfo.cs
```diff
@@ -298,19 +343,33 @@ public void RunClassInitialize(TestContext testContext)
...
         /// </returns>
-        [SuppressMessageAttribute("Microsoft.Design", "CA1031:DoNotCatchGeneralExceptionTypes", Justification = "Requirement is to handle all kinds of user exceptions and message appropriately.")]
+        [SuppressMessage("Microsoft.Design", "CA1031:DoNotCatchGeneralExceptionTypes", Justification = "Requirement is to handle all kinds of user exceptions and message appropriately.")]
         public string RunClassCleanup()
         {
-            if (this.ClassCleanupMethod == null)
+            if (this.ClassCleanupMethod is null)
             {
-                return null;
+                if (this.BaseClassCleanupMethodsQueue.Count == 0)
```
AbhitejJohn: similar to the above. .Any() and can merge with the if on top.

src/Adapter/MSTest.CoreAdapter/Execution/TestClassInfo.cs
```diff
@@ -298,19 +343,33 @@ public void RunClassInitialize(TestContext testContext)
...
                 try
                 {
-                    this.ClassCleanupMethod.InvokeAsSynchronousTask(null);
+                    classCleanupMethod?.InvokeAsSynchronousTask(null);
+                    var baseClassCleanupQueue = new Queue<MethodInfo>(this.BaseClassCleanupMethodsQueue);
+                    while (baseClassCleanupQueue.Count > 0)
+                    {
+                        classCleanupMethod = baseClassCleanupQueue.Dequeue();
+                        if (this.BaseClassInitializeMethodsDict.Any(method => method.Key.DeclaringType
+                                == classCleanupMethod.DeclaringType && method.Value))
```
AbhitejJohn: This would probably not work in the following case. We wouldn't call the base class cleanup cause we haven't run its initialize.
```
class A
{
 [ClassCleanup]
public static void CC() {}
}

public class B : A
{
[ClassInitialize]
public static void DCI(TestContext tc) {}
[ClassCleanup]
public static void DCC() {}
}
```

AbhitejJohn: I like that we aren't running cleanup for initializes that haven't happened but this would probably only work correctly if we keep track of the base types hierarchy.

AbhitejJohn: It might be ok to just run Cleanups blind like TestCleanups to start with. Not my favorite choice but should be ok to start with.

src/Adapter/MSTest.CoreAdapter/Execution/TestClassInfo.cs
```diff
@@ -48,8 +49,11 @@ public class TestClassInfo
             this.ClassType = type;
             this.Constructor = constructor;
             this.TestContextProperty = testContextProperty;
+            this.BaseClassInitializeMethodsDict = new Dictionary<MethodInfo, bool>();
```
AbhitejJohn: nit: maybe rename this to be `ExecutedBaseClassInitializeMethods` and declare it as a HashSet<string> since we only have a DeclaringType check?

src/TestFramework/MSTest.Core/Attributes/VSTestAttributes.cs
jayaranigarg: Agree. We should not have name specific to initialize when are using it for cleanup as well. 
Having this enum for cleanup as well would be good but in that case should we have a check that `ClassInitializeInheritance` is same for initialize and cleanup?

AbhitejJohn: Did not quite get the last part.

src/Adapter/MSTest.CoreAdapter/Execution/TypeCache.cs
parrainc: I was leveraging the use of the `methodInfo ` extension method (`HasCorrectClassOrAssemblyInitializeSignature`) inside the `this.IsAssemblyOrClassInitilizeMethod()`

It's already doing the check there to see if the method satisfies the right signature. So, in that case it would be a duplicated check for static and public. What do you think?

src/Adapter/MSTest.CoreAdapter/Execution/TestClassInfo.cs
parrainc: fixed! thanks.

src/Adapter/MSTest.CoreAdapter/Execution/TestClassInfo.cs
parrainc: fixed! thanks.

src/Adapter/MSTest.CoreAdapter/Execution/TestClassInfo.cs
parrainc: Done!

parrainc: done, but actually, it should be && instead, because we want to execute base class initialize even though the class initialize method is null. Otherwise, we just don't run any class initialize.

src/Adapter/MSTest.CoreAdapter/Execution/TestClassInfo.cs
parrainc: true, already removed!

src/Adapter/MSTest.CoreAdapter/Execution/TestClassInfo.cs
parrainc: agree, already updated.

src/Adapter/MSTest.CoreAdapter/Execution/TestClassInfo.cs
parrainc: > So shouldn't the logic above also be guarded by the `IsClassInitializeExecuted` so it executes only once?

Given the way base initialize class methods are being executed: `Parent -> Child -> etc...` I think we shouldn't. I make sure to pop the element from the `baseClassInitializeStack`, so every time is passes through the while loop, it'll has -1 element.

Does it make sense?

EDIT:
After taking a second look, I think it would be wise to first make sure the `IsClassInitializeExecuted`. So, if it is not, then the base classes will go through.


> and maybe we do not want to run the current classes initialize if any of the ones above failed. `ClassInitializationException` is non null.

you're absolutely right here. Added `this.ClassInitializationException` as part of the if check for the initialize method on the not parent test class.

src/Adapter/MSTest.CoreAdapter/Execution/TestClassInfo.cs
parrainc: yup, that's true. missed that part. Created a local variable to hold the failed initialize method name.

src/Adapter/MSTest.CoreAdapter/Execution/TestClassInfo.cs
parrainc: Got it, done!

src/Adapter/MSTest.CoreAdapter/Execution/TypeCache.cs
parrainc: hmm, right it could be. Changed to continue instead which i think make more sense.
make sense?

src/Adapter/MSTest.CoreAdapter/Execution/TestClassInfo.cs
parrainc: well, not mine either honestly. I think that in order to be symmetrical with initialize the clean up should only be executed when its initialize has happened. 

I think that the if check:
`this.ExecutedBaseClassInitializeMethods.Any(method => method.DeclaringType == classCleanupMethod.DeclaringType)` will only let cleanup to execute if both initialize and cleanup method share same declaring type.

I'm open to make the change either way. Let me know what you think.

src/TestFramework/MSTest.Core/Attributes/VSTestAttributes.cs
parrainc: well, I'm totally open to change the name to something like: `InheritanceBehavior` maybe?
I think that since initialize and cleanup are related, so their execution is/should be symmetrical and there would no need to specify the attribute in the cleanup as well, it'll do it the same fashion as initialize.

And, if we want cleanup always to be initialize (or at least every time the initialize is done) we shouldn't change this by using this. Plus, we avoid additional checks.

let me know your thoughts...

parrainc:
@dotnet-bot test Windows / Release Build please

src/Adapter/MSTest.CoreAdapter/Execution/TestClassInfo.cs
AbhitejJohn: I have another idea - how about storing init and cleanup as tuples? That way cases like the ones above can be accounted for as well. You'd only execute cleanups whose corresponding initialize tuple has been executed without any exceptions.

src/Adapter/MSTest.CoreAdapter/Execution/TypeCache.cs
AbhitejJohn: well that first seems to be checking if the Assembly/ClassInitialize attribute is defined on the method. Attribute retrieval is generally costly and should be avoided if we can.

src/Adapter/MSTest.CoreAdapter/Execution/TypeCache.cs
AbhitejJohn: I imagine you'd set a flag to not process any more Class Initializes/Class Cleanups within that loop.

src/TestFramework/MSTest.Core/Attributes/VSTestAttributes.cs
AbhitejJohn: I agree to the most part. I'm just not certain how common the case I described earlier is - Just cleanups without initialize and would users need some form of control in those cases as well? I'm inclined to have it on both. The way I see it, we hopefully would default to always use Base Class Init/Cleanup functionality soon (a major release) and not many folks would need to switch back unless its an unrecoverable break. If they do they could use as much flexibility as they can get.

src/Adapter/MSTest.CoreAdapter/Execution/TestClassInfo.cs
parrainc: it sounds good. one question, how different would be using a dictionary instead of a tuple? will we gain any advantage for using it?

src/TestFramework/MSTest.Core/Attributes/VSTestAttributes.cs
parrainc: That would be great having that in a near future with the default value to `BeforeEachDerivedClass`, that way the base init and cleanup methods always execute without having to do any code change (from users perspective). On regards to the case you mentioned, given the way the current behavior is (just base class cleanup are executed only if its initialize has been executed) I don't think users would need to do anything here. Unless we change that behavior and let class cleanup method to execute even though an initialize haven't been specified.

src/Adapter/MSTest.CoreAdapter/Execution/TypeCache.cs
parrainc: hmm, that's true. I've changed this and also in previous lines when checking for init/cleanup in derived class. And, since this was similar in both derived class init/cleanup method and base, I extracted a method similar to `UpdateInfoIfTestInitializeOrCleanupMethod` but for class initialize/cleanup. That way we check for the class initialize/cleanup attribute only if the method is public and static.

I'll be pushing those changes after other comments are resolved.

src/Adapter/MSTest.CoreAdapter/Execution/TypeCache.cs
parrainc: This was affected as part of the changes mentioned in below comment. 
So, instead of checking if the attribute is None, changed to see if the attribute is BeforeEachDerivedClass, then added the base method to the queue. Otherwise, nothing will be done and TestInitialize methodinfo gathering won't be affected at all by breaking or anything.

I'll push this changes after other review comments are resolved.

src/Adapter/MSTest.CoreAdapter/Execution/TestClassInfo.cs
AbhitejJohn: So the tuple would help you associate Init,Cleanup pairs so cases like the above where you have a cleanup but no init would still work(you;d just push in a null init with the cleanup). You'd probably still need some form a linear data structure(a stack should work) to determine what cleanups to run.

sturlath:
So when can we expect this to be merged? I really need to be able to use [ClassInitialize] in my custom base class and if I'm correct this will make that possible. Right?

parrainc:
@sturlath 
> So when can we expect this to be merged? I really need to be able to use [ClassInitialize] in my custom base class and if I'm correct this will make that possible. Right?

From my side, I can tell you that since I've been kind of busy, i'm working a little bit slower than before on this, but even though, I think by end of next week I'll be done with this (unless there's more review to solve).

This changes will let you tell the framework to take into account your custom base class' initialize method when running your tests. So, yeah, you're correct.

I think @jayaranigarg and @AbhitejJohn could expand a bit more when this would be released after merged into master.

src/Adapter/MSTest.CoreAdapter/Execution/TestClassInfo.cs
parrainc: hey @AbhitejJohn 
latest changes i pushed cover what we discussed through the review comments.
Please, take a look at the code and let me know if further changes are required, otherwise, I think we should be good.

src/TestFramework/MSTest.Core/Attributes/VSTestAttributes.cs
parrainc: --
updated in latest changes.

src/Adapter/MSTest.CoreAdapter/Execution/TestClassInfo.cs
```diff
@@ -230,8 +243,38 @@ public void RunClassInitialize(TestContext testContext)
...
-            if (!this.IsClassInitializeExecuted)
+            MethodInfo initializeMethod = null;
+            string failedClassInitializeMethodName = string.Empty;
+
+            try
+            {
+                // If class initialization is not done, we can proceed to perform
+                // base class initialization first
+                if (!this.IsClassInitializeExecuted)
+                {
```
AbhitejJohn, line 266: nit: might be cleaner to just return from this method if `isClassInitializeExecuted`

src/Adapter/MSTest.CoreAdapter/Execution/TestClassInfo.cs
```diff
@@ -230,8 +243,38 @@ public void RunClassInitialize(TestContext testContext)
...
+                    // ClassInitialize methods for base classes are called in reverse order of discovery
+                    // Base -> Child TestClass
+                    var baseClassInitializeStack = new Stack<Tuple<MethodInfo, MethodInfo>>(this.BaseClassInitAndCleanupMethods);
+                    while (baseClassInitializeStack.Count > 0)
+                    {
+                        initializeMethod = baseClassInitializeStack.Pop().Item1;
+                        if (initializeMethod != null)
+                        {
+                            initializeMethod?.InvokeAsSynchronousTask(null, testContext);
```
AbhitejJohn: We do not need this null check operator here given that we are already performing a null check above.

src/Adapter/MSTest.CoreAdapter/Execution/TestClassInfo.cs
```diff
@@ -230,8 +243,38 @@ public void RunClassInitialize(TestContext testContext)
...
+            {
+                this.ClassInitializationException = ex;
+                failedClassInitializeMethodName = initializeMethod.Name;
+            }
+
+            // If class initialization is not done and class initialize method is not null,
+            // and class initialization exception is null, then do it.
+            if (!this.IsClassInitializeExecuted && this.classInitializeMethod != null && this.ClassInitializationException == null)
             {
                 // Aquiring a lock is usually a costly operation which does not need to be
```
AbhitejJohn, line 291: Does the base class init need the same locking mechanism that the current class init uses below? Would be useful when tests are run in parallel.

src/Adapter/MSTest.CoreAdapter/Execution/TestClassInfo.cs
```diff
@@ -298,19 +340,26 @@ public void RunClassInitialize(TestContext testContext)
         /// <returns>
         /// Any exception that can be thrown as part of a class cleanup as warning messages.
         /// </returns>
-        [SuppressMessageAttribute("Microsoft.Design", "CA1031:DoNotCatchGeneralExceptionTypes", Justification = "Requirement is to handle all kinds of user exceptions and message appropriately.")]
+        [SuppressMessage("Microsoft.Design", "CA1031:DoNotCatchGeneralExceptionTypes", Justification = "Requirement is to handle all kinds of user exceptions and message appropriately.")]
```
AbhitejJohn, line 355: thanks 👍

src/Adapter/MSTest.CoreAdapter/Execution/TestClassInfo.cs
```diff
@@ -298,19 +340,26 @@ public void RunClassInitialize(TestContext testContext)
...
         public string RunClassCleanup()
         {
-            if (this.ClassCleanupMethod == null)
+            if (this.ClassCleanupMethod is null && !this.BaseClassInitAndCleanupMethods.Any(p => p.Item2 != null))
```
AbhitejJohn, line 358: not that this would affect the code here but [this](https://stackoverflow.com/questions/40676426/what-is-the-difference-between-x-is-null-and-x-null) is an interesting discussion on the difference between `== null` and `is null`

src/Adapter/MSTest.CoreAdapter/Execution/TestClassInfo.cs
```diff
@@ -298,19 +340,26 @@ public void RunClassInitialize(TestContext testContext)
...
             {
                 return null;
             }
 
-            if (this.IsClassInitializeExecuted || this.ClassInitializeMethod == null)
+            if (this.IsClassInitializeExecuted || this.ClassInitializeMethod is null || this.ExecutedBaseClassInitializeMethods.Any())
             {
+                var classCleanupMethod = this.classCleanupMethod;
```
AbhitejJohn: nit: just to be consistent with other usages in this method, lets use the property here(`ClassCleanupMethod`).

src/Adapter/MSTest.CoreAdapter/Execution/TestClassInfo.cs
```diff
@@ -298,19 +340,26 @@ public void RunClassInitialize(TestContext testContext)
...
                 try
                 {
-                    this.ClassCleanupMethod.InvokeAsSynchronousTask(null);
+                    classCleanupMethod?.InvokeAsSynchronousTask(null);
+                    var baseClassCleanupQueue = new Queue<Tuple<MethodInfo, MethodInfo>>(this.BaseClassInitAndCleanupMethods);
+                    while (baseClassCleanupQueue.Count > 0)
+                    {
+                        classCleanupMethod = baseClassCleanupQueue.Dequeue().Item2;
+                        classCleanupMethod?.InvokeAsSynchronousTask(null);
```
AbhitejJohn, line 373: shouldn't we be only executing cleanups whose inits have run? This seems to be running all cleanups.

src/TestFramework/MSTest.Core/Attributes/VSTestAttributes.cs
```diff
@@ -164,6 +195,20 @@ public sealed class ClassInitializeAttribute : Attribute
     [AttributeUsage(AttributeTargets.Method, AllowMultiple = false)]
     public sealed class ClassCleanupAttribute : Attribute
     {
+        /// <summary>
+        /// Initializes a new instance of the <see cref="ClassCleanupAttribute"/> class.
+        /// ClassCleanupAttribute
+        /// </summary>
+        /// <param name="inheritanceBehavior">none</param>
+        public ClassCleanupAttribute(InheritanceBehavior inheritanceBehavior = InheritanceBehavior.None)
```
AbhitejJohn: A more friendly back compat change would be to add two ctors here - the default and the one that takes in InheritanceBehavior enum. What we currently have here would break binary compat.

src/Adapter/MSTest.CoreAdapter/Execution/TestClassInfo.cs
AbhitejJohn: Maybe we could just push Cleanups that need to be executed into a Stack instead of `ExecutedBaseClassInitializeMethods`

src/Adapter/MSTest.CoreAdapter/Execution/TestClassInfo.cs
parrainc: Indeed, pretty interesting discussion. When I first saw the new (not that new already) `is` expression I read some posts about what differentiates it from the `==` on null comparison and was kind of similar to this discussion except for the part that they didn't show how resulted IL code is. Thanks for sharing that out :)

And yeah, as you first pointed out this wouldn't affect because no equality overloaded is in the code checking if the type we're is null.

src/Adapter/MSTest.CoreAdapter/Execution/TestClassInfo.cs
parrainc: true, i was following same pattern as few lines below but that way seems little bit cleaner. Changes coming up in next push.

src/Adapter/MSTest.CoreAdapter/Execution/TestClassInfo.cs
parrainc: actually, i think we might needed it since the null check above is to make sure there are any non-null init method in the tuple that can be executed. In any case, I changed this for a `.Where` call when newing up the `baseClassInitializeStack `(on the `this.BaseClassInitAndCleanupMethods`) that way, we just pass executable init methods, and we don't need to do any extra thing to pass its corresponding cleanup method to the stack for cleanup methods to execute in the `RunClassCleanup` method.

src/Adapter/MSTest.CoreAdapter/Execution/TestClassInfo.cs
parrainc: You're right. Didn't considered it in the first place when changing the RunClassInitialize method for base class init.
So, taking the same locking logic for the base class init methods, considering parallel runs. thanks for that catch.

src/Adapter/MSTest.CoreAdapter/Execution/TestClassInfo.cs
parrainc: gotcha! thanks!

src/Adapter/MSTest.CoreAdapter/Execution/TestClassInfo.cs
parrainc: Yup, I had a block of code that I removed right before pushing the latest changes, handling the logic for deciding those cleanup method that will be executed. But, now that you mentioned that last part, I think that'd be better. For that, renamed the property you mentioned, now it's called: `BaseClassCleanupMethodsStack `not sure if the name is 100% ok, and in the `RunClassInitialize `method, I'll be pushing the cleanups into that stack to then execute them in the `RunClassCleanup `method.

src/TestFramework/MSTest.Core/Attributes/VSTestAttributes.cs
parrainc: Done, added one more non-param ctor in both ClassInitialize and Cleanup attrs.

src/Adapter/MSTest.CoreAdapter/Execution/TestClassInfo.cs
AbhitejJohn: It looks like the updated code wouldn't run Cleanups from base classes that do not have an initialize method...

src/Adapter/MSTest.CoreAdapter/Execution/TestClassInfo.cs
AbhitejJohn: Sounds good. It doesn't look like we need a Queue here. We could just pop the methods off the stack..

paulmccormack:
Hi

Do you know when this change will be merged and released?

Cheers

Paul

AbhitejJohn:
I'd like to see this in too.
/cc: @acesiddhu , @singhsarab .

NGloreous:
@parrainc  Any updates on this change? Look like there hasn't been progress for a bit. I'm in need of this and would be happy to help if needed.

AbhitejJohn:
So, I've checked offline with the team on getting this in. The plan is to do the needful to merge this in early next week. I believe @parrainc has addressed most of the feedback here and I'd love to have this in the product.
/cc: @cltshivash , @jayaranigarg

parrainc:
I think most of the comments (if not, all of them) were addressed. I've been a little busy but if something still missing I could fix it, so it could be merged. So, basically, I'm waiting on @jayaranigarg @cltshivash inputs. If this looks good for them as well, then we should be good.

jayaranigarg:
@sturlath @paulmccormack @NGloreous You can consume this change using following nuget packet versions:
https://dotnet.myget.org/feed/mstestv2/package/nuget/MSTest.TestAdapter/2.0.0-build-20190708-01
https://dotnet.myget.org/feed/mstestv2/package/nuget/MSTest.TestFramework/2.0.0-build-20190708-01

elgatov:
@parrainc must classinitialize be static for it to run? i do not have it marked static and it will not run from a base class

parrainc:
> @parrainc must classinitialize be static for it to run? i do not have it marked static and it will not run from a base class

@elgatov yes, it must has this signature:
 ```
[ClassInitialize(UTF.InheritanceBehavior.BeforeEachDerivedClass)]
public static void InitBaseClassMethod(TestContext testContext) { }
```

Issue #549, opened 2019-01-07
French resources in Italian


mikelapierre:
## Description
>The following French resource file contains Italian translations:
https://github.com/Microsoft/testfx/blob/master/src/Adapter/PlatformServices.Desktop/Resources/xlf/Resource.fr.xlf.
>Translations are the same as the Italian resource file:
https://github.com/Microsoft/testfx/blob/master/src/Adapter/PlatformServices.Desktop/Resources/xlf/Resource.it.xlf

## Steps to reproduce
>Run tests generating messages in a French environment.

## Expected behavior
>The messages should be in French.

## Actual behavior
>The messages are in Italian.

vagisha-nidhi:
@v-zbsail Can you please look at this?

janbaltus:
Fixed the #549 issue and updated PR

PR #581, opened 2019-03-13
LOC CHECKIN | Microsoft/testfx master | 20190313

v-zbsail:


jayaranigarg:
@v-zbsail : Was there any specific reason for moving file contents up and down? It is very difficult to track which strings were changed/translated in this PR.

jayaranigarg:
Also, as part of this PR can you please address this bug as well : https://github.com/Microsoft/testfx/issues/549

janbaltus:
Fixed the #549 issue and updated PR

PR #590, opened 2019-03-28
Generated the xlf files for the CancellationToken changes

cltshivash:
Generated the xlf files for the CancellationToken changes


jayaranigarg:
Loc team has recently moved to a new tool and these things are happening after that. They said it would be a one time thing.

PR #593, opened 2019-04-05
Bugfix for CancellationTokenSource

cltshivash:
Added initialization that was missing in the NetCoreTestContext and added a test to cover the net core scenario as well.

### DIFF
--- a/test/E2ETests/TestAssets/TimeoutTestProjectNetCore/TimeoutTestProjectNetCore.csproj
+++ b/test/E2ETests/TestAssets/TimeoutTestProjectNetCore/TimeoutTestProjectNetCore.csproj
@@ -1,0 +1,17 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+  <PropertyGroup>
+    <TestFxRoot Condition="$(TestFxRoot) == ''">..\..\..\..\</TestFxRoot>
+  </PropertyGroup>
+
+  <PropertyGroup>
+    <TargetFramework>net452</TargetFramework>
+    <IsPackable>false</IsPackable>
+    <AppendTargetFrameworkToOutputPath>false</AppendTargetFrameworkToOutputPath>
+    <OutputPath>$(TestFxRoot)artifacts\TestAssets\</OutputPath>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <ProjectReference Include="$(TestFxRoot)\src\TestFramework\Extension.Desktop\Extension.Desktop.csproj" />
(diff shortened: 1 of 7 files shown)


test/E2ETests/TestAssets/TimeoutTestProjectNetCore/TimeoutTestProjectNetCore.csproj
```diff
@@ -0,0 +1,17 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+  <PropertyGroup>
+    <TestFxRoot Condition="$(TestFxRoot) == ''">..\..\..\..\</TestFxRoot>
+  </PropertyGroup>
+
+  <PropertyGroup>
+    <TargetFramework>net452</TargetFramework>
+    <IsPackable>false</IsPackable>
```
ankit-nitd-zz, line 9: Shouldn't this project have target framework of net core type?

cltshivash: Thanks for catching this. Fixed it.

test/E2ETests/TestAssets/TimeoutTestProjectNetCore/TimeoutTestProjectNetCore.csproj
```diff
@@ -0,0 +1,22 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+  <PropertyGroup>
+    <TestFxRoot Condition="$(TestFxRoot) == ''">..\..\..\..\</TestFxRoot>
+  </PropertyGroup>
+
+  <PropertyGroup>
+    <TargetFramework>netcoreapp1.1</TargetFramework>
+    <FrameworkIdentifier>NetCore</FrameworkIdentifier>
+    <IsPackable>false</IsPackable>
+    <!--<AppendTargetFrameworkToOutputPath>true</AppendTargetFrameworkToOutputPath>-->
```
jayaranigarg, line 10: If we make this as false, then we will not need to specify 'netcoreapp1.1' in TimeoutTestAssemblyNetCore path here:https://github.com/Microsoft/testfx/blob/37b1ebf822786771fbcf641cfa84a39f3c83204b/test/E2ETests/Smoke.E2E.Tests/TimeoutTests.cs#L14

cltshivash: Thought about it. But it is better to have a separate folder per test assembly.

PR #610, opened 2019-04-29
Setting MapNotRunnableToFailed to true by default

vagisha-nidhi:
Setting MSTest settings option MapNotRunnableToFailed to true by default. This will lead all tests that cannot be executed to result in failed outcome.

### DIFF
--- a/test/UnitTests/MSTest.CoreAdapter.Unit.Tests/ObjectModel/UnitTestResultTests.cs
+++ b/test/UnitTests/MSTest.CoreAdapter.Unit.Tests/ObjectModel/UnitTestResultTests.cs
@@ -292,7 +292,7 @@ namespace Microsoft.VisualStudio.TestPlatform.MSTestAdapter.UnitTests.ObjectMode
         }
 
         [TestMethod]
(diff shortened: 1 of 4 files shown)


test/UnitTests/MSTest.CoreAdapter.Unit.Tests/ObjectModel/UnitTestResultTests.cs
```diff
@@ -292,11 +292,12 @@ public void UniTestHelperToTestOutcomeForUnitTestOutcomeErrorShouldReturnTestOut
         }
 
         [TestMethod]
-        public void UniTestHelperToTestOutcomeForUnitTestOutcomeNotRunnableShouldReturnTestOutcomeNoneWhenNotSpecified()
+        public void UniTestHelperToTestOutcomeForUnitTestOutcomeNotRunnableShouldReturnTestOutcomeNoneWhenSpecified()
```
singhsarab: Please change the name for this test.
Should start with UnitTestResults* and just specified, does not make sense. Make sure you call out it's set to false.

singhsarab: Similar for the test below.

singhsarab:
I want us to be careful with this change, please make sure it is available as part of the beta release

Issue #617, opened 2019-06-06
Assert.IsNotInstanceOfType fails on value null


Saeras:
## Description
Assert.IsInstanceOfType succeeds when value is null, while the MSDN specifies it should throw an error.
AssertFailedException : 'value is null or expectedType is not found in the inheritance hierarchy of value.'
This behavior was broken under issue #178 where Assert.IsNotInstanceOfType was incorrectly changed to fail when the value is null.

## Steps to reproduce
Create a standard .net core test project from visual studio 2017, and create a dummy class. then use the following method:
`public void TestMethod1() { Assert.IsNotInstanceOfType(null, typeof(Dummy)); }`

## Expected behavior
The test should succeed

## Actual behavior
The test throws AssertFailedException and fails

## Environment
Windows 10 version 1709 (build 16299.125)
Visual Studio version 15.9.11
MSTest.TestAdapter version 1.3.2
MSTest.TestFramework version 1.3.2

ShreyasRmsft:
@Saeras thank you for reporting this. We will be fixing this.

ShreyasRmsft:
Fixed in PR https://github.com/microsoft/testfx/pull/620

PR #620, opened 2019-06-17
fixed IsNotInstanceOfType failing when objected being asserted on is null

ShreyasRmsft:
Fixed IsNotInstanceOfType failing when objected being asserted on is null

Null is not an instance of any type.

### DIFF
--- a/src/TestFramework/MSTest.Core/Assertions/Assert.cs
+++ b/src/TestFramework/MSTest.Core/Assertions/Assert.cs
@@ -1754,9 +1754,14 @@ namespace Microsoft.VisualStudio.TestTools.UnitTesting
         /// </exception>
         public static void IsNotInstanceOfType(object value, Type wrongType, string message, params object[] parameters)
         {
-            if (wrongType == null || value == null)
+            if (wrongType == null)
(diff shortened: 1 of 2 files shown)


ShreyasRmsft:
Adding tests in a bit

src/TestFramework/MSTest.Core/Assertions/Assert.cs
```diff
@@ -1754,10 +1754,15 @@ public static void IsNotInstanceOfType(object value, Type wrongType, string mess
         /// </exception>
         public static void IsNotInstanceOfType(object value, Type wrongType, string message, params object[] parameters)
         {
-            if (wrongType == null || value == null)
+            if (wrongType == null)
```
singhsarab, line 1757: Why was this required ? Do we have an issue ?

ShreyasRmsft: Yes. Also the functionality was clearly wrong.

ShreyasRmsft:
Issue https://github.com/microsoft/testfx/issues/617

Issue #464, opened 2018-07-23
MSTest V2 and data driven tests


AbhitejJohn:
## Description
Filing this to better the experience for MSTestV2's data driven tests in VS IDE. Currently with 1.3.2 version of the Fx/Adapter pair here is how the tests are displayed:

![image](https://user-images.githubusercontent.com/18014426/43106155-54667ade-8e8c-11e8-8a22-f4e715057ce8.png)

Here are the problems that this notation has:
1. One does not have a way of running/debugging an individual test data. (#171 details this as well)
2. The summary test result that started showing up with the latest package isn't much of a help since the TE already has a node indicating that one of the tests has failed. 

For the first issue, if mstest starts associating each data row with a separate test case object that should start showing tests similar to how nunit/xunit does today which is more granular and consistent with the Test Explorers tree view. Open to suggestions on better ways of achieving this.
The second issue is kind of tied to the first. If we must send out a summary result though, is there a way that clients can filter this out? Is there a property that marks these results?

/cc: @jayaranigarg , @pvlakshm , @cltshivash

michael-hawker:
Just found out about `DataTestMethod` and `DataRow`, it's super powerful, but it would be great if the individual data rows appeared underneath the test case.

Basically it'd be great if the test case became a new parent in the tree in order to:
1. Better navigate the results of individual test scenarios (and be able to jump to the DataRow attribute where it was declared on double-click like any other test)
2. Providing easy access to the individual scenario's result of expected and actual
3. Providing a click/UI point to be able to run/debug a specific row

Open Questions:
1. Should each DataRow test count towards the number of tests? For consistency in bubbling the total number of test-cases, I think yes.
2. Changed the above to have a `-` separator as keeping the name as `DataRow(1,2)` seems like it would lead to confusion with the parenthesis being used everywhere else to indicate number of tests.

I.e.
- UnitTestProject20 (2 tests) 1 failed
  - UnitTestProject20 (2)
    - UnitTestProject20 (2)
      - UnitTest1 (2)
        - TestMethod1 (2)
          - DataRow - 1,2
          - DataRow - 1, 1

nohwnd:
@michael-hawker @peterwald Is this someting you should take into consideration in your FQN proposal? 

I thought this is already happening. That is, we already produce a test case per datarow, are 1) and 2) "only" matter of presenting it differently?

[...]

Issue #340, opened 2017-12-27
Test Adapter: parametrized tests are discovered/logged to data collector as 1 test (no information about specific parameters)


pkruk2:
## Description
Hi, 

MsTest tests may be parametrizable, for example a method with DataTestMethod and DataRow attributes. For example:
```
        [DataTestMethod]
        [DataRow("baz", false)]
        [DataRow("foo", true)]
        public void TestMethod(string name, bool expectedResult)
```

## Current behavior

For such test, MsTest Test Adapter:
1. during test discovery, only 1 VSTest TestCase is discovered (which represents whole TestMethod)
2. during test execution, the method is executed for data row, but VSTest's DataCollectors (both In-Proc and out of process) receive only 1 TestCaseStart/TestCaseStop event (for TestMethod)

## Desired behavior

Would it be possible to have per data row discovery/logging resolution? That is, ideally:
1. discovery would discover 2 test cases: TestMethod("baz", false) and TestMethod("foo", true)
2. DataCollector would receive TestCaseStart/TestCaseStop for both above test cases

In fact, the 2nd point is more important for me - I would like to change environment and monitor each execution of TestMethod (for each data row separately).

If it is not possible, could you please share the reason? 

Thanks

## Environment
mstest.testadapter.1.2.0
microsoft.testplatform.15.6.0-preview-20171211-02

cltshivash:
@pkruk2 Having separate tests for each datarow is not in plan for the near future. Would it be possible to check the parameter value in your test code (or) does this need to be done outside the test code ?

pkruk2:
@cltshivash It must be outside of the test code.

In case I need to add this myself to the test adapter, do you have any suggestions? Ideally it'd be accepted upstream.

I understand it'd also change VS behavior, because currently when you discover/run such parametrized test, VS shows 1 test in Test Explorer, and all data rows results are visible in test details section.

jayaranigarg:
Tagging @pvlakshm , @cltshivash to see if and how can we bring this in?

AbhitejJohn:
@jayaranigarg : Another reason why it would be really nice to have separate test cases is in #171. To be able to run/debug an individual test case would greatly improve dev efficiency especially if these are long running tests.

ShreyasRmsft:
@pkruk2 this is by design and it is a non-goal to have discovery by data row.

[...]

PR #631, opened 2019-08-16
Record test start/end events for data driven tests

NGloreous:
This is a fix for #340 so that data collectors can get start/end events for individual execution of data driven tests. Also solves to allow the custom test execution extension to run a test multiple times and get multiple start/end events. 

Solving for #464 might be a better approach but seems like a much bigger change. 

One side effect of this change is ClassInitialize happens before any start event gets fired. IMO this is fine, it always seemed weird that this execution would be lumped into the first test that happened to execute. This also make it so ClassInitialize is now identical to ClassCleanup which doesn't get captured.

### DIFF
--- a/src/Adapter/MSTest.CoreAdapter/Execution/TestExecutionManager.cs
+++ b/src/Adapter/MSTest.CoreAdapter/Execution/TestExecutionManager.cs
@@ -154,6 +154,7 @@ namespace Microsoft.VisualStudio.TestPlatform.MSTest.TestAdapter.Execution
                 return;
             }
 
+            Guid orginalTestCaseId = test.Id;
             foreach (var unitTestResult in unitTestResults)
             {
                 if (test == null)
@@ -161,6 +162,7 @@ namespace Microsoft.VisualStudio.TestPlatform.MSTest.TestAdapter.Execution
                     continue;
                 }
 
+                test.Id = unitTestResult.TestId == Guid.Empty ? orginalTestCaseId : unitTestResult.TestId;
                 var testResult = unitTestResult.ToTestResult(test, startTime, endTime, MSTestSettings.CurrentSettings);
 
                 if (unitTestResult.DatarowIndex >= 0)
@@ -183,6 +185,8 @@ namespace Microsoft.VisualStudio.TestPlatform.MSTest.TestAdapter.Execution
                     // Ignore this exception
                 }
             }
+
+            test.Id = orginalTestCaseId;
         }
 
         private static bool MatchTestFilter(ITestCaseFilterExpression filterExpression, TestCase test, TestMethodFilter testMethodFilter)
--- a/src/Adapter/MSTest.CoreAdapter/Execution/TestExecutionManager.cs
+++ b/src/Adapter/MSTest.CoreAdapter/Execution/TestExecutionManager.cs
@@ -165,7 +165,7 @@ namespace Microsoft.VisualStudio.TestPlatform.MSTest.TestAdapter.Execution
                 test.Id = unitTestResult.TestId == Guid.Empty ? orginalTestCaseId : unitTestResult.TestId;
                 var testResult = unitTestResult.ToTestResult(test, startTime, endTime, MSTestSettings.CurrentSettings);
 
-                if (unitTestResult.DatarowIndex >= 0 && string.IsNullOrEmpty(unitTestResult.DisplayName))
+                if (unitTestResult.DatarowIndex >= 0)
                 {
                     testResult.DisplayName = string.Format(CultureInfo.CurrentCulture, Resource.DataDrivenResultDisplayName, test.DisplayName, unitTestResult.DatarowIndex);
                 }
--- a/src/Adapter/MSTest.CoreAdapter/Execution/TestExecutionManager.cs
+++ b/src/Adapter/MSTest.CoreAdapter/Execution/TestExecutionManager.cs
@@ -172,6 +172,11 @@ namespace Microsoft.VisualStudio.TestPlatform.MSTest.TestAdapter.Execution
                     testResult.DisplayName = string.Format(CultureInfo.CurrentCulture, Resource.DataDrivenResultDisplayName, test.DisplayName, unitTestResult.DatarowIndex);
                 }
 
+                // Fire a RecordEnd here even though we also fire RecordEnd inside TestMethodInfo
+                // this is to ensure we fire end events for error cases. This is acceptable because
+                // vstest ignores multiple end events.
+                testExecutionRecorder.RecordEnd(test, testResult.Outcome);
+
                 if (testResult.Outcome == TestOutcome.Failed)
                 {
                     PlatformServiceProvider.Instance.AdapterTraceLogger.LogInfo("MSTestExecutor:Test {0} failed. ErrorMessage:{1}, ErrorStackTrace:{2}.", testResult.TestCase.FullyQualifiedName, testResult.ErrorMessage, testResult.ErrorStackTrace);
@@ -369,6 +374,11 @@ namespace Microsoft.VisualStudio.TestPlatform.MSTest.TestAdapter.Execution
                 }
 
                 var unitTestElement = currentTest.ToUnitTestElement(source);
+
+                // Fire a RecordStart here even though we also fire RecordStart inside TestMethodInfo
+                // this is to ensure we fire start events for error cases. This is acceptable because
+                // vstest ignores multiple start events.
+                testExecutionRecorder.RecordStart(currentTest);
 
                 var startTime = DateTimeOffset.Now;
 
--- a/src/Adapter/MSTest.CoreAdapter/Execution/TestExecutionManager.cs
+++ b/src/Adapter/MSTest.CoreAdapter/Execution/TestExecutionManager.cs
@@ -154,7 +154,7 @@ namespace Microsoft.VisualStudio.TestPlatform.MSTest.TestAdapter.Execution
                 return;
             }
 
-            Guid orginalTestCaseId = test.Id;
+            Guid originalTestCaseId = test.Id;
             foreach (var unitTestResult in unitTestResults)
             {
                 if (test == null)
@@ -162,7 +162,9 @@ namespace Microsoft.VisualStudio.TestPlatform.MSTest.TestAdapter.Execution
                     continue;
                 }
 
-                test.Id = unitTestResult.TestId == Guid.Empty ? orginalTestCaseId : unitTestResult.TestId;
+                // If the result has a TestId set then update the TestCase to have that id
+                // this is done for data driven results so we can fire unique start/end events for each iteration
+                test.Id = unitTestResult.TestId == Guid.Empty ? originalTestCaseId : unitTestResult.TestId;
                 var testResult = unitTestResult.ToTestResult(test, startTime, endTime, MSTestSettings.CurrentSettings);
 
                 if (unitTestResult.DatarowIndex >= 0)
@@ -186,7 +188,8 @@ namespace Microsoft.VisualStudio.TestPlatform.MSTest.TestAdapter.Execution
                 }
             }
 
-            test.Id = orginalTestCaseId;
+            // Reset the test id back to the original id in case we changed it above
+            test.Id = originalTestCaseId;
         }
 
         private static bool MatchTestFilter(ITestCaseFilterExpression filterExpression, TestCase test, TestMethodFilter testMethodFilter)
--- a/src/Adapter/MSTest.CoreAdapter/Execution/TestExecutionManager.cs
+++ b/src/Adapter/MSTest.CoreAdapter/Execution/TestExecutionManager.cs
@@ -168,8 +168,6 @@ namespace Microsoft.VisualStudio.TestPlatform.MSTest.TestAdapter.Execution
                     testResult.DisplayName = string.Format(CultureInfo.CurrentCulture, Resource.DataDrivenResultDisplayName, test.DisplayName, unitTestResult.DatarowIndex);
                 }
 
-                testExecutionRecorder.RecordEnd(test, testResult.Outcome);
-
                 if (testResult.Outcome == TestOutcome.Failed)
                 {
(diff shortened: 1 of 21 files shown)


src/Adapter/MSTest.CoreAdapter/ObjectModel/TestMethod.cs
```diff
@@ -44,6 +46,16 @@ public TestMethod(string name, string fullClassName, string assemblyName,  bool
             this.IsAsync = isAsync;
         }
 
+        internal TestMethod(TestMethodInfo testMethodInfo)
```
singhsarab, line 53: Comment on when this gets used

src/Adapter/MSTest.CoreAdapter/Execution/TestExecutionManager.cs
```diff
@@ -154,22 +154,22 @@ internal void SendTestResults(TestCase test, UnitTestResult[] unitTestResults, D
                 return;
             }
 
+            Guid orginalTestCaseId = test.Id;
```
singhsarab: >orginalTestCaseId [](start = 17, length = 17)

nit: original*

src/Adapter/MSTest.CoreAdapter/Execution/TestExecutionManager.cs
```diff
@@ -185,6 +185,8 @@ internal void SendTestResults(TestCase test, UnitTestResult[] unitTestResults, D
                     // Ignore this exception
                 }
             }
+
+            test.Id = orginalTestCaseId;
```
singhsarab: Please put comments above to explain why this is required.
This is being done only for the record result ?

src/Adapter/MSTest.CoreAdapter/Execution/TestMethodRunner.cs
```diff
@@ -145,6 +145,7 @@ internal UnitTestResult[] Execute()
 
             if (isIgnoreAttributeOnClass || isIgnoreAttributeOnMethod)
             {
+                this.testMethodInfo.TestExecutionRecorder.RecordStartAndEnd(new UnitTestElement(this.test), UnitTestOutcome.Ignored);
```
singhsarab: Why do we need to log these ?

src/Adapter/MSTest.CoreAdapter/Execution/UnitTestRunner.cs
```diff
@@ -80,17 +87,19 @@ internal UnitTestResult[] RunSingleTest(TestMethod testMethod, IDictionary<strin
                     testContext.SetOutcome(TestTools.UnitTesting.UnitTestOutcome.InProgress);
 
                     // Get the testMethod
-                    var testMethodInfo = this.typeCache.GetTestMethodInfo(testMethod, testContext, MSTestSettings.CurrentSettings.CaptureDebugTraces);
+                    var testMethodInfo = this.typeCache.GetTestMethodInfo(testMethod, testContext, testExecutionRecorder, MSTestSettings.CurrentSettings.CaptureDebugTraces);
 
                     // If the specified TestMethod could not be found, return a NotFound result.
                     if (testMethodInfo == null)
                     {
+                        testExecutionRecorder.RecordStartAndEnd(new UnitTestElement(testMethod), UnitTestOutcome.NotFound);
```
singhsarab: I am not able to understand the use case for this.

src/Adapter/MSTest.CoreAdapter/Execution/TestMethodRunner.cs
NGloreous: @singhsarab strictly speaking logging the start/end events for these error cases probably isn't useful but I was trying to keep parity with the existing logic I'm replacing. One thing I should check is if VSTEST will log the result if no start/end event happens (I think it will though). If you are fine with the slight behavior change then I can remove these.

NGloreous: @singhsarab let me know what you think about this. I've addressed all other comments.

singhsarab: @NGloreous I had a chat with @jayaranigarg and she has certain concerns regarding the current implementation.

src/Adapter/MSTest.CoreAdapter/Execution/TestExecutionRecorderWrapper.cs
```diff
@@ -0,0 +1,55 @@
...
+    using System.Diagnostics;
+    using Microsoft.VisualStudio.TestPlatform.MSTest.TestAdapter.Helpers;
+    using Microsoft.VisualStudio.TestPlatform.MSTest.TestAdapter.ObjectModel;
+    using Microsoft.VisualStudio.TestPlatform.ObjectModel;
+    using Microsoft.VisualStudio.TestPlatform.ObjectModel.Adapter;
+
+    /// <summary>
+    /// Wraps calls to ITestExecutionRecorder using types that can be serialized across AppDomains.
+    /// </summary>
+    internal class TestExecutionRecorderWrapper : MarshalByRefObject
```
jayaranigarg, line 16: Unsure about the purpose of this class. Why couldn't we leverage the existing ITestExecutionRecorder  itself?

src/Adapter/MSTest.CoreAdapter/Execution/TestMethodRunner.cs
jayaranigarg: @NGloreous : Can you explain a bit about keeping parity with the existing logic? I don't think recording testStartAndEnd for such error/exception cases should be required, since the wrapper over this class i.e. TestExecutionManager.cs is already taking care of handling Start and End.

src/Adapter/MSTest.CoreAdapter/Execution/TestMethodRunner.cs
```diff
@@ -159,6 +160,11 @@ internal UnitTestResult[] Execute()
                         this.testMethodInfo.Parent.Parent.RunAssemblyInitialize(this.testContext.Context);
                         this.testMethodInfo.Parent.RunClassInitialize(this.testContext.Context);
                     }
+                    catch
+                    {
```
jayaranigarg: Same here. Not required.

src/Adapter/MSTest.CoreAdapter/Execution/UnitTestRunner.cs
jayaranigarg: Same for this class. Not required to record here.

src/Adapter/MSTest.CoreAdapter/Execution/TestMethodRunner.cs
NGloreous: @jayaranigarg TestExecutionManager.cs no longer fires start/end events which is the whole point of my change (to move the start/end event firing closer to where the test actually executes so we capture these events for each data driven iteration).

src/Adapter/MSTest.CoreAdapter/Execution/TestExecutionManager.cs
```diff
@@ -154,22 +154,22 @@ internal void SendTestResults(TestCase test, UnitTestResult[] unitTestResults, D
...
 
+                test.Id = unitTestResult.TestId == Guid.Empty ? orginalTestCaseId : unitTestResult.TestId;
                 var testResult = unitTestResult.ToTestResult(test, startTime, endTime, MSTestSettings.CurrentSettings);
 
                 if (unitTestResult.DatarowIndex >= 0)
                 {
                     testResult.DisplayName = string.Format(CultureInfo.CurrentCulture, Resource.DataDrivenResultDisplayName, test.DisplayName, unitTestResult.DatarowIndex);
                 }
 
-                testExecutionRecorder.RecordEnd(test, testResult.Outcome);
```
NGloreous, line 171: @jayaranigarg as you can see I removed the RecordEnd here

src/Adapter/MSTest.CoreAdapter/Execution/TestExecutionManager.cs
```diff
@@ -363,7 +366,6 @@ private void ExecuteTestsInSource(IEnumerable<TestCase> tests, IRunContext runCo
                 }
 
                 var unitTestElement = currentTest.ToUnitTestElement(source);
-                testExecutionRecorder.RecordStart(currentTest);
```
NGloreous, line 366: @jayaranigarg and here I removed RecordStart.

src/Adapter/MSTest.CoreAdapter/Execution/TestExecutionRecorderWrapper.cs
NGloreous: @jayaranigarg the existing recorder can't be marshaled across app domains (notice this extends from MarshalByRefObject).

src/Adapter/MSTest.CoreAdapter/Execution/TestMethodInfo.cs
```diff
@@ -137,9 +151,13 @@ public virtual TestResult Invoke(object[] arguments)
 
             using (LogMessageListener listener = new LogMessageListener(this.TestMethodOptions.CaptureDebugTraces))
             {
+                var testElement = new UnitTestElement(new TestMethod(this)) { TestId = this.TestId };
+
                 watch.Start();
                 try
                 {
+                    this.TestExecutionRecorder.RecordStart(testElement);
```
NGloreous, line 159: @jayaranigarg this is where I moved firing the start event to. This happens right before we invoke test setup/execute/cleanup. This ensures we fire an event for each iteration of a data driven tests. It also happens after the custom test execution extension is called which means if someone writes an extension to run the same test 10 times we will fire 10 start/end events (which IMO is expected).

src/Adapter/MSTest.CoreAdapter/Execution/TestMethodInfo.cs
```diff
@@ -162,7 +180,12 @@ public virtual TestResult Invoke(object[] arguments)
                         result.LogError = listener.StandardError;
                         result.TestContextMessages = this.TestMethodOptions.TestContext.GetAndClearDiagnosticMessages();
                         result.ResultFiles = this.TestMethodOptions.TestContext.GetResultFiles();
+                        result.TestId = this.TestId;
                     }
+
+                    this.TestExecutionRecorder.RecordEnd(
```
NGloreous, line 186: @jayaranigarg and here is where the RecordEnd event got moved to.

src/Adapter/MSTest.CoreAdapter/Execution/TestMethodRunner.cs
jayaranigarg: @NGloreous : In the process of removing start/end events from TestExecutionManager.cs, we might end up missing certain scenarios. Moreover, adding RecordStartAndEnd() at so many places doesn't feel right to me.
To have start/end events for data driven testing, I would recommend adding events in TestMethdRunner.cs [here](https://github.com/microsoft/testfx/blob/master/src/Adapter/MSTest.CoreAdapter/Execution/TestMethodRunner.cs#L315) and [here](https://github.com/microsoft/testfx/blob/master/src/Adapter/MSTest.CoreAdapter/Execution/TestMethodRunner.cs#L257). I would let TestExecutionManager.cs still fire start/end events which can act as parent results and firing from TestMethodRunner.cs can act as child results.

NGloreous: @jayaranigarg letting TestExecutionManager.cs still fire start/end events for the parent result seems reasonable. Firing events for data driven tests at the location you suggest doesn't solve for the case that a custom test execution extension runs a test multiple times (we plan to leverage this extension to do test retry and want a start/end event for each try). What are your concerns with the current placement of start/end event firing inside TestMethodInfo.cs? Please advise and I will update the PR.

jayaranigarg: @NGloreous : Good catch about custom test execution extension. In that case, let's keep the start/end event firing inside TestMethodInfo.cs itself.
Let TestExecutionManager.cs still fire start and end events for parent result, which will also cover any error scenarios. So we can safely remove any event firing from TestMethodRunner.cs.

NGloreous: Will do. Thanks for the feedback. I'll wok on this later in the week and update the PR.

src/Adapter/MSTest.CoreAdapter/Execution/TestExecutionManager.cs
NGloreous: Added back

src/Adapter/MSTest.CoreAdapter/Execution/TestExecutionManager.cs
NGloreous: Added back.

test/UnitTests/MSTest.CoreAdapter.Unit.Tests/Execution/TestExecutionManagerTests.cs
```diff
@@ -148,13 +148,9 @@ public void RunTestsForASingleTestShouldSendSingleResult()
...
             this.TestExecutionManager.RunTests(tests, this.runContext, this.frameworkHandle, new TestRunCancellationToken());
 
-            List<string> expectedTestCaseStartList = new List<string>() { "PassingTest" };
-            List<string> expectedTestCaseEndList = new List<string>() { "PassingTest:Passed" };
-            List<string> expectedResultList = new List<string>() { "PassingTest  Passed" };
-
-            CollectionAssert.AreEqual(expectedTestCaseStartList, this.frameworkHandle.TestCaseStartList);
-            CollectionAssert.AreEqual(expectedTestCaseEndList, this.frameworkHandle.TestCaseEndList);
-            CollectionAssert.AreEqual(expectedResultList, this.frameworkHandle.ResultsList);
+            CollectionAssert.Contains(this.frameworkHandle.TestCaseStartList, "PassingTest");
```
NGloreous, line 151: @jayaranigarg I added back the start/end events in TestExecutionManager. The side effect is that we now have duplicate start/end events for non-data driven tests. It's a bit odd but shouldn't effect execution because vstest ignores duplicate events. I don't think there is much I can do here because TestMethodInfo doesn't really  know if the test is being executed more than once.

Issue #606, opened 2019-04-22
Make TestFx repo CPS based


jayaranigarg:
## CPS backlog
This should address other related issues like -

- [ ] Running build.cmd locally results in failure [#605](https://github.com/Microsoft/testfx/issues/605)
- [ ] [NetCore] Enable support for messages/warnings related to Deployment Items for languages other than English [#591](https://github.com/Microsoft/testfx/issues/591)
- [ ] Consolidate the adapter and framework nuget packages (with framework being the longterm one) [#193](https://github.com/Microsoft/testfx/issues/193)
- [ ] Set DisableAppDomain to true, by default

jayaranigarg:
@cltshivash : Please update this checklist as we uncover/identify more issues that needs to be resolved as part of CPS change.

parrainc:
@jayaranigarg just wondering... will be PR #411 taken into consideration as part of this feature? Or maybe it's out of scope?

jayaranigarg:
@parrainc : We are still working on the design aspect of the CPS change, so mostly PR #411 will not be considered.

parrainc:
Oh, I see. Just wanted to know if still worth it to invest time on #411 but i think is not.
Thanks for the update. I'll be checking at the issues to see if there more things I can help to get the repo CPS :)

StingyJack:
What is a CPS?

parrainc:
@StingyJack late but, CPS stands for: Common Project System ([check here](https://github.com/microsoft/vsprojectsystem))

Evangelink:
I will move forward by closing this ticket because it seems like most issues have already been handled.

Issue #605, opened 2019-04-20
Running build.cmd locally results in failure.


mitchdenny:
## Description
I just cloned down the netfx repository to have a go at implementing #579 and when I ran build.cmd I got the following error.

## Steps to reproduce
Environment: Visual Studio 2017 and 2019 installed.

1. Clone repo.
2. Open VS2019 developer command-prompt.
3. Run build.cmd

```
C:\Code\testfx>build
```

## Expected behavior
I expected a successful initial build before I started changing anything.

## Actual behavior
Build failed with:

```
... Perform-Restore: Started.
...     Starting toolset restore...
All packages listed in packages.config are already installed.
Error parsing solution file at C:\Code\testfx\TestFx.sln: The system cannot find the file specified. (Exception from HRESULT: 0x80070002)
The restore failed with an exit code of '1'.
At C:\Code\testfx\scripts\build.ps1:166 char:5
+     throw "The restore failed with an exit code of '$lastExitCode'."
+     ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : OperationStopped: (The restore fai...it code of '1'.:String) [], RuntimeException
    + FullyQualifiedErrorId : The restore failed with an exit code of '1'.
```

## Environment

```
C:\Users\mitch> ver
Microsoft Windows [Version 10.0.18877.1000]
```

Both Visual Studio 2017 and 2019 installed. I noticed by spelunking through build.ps1 and the files it references that is looking in the worng place for MSBuild (at least for VS2019).

mayankbansal018:
can you please run you build.cmd with -verbose, & share those with us, it should list out in detail that from where did it pick msbuild, & what error was thrown.

jayaranigarg:
@mitchdenny : This is a known issue https://github.com/Microsoft/testfx/issues/596 and comes when we have both VS2017 and VS2019 installed on the machine. 
For now, the only workaround would be to uninstall VS2019. This is being tracked as part of this issue: https://github.com/Microsoft/testfx/issues/606 , which is planned for this quarter.

singhsarab:
We will be tracking this with CPS change. Closing.

PR #641, opened 2019-09-21
Fix build script to work with VS2019

jimmylewis:
The main change is to try to find msbuild using the 'Current' version
instead of the '15.0' version.  The MSBuild team has stated that
'Current' will be used going forward to avoid a breaking change to the
path when the version revs.

Also necessary in this change was to drop the /toolsVersion flag from
msbuild.  It was passing 15.0 but that tools version is not recognized in the
current MSBuild.

Lastly, added the UWP.Support component as a required build component.
When the build script picks an instance of VS missing this feature, the build
will fail when trying to import Microsoft.Windows.UI.Xaml.CSharp.Targets.

### DIFF
--- a/scripts/common.lib.ps1
+++ b/scripts/common.lib.ps1
@@ -54,8 +54,18 @@ function Locate-MSBuild($hasVsixExtension = "false") {
 
 function Locate-MSBuildPath($hasVsixExtension = "false") {
   $vsInstallPath = Locate-VsInstallPath -hasVsixExtension $hasVsixExtension
-  $msbuildPath = Join-Path -path $vsInstallPath -childPath "MSBuild\$msbuildVersion\Bin"
-  return Resolve-Path -path $msbuildPath
+
+  # first try to find the VS2019+ path
+  $msbuildPath = Join-Path -path $vsInstallPath -childPath "MSBuild\Current\Bin"
+  $msbuildPath = Resolve-Path $msbuildPath
+
+  # otherwise fall back to the VS2017 path
+  if(!(Test-Path -path $msbuildPath)) {
+    $msbuildPath = Join-Path -path $vsInstallPath -childPath "MSBuild\$msbuildVersion\Bin"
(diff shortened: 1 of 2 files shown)


jimmylewis:
I submitted this as a possible fix for #605 before waiting for all the tasks in #606 to be completed.  This issue has been filed a few times and is a pain point trying to debug the test framework or make other contributions.

singhsarab:
@jimmylewis The changes seem to have broken the build. Can you please take a look

singhsarab:
There you go https://dev.azure.com/vstestplatform/TestFx/_build/results?buildId=769

scripts/common.lib.ps1
```diff
@@ -117,6 +128,7 @@ function Locate-VsInstallPath($hasVsixExtension ="false"){
 
   $requiredPackageIds += "Microsoft.Component.MSBuild" 
   $requiredPackageIds += "Microsoft.Net.Component.4.6.TargetingPack"
+  $requiredPackageIds += "Microsoft.VisualStudio.Windows.Build"
```
jayaranigarg, line 131: > $requiredPackageIds += "Microsoft.VisualStudio.Windows.Build" [](start = 1, length = 62)

can you help me understand why this package is required?

jimmylewis: scripts/build/TestFx.targets contains:
    
    <Import Project="$(MSBuildExtensionsPath)\Microsoft\WindowsXaml\v$(VisualStudioVersion)\Microsoft.Windows.UI.Xaml.CSharp.targets" Condition="$(TargetPlatformIdentifier) == 'UAP'" />

This targets file is part of the Microsoft.VisualStudio.Windows.Build component.  I ran into this because I have two instances of VS installed but the "latest" one has a very limited set of components.  The build script selected it because it was installed more recently and matched the other two component IDs, but the build failed because it couldn't find the targets file.

PR #466, opened 2018-07-27
Friendly test names

spottedmahn:
related to [#410 Allow Overriding the Test Name](https://github.com/Microsoft/testfx/issues/410)

Issue #515, opened 2018-11-09
Test Display Name should be Extensible


ahmedalejo:
Description
Following the extensibility initiative in MsTest V2. It would help to be able to define Display Names (i.e TestName) for tests thereby making human-readable(serves as better documentation of codebase as well) names that show up in the test explorer .

Test frameworks like xUnit this via configuration https://github.com/xunit/xunit/issues/759

**Expected behavior**

```csharp
[TestMethod("Consecutive authentication attempts with invalid credentials should fail")]
public void MyTest()
{ ... }
//or

[MyAppTestMethod]
public void Consecutive_authentication_attempts_with_invalid_credentials_should_fail()
{ ... }

//Given the following definitions
public class TestMethodAttrubte : Attribute
{
    public string DisplaName { get; set; }

    public virtual string GetDisplayName(System.Reflection.MethodInfo methodInfo) =>
        string.IsNullOrWhiteSpace(this.DisplaName) 
        ? methodInfo.Name 
        : this.DisplaName?.Trim();

   ...
}

public class MyAppsTestMethodAttrubte : TestMethodAttrubte
{
    public override string GetDisplayName(MethodInfo methodInfo) =>
        methodInfo
            .Name
            .Replace("_", " ")
            .Replace(".", ", ");
}
```
![image](https://user-images.githubusercontent.com/7012074/48248473-9cefb480-e3de-11e8-9025-664a00070f0d.png)

**Actual behavior**
The display name is "machine" readable.
![image](https://user-images.githubusercontent.com/7012074/48248171-68c7c400-e3dd-11e8-9562-146f2340f52f.png)

This proposal is also similar to #410

karanjitsingh:
Hi @ahmedalejo, thanks for suggesting that. I've marked this as an enhancement and we'll be happy to accept any contribution.

jadaml:
It might be me reading GitHub incorrectly, but I do not quite understand what is the decision here; are you saying that the feature request&mdash;to apply a display name altering the test case name displayed&mdash;is the same as #644 or it was just mentioned there and the decision is then ... what exactly?

ahmedalejo:
@jadaml like metioned in #644, 

Merging #644 would complement the implementation of #515, i.e if or when it is done.

jadaml:
So the decision is then is that you'll solve these two under the same issue, as both issues affects the display of the test, but both method of controlling the name (prettifying the method name, and overriding it with an attribute) will be available in the end?

Issue #410, opened 2018-04-25
Allow Overriding the Test Name


spottedmahn:
## Description
It would be very handy if we could control the `TestName` so we could create nice human-readable test names that show up in the test explorer.

Test frameworks like Mocha support this 😊.  [One example.](https://codeburst.io/how-to-test-javascript-with-mocha-the-basics-80132324752e)

Reference: [SO Post](https://stackoverflow.com/questions/16314845/mstest-how-to-change-name-of-testmethod-in-testexplorer)

## Expected behavior
```csharp
[TestMethod("My user friendly test name with special characters like | or * or \\ etc.")]
public void MyTest()
{
    ...
}
```

or

```csharp
[TestMethod()]
[TestName("My user friendly test name with special characters like | or * or \\ etc.")]
public void MyTest()
{
    ...
}
```

## Actual behavior
The method name is the test name 😪.

spottedmahn:
Is the code that decides what to display in Visual Studio in this repo?  I would imagine not.  I would imagine, it is part the of Visual Studio code base?

spottedmahn:
**Example:** Test name from [From testx repo](https://github.com/Microsoft/testfx/blob/3e4ca97f631eb04587ec62d889998fb08f5b2043/test/UnitTests/MSTest.CoreAdapter.Unit.Tests/Execution/TestMethodRunnerTests.cs#L138)

`ExecuteShouldSkipTestAndSkipFillingIgnoreMessageIfIgnoreAttributeIsPresentOnTestClassButHasNoMessage`  

**vs**

`Execute should skip this test and skip filling IgnoreMessage If IgnoreAttribute is present on the TestClass but has no message`

which one is easier for you to read? 😜

cltshivash:
@spottedmahn : This repo has the code for the test framework and decides the name that will be decided in VS. The ask is valid and we will look into it. we are open for contributions as well.

spottedmahn:
>How do I point Visual Studio at my new custom assemblies?

nvm, I think I found it (now that I'm a full pc and not my phone).

[Contribution Guide](https://github.com/Microsoft/testfx-docs/blob/master/docs/contribute.md)

>This article will help you build, test and consume local builds of the MSTest Framework and Adapter.

spottedmahn:
Making progress!!

![image](https://user-images.githubusercontent.com/14079228/40891919-bebb29cc-675c-11e8-913f-483183b8e34d.png)

spottedmahn:
[...]

PR #644, opened 2019-09-26
Prettify Test Method Display Name

ahmedalejo:
Friendly "Display Name" for test method names 

```csharp
[TestMethod]
void Method_names_should_be_readable_when_hinted()
{
}

"Method_names_should_be_readable_when_hinted" => "Method names should be readable when hinted"


[TestMethod]
void Say_What_You_Have_Spaces_in_Your_CSharp_Test_Names()
{
}

"Say_What_You_Have_Spaces_in_Your_CSharp_Test_Names" => "Say What You Have Spaces in Your CSharp Test Names"
```
similar to

image

Complements #466

This spec can be improved for other cases PascalCase, CamelCase.
and specified DisplayName should be favoured

### DIFF
--- a/src/Adapter/MSTest.CoreAdapter/ObjectModel/TestMethod.cs
+++ b/src/Adapter/MSTest.CoreAdapter/ObjectModel/TestMethod.cs
@@ -109,6 +109,6 @@ namespace Microsoft.VisualStudio.TestPlatform.MSTest.TestAdapter.ObjectModel
         /// </summary>
         public bool IsAsync { get; private set; }
         
-        string GetFriendlyName(string name) => name?.Replace('_', ' ');
+        string GetFriendlyName(string name) => name?.Replace('_', ' ')?.Trim();
     }
 }
--- a/src/Adapter/MSTest.CoreAdapter/ObjectModel/TestMethod.cs
+++ b/src/Adapter/MSTest.CoreAdapter/ObjectModel/TestMethod.cs
@@ -44,12 +44,12 @@ namespace Microsoft.VisualStudio.TestPlatform.MSTest.TestAdapter.ObjectModel
             this.AssemblyName = assemblyName;
             this.IsAsync = isAsync;
         }
-        
+
         /// <summary>
         /// Gets the name of the test method
         /// </summary>
         public string Name { get; private set; }
-        
+
         /// <summary>
         /// Gets the display name of the test method
         /// </summary>
@@ -108,7 +108,7 @@ namespace Microsoft.VisualStudio.TestPlatform.MSTest.TestAdapter.ObjectModel
         /// Gets a value indicating whether specifies test method is async
         /// </summary>
         public bool IsAsync { get; private set; }
-        
-        string GetFriendlyName(string name) => name?.Replace('_', ' ')?.Trim();
+
+        private string GetFriendlyName(string name) => name?.Replace('_', ' ')?.Trim();
     }
 }
--- a/src/Adapter/MSTest.CoreAdapter/ObjectModel/TestMethod.cs
+++ b/src/Adapter/MSTest.CoreAdapter/ObjectModel/TestMethod.cs
@@ -39,7 +39,7 @@ namespace Microsoft.VisualStudio.TestPlatform.MSTest.TestAdapter.ObjectModel
             Debug.Assert(!string.IsNullOrEmpty(fullClassName), "Full className cannot be empty");
 
             this.Name = name;
-            this.DisplayName = GetFriendlyName(name);
+            this.DisplayName = this.GetFriendlyName(name);
             this.FullClassName = fullClassName;
             this.AssemblyName = assemblyName;
             this.IsAsync = isAsync;
--- a/src/Adapter/MSTest.CoreAdapter/ObjectModel/TestMethod.cs
+++ b/src/Adapter/MSTest.CoreAdapter/ObjectModel/TestMethod.cs
@@ -39,15 +39,21 @@ namespace Microsoft.VisualStudio.TestPlatform.MSTest.TestAdapter.ObjectModel
             Debug.Assert(!string.IsNullOrEmpty(fullClassName), "Full className cannot be empty");
 
             this.Name = name;
+            this.DisplayName = GetFriendlyName(name);
             this.FullClassName = fullClassName;
             this.AssemblyName = assemblyName;
             this.IsAsync = isAsync;
         }
-
+        
         /// <summary>
         /// Gets the name of the test method
         /// </summary>
         public string Name { get; private set; }
+        
+        /// <summary>
+        /// Gets the display name of the test method
+        /// </summary>
+        public string DisplayName { get; private set; }
 
         /// <summary>
         /// Gets the full classname of the test method
@@ -102,5 +108,7 @@ namespace Microsoft.VisualStudio.TestPlatform.MSTest.TestAdapter.ObjectModel
         /// Gets a value indicating whether specifies test method is async
         /// </summary>
         public bool IsAsync { get; private set; }
+        
+        string GetFriendlyName(string name) => name?.Replace('_', ' ');
     }
 }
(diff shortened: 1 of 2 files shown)


ahmedalejo:
There also been other specs #515
and [[Proposal Support Automatic "Pretty" Display Name for Test Cases]](https://github.com/xunit/xunit/issues/759) by https://github.com/xunit/xunit/issues/759

spottedmahn:
I love this idea.  I was thinking about this same topic.

- How do we "fix" all of the existing test names?  I'm thinking a lot of test method names are written in PascalCase.
- Do method names matter anymore?  What conventions should we be using if using `TestMethod(displayName: "Some name")`?

```c#
[TestMethod("Now I can read my test names like a human not a computer 😜")]
public void Blah1()
{
}
```

ahmedalejo:
@spottedmahn 

I have two solutions

1. allow the `DisplayName` of the `TestMethodAttribute` to be extensible

1. a


```csharp
[TestMethod("Consecutive authentication attempts with invalid credentials should fail")]
public void MyTest()
{ ... }
//or

[MyAppTestMethod]
public void Consecutive_authentication_attempts_with_invalid_credentials_should_fail()
{ ... }

//Given the following definitions
public class TestMethodAttrubte : Attribute
{
    public string DisplaName { get; set; }

    public virtual string GetDisplayName(System.Reflection.MethodInfo methodInfo) =>
        string.IsNullOrWhiteSpace(this.DisplaName) 
        ? methodInfo.Name 
        : this.DisplaName?.Trim();

   ...
}

public class MyAppsTestMethodAttrubte : TestMethodAttrubte
{
    public override string GetDisplayName(MethodInfo methodInfo) =>
        methodInfo
            .Name
            .Replace("_", " ")
            .Replace(".", ", ");
}
```

nohwnd:
I like the general idea, and the extension point. But looking at the xunit spec, and the amount of work and iteration they did on it, I am inclined to keep the implementation at just that. No built in automatic rewrites. 

That way you can provide your own implementation, which can be simple enough to cover just the conventions you use. This would get the change merged quicker because we will not have to provide a good generic implementation, and multiple configuration options to cater every possible edge case.

src/Adapter/MSTest.CoreAdapter/ObjectModel/TestMethod.cs
```diff
@@ -49,6 +50,11 @@ public TestMethod(string name, string fullClassName, string assemblyName,  bool
         /// </summary>
         public string Name { get; private set; }
 
+        /// <summary>
+        /// Gets the display name of the test method
+        /// </summary>
+        public string DisplayName { get; private set; }
```
sebainones, line 56: As **DisplayName** has a **public** access modifier I believe it should be reflected as well in the _interface_ **ITestMethod**.
That's the _interface_ **ITestMethod** should declare:

```
public interface ITestMethod
    {
        /// <summary>
        /// Gets the display name of the test method
        /// </summary>
        string DisplayName { get; private set; }

       // The rest of the properties...
}
```

spottedmahn:
[Java JUnit 5 Display Name Generators](https://junit.org/junit5/docs/current/user-guide/#writing-tests-display-name-generator)

#interesting 🤔

nohwnd:
@spottedmahn could you elaborate a bit more on why you find that interesting? 🙂 as I read it it is basically the same concept as here, or am I missing something important?

src/Adapter/MSTest.CoreAdapter/ObjectModel/TestMethod.cs
nohwnd: I think you are right. While not ever public member on a class needs  to be part of it's interface, this seems to make a lot of sense. 🙂

sebainones: > I think you are right. While not ever public member on a class needs to be part of it's interface, this seems to make a lot of sense. 🙂

Yes, I agree on that but I think it could be useful if the interface can offer as well this **DisplayName** in this case.
So, how can I contribute (with code) to this PR? I have  already cloned @ahmedalejo fork just in case ;)

ahmedalejo: @sebainones your are correct, that should be it.

ahmedalejo:
Hi guys

cc: @spottedmahn 

I plan to close this PR.  I prefer sending in one that satisfies the below use case.

What do you think as the following partially implemented this.
#410 **issue**
#466 **PR** 

```csharp
[TestMethod("Consecutive authentication attempts with invalid credentials should fail")]
public void MyTest()
{ ... }
//or

[MyAppTestMethod]
public void Consecutive_authentication_attempts_with_invalid_credentials_should_fail()
{ ... }

//Given the following definitions
public class TestMethodAttrubte : Attribute
{
    public string DisplayName { get; set; }

    public virtual string GetDisplayName(System.Reflection.MethodInfo methodInfo) =>
        string.IsNullOrWhiteSpace(this.DisplayName) 
        ? methodInfo.Name 
        : this.DisplayName.Trim();

   ...
}

public class MyAppsTestMethodAttrubte : TestMethodAttrubte
{
    public override string GetDisplayName(MethodInfo methodInfo) =>
        methodInfo
            .Name
            .Replace("_", " ")
            .Replace(".", ", ");
}
public class MyHappyAppsTestMethodAttrubte : MyAppsTestMethodAttrubte 
{
    public override string GetDisplayName(MethodInfo methodInfo) =>
        base.GetDisplayName(methodInfo)
            .Replace("fails", "😢")
            .Replace("should succeed", "😅");
}
```

src/Adapter/MSTest.CoreAdapter/ObjectModel/TestMethod.cs
sebainones: @ahmedalejo please let me know how could I help you (contribute) with this. 
If you are opening a new PR just tag me so I can review it or if I can code something even better ;)

spottedmahn:
>🙂 as I read it it is basically the same concept as here

Yep, seems like it.  It was interesting to me that another "team" came to the same "discovery" independently.

There might also be some ideas in their approach we could learn from

ahmedalejo:
> @nohwnd  I missing something important?
The main thing is that there is more traction in other libraries about this topic. Basically, they´ve implemented it one way or the other.
It's a common need that we have to pull off for **testfx**

ahmedalejo:
@spottedmahn @nohwnd 

Found a better extension point that could be baked into `TestMethodAttribute` with ease.

The below is what i´m currently using

cheers.
```csharp
public class PrettyTestClassAttribute : TestClassAttribute
{
    public override TestMethodAttribute GetTestMethodAttribute(TestMethodAttribute wrappedTestMethodAttribute)
    {
        var attribute = base.GetTestMethodAttribute(wrappedTestMethodAttribute);

        return attribute is PrettyTestMethodAttribute
            ? attribute
            : new PrettyTestMethodAttribute(attribute);
    }
}

public class PrettyTestMethodAttribute : TestMethodAttribute
{
    private readonly TestMethodAttribute wrappedTestMethodAttribute;

    public PrettyTestMethodAttribute(){ }

    public PrettyTestMethodAttribute(TestMethodAttribute wrappedTestMethodAttribute) =>
        this.wrappedTestMethodAttribute = wrappedTestMethodAttribute;

    public override TestResult[] Execute(ITestMethod testMethod)
    {
        TestResult[] results;

        if (wrappedTestMethodAttribute is null)
            results = base.Execute(testMethod);
        else
            results = wrappedTestMethodAttribute.Execute(testMethod);

        if(results.Any())
            results[0].DisplayName = testMethod.TestMethodName
                .Replace("_eq_", " == ")
                .Replace("_neq_", " != ")
                .Replace("_gt_", " > ")
                .Replace("_gte_", " >= ")
                .Replace("_lt_", " < ")
                .Replace("_lte_", " <= ")
                .Replace("Bug_", "🐞");
        return results;
    }
}
```

PR #659, opened 2019-11-18
Configuring child domain's appbase to be test source location

jayaranigarg:
As part of this fix, we are setting child's appdomain to point to the test source location so that the dependencies gets picked up from the the test source location itself or we honor the relative path  specified in app.config file.

In scenario where we have multiple projects in a solution each using different versions of adapter, currently we run the projects with highest version of adapter referenced in that solution irrespective of the version a project references. After this fix, there will be a behavior change where each project will be run using the exact adapter version that particular project references and not the highest one in that solution.

### DIFF
--- a/src/Adapter/MSTest.CoreAdapter/Execution/TestExecutionManager.cs
+++ b/src/Adapter/MSTest.CoreAdapter/Execution/TestExecutionManager.cs
@@ -252,7 +252,7 @@ namespace Microsoft.VisualStudio.TestPlatform.MSTest.TestAdapter.Execution
                         typeof(TestAssemblySettingsProvider),
                         null) as TestAssemblySettingsProvider;
                 }
-                catch (TypeLoadException ex)
+                catch (Exception ex)
                 {
                     PlatformServiceProvider.Instance.AdapterTraceLogger.LogInfo(
                         "Could not create TestAssemblySettingsProvider instance in child app-domain",
--- a/src/Adapter/MSTest.CoreAdapter/Execution/TestExecutionManager.cs
+++ b/src/Adapter/MSTest.CoreAdapter/Execution/TestExecutionManager.cs
@@ -244,8 +244,31 @@ namespace Microsoft.VisualStudio.TestPlatform.MSTest.TestAdapter.Execution
                     sourceLevelParameters = sourceLevelParameters.Concat(this.sessionParameters).ToDictionary(x => x.Key, x => x.Value);
                 }
 
-                var sourceSettingsProvider = new TestAssemblySettingsProvider();
-                var sourceSettings = sourceSettingsProvider.GetSettings(source);
+                TestAssemblySettingsProvider sourceSettingsProvider = null;
+
+                try
+                {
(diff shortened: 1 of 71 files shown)


src/Adapter/MSTest.CoreAdapter/Execution/TestExecutionManager.cs
```diff
@@ -247,11 +244,31 @@ private void ExecuteTestsInSource(IEnumerable<TestCase> tests, IRunContext runCo
...
+                {
+                    sourceSettings = sourceSettingsProvider.GetSettings(source);
+                }
+                else
+                {
+                     sourceSettings = new TestAssemblySettings();
+                }
 
-                var sourceSettings = sourceSettingsProvider.GetSettings(source);
                 var parallelWorkers = sourceSettings.Workers;
```
mayankbansal018, line 263: Ensure that default values for sourceSettings.Workers, & Scope is set appropriately in case we fail to create instance of "TestAssemblySettingsProvider"

jayaranigarg: Yes, checked that.

src/Adapter/MSTest.CoreAdapter/Execution/TestExecutionManager.cs
```diff
@@ -247,11 +244,31 @@ private void ExecuteTestsInSource(IEnumerable<TestCase> tests, IRunContext runCo
...
-                    null) as TestAssemblySettingsProvider;
+                TestAssemblySettingsProvider sourceSettingsProvider = null;
+
+                try
+                {
+                    sourceSettingsProvider = isolationHost.CreateInstanceForType(
+                        typeof(TestAssemblySettingsProvider),
+                        null) as TestAssemblySettingsProvider;
+                }
+                catch (TypeLoadException ex)
```
singhsarab: >TypeLoadException [](start = 23, length = 17)

Generic exception ?

src/Adapter/MSTest.CoreAdapter/Execution/TestExecutionManager.cs
```diff
@@ -247,11 +244,31 @@ private void ExecuteTestsInSource(IEnumerable<TestCase> tests, IRunContext runCo
...
+                {
+                    PlatformServiceProvider.Instance.AdapterTraceLogger.LogInfo(
+                        "Could not create TestAssemblySettingsProvider instance in child app-domain",
+                        ex);
+                }
+
+                TestAssemblySettings sourceSettings;
+                if (sourceSettingsProvider != null)
```
singhsarab: Inline

src/Adapter/MSTest.CoreAdapter/Execution/TestExecutionManager.cs
jayaranigarg: Done

src/Adapter/MSTest.CoreAdapter/Execution/TestExecutionManager.cs
jayaranigarg: Done

PR #660, opened 2019-11-19
Fix bugs in parent class init/cleanup logic

NGloreous:
Fixing two bugs related to feature #143 
- Call base class cleanup even if top level test class doesn't have class cleanup
- Fix null ref when a class more than 1 level of inheritance has a class init/cleanup method.

### DIFF
--- a/src/Adapter/MSTest.CoreAdapter/Execution/TestClassInfo.cs
+++ b/src/Adapter/MSTest.CoreAdapter/Execution/TestClassInfo.cs
@@ -149,6 +149,12 @@ namespace Microsoft.VisualStudio.TestPlatform.MSTest.TestAdapter.Execution
         {
             get
             {
+                if (this.BaseClassCleanupMethodsStack.Any())
+                {
+                    // If any base cleanups were pushed to the stack we need to run them
+                    return true;
+                }
+
                 // If no class cleanup, then continue with the next one.
                 if (this.ClassCleanupMethod == null)
                 {
@@ -274,7 +280,11 @@ namespace Microsoft.VisualStudio.TestPlatform.MSTest.TestAdapter.Execution
                         var baseInitCleanupMethods = baseClassInitializeStack.Pop();
                         initializeMethod = baseInitCleanupMethods.Item1;
                         initializeMethod?.InvokeAsSynchronousTask(null, testContext);
-                        this.BaseClassCleanupMethodsStack.Push(baseInitCleanupMethods.Item2);
+
+                        if (baseInitCleanupMethods.Item2 != null)
+                        {
+                            this.BaseClassCleanupMethodsStack.Push(baseInitCleanupMethods.Item2);
+                        }
                     }
                 }
                 catch (Exception ex)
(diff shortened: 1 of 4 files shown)


NGloreous:
@parrainc FYI since this fixes some bugs from your change in case you have feedback.

src/Adapter/MSTest.CoreAdapter/Execution/TestClassInfo.cs
```diff
@@ -355,16 +365,16 @@ public void RunClassInitialize(TestContext testContext)
         [SuppressMessage("Microsoft.Design", "CA1031:DoNotCatchGeneralExceptionTypes", Justification = "Requirement is to handle all kinds of user exceptions and message appropriately.")]
         public string RunClassCleanup()
         {
-            if (this.ClassCleanupMethod is null && !this.BaseClassInitAndCleanupMethods.Any(p => p.Item2 != null))
+            if (this.ClassCleanupMethod is null && !this.BaseClassCleanupMethodsStack.Any())
```
jayaranigarg: We should keep this check as it is for the cases where we don't have classInitalize but we do have classCleanUp methods.

We are checking for this.BaseClassCleanupMethodsStack.Any() in the below if check as well.

src/Adapter/MSTest.CoreAdapter/Execution/TypeCache.cs
```diff
@@ -485,17 +485,15 @@ private bool IsAssemblyOrClassCleanupMethod(MethodInfo methodInfo, Type cleanupA
             TestClassInfo classInfo,
             ref MethodInfo[] initAndCleanupMethods)
         {
-            if (initAndCleanupMethods is null)
+            if (initAndCleanupMethods.Any(x => x != null))
```
jayaranigarg, line 488: What is the motivation for this change? Why do we need to reinitialize initAndCleanupMethods?

test/UnitTests/MSTest.CoreAdapter.Unit.Tests/Execution/TestClassInfoTests.cs
```diff
@@ -500,6 +500,20 @@ public void RunClassCleanupShouldReturnExceptionDetailsOfNonAssertExceptions()
                 "Class Cleanup method DummyTestClass.ClassCleanupMethod failed. Error Message: System.ArgumentException: Argument Exception. Stack Trace:     at Microsoft.VisualStudio.TestPlatform.MSTestAdapter.UnitTests.Execution.TestClassInfoTests.<>c.<RunClassCleanupShouldReturnExceptionDetailsOfNonAssertExceptions>");
         }
 
+        [TestMethod]
+        public void RunBaseClassCleanupEvenIfThereIsNoTopLevelClassCleanup()
```
jayaranigarg: Should we rename this to 'RunBaseClassCleanupEvenIfThereIsNoDerivedClassCleanup' ? 
TopLevel seems a bit unclear.

src/Adapter/MSTest.CoreAdapter/Execution/TypeCache.cs
MatthewSteeples: If you run the included tests but without the fixes you can see the problems that are being solved. Because this method used to reset `initAndCleanupMethods` back to null, it would cause a NullReferenceException elsewhere in the loop. So rather than the array being null, just the elements are null in this new process

MatthewSteeples:
We're seeing some problems with this change (we needed the same fix, so we've got a build of it working). I believe (not 100% sure) that it's changed the order of execution for TestInitialize attributes

MatthewSteeples:
Turns out that the scenario that these changes had enabled for us exposed a flaw in our tests that had runaway memory usage, so the problem was in our code! I think it just exhausted the memory on the Pipeline servers until paging stuff in and out to disk caused them to be unresponsive. When we got it working on a locally installed build agent it topped out at 24gb of RAM in use!

test/UnitTests/MSTest.CoreAdapter.Unit.Tests/Execution/TestClassInfoTests.cs
NGloreous: Sure

src/Adapter/MSTest.CoreAdapter/Execution/TestClassInfo.cs
NGloreous: I can revert it back but it's basically an equivalent check because things don't get pushed into BaseClassCleanupMethodsStack unless Item2 in BaseClassInitAndCleanupMethods was not null.

src/Adapter/MSTest.CoreAdapter/Execution/TypeCache.cs
NGloreous: Yes what Matt says is correct. This is the main fix for the null ref issue.

parrainc:
> @parrainc FYI since this fixes some bugs from your change in case you have feedback.

I think already provided review comments covers pretty much it. For me, it looks good! Thanks for addressing that issue :)

PR #661, opened 2019-11-20
Handle more than 2 levels of inheritance (fixes nullref)

MatthewSteeples:
We were seeing an error when using MSTest with our hierarchy when trying to add ClassInit and Cleanup attributes.

Have re-created the problem in tests and fixed the code (and also fixed a typo in another test)


NGloreous:
@MatthewSteeples FYI I also fixed this in my pr #660

MatthewSteeples:
So you did! Can't believe I missed that as I even skimmed through your PR looking to see if you'd made similar changes to me! Will close in favour of yours, no point the team reviewing 2 PRs!

PR #671, opened 2020-01-03
Fix build

nohwnd:
Fix build failing after merged pull request.

### DIFF
--- a/test/E2ETests/Automation.CLI/CLITestBase.cs
+++ b/test/E2ETests/Automation.CLI/CLITestBase.cs
@@ -77,6 +77,7 @@ namespace Microsoft.MSTestV2.CLIAutomation
             // this step of Initializing extensions should not be required after this issue: https://github.com/Microsoft/vstest/issues/236 is fixed
             vsTestConsoleWrapper.InitializeExtensions(Directory.GetFiles(this.GetTestAdapterPath(), "*TestAdapter.dll"));
             vsTestConsoleWrapper.RunTests(sources, runSettingXml, new TestPlatformOptions { TestCaseFilter = testCaseFilter }, this.runEventsHandler);
+            this.runEventsHandler.EnsureSuccess();
         }
 
         /// <summary>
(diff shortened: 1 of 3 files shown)


nohwnd:
```cs 
Error Message:
 Test method MSTestAdapter.Smoke.E2ETests.CompatTests.RunAllCompatTests threw exception: 
System.InvalidOperationException: Test run ended with 1 errors.
An exception occurred while invoking executor 'executor://mstestadapter/v2': 3 arguments were passed to 'Microsoft.VisualStudio.TestPlatform.MSTest.TestAdapter.Execution.UnitTestRunner::RunSingleTest'. 2 arguments were expected by this method.
Stack Trace:
    at Microsoft.MSTestV2.CLIAutomation.RunEventsHandler.EnsureSuccess() in D:\a\1\s\test\E2ETests\Automation.CLI\RunEventsHandler.cs:line 125
   at Microsoft.MSTestV2.CLIAutomation.CLITestBase.InvokeVsTestForExecution(String[] sources, String runSettings, String testCaseFilter) in D:\a\1\s\test\E2ETests\Automation.CLI\CLITestBase.cs:line 80
   at MSTestAdapter.Smoke.E2ETests.CompatTests.RunAllCompatTests() in D:\a\1\s\test\E2ETests\Smoke.E2E.Tests\CompatTests.cs:line 34
```
The build fails becase the api is no longer compatible with the api published by the legacy adapter that is used in the test. Investigating what is the best way to fix this.

nohwnd:
@jayaranigarg I can't figure this out, I can't locate where exactly it fails and figure out how to better debug this. Any suggestions who could help me with this?

AbhitejJohn:
@nohwnd: Sorry I usually tend to miss github tags. Please hit me up on teams if you need anything I haven't responded on. This specific issue does indeed sound like we have two different versions of the adapter on either side of the app domain which is a recipe for disaster. I dont know if these smoke tests test the adapter against its latest self - It ideally should. Maybe that's the problem.

nohwnd:
@AbhitejJohn no problem, it's stale, I ended up fixing it and we even discussed it in an internal mail thread about how to evolve the test platform interfaces. :)


The problem was that there is non-explicit dependency on an interface, and because we compile against one version of the library, but end up resolving to another version on runtime it fails when non compatible change is done to that interface. Closing this.

AbhitejJohn:
ah sounds good.

PR #673, opened 2020-01-10
Revert "Record test start/end events for data driven tests"

nohwnd:
Reverts microsoft/testfx#631
Reverts the PR because master fails in compat tests after the PR was merged.


NGloreous:
@nohwnd what's failing?

nohwnd:
@NGloreous sorry for the lack of info, I was busy with release when I was reverting this.  

This is where it started failing [this build](https://dev.azure.com/vstestplatform/TestFx/_build/results?buildId=915) which was triggered after merging your PR to master. It fails on RunAllCompatTests. 

I was trying to fix it in #671 because I hoped it will be just a quick fix (the original error I saw was just a missing parameter on a method call), but it turned out that the error is cryptic because the test base does not print errors when they are written in the log. So the message in the above build is confusing (because no tests were run, so they are none are in the list) and the error is actually (this)[https://github.com/microsoft/testfx/pull/671#issuecomment-571048003]. 

If you have any insight into how to fix this, please help me out in #671, so we can revert back the commit I am reverting here. 🙂

NGloreous:
@nohwnd thanks for taking care of this and sorry for the hassle. I've signed off here. I will take a look at #671 but it may have to wait until later in the week.

PR #680, opened 2020-01-30
Fix parameters in tests

nohwnd:


nohwnd:
Forced the merge, reverting to broken state that happened because of not updated merge.

PR #690, opened 2020-02-26
Assert.IsTrue() & False() to handle nullable bools

### LINKED ISSUE #652: Assert.IsTrue/IsFalse for nullable booleans
I propose adding the following overloads for handling nullable Booleans:

        public static void IsTrue(bool? condition);
        public static void IsTrue(bool? condition, string message);
        public static void IsTrue(bool? condition, string message, params object[] parameters);
        public static void IsFalse(bool? condition);
        public static void IsFalse(bool? condition, string message);
        public static void IsFalse(bool? condition, string message, params object[] parameters);

They would work exactly like their non-nullable versions.

**Reasoning**

When checking a Nullable<bool> value, we can't use the `IsTrue` and `IsFalse` shortcuts. Instead we have to use the longform of `Assert.AreEqual( [true|false]`, which is more error prone because it isn't type safe. 

There is also a very minor performance hit with `AreEqual` because it boxes by default. (You can use `AreEqual( (bool?) true, ` but no one is going to.)

HeroMaxPower:
Description:
An attempt on handling nullable bools for Assert.IsTrue() Assert.IsFalse()

Implemented these overrides and added some tests for them

`public static void IsTrue(bool? condition);`
`public static void IsTrue(bool? condition, string message);`
`public static void IsTrue(bool? condition, string message, params object[] parameters);`
`public static void IsFalse(bool? condition);`
`public static void IsFalse(bool? condition, string message);`
`public static void IsFalse(bool? condition, string message, params object[] parameters);`

Resolves #652

### DIFF
--- a/src/TestFramework/MSTest.Core/Assertions/Assert.cs
+++ b/src/TestFramework/MSTest.Core/Assertions/Assert.cs
@@ -162,14 +162,9 @@ namespace Microsoft.VisualStudio.TestTools.UnitTesting
         /// </exception>
         public static void IsTrue(bool? condition, string message, params object[] parameters)
         {
-            if (condition == false)
+            if (condition == false || condition == null)
             {
                 HandleFail("Assert.IsTrue", message, parameters);
-            }
-
-            if (condition == null)
-            {
-                HandleFail("Assert.IsNull", message, parameters);
             }
         }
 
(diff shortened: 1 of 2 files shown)


Haplois (approved):

HeroMaxPower:
Thanks Haplois

PR #691, opened 2020-02-27
Added StringComparison to StringAssert Contains(), EndsWith() & StartsWith()

### LINKED ISSUE #154: StringAssert APIs do not have ignorecase overloads
## Description
Like Assert.AreEqual APIs StringAssert does not have APIs that Ignore case in the default set. Although it is possible to add an extension through the Assertion extensibility [here](https://github.com/Microsoft/testfx-docs/blob/master/RFCs/002-Framework-Extensibility-Custom-Assertions.md), this is more of an ask to include this in the default set of APIs.

HeroMaxPower:
Description:

Added StringComparison to StringAssert Contains, EndsWith & StartsWith
Added tests to test functionality
Added API comment on StringComparison Enum
Updated APIs to default to StringComparison.Ordinal 

Probably needs API documentation updates if this gets merged.

Resolves #154

### DIFF
--- a/src/TestFramework/MSTest.Core/Assertions/StringAssert.cs
+++ b/src/TestFramework/MSTest.Core/Assertions/StringAssert.cs
@@ -116,9 +116,40 @@ namespace Microsoft.VisualStudio.TestTools.UnitTesting
         /// </exception>
         public static void Contains(string value, string substring, string message, params object[] parameters)
         {
+            Contains(value, substring, message, StringComparison.Ordinal, parameters);
+        }
+
+        /// <summary>
+        /// Tests whether the specified string contains the specified substring
+        /// and throws an exception if the substring does not occur within the
+        /// test string.
+        /// </summary>
+        /// <param name="value">
+        /// The string that is expected to contain <paramref name="substring"/>.
+        /// </param>
+        /// <param name="substring">
+        /// The string expected to occur within <paramref name="value"/>.
+        /// </param>
+        /// <param name="message">
+        /// The message to include in the exception when <paramref name="substring"/>
+        /// is not in <paramref name="value"/>. The message is shown in
+        /// test results.
+        /// </param>
+        /// <param name="comparisonType">
+        /// The comparison method to compare strings <paramref name="comparisonType"/>.
+        /// </param>
+        /// <param name="parameters">
+        /// An array of parameters to use when formatting <paramref name="message"/>.
+        /// </param>
+        /// <exception cref="AssertFailedException">
+        /// Thrown if <paramref name="substring"/> is not found in
+        /// <paramref name="value"/>.
+        /// </exception>
+        public static void Contains(string value, string substring, string message, StringComparison comparisonType, params object[] parameters)
+        {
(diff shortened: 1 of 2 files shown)


src/TestFramework/MSTest.Core/Assertions/StringAssert.cs
```diff
@@ -107,18 +158,21 @@ public static void Contains(string value, string substring, string message)
...
+        /// The comparison method to compare strings <paramref name="comparisonType"/>.
+        /// </param>
         /// <param name="parameters">
         /// An array of parameters to use when formatting <paramref name="message"/>.
         /// </param>
         /// <exception cref="AssertFailedException">
         /// Thrown if <paramref name="substring"/> is not found in
         /// <paramref name="value"/>.
         /// </exception>
-        public static void Contains(string value, string substring, string message, params object[] parameters)
```
304NotModified, line 117: please note, `Contains(string value, string substring, string message, params object[] parameters)` is now gone, so this is breaking change. 

Please also add  `Contains(string value, string substring, string message, params object[] parameters)` back

src/TestFramework/MSTest.Core/Assertions/StringAssert.cs
```diff
@@ -88,7 +111,35 @@ public static void Contains(string value, string substring)
...
+        /// <param name="comparisonType">
+        /// The comparison method to compare strings <paramref name="comparisonType"/>.
+        /// </param>
+        /// <exception cref="AssertFailedException">
+        /// Thrown if <paramref name="substring"/> is not found in
+        /// <paramref name="value"/>.
+        /// </exception>
+        public static void Contains(string value, string substring, string message, StringComparison comparisonType)
+        {
+            Contains(value, substring, message, comparisonType, null);
```
304NotModified: are we sending `null` to ` params object[] parameters`? That isn't needed nor recommend? 

(same for StartsWith and EndsWith)

nohwnd:
@HeroMaxPower are you going to address the PR comments or should I?

HeroMaxPower:
Yes please, i haven't had time in my personal life to look at these. 👍

src/TestFramework/MSTest.Core/Assertions/StringAssert.cs
nohwnd: Added

src/TestFramework/MSTest.Core/Assertions/StringAssert.cs
nohwnd: added

nohwnd:
@304NotModified I had to pass empty array to disambiguate the calls (before we passed null). The compiler will pass empty array for us, so we should not be doing anything wrong. Unless the user passed StringComparison as the actual first param, then they now would get different overload than before. But that is unlikely scenario.

HeroMaxPower:
@304NotModified any updates?

304NotModified (approved):
Looks good to me! Thanks for the reminder!

HeroMaxPower:
@nohwnd i think this is ready to be merged 👍

nohwnd:
@Haplois could you review and merge please.

Haplois (approved):

PR #697, opened 2020-03-06
Added an EnumDataSourceAttribute

joboivin:
Sometimes, in our automated tests, it's possible we want to execute a given test for all values of an `enum`, or for all values except one or two. However, there are no easy and scalable way to do so.

For example, let's consider an arbitraty enum : 
``` csharp
private enum Value
{
    First,
    Second,
    Third,
    Fourth
}
```

We can use the `DataRow` attribute, but we have to keep it in sync everytime we add a new value to the `enum`. And, it is not always explicit that we want to exclude a particuliar valu, especially for larger `enum`s (imagine you have 50 values and 49 `DataRow`, it's not easy to see which value isn't there):
``` csharp
[DataTestMethod]
[DataRow(Value.First)]
[DataRow(Value.Second)]
[DataRow(Value.Fourth)]
public void MyMethodShouldReturnTrueIfValueIsNotThirdValue(Value value)
{
    ...
}
```

The same result can also be done with `DynamicData`, which can already be more explicit and be kept in sync with new values:
``` csharp
[DataTestMethod]
[DynamicData(nameof(AllValuesExceptThird),  DynamicDataSourceType.Method)]
public void MyMethodShouldReturnTrueIfValueIsNotThirdValue(Value value)
{
    ...
}

private static IEnumerable<object[]>  AllValuesExceptThird()
{
    var allValuesExceptThird = ((Value[])Enum.GetValues(typeof(Value))).Where(x => x != Value.Third);
    foreach (var value in allValuesExceptThird)
    {
        yield return new object[] { value};
    }
}
``` 

The problem with this approach is that it's not really scalable. If I want to exclude all `Value`s except `Value.First` in another test, I have to create a new method. And if I want all `Value`s except `Value.First` and `Value.Third`, I have to create yet a third method. It's going to take quite some place at the bottom of my test class, and doesn't really contribute to anything in the class.

Following this reflection, I came up with an `EnumDataSource`. With only 1 line, we can iterate on all values in an `enum`, or exclude some values:

``` csharp
[DataTestMethod]
[EnumDataSource(typeof(Value))]
public void MyMethodShouldReturnTrueForAllValues(Value value)
{
   ...
}

[DataTestMethod]
[EnumDataSource(typeof(Value), exclusions: Value.Third)]
public void MyMethodShouldReturnTrueIfValueIsNotThirdValue(Value value)
{
    ...
}
```

Contraty to `DataRow`, it's really explicit which values are excluded, and contrary to `DynamicData`, I don't have to create a new method each time I want a test to exclude a new value. No matter the size of the `enum`, it's always going to be a single line, that tells us what's going on in the test.

### DIFF
--- a/src/TestFramework/MSTest.Core/Attributes/EnumDataSourceAttribute.cs
+++ b/src/TestFramework/MSTest.Core/Attributes/EnumDataSourceAttribute.cs
@@ -1,0 +1,72 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+namespace Microsoft.VisualStudio.TestTools.UnitTesting
+{
+    using System;
+    using System.Collections.Generic;
+    using System.Linq;
+    using System.Reflection;
+
+    /// <summary>
+    /// Attribute to define dynamic data for a test method.
+    /// </summary>
+    [AttributeUsage(AttributeTargets.Method, AllowMultiple = false)]
+    public class EnumDataSourceAttribute : Attribute, ITestDataSource
+    {
+        private readonly Type enumDataSource;
+        private readonly object[] exclusions;
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="EnumDataSourceAttribute"/> class.
+        /// </summary>
+        /// <param name="enumDataSource">
+        /// The type of the <see cref="Enum"/> to iterate for the test.
+        /// </param>
+        public EnumDataSourceAttribute(Type enumDataSource)
+            : this(enumDataSource, new object[0])
+        {
+            // Need to have this constructor explicitely to fix a CLS compliant error.
+        }
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="EnumDataSourceAttribute"/> class.
+        /// </summary>
+        /// <param name="enumDataSource">
+        /// The type of the <see cref="Enum"/> to iterate for the test.
+        /// </param>
+        /// <param name="exclusions">
+        /// Values to exclude during the iteration.
+        /// </param>
+        public EnumDataSourceAttribute(Type enumDataSource, params object[] exclusions)
+        {
+            this.enumDataSource = enumDataSource ??
+                                  throw new ArgumentNullException(nameof(enumDataSource), "The enum to iterate is not specified.");
+
+            this.exclusions = exclusions;
+        }
+
+        /// <inheritdoc />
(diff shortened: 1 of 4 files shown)


src/TestFramework/MSTest.Core/Attributes/EnumDataSourceAttribute.cs
```diff
@@ -0,0 +1,72 @@
...
+        /// <summary>
+        /// Initializes a new instance of the <see cref="EnumDataSourceAttribute"/> class.
+        /// </summary>
+        /// <param name="enumDataSource">
+        /// The type of the <see cref="Enum"/> to iterate for the test.
+        /// </param>
+        /// <param name="exclusions">
+        /// Values to exclude during the iteration.
+        /// </param>
+        public EnumDataSourceAttribute(Type enumDataSource, params object[] exclusions)
```
joboivin, line 41: I would have loved to create a generic attribute `EnumDataSourceAttribute<T> where T: System.Enum`, and then enforce the type here (`params T[] exclusions`), but sadly en attribute cannot be a generic type 😢

nohwnd:
This seems to be quite specialized. What happens when the enum is Flags? What if you want to do inclusions instead of exclusions? 

Maybe it would be a better solution to allow the DynamicData attribute to take parameters that would be passed to the method? 🙂  Because that is way more generic.

Ultimately the extension point `IDataSource` is there, so you can implement it in any way you want.

joboivin:
@nohwnd you make some good points. The way I saw it for the inclusions was that, if I have an `enum` with around 10 values, and I want to use just 1 or 2 values for my unit test, then I'm just gonna go with
```csharp
[DataTestMethod]
[DataRow(Value.First)]
[DataRow(Value.Second)]
public void MyMethodShouldReturnTrueForFirstOrSecond(Value value)
{
    ...
}
```

It's really easy to use, and we can see which values are used. This is not the case if I want to exclude values, if I put all the `Value`s except 1 as `DataRow`s, it's not always easy to see. For smaller `enum`s it is, but for larger it's not really, that's why I went with an exclusion thinking.

You make a good point with the `Flag`, I haven't though about it. It depends whether or not the need to test all values of an `enum` is present with flagged `enum`s. To give you a more concrete example, we have an `enum` that represents all product sold by a company. For 1 particular product, we want to do a specific treatment. So, for my unit test, to ensure that this treatment isn't done for other products, it's really easy to do with `EnumDataSource`, as there are around 50 different products in the `enum`. 

Obviously, if we realy follow OOP principles, each product should be a different `class` instead of an `enum` value, but with legacy systems it's not always the case, that's why I though this attribute could be of good use.

PR #698, opened 2020-03-07
Implement ThreadSafety in TestContext.

### LINKED ISSUE #386: TestContext not Thread Safe
## Description
The concept of the TestContext as a property of a class is not thread safe.  This is especially true if it is made static, but is even true if it is class instance specific.  It needs to be made an optional parameter to the test level functions to make it so.  This bug is related to #385.  Although they address slightly different problems, if you had to choose between the two to fix, I think this way of addressing the threading issues would be cleaner.

## Steps to reproduce
Add the .cs file in the attached zip into a MSTest 1.3.0-beta2 project with MethodLevel scope parallelize functionality activated.  

[ParallelTestContextBug.zip](https://github.com/Microsoft/testfx/files/1842835/ParallelTestContextBug.zip)

## Expected behavior
The Test Initialize and Test Cleanup functions, at the very least, should be able to take a TestContext parameter.  For Data Driven tests with attached data sources, the Test Method functions should also get one, as the DataRow property in a Class level TestContext variable is bound to get jumbled between the functions.  If you get to the point of parallelizing a data driven function's test cases, this separation will become even more critical.

## Actual behavior
Because the TestContext object is shared at the class level, the output from the two test functions and their respective Test Initialize and Cleanup functions gets jumbled.

HeroMaxPower:
Implements some locking in reference types called in test context, mostly StringBuilder and Properties Dictionary

Fixes #386

### DIFF
--- a/src/Adapter/PlatformServices.NetCore/Services/NetCoreTestContextImplementation.cs
+++ b/src/Adapter/PlatformServices.NetCore/Services/NetCoreTestContextImplementation.cs
@@ -93,7 +93,10 @@ namespace Microsoft.VisualStudio.TestPlatform.MSTestAdapter.PlatformServices
         {
             get
             {
-                return this.GetStringPropertyValue(TestContextPropertyStrings.TestRunDirectory);
+                lock (TestContextPropertyStrings.TestRunDirectory)
+                {
+                    return this.GetStringPropertyValue(TestContextPropertyStrings.TestRunDirectory);
+                }
             }
         }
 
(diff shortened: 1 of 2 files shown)


HeroMaxPower:
@nohwnd 
Im new to thread safety, am i on the right track with this initial implementation. Something tells me there maybe more to it.

nohwnd:
What issue are you trying to solve here? I don't think adding a bunch of locks is the way to go. Do we have a case where this becomes a problem?

HeroMaxPower:
The problem is described in #386, I'm not sure if this is the right way to go about it. My guess is when running test methods in parallel and calling the WriteLine() that people are running into some issues with writing and accessing properties in the test context. I thought that locking some of these would solve this. but i'm a novice 🙂

nohwnd:
It would probably solve it, but with a lot of overhead and potential to cause deadlocks :)

HeroMaxPower:
Not sure what to do with this now, my only thoughts would be to Lazy load test context as that could be thread safe.

PR #699, opened 2020-03-12
Code Maintenance with static analysers

HeroMaxPower:
IDE0029 Null check can be simplified.
IDE0016 Null check can be simplified
IDE0017 Object initialization can be simplified.
IDE0019 Use pattern matching.
IDE0028 Collection initialization can be simplified.
IDE0054 Use compound assignment.
IDE1005 Delegate invocation can be simplified.

CA1507 Use nameof in place of string literal
CA1825: Avoid zero-length array allocations
CA1802 Field is declared as 'static readonly' but is initialized with a constant value. Mark this field as 'const' instead.
CA1813 Avoid unsealed attributes.

S1066 Merge this if statement with the enclosing one.
S2971 Drop 'Where' and move the condition into the 'Count'.
S2971 Drop 'Where' and move the condition into the 'FirstOrDefault'.
S2971 Drop this useless call to 'ToList' or replace it by 'AsEnumerable' if you are using LINQ to Entities.
S1125 Remove the unnecessary Boolean literal(s).
S1066 Merge this if statement with the enclosing one.
S1128 Remove this unnecessary 'using'.

SA1623 The property's documentation summary text must begin with: 'Gets'
SA1611 The documentation for parameter is missing.
SA1642 Constructor summary documentation must begin with standard text.


HeroMaxPower:
Closing this as I have another PR #807

PR #724, opened 2020-07-16
WIP: Implement Class Cleanup Lifecycle selection

jhenkens:
I'm soliciting early feedback on a branch to introduce a lifecycle policy for the `ClassCleanup` attribute. This allows a test author to set at the command line, assembly, or class level, whether a the `ClassCleanup` should happen at the end of a class, or at the end of the assembly.

I've lightly tested this, and it seems to work as expected in my tests, though I am not confident in its robustness in its current form. Specifically, my current implementation relying on a dictionary of the strings of the full test names seems fragile to me, though I was unable to determine how to instead track via the guids on the completion side of the dictionary.

If this feature seems generally appeciated, I'd love guidance from those more familar with the MSTest codebase in how I might improve my implementation.

A part of any final submission I will surely include reverting the many changed build files - those were needed to build in my environment, and are not to be present in any final reviews.

Thank you.


src/Adapter/MSTest.CoreAdapter/MSTestSettings.cs
```diff
@@ -338,6 +344,26 @@ private static MSTestSettings ToSettings(XmlReader reader)
                                 break;
                             }
 
+                        case "CLASSCLEANUPLIFECYCLE":
+                            {
+                                var value = reader.ReadInnerXml();
+                                if (Enum.TryParse(value, out ClassCleanupLifecycle lifecycle))
```
nohwnd, line 350: Make this case insensitive, and check the the parsed value IsDefined, otherwise any value that is parsable to the underlying type (int) would be a valid enum value, e.g. "10".
```suggestion
                                if (Enum.TryParse(value, true, out ClassCleanupLifecycle lifecycle) && Enum.IsDefined(lifecycle))
```

src/Adapter/MSTest.CoreAdapter/MSTestSettings.cs
```diff
@@ -338,6 +344,26 @@ private static MSTestSettings ToSettings(XmlReader reader)
...
+                                {
+                                    settings.ClassCleanupLifecycle = lifecycle;
+                                }
+                                else
+                                {
+                                    throw new AdapterSettingsException(
+                                        string.Format(
+                                            CultureInfo.CurrentCulture,
+                                            Resource.InvalidParallelScopeValue,
```
nohwnd, line 359: This uses an incorrect resource.
```suggestion
                                            Resource.InvalidClassCleanupLifecycleValue,
```

src/TestFramework/MSTest.Core/Attributes/ClassCleanupSequencingAttribute.cs
```diff
@@ -0,0 +1,33 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+namespace Microsoft.VisualStudio.TestTools.UnitTesting
+{
+    using System;
+
+    /// <summary>
+    /// Specification for parallelization level for a test run.
```
nohwnd, line 9: Fix description, please.

src/TestFramework/MSTest.Core/Attributes/ClassCleanupSequencingAttribute.cs
```diff
@@ -0,0 +1,33 @@
...
+        /// <summary>
+        /// Initializes a new instance of the <see cref="ClassCleanupSequencingAttribute"/> class.
+        /// </summary>
+        public ClassCleanupSequencingAttribute()
+        {
+            this.LifecyclePosition = DefaultClassCleanupLifecycle;
+        }
+
+        /// <summary>
+        /// Gets or sets the number of workers to be used for the parallel run.
```
nohwnd, line 25: Also wrong description.

src/Adapter/MSTest.CoreAdapter/Execution/TestAssemblySettingsProvider.cs
```diff
@@ -56,6 +56,13 @@ internal TestAssemblySettings GetSettings(string source)
 
             testAssemblySettings.CanParallelizeAssembly = !this.reflectHelper.IsDoNotParallelizeSet(testAssembly);
 
+            var classCleanupSequencingAttribute = this.reflectHelper.GetClassCleanupAttribute(testAssembly);
```
nohwnd, line 59: Is this naming based on some existing settings we have? I think we should choose between Sequencing and Lifecycle if there is no existing example of this and in that case Lifecycle seems easier to grasp.

Or possibly ClassCleanupExecution?

nohwnd (commented):
Some early thoughts, I did not get to dig deeper into the code and think about edge cases yet, unfortunately.

Haplois:
Code moved to #968. @jhenkens thank you for the contribution!

PR #730, opened 2020-08-14
Fix XML doc comments (code -> c)

Youssef1313:
Fixes https://github.com/MicrosoftDocs/feedback/issues/2943

@mairaw, would `<see cref="TypeName" />` be more suitable than `<c>TypeName</c>`?

I'm currently using `<c>` per your comment https://github.com/MicrosoftDocs/feedback/issues/2943#issuecomment-663307767


Youssef1313:
The doc page shows versions 1.2.0 and 1.3.2, will the change made here reflect on these versions?

mairaw:
If you use see cref, then just make sure to use the fully qualified name. But c would already fix the bad layout on docs. see cref would add the link to the API.

Haplois (approved):

Haplois:
Thank you for your contribution, @Youssef1313!

PR #737, opened 2020-10-22
FQN work

Haplois:
https://github.com/microsoft/vstest/pull/2611 needs to be merged, and published into the package repository before merging this PR.


src/Adapter/MSTest.CoreAdapter/ObjectModel/UnitTestElement.cs
```diff
@@ -114,10 +114,21 @@ internal TestCase ToTestCase()
                 this.TestMethod.FullClassName,
                 this.TestMethod.Name);
 
-            TestCase testCase = new TestCase(fullName, TestAdapter.Constants.ExecutorUri, this.TestMethod.AssemblyName);
-            testCase.DisplayName = string.IsNullOrEmpty(this.DisplayName) ? this.TestMethod.Name : this.DisplayName;
+            TestCase testCase;
 
+            if (string.IsNullOrWhiteSpace(this.TestMethod.ManagedType) || string.IsNullOrWhiteSpace(this.TestMethod.ManagedMethod))
```
peterwald: Similar comment in vstest repo. Consider creating a shared helper that is used everywhere to determine if we have `ManagedType`/`ManagedMethod`

src/Adapter/PlatformServices.NetCore/PlatformServices.NetCore.csproj
```diff
@@ -66,11 +66,13 @@
   </ItemGroup>
 
   <ItemGroup>
-    <PackageReference Include="Microsoft.TestPlatform.ObjectModel" Version="15.0.0" />
+    <PackageReference Include="Microsoft.TestPlatform.ObjectModel" Version="$(TestPlatfromVersion)" />
```
peterwald: nit: should be `$(TestPlatformVersion)`

peterwald:
LGTM with a few minor comments. It's encouraging to see that the addition of the two properties seems to be relatively simple. Of course, we still need to add support for parameter/argument sets.

src/Adapter/PlatformServices.NetCore/PlatformServices.NetCore.csproj
Haplois: Thanks 👍

scripts/Build.ps1
```diff
@@ -307,7 +295,37 @@ function Create-NugetPackages
...
+  if(-not $content.Equals($newContent)) {
+    Write-Log "Updating TestPlatform version in $File"
+    $newContent | Set-Content -Encoding utf8 $File -NoNewline
+  }
+}
+
+function Sync-PackageVersions {
+  $versionsFile = "$PSScriptRoot\build\TestFx.Versions.targets"
+
+  $versionsRegex = '(?mi)<(TestPlatformVersion.*?)>(.*?)<\/TestPlatformVersion>'
```
nohwnd, line 310: nit: Not sure what the ?mi option would do here. It is definitely case insensitive by default ( -split and -isplit is case insensitive, -csplit is case sensitive). And you are passing the whole file as one string (you used -Raw).  If you have a strong reason I would like to know to learn.

scripts/Build.ps1
```diff
@@ -307,7 +295,37 @@ function Create-NugetPackages
...
+  $sourceRegex = '(?mi)(.+[a-z =]+\@\")Microsoft\.TestPlatform\.([0-9.-a-z]+)\";'
+
+  if ([String]::IsNullOrWhiteSpace($TestPlatformVersion)) {
+    $TestPlatformVersion = (([XML](Get-Content $versionsFile)).Project.PropertyGroup.TestPlatformVersion).InnerText
+  } else {
+    Replace-InFile -File $versionsFile -RegEx $versionsRegex -ReplaceWith "<`$1>$TestPlatformVersion</TestPlatformVersion>"
+  }
+
+  (Get-ChildItem "$PSScriptRoot\..\src\*packages.config","$PSScriptRoot\..\test\*packages.config" -Recurse) | ForEach-Object {
+    Replace-InFile -File $_ -RegEx $packageRegex -ReplaceWith "<package id=`"Microsoft.TestPlatform`$1`" version=`"$TestPlatformVersion`""
```
nohwnd: nit: consider using string format: `'some string with literal $1 and {0}' -f $TestPlatformVersion` to avoid all that ugly escaping.

scripts/common.lib.ps1
```diff
@@ -127,7 +127,7 @@ function Locate-VsInstallPath($hasVsixExtension ="false"){
   $requiredPackageIds = @()
 
   $requiredPackageIds += "Microsoft.Component.MSBuild" 
-  $requiredPackageIds += "Microsoft.Net.Component.4.6.TargetingPack"
+  $requiredPackageIds += "Microsoft.Net.Component.4.5.2.TargetingPack"
```
nohwnd, line 130: Didn't this work as is? This will require everyone to install more targeting packs on their system won't it?

src/Adapter/PlatformServices.Desktop/AssemblyResolver.cs
```diff
@@ -577,7 +587,7 @@ private bool TryLoadFromCache(string assemblyName, bool isReflectionOnly, out As
         /// <param name="loggerAction">The logger function.</param>
         private void SafeLog(string assemblyName, Action loggerAction)
         {
-            if (!string.IsNullOrEmpty(assemblyName) && !assemblyName.StartsWith(LoggerAssemblyName))
+            if (!string.IsNullOrEmpty(assemblyName) && !assemblyName.StartsWith(LoggerAssemblyName) && !assemblyName.StartsWith(LoggerAssemblyNameLegacy))
```
nohwnd, line 591: Please add comment with the reason why this was done.

test/UnitTests/MSTest.CoreAdapter.TestUtilities/ActionUtility.cs
```diff
@@ -3,7 +3,7 @@
 
 namespace MSTestAdapter.TestUtilities
 {
-#if NETCOREAPP1_1
+#if NETCOREAPP1_1 || NETCOREAPP2_1
```
nohwnd: should this be any NETCOREAPP or just those two specific versions?

nohwnd (commented):
Looks mostly good, I would review if #if NETCOREAPP1_1 || NETCOREAPP2_1 can be replaced witch check for just netcoreapp to help us in the future if we need to multitarget this further.

scripts/Build.ps1
Haplois: In some (like test/E2ETests/Automation.CLI/packages.config) files we reference more than one package, this is to catch all references and update those.

test/UnitTests/MSTest.CoreAdapter.TestUtilities/ActionUtility.cs
Haplois: We can just use `NETCOREAPP`, updated.

scripts/Build.ps1
nohwnd: Looks like that is the default behavior of `-replace`:

```
PS > Get-Content ~/Desktop/abc.txt

abc
abc abcabc
abc

PS > (Get-Content ~/Desktop/abc.txt -Raw) -replace "abc", "ggg"

ggg       
ggg gggggg
ggg    
```

scripts/common.lib.ps1
Haplois: I am not sure about this, but we target v4.5.2 but v4.6 in our projects. Let's say if a user has the v4.6 targeting pack installed but not v4.5.2 won't we fail anyway in that case?

src/Adapter/PlatformServices.Desktop/AssemblyResolver.cs
Haplois: done.

scripts/common.lib.ps1
nohwnd: I think I confused this with the 4.6 Portable pack. I looked into my VS installation and I have all targeting packs installed, to avoid those issues, so no problem for me. Ignore this then.

nohwnd (approved):

scripts/Build.ps1
Haplois: I'll have another PR after this and move to use `PackageReference` instead of "NuGet.config" files. With that PR, I'll remove this logic too.

src/Adapter/MSTest.CoreAdapter/ObjectModel/UnitTestElement.cs
```diff
@@ -114,10 +114,21 @@ internal TestCase ToTestCase()
                 this.TestMethod.FullClassName,
                 this.TestMethod.Name);
 
-            TestCase testCase = new TestCase(fullName, TestAdapter.Constants.ExecutorUri, this.TestMethod.AssemblyName);
-            testCase.DisplayName = string.IsNullOrEmpty(this.DisplayName) ? this.TestMethod.Name : this.DisplayName;
+            TestCase testCase;
 
+            if (this.TestMethod.HasManagedMethodAndType)
+            {
+                testCase = new TestCase(fullName, this.TestMethod.ManagedType, this.TestMethod.ManagedMethod, TestAdapter.Constants.ExecutorUri, this.TestMethod.AssemblyName);
```
shyamnamboodiripad, line 121: @Haplois Won't this fail If the version of mstest adapter that includes the current FQN changes is loaded in VS 2017 where the TP OM that will be loaded will not contain this `TestCase` constructor?

src/Adapter/MSTest.CoreAdapter/ObjectModel/TestMethod.cs
```diff
@@ -44,6 +47,20 @@ public TestMethod(string name, string fullClassName, string assemblyName,  bool
...
 
+        public TestMethod(MethodBase method, string name, string fullClassName, string assemblyName, bool isAsync)
+            : this(name, fullClassName, assemblyName, isAsync)
+        {
+            if (method == null)
+            {
+                throw new ArgumentNullException(nameof(method));
+            }
+
+            ManagedNameHelper.GetManagedName(method, out var managedType, out var managedMethod);
```
shyamnamboodiripad, line 58: @Haplois Same comment here. `ManagedNameHelper` will not be available in the TP OM that ships with older versions of VS.

src/Adapter/MSTest.CoreAdapter/ObjectModel/UnitTestElement.cs
Haplois: It will fail. I am moving these changes to a separate assembly and NuGet package. I'll have a PR soon.

PR #744, opened 2020-11-12
Add nullable-annotated Assert.IsNotNull

### LINKED ISSUE #714: Nullable Reference Type-aware support?
## Description
I've been making some updates to some code I work on to see what it would take to actually use C# 8.0's Nullable Reference Types feature.  One issue I've run into is that some of the `Assert.*` methods don't use the new [CodeAnalysis attributes](https://docs.microsoft.com/en-us/dotnet/api/system.diagnostics.codeanalysis?view=netcore-3.1), and our unit test projects end up needing extra code to help the compiler know a nullable variable isn't actually null.

## Steps to reproduce
.NET Core 3.1 unit test project (with NRT enabled)
```csharp
string? str = null;
Assert.IsNotNull(str);
Assert.AreEqual(5, str.Length);
```

## Expected behavior
No compiler warnings.

## Actual behavior
Compiler sees `Assert.IsNotNull` is null-oblivious, is not able to infer anything about `str` from it, and issues a warning on `str.Length` because thinks `str` could be null.

## Workaround
I've written a shim to unblock me for now, but it'd be nice if the library itself had all the right annotations. :smile:

```csharp
public static class NullableAssert
{
    public static void IsNotNull([NotNull] object? value)
    {
        Assert.IsNotNull(value);
    }
}

// ...elsewhere...

string? str = null;
NullableAssert.IsNotNull(str);
Assert.AreEqual(5, str.Length); // no warning this time!
```

I suspect there are other annotations that could also be added elsewhere, but I've not tried to think through those other cases.  I am but a humble app developer, not a library developer. 😇 

Also, I think fixing this requires a new platform target to be added, since the attributes were added in `netcoreapp3.0`/`netstandard2.1`.

Haplois:
- Fixes #714

### DIFF
--- a/src/TestFramework/MSTest.Core/Assertions/Assert.cs
+++ b/src/TestFramework/MSTest.Core/Assertions/Assert.cs
@@ -359,7 +359,7 @@ namespace Microsoft.VisualStudio.TestTools.UnitTesting
         /// <exception cref="AssertFailedException">
         /// Thrown if <paramref name="value"/> is null.
         /// </exception>
-        public static void IsNotNull(object value)
+        public static void IsNotNull([NotNull] object value)
         {
             IsNotNull(value, string.Empty, null);
         }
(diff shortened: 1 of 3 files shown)


nohwnd (approved):
LGTM

CodeBlanch:
Is there any timeline for this being available in the NuGet package? I'm getting a lot of CS8602 warnings now in my tests (using VS16.9.0 Preview 1) which cause build breakage w/ TreatWarningsAsErrors enabled.

Haplois:
We will release a preview package tomorrow.

CodeBlanch:
I updated to the latest preview and the warnings are gone thanks @Haplois!

PR #749, opened 2020-12-03
BugFix: WorkItemAttribute not extracted 

jflepp:
The `TypeEnumerator` class returned an `Attribute[]`, which cannot be casted to an `WorkItemAttribute[]`. Even though the method signature specified an `Attribute[]`, the mock was set up to return a `WorkItemAttribute[]` array, which causes the test to pass but the `as` conversion to fail at runtime.

### DIFF
--- a/src/Adapter/MSTest.CoreAdapter/Discovery/TypeEnumerator.cs
+++ b/src/Adapter/MSTest.CoreAdapter/Discovery/TypeEnumerator.cs
@@ -192,8 +192,8 @@ namespace Microsoft.VisualStudio.TestPlatform.MSTest.TestAdapter.Discovery
                 testElement.Description = descriptionAttribute.Description;
             }
 
-            var workItemAttributeArray = this.reflectHelper.GetCustomAttributes(method, typeof(WorkItemAttribute)).Cast<WorkItemAttribute>();
-            testElement.WorkItemIds = workItemAttributeArray.Select(x => x.Id.ToString()).ToArray();
+            var workItemAttributes = this.reflectHelper.GetCustomAttributes(method, typeof(WorkItemAttribute)).Cast<WorkItemAttribute>();
+            testElement.WorkItemIds = workItemAttributes.Select(x => x.Id.ToString()).ToArray();
 
             // Get Deployment items if any.
             testElement.DeploymentItems = PlatformServiceProvider.Instance.TestDeployment.GetDeploymentItems(
--- a/src/Adapter/MSTest.CoreAdapter/Discovery/TypeEnumerator.cs
+++ b/src/Adapter/MSTest.CoreAdapter/Discovery/TypeEnumerator.cs
@@ -192,11 +192,8 @@ namespace Microsoft.VisualStudio.TestPlatform.MSTest.TestAdapter.Discovery
                 testElement.Description = descriptionAttribute.Description;
             }
 
-            var workItemAttributeArray = this.reflectHelper.GetCustomAttributes(method, typeof(WorkItemAttribute)) as WorkItemAttribute[];
-            if (workItemAttributeArray != null)
-            {
-                testElement.WorkItemIds = workItemAttributeArray.Select(x => x.Id.ToString()).ToArray();
-            }
+            var workItemAttributeArray = this.reflectHelper.GetCustomAttributes(method, typeof(WorkItemAttribute)).Cast<WorkItemAttribute>();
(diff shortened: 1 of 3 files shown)


schibu007:
I'd really appreciate this PR.

src/Adapter/MSTest.CoreAdapter/Discovery/TypeEnumerator.cs
```diff
@@ -192,11 +192,8 @@ internal UnitTestElement GetTestFromMethod(MethodInfo method, bool isDeclaredInT
                 testElement.Description = descriptionAttribute.Description;
             }
 
-            var workItemAttributeArray = this.reflectHelper.GetCustomAttributes(method, typeof(WorkItemAttribute)) as WorkItemAttribute[];
-            if (workItemAttributeArray != null)
-            {
-                testElement.WorkItemIds = workItemAttributeArray.Select(x => x.Id.ToString()).ToArray();
-            }
+            var workItemAttributeArray = this.reflectHelper.GetCustomAttributes(method, typeof(WorkItemAttribute)).Cast<WorkItemAttribute>();
```
Haplois: Nit-picking: This is no longer an array. Please rename the variable.

src/Adapter/MSTest.CoreAdapter/Discovery/TypeEnumerator.cs
```diff
@@ -192,11 +192,8 @@ internal UnitTestElement GetTestFromMethod(MethodInfo method, bool isDeclaredInT
...
+            testElement.WorkItemIds = workItemAttributeArray.Select(x => x.Id.ToString()).ToArray();
```
Haplois: We're not doing a null check here, What happens if `workItemAttributeArray` is empty?

Haplois (changes requested):

src/Adapter/MSTest.CoreAdapter/Discovery/TypeEnumerator.cs
fl3pp: Not nit-picking, the variable name _is_ wrong

Removed `Attribute` Suffix

src/Adapter/MSTest.CoreAdapter/Discovery/TypeEnumerator.cs
fl3pp: - The `GetCustomAttributes` method is simply mapped to `MemberInfo.GetCustomAttributes()`, which doesn't return null ([docs](https://docs.microsoft.com/en-us/dotnet/api/system.reflection.memberinfo.getcustomattributes?view=netframework-4.5.1)). 
- As we are not expecting the `GetCustomAttributes` method to return any `Attribute` other than `WorkItemAttribute`, the `.Cast()` extension method is the right choice here. It returns the casted enumerable or throws if any object is not of the specified type.

So a null check as it was implemented before doesn't make sense anymore.

If the `WorkItemAttribute` is empty, the `testElement.WorkItemIds` property is set to an empty array. IMO this makes sense, but then I remembered the following snippet in the `UnitTestElement` class:

``` cs
if (this.WorkItemIds != null)
{
    testCase.SetPropertyValue(TestAdapter.Constants.WorkItemIdsProperty, this.WorkItemIds);
}
```

So, we found a bug here: The `UnitTestElement` class sets the test case property only if the `WorkItemIds` property is not null, which cannot happen anymore in this implementation.

In order to fix the bug and stick with the existing conventions, I altered the implementation to set the `WorkItemIds` property to null if no WorkitemIds exist. I added the UnitTest `GetTestFromMethodShouldSetWorkItemIdsToNullIfNotAny` to ensure this behavior is not altered.

Haplois: Thanks for the fix. I'll try to approve this PR today _or tomorrow._

Haplois (approved):
👍 LGTM. Thank you for your contribution.

jflepp:
Thank you for your review!

Is there any chance you could tell when a preview will be available?

Haplois:
There is a back-compatibility issue in https://github.com/microsoft/vstest; after the fix is merged there, we can release a preview. Hopefully on this Friday, or mid next week at the latest.

We're also continuously integrating into our [daily feed][1]. You can use that for now. The version should be "MSTest.TestAdapter 2.2.0-preview-20201209-01" after the [build][2] completes.

[1]: https://dev.azure.com/dnceng/public/_packaging?_a=feed&feed=test-tools
[2]: https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=4285992

PR #739, opened 2020-10-22
Updated dependencies. Added new assembly level attribute.

Sanan07:

PR #746, opened 2020-11-24
Load specific types from adapter

Sanan07:
Added functionality to MSTestAdapter from Test Platform, where we load only necessary types from assemblies.

PR #753, opened 2020-12-14
Fixed TypesToLoadAttribute compatibility

Haplois:
TestFx updated to implement changes in https://github.com/microsoft/vstest/pull/2674; this should fix the compatibility issue introduced in https://github.com/microsoft/vstest/pull/2537 and https://github.com/microsoft/testfx/pull/739.

### DIFF
--- a/src/Adapter/MSTest.CoreAdapter/Properties/TypesToLoadAttribute.cs
+++ b/src/Adapter/MSTest.CoreAdapter/Properties/TypesToLoadAttribute.cs
@@ -1,0 +1,26 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System;
+
+using Microsoft.VisualStudio.TestPlatform.MSTest.TestAdapter;
+using Microsoft.VisualStudio.TestPlatform.ObjectModel.Utilities;
+
+[assembly: TypesToLoad(typeof(MSTestDiscoverer), typeof(MSTestExecutor))]
+
+namespace Microsoft.VisualStudio.TestPlatform.ObjectModel.Utilities
+{
+    /// <summary>
+    /// Custom Attribute to specify the exact types which should be loaded from assembly
+    /// </summary>
+    [AttributeUsage(AttributeTargets.Assembly, Inherited = false, AllowMultiple = false)]
+    internal sealed class TypesToLoadAttribute : Attribute
+    {
+        public TypesToLoadAttribute(params Type[] types)
+        {
+            this.Types = types;
+        }
+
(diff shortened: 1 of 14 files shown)


peterwald (approved):

src/Adapter/MSTest.CoreAdapter/Properties/TypesToLoadAttribute.cs
```diff
@@ -0,0 +1,26 @@
...
+
+[assembly: TypesToLoad(typeof(MSTestDiscoverer), typeof(MSTestExecutor))]
+
+namespace Microsoft.VisualStudio.TestPlatform.ObjectModel.Utilities
+{
+    /// <summary>
+    /// Custom Attribute to specify the exact types which should be loaded from assembly
+    /// </summary>
+    [AttributeUsage(AttributeTargets.Assembly, Inherited = false, AllowMultiple = false)]
+    internal sealed class TypesToLoadAttribute : Attribute
```
shyamnamboodiripad, line 17: I'd [mentioned](https://github.com/microsoft/testfx/pull/746#discussion_r533752933) this to @Sanan07, but perhaps `TestExtensionTypesAttribute` would be be a better name (i.e. a bit less confusing for adapter implementers).

src/Adapter/MSTest.CoreAdapter/Properties/TypesToLoadAttribute.cs
```diff
@@ -0,0 +1,26 @@
...
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System;
+
+using Microsoft.VisualStudio.TestPlatform.MSTest.TestAdapter;
+using Microsoft.VisualStudio.TestPlatform.ObjectModel.Utilities;
+
+[assembly: TypesToLoad(typeof(MSTestDiscoverer), typeof(MSTestExecutor))]
+
+namespace Microsoft.VisualStudio.TestPlatform.ObjectModel.Utilities
```
shyamnamboodiripad, line 11: Could we make the namespace simply `Microsoft.VisualStudio.TestPlatform` or `Microsoft.VisualStudio.TestPlatform.ObjectModel`? The `Utilities` part of the name seems verbose and I may not intuitively remember that the attribute needs to be created in that namespace...

shyamnamboodiripad (commented):
LGTM modulo comments about naming

src/Adapter/MSTest.CoreAdapter/Properties/TypesToLoadAttribute.cs
Haplois: That's a good idea and a relatively simple change. I'll do it tomorrow morning.

PR #763, opened 2021-01-22
Updated build status badge.

Haplois:


README.md
```diff
@@ -8,7 +8,7 @@ This is a fully supported, open source and cross-platform implementation of the
 
 ### Build status
 
-[![Build Status](https://dev.azure.com/vstestplatform/TestFx/_apis/build/status/TestFx.CI?branchName=master)](https://dev.azure.com/vstestplatform/TestFx/_build/latest?definitionId=8&branchName=master)
+[![Build Status](https://dev.azure.com/dnceng/public/_apis/build/status/Microsoft/testfx/microsoft.testfx.ci?branchName=master)](https://dev.azure.com/dnceng/public/_build/latest?definitionId=937&branchName=master)
```
nohwnd, line 11: The only confusing part about this is that typically the private part of the build is in the dnceng private (or is it internal?), but ours will be in devdiv. Maybe we should move vstest under dnceng as part of the arcade migration?

Sanan07 (approved):

README.md
Haplois: I checked all organizations we have. Only public org. we have is in dnceng. It already had some `microsoft/*` repositories, and all of dotnet builds. So I just used it. arcade is also there.

PR #771, opened 2021-02-15
Merge settings safely

nohwnd:
Merge session and source level settings safely. Linq.Concat will throw "key has already been added" exception when there are duplicates. This add a helper method that will merge the dictionaries safely.


Haplois (approved):
LGTM

PR #776, opened 2021-02-17
Forward logs to EqtTrace

nohwnd:
When running on .NET Core forward logs to EqtTrace to be written into diag log when it is enabled.

### DIFF
--- a/src/Adapter/MSTest.CoreAdapter/MSTestExecutor.cs
+++ b/src/Adapter/MSTest.CoreAdapter/MSTestExecutor.cs
@@ -44,6 +44,8 @@ namespace Microsoft.VisualStudio.TestPlatform.MSTest.TestAdapter
 
         public void RunTests(IEnumerable<TestCase> tests, IRunContext runContext, IFrameworkHandle frameworkHandle)
         {
+            PlatformServiceProvider.Instance.AdapterTraceLogger.LogInfo("MSTestExecutor.RunTests: Running tests from testcases.");
+
             ValidateArg.NotNull(frameworkHandle, "frameworkHandle");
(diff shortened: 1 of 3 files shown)


nohwnd:
Example of logs:

image

nohwnd:
@Haplois  Do you know how we could safely figure out the exact version of the used adapter from the assembly metadata and print it into the log, as we do with VSTest platform? In vstest it is in Product class. But simple reflection does not work in mstest because it is portable. So I guess I need multiple implementations?

Haplois:
> @Haplois Medeni Baykal FTE  Do you know how we could safely figure out the exact version of the used adapter from the assembly metadata and print it into the log, as we do with VSTest platform? In vstest it is in Product class. But simple reflection does not work in mstest because it is portable. So I guess I need multiple implementations?

The easiest way to get it is using `System.Diagnostics.FileVersionInfo` class, we can't directly map that into a commit but it's unique. In MSTest it's quite clear like to the build. Use the `ProductVersion`.

Examples:
VSTest: 14.0.5717.01
MSTest: 16.10.0-preview-20210216-05

Haplois (approved):

nohwnd:
Is that MSTest version correct? Shouldn't that be 2.1.1 or something?

nohwnd:
Anyways let's merge that now, and I will look into the version later.

Haplois:
> Is that MSTest version correct? Shouldn't that be 2.1.1 or something?

Ah sorry, I just realized that I mislabeled; MSTest version we set at assemblies are versions of `14.*.*.*` they are unique across packages but wrong.

Issue #789, opened 2021-03-02
Cannot run tests anymore after upgrading to 2.2.1


avivanoff:
Attempt to upgrade our test projects to version from 2.1.2 to 2.2.1 results in the following test execution error:

`An exception occurred while invoking executor 'executor://mstestadapter/v2': Could not load file or assembly 'Microsoft.TestPlatform.AdapterUtilities, Version=15.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a' or one of its dependencies. The system cannot find the file specified.`

[AB#1287803](https://devdiv.visualstudio.com/0bdbc590-a062-4c3f-b0f6-9383f67865ee/_workitems/edit/1287803)

avivanoff:
A very simple sample project that reproduces the issue: [UnitTestProject1.zip](https://github.com/microsoft/testfx/files/6072972/UnitTestProject1.zip).

To get it to execute, just comment out DeploymentItem in UnitTest1.cs.

kwilson-93:
I am also having issues successfully executing tests when upgrading from v2.1.2 to v2.2.1 (I am also using DeploymentItem mentioned above) 👍

mprokopy:
Same for me.

Sanan07:
You need to install this package [Microsoft.TestPlatform.AdapterUtilities](https://dev.azure.com/dnceng/public/_packaging?_a=package&feed=test-tools&package=Microsoft.TestPlatform.AdapterUtilities&protocolType=NuGet&version=16.10.0-preview-20210303-02)

@Haplois we need to think how to prevent this issue. Everyone who update MsTestAdapter to 2.2.1 will get this error.
Maybe somehow automatically download this package while updating MsTestAdapter

mprokopy:
@Sanan07 for the following project, the installation of the package did not help: 
[UnitTestProject2.zip](https://github.com/microsoft/testfx/files/6078720/UnitTestProject2.zip)

Haplois:
Hi @avivanoff and @mprokopy 

Thank you for reporting this bug. There is an issue with the `MSTest.TestAdapter` package. Currently, you can disable app domains to make your tests functional.

```xml
<RunSettings>
  <RunConfiguration>
    <DisableAppDomain>True</DisableAppDomain>
  </RunConfiguration>
</RunSettings>
```

avivanoff:
@Haplois, when do you expect to have a fix?

Haplois:
[...]

PR #794, opened 2021-03-09
Moved ITestDataSource discovery to test discovery.

Haplois:
## Description
This PR moves discovery for `IDataSource` based data test methods to test discovery instead of execution. See the below examples for details.

### Work done:
1. `IDataSource` data tests (includes `DataRow` tests) are now discovered before test execution; and individually runnable.
2. Parent test results returned for data tests are not generated now. (See the new behavior of DataSourceAttribute in pictures.)

|       | IDataSource | DataSourceAttribute |
|----:|:------:|:------:|
| Current behavior | Current style with parent result (DataRow) | Current style with parent result (DataSource) |
| New behavior: | New implementation without a parent result (DataRow) | New implementation without a parent result (DataSource) |


src/Adapter/MSTest.CoreAdapter/Resources/Resource.resx
```diff
@@ -320,4 +320,42 @@ Error: {1}</value>
   <data name="Execution_Test_Cancelled" xml:space="preserve">
     <value>Test '{0}' execution has been aborted.</value>
   </data>
+  <data name="CannotEnumarateDataSourceAttribute" xml:space="preserve">
```
peterwald: `Enumerate` should be spelled with an 'e'.

src/Adapter/MSTest.CoreAdapter/Resources/Resource.resx
```diff
@@ -320,4 +320,42 @@ Error: {1}</value>
...
+  <data name="FailedFetchAttributeCache" xml:space="preserve">
+    <value>Failed to get attribute cache. Ignoring attribute inheritance and falling into 'type defines Attribute model', so that we have some data.</value>
+  </data>
+  <data name="FailedToGetCustomAttribute" xml:space="preserve">
+    <value>Getting custom attributes for type {0} threw exception (will ignore and use the reflection way): {1}</value>
+    <comment>{0}: Attribute full type name.
+{1}: Exception description</comment>
+  </data>
+  <data name="OlderTFMVersionFound" xml:space="preserve">
+    <value>An older version of MSTestV2 package is loaded in asssembly, test discovery might fail to discover all data tests if they depend on `.runsettings` file.</value>
```
peterwald: 2 's' in `assembly`

src/Adapter/MSTest.CoreAdapter/Helpers/DataSerializationHelper.cs
```diff
@@ -0,0 +1,94 @@
...
+
+            return serializedData;
+        }
+
+        /// <summary>
+        /// Deserialzes the data serialzed by <see cref="Serialize(object[])" /> method.
+        /// </summary>
+        /// <param name="serializedData">Serialized data array to deserialize.</param>
+        /// <returns>Deserialized array.</returns>
+        public static object[] Deserialze(string[] serializedData)
```
peterwald: nit: `Deserialize()`

src/Adapter/MSTest.CoreAdapter/Helpers/DataSerializationHelper.cs
```diff
@@ -0,0 +1,94 @@
...
+                if (serializedData[i] == null)
+                {
+                    data[i] = null;
+                    continue;
+                }
+
+                var serialzedDataBytes = Encoding.UTF8.GetBytes(serializedData[i]);
+                using (var memoryStream = new MemoryStream(serialzedDataBytes))
+                {
+                    data[i] = serializer.ReadObject(memoryStream);
```
peterwald, line 74: It would be nice to add an overload to the serializer to allow for ReadObject to operate on a `byte[]` or even better a `ReadOnlySpan<byte>`. That would eliminate the extra allocations here for the `MemoryStream`.

Same holds for the Serializer as well...adding a method to serialize directly to a `byte[]`

src/Adapter/MSTest.CoreAdapter/Helpers/ReflectHelper.cs
```diff
@@ -113,7 +113,7 @@ public bool HasAttributeDerivedFrom(MemberInfo memberInfo, Type baseAttributeTyp
             Dictionary<string, object> attributes = this.GetAttributes(memberInfo, inherit);
             if (attributes == null)
             {
-                PlatformServiceProvider.Instance.AdapterTraceLogger.LogWarning("ReflectHelper.HasAttributeDerivedFrom: Failed to get attribute cache. Ignoring attribute inheritance and falling into 'type defines Attribute model', so that we have some data.");
+                PlatformServiceProvider.Instance.AdapterTraceLogger.LogWarning($"ReflectHelper.HasAttributeDerivedFrom: {Resource.FailedFetchAttributeCache}");
```
peterwald:
```csharp
$"{nameof(ReflectHelper)}.{nameof(HasAttributeDerivedFrom)}: {Resource.FailedFetchAttributeCache}"
```

would be more future-proof regarding renamings.

src/Adapter/MSTest.CoreAdapter/ObjectModel/UnitTestElement.cs
```diff
@@ -188,16 +204,54 @@ internal TestCase ToTestCase()
...
+                testFullId += testCase.FullyQualifiedName;
+            }
+
+            if (this.TestMethod.DataType != DynamicDataType.None)
+            {
+                testFullId += testCase.DisplayName;
+            }
+
+            testCase.Id = EqtHash.GuidFromString(testFullId);
+
```
peterwald, line 251: Can we accumulate string values into the hash one by one, and grab the hash at the end? This code will allocate a new string each time a new value is appended. So a lot of string allocations for each id.

src/Adapter/MSTest.CoreAdapter/ObjectModel/UnitTestElement.cs
```diff
@@ -188,16 +204,54 @@ internal TestCase ToTestCase()
...
         private string GetDisplayName()
         {
             if (string.IsNullOrWhiteSpace(this.DisplayName))
             {
-                return string.IsNullOrWhiteSpace(this.TestMethod.ManagedMethodName)
-                     ? this.TestMethod.Name
-                     : this.TestMethod.ManagedMethodName;
+                return this.TestMethod.Name;
+
+                // This causes compatibility problems with older runners.
```
peterwald: What is the compatibility issue? Is it because of unexpected characters? Or something different.

test/E2ETests/Smoke.E2E.Tests/CustomTestExecutionExtensibilityTests.cs
```diff
@@ -40,8 +40,9 @@ public void ExecuteCustomTestExtensibilityWithTestDataTests()
                 "CustomTestMethod2 (B)",
                 "CustomTestMethod2 (B)");
 
+            // TODO :: ASK REVIEW :: SHOULD WE CONTINUE RETURNING A PARENT TEST FOR DATA BOUND TESTS?
             // Parent results should fail and thus failed count should be 7.
-            this.ValidateFailedTestsCount(7);
+            // this.ValidateFailedTestsCount(7);
```
peterwald: Do we not need this check any more?

test/E2ETests/Smoke.E2E.Tests/CustomTestExecutionExtensibilityTests.cs
```diff
@@ -40,8 +40,9 @@ public void ExecuteCustomTestExtensibilityWithTestDataTests()
                 "CustomTestMethod2 (B)",
                 "CustomTestMethod2 (B)");
 
+            // TODO :: ASK REVIEW :: SHOULD WE CONTINUE RETURNING A PARENT TEST FOR DATA BOUND TESTS?
```
peterwald: I don't think we should continue to return parent test, as there really is no such concept elsewhere. It was primarily used to support the summary feature which is no longer helpful.

peterwald (approved):
Looks like `test/E2ETests/DiscoveryAndExecutionTests/Utilities/CLITestBase.discovery.cs` is empty?

Mostly LGTM with comments noted.

src/Adapter/MSTest.CoreAdapter/ObjectModel/UnitTestElement.cs
Haplois: Yes, basically because unexpected and unescaped char in name https://github.com/microsoft/testfx/issues/789

test/E2ETests/Smoke.E2E.Tests/CustomTestExecutionExtensibilityTests.cs
Haplois: We're not returning a parent test anymore.

Haplois:
> @peterwald: 
> Looks like `test/E2ETests/DiscoveryAndExecutionTests/Utilities/CLITestBase.discovery.cs` is empty?

It actually is not; I don't know why it's listed as a binary file...

scripts/Build.ps1
```diff
@@ -48,6 +48,9 @@ Param(
 
   [Alias("tpv")]
   [string] $TestPlatformVersion = $null,
+  
+  [Alias("np")]
+  [Switch] $DisallowPrereleaseMSBuild,
```
shyamnamboodiripad: This seems to be duplicated below. Was this intentional?

src/Adapter/MSTest.CoreAdapter/Discovery/AssemblyEnumerator.cs
```diff
@@ -40,6 +53,11 @@ public AssemblyEnumerator(MSTestSettings settings)
             MSTestSettings.PopulateSettings(settings);
         }
 
+        /// <summary>
+        /// Sets run settings to use for current discovery session.
+        /// </summary>
+        public string RunSettingsXml { private get; set; }
```
shyamnamboodiripad: nit: Might as well make the getter also public. I find set-only properties a bit weird :) although this is not in public API so may be ok...

src/Adapter/MSTest.CoreAdapter/Discovery/AssemblyEnumerator.cs
```diff
@@ -55,14 +73,15 @@ public override object InitializeLifetimeService()
...
-        /// <returns> A collection of Test Elements. </returns>
-        [SuppressMessage("Microsoft.Design", "CA1031:DoNotCatchGeneralExceptionTypes", Justification = "Catching a generic exception since it is a requirement to not abort discovery in case of any errors.")]
+        /// <param name="assemblyFileName">The assembly file name.</param>
+        /// <param name="warnings">Contains warnings if any, that need to be passed back to the caller.</param>
+        /// <returns>A collection of Test Elements.</returns>
         internal ICollection<UnitTestElement> EnumerateAssembly(string assemblyFileName, out ICollection<string> warnings)
         {
             Debug.Assert(!string.IsNullOrWhiteSpace(assemblyFileName), "Invalid assembly file name.");
 
+            string runSettingsXml = this.RunSettingsXml;
```
shyamnamboodiripad: nit: var

Also this local can probably be deleted i.e. the callsite below can probably access the field directly.

src/Adapter/MSTest.CoreAdapter/Discovery/AssemblyEnumerator.cs
```diff
@@ -6,21 +6,34 @@ namespace Microsoft.VisualStudio.TestPlatform.MSTest.TestAdapter.Discovery
...
+
     /// <summary>
     /// Enumerates through all types in the assembly in search of valid test methods.
     /// </summary>
     internal class AssemblyEnumerator : MarshalByRefObject
     {
+        /// <summary>
+        /// Helper for reflection API's.
+        /// </summary>
+        private static readonly ReflectHelper ReflectHelper = new ReflectHelper();
```
shyamnamboodiripad: Consider: Perhaps you could define a static instance in the type itself that all callers can then use i.e. `ReflectHelper.Instance`

src/Adapter/MSTest.CoreAdapter/Discovery/AssemblyEnumeratorWrapper.cs
```diff
@@ -22,8 +21,7 @@ internal class AssemblyEnumeratorWrapper
         /// <summary>
         /// Assembly name for UTF
         /// </summary>
-        private static readonly AssemblyName UnitTestFrameworkAssemblyName =
-            typeof(TestMethodAttribute).GetTypeInfo().Assembly.GetName();
+        private static readonly AssemblyName UnitTestFrameworkAssemblyName = typeof(TestMethodAttribute).GetTypeInfo().Assembly.GetName();
```
shyamnamboodiripad: nit: I would recommend moving refactoring / formatting / code hygiene changes that are unrelated to the primary change to a separate PR (especially when there are many such refactoring changes). Otherwise, it can be difficult for reviewers to figure out and focus on exactly what is changing 😄

src/Adapter/MSTest.CoreAdapter/Discovery/AssemblyEnumeratorWrapper.cs
```diff
@@ -71,33 +59,18 @@ internal class AssemblyEnumeratorWrapper
...
-                    fullFilePath,
-                    ex);
-
-                var message = string.Format(
-                    CultureInfo.CurrentCulture,
-                    Resource.TestAssembly_AssemblyDiscoveryFailure,
-                    fullFilePath,
-                    ex.Message);
+                var message = string.Format(CultureInfo.CurrentCulture, Resource.TestAssembly_AssemblyDiscoveryFailure, fullFilePath, ex.Message);
+                PlatformServiceProvider.Instance.AdapterTraceLogger.LogWarning($"{nameof(AssemblyEnumeratorWrapper)}.{nameof(this.GetTests)}: {Resource.TestAssembly_AssemblyDiscoveryFailure}", fullFilePath, ex);
```
shyamnamboodiripad: Is it by design that the trace messages are now localized? I am assuming the trace messages were mainly intended for our own debugging (as opposed to for a user trying to debug problems on their own). If yes, then localizing the trace logs may make it more difficult to analyze logs supplied by users...

Just checking to make sure this was an intentional change that has been considered by the team.

src/Adapter/MSTest.CoreAdapter/Discovery/AssemblyEnumeratorWrapper.cs
```diff
@@ -141,8 +106,17 @@ private ICollection<UnitTestElement> GetTestsInIsolation(string fullFilePath, IR
             using (var isolationHost = PlatformServiceProvider.Instance.CreateTestSourceHost(fullFilePath, runSettings, frameworkHandle: null))
             {
                 // Create an instance of a type defined in adapter so that adapter gets loaded in the child app domain
-                var assemblyEnumerator = isolationHost.CreateInstanceForType(
-                    typeof(AssemblyEnumerator), new object[] { MSTestSettings.CurrentSettings }) as AssemblyEnumerator;
+                var assemblyEnumerator = isolationHost.CreateInstanceForType(typeof(AssemblyEnumerator), new object[] { MSTestSettings.CurrentSettings }) as AssemblyEnumerator;
+
+                // This might not be supported if an older version of "PlatformServices
```
shyamnamboodiripad: Comment seems incomplete. Was some text deleted by accident?

src/Adapter/MSTest.CoreAdapter/Discovery/TypeEnumerator.cs
```diff
@@ -199,6 +198,8 @@ internal UnitTestElement GetTestFromMethod(MethodInfo method, bool isDeclaredInT
                 testElement.WorkItemIds = workItemAttributes.Select(x => x.Id.ToString()).ToArray();
             }
 
+            testElement.Ignored = this.reflectHelper.IsAttributeDefined(method, typeof(IgnoreAttribute), false);
```
shyamnamboodiripad, line 201: Just curious. Was it a bug that we were not populating ignore info before? Or did this code exist elsewhere before?

src/Adapter/MSTest.CoreAdapter/Discovery/AssemblyEnumerator.cs
```diff
@@ -220,11 +196,191 @@ internal string GetLoadExceptionDetails(ReflectionTypeLoadException ex)
...
+                            tests.Add(test);
+                            continue;
+                        }
+
+                        if (this.DynamicDataAttached(sourceLevelParameters, assembly, test, tests))
+                        {
+                            continue;
+                        }
+
+                        tests.Add(test);
```
shyamnamboodiripad, line 234: The above logic is slightly confusing. Perhaps adding comments would help.

I am not sure I understand why its important to check whether test is ignored in this case - in other words, why are dynamic data tests only added if they are ignored (and skipped otherwise)?

Why can't we do the following instead? In other words, can we skip all dynamic tests and add add all remaining tests (regardless of whether they are ignored).

```
foreach (var test in unitTestCases)
{
    if (this.DynamicDataAttached(sourceLevelParameters, assembly, test, tests))
    {
        continue;
    }

    tests.Add(test);
}
```

src/Adapter/MSTest.CoreAdapter/Discovery/AssemblyEnumerator.cs
```diff
@@ -220,11 +196,191 @@ internal string GetLoadExceptionDetails(ReflectionTypeLoadException ex)
...
+                    }
+                }
+            }
+            catch (Exception exception)
+            {
+                // If we fail to discover type from a class, then don't abort the discovery
+                // Move to the next type.
+                string message = string.Format(CultureInfo.CurrentCulture, Resource.CouldNotInspectTypeDuringDiscovery, typeFullName, assemblyFileName, exception.Message);
+                PlatformServiceProvider.Instance.AdapterTraceLogger.LogInfo($"AssemblyEnumerator: {message}");
+                warningMessages.Add(message);
```
shyamnamboodiripad, line 244: Will this warning be relayed to the Test Explorer when client is VS?

src/Adapter/MSTest.CoreAdapter/Discovery/AssemblyEnumerator.cs
```diff
@@ -220,11 +196,191 @@ internal string GetLoadExceptionDetails(ReflectionTypeLoadException ex)
...
+            }
+
+            return tests;
+        }
+
+        private bool DynamicDataAttached(IDictionary<string, object> sourceLevelParameters, Assembly assembly, UnitTestElement test, List<UnitTestElement> tests)
+        {
+            if (test.TestMethod.HasManagedMethodAndTypeProperties == false)
```
shyamnamboodiripad: When is this expected to be false? Aren't all test methods supposed to have this now? Perhaps a comment would help...

shyamnamboodiripad: minor: `if (!test.TestMethod.HasManagedMethodAndTypeProperties)` (assuming this is a simple bool)

src/Adapter/MSTest.CoreAdapter/Discovery/AssemblyEnumerator.cs
```diff
@@ -220,11 +196,191 @@ internal string GetLoadExceptionDetails(ReflectionTypeLoadException ex)
...
+                var testContext = PlatformServiceProvider.Instance.GetTestContext(testMethod, writer, sourceLevelParameters);
+                var testMethodInfo = this.typeCache.GetTestMethodInfo(testMethod, testContext, MSTestSettings.CurrentSettings.CaptureDebugTraces);
+                if (testMethodInfo == null)
+                {
+                    return false;
+                }
+
+                return false /* DataSourceAttribute discovery is disabled for now, since we cannot serialize DataRow values.
+                    || this.TryProcessDataSource(test, testMethodInfo, testContext, tests) */
+                    || this.TryProcessTestDataSourceTests(test, testMethodInfo, tests);
```
shyamnamboodiripad: Can we omit the false here?

```
return /* DataSourceAttribute discovery is disabled for now, since we cannot serialize DataRow values.
    this.TryProcessDataSource(test, testMethodInfo, testContext, tests) || */
    this.TryProcessTestDataSourceTests(test, testMethodInfo, tests);
```

src/Adapter/MSTest.CoreAdapter/Discovery/AssemblyEnumerator.cs
```diff
@@ -220,11 +196,191 @@ internal string GetLoadExceptionDetails(ReflectionTypeLoadException ex)
...
+            }
+        }
+
+        private bool TryProcessDataSource(UnitTestElement test, TestMethodInfo testMethodInfo, ITestContext testContext, List<UnitTestElement> tests)
+        {
+            UTF.DataSourceAttribute[] dataSourceAttributes = ReflectHelper.GetAttributes<UTF.DataSourceAttribute>(testMethodInfo.MethodInfo, false);
```
shyamnamboodiripad: consider: `var` (also other occurrences below)

src/Adapter/MSTest.CoreAdapter/Discovery/AssemblyEnumerator.cs
```diff
@@ -220,11 +196,191 @@ internal string GetLoadExceptionDetails(ReflectionTypeLoadException ex)
...
+                    return this.ProcessDataSourceTests(test, testMethodInfo, testContext, tests);
+                }
+                catch (Exception ex)
+                {
+                    var message = string.Format(CultureInfo.CurrentCulture, Resource.CannotEnumerateDataSourceAttribute, test.TestMethod.ManagedTypeName, test.TestMethod.ManagedMethodName, ex);
+                    PlatformServiceProvider.Instance.AdapterTraceLogger.LogInfo($"DynamicDataEnumarator: {message}");
+                    return false;
+                }
+            }
+            else if (dataSourceAttributes != null && dataSourceAttributes.Length > 1)
```
shyamnamboodiripad: consider: Add early outs

```
if (dataSourceAttributes == null || dataSourceAttributes.Length == 0)
{
    return false;
}

if (dataSourceAttributes.Length > 1)
{
    ...
    throw new InvalidOperationException(...);
}

// dataSourceAttributes.Length == 1
try
{
    return this.ProcessDataSourceTests(test, testMethodInfo, testContext, tests);
}
catch
{
    ...
    return false;
}
```

src/Adapter/MSTest.CoreAdapter/Discovery/AssemblyEnumerator.cs
```diff
@@ -220,11 +196,191 @@ internal string GetLoadExceptionDetails(ReflectionTypeLoadException ex)
...
+            }
+
+            try
+            {
+                int rowIndex = 0;
+
+                foreach (var dataRow in dataRows)
+                {
+                    // TODO: Test serialization
+                    rowIndex++;
```
shyamnamboodiripad, line 317: consider: Use a `for` loop in place of `foreach`

src/Adapter/MSTest.CoreAdapter/Discovery/AssemblyEnumerator.cs
```diff
@@ -220,11 +196,191 @@ internal string GetLoadExceptionDetails(ReflectionTypeLoadException ex)
...
+            }
+
+            return false;
+        }
+
+        private bool TryProcessTestDataSourceTests(UnitTestElement test, TestMethodInfo testMethodInfo, List<UnitTestElement> tests)
+        {
+            var methodInfo = testMethodInfo.MethodInfo;
+
+            UTF.ITestDataSource[] testDataSources = ReflectHelper.GetAttributes<Attribute>(methodInfo, false)?.Where(a => a is UTF.ITestDataSource).OfType<UTF.ITestDataSource>().ToArray();
```
shyamnamboodiripad: Looks like testDataSources can be null. Do we need an early return for that?

src/Adapter/MSTest.CoreAdapter/Discovery/AssemblyEnumerator.cs
```diff
@@ -220,11 +196,191 @@ internal string GetLoadExceptionDetails(ReflectionTypeLoadException ex)
...
+                    var discoveredTest = test.Clone();
+                    discoveredTest.DisplayName = displayName;
+                    discoveredTest.TestMethod.DataType = DynamicDataType.DataSourceAttribute;
+                    discoveredTest.TestMethod.Data = new[] { (object)rowIndex };
+                    tests.Add(discoveredTest);
+                }
+
+                return true;
+            }
+            catch
```
shyamnamboodiripad: Should this exception be logged (similar to other methods above and below)?

test/E2ETests/Smoke.E2E.Tests/DataRowTests.cs
```diff
@@ -26,7 +26,7 @@ public void ExecuteOnlyDerivedClassDataRowsWhenBothBaseAndDerviedClassHasDataRow
             // 4 tests of BaseClass.DataRowTestMethod - 3 data row results and 1 parent result
             // 3 tests of DerivedClass.DataRowTestMethod - 2 data row results and 1 parent result
             // Total 7 tests - Making sure that DerivedClass doesn't run BaseClass tests
-            this.ValidatePassedTestsCount(7);
+            this.ValidatePassedTestsCount(7 - 2);
```
shyamnamboodiripad: Any reason to not just say 5 instead of 7 - 2?

(Same question for below changes)

src/Adapter/MSTest.CoreAdapter/Helpers/ReflectHelper.cs
```diff
@@ -200,6 +200,20 @@ public override object InitializeLifetimeService()
...
+        {
+            Attribute[] attributeArray = GetCustomAttributes(methodBase, typeof(T), inherit);
+
+            var attributes = attributeArray as T[];
+            if (attributes != null)
+            {
+                return attributes;
+            }
+
+            return attributeArray?.Select(a => a as T)?.Where(a => a != null)?.ToArray();
```
shyamnamboodiripad: This can be simplified as `return attributeArray?.Where(a => a is T).ToArray()`

src/Adapter/MSTest.CoreAdapter/Helpers/ReflectHelper.cs
```diff
@@ -200,6 +200,20 @@ public override object InitializeLifetimeService()
             return null;
         }
 
+        internal static T[] GetAttributes<T>(MethodBase methodBase, bool inherit)
+            where T : Attribute
+        {
+            Attribute[] attributeArray = GetCustomAttributes(methodBase, typeof(T), inherit);
+
+            var attributes = attributeArray as T[];
```
shyamnamboodiripad: Looks like this will never succeed since GetCustomAttributes below always returns `attributesArray.OfType<Attribute>().ToArray();`. Consider removing this block to avoid confusion.

src/Adapter/MSTest.CoreAdapter/Helpers/ReflectHelper.cs
shyamnamboodiripad: Would it make sense to add an early return for `attributeArrray == null` above and omit the `?.` operators on this line?

shyamnamboodiripad (commented):
I haven't finished reviewing all the changes but wanted to post the comments so far. Will continue reviewing the remaining code tomorrow.

src/Adapter/MSTest.CoreAdapter/Discovery/TypeEnumerator.cs
Haplois: No, this was set during execution.

src/Adapter/MSTest.CoreAdapter/Discovery/AssemblyEnumerator.cs
Haplois: So, should we discover all tests, even if they are ignored or in an ignored class?

scripts/Build.ps1
Haplois: No, it was a merge error, fixed.

src/Adapter/MSTest.CoreAdapter/Discovery/AssemblyEnumeratorWrapper.cs
Haplois: Thanks for pointing my laziness out :D I moved all those into a separate PR, though it's too late for your review now :)

AbhitejJohn:
Is this going to be configurable through runsettings? For instance, Can a user choose not to discover tests from a large excel file if they do not want to unless they run it? That would also be useful given this changes the public extensibility layer.

shyamnamboodiripad:
> Is this going to be configurable through runsettings? For instance, Can a user choose not to discover tests from a large excel file if they do not want to unless they run it? That would also be useful given this changes the public extensibility layer.

That's a good point and I understand why this could be useful. However, if the test is not split at discovery time, this means that to run a single dynamic test one would have to always re-discover and run all of them. This could be equally annoying I think. A setting would also add concept count + difference in behavior when the setting is on v/s off could cause confusion as one moves between different repos.

If we do add such a setting, we should probably limit it to be applicable for dynamic tests only. Tests with data points defined in source should be expanded regardless.

I also think a better alternative for the issue you raise would be to fully decouple test discovery from build in VS. If there were a way to do that (and trigger discovery explicitly via some sort of refresh button or via the run all tests button), one would avoid paying any discovery cost until one is ready to run tests, while at the same time retaining the ability to run individual dynamic tests.

@AbhitejJohn @Haplois Do you know if other frameworks (xUnit / NUnit) allow to configure the splitting behavior?

src/Adapter/MSTest.CoreAdapter/Discovery/AssemblyEnumerator.cs
shyamnamboodiripad: I think we should discover all tests. Ignored tests should be discovered, but should return the "skipped" status upon execution (and should be displayed with a yellow skipped icon in VS). This would match the behavior for other frameworks.

If we skip discovery for ignored tests they would not be displayed in VS. This means it would be impossible to un-skip and execute them again without running a build to make them appear first after un-skipping.

One may also want to know how many tests in a given solution are currently ignored by looking at total count of ignored tests in VS (and fix them up / un-skip by navigating to them one-by-one). This would again not be impossible if the skipped tests are not discovered in the first place.

Some frameworks also have a conditional skip attribute (e.g. skip if OS=Linux) where the condition is evaluated at run time. I am not sure if mstest has such capability - but this would be another reason to only skip at run time.

src/Adapter/MSTest.CoreAdapter/Discovery/TypeEnumerator.cs
shyamnamboodiripad: Ah thanks please see the other comment around ignores.

src/Adapter/MSTest.CoreAdapter/Discovery/AssemblyEnumerator.cs
Haplois: It should always be `true`, but any part of the chain is obsolete; it might not contain those. Since we depend on those properties, if they don't exist, we bail out early—added comment.

src/Adapter/MSTest.CoreAdapter/Discovery/AssemblyEnumerator.cs
Haplois: `dataRows` is `IEnumerable<object>`, `foreach` is more convinient here.

src/Adapter/MSTest.CoreAdapter/Discovery/AssemblyEnumerator.cs
Haplois: Converted to `finally` since we already `catch` the error and log it outer levels.

src/Adapter/MSTest.CoreAdapter/Helpers/DataSerializationHelper.cs
Haplois: `DataContractJsonSerializer` doesn't support `Span`.

src/Adapter/MSTest.CoreAdapter/Helpers/ReflectHelper.cs
Haplois: In `~/src/Adapter/PlatformServices.Desktop/Utilities/DesktopReflectionUtility.cs` it might return a `T[]`, and that might be the most common path for this method.

PR #813, opened 2021-04-10
Added WinUI TestAdapter.

azchohfi:
This adds a new adapter TFM (net5.0-windows10.0.18362.0), which is used for WinUI projects.
Complemented by https://github.com/microsoft/vstest/pull/2849


Haplois (approved):

dotMorten:
@azchohfi Thank you!

jlaanstra:
This PR has broken running any tests targeting net5.0-windows that do not use WinUI3. Supported for packaged WinUI3 tests doesn't seem worth this regression. Can we revert until there is a solution that works for both cases?

PR #819, opened 2021-04-15
Handle possible null ref in error handler

nohwnd:
I have a dump that throws access violation exception in native code and the test run fails with: 

`An exception occurred while invoking executor 'executor://mstestadapter/v2': Object reference not set to an instance of an object.`

I looked through the code and the only way I can get this result is when ex would be null in the handler. Otherwise it fails with different error, either from the asserts in the `StackTraceHelper` methods or in upstream error handlers.

### DIFF
--- a/src/Adapter/MSTest.CoreAdapter/Execution/TestMethodInfo.cs
+++ b/src/Adapter/MSTest.CoreAdapter/Execution/TestMethodInfo.cs
@@ -717,11 +717,27 @@ namespace Microsoft.VisualStudio.TestPlatform.MSTest.TestAdapter.Execution
             }
             catch (Exception ex)
             {
+                if (ex == null)
+                {
+                    // It seems that ex can be null in some rare cases when initialization fails in native code.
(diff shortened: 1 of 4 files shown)


src/Adapter/MSTest.CoreAdapter/Execution/TestMethodInfo.cs
```diff
@@ -735,6 +751,11 @@ private object CreateTestClassInstance(TestResult result)
             return classInstance;
         }
 
+        private Exception NullReferenceException()
```
Haplois: Was this for testing purposes? Is it used anywhere?

Haplois (approved):
LGTM, with a comment.

src/Adapter/MSTest.CoreAdapter/Execution/TestMethodInfo.cs
nohwnd: It's code intellisense that was too clever, reverting.

PR #826, opened 2021-05-12
Reverting an unintended breaking-change.

Haplois:
Reverting an unintended breaking change in reflection.

Thank you, @azchohfi, for reporting it; this makes #824 unnecessary.


nohwnd (approved):
Would have been nice to have more details on what is being fixed, or a linked issue.

Haplois:
@nohwnd, [This][1] method returned an empty array when it couldn't find any matching attributes. We refactored it, but it started returning `null` as an unintended consequence. Fixed that.

[1]: https://github.com/microsoft/testfx/pull/826/files#diff-f9e4ebe7891ad006c4be07397779eebbc20eebe85977eff03404e0718516dc8bL96

PR #855, opened 2021-06-01
Resolves: Add native GitHub continuous code security and quality analysis

### LINKED ISSUE #854: Add native GitHub continuous code security and quality analysis
- [CodeQL](https://github.com/github/codeql) is an open source code scanning tool for discovering vulnerabilities across a codebase

Pipeline Foundation is **a non-profit initiative** with the sole purpose of giving back to the IT community by assisting OSS projects with DevOps implementations and best practices.

Built with ❤ by [Pipeline Foundation](https://pipeline.foundation).

aleks-ivanov:
- add `codeql-analysis.yml` which automatically enables CodeQL code security and quality scanner. It executes on every push commit, PR, manually and every day at 8:00AM UTC. A scan check can be viewed on commits and PRs. All alerts and fix suggestion can be viewed at **Security** tab -> **Code scanning alerts** -> **CodeQL**

Resolves #854

### DIFF
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -1,0 +1,39 @@
+name: CodeQL Analysis
+
+on:
+  push:
+  pull_request:
+  schedule:
+    - cron: '0 8 * * *'
+  workflow_dispatch:
+
+jobs:
+  analyze:
+    name: Analyze
+    runs-on: windows-latest
+    steps:
+      - name: Add msbuild to PATH
+        uses: microsoft/setup-msbuild@v1.0.2
+
+      - name: Checkout repository
+        id: checkout_repo
+        uses: actions/checkout@v2
+
+      - name: Initialize CodeQL
+        id: init_codeql
+        uses: github/codeql-action/init@v1
+        with:
+          language: csharp
+          queries: security-and-quality
+
+      - name: Build solution
+        id: build_solution
+        run: |
+          nuget restore .\TestFx.sln
+          msbuild .\TestFx.sln /p:UseSharedCompilation=false
+
+      - name: Perform CodeQL Analysis
+        id: analyze_codeql
+        uses: github/codeql-action/analyze@v1
+
+# Built with ❤ by [Pipeline Foundation](https://pipeline.foundation)


.github/workflows/codeql-analysis.yml
```diff
@@ -0,0 +1,39 @@
...
+        id: init_codeql
+        uses: github/codeql-action/init@v1
+        with:
+          language: csharp
+          queries: security-and-quality
+
+      - name: Build solution
+        id: build_solution
+        run: |
+          nuget restore .\TestFx.sln
```
Haplois, line 32:
```suggestion
```

.github/workflows/codeql-analysis.yml
```diff
@@ -0,0 +1,39 @@
...
+          msbuild .\TestFx.sln /p:UseSharedCompilation=false
```
Haplois, line 33:
```suggestion
          .\build.cmd -configuration Debug -Full
```

.github/workflows/codeql-analysis.yml
```diff
@@ -0,0 +1,39 @@
...
+
+      - name: Perform CodeQL Analysis
+        id: analyze_codeql
+        uses: github/codeql-action/analyze@v1
```
Haplois, line 37: Can you exclude `~/test` folder from analysis?

Haplois (changes requested):

Haplois:
Thanks again, we don't have edit permissions on your PR could you review the changes we requested.

.github/workflows/codeql-analysis.yml
aleks-ivanov: Without the property `/p:UseSharedCompilation=false` applied to the build process, CodeQL doesn't recognize the build and [throws an error](https://github.com/aleks-ivanov/testfx/runs/2819883514?check_suite_focus=true).

If the script is mandatory, a possible solution might be a custom parameter for build command properties that accepts whole declarations and applies them to the build command.

If the script is not mandatory, we think it would be easier to implement and maintain, just to use the `nuget restore` and `msbuild` commands with the required properties like configuration, platform etc. so the `UseSharedCompilation` property could also be added, something like this:

`msbuild .\TestFx.sln /p:Configuration=Debug /p:Platform='Any CPU' /p:UseSharedCompilation=false`

.github/workflows/codeql-analysis.yml
aleks-ivanov: CodeQL doesn't have the option to exclude folders from the analysis for compiled languages. The suggestion in [their documentation](https://docs.github.com/en/code-security/secure-coding/automatically-scanning-your-code-for-vulnerabilities-and-errors/configuring-code-scanning#specifying-directories-to-scan) is to implement this within the build.

The solution we came up with is during the execution of the run, before the build of the solution file is started, all the test projects to be removed from the solution file. This results in only the `src` and one `sample` cs-projects to be built and [CodeQL analyzing only them](https://github.com/aleks-ivanov/testfx/actions/runs/936125970).

The solution looks like this:
```
Get-ChildItem -Path test/ -Include *.csproj, *.fsproj -Recurse | ForEach-Object { 
    dotnet sln .\TestFx.sln remove $_.FullName 
}

nuget restore .\TestFx.sln

msbuild .\TestFx.sln /p:Configuration=Debug /p:Platform='Any CPU' /p:UseSharedCompilation=false
```

Which results in these project be left in the solution file and only them be restored, built and analyzed:
```
src\Adapter\MSTest.CoreAdapter\MSTest.CoreAdapter.csproj
src\Adapter\PlatformServices.Interface\PlatformServices.Interface.csproj
src\Adapter\PlatformServices.Portable\PlatformServices.Portable.csproj
src\Adapter\PlatformServices.Universal\PlatformServices.Universal.csproj
src\TestFramework\MSTest.Core\MSTest.Core.csproj
src\Adapter\PlatformServices.Desktop\PlatformServices.Desktop.csproj
src\Adapter\PlatformServices.Shared\PlatformServices.Shared.shproj
src\TestFramework\Extension.Desktop\Extension.Desktop.csproj
src\TestFramework\Extension.Core\Extension.Core.csproj
src\TestFramework\Extension.Shared\Extension.Shared.shproj
src\TestFramework\Extension.UWP\Extension.UWP.csproj
samples\FxExtensibility\FxExtensibility.csproj
src\Adapter\PlatformServices.NetCore\PlatformServices.NetCore.csproj
src\TestFramework\Extension.WinUI\Extension.WinUI.csproj
src\Adapter\PlatformServices.WinUI\PlatformServices.WinUI.csproj
```

.github/workflows/codeql-analysis.yml
Haplois: Can you enable commits from maintainers to these PRs? I can fix the scripts to support this.

aleks-ivanov: @Haplois since I've created this PR from our organization, this option is ignored by default. To speed up the process we've made you a maintainer to the fork, so after you accept the invitation, you should be able to commit to this PR 🙂

Evangelink:
@aleks-ivanov Hey thank you for the contribution. We made some tests and it seems like the target machine is missing .net 4.5 so restore is failing - preventing us to fully test the codeql feature. Is it something you could fix? Otherwise we may want to close this PR to avoid wasting your time with constant rebases.

aleks-ivanov:
> @aleks-ivanov Hey thank you for the contribution. We made some tests and it seems like the target machine is missing .net 4.5 so restore is failing - preventing us to fully test the codeql feature. Is it something you could fix? Otherwise we may want to close this PR to avoid wasting your time with constant rebases.

The `windows-2022` (which resolves from `windows-latest`) environment includes only .NET 4.8, which does not include .NET version 4.0-4.5.
However earlier environment version `windows-2019` includes .NET 4.7.2 as well, which should include .NET 4.5.

.github/workflows/codeql-analysis.yml
Evangelink: @aleks-ivanov I think we might still need to rely on `build.cmd` because simply doing nuget restore doesn't restore enough of the dependencies we need.

Evangelink:
Hi @aleks-ivanov! I have just learn that our infra teams have enabled codeql on our AzDo builds. We will sync with them and will probably enable some analysis on the public builds too.

Based on that I will have to close this PR. We are sorry to have made you work for nothing but we really appreciate your contribution and commitment to improve quality.

Have a lovely day!

Issue #875, opened 2021-06-14
MSTest.TestFramework 2.2.5-preview-20210605-01: 'DateTime' in DynamicData looses significant digits less than milliseccond


jvrooyen:
## Description
Significant digits smaller than milliseconds of `DateTime` values in DynamicData are lost when the value enters the TestMethod. 

## Steps to reproduce
- MSTest.TestFramework 2.2.5-preview-20210605-01
- MSTest.TestAdapter 2.2.5-preview-20210605-01
- Microsoft.NET.Test.Sdk 16.10.0
- Net Core 3.1

```
[TestMethod]
[DynamicData(nameof(TestCases), DynamicDataSourceType.Method)]
public void DateTimeTest(DateTime value1, DateTime value2, string _)
{
    var actual = value1 == value2;

    Assert.AreEqual(false, actual);
}

public static IEnumerable<object[]> TestCases()
{
    var now = new DateTime(2021, 06, 14, 12, 13, 14);

    yield return new object[] { now, now + TimeSpan.FromTicks(1),       "100 ns" };
    yield return new object[] { now, now + TimeSpan.FromTicks(10),      "  1 us" };
    yield return new object[] { now, now + TimeSpan.FromTicks(100),     " 10 us" };
    yield return new object[] { now, now + TimeSpan.FromTicks(1000),    "100 us" };
    yield return new object[] { now, now + TimeSpan.FromTicks(10000),   "  1 ms" };
    yield return new object[] { now, now + TimeSpan.FromTicks(100000),  " 10 ms" };
    yield return new object[] { now, now + TimeSpan.FromTicks(1000000), "100 ms" };
}
```

## Expected behavior
All 7 test cases should pass

## Actual behavior
The first 4 test cases fail, the last 3 pass.

## Environment
Windows 2004 build 19041.1052

Haplois:
This is because the default serializer for `DateTime` doesn't preserve the least significant digits.

Haplois:
This is fixed by @jvrooyen on #907. Thank you for your contribution!

PR #907, opened 2021-07-08
Fix DateTime looses significant digits in DynamicData (#875)

jvrooyen:
Changed the DateTime formatter for the DataSerializer
Added unit tests


Haplois (approved):
@jvrooyen thank you for the contribution! LGTM.

PR #937, opened 2021-08-05
Enable internal testclass discovery

### LINKED ISSUE #934: Enable discovery of internal test classes
My team finds it convenient to extract tests for common attributes into reusable, abstract, generic classes and inherit from these classes in the type-specific tests. This approach doesn’t work for internal types, however, because the test class must be public in order for MSTest to discover it.

In the below example, FancyString is not intended to be a public type of the assembly. (It’s marked internal just so that it can be tested.) Because FancyString is marked internal, it’s a compile-time error to mark the class FancyStringsAreCaseInsensitive as public. But since FancyStringsAreCaseInsensitive is marked internal it’s not discovered.

FancyStringTests.cs:
```C#
using System;
using DevProject1;

namespace TestProject1
{
    [TestClass]
    public class FancyStringTests
    { 
        [TestClass]
        // This class will not be discovered by MSTest because it is internal rather than public.
        // But, since FancyString is internal, marking the test class public will produce a compilation error.
        // This forces us to either make FancyString public or abandon this technique for reusing test classes.
        internal class FancyStringsAreCaseInsensitive : CaseInsensitivityTests<FancyString>
        {
            protected override Tuple<FancyString, FancyString> EquivalentInstancesWithDistinctCasing { get; }
        }
    }
}
```

FancyString.cs:
```C#
using System;

namespace DevProject1
{
    internal class FancyString
    {
        // TODO
    }
}
```

CaseInsensitivityTests.cs:
```C#
using System;

namespace TestProject1
{
    internal abstract class CaseInsensitivityTests<T>
    {
        protected abstract Tuple<T, T> EquivalentInstancesWithDistinctCasing { get; }

        [TestMethod]
        public void EqualityIsCaseInsensitive()
        {
            var (left, right) = EquivalentInstancesWithDistinctCasing;
            Assert.AreEqual(left, right);
        }

        [TestMethod]
        public void HashCodeIsCaseInsensitive()
        {
            // TODO
        }
    }
}
```

cimryan:
This PR resolves #934 

Key points:
* Discovery of internal test classes (i.e. classes with the ```internal``` accessibility declaration and the ```TestClass``` attribute) is enabled via the application of the newly-introduced ```DiscoverInternals``` attribute to the assembly containing the tests.
* Similarly, discovery of internal test methods (i.e. methods with the ```internal``` accessibility declaration and the ```TestMethod``` attribute) is enabled via the same ```DiscoverInternals``` attribute. I.e., the application of the ```DiscoverInternals``` attribute to an assembly enables discovery of both internal test classes and internal test methods within that assembly.
* The change is backward-compatible: Unless the attribute is applied to the assembly internal test classes and methods continue to be ignored.
* Performance impact is minimized: In existing scenarios, the only impact is one check for the presence of the attribute on the already-loaded assembly object. In the newly-enabled scenario, additional discovery is only performed on classes which are marked [TestClass]. (I.e., this change _does not_ add analysis of all internal types in the test assembly.)

To reiterate the issue: before this PR the ```FancyStringsAreCaseInsensitive``` tests in the following example will not be discovered, and afterward, they will be.

### Test project:
FancyStringTests.cs:
```C#
using System;
using DevProject1;

namespace TestProject1
{
    [TestClass]
    public class FancyStringTests
    { 
        [TestClass]
        // Before this PR the following tests would not be discovered. After this PR they will be. 
        // The test class must be marked internal because FancyString is internal, and it is a compile-time error
        // to have a public class derived from a generic class with an internal type parameter.
        internal class FancyStringsAreCaseInsensitive : CaseInsensitivityTests<FancyString>
        {
            protected override Tuple<FancyString, FancyString> EquivalentInstancesWithDistinctCasing { get; }
        }
    }
}
```

AssemblyAttributes.cs
```C#
// This PR introduces this attribute type.
[assembly: Microsoft.VisualStudio.TestTools.UnitTesting.DiscoverInternals]
```

CaseInsensitivityTests.cs:
```C#
using System;

namespace TestProject1
{
    internal abstract class CaseInsensitivityTests<T>
    {
        protected abstract Tuple<T, T> EquivalentInstancesWithDistinctCasing { get; }

        [TestMethod]
        public void EqualityIsCaseInsensitive()
        {
            var (left, right) = EquivalentInstancesWithDistinctCasing;
            Assert.AreEqual(left, right);
        }

        [TestMethod]
        public void HashCodeIsCaseInsensitive()
        {
            // TODO
        }
    }
}
```

### Dev project:
FancyString.cs:
```C#
using System;

namespace DevProject1
{
    // This type is not intended to be part of the public interface of the assembly. It's declared internal
    // just for testability.
    internal class FancyString
    {
        // TODO
    }
}
```


Haplois (commented):
LGTM, but we need to make sure it backwards and forwards compatible.

I am just wondering should we allow this attribute on classes too, to control this class-by-class basis. Another, improvement might be having way to enable it using RunSettings file..

nohwnd (changes requested):
The code currently does not allow public classes in internal types (also why the public class nested in a nested private class does not discover, but internal class there would discover). I have the code to make this more consistent so we allow any public or internal classes as long as they are "visible". Will push that to this branch.

I would also like this to extend to internal methods on those classes that have the attribute, and possibly rename it to DiscoverInternalTestClassesAndTestMethods. This will enable tests that generate data to use internal classes for the payload, and also gives the option a little bit more significance, because then we are introducing a slightly more breaking change. (Still user would have easy way out, just remove the attribute from your internal classes).

nohwnd:
image

Code: https://gist.github.com/nohwnd/eebd17f83a7498ae9d55988cddc95936 
A nice infographic of the visibility levels: https://stackoverflow.com/a/22915745/3065397 

I think I covered all the reasonable combinations.

nohwnd:
@cimryan pushed my fixes, and few more tests. Hopefully it covers every reasonable edge case. Would you be able to work on the internal method discovery so we can merge this a cohesive change?

cimryan:
@Haplois Thanks! I haven't been able to come up with a scenario in which putting the attribute on the test class would be necessary or preferable. If you have one (or more!) in mind, could you sketch the classes/methods that would be involved to help me understand it?

cimryan:
@nohwnd Thanks! I think the changes you pushed make sense.

I'm not sure I understand what you mean by "This will enable tests that generate data to use internal classes for the payload".

Can you provide a few simple classes and methods that would demonstrate the additional scenario(s) that you'd like to see supported? I'd like to see the code that doesn't work today but that you think should. I think that'd really make it clear, for me.

Perhaps create another Issue for the new scenario? (Which could also be resolved by this PR if we think that makes sense.)

cimryan:
@nohwnd I thought you might be referring to using ```internal``` types with the ```DynamicData``` attribute, but the following seems to work when the discovery of ```internal``` test classes is enabled, so I'm still not sure what additional scenario you'd like supported.

```C#
    [TestClass]
    internal class DynamicDataTest
    {
        [DataTestMethod]
        [DynamicData(nameof(DynamicData), DynamicDataSourceType.Property)]
        public void DynamicDataTestMethod(FancyString fs)
        {

        }

        public static IEnumerable<object[]> DynamicData => new []
        {
            new object[] { new FancyString() }
        };
    }
```

Will you please provide an example of a test that isn't discovered?

nohwnd:
Pretty much like your example but with the test method internal and not public. I agree it works when your method is public in internal class, but IMHO internal methods should be discovered as well. It avoids a lot of confusion, and makes it much simpler to explain the behavior, and more convenient for the user:

A user probably starts with a public test class, and public test method. They add internal parameter and project fails to compile because the parameter type is internal (inconsistent accessibility). They will most likely make the method internal. It will compile but the test won't be discovered. They google for the assembly attribute, add it, and still nothing. They will make the whole class internal, and still nothing. They will finally read our documentation saying that the class needs to be internal but the method needs to be public, confused they will do that and it will start to work. 

OR: 

A user probably starts with a public test class, and public test method. They add internal parameter and project fails to compile because the parameter type is internal (inconsistent accessibility). They will most likely make the method internal. It will compile but the test won't be discovered. They google for the assembly attribute, add it, and it works.


```csharp
using System;
using System.Reflection;

[assembly: Microsoft.VisualStudio.TestTools.UnitTesting.DiscoverInternalTestClasses]

namespace TestProject26
{
    // this would be internal class coming from another project
    // via IVTs
    internal class User
    {
        internal string Name { get; set; }
    }

    [TestClass]
    internal class UnitTest1
    {
        [TestMethod]
        [UserDataFixture]
        internal void TestMethod1(User user, string name)
        {
             Assert.AreEqual(name, user.Name);
        }
    }

    internal class UserDataFixtureAttribute : Attribute, ITestDataSource
    {
        {
            return new List<object[]> {
                new object[] {
                    new User { Name = "Jakub"},
                    "Jakub"

                },
                new object[] {
                    new User { Name = "Medeni"},
                    "Medeni"
                },
            };
        }

        {
            return methodInfo.Name;
        }
    }

    [TestClass]
    internal class UnitTest2
    {
        [TestMethod]
        [DynamicData(nameof(Data), DynamicDataSourceType.Method)]       
        internal void TestMethod2(User user, string name)
        {

        }

        internal static IEnumerable<object[]> Data()
        {
            return new List<object[]> {
                new object[] {
                    new User { Name = "Jakub"},
                    "Jakub"

                },
                new object[] {
                    new User { Name = "Medeni"},
                    "Medeni"
                },
            };
        }
    }
}
```

cimryan:
@nohwnd, Thanks!

I'll make the change to also discover internal test methods.

I propose changing the attribute name in this PR to ```DiscoverInternals```. This attribute is in the namespace ```Microsoft.VisualStudio.TestTools.UnitTesting```, and "Discover" is implicitly about test discovery, so I think that name is sufficiently specific without being overwhelmingly verbose.

nohwnd:
Will give it a test tomorrow hopefully. Today is already full, sorry :/ 

I need to look at the other assembly options to see how I feel about the new name. I like that it is not overly verbose.

nohwnd:
Kicked off a new build to see if the timeout in e2e tests was just flakiness.

nohwnd:
Still could not test this, sorry :/

cimryan:
@nohwnd I'd hope that the automated tests would obviate additional testing. What tests do you think are missing?

nohwnd:
I am missing some time sitting down and trying to come up with additional edge cases, much like with first draft where the behavior seemed correct on first look, but needed some additional work. 🙂

nohwnd:
@cimryan LGTM, please fix the description of the PR to reflect the final attribute name DiscoverInternals.

nohwnd (approved):

Haplois (changes requested):
I am OOF, but I like to quickly review this tonight or tomorrow 😊

cimryan:
@nohwnd I updated the description, per your request.

cimryan:
@Haplois are you ready to approve?

Haplois (approved):
LGTM

PR #940, opened 2021-08-18
Add task aware string writer with asynclocal

nohwnd:
Change implementation of ThreadSafeStringWriter to use AsyncLocal static that will avoid stealing the "context" when another ThreadSafeStringWriter is created. Add parameter to distinguish the different kinds of output that we capture because otherwise all output types would all go into the same string writer (e.g., Debug and Console output for one test would be put together). 

This implementation is not great, but it avoids putting the implementation into the .Core which is Portable and can't have AsyncLocal. 

It works nicely with non-async parallel methods as well as with async parallel methods, and works with non-parallel execution as well.

The setup and cleanup output is still a bit problematic, because setup runs in the context of one of the tests, and we don't know which test will end up capturing it. The cleanup output is not shown in TE but is shown in console because it is sent out as a message (not my change, this is how it always worked in v2). 

It would be also nice to surface All output (currently only capturing this in Debug build), to see in which order the output was produced, which might be useful in some cases, but would get overwhelming fast.

- [x] implement for net core
- [x] share implementation with .net framework - we build against net45 which does not support asynclocal, net46 is the minimal version.
- [x] change the identifiers of the different outputs?
- [x] keep the old implementation for ns10 working 

Further improvements: 
- [ ] associate teardown output to something, or at least add the type name
- [ ] have all output in chronological order associated with the run and shown in TE somewhere


nohwnd:
Test proj: 

```csproj
<!-- file TestProject29.csproj -->

  <PropertyGroup>
    <TargetFramework>net5.0</TargetFramework>

  </PropertyGroup>

  <ItemGroup>
    <PackageReference Include="Microsoft.NET.Test.Sdk" Version="16.9.4" />
    <PackageReference Include="MSTest.TestAdapter" Version="99.99.99-dev" />
    <PackageReference Include="MSTest.TestFramework" Version="99.99.99-dev" />
    <PackageReference Include="coverlet.collector" Version="3.0.2" />
  </ItemGroup>

</Project>

```

```cs
// file UnitTest1.cs
using System;
using System.Threading.Tasks;

[assembly: Parallelize(Workers = 0, Scope = ExecutionScope.MethodLevel)]

namespace TestProject29
{
    [TestClass]
    public class UnitTest1
    {
        public TestContext TestContext { get; set; }


        [ClassInitialize]
        public static async Task SetupTests(TestContext testContext)
        {
            testContext.WriteLine("test context in setup");
            Console.WriteLine("class setup");
            System.Threading.Thread.Sleep(1);
        }

        [ClassCleanup]
        public static async Task CleanupTests()
        {
            Console.WriteLine("class cleanup");
        }

        public static bool Flag2 { get; private set; }
        public static bool Flag3 { get; private set; }

        [TestMethod]
        [DataRow(1)]
        [DataRow(2)]
        public async Task TestMethod1(int a)
        {
            TestContext.WriteLine("test context in test");
            Console.WriteLine($"test1-{a}");
            Console.WriteLine($"test1-{a}");
            Console.WriteLine($"test1-{a}");
            while (!UnitTest1.Flag2)
            {
            }
            Console.WriteLine($"test1-{a}");
            Console.WriteLine($"test1-{a}");
            Console.WriteLine($"test1-{a}");
            while (!UnitTest1.Flag3)
            {
            }
            Console.WriteLine($"test1-{a}");
            Console.WriteLine($"test1-{a}");
            Console.WriteLine($"test1-{a}");
        }

        [TestMethod]
        public async Task TestMethod2()
        {
            Console.WriteLine("test2");
            Console.WriteLine("test2");

            UnitTest1.Flag2 = true;

            Console.WriteLine("test2");
            Console.WriteLine("test2");
        }

        [TestMethod]
        public async Task TestMethod3()
        {
            Console.WriteLine("test3");
            Console.WriteLine("test3");

            UnitTest1.Flag3 = true;

            Console.WriteLine("test3");
            Console.WriteLine("test3");
        }

    }

    [TestClass]
    public class UnitTest4
    {
        [TestMethod]
        public async Task TestMethod4()
        {
            Console.WriteLine("test4");
            Console.WriteLine("test4");
            Console.WriteLine("test4");
            throw new Exception("aaa");
        }
    }
}

```

# Current version: 

image
image

# 2.1.0: 

image

nohwnd:
[TestProject42.zip](https://github.com/microsoft/testfx/files/7582035/TestProject42.zip)
Here is a test proj if you want to take it for spin.

Haplois (changes requested):
LGTM, verifying signing - we can merge after that.

Haplois (approved):

PR #944, opened 2021-08-24
Enable internal testclass discovery

Haplois:
* Enable discovery of internal test classes

It's useful to extract sets of test methods which establish some common criteria into reusable, generic, abstract test classes from which concrete test classes for specific types can be derived.

However, this technique requires that the type under test be as accessible as the test class, itself. In particular, it's a compile-time error to have a public class derived from a generic class parameterized with an internal type. So, either the type must be made public or the test class must be made internal, in order for the compiler to be satisfied.

Before the change in this PR, the test framework would only discover tests which were declared in public classes. This meant that the option of declaring the test class internal wasn't really an option at all: doing so would prevent the tests from running. So, types which should have been internal were marked public, just so they could be tested.

The following code reiterates the dilemma. After this commit, the tests in the ```FancyStringsAreCaseInsensitive``` class will be discovered.

FancyStringTets.cs
```C#
using System;
using DevProject1;

namespace TestProject1
{
    [TestClass]
    public class FancyStringTests
    {
        [TestClass]
        // This class will not be discovered by MSTest because it is internal rather than public.
        // But, since FancyString is internal, marking the test class public will produce a compilation error.
        // This forces us to either make FancyString public or abandon this technique for reusing test classes.
        internal class FancyStringsAreCaseInsensitive : CaseInsensitivityTests<FancyString>
        {
            protected override Tuple<FancyString, FancyString> EquivalentInstancesWithDistinctCasing { get; }
        }
    }
}
```

FancyString.cs
```C#
using System;

namespace DevProject1
{
    internal class FancyString
    {
        // TODO
    }
}
```

CaseInsensitivityTests.cs
```C#
using System;

namespace TestProject1
{
    internal abstract class CaseInsensitivityTests<T>
    {
        protected abstract Tuple<T, T> EquivalentInstancesWithDistinctCasing { get; }

        [TestMethod]
        public void EqualityIsCaseInsensitive()
        {
            var (left, right) = EquivalentInstancesWithDistinctCasing;
            Assert.AreEqual(left, right);
        }

        [TestMethod]
        public void HashCodeIsCaseInsensitive()
        {
            // TODO
        }
    }
}
```

* Add AnInternalTestClassDerivedFromAPublicAbstractGenericBaseClassForAnInternalTypeIsDiscovered

This test establishes support for the principal motivating scenario for the discovery of internal test classes.

* Add support for the discovery of internal test methods

This commit renames DiscoverInternalTestClassesAttribute to DiscoverInternalsAttribute, and causes the presence of that attribute in the assembly to additionally discover test methods which are declared internal, in addition to test methods which are declared public.

Co-authored-by: nohwnd <me@jakubjares.com>
Co-authored-by: Medeni Baykal <433724+Haplois@users.noreply.github.com>


nohwnd:
To add to the reasoning for this PR. When you have an enum that is internal, and you use it as a parameter to a test, you cannot compile when you don't discover internal classes / methods. This is apparent especially when your production library uses InternalsVisibleTo to show types to your test library that would normally not be visible.

PR #950, opened 2021-08-27
Resolve dependencies from GAC

### LINKED ISSUE #949: Deployment items are not copied when they are dependency of a GAC assembly
## Description
When a Test has deployment item attribute, the dlls from the source are copied to Out folder in TestResults. The test project can directly or transitively depend on assembly that is normally brought in by netstandard, such as System.ValueTuple. When dependencies are copied over, System.ValueTuple is not copied to the Out folder because no dependencies of GAC assemblies are copied over. 

When the test project uses a newer version of that assembly, the assembly will fail to resolve from GAC and won't be present in the Out folder and the test will fail. 

## Expected behavior
System.ValueTuple is copied into the output folder. 

## Environment
NET Framework project with MSTest

nohwnd:
When there are dependencies that the project installs, but they would only resolve via a reference that is located in GAC (e.g. System.ValueTuple via netstandard) then the deployment into Out folder would ignore it. 

If user then installs a version that is newer than what is in GAC it will get into their assembly redirects, but won't get copied into Out folder, resulting into TypeLoad exception for tests with DeploymentItem attribute. 

This change fixes the resolver to look through all the dependencies including GAC dependencies and then copy over only the ones that are found in the bin folder, because ultimately Out should be just a subset (or the same) as the contents in bin folder. 

Looking through GAC assemblies does not seem to add significant overhead the whole resolve is under 300ms so hopefully we don't need an option to configure enabling this.

(The new log messages won't go into diag log, because it probably is not correctly initialized in the new appdomain that loads the dll, but they will be written to Debug Trace and can be observed by DebugView++ or DebugView.)

Fix #949

### DIFF
--- a/src/Adapter/PlatformServices.Desktop/Deployment/AssemblyLoadWorker.cs
+++ b/src/Adapter/PlatformServices.Desktop/Deployment/AssemblyLoadWorker.cs
@@ -154,7 +154,7 @@ namespace Microsoft.VisualStudio.TestPlatform.MSTestAdapter.PlatformServices.Dep
         /// <param name="result"> The result. </param>
         /// <param name="visitedAssemblies"> The visited Assemblies. </param>
         /// <param name="warnings"> The warnings. </param>
-        private void ProcessChildren(Assembly assembly, IList<string> result, HashSet<string> visitedAssemblies, IList<string> warnings)
+        private void ProcessChildren(Assembly assembly, IList<string> result, ISet<string> visitedAssemblies, IList<string> warnings)
         {
             Debug.Assert(assembly != null, "assembly");
 
@@ -226,7 +226,7 @@ namespace Microsoft.VisualStudio.TestPlatform.MSTestAdapter.PlatformServices.Dep
         /// <param name="visitedAssemblies"> The visited Assemblies. </param>
         /// <param name="warnings"> The warnings. </param>
         [System.Diagnostics.CodeAnalysis.SuppressMessage("Microsoft.Design", "CA1031:DoNotCatchGeneralExceptionTypes", Justification = "Requirement is to handle all kinds of user exceptions and message appropriately.")]
-        private void GetDependentAssembliesInternal(string assemblyString, IList<string> result, HashSet<string> visitedAssemblies, IList<string> warnings)
+        private void GetDependentAssembliesInternal(string assemblyString, IList<string> result, ISet<string> visitedAssemblies, IList<string> warnings)
         {
             Debug.Assert(!string.IsNullOrEmpty(assemblyString), "assemblyString");
 
--- a/src/Adapter/PlatformServices.Desktop/Deployment/AssemblyLoadWorker.cs
+++ b/src/Adapter/PlatformServices.Desktop/Deployment/AssemblyLoadWorker.cs
@@ -46,11 +46,16 @@ namespace Microsoft.VisualStudio.TestPlatform.MSTestAdapter.PlatformServices.Dep
             Assembly assembly = null;
             try
             {
+                EqtTrace.Verbose($"AssemblyLoadWorker.GetFullPathToDependentAssemblies: Reflection loading {assemblyPath}.");
+
                 // First time we load in LoadFromContext to avoid issues.
                 assembly = this.assemblyUtility.ReflectionOnlyLoadFrom(assemblyPath);
             }
             catch (Exception ex)
             {
+                EqtTrace.Error($"AssemblyLoadWorker.GetFullPathToDependentAssemblies: Reflection loading of {assemblyPath} failed:");
+                EqtTrace.Error(ex);
+
                 warnings.Add(ex.Message);
                 return new string[0]; // Otherwise just return no dependencies.
             }
@@ -58,7 +63,7 @@ namespace Microsoft.VisualStudio.TestPlatform.MSTestAdapter.PlatformServices.Dep
             Debug.Assert(assembly != null, "assembly");
 
             List<string> result = new List<string>();
-            List<string> visitedAssemblies = new List<string>();
+            HashSet<string> visitedAssemblies = new HashSet<string>();
 
             visitedAssemblies.Add(assembly.FullName);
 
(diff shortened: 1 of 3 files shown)


src/Adapter/PlatformServices.Desktop/Deployment/AssemblyLoadWorker.cs
```diff
@@ -149,9 +154,11 @@ private string GetTargetFrameworkStringFromAssembly(Assembly assembly)
         /// <param name="result"> The result. </param>
         /// <param name="visitedAssemblies"> The visited Assemblies. </param>
         /// <param name="warnings"> The warnings. </param>
-        private void ProcessChildren(Assembly assembly, IList<string> result, IList<string> visitedAssemblies, IList<string> warnings)
+        private void ProcessChildren(Assembly assembly, IList<string> result, HashSet<string> visitedAssemblies, IList<string> warnings)
```
Sanan07:
```suggestion
        private void ProcessChildren(Assembly assembly, IList<string> result, ISet<string> visitedAssemblies, IList<string> warnings)
```

src/Adapter/PlatformServices.Desktop/Deployment/AssemblyLoadWorker.cs
```diff
@@ -219,20 +226,21 @@ private void ProcessChildren(Assembly assembly, IList<string> result, IList<stri
         /// <param name="visitedAssemblies"> The visited Assemblies. </param>
         /// <param name="warnings"> The warnings. </param>
         [System.Diagnostics.CodeAnalysis.SuppressMessage("Microsoft.Design", "CA1031:DoNotCatchGeneralExceptionTypes", Justification = "Requirement is to handle all kinds of user exceptions and message appropriately.")]
-        private void GetDependentAssembliesInternal(string assemblyString, IList<string> result, IList<string> visitedAssemblies, IList<string> warnings)
+        private void GetDependentAssembliesInternal(string assemblyString, IList<string> result, HashSet<string> visitedAssemblies, IList<string> warnings)
```
Sanan07:
```suggestion
        private void GetDependentAssembliesInternal(string assemblyString, IList<string> result, ISet<string> visitedAssemblies, IList<string> warnings)
```

src/Adapter/PlatformServices.Desktop/Utilities/DesktopAssemblyUtility.cs
```diff
@@ -227,7 +228,18 @@ internal virtual string[] GetFullPathToDependentAssemblies(string assemblyPath,
 
                     EqtTrace.InfoIf(EqtTrace.IsInfoEnabled, "AssemblyDependencyFinder.GetDependentAssemblies: loaded the worker.");
 
-                    return worker.GetFullPathToDependentAssemblies(assemblyPath, out warnings);
+                    var allDependencies = worker.GetFullPathToDependentAssemblies(assemblyPath, out warnings);
+                    var dependenciesFromDllDirectory = new List<string>();
+                    var dllDirectoryUppercase = dllDirectory.ToUpperInvariant();
+                    foreach (var dependency in allDependencies)
+                    {
+                        if (dependency.ToUpperInvariant().Contains(dllDirectoryUppercase))
```
Sanan07, line 236: Why to convert to UpperCase, if you can use something like this
`  dependency.Contains(dllDirectoryUppercase, StringComparison.OrdinalIgnoreCase);`

nohwnd:
@Sanan07 

There is DeploymentItem attribute, and this resolver is trying to figure out which assemblies from the bin folder should be copied into Out folder to reduce the amount of copying of unnecessary dependencies. 

Before .NET Core it was rare that a GAC assembly (GAC = Global Assembly Cache) would be distributed and acquired outside of the project. But in modern times you can get your dependencies from Nuget rather than from GAC when upgrading them. And then there is .NET Standard, which is a kind of meta assembly that does not pack any implementations, but rather just forwards to other dlls based on the current platform, this gives you unified API surface across platforms, with the flexibility of forwarding to the actual implementation as needed. 

The original algorithm searches dependencies of your test assembly recursively, and as soon as it finds a GAC assembly, it will stop searching in that branch of the tree. netstdard.dll is a GAC assembly, so the algo would terminate the search in that branch. BUT under netstandard there are assemblies that can be acquired as Nugets, e.g. System.ValueTuple.dll, which can be a newer version than the one linked by net standard. 

Say your project looks like this:

```
My.Tests.dll
|
---- netstandard.dll (isFromGAC = true)
|       |
|       ----- System.ValueTuple.dll 4.0.0 (IsFromGAC = true) -> System.ValueTuple.dll 4.0.3 (IsFromGAC = false)
|
---- Newtonsoft.Json.dll (IsFromGAC = false)
```

You have your My.Tests.dll test assembly, which depend on netstandard.dll. Netstandard normally depends on System.ValueTuple.dll from GAC, but because you installed it from Nuget, it will actually end up with System.Value.Tuple 4.0.3 in the bin folder, and a bindind redirect from 4.0.0 to 4.0.3.

When the dependency resolver algorithm kicks in, it will look at My.Tests.dll, will see that it depends on netstandard, and will see that it is from GAC and will search no furher. This will result in System.ValueTuple not being copied into the Out folder. 

Instead what needs to happen, is that we will look at the dependencies of the netstandard GAC assembly, and all of it's dependencies to see if any of them is coming from the bin folder and copy all of those over. 

Which is what this change does, instead of stopping at any GAC assembly, it resolves all dependencies of all assemblies, and returns all of them that are coming from the bin folder. This is because the bin folder is the king, and we are just trying to reduce what we copy, but never add more.

nohwnd:
The change to hashset is just because sets are better structure to use when you are repeatedly looking for "contains" because they are made for that purpose.

src/Adapter/PlatformServices.Desktop/Utilities/DesktopAssemblyUtility.cs
nohwnd: That would be nice, but this is netstandard1.0 code and that overload does not exist there. 

(Actually it is portable code that is equivalent to netstandard1.0, but the api problems are the same).

Sanan07 (approved):

Haplois (approved):

PR #968, opened 2021-09-06
Implement Class Cleanup Lifecycle selection

Haplois:
Originally posted by jhenkens on #724. Creating a new branch to be able to update it and merge.

> I'm soliciting early feedback on a branch to introduce a lifecycle policy for the `ClassCleanup` attribute. This allows a test author to set at the command line, assembly, or class level, whether a the `ClassCleanup` should happen at the end of a class, or at the end of the assembly.
> 
> I've lightly tested this, and it seems to work as expected in my tests, though I am not confident in its robustness in its current form. Specifically, my current implementation relying on a dictionary of the strings of the full test names seems fragile to me, though I was unable to determine how to instead track via the guids on the completion side of the dictionary.
> 
> If this feature seems generally appeciated, I'd love guidance from those more familar with the MSTest codebase in how I might improve my implementation.
> 
> A part of any final submission I will surely include reverting the many changed build files - those were needed to build in my environment, and are not to be present in any final reviews.
> 
> Thank you.


nohwnd:
I am missing some tests for the new behavior. There seem to be a lot of tests moved from one place to another. But are there any new tests to help us keep this in check?

src/TestFramework/MSTest.Core/Attributes/ClassCleanupLifecycle.cs
```diff
@@ -0,0 +1,21 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+namespace Microsoft.VisualStudio.TestTools.UnitTesting
+{
+    /// <summary>
+    /// When to run ClassCleanup during test execution
+    /// </summary>
+    public enum ClassCleanupLifecycle
```
nohwnd: Is this naming based on some existing settings we have? I think we should choose between Sequencing and Lifecycle if there is no existing example of this and in that case Lifecycle seems easier to grasp.

Or possibly ClassCleanupExecution?

nohwnd: image

Tried it out, this is how it looks like on the class. Which is imho bit ridiculous. Like I would have no idea what this means. :) 

These are the others that we have. 
TestDataSourceDiscoveryAttribute
ParallelizeAttribute
DoNotParallelizeAttribute
TestExtensionTypesAttribute
DiscoverInternalsAttribute

This also permits positioning it on method. Other assembly level attributes do not allow that, except for DoNotParallelize. And we have ClassCleanup, attribute, that already takes option for InheritanceBehavior , but not this one.

So imho we should: 
- [x] only allow this new attribute on assembly level
- [x] review the syntax
- [x] add option to ClassCleanup attribute

Not sure about this, but it seems to more or less fit both use cases:
```cs
[assembly: ClassCleanupExecution(ExecuteAt = ClassCleanupBehavior.EndOfAssembly)]
[ClassCleanup(InheritanceBehavior.BeforeEachDerivedClass, ClassCleanupBehavior.EndOfAssembly)]
[ClassCleanup(InheritanceBehavior.BeforeEachDerivedClass, ClassCleanupBehavior.EndOfClass)][ClassCleanup(ClassCleanupBehavior.EndOfAssembly)]
```

Not sure why, but our attributes don't expose properties, so hopefully we won't have more options, because that would lead to constructor boom.

nohwnd:
Imho we should also take into account the Standard output here, which we would normally not send because we don't have the test object anymore, and send it because we do have the test object.

src/TestFramework/MSTest.Core/Attributes/ClassCleanupLifecycle.cs
Haplois: Renamed the original attribute to `ClassCleanupExecution`, but I am hesitant to add this to `ClassCleanup` since it's method level it's hard see it.

Haplois: I turns out it was relatively easy. Done.

PR #1005, opened 2021-11-02
Add DoesNotReturnIf to Assert.IsTrue/Assert.IsFalse and DoesNotReturn to Assert.Fail

### LINKED ISSUE #1004: Additional Nullable Reference Type-aware support
## Description
Following https://github.com/microsoft/testfx/issues/714
and borrowing from https://devblogs.microsoft.com/dotnet/try-out-nullable-reference-types/#flow-attributes-doesnotreturn-and-doesnotreturnifbool
Can `DoesNotReturnIf` be added to `Assert.IsTrue` / `Assert.IsFalse` ? And `DoesNotReturn` be added to `Assert.Fail()` ? There are some other null checks should be able to be skipped, such as:

## Steps to reproduce
```c#
Dictionary<string, string> dictionary = ...
Assert.IsTrue(dictionary.TryGetValue(key, out var value));
Assert.AreEqual(5, value.Length); // should know that value is not null, but still warns currently
```
similarly:
```c#
Dictionary<string, string> dictionary = ...
var found = dictionary.TryGetValue(key, out var value);
if (!found) Assert.Fail();
Assert.AreEqual(5, value.Length); // should know that value is not null, but still warns currently
```

## Expected behavior
No compiler warnings.

## Actual behavior
Compiler warnings issued on `value.Length`.

## Environment
Using version `2.2.7` of MSTest.TestAdapater and MSTest.TestFramework

## Workaround

Similar shims can be written such as
```c#
[DoesNotReturn]
public static void AssertFail() => Assert.Fail();

public static void AssertIsTrue([DoesNotReturnIf(false)] bool condition) => Assert.IsTrue(condition);

public static void AssertIsFalse([DoesNotReturnIf(true)] bool condition) => Assert.IsFalse(condition);
```

dferretti:
Fixes #1004 

When adding these attributes, I did hesitate on the IsTrue/IsFalse overloads accepting nullable bools, because it does look odd to see
`[DoesNotReturnIf(true)] bool? condition` 
right next to a check like
`if (condition == true || condition == null)`
knowing that for both `true` and `null` the method will not return - but only for `true` will the nullability analysis kick in. Passing `null` will result in the call site still not knowing if the method returns or not so the user will still have to perform their own additional check.

But my thought process was that the attribute is still accurate and still helps. `[DoesNotReturnIf(true)]` means just that - the method does not return if the parameter exactly matches `true`. It does not say whether or not the method returns if `false` or `null` are passed in. 
Even though in this case we know that `IsFalse(null)` will not return, there is not a way to represent that with the current `DoesNotReturnIf` attribute.

### DIFF
--- a/src/TestFramework/MSTest.Core/Assertions/Assert.cs
+++ b/src/TestFramework/MSTest.Core/Assertions/Assert.cs
@@ -60,7 +60,7 @@ namespace Microsoft.VisualStudio.TestTools.UnitTesting
         /// <exception cref="AssertFailedException">
         /// Thrown if <paramref name="condition"/> is false.
         /// </exception>
-        public static void IsTrue(bool condition)
+        public static void IsTrue([DoesNotReturnIf(false)] bool condition)
         {
             IsTrue(condition, string.Empty, null);
         }
@@ -75,7 +75,7 @@ namespace Microsoft.VisualStudio.TestTools.UnitTesting
         /// <exception cref="AssertFailedException">
         /// Thrown if <paramref name="condition"/> is false.
         /// </exception>
-        public static void IsTrue(bool? condition)
+        public static void IsTrue([DoesNotReturnIf(false)] bool? condition)
         {
             IsTrue(condition, string.Empty, null);
         }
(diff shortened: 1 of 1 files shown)


dferretti:
cc @Haplois

Haplois (approved):

Haplois:
Thank you for the contribution @dferretti, merging!

Issue #927, opened 2021-07-19
Tests fail to run for UWP projects with Minimum Versions set to RS2 (15063) or earlier - Regressed with TestAdapter 2.2.1


MSLukeWest:
## Description
Attempting to run tests for a UWP unit test project with Minimum Version set to RS2 (15063) or earlier fails with an error about not being able to load System.Xml.ReaderWriter.  This regressed starting with MSTest.TestAdapter version 2.2.1.  It reproes both in and outside of Visual Studio.

## Steps to reproduce
1. Create a C# UWP test project.  This brings up a wizard where you select two values.  You can leave "Target Version" set to the default, but for "Minimum Version" change it to "Windows 10 Creators Update (10.0; Build 15063)".  Anything older will repro the issue as well.
2. Manage NuGet Packages for the project and update the version used for MSTest.TestAdapter to 2.1.2.
3. Build the project and run all tests, it should work fine.
4. Go back to Manage NuGet Packages and update the version used for MSTest.TestAdapter to 2.2.1 or anything newer.
5. Rebuild and run all tests again.

## Expected behavior
The test should pass.

## Actual behavior
The test fails to run, with the following error listed in the Test pane of the output window:

`An exception occurred while invoking executor 'executor://mstestadapter/v2': Could not load file or assembly 'System.Xml.ReaderWriter, Version=4.1.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a'. The located assembly's manifest definition does not match the assembly reference. (Exception from HRESULT: 0x80131040)`

## Environment
>  - Operating system: Win10, repro'd on various versions between RS5 and 21H1
>  - Build version of vstest.console: Repro'd with 16.10.0-release-20210520-03 and 16.11.0-release-20210626-04
>  - Package version of MSTest.TestAdapter: 2.2.1 - 2.2.5

MSLukeWest:
Debugging observations:

In the working case Framework assemblies (System.*) are being loaded from the CoreFramework app directory, for example System.Xml.ReaderWriter.dll is loaded from "C:\Program Files\WindowsApps\Microsoft.NET.CoreFramework.Debug.2.2_2.2.29301.2_x86__8wekyb3d8bbwe\System.Xml.ReaderWriter.dll".

[...]

PR #1008, opened 2021-11-12
Downgrade uwp

nohwnd:
Use testplatform that downgrades the dependencies for UWP so older version of UWP can run with the latest version of object model.

Pre-requisite for: https://github.com/microsoft/testfx/issues/927


eng/Versions.props
```diff
@@ -6,10 +6,10 @@
     <PreReleaseVersionLabel>beta</PreReleaseVersionLabel>
     <!-- Opt-out repo features -->
     <UsingToolXliff>false</UsingToolXliff>
-    <MicrosoftNETTestSdkVersion>17.0.0</MicrosoftNETTestSdkVersion>
+    <MicrosoftNETTestSdkVersion>17.1.0-preview-20211118-01</MicrosoftNETTestSdkVersion>
```
Haplois, line 9: This version will be overridden by Arcade,

Haplois (commented):
I think we should base this PR from "rel/2.2.7" and merge it on 2.2.8 if you need that version downgrade in TP?

eng/Versions.props
Haplois: See: https://github.com/microsoft/testfx/pull/1007/files#diff-1ea18ff65faa2ae6fed570b83747086d0317f5e4bc325064f6c14319a9c4ff67s :)

nohwnd: Better?

nohwnd: I edited the file manually, now it should give warning about downgrade if Maestro tries to patch over it.

PR #1014, opened 2021-11-29
Fix debug output clear

nohwnd:
Debug output clear had race condition that could cause the StringBuilder to throw, now we lock over both operations to prevent it.


Haplois (dismissed):

nohwnd:
The behavior is now what I would expect, and running on repro that crashed before no longer crashes, but tests are broken. I think there are just different assumptions about the behavior and how we receive data. There are also few tests that pass by accident, when dispose runs on the class at just the right moment to trigger Remove count in Dispose to be 2 instead of the more logical 1. Perfect local debugging experience because the test does not pass when you run it alone 🙄

Haplois (approved):

PR #1020, opened 2021-12-08
Fix EndOfClass cleanup for DataRow

nohwnd:
Fix cleanup for DataRow, to remove tests from the dictionary by using the same name (DisplayName) as what they were added. This affects DataRow tests and probably any test that uses DisplayName. Because then the entry in the hashmap and the key used to remove it (MethodName) are not the same.

Fixes https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1448552

- [x] Fix it 
- [ ] Add tests

### DIFF
--- a/src/Adapter/MSTest.CoreAdapter/Execution/UnitTestRunner.cs
+++ b/src/Adapter/MSTest.CoreAdapter/Execution/UnitTestRunner.cs
@@ -350,7 +350,7 @@ namespace Microsoft.VisualStudio.TestPlatform.MSTest.TestAdapter.Execution
                 this.remainingTestsByClass = testsToRun.GroupBy(t => t.TestMethod.FullClassName)
                     .ToDictionary(
                         g => g.Key,
-                        g => new HashSet<string>(g.Select(t => t.DisplayName)));
+                        g => new HashSet<string>(g.Select(t => t.TestMethod.Name)));
                 this.lifecycleFromMsTest = lifecycleFromMsTest;
                 this.lifecycleFromAssembly = lifecycleFromAssembly;
                 this.reflectHelper = reflectHelper ?? new ReflectHelper();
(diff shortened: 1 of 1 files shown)


nohwnd:
This seems to fix the issue, but I won't manage to add tests. Please add some @haplois. 

This was my repro, the Display name change in the attribute is not necessary.: 

```csproj
<!-- file TestProject3\TestProject3.csproj -->

  <PropertyGroup>
    <TargetFramework>netcoreapp3.1</TargetFramework>

  </PropertyGroup>

  <ItemGroup>
    <PackageReference Include="Microsoft.NET.Test.Sdk" Version="16.5.0" />
    <PackageReference Include="MSTest.TestAdapter" Version="99.99.99-dev" />
    <PackageReference Include="MSTest.TestFramework" Version="99.99.99-dev" />
	  <!--<PackageReference Include="MSTest.TestAdapter" Version="2.2.8" />
	  <PackageReference Include="MSTest.TestFramework" Version="2.2.8" />-->
    <PackageReference Include="coverlet.collector" Version="1.2.0" />
  </ItemGroup>

</Project>

```

```cs
// file TestProject3\UnitTest1.cs
using System;
using System.Diagnostics;

[assembly: ClassCleanupExecution(cleanupBehavior: ClassCleanupBehavior.EndOfClass)]

namespace TestProject3
{
    [TestClass]
    public class UnitTest1
    {
        static long Start;

        static UnitTest1()
        {
            Start = Stopwatch.GetTimestamp();
        }

        [ClassCleanup]
        public static void ClassCleanup() { 
            Console.WriteLine("cleanup");
        }

        [TestMethod(displayName: "TestMethod1")]
        [DataRow(1)]
        [DataRow(2)]
        public void TestMethod1(int _)
        {
            // the cleanup output should be on the test with the bigger number
            // because it needs to run on the last test
            Console.WriteLine(Stopwatch.GetTimestamp() - Start);
        }
    }

    [TestClass]
    public class UnitTest2
    {
        static long Start;
        static UnitTest2()
        {
            Start = Stopwatch.GetTimestamp();
        }


        [ClassCleanup]
        {
            Console.WriteLine("cleanup");
        }

        [TestMethod(displayName: "TestMethod1")]
        public void TestMethod1()
        {
            Console.WriteLine(Stopwatch.GetTimestamp() - Start);
        }

        [TestMethod(displayName: "TestMethod2")]
        public void TestMethod2()
        {
            Console.WriteLine(Stopwatch.GetTimestamp() - Start);
        }
    }
}

```

volleyms (commented):
Hi, I wonder why you would use the display name which could be any string, even as duplicate within a test class, instead of the unique method name... just my 2 cents...

Haplois (approved):
Tests to be added later on...

Issue #808, opened 2021-04-04
Add Assert.Equals & AssertNotEquals for decimals


HeroMaxPower:
## Description

I discovered that the assert.equals doesn't have a override for decimal and delta compare, I thought this could be a useful enhancement. More than happy to contribute.

nohwnd:
@HeroMaxPower you are more than welcome to add it :)

PR #1021, opened 2021-12-11
Add AreEqual/AreNotEqual overloads for decimal and long.

Exoow:
Long  overloads are needed, otherwise method calls with longs are ambiguous between the float/decimal parameter signatures.
#808


Haplois (approved):

Haplois:
@exoow Thank you for your contribution.

PR #1029, opened 2022-01-16
Remove nullability option from cleanupBehavior parameter

badcel:
Attribute enum parameters must be constant. The parameter "cleanupBehavior" can't be of type Nullable<ClassCleanupBehavior> as it is not possible to create a constant attribute parameter for it.


Haplois (approved):
Thank you for the contribution!

Issue #1053, opened 2022-03-14
tests fail intermittently when run in parallell


krijohan:
## Description

When running `dotnet test` in a project with `[assembly: Parallelize(Workers = 4, Scope = ExecutionScope.ClassLevel)]`, you will occasionally get unwarranted test failures from InvalidOperationException / "Collection was modified; enumeration operation may not execute". This makes the feature of running tests in parallell not very useful, unless you are willing to accept flaky tests.

While not clear from the stack trace in the resulting test failure, debugging indicates that the exception originates from https://github.com/microsoft/testfx/blob/rel/2.2.8/src/Adapter/MSTest.CoreAdapter/Execution/StackTraceHelper.cs#L239, where there is a `foreach` on a `static List<string>` member.

## Steps to reproduce
1. Clone https://github.com/krijohan/BugRepro1/ (2 test classes, with 1 always passing and 1 always skipped test in each).
2. Run `dotnet test` over and over again until it fails, e.g. by running the included script [repro.ps1](https://github.com/krijohan/BugRepro1/blob/main/repro.ps1) in powershell. On my laptop this takes about 5 minutes.

## Expected behavior
`dotnet test` never fails (as the included test cases are always passed or skipped)

## Actual behavior
Eventually, one of the skipped test cases will fail with

```
  Failed AlwaysSkip
  Error Message:
   Exception thrown while executing test. If using extension of TestMethodAttribute then please contact vendor. Error message: Collection was modified; enumeration operation may not execute., Stack trace:    at Microsoft.VisualStudio.TestPlatform.MSTest.TestAdapter.Execution.TestMethodInfo.ExecuteInternal(Object[] arguments)
   at Microsoft.VisualStudio.TestPlatform.MSTest.TestAdapter.Execution.TestMethodInfo.Invoke(Object[] arguments)
   at Microsoft.VisualStudio.TestTools.UnitTesting.TestMethodAttribute.Execute(ITestMethod testMethod)
   at Microsoft.VisualStudio.TestPlatform.MSTest.TestAdapter.Execution.TestMethodRunner.ExecuteTest(TestMethodInfo testMethodInfo)
```

[...]

PR #1055, opened 2022-03-17
Static init of StackTraceHelper.typesToBeExcluded

krijohan:
Address thread-safety issue https://github.com/microsoft/testfx/issues/1053 by initializing the static field StackTraceHelper.typesToBeExcluded statically instead of on demand.


Haplois:
Thank you for fix @krijohan, merged.

Haplois (approved):

PR #1071, opened 2022-04-14
Test execution bugs in specific TFMs addressed.

### LINKED ISSUE #750: .NET 5 TFM Support?
## Description
I'm trying to follow [these docs](https://docs.microsoft.com/en-us/windows/apps/desktop/modernize/desktop-to-uwp-enhance#net-5-preview-8-and-later-use-the-target-framework-moniker-option) about using the new .NET 5 TFMs for access to the Windows APIs (specifically [App Service](https://docs.microsoft.com/uwp/api/Windows.ApplicationModel.AppService.AppServiceConnection)).

The same setup works fine if I use the `netcoreapp3.1` TFM with the `PackageReference` (also described in that document):

```xml
<PackageReference Condition="'$(TargetFramework)' == 'netcoreapp3.1'"
                  Include="Microsoft.Windows.SDK.Contracts"
                  Version="10.0.19041.0" />
```

Trying to upgrade this to the recommended .NET 5 path doesn't seem to work with the MSTest project type in VS or on the command line.

I discovered this in my PR here: https://github.com/windows-toolkit/WindowsCommunityToolkit/pull/3580/files#diff-bcd2accc8655a3a705591da25274a0ebe6740eaf0dac109958fee2f4f553bf1eR6

And reverted back to using the .NET Core 3.1 approach.

## Steps to reproduce
Revert the commit here from this PR: https://github.com/windows-toolkit/WindowsCommunityToolkit/pull/3580/commits/3bf27d9740c588913d9ccc24155bbcc847d7663e

Otherwise, if you start a new MSTest project with .NET Core 3.1/PackageReference to include a Windows API call, try upgrading to .NET 5 TFM after...

## Expected behavior
Able to build MSTest project which references Windows API with .NET 5 Windows TFM.

## Actual behavior
Fails to build, Windows APIs can't be found.

## Environment
VS 2019 16.8.2
Azure Hosted environment and Windows 10 19042 (20H2).

FYI @andrewleader think this would be an issue for the Notifications tests too, eh?

### LINKED ISSUE #982: Tests are ignored for OS version-specific TFMs
## Steps to reproduce

In an MSTest project, set:

```<TargetFramework>net5.0-windows10.0.19041</TargetFramework>```

## Expected behavior

Tests should run

## Actual behavior

Tests do not run

Windows 10 Version	10.0.19043 Build 19043
>  - Build version of vstest.console
16.11.0
>  - Package version of MSTest framework and adapter

```    
    <PackageReference Include="Microsoft.NET.Test.Sdk" Version="16.11.0" />
    <PackageReference Include="MSTest.TestAdapter" Version="2.2.7" />
    <PackageReference Include="MSTest.TestFramework" Version="2.2.7" />
    <PackageReference Include="coverlet.collector" Version="3.1.0">
```

Please find a trivial test project that reproduces the issue attached
[TestProject1.zip](https://github.com/microsoft/testfx/files/7214925/TestProject1.zip)

### LINKED ISSUE #991: MSTest V2 tests won't run when targeting net5.0-windows10.0.19041
## Description
Tests will not run when targeting net5.0-windows10.0.19041 with MSTest version greater than 2.2.3. 

## Steps to reproduce
- Create MSTest project targeting net5.0-windows10.0.19041.
- Attempt to run tests
- Change MSTest version to 2.2.3
- Clean solution and run tests again

## Expected behavior
Tests should run in the test explorer

## Actual behavior
All tests are skipped. VS Test log incorrectly reports .dll does not exist.

VS Test Log Output:
```
[MSTest][Discovery][WinUI3-Demo\WinUI3-UnitTests\bin\Debug\net5.0-windows10.0.19041\WinUI3-UnitTests.dll] Failed to discover tests from assembly WinUI3-Demo\WinUI3-UnitTests\bin\Debug\net5.0-windows10.0.19041\WinUI3-UnitTests.dll. Reason:File does not exist: WinUI3-Demo\WinUI3-UnitTests\bin\Debug\net5.0-windows10.0.19041\WinUI3-UnitTests.dll
```

MSTest log:
```
pTrace Information: 0 : 4580, 4, 2021/10/01, 16:05:24.725, 12341249999384, testhost.x86.dll, AssemblyResolver.OnResolve: Resolved assembly: Microsoft.VisualStudio.TestPlatform.MSTestAdapter.PlatformServices, from path: \WinUI3-Demo\WinUI3-UnitTests\bin\Debug\net5.0-windows10.0.19041\Microsoft.VisualStudio.TestPlatform.MSTestAdapter.PlatformServices.dll
TpTrace Information: 0 : 4580, 4, 2021/10/01, 16:05:24.725, 12341250002103, testhost.x86.dll, MSTestExecutor.RunTests: Running tests from sources.
TpTrace Information: 0 : 4580, 4, 2021/10/01, 16:05:24.728, 12341250024421, testhost.x86.dll, CollectSourceInformation value Found : False 
TpTrace Error: 0 : 4580, 4, 2021/10/01, 16:05:24.773, 12341250480636, testhost.x86.dll, Exception of type 'System.InvalidOperationException' was thrown.
```

## Environment
    <PackageReference Include="Microsoft.NET.Test.Sdk" Version="16.11.0" />
    <PackageReference Include="MSTest.TestAdapter" Version="2.2.5" />
    <PackageReference Include="MSTest.TestFramework" Version="2.2.5" />

[AB#1415159](https://devdiv.visualstudio.com/0bdbc590-a062-4c3f-b0f6-9383f67865ee/_workitems/edit/1415159)

### PR DESCRIPTION
**Pull request by Haplois**
- Fixed test execution under net5+ TFMs.
 - Fixed UWP test execution.

 - Fixed #991
 - Fixed #982
 - Fixed #750

# Remarks
|Sample|Type|
|-------|-----|
|[PackagedAppTestSample.zip](https://github.com/microsoft/testfx/files/8526142/PackagedAppTestSample.zip)| Packaged Test |
|[UnpackagedAppTestSample.zip](https://github.com/microsoft/testfx/files/8526144/UnpackagedAppTestSample.zip)| Unpackaged Test |

From now on, WinUI packaged tests are required to specify `UseWinUI` property in `*.csproj` and set it true.

```xml
<PropertyGroup>
    <UseWinUI>true</UseWinUI>
</PropertyGroup>
```

## Packaged apps
[This](https://devblogs.microsoft.com/ifdef-windows/winui-desktop-unit-tests/) blog post form @azchohfi is still current, just make sure you're using WinUI 1.0.0 or later.

Also, `UseWinUI` property needs to be specified.

## Unpackaged apps

1. Create a packaged WinUI application.
2. Open `.csproj` file and add
```xml
    <ItemGroup>
        <ProjectCapability Include="TestContainer" />
    </ItemGroup>

    <PropertyGroup>
        <TestPlatformVersion>17.1.0-release-20220113-05</TestPlatformVersion>
        <MSTestVersion>5.0.30-dev</MSTestVersion>
        <GenerateProgramFile>false</GenerateProgramFile>
        <WindowsPackageType>None</WindowsPackageType>
    </PropertyGroup>

    <ItemGroup>
        <PackageReference Include="Microsoft.NET.Test.Sdk" Version="$(TestPlatformVersion)" />
        <PackageReference Include="MSTest.TestAdapter" Version="$(MSTestVersion)" />
        <PackageReference Include="MSTest.TestFramework" Version="$(MSTestVersion)" />
    </ItemGroup>
```
3. In `.csproj` file make sure you're referencing `1.0.1+` version of `Microsoft.WindowsAppSDK` package.
4. Add your test classes, if you need to run any tests in UI thread you need to specify this attribute in your test application
```csharp
using Microsoft.VisualStudio.TestTools.UnitTesting.AppContainer;

[assembly: WinUITestTarget(typeof(App4.App))]
```

## Building this branch to test
1. Open a terminal window at your source root
2. Run
```powershell
.\scripts\Build.ps1 -FrameworkVersion "5.0.30" -AdapterVersion "5.0.30" 

# If a preview version of VS fails to compile you can instead run:
# .\scripts\Build.ps1 -FrameworkVersion "5.0.30" -AdapterVersion "5.0.30" -DisallowPrereleaseMSBuild
```
3. You should increment the version to bypass package cache when testing - also add `./artifacts/Debug/MSTestPackages/` to your nuget.config file.


Scottj1s:
I'm concerned about false positives/negatives, as using WinUI is not equivalent to executing in a packaged app.

Haplois:
From now on, WinUI packaged tests are required to specify `UseWinUI` property in `*.csproj` and set it true.

```xml
<PropertyGroup>
    <UseWinUI>true</UseWinUI>
</PropertyGroup>
```

Haplois:
> @Scottj1s 
> I'm concerned about false positives/negatives, as using WinUI is not equivalent to executing in a packaged app.

Currently we only support packaged WinUI applications. I am experimenting with non packaged ones. I might amend this PR on weekend.

Scottj1s:
> > @Scottj1s
> > I'm concerned about false positives/negatives, as using WinUI is not equivalent to executing in a packaged app.
> 
> Currently we only support packaged WinUI applications. I am experimenting with non packaged ones. I might amend this PR to include support this weekend.

Sounds good.  Also note that in addition to packaged non-WinUI apps, you can have an unpackaged WinUI app.

Haplois:
Support for unpackaged-apps are also here. See PR description for details.

src/Adapter/PlatformServices.WinUI/PackageHelpers.cs
```diff
@@ -0,0 +1,50 @@
...
+
+#pragma warning disable SA1310 // Field names must not contain underscore
+
+namespace Microsoft.VisualStudio.TestPlatform.MSTestAdapter.PlatformServices
+{
+    using System;
+    using System.Runtime.InteropServices;
+    using System.Text;
+
+    // See https://github.com/microsoft/WindowsAppSDK/blob/f1cd5fc8ce6a6fed8442b3f64978b9e65b1d43fd/dev/Common/AppModel.Identity.h#L11
```
DrusTheAxe: What's this related to?

This should be line 24.5 as

```
/// @see [https](https://github.com/microsoft/WindowsAppSDK/blob/f1cd5fc8ce6a6fed8442b3f64978b9e65b1d43fd/dev/Common/AppModel.Identity.h#L11)
```

?

src/Adapter/PlatformServices.WinUI/PackageHelpers.cs
```diff
@@ -0,0 +1,50 @@
...
+
+    /// <summary>
+    /// Global helpers for queriying package information
+    /// </summary>
+    internal static class PackageHelpers
```
DrusTheAxe: 'Helpers' is a trigger word for me. Along with "util" and "temp" it's a poor term one should actively avoid.

Suggestion: `internal static class AppModel`

then callers are `if (AppModel.IsPackagedProcess())`

src/Adapter/PlatformServices.WinUI/PackageHelpers.cs
```diff
@@ -0,0 +1,50 @@
...
+        /// <returns>Returns true if application is packaged.</returns>
+        public static bool IsPackagedProcess()
+        {
+            // This API is supported in Windows 6.2+
+            // See https://docs.microsoft.com/en-us/windows/win32/api/appmodel/nf-appmodel-getcurrentpackagefullname#requirements
+            var os = (Environment.OSVersion.Version.Major * 1000) + Environment.OSVersion.Version.Minor;
+            if (os < 6002)
+            {
+                return false;
+            }
```
DrusTheAxe: You're already dynamically accessing it via P/Invoke. No need to try to probe product/version to determine feature; probe for the feature to use the feature.

In this case

```c#
try
{
   ...GetCurrentPackageFullName(...)...
}
catch (WhatExceptionDoesDotNetThrowWhenPInvokeDoesntFindTheTargetFunction e)
{
    return false;
}
```

src/Adapter/PlatformServices.WinUI/PackageHelpers.cs
```diff
@@ -0,0 +1,50 @@
...
+
+            StringBuilder packageFullName = new StringBuilder(0);
```
DrusTheAxe: If pacckageFullNameLength is 0 then packageFullName can (should) be NULL

This doesn't pass NULL to the native function, does it? `...GetCurrentPackageFullName(ref packageFullNameLength, null)` ?

src/Adapter/PlatformServices.WinUI/PackageHelpers.cs
```diff
@@ -0,0 +1,50 @@
...
+
+            /* Second call will return the package name.
+            packageFullName = new StringBuilder(packageFullNameLength);
+            result = GetCurrentPackageFullName(ref packageFullNameLength, packageFullName);
+            */
+
+            return result != APPMODEL_ERROR_NO_PACKAGE && packageFullNameLength != 0;
+        }
+
+        [DllImport("kernel32.dll", CharSet = CharSet.Unicode, SetLastError = true)]
```
DrusTheAxe: Why `SetLastError`? GetCurrentPackageFullName returns a Win32 error, it doesn't set LastError

src/Adapter/PlatformServices.WinUI/Services/WinUIFileOperations.cs
```diff
@@ -108,7 +109,14 @@ public void DisposeNavigationSession(object navigationSession)
         /// </returns>
         public string GetFullFilePath(string assemblyFileName)
         {
-            return (SafeInvoke<string>(() => Path.Combine(Windows.ApplicationModel.Package.Current.InstalledLocation.Path, assemblyFileName)) as string) ?? assemblyFileName;
+            string packagedPath()
+            {
+                return SafeInvoke<string>(() => Path.Combine(Windows.ApplicationModel.Package.Current.InstalledLocation.Path, assemblyFileName)) as string ?? assemblyFileName;
```
DrusTheAxe: `InstalledLocation` == `StorageFolder`. This is a painful way to get the path

You want to use `GetPackagePathByFullName2(...PackagePathType_Effective...)`

Win11 adds `InstalledPath` returning a `string` instead of `StorageFolder` - more efficient and reliable.

Better to call [GetPackagePath()](https://docs.microsoft.com/windows/win32/api/appmodel/nf-appmodel-getpackagepath), first appearing in Win8

But that's still flawed as it only gets the 'installed' location. It doesn't support Mutable or External Location. GetPackagePathByFullName2 solves that.

Pseudo-code you want:

```
try
{
    path = GetPackagePathByFullName2(...PackagePathtype_Effective...)
}
catch (FunctionNotFound e)
{
    try
    {
        path = GetPackagePath()
    }
    catch (FunctionNotFound e)
    {
        return assemblyFileName;
    }
}
```

src/Adapter/PlatformServices.WinUI/Services/WinUIFileOperations.cs
```diff
@@ -108,7 +109,14 @@ public void DisposeNavigationSession(object navigationSession)
...
+            }
+
+            return this.isPackaged
+                   ? packagedPath()
+                   : assemblyFileName;
```
DrusTheAxe: assemblyFileName == filename

but packaged path == path

odd to be mixing apples and oranges like that. That should be `Directory.Parent(assemblyName)` ?

src/Adapter/PlatformServices.WinUI/Services/WinUIFileOperations.cs
```diff
@@ -129,6 +137,25 @@ private static object SafeInvoke<T>(Func<T> action, string messageFormatOnExcept
...
         }
+
+        private bool DoesFileExistPackaged(string assemblyFileName)
+        {
+            var fileExists = false;
+
+            try
+            {
+                var fileNameWithoutPath = Path.GetFileName(assemblyFileName);
+                var searchTask = Windows.ApplicationModel.Package.Current.InstalledLocation.GetFileAsync(fileNameWithoutPath).AsTask();
```
DrusTheAxe: See previous comment re InstalledLocation vs path

DrusTheAxe: Why not `System.IO.File.Exists(assemblyFileName)` ?

src/Adapter/Build/Net5/MSTest.TestAdapter.props
```diff
@@ -0,0 +1,4 @@
+<?xml version="1.0" encoding="utf-8"?>
+<Project ToolsVersion="12.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
```
Evangelink: Might be good to add a comment about why we want to prevent using base props.

src/Adapter/Build/Net5/MSTest.TestAdapter.targets
```diff
@@ -28,7 +64,7 @@
       <MSTestV2Files Include="$(MSBuildThisFileDirectory)..\_common\*.dll" />
     </ItemGroup>
 	
-	<ItemGroup>
+	  <ItemGroup>
```
Evangelink: Nit:

```suggestion
    <ItemGroup>
```

src/Adapter/PlatformServices.WinUI/PackageHelpers.cs
Evangelink: You will want to also update the type xml doc comment.

src/Adapter/PlatformServices.WinUI/PackageHelpers.cs
```diff
@@ -0,0 +1,50 @@
...
+    /// Global helpers for queriying package information
+    /// </summary>
+    internal static class PackageHelpers
+    {
+        private const int APPMODEL_ERROR_NO_PACKAGE = 15700;
+
+        /// <summary>
+        /// Checks whether the current application is packaged or not.
+        /// </summary>
+        /// <returns>Returns true if application is packaged.</returns>
```
Evangelink: Nit
```suggestion
        /// <returns>Returns true if application is packaged; false otherwise.</returns>
```

src/TestFramework/Extension.WinUI/UITestMethodAttribute.cs
```diff
@@ -3,19 +3,48 @@
...
     {
+        private static Type applicationType;
+        private static bool isApplicationInitialized = false;
+        private static UI.Dispatching.DispatcherQueue applicationDispatcherQueue;
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="UITestMethodAttribute"/> class.
+        /// </summary>
+        /// <param name="displayName">
+        /// Display Name for the Test Window
```
Evangelink: Is it only used in the Test Window? Not in logs/CI etc?

```suggestion
        /// The display name used within the Test Window.
```

src/TestFramework/Extension.WinUI/UITestMethodAttribute.cs
```diff
@@ -81,5 +110,106 @@ public override TestResult[] Execute(ITestMethod testMethod)
...
+                return applicationType;
+            }
+
+            var attribute = assembly.GetCustomAttribute<WinUITestTargetAttribute>();
+            if (attribute == null || attribute.ApplicationType == null)
+            {
+                return null;
+            }
+
+            applicationType = attribute.ApplicationType;
```
Evangelink: This is un-needed as you are doing the same thing at the caller location.

src/TestFramework/Extension.WinUI/UITestMethodAttribute.cs
```diff
@@ -81,5 +110,106 @@ public override TestResult[] Execute(ITestMethod testMethod)
 
             return new TestResult[] { result };
         }
+
+        private static Type TryGetApplicationType(Assembly assembly)
```
Evangelink: Nit: Usually `TryXXX` is synonym of returning `bool` and having some `out` parameter, I would simply go with `GetApplicationType`.

src/TestFramework/Extension.WinUI/UITestMethodAttribute.cs
```diff
@@ -81,5 +110,106 @@ public override TestResult[] Execute(ITestMethod testMethod)
...
+                // We need to execute all module initializers before doing any WinRT calls.
+                // This will cause the [ModuleInitialzer]s to execute, if they haven't yet.
+                var id = applicationType.Assembly.GetType("Microsoft.WindowsAppSDK.Runtime.Identity");
+                if (id != null)
+                {
+                    _ = Activator.CreateInstance(id);
+                }
+            }
+            catch
+            {
```
Evangelink, line 155: Could we log somehow?

src/TestFramework/Extension.WinUI/UITestMethodAttribute.cs
```diff
@@ -81,5 +110,106 @@ public override TestResult[] Execute(ITestMethod testMethod)
...
+            if (attribute == null || attribute.ApplicationType == null)
+            {
+                return null;
+            }
+
+            applicationType = attribute.ApplicationType;
+            return applicationType;
+        }
+
+        private static UI.Dispatching.DispatcherQueue TryCreateApplicationDispatcherQueue(Assembly assembly)
```
Evangelink: Same nit remark about `TryXXX`.

src/TestFramework/Extension.WinUI/UITestMethodAttribute.cs
```diff
@@ -81,5 +110,106 @@ public override TestResult[] Execute(ITestMethod testMethod)
...
+        {
+            if (DispatcherQueue != null)
+            {
+                return DispatcherQueue;
+            }
+
+            if (TryCreateApplicationDispatcherQueue(assembly) is { } appQueue)
+            {
+                return appQueue;
+            }
```
Evangelink:
```suggestion
            if (TryCreateApplicationDispatcherQueue(assembly) is { } appDispatcher)
            {
                return appDispatcher;
            }
```

src/TestFramework/Extension.WinUI/UITestMethodAttribute.cs
```diff
@@ -81,5 +110,106 @@ public override TestResult[] Execute(ITestMethod testMethod)
...
+
+            try
+            {
+                if (Window.Current?.DispatcherQueue is { } queue)
+                {
+                    return queue;
+                }
```
Evangelink:
```suggestion
                if (Window.Current?.DispatcherQueue is { } windowDispatcher)
                {
                    return windowDispatcher;
                }
```

src/TestFramework/Extension.WinUI/UITestMethodAttribute.cs
```diff
@@ -81,5 +110,106 @@ public override TestResult[] Execute(ITestMethod testMethod)
...
+            }
+
+            var tsc = new TaskCompletionSource<UI.Dispatching.DispatcherQueue>();
+            var uiMagicThread = new Thread(new ThreadStart(() =>
+            {
+                Application.Start(p =>
+                {
+                    // TODO: @haplois before merging this PR
+                    // ADD ERROR HANDLING HERE
+                    isApplicationInitialized = true;
```
Evangelink: I would move this to later in the flow to ensure that if we are failing we don't consider it as initialized.

src/TestFramework/Extension.WinUI/UITestMethodAttribute.cs
```diff
@@ -3,19 +3,48 @@
...
+    using Microsoft.UI.Xaml;
 
     /// <summary>
     /// Execute test code in UI thread for Windows store apps.
     /// </summary>
     public class UITestMethodAttribute : TestMethodAttribute
     {
+        private static Type applicationType;
+        private static bool isApplicationInitialized = false;
+        private static UI.Dispatching.DispatcherQueue applicationDispatcherQueue;
```
Evangelink, line 21: Looking back at the code, I think this field might not be needed. We could set the property instead.

src/TestFramework/Extension.WinUI/UITestMethodAttribute.cs
```diff
@@ -81,5 +110,106 @@ public override TestResult[] Execute(ITestMethod testMethod)
...
+            {
+                return applicationDispatcherQueue;
+            }
+
+            if (isApplicationInitialized)
+            {
+                return null;
+            }
+
+            applicationType = TryGetApplicationType(assembly);
```
Evangelink: I don't think we need this to be a field because the fact that we cache the `applicationDispatcherQueue` gives us guarantee we won't call this multiple times.

src/TestFramework/Extension.WinUI/WinUITestTargetAttribute.cs
```diff
@@ -0,0 +1,42 @@
...
+    {
+        private readonly Type applicationType;
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="WinUITestTargetAttribute"/> class.
+        /// </summary>
+        /// <param name="application">
+        /// Specifies <see cref="Microsoft.UI.Xaml.Application" /> derived class to run UI tests on.
+        /// </param>
+        public WinUITestTargetAttribute(Type application)
```
Evangelink: Nit: Rename `application` into `applicationType`.

src/TestFramework/Extension.WinUI/WinUITestTargetAttribute.cs
```diff
@@ -0,0 +1,42 @@
...
+        {
+            if (application == null)
+            {
+                throw new ArgumentNullException(nameof(application));
+            }
+
+            if (!typeof(Microsoft.UI.Xaml.Application).IsAssignableFrom(application))
+            {
+                throw new ArgumentException(string.Format(FrameworkMessages.ArgumentXMustDeriveFromClassY, nameof(application), "Microsoft.UI.Xaml.Application"), nameof(application));
```
Evangelink:
```suggestion
                throw new ArgumentException(string.Format(CultureInfo.InvariantCulture, FrameworkMessages.ArgumentXMustDeriveFromClassY, nameof(application), "Microsoft.UI.Xaml.Application"), nameof(application));
```

src/TestFramework/Extension.WinUI/WinUITestTargetAttribute.cs
```diff
@@ -0,0 +1,42 @@
...
+            }
+
+            this.applicationType = application;
+        }
+
+        /// <summary>
+        /// Gets the <see cref="Microsoft.UI.Xaml.Application" /> class.
+        /// </summary>
+        public Type ApplicationType => this.applicationType;
```
Evangelink: I would remove the field and transform this property in readonly property.

```suggestion
        public Type ApplicationType { get; }
```

src/TestFramework/MSTest.Core/Resources/FrameworkMessages.resx
```diff
@@ -292,4 +292,11 @@ Stack Trace: {4}</value>
   <data name="DynamicDataIEnumerableEmpty" xml:space="preserve">
     <value>Property or method {0} on {1} returns empty IEnumerable&lt;object[]&gt;.</value>
   </data>
+  <data name="ArgumentXMustDeriveFromClassY" xml:space="preserve">
+    <value>{0} must derive from {1}!</value>
```
Evangelink: What about

```suggestion
    <value>Type '{0}' is not assignable to '{1}'.</value>
```

Evangelink: And maybe I would make the resource entry less generic and directly hardcode the `{1}` (and rename the resource entry into `TypeIsNotAssignableToApplication`).

src/Adapter/PlatformServices.WinUI/Services/WinUIFileOperations.cs
Haplois: I'll double check this, but in unpackaged-apps and fullfw `assemblyFileName` is a full path. In UWP and WinUI it's a filename.

Evangelink (commented):
Could we add some integration test (probably the test case we used for manual testing).

src/TestFramework/Extension.WinUI/UITestMethodAttribute.cs
Haplois: It's it's used everywhere TestName is used- we just copied it from somewhere else.

src/TestFramework/Extension.WinUI/UITestMethodAttribute.cs
Haplois: I don't like how it looks - without that. Introduced a local instead?

src/TestFramework/Extension.WinUI/UITestMethodAttribute.cs
Haplois: We don't have easy access to logging here, and I don't think it's required here.

src/TestFramework/Extension.WinUI/UITestMethodAttribute.cs
Haplois: If Application.Start succeeds, application is initialized. If we fail - we just don't have the dispatcher - nothing to do. A new call to Application.Start will bring down the process.

src/TestFramework/Extension.WinUI/UITestMethodAttribute.cs
Haplois: which field?

src/Adapter/Build/Net5/MSTest.TestAdapter.props
Haplois: What do you mean?

src/Adapter/PlatformServices.WinUI/AppModel.cs
```diff
@@ -0,0 +1,125 @@
...
+#pragma warning disable SA1310 // Field names must not contain underscore
+
+namespace Microsoft.VisualStudio.TestPlatform.MSTestAdapter.PlatformServices.AppContainer
+{
+    using System;
+    using System.Runtime.InteropServices;
+    using System.Text;
+
+    /// <summary>
+    /// Global helpers for queriying package information
```
Evangelink:
```suggestion
    /// Provides package information about the application.
```

src/Adapter/PlatformServices.WinUI/AppModel.cs
```diff
@@ -0,0 +1,125 @@
...
+                var packageFullName = new StringBuilder(packageFullNameLength);
+                result = GetCurrentPackageFullName(ref packageFullNameLength, packageFullName);
+                */
+            }
+            catch (EntryPointNotFoundException)
+            {
+                return false;
+            }
+
+            return result != Interop.APPMODEL_ERROR_NO_PACKAGE && packageFullNameLength != 0;
```
Evangelink, line 41: I am not sure to understand why they also check `ERROR_INSUFFICIENT_BUFFER`, but is it something we should care about here?

src/TestFramework/Extension.WinUI/UITestMethodAttribute.cs
Evangelink: I meant `applicationDispatcherQueue`. We could directly set `DispatcherQueue` property instead as we would only consider the app one if that's not set by user (so no chance of overriding their input).

src/TestFramework/Extension.WinUI/UITestMethodAttribute.cs
Evangelink: I see.

src/TestFramework/Extension.WinUI/UITestMethodAttribute.cs
```diff
@@ -81,5 +109,100 @@ public override TestResult[] Execute(ITestMethod testMethod)
...
+            catch
+            {
+            }
+
+            var tsc = new TaskCompletionSource<UI.Dispatching.DispatcherQueue>();
+            var uiMagicThread = new Thread(new ThreadStart(() =>
+            {
+                Application.Start(p =>
+                {
+                    // TODO: @haplois before merging this PR
```
Evangelink: Don't forget this comment.

src/Adapter/PlatformServices.WinUI/AppModel.cs
Haplois: I don't know the API, but I assume if the package name is null - it's not a package :)

src/TestFramework/Extension.Shared/TestContext.cs
```diff
@@ -57,6 +57,14 @@ public abstract class TestContext
         /// </summary>
         public virtual UnitTestOutcome CurrentTestOutcome => UnitTestOutcome.Unknown;
 
+        /// <summary>
+        /// Adds a file name to the list in TestResult.ResultFileNames
```
Evangelink:
```suggestion
        /// Adds a file name to the list in <see cref="TestResult.ResultFileNames" />.
```

src/TestFramework/Extension.Shared/TestContext.cs
```diff
@@ -57,6 +57,14 @@ public abstract class TestContext
...
+        /// </summary>
+        /// <param name="fileName">
+        /// The file Name.
```
Evangelink:
```suggestion
        /// The file name.
```

src/TestFramework/Extension.WinUI/UITestMethodAttribute.cs
```diff
@@ -81,5 +110,108 @@ public override TestResult[] Execute(ITestMethod testMethod)
...
+                }
+            }
+            catch
+            {
+            }
+
+            return null;
+        }
+
+        private static DispatcherQueue InitializeApplication(Type applicationType)
```
Evangelink, line 187: Just wondering if this should be a `void` method simply doing the app initialization and the caller would then return `applicationDispatcherQueue`.

src/TestFramework/MSTest.Core/Attributes/TestMethod/TestMethodAttribute.cs
```diff
@@ -23,17 +23,17 @@ public TestMethodAttribute()
         /// Initializes a new instance of the <see cref="TestMethodAttribute"/> class.
         /// </summary>
         /// <param name="displayName">
-        /// Display Name for the Test Window
+        /// Display Name for the test
```
Evangelink:
```suggestion
        /// Display name for the test.        
```

src/TestFramework/MSTest.Core/Attributes/TestMethod/TestMethodAttribute.cs
```diff
@@ -23,17 +23,17 @@ public TestMethodAttribute()
...
         /// </param>
         public TestMethodAttribute(string displayName)
         {
             this.DisplayName = displayName;
         }
 
         /// <summary>
-        /// Gets display Name for the Test Window
+        /// Gets display Name for the test
```
Evangelink:
```suggestion
        /// Gets display name for the test.
```

Evangelink (approved):
A few nit comments - feel free to ignore them.

Evangelink (approved):

Haplois:
It also supports unpackaged apps, you can find a sample [here](https://github.com/microsoft/testfx/files/8526144/UnpackagedAppTestSample.zip). I'll also update the documentation.

PR #1076, opened 2022-04-22
Added MSTest meta-package

Haplois:
Our consumers can simplify their test project by depending on `MSTest` package instead of `MSTest.TestAdapter`, `MSTest.TestFramework` and `Microsoft.NET.Test.Sdk` packages. 

This meta package will guarantee all the parts are compatible.

[This](https://github.com/microsoft/testfx/files/8540471/MSTestProjects.zip) solution can be used to verify.

### DIFF
--- a/src/Package/MSTest.nuspec
+++ b/src/Package/MSTest.nuspec
@@ -1,0 +1,68 @@
+<?xml version="1.0"?>
+<package >
+  <metadata>
+    <id>MSTest</id>
+    <version>1.1.17</version>
+    <title>MSTest</title>
+    <requireLicenseAcceptance>true</requireLicenseAcceptance>
+    <authors>Microsoft</authors>
+    <owners>Microsoft</owners>
+    <description>
+      A meta package to simplify test projects.
+
+      Supported platforms:
+      - .NET 4.5.0+
+      - .NET Core 1.0+ (Universal Windows Apps 10+) (Visual Studio 2017)
+      - .NET 5.0 (Visual Studio 2019)
+      - .NET 5.0 Windows.18362+ (WinUI Desktop Apps) (Visual Studio 2019)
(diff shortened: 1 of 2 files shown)


src/Package/MSTest.nuspec
```diff
@@ -0,0 +1,68 @@
+<?xml version="1.0"?>
+<package >
+  <metadata>
+    <id>MSTest</id>
+    <version>1.1.17</version>
```
Evangelink, line 5: I see it is defined as the other files but shouldn't this be `2.2.10`? If the version is always replaced during build, I would vote for (ok to be a follow up PR) making it more explicit: blank, `42.42.42`, `version-set-by-build`...

src/Package/MSTest.nuspec
```diff
@@ -0,0 +1,68 @@
...
+
+      <group targetFramework="net5.0">
+        <dependency id="MSTest.TestAdapter" version="[$version$]" />
+        <dependency id="MSTest.TestFramework" version="[$version$]" />
+        <dependency id="Microsoft.NET.Test.Sdk" version="[$TestPlatformVersion$]" />
+      </group>
+    </dependencies>
+  </metadata>
+  <files>
+    <file src="Icon.png" target="" />
```
Evangelink, line 66: We should ask for some icon to be designed for mstest.

src/Package/MSTest.nuspec
Haplois: Version is always replaced during the build. I agree with you, but let’s do that in a separate Pr?

Evangelink (approved):

cremor:
I just found this in the release notes of MSTest 2.2.10. But the package https://www.nuget.org/packages/MSTest/ was unlisted. So I assume this means it should not be used yet?

PR #1084, opened 2022-04-26
Added more fail paths for data serialization.

Haplois:
IDataSource tests don't support duplicate test names, or unserializable objects. Added bail out paths and warnings for those scnerios.

### DIFF
--- a/src/Adapter/MSTest.CoreAdapter/Discovery/AssemblyEnumerator.cs
+++ b/src/Adapter/MSTest.CoreAdapter/Discovery/AssemblyEnumerator.cs
@@ -357,13 +357,27 @@ namespace Microsoft.VisualStudio.TestPlatform.MSTest.TestAdapter.Discovery
             foreach (var dataSource in testDataSources)
             {
                 var data = dataSource.GetData(methodInfo);
-                var discoveredTests = new List<UnitTestElement>();
+                var discoveredTests = new Dictionary<string, UnitTestElement>();
+                var discoveredIndex = new Dictionary<string, int>();
                 var serializationFailed = false;
+                var index = 0;
 
                 foreach (var d in data)
                 {
                     var discoveredTest = test.Clone();
-                    discoveredTest.DisplayName = dataSource.GetDisplayName(methodInfo, d);
+                    discoveredTest.DisplayName = dataSource.GetDisplayName(methodInfo, d) ?? discoveredTest.DisplayName;
+
+                    // if we have a duplicate test name don't expand the test, bail out.
+                    if (discoveredTests.ContainsKey(discoveredTest.DisplayName))
+                    {
+                        var firstSeen = discoveredIndex[discoveredTest.DisplayName];
+                        var warning = string.Format(CultureInfo.CurrentCulture, Resource.CannotExpandIDataSourceAttribute_DuplicateDisplayName, firstSeen, index, discoveredTest.DisplayName);
+                        warning = string.Format(CultureInfo.CurrentUICulture, Resource.CannotExpandIDataSourceAttribute, test.TestMethod.ManagedTypeName, test.TestMethod.ManagedMethodName, warning);
+                        PlatformServiceProvider.Instance.AdapterTraceLogger.LogWarning($"DynamicDataEnumarator: {warning}");
+
+                        serializationFailed = true;
+                        break;
+                    }
 
                     try
                     {
@@ -372,22 +386,27 @@ namespace Microsoft.VisualStudio.TestPlatform.MSTest.TestAdapter.Discovery
                     }
                     catch (SerializationException)
                     {
+                        var firstSeen = discoveredIndex[discoveredTest.DisplayName];
+                        var warning = string.Format(CultureInfo.CurrentCulture, Resource.CannotExpandIDataSourceAttribute_CannotSerialize, index, discoveredTest.DisplayName);
+                        warning = string.Format(CultureInfo.CurrentUICulture, Resource.CannotExpandIDataSourceAttribute, test.TestMethod.ManagedTypeName, test.TestMethod.ManagedMethodName, warning);
+                        PlatformServiceProvider.Instance.AdapterTraceLogger.LogWarning($"DynamicDataEnumarator: {warning}");
+
                         serializationFailed = true;
                         break;
                     }
 
-                    discoveredTests.Add(discoveredTest);
+                    discoveredTests[discoveredTest.DisplayName] = discoveredTest;
+                    discoveredIndex[discoveredTest.DisplayName] = index++;
                 }
 
                 // Serialization failed for the type, bail out.
                 if (serializationFailed)
                 {
                     tests.Add(test);
-
                     break;
                 }
 
-                tests.AddRange(discoveredTests);
+                tests.AddRange(discoveredTests.Values);
             }
 
             return true;
--- a/src/Adapter/MSTest.CoreAdapter/Resources/Resource.resx
+++ b/src/Adapter/MSTest.CoreAdapter/Resources/Resource.resx
@@ -368,4 +368,20 @@ Error: {1}</value>
   <data name="OlderTFMVersionFoundClassCleanup" xml:space="preserve">
     <value>An older version of MSTestV2 package is loaded in assembly, test cleanup methods might not run as expected. Please make sure all your test projects references MSTest pacakges newer then version 2.2.8.</value>
   </data>
+  <data name="CannotExpandIDataSourceAttribute" xml:space="preserve">
+    <value>Exception occurred while expanding IDataSource rows from attribute on "{0}.{1}": {2}</value>
+    <comment>{0}: TypeName with namespace, 
+{1}: Method name,
+{2}: CannotExpandIDataSourceAttribute_DuplicateDisplayName or CannotExpandIDataSourceAttribute_CannotSerialize</comment>
+  </data>
+  <data name="CannotExpandIDataSourceAttribute_DuplicateDisplayName" xml:space="preserve">
+    <value>Display name "{2}" on indexes {0} and {1} are duplicate. Display names should be unique.</value>
+    <comment>{0}, {1}: Zero based index if an element inside of an array
+{2}: Test display name.</comment>
+  </data>
+  <data name="CannotExpandIDataSourceAttribute_CannotSerialize" xml:space="preserve">
+    <value>Data on index {0} for "{1}" cannot be serialized. All data provided through "IDataSource" should be serializable. If you need to test non-serializable data sources, please make sure you add "TestDataSourceDiscovery" attribute on your test assembly and set the discovery option to "DuringExecution".</value>
+    <comment>{0}: Zero based index if an element inside of an array,
+{1}: Test name</comment>
+  </data>
 </root>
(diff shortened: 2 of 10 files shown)


src/Adapter/MSTest.CoreAdapter/Resources/Resource.resx
```diff
@@ -368,4 +368,19 @@ Error: {1}</value>
   <data name="OlderTFMVersionFoundClassCleanup" xml:space="preserve">
     <value>An older version of MSTestV2 package is loaded in assembly, test cleanup methods might not run as expected. Please make sure all your test projects references MSTest pacakges newer then version 2.2.8.</value>
   </data>
+  <data name="CannotExpandIDataSourceAttribute" xml:space="preserve">
+    <value>Exception occurred while expanding IDataSource rows from attribute on "{0}.{1}": {2}</value>
+    <comment>{0}: TypeName with namespace, 
+{1}: Method name,</comment>
```
Evangelink: Missing info about `{2}`.

src/Adapter/MSTest.CoreAdapter/Discovery/AssemblyEnumerator.cs
```diff
@@ -357,37 +357,56 @@ private bool ProcessTestDataSourceTests(UnitTestElement test, MethodInfo methodI
...
                     var discoveredTest = test.Clone();
                     discoveredTest.DisplayName = dataSource.GetDisplayName(methodInfo, d);
 
+                    // if we have a duplicate test name don't expand the test, bail out.
+                    if (discoveredTests.ContainsKey(discoveredTest.DisplayName))
+                    {
+                        var firstSeen = discoveredIndex[discoveredTest.DisplayName];
+                        var warning = string.Format(CultureInfo.CurrentCulture, Resource.CannotExpandIDataSourceAttribute_DuplicateDisplayName, firstSeen, index, discoveredTest.DisplayName);
+                        warning = string.Format(CultureInfo.CurrentUICulture, Resource.CannotExpandIDataSourceAttribute, test.TestMethod.ManagedTypeName, test.TestMethod.ManagedMethodName, warning);
+                        PlatformServiceProvider.Instance.AdapterTraceLogger.LogWarning($"DynamicDataEnumarator: {warning}");
```
Evangelink, line 376: It's weird to have the message translated but not this prefix part, I would like to have everything translated.

src/Adapter/MSTest.CoreAdapter/Discovery/AssemblyEnumerator.cs
```diff
@@ -357,37 +357,56 @@ private bool ProcessTestDataSourceTests(UnitTestElement test, MethodInfo methodI
...
                     {
                         discoveredTest.TestMethod.SerializedData = DataSerializationHelper.Serialize(d);
                         discoveredTest.TestMethod.DataType = DynamicDataType.ITestDataSource;
                     }
                     catch (SerializationException)
                     {
+                        var firstSeen = discoveredIndex[discoveredTest.DisplayName];
+                        var warning = string.Format(CultureInfo.CurrentCulture, Resource.CannotExpandIDataSourceAttribute_CannotSerialize, index, discoveredTest.DisplayName);
+                        warning = string.Format(CultureInfo.CurrentUICulture, Resource.CannotExpandIDataSourceAttribute, test.TestMethod.ManagedTypeName, test.TestMethod.ManagedMethodName, warning);
+                        PlatformServiceProvider.Instance.AdapterTraceLogger.LogWarning($"DynamicDataEnumarator: {warning}");
```
Evangelink, line 392: Same remark about translated text.

src/Adapter/MSTest.CoreAdapter/Resources/Resource.resx
```diff
@@ -368,4 +368,19 @@ Error: {1}</value>
...
+  </data>
+  <data name="CannotExpandIDataSourceAttribute_DuplicateDisplayName" xml:space="preserve">
+    <value>Display name "{2}" on indexes {0} and {1} are duplicate.</value>
+    <comment>{0}, {1}: Zero based index if an element inside of an array
+{2}: Display name of a test.</comment>
```
Evangelink:
```suggestion
    <comment>{0}: Zero-based index of the first data row with the given display name
{1}: Zero-based index of the duplicate data row display name
{2}: Test display name.</comment>
```

src/Adapter/MSTest.CoreAdapter/Resources/Resource.resx
```diff
@@ -368,4 +368,19 @@ Error: {1}</value>
   <data name="OlderTFMVersionFoundClassCleanup" xml:space="preserve">
     <value>An older version of MSTestV2 package is loaded in assembly, test cleanup methods might not run as expected. Please make sure all your test projects references MSTest pacakges newer then version 2.2.8.</value>
   </data>
+  <data name="CannotExpandIDataSourceAttribute" xml:space="preserve">
+    <value>Exception occurred while expanding IDataSource rows from attribute on "{0}.{1}": {2}</value>
+    <comment>{0}: TypeName with namespace, 
+{1}: Method name,</comment>
+  </data>
+  <data name="CannotExpandIDataSourceAttribute_DuplicateDisplayName" xml:space="preserve">
+    <value>Display name "{2}" on indexes {0} and {1} are duplicate.</value>
```
Evangelink: Duplicated display name "{2}" on indexes {0} and {1}. Display name should be unique.

src/Adapter/MSTest.CoreAdapter/Resources/Resource.resx
```diff
@@ -368,4 +368,19 @@ Error: {1}</value>
...
+    <comment>{0}, {1}: Zero based index if an element inside of an array
+{2}: Display name of a test.</comment>
+  </data>
+  <data name="CannotExpandIDataSourceAttribute_CannotSerialize" xml:space="preserve">
+    <value>Data on index {0} for "{1}" cannot be serialized. All data provided to thru a IDataSource should be serializable, shouldn't lose it's state when deserialized. If you need to test non-serializable data sources, please make sure you define `TestDataSourceDiscovery` attribute on your test assembly and set discovery option to `DuringExecution`.</value>
```
Evangelink:
```suggestion
    <value>Test "{1}" with data row index {0} cannot be serialized. All data provided through "IDataSource" should be serializable. If you need to test non-serializable data sources, please make sure you add "TestDataSourceDiscovery" attribute on your test assembly and set the discovery option to "DuringExecution".</value>
```

src/Adapter/MSTest.CoreAdapter/Resources/Resource.resx
```diff
@@ -368,4 +368,19 @@ Error: {1}</value>
   <data name="OlderTFMVersionFoundClassCleanup" xml:space="preserve">
     <value>An older version of MSTestV2 package is loaded in assembly, test cleanup methods might not run as expected. Please make sure all your test projects references MSTest pacakges newer then version 2.2.8.</value>
   </data>
+  <data name="CannotExpandIDataSourceAttribute" xml:space="preserve">
+    <value>Exception occurred while expanding IDataSource rows from attribute on "{0}.{1}": {2}</value>
```
Evangelink, line 372:
```suggestion
    <value>Exception occurred while expanding "IDataSource" rows from attribute on "{0}.{1}". Error: {2}</value>
```

Evangelink: But maybe I would remove the `Error: {2}` and print 2 warnings or add the relevant test info (i.e. `{0}.{1}`) inside the other 2 messages.

Evangelink (commented):
Could we add tests for these new conditions?

src/Adapter/MSTest.CoreAdapter/Discovery/AssemblyEnumerator.cs
Haplois: Prefix was actually method name one, it's still something like that. (Consistent with other warnings in TestFx)

src/Adapter/MSTest.CoreAdapter/Discovery/AssemblyEnumerator.cs
Haplois: See the prev comment.

src/Adapter/MSTest.CoreAdapter/Resources/Resource.resx
Haplois: It's not an error, it's the continuation of the message. There will be more edge cases here, so it's easier on localization to do this way. A standard part, and a detail part.

src/Adapter/MSTest.CoreAdapter/Discovery/AssemblyEnumerator.cs
Evangelink: Ok then let's ignore that but that's something we should probably address in the future, that's weird to mix technical (trace log) and functional to me.

src/Adapter/MSTest.CoreAdapter/Resources/Resource.resx
Evangelink: Yeah error was probably not the best wording. Let's continue as-is but IMO the full message reads weird.

src/Adapter/MSTest.CoreAdapter/Resources/Resource.resx
Haplois: Comments are not that important, it's to help loc team.

src/Adapter/MSTest.CoreAdapter/Resources/Resource.resx
Haplois: It's the data which we can't serialize, not the test, and it's not always a DataRow. That was the reason I didn't use the word `row`.

src/Adapter/MSTest.CoreAdapter/Resources/Resource.resx
```diff
@@ -368,4 +368,20 @@ Error: {1}</value>
...
+{1}: Method name,
+{2}: CannotExpandIDataSourceAttribute_DuplicateDisplayName or CannotExpandIDataSourceAttribute_CannotSerialize</comment>
+  </data>
+  <data name="CannotExpandIDataSourceAttribute_DuplicateDisplayName" xml:space="preserve">
+    <value>Display name "{2}" on indexes {0} and {1} are duplicate. Display names should be unique.</value>
+    <comment>{0}, {1}: Zero based index if an element inside of an array
+{2}: Test display name.</comment>
+  </data>
+  <data name="CannotExpandIDataSourceAttribute_CannotSerialize" xml:space="preserve">
+    <value>Data on index {0} for "{1}" cannot be serialized. All data provided to through "IDataSource" should be serializable, shouldn't lose it's state when deserialized. If you need to test non-serializable data sources, please make sure you add "TestDataSourceDiscovery" attribute on your test assembly and set the discovery option to "DuringExecution".</value>
```
Evangelink: Either `provided to "IDataSource"` or `provided through "IDataSource"`. Also `should not lose its state` although as on my suggestion I would remove this part because that's implied by being serializable.

Haplois: I removed that part.

Evangelink (approved):

PR #1089, opened 2022-04-26
Remove unused classes

Evangelink:


Haplois (commented):
Are you sure this classes are not used? How did you check it

Evangelink:
I detected them when doing the sdk project style change, they were not referenced by the old project and the sdk style picked them up and it was causing conflicts.

Haplois (approved):

PR #1119, opened 2022-06-01
TestResults folder names are now cross platform compatible, as per #678

### LINKED ISSUE #678: TestResults folder names are not cross platform compatible
## Description

Hi, i'm from the Docker team. We are getting user issues creating test results folders on Linux which are then synchronised with the host (Windows) filesystem.

Using `dotnet test` with the default out of the box MSTest configuration creates TestResults folders which are not cross platform compatible due to using invalid characters in the names, specifically the colon character `:`, for example: `TestResults/Deploy_ 2020-01-22 14:14:56`

I believe the offending code is here https://github.com/microsoft/testfx/blob/master/src/Adapter/PlatformServices.NetCore/Utilities/NetCoreDeploymentUtility.cs#L51 `string dateTimeSufix = DateTime.Now.ToString("yyyy-MM-dd HH:mm:ss", DateTimeFormatInfo.InvariantInfo);`

Despite the code running this result through `ReplaceInvalidFileNameCharacters` this runs through a platform specific `GetInvalidFileCharacters` for the current platform and does not consider the cross platform file sync case.

You could argue this is a problem for the file sharing sync system (and I would somewhat agree), however whilst we can catch and translate invalid file name characters whenever we synchronise files going across the OS boundary, this is a lossy process and it's possible for users to create 2 different folders or files on Linux which map to the same filename on Windows (and mapping back the other way perfectly becomes impossible). This also means any scripts that run on linux and windows or export artifacts across the boundary will need to reference different folder names.

My suggestion is to have cross platform compatible names from the start and not include `:` or any other invalid windows characters in the directory names regardless of which platform the code is currently running on. We're going to implement as much as we can our side too but I suggest to prevent further issues we just stop using `:` in directory names, there is no real need to do it and anything that reduces friction working with cross-platform environments is a good thing.

JasonMurrayCole:
I ran into the issue in #678 today, so we wanted to contribute a fix for it. I'm happy to revise if I've not followed the contributor guidelines.

- Fixes #678

### DIFF
--- a/src/Adapter/PlatformServices.Desktop.Legacy/Utilities/DesktopDeploymentUtility.cs
+++ b/src/Adapter/PlatformServices.Desktop.Legacy/Utilities/DesktopDeploymentUtility.cs
@@ -57,8 +57,8 @@ namespace Microsoft.VisualStudio.TestPlatform.MSTestAdapter.PlatformServices.Uti
         /// <returns>Root deployment directory.</returns>
         public override string GetRootDeploymentDirectory(string baseDirectory)
         {
-            string dateTimeSufix = DateTime.Now.ToString("yyyy-MM-dd HH:mm:ss", DateTimeFormatInfo.InvariantInfo);
-            string directoryName = string.Format(CultureInfo.CurrentCulture, Resource.TestRunName, DeploymentFolderPrefix, Environment.UserName, dateTimeSufix);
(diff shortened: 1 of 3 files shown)


Haplois:
@JasonMurrayCole, thanks for the contribution, we'll be reviewing this.

PR #1126, opened 2022-06-09
Ensure assertions do not fail with FormatException

Evangelink:
Update the assertion message construction order to ensure user message is built first (user message could fail with FormatException) and then include this user message in our messages. This switch of order ensure we will never throw FormatException because our built message becomes an invalid format composite.

Fixes https://github.com/dotnet/sdk/issues/25373


src/TestFramework/MSTest.Core/Assertions/Assert.cs
```diff
@@ -88,8 +88,8 @@ public static void IsTrue([DoesNotReturnIf(false)] bool? condition)
         /// The condition the test expects to be true.
         /// </param>
         /// <param name="message">
-        /// The message to include in the exception when <paramref name="condition"/>
```
Evangelink: Looks like my rename renamed too many things, I will revert.

PR #1129, opened 2022-06-13
Removed conflicting dependencies

Haplois:


Evangelink (commented):
Is it normal to be at 17.2 for TP?

Haplois:
Yes, we usually use the last stable version.

Evangelink (approved):

PR #1147, opened 2022-07-11
Better messages for XXXInitialize and XXXCleanup

### LINKED ISSUE #1142: Why are AssertExceptions thrown within Assembly/Class/Test Initializations differentiated?
## Description
When an AssemblyInitialize or ClassInitialize throws an Exception (in most cases), it is formatted like:

```Assembly Initialization method {0}.{1} threw exception. {2}: {3}. Aborting test execution.```

When an AssemblyInitialize or ClassInitialize fails with an Assert exception, the message is like:

```Assert.IsTrue failed.```

## Question
Before, we use ```Assembly/Class Initialization method``` to determine if the test result contains an assembly/class initialization failure.
But I found if there doesn't throw an error, just like Assert failed or others, the results will not contain this message.
Does there exist a way to determine whether a test failed because assembly init or class init failed. I find all properties and fields in test results, but none of them works.

[AB#1568808](https://devdiv.visualstudio.com/0bdbc590-a062-4c3f-b0f6-9383f67865ee/_workitems/edit/1568808)

### PR DESCRIPTION
**Pull request by Evangelink**
# Description

Use the same informative message for all XXXInitialize and XXXCleanup methods whatever the kind of exception. We used to have a simpler handling for Assert exceptions that was removing the information about the method throwing the exception.

# Related issue

Fixes #1142 and fixes [AB#1543904](https://devdiv.visualstudio.com/0bdbc590-a062-4c3f-b0f6-9383f67865ee/_workitems/edit/1543904)

### DIFF
--- a/src/Adapter/MSTest.CoreAdapter/Execution/TestAssemblyInfo.cs
+++ b/src/Adapter/MSTest.CoreAdapter/Execution/TestAssemblyInfo.cs
@@ -169,25 +169,22 @@ namespace Microsoft.VisualStudio.TestPlatform.MSTest.TestAdapter.Execution
                 throw this.AssemblyInitializationException;
             }
 
-            var realException = this.AssemblyInitializationException.InnerException
-                                ?? this.AssemblyInitializationException;
+            var realException = this.AssemblyInitializationException.InnerException ?? this.AssemblyInitializationException;
 
-            var outcome = UnitTestOutcome.Failed;
-            if (!realException.TryGetUnitTestAssertException(out outcome, out var errorMessage, out var stackTraceInfo))
-            {
-                var exception = realException.GetType().ToString();
-                var message = StackTraceHelper.GetExceptionMessage(realException);
-                errorMessage = string.Format(
-                    CultureInfo.CurrentCulture,
-                    Resource.UTA_AssemblyInitMethodThrows,
-                    this.AssemblyInitializeMethod.DeclaringType.FullName,
-                    this.AssemblyInitializeMethod.Name,
-                    exception,
-                    message);
-                stackTraceInfo = StackTraceHelper.GetStackTraceInformation(realException);
-            }
+            var outcome = realException is AssertInconclusiveException ? UnitTestOutcome.Inconclusive : UnitTestOutcome.Failed;
 
-            var testFailedException = new TestFailedException(outcome, errorMessage, stackTraceInfo);
+            // Do not use StackTraceHelper.GetExceptionMessage(realException) as it prefixes the message with the exception type name.
+            var exceptionMessage = realException.TryGetMessage();
+            var errorMessage = string.Format(
+                CultureInfo.CurrentCulture,
+                Resource.UTA_AssemblyInitMethodThrows,
+                this.AssemblyInitializeMethod.DeclaringType.FullName,
+                this.AssemblyInitializeMethod.Name,
+                realException.GetType().ToString(),
+                exceptionMessage);
+            var exceptionStackTraceInfo = StackTraceHelper.GetStackTraceInformation(realException);
+
+            var testFailedException = new TestFailedException(outcome, errorMessage, exceptionStackTraceInfo, realException);
             this.AssemblyInitializationException = testFailedException;
 
             throw testFailedException;
--- a/src/Adapter/MSTest.CoreAdapter/Execution/TestClassInfo.cs
+++ b/src/Adapter/MSTest.CoreAdapter/Execution/TestClassInfo.cs
@@ -10,10 +10,12 @@ namespace Microsoft.VisualStudio.TestPlatform.MSTest.TestAdapter.Execution
     using System.Globalization;
     using System.Linq;
     using System.Reflection;
+
     using Extensions;
     using Microsoft.VisualStudio.TestTools.UnitTesting;
     using ObjectModel;
-    using UnitTestOutcome = Microsoft.VisualStudio.TestPlatform.MSTest.TestAdapter.ObjectModel.UnitTestOutcome;
+
+    using ObjectModelUnitTestOutcome = ObjectModel.UnitTestOutcome;
 
     /// <summary>
     /// Defines the TestClassInfo object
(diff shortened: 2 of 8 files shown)

### CODE REVIEW

File: src/Adapter/MSTest.CoreAdapter/Execution/TestAssemblyInfo.cs [hunk 1]
```diff
@@ -169,25 +169,22 @@ public void RunAssemblyInitialize(TestContext testContext)
...
-            var testFailedException = new TestFailedException(outcome, errorMessage, stackTraceInfo);
+            var realException = this.AssemblyInitializationException.InnerException ?? this.AssemblyInitializationException;
+
+            var outcome = realException is AssertInconclusiveException ? UnitTestOutcome.Inconclusive : UnitTestOutcome.Failed;
+
+            // Do not use StackTraceHelper.GetExceptionMessage(realException) as it prefixes the message with the exception type name.
+            var exceptionMessage = realException.TryGetMessage();
+            var errorMessage = string.Format(
+                CultureInfo.CurrentCulture,
+                Resource.UTA_AssemblyInitMethodThrows,
```
Evangelink, line 180: I suggest to edit the resource to remove the requirement to pass the exception type before the exception message as I think that we don't need to display the type for assert exceptions.

src/Adapter/MSTest.CoreAdapter/Execution/TestAssemblyInfo.cs
```diff
@@ -169,25 +169,22 @@ public void RunAssemblyInitialize(TestContext testContext)
...
+                this.AssemblyInitializeMethod.DeclaringType.FullName,
+                this.AssemblyInitializeMethod.Name,
+                realException.GetType().ToString(),
+                exceptionMessage);
+            var exceptionStackTraceInfo = StackTraceHelper.GetStackTraceInformation(realException);
+
+            var testFailedException = new TestFailedException(outcome, errorMessage, exceptionStackTraceInfo, realException);
```
Evangelink, line 187: I am now always passing the original exception as inner exception so that if we are debugging a complex scenario we can keep all the information.

src/Adapter/MSTest.CoreAdapter/Execution/TestClassInfo.cs
```diff
@@ -319,21 +321,20 @@ public void RunClassInitialize(TestContext testContext)
...
-
-                exceptionStackTraceInfo = realException.TryGetStackTraceInformation();
-            }
+            var outcome = realException is AssertInconclusiveException ? ObjectModelUnitTestOutcome.Inconclusive : ObjectModelUnitTestOutcome.Failed;
+
+            // Do not use StackTraceHelper.GetExceptionMessage(realException) as it prefixes the message with the exception type name.
+            var exceptionMessage = realException.TryGetMessage();
+            var errorMessage = string.Format(
+                CultureInfo.CurrentCulture,
+                Resource.UTA_ClassInitMethodThrows,
```
Evangelink, line 330: I suggest to edit the resource to remove the requirement to pass the exception type before the exception message as I think that we don't need to display the type for assert exceptions.

src/Adapter/MSTest.CoreAdapter/Execution/TestClassInfo.cs
```diff
@@ -319,21 +321,20 @@ public void RunClassInitialize(TestContext testContext)
...
+                this.ClassType.FullName,
+                failedClassInitializeMethodName,
+                realException.GetType().ToString(),
+                exceptionMessage);
+            var exceptionStackTraceInfo = StackTraceHelper.GetStackTraceInformation(realException);
 
-            var testFailedException = new TestFailedException(outcome, errorMessage, exceptionStackTraceInfo);
+            var testFailedException = new TestFailedException(outcome, errorMessage, exceptionStackTraceInfo, realException);
```
Evangelink, line 337: Also passing the original exception as inner exception.

src/Adapter/MSTest.CoreAdapter/Execution/TestMethodInfo.cs
```diff
@@ -441,24 +443,17 @@ private Exception HandleMethodException(Exception ex, string className, string m
             }
             else
             {
-                string errorMessage;
+                string errorMessage = string.Format(
+                    CultureInfo.CurrentCulture,
+                    Resource.UTA_TestMethodThrows,
+                    className,
+                    methodName,
+                    StackTraceHelper.GetExceptionMessage(realException));
```
Evangelink, line 451: We are not using this helper in a uniform way in the codebase. I would suggest to avoid using it (it prefixes the exception type name and also prints all inner exception type names and messages).

src/Adapter/MSTest.CoreAdapter/Execution/TestMethodInfo.cs
```diff
@@ -509,70 +504,66 @@ private void RunTestCleanupMethod(object classInstance, TestResult result)
...
                 Exception realException = ex.GetInnerExceptionOrDefault();
 
-                // special case UnitTestAssertException to trim off part of the stack trace
-                if (!realException.TryGetUnitTestAssertException(out cleanupOutcome, out var exceptionMessage, out var realExceptionStackTraceInfo))
+                if (testCleanupMethod != null)
+                {
+                    // Do not use StackTraceHelper.GetExceptionMessage(realException) as it prefixes the message with the exception type name.
+                    cleanupError.AppendFormat(
+                        CultureInfo.CurrentCulture,
+                        Resource.UTA_CleanupMethodThrows,
```
Evangelink, line 533: I suggest to edit the resource to remove the requirement to pass the exception type before the exception message as I think that we don't need to display the type for assert exceptions.

src/Adapter/MSTest.CoreAdapter/Execution/TestMethodInfo.cs
```diff
@@ -509,70 +504,66 @@ private void RunTestCleanupMethod(object classInstance, TestResult result)
...
-                            StackTraceHelper.GetExceptionMessage(exception));
-            }
-            else
-            {
-                return string.Format(
-                            CultureInfo.CurrentCulture,
-                            Resource.UTA_CleanupMethodThrowsGeneralError,
-                            this.TestClassName,
-                            StackTraceHelper.GetExceptionMessage(exception));
+                result.Outcome = result.Outcome.GetMoreImportantOutcome(realException is AssertInconclusiveException ? UTF.UnitTestOutcome.Inconclusive : UTF.UnitTestOutcome.Failed);
```
Evangelink, line 565: This was always forcing `Failed` for the result (but was preserving the `Inconclusive` at the exception level). I am uniformizing that.

Evangelink (commented):
DAMN! I forgot to publish my comments...

@Haplois Please review them, I will do a follow-up PR if needed.

src/Adapter/MSTest.CoreAdapter/Execution/TestAssemblyInfo.cs
Haplois: Sounds good for Assert, but don't you think losing exception type in logs will hurt us?

Evangelink: I want to keep the prefix for exceptions not inheriting from `TestAssetException` but remove it for ones inheriting as we know message is sufficient.

Haplois: Go ahead :shipit:

PR #1171, opened 2022-08-10
Fixed issues with SDK style projects.

Haplois:
- [x] Add commit to assemblies
 - [x] Fix `AssemblyInfo.cs` issues when marking assemblies with correct versions.
 - [x] Fix TestId bug.


scripts/build/TestFx.props
```diff
@@ -8,6 +8,12 @@
     <IsTest Condition=" '$(IsTest)' == '' ">false</IsTest>
   </PropertyGroup>
 
+  <PropertyGroup>
+    <BranchName Condition=" '$(BranchName)' == '' ">BRANCHNAME</BranchName>
+    <CommitHash Condition=" '$(CommitHash)' == '' ">COMMITHASH</CommitHash>
+    <CommitId>$(CommitHash.SubString(0, 8))</CommitId>
```
cvpoienaru, line 14: I understand that trimming the commit hash makes it easier to look at, but should we use the full commit hash instead ? I know it shouldn't come to it but I'm wondering what happens in case of a collision.

src/Adapter/MSTest.CoreAdapter/ObjectModel/UnitTestElement.cs
```diff
@@ -229,25 +229,26 @@ internal TestCase ToTestCase()
...
-                idProvider.AppendString(testCase.FullyQualifiedName);
-            }
-
-            if (this.TestMethod.DataType != DynamicDataType.None)
-            {
-                idProvider.AppendString(testCase.DisplayName);
-            }
-
-            testCase.Id = idProvider.GetId();
+            // This code is disabled for compatibility.
```
cvpoienaru, line 232: Would it be better to remove this code instead of commenting it out ? We'll always have the git log should we wish to restore it.

scripts/build/TestFx.props
Haplois: It won't happen. (Too low of a chance to happen.) But in that case we also store full commit hash in binary too.

src/Adapter/MSTest.CoreAdapter/ObjectModel/UnitTestElement.cs
Haplois: I almost sure that change will make people mad, so the commenting. If we're not asked to revert this change after preview - I'll remove the code.

MarcoRossignoli (approved):

PR #1192, opened 2022-08-29
Adapters: prefer specific tfm over generic ones

Evangelink:


Evangelink (commented):
Remove un-needed file

Haplois (approved):

PR #1203, opened 2022-08-31
Review compiler directives

Evangelink:


src/Adapter/PlatformServices.Shared/netstandard1.0/Services/ns10FileOperations.cs
```diff
@@ -36,11 +36,7 @@ public Assembly LoadAssembly(string assemblyName, bool isReflectionOnly)
     /// <returns>Path to the .DLL of the assembly.</returns>
     public string GetAssemblyPath(Assembly assembly)
     {
-#if NETSTANDARD1_5
```
Evangelink, line 39: @Haplois If I follow I should be returning `null` always now but it seems weird. I think the condition was forgotten when we evolved supported TFMs.

Please confirm my hypothesize.

Haplois: The main reason was that ns10 doesn't support `Assembly.Location`. It should always return the location now.

Haplois (approved):

Issue #1033, opened 2022-01-31
Assert.AreEqual(String, String, Boolean, CultureInfo) ignores the culture parameter


stogle:
## Description
The `Assert.AreEqual` overloads that accept strings and a `CultureInfo` parameter always perform an ordinal comparison, regardless of the culture parameter.

## Steps to reproduce
Run all tests in the following test class:
```
    [TestClass]
    public class AssertTests
    {
        private static readonly string Expected = "i";
        private static readonly string Actual = "I";
        private static readonly CultureInfo TurkishCulture = new CultureInfo("tr-TR");

        [TestMethod]
        public void StringEquals_WhenInvariantCultureIgnoreCase_ReturnsTrue()
        {
            Assert.IsTrue(Expected.Equals(Actual, StringComparison.InvariantCultureIgnoreCase));
        }

        [TestMethod]
        public void StringEquals_WhenTurkishCultureIgnoreCase_ReturnsFalse()
        {
            CultureInfo.CurrentCulture = TurkishCulture;
            Assert.IsFalse(Expected.Equals(Actual, StringComparison.CurrentCultureIgnoreCase));
        }

        [TestMethod]
        public void AssertAreEqual_WhenInvariantCultureIgnoreCase_ReturnsTrue()
        {
            Assert.AreEqual(Expected, Actual, true, CultureInfo.InvariantCulture);
        }

        [TestMethod]
        public void AssertAreEqual_WhenTurkishCultureIgnoreCase_ThrowsException()
        {
            var ex = Assert.ThrowsException<AssertFailedException>(() =>
            {
                Assert.AreEqual(Expected, Actual, true, TurkishCulture);
            });
            Assert.AreEqual($"Assert.AreEqual failed. Expected:<{Expected}>. Actual:<{Actual}>. ", ex.Message);
        }
    }
```

## Expected behavior
All tests should pass. In the `tr-TR` culture, "i" and "I" are not considered equal when doing a case-insensitive comparison.

## Actual behavior
The final test (`AssertAreEqual_WhenTurkishCultureIgnoreCase_ThrowsException`) fails with "Assert.ThrowsException failed. No exception thrown. AssertFailedException exception was expected."

This is because the `AreEqual` overloads that accept a `CultureInfo` parameter, pass it to the private `CompareInternal` method, and that method ignores the parameter and always does an ordinal comparison:

[...]

PR #1216, opened 2022-09-05
Fixed the case ignoring in AreEqual() with culture parameter

engyebrahim:
Fix: https://github.com/microsoft/testfx/issues/1033

### DIFF
--- a/src/TestFramework/MSTest.Core/Assertions/Assert.cs
+++ b/src/TestFramework/MSTest.Core/Assertions/Assert.cs
@@ -2946,7 +2946,7 @@ public sealed class Assert
 
     private static int CompareInternal(string expected, string actual, bool ignoreCase, CultureInfo culture)
     {
-        return string.Compare(expected, actual, ignoreCase ,culture);
+        return string.Compare(expected, actual, ignoreCase, culture);
     }
 
     #endregion
--- a/src/TestFramework/MSTest.Core/Assertions/Assert.cs
+++ b/src/TestFramework/MSTest.Core/Assertions/Assert.cs
@@ -2946,8 +2946,7 @@ public sealed class Assert
 
     private static int CompareInternal(string expected, string actual, bool ignoreCase, CultureInfo culture)
     {
-        CultureInfo.CurrentCulture = culture;
-        return string.Compare(expected, actual, ignoreCase ? StringComparison.CurrentCultureIgnoreCase : StringComparison.Ordinal);
+        return string.Compare(expected, actual, ignoreCase ,culture);
     }
 
     #endregion
--- a/src/TestFramework/MSTest.Core/Assertions/Assert.cs
+++ b/src/TestFramework/MSTest.Core/Assertions/Assert.cs
@@ -2946,7 +2946,8 @@ public sealed class Assert
 
     private static int CompareInternal(string expected, string actual, bool ignoreCase, CultureInfo culture)
     {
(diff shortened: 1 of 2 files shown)


Evangelink (changes requested):
As discussed offline, there are a couple of issues with current implementation so I am requesting changes for now.

src/TestFramework/MSTest.Core/Assertions/Assert.cs
```diff
@@ -2946,7 +2946,7 @@ internal static string ReplaceNulls(object input)
 
     private static int CompareInternal(string expected, string actual, bool ignoreCase, CultureInfo culture)
     {
-        return string.Compare(expected, actual, ignoreCase ? StringComparison.OrdinalIgnoreCase : StringComparison.Ordinal);
+        return string.Compare(expected, actual, ignoreCase ,culture);
```
Evangelink: Nit

```suggestion
        return string.Compare(expected, actual, ignoreCase, culture);
```

test/UnitTests/MSTest.Core.Unit.Tests/Assertions/AssertTests.cs
```diff
@@ -482,6 +483,50 @@ public void AreEqualShouldFailWhenNotEqualTypeWithMessage()
...
+    public void AreEqualShouldIgnorCaseBasedOnTheCurrentCulture()
+    {
+        var expected = "i";
+        var actual = "I";
+        var turkishCulture = new CultureInfo("tr-TR");
+
+        CultureInfo currentCulture = CultureInfo.CurrentCulture;
+        try
+        {
+            CultureInfo.CurrentCulture = turkishCulture;
```
Evangelink: Below you are using the API that accepts the culture parameter so there is no need to set (and then reset) the thread culture. You can simplify this test and the one below.

test/UnitTests/MSTest.Core.Unit.Tests/Assertions/AssertTests.cs
```diff
@@ -482,6 +483,50 @@ public void AreEqualShouldFailWhenNotEqualTypeWithMessage()
         StringAssert.Contains(ex.Message, "A Message");
     }
 
+    [TestMethod]
+    public void AreEqualShouldIgnorCaseBasedOnTheCurrentCulture()
```
Evangelink:
```suggestion
    public void AreEqualWithTurkishCultureAndIgnoreCaseReturnsFalse()
```

Evangelink: It would be nice to test that with English culture it's equal (you can either duplicate the test or take some `DataRow` with culture as string and expected result as bool). If you go with only one test, then test method name should be updated.

test/UnitTests/MSTest.Core.Unit.Tests/Assertions/AssertTests.cs
```diff
@@ -482,6 +483,50 @@ public void AreEqualShouldFailWhenNotEqualTypeWithMessage()
...
+        }
+        finally
+        {
+            // Reset the culture.
+            CultureInfo.CurrentCulture = currentCulture;
+        }
+    }
+
+    [TestMethod]
+    public void AreEqualShouldNotIgnorCaseBasedOnTheCurrentCulture()
```
Evangelink: Same comment here as before, it would be nice to test both cultures.

Evangelink (approved):

PR #1231, opened 2022-09-07
Improve assembly versions and available metadata

### LINKED ISSUE #1051: Provide an easy way to see the public version of MSTest
MSTest dlls do not contain the version under which it is published, which makes it really hard to see what is being used. Here I am using 2.2.9-preview-20220210-07.  I would like to be able to easily see the version in ilspy, and hopefully in file properties as well: 

image

image

Ideally this would also be accompanied by docs page that lists versions from all the available nugets on nuget.org and their publish version, **until we started embedding that data**. I can help automating that if you'd like.

Evangelink:
Fixes #1051 

Example when running locally:
image

Obviously, version number will be the correct ones when coming from CI.

### DIFF
--- a/scripts/build/TestFx.props
+++ b/scripts/build/TestFx.props
@@ -11,7 +11,6 @@
   <PropertyGroup>
     <BranchName Condition=" '$(BranchName)' == '' ">BRANCHNAME</BranchName>
     <CommitHash Condition=" '$(CommitHash)' == '' ">COMMITHASH</CommitHash>
-    <CommitId>$(CommitHash.SubString(0, 8))</CommitId>
   </PropertyGroup>
 
   <Import Project="$(RepoRoot)eng\Versions.props" />
--- a/scripts/build/TestFx.targets
+++ b/scripts/build/TestFx.targets
@@ -35,20 +35,21 @@
   <!-- Generate AssemblyInfo.cs -->
   <PropertyGroup>
     <TFBuildNumber Condition=" '$(TFBuildNumber)' == '' ">0.1</TFBuildNumber>
-    <MajorVersion>14.0</MajorVersion>
+    <LegacyMajorVersion>14</LegacyMajorVersion>
+    <LegacyMinorVersion>0</LegacyMinorVersion>
     <GeneratedAssemblyInfoFile Condition="'$(GeneratedAssemblyInfoFile)' ==''">$(IntermediateOutputPath)$(MSBuildProjectName).AssemblyInfo.cs</GeneratedAssemblyInfoFile>
     <Company>Microsoft Corporation</Company>
     <Copyright>© Microsoft Corporation. All rights reserved.</Copyright>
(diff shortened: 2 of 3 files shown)


scripts/build/TestFx.targets
```diff
@@ -35,20 +35,21 @@
   <!-- Generate AssemblyInfo.cs -->
   <PropertyGroup>
     <TFBuildNumber Condition=" '$(TFBuildNumber)' == '' ">0.1</TFBuildNumber>
-    <MajorVersion>14.0</MajorVersion>
+    <LegacyMajorVersion>14</LegacyMajorVersion>
+    <LegacyMinorVersion>0</LegacyMinorVersion>
```
Evangelink, line 40: It seems we are passing the version from the command line, is there any reason to keep these?

scripts/build/TestFx.targets
```diff
@@ -35,20 +35,21 @@
...
     <GeneratedAssemblyInfoFile Condition="'$(GeneratedAssemblyInfoFile)' ==''">$(IntermediateOutputPath)$(MSBuildProjectName).AssemblyInfo.cs</GeneratedAssemblyInfoFile>
     <Company>Microsoft Corporation</Company>
     <Copyright>© Microsoft Corporation. All rights reserved.</Copyright>
-    <AssemblyVersion>$(MajorVersion).0.0</AssemblyVersion>
-    <BuildVersion Condition=" '$(BuildVersion)' == '' ">$(MajorVersion).$(TFBuildNumber)</BuildVersion>
+    <AssemblyVersion>$(LegacyMajorVersion).$(LegacyMinorVersion).0.0</AssemblyVersion>
+    <BuildVersion Condition=" '$(BuildVersion)' == '' ">$(LegacyMajorVersion).$(LegacyMinorVersion).$(TFBuildNumber)</BuildVersion>
     <FileVersion Condition=" '$(FileVersion)' == '' ">$(BuildVersion)</FileVersion>
```
Evangelink, line 48: I haven't touched the `FileVersion` so far, as discussed we need to make tests to see if this can be safely updated.

scripts/build/TestFx.props
```diff
@@ -11,7 +11,6 @@
   <PropertyGroup>
     <BranchName Condition=" '$(BranchName)' == '' ">BRANCHNAME</BranchName>
     <CommitHash Condition=" '$(CommitHash)' == '' ">COMMITHASH</CommitHash>
-    <CommitId>$(CommitHash.SubString(0, 8))</CommitId>
```
Haplois, line 14: No one liked my short hashes :(

Haplois (dismissed):
:shipit:

scripts/build/TestFx.targets
Haplois: Yes, these are defaults for when you build it from VS.

Evangelink: Ah right! Let me add a comment and do the same for the other properties I have added.

PR #1232, opened 2022-09-08
Merge various PlatformServices.XXX.UnitTests projects together

Evangelink:
Aside from `PlatformServices.Universal.UnitTest`, merge all other TFM projects in one and ensure we properly run NetCore tests (they were broken).

NOTE: For .NET fwk we use MSTest v1 and for netcoreapp we use current built version of MSTest (that's not perfect because that means we won't catch some problems). The idea would be to build a small/simple test framework we can use to test MSTest whatever the TFM we use.


Haplois (approved):
:shipit:

PR #1235, opened 2022-09-08
Update mstest adapter tests to use correct implementation

Evangelink:


test/UnitTests/MSTest.CoreAdapter.Unit.Tests/Execution/LogMessageListenerTests.cs
```diff
@@ -92,21 +97,4 @@ public void DisposeShouldDisposeTraceListener()
     }
 
     #endregion
-
-    [TestMethod]
-    public void LogMessageListenerShouldCaptureLogMessagesInAllListeningScopes()
```
Evangelink, line 97: This test was testing the behavior of `ns10ThreadSafeStringWriter` which was used only by PlatformServices.Portable. With real implementation the behavior is totally different and the test makes no sense.

src/Adapter/PlatformServices/Services/FileOperations.cs
```diff
@@ -36,17 +36,17 @@ public FileOperations()
     /// <exception cref="NotImplementedException"> This is currently not implemented. </exception>
     public Assembly LoadAssembly(string assemblyName, bool isReflectionOnly)
     {
+#if NETSTANDARD || NETCOREAPP || WINDOWS_UWP
```
Evangelink, line 39: Simple refactoring to avoid confusion around the fact that `WIN_UI` is subset of `NETCOREAPP`.

src/Adapter/PlatformServices/Services/TestSourceHost.cs
```diff
@@ -252,16 +252,19 @@ private void SetContext(string source)
         _currentDirectory = Environment.CurrentDirectory;
         try
         {
+            // If the source is in the format of an assembly qualified name, then calls to
+            // Path.GetDirectoryName will return empty string. But if we use Path.GetFullPath first
+            // then directory resolution works properly.
+            var dirName = Path.GetDirectoryName(Path.GetFullPath(source));
```
Evangelink, line 258: This behavior fixes some of the tests. I don't think it has any bad consequences compared to existing code.

Haplois (approved):

PR #1252, opened 2022-09-14
Update PlatformServices.Desktop.Component.Tests to use new test fwk

Evangelink:
This PR also touches some other projects because these tests are referencing old versions of MSTest which is loaded in memory preventing the correct version (local) one to be loaded. This was ok up to now because we didn't do any big change (e.g. file rename).


TestFx.sln
```diff
@@ -157,8 +163,6 @@ Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "DoNotParallelizeTestProject
 EndProject
 Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "DeploymentTestProject", "test\E2ETests\TestAssets\DeploymentTestProject\DeploymentTestProject.csproj", "{3FCE3987-7C8C-4E12-B54A-C6EC2D9FC7A6}"
 EndProject
-Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "CompatTestProject", "test\E2ETests\TestAssets\CompatTestProject\CompatTestProject.csproj", "{2D2C5B73-F1F1-47C8-BC5C-A172E9BB3D16}"
```
Evangelink, line 158: Removed from solution but kept locally so we avoid its content to be discovery and so loaded in memory.

test/ComponentTests/TestAssets/TestProjectForAssemblyResolution/TestProjectForAssemblyResolution.csproj
```diff
@@ -7,9 +8,4 @@
     <AssemblyOriginatorKeyFile>$(RepoRoot)scripts\build\key.snk</AssemblyOriginatorKeyFile>
   </PropertyGroup>
 
-  <ItemGroup>
-    <PackageReference Include="Microsoft.NET.Test.Sdk" Version="$(Legacy__MicrosoftNETTestSdkVersion)" />
```
Evangelink, line 11: We actually don't care if that's a test project, all we care about is that the project type is loaded in memory because specified in the runsettings. Removing this reference avoid Test Explorer to discover and load the dlls in memory, causing side effects to our integration test.

test/ComponentTests/PlatformServices.Desktop.Component.Tests/App.config
```diff
@@ -1,18 +0,0 @@
-<?xml version="1.0" encoding="utf-8"?>
```
engyebrahim, line 1: please can you explain, why did you delete this file?

test/ComponentTests/PlatformServices.Desktop.Component.Tests/DesktopTestSourceHostTests.cs
```diff
@@ -105,7 +96,7 @@ private static string GetTestAssemblyPath(string assemblyName)
         return Path.Combine(testAssetPath, assemblyName);
     }
 
-    private Mock<IRunSettings> GetMockedIRunSettings(string runSettingxml)
+    private static Mock<IRunSettings> GetMockedIRunSettings(string runSettingxml)
```
engyebrahim, line 99: please can you explain, why we need it to be static?

test/ComponentTests/PlatformServices.Desktop.Component.Tests/DesktopTestSourceHostTests.cs
```diff
@@ -75,26 +67,25 @@ public void ChildDomainResolutionPathsShouldHaveSearchDirectoriesSpecifiedInRuns
 
         var assemblyResolution = "ComponentTests\\TestProjectForAssemblyResolution.dll";
         var asm = Assembly.LoadFrom(assemblyResolution);
-        var type = asm.GetType("PlatformServices.Desktop.ComponentTests.TestProjectForAssemblyResolution");
+        var type = asm.GetType("PlatformServices.Desktop.ComponentTests.SerializableTypeThatShouldBeLoaded");
```
engyebrahim: please can you explain, the difference between the two of them

test/ComponentTests/PlatformServices.Desktop.Component.Tests/ReflectionUtilityTests.cs
```diff
@@ -294,33 +268,33 @@ private Assembly ReflectionOnlyOnResolve(object sender, ResolveEventArgs args)
         return null;
     }
 
-    private string[] GetAttributeValuePairs(object[] attributes)
+    private static string[] GetAttributeValuePairs(object[] attributes)
```
engyebrahim, line 271: the same question why we need using static

test/ComponentTests/TestAssets/SampleFrameworkExtensions/Properties/AssemblyInfo.cs
```diff
@@ -1,27 +0,0 @@
-﻿// Copyright (c) Microsoft Corporation. All rights reserved.
```
engyebrahim, line 1: and please also explain, why we deleted this

engyebrahim (dismissed):

test/ComponentTests/PlatformServices.Desktop.Component.Tests/App.config
Evangelink: I did some cleaning, this file is actually no longer relevant (was already the case with main).

test/ComponentTests/PlatformServices.Desktop.Component.Tests/DesktopTestSourceHostTests.cs
Evangelink: Just took the liberty to address some code analyzers recommendation (method doesn't use any member so can be made static. `static` members are supposed to be faster so it's good to use them when possible).

test/ComponentTests/PlatformServices.Desktop.Component.Tests/ReflectionUtilityTests.cs
Evangelink: Just took the liberty to address some code analyzers recommendation (method doesn't use any member so can be made static. `static` members are supposed to be faster so it's good to use them when possible).

test/ComponentTests/TestAssets/SampleFrameworkExtensions/Properties/AssemblyInfo.cs
Evangelink: This file is not required and wasn't bringing any benefits so I removed it to cleanup :)

test/ComponentTests/PlatformServices.Desktop.Component.Tests/DesktopTestSourceHostTests.cs
Evangelink: Definitely :) Basically we don't need the project we load to be a test (see https://github.com/microsoft/testfx/pull/1252/files#r970936904) so to make it more obvious I decided to rename the type so we don't think it's a test that is missing attributes or something.

PR #1267, opened 2022-09-22
Ensure we run adapter tests for all supported TFMs

Evangelink:


src/Adapter/MSTest.TestAdapter/Discovery/AssemblyEnumerator.cs
```diff
@@ -160,7 +160,7 @@ internal static string GetLoadExceptionDetails(ReflectionTypeLoadException ex)
         var map = new Dictionary<string, object>(StringComparer.OrdinalIgnoreCase); // Exception -> null.
         var errorDetails = new StringBuilder();
 
-        if (ex.LoaderExceptions != null)
+        if (ex.LoaderExceptions?.Length > 0)
```
Evangelink, line 163: Looks like a change of behavior on .NET where we now have empty array instead of null.

Issue #1164, opened 2020-05-26
Test Classes should support IAsyncDisposable in addition to IDisposable.


invino4:
## Description
> **IMPORTANT**: if the defect is reproduced only in a workflow from within the Visual Studio IDE then do not report the issue here - instead, please report it using Visual Studio's "Send Feedback" option that can be accessed from the Help menu OR using this link https://developercommunity.visualstudio.com.
>
> For a defect reproducable from the vstest command line, describe the issue you've observed.

## Steps to reproduce
> What steps can reproduce the defect?
> Please share the setup, commandline for vstest.console, sample project, target
> framework etc.

````cs
    [TestClass]
    public sealed class SomeTests : IAsyncDisposable
    {
        public ValueTask DisposeAsync()
        {
            // should be called during test cleanup.
        }
    }
````

## Expected behavior
> Share the expected output
The DisposeAsync method should be called during test cleanup in the same manner as the Dispose method from IDisposable is called.  Either Dispose, DisposeAsync or both can be implemented by the test class.  If both are implemented then DisposeAsync should be preferred.  Only one is actually called.

## Actual behavior
> What is the behavior observed?
DisposeAsync is ignored and never executed.  See code [here](https://github.com/microsoft/testfx/blob/088fcd90d11466e416c1ddbf7dbd563c9e9830d9/src/Adapter/MSTest.CoreAdapter/Execution/TestMethodInfo.cs#L529).

## Diagnostic logs
> Please share test platform diagnostics logs. Instructions to collect logs are [here](https://github.com/Microsoft/vstest-docs/blob/master/docs/diagnose.md#test-platform-diagnostics).  
> The logs may contain test assembly paths, kindly review and mask those before sharing.

## Environment
> Please share additional details about the test environment.
> Operating system, Build version of vstest.console

[AB#1634620](https://devdiv.visualstudio.com/0bdbc590-a062-4c3f-b0f6-9383f67865ee/_workitems/edit/1634620)

PR #1288, opened 2022-09-27
Add support for AsyncDisposable cleanup

engyebrahim:
fix: https://github.com/microsoft/testfx/issues/1164


src/Adapter/MSTest.TestAdapter/Execution/TestMethodInfo.cs
```diff
@@ -493,6 +493,9 @@ private void RunTestCleanupMethod(object classInstance, TestResult result)
             }
             finally
             {
+#if NET6_0_OR_GREATER
```
Evangelink, line 495: It would be nice to capture the reasoning we had about why we think it's a good idea to keep the call to `Dispose()` after the call to `DisposeAsync()`. This will help us and others in the future :)

As a "general" rule, if you are doing something that is not straightforward, it's always better to explain the reasoning. It helps when doing review and when debugging/investigating later.

test/UnitTests/MSTest.CoreAdapter.Unit.Tests/Execution/TestMethodInfoTests.cs
```diff
@@ -806,6 +806,40 @@ public void TestMethodInfoInvokeShouldCallDiposeForDisposableTestClass()
         Verify(disposeCalled);
     }
 
+#if NET6_0_OR_GREATER
+    public void TestMethodInfoInvokeShouldCallDiposeAsyncForAsyncDisposableTestClass()
```
Evangelink: Suggestion:

```suggestion
    public void TestMethodInfoInvoke_WhenTestClassIsAsyncDisposable_ShouldDisposeAsync()
```

test/UnitTests/MSTest.CoreAdapter.Unit.Tests/Execution/TestMethodInfoTests.cs
```diff
@@ -806,6 +806,40 @@ public void TestMethodInfoInvokeShouldCallDiposeForDisposableTestClass()
...
+        var ctorInfo = typeof(DummyTestClassWithAsyncDisposable).GetConstructors().Single();
+        var testClass = new TestClassInfo(typeof(DummyTestClassWithAsyncDisposable), ctorInfo, null, _classAttribute, _testAssemblyInfo);
+        var method = new TestMethodInfo(typeof(DummyTestClassWithAsyncDisposable).GetMethod("DummyTestMethod"), testClass, _testMethodOptions);
+
+        method.Invoke(null);
+
+        Verify(asyncDisposeCalled);
+    }
+
+    public void TestMethodInfoInvokeShouldCallAsyncDiposeThenDiposeForDisposableAsyncDisposableTestClass()
```
Evangelink: Suggestion:

```suggestion
    public void TestMethodInfoInvoke_WhenTestClassIsDisposableAndAsyncDisposable_ShouldCallAsyncDisposeThenDispose()
```

test/UnitTests/MSTest.CoreAdapter.Unit.Tests/Execution/TestMethodInfoTests.cs
```diff
@@ -806,6 +806,40 @@ public void TestMethodInfoInvokeShouldCallDiposeForDisposableTestClass()
...
+        DummyTestClassWithAsyncDisposableAndDisposable.DisposeAsyncMethodBody = () => disposeAsyncCalledOrder = ++order;
+
+        var ctorInfo = typeof(DummyTestClassWithAsyncDisposableAndDisposable).GetConstructors().Single();
+        var testClass = new TestClassInfo(typeof(DummyTestClassWithAsyncDisposableAndDisposable), ctorInfo, null, _classAttribute, _testAssemblyInfo);
+        var method = new TestMethodInfo(typeof(DummyTestClassWithAsyncDisposableAndDisposable).GetMethod("DummyTestMethod"), testClass, _testMethodOptions);
+
+        method.Invoke(null);
+
+        Verify(disposeCalledOrder == 2);
+        Verify(disposeAsyncCalledOrder == 1);
```
Evangelink: The test method is clear as-is but as it starts to have a couple lines of code, I want to use this as an example for you to use the AAA comments (Arrange, Act, Assert). The goal is to help identify which part of the test is there to "arrange"/"setup" the test to be in the right state, which part is the actual behavior you want to test and which part is the assertion of the behavior.

```suggestion
        // Arrange
        var order = 0;
        var disposeCalledOrder = 0;
        var disposeAsyncCalledOrder = 0;

        DummyTestClassWithAsyncDisposableAndDisposable.DisposeMethodBody = () => disposeCalledOrder = ++order;
        DummyTestClassWithAsyncDisposableAndDisposable.DisposeAsyncMethodBody = () => disposeAsyncCalledOrder = ++order;

        var ctorInfo = typeof(DummyTestClassWithAsyncDisposableAndDisposable).GetConstructors().Single();
        var testClass = new TestClassInfo(typeof(DummyTestClassWithAsyncDisposableAndDisposable), ctorInfo, null, _classAttribute, _testAssemblyInfo);
        var method = new TestMethodInfo(typeof(DummyTestClassWithAsyncDisposableAndDisposable).GetMethod("DummyTestMethod"), testClass, _testMethodOptions);

        // Act
        method.Invoke(null);

        // Assert
        Verify(disposeCalledOrder == 2);
        Verify(disposeAsyncCalledOrder == 1);
```

test/UnitTests/MSTest.CoreAdapter.Unit.Tests/Execution/TestMethodInfoTests.cs
```diff
@@ -1562,5 +1596,48 @@ protected override void Verify(Exception exception)
     }
 
     #endregion
+
+#if NET6_0_OR_GREATER
+    public class DummyTestClassWithAsyncDisposable : IAsyncDisposable
+    {
+
```
Evangelink: Nit: You can remove the empty blank line.

```suggestion
```

Evangelink (commented):
That's a super good work @engyebrahim!

A few extra comments:
- Try to have a more "functional" title to your PRs so that it's easier what this is bringing without having to look at the implementation.

- The implementation with dummies and static Func/Action is a good idea but we need to be careful with them! If we add a new test reusing the dummy and the two tests are executed in parallel we might have test failure because the actual body was changed by another test. I need to investigate if we could do something similar to roslyn/roslyn-analyzers where we would compile classes on the fly and execute the content.

Evangelink (approved):

abatishchev:
Can you please update the docs with an example of what has been made possible?

Evangelink:
@abatishchev Sure! May I ask which doc you are referring to?

abatishchev:
@Evangelink that's what I'm looking for, any doc that would shed light on the new feature. The release notes only point to this issue. I couldn't find a sample or anything like that.

PR #1306, opened 2022-10-03
Introduce strategies for test ID generation

### LINKED ISSUE #1286: Introduce strategies for test ID generation
## Description

MSTest has been through various strategies to generate test IDs and so far the changes have been done in a one choice only behavior. Through this ticket, we want to fix this issue by introducing strategies (via some enum) to let users decide how we should generate test IDs.

We have identified the following strategies:
- Legacy: strategy used by from v1 to v2.2.3
- Display name: strategy used between v2.2.4 and v2.2.10
- Data: new default strategy used from v3 forward

We have identified the following issues as impacted/fixed by this change:

- https://github.com/microsoft/testfx/issues/1016
- https://github.com/microsoft/testfx/issues/1028
- https://github.com/microsoft/testfx/issues/1121
- https://github.com/microsoft/testfx/issues/1149

[AB#1634557](https://devdiv.visualstudio.com/0bdbc590-a062-4c3f-b0f6-9383f67865ee/_workitems/edit/1634557)

Evangelink:
Fixes #1286 and fixes [AB#1634557](https://devdiv.visualstudio.com/0bdbc590-a062-4c3f-b0f6-9383f67865ee/_workitems/edit/1634557)

TODO:

- [x] Add comments
- [x] Add tests


src/Adapter/MSTest.TestAdapter/Extensions/TestCaseExtensions.cs
```diff
@@ -76,15 +77,16 @@ internal static UnitTestElement ToUnitTestElement(this TestCase testCase, string
         var isAsync = (testCase.GetPropertyValue(Constants.AsyncTestProperty) as bool?) ?? false;
         var testClassName = testCase.GetPropertyValue(Constants.TestClassNameProperty) as string;
         var name = testCase.GetTestName(testClassName);
+        var testIdGenerationStrategy = (TestIdGenerationStrategy)testCase.GetPropertyValue(Constants.TestIdGenerationStrategyProperty, (int)TestIdGenerationStrategy.DisplayName);
```
Evangelink: @Haplois Why are you not retrieving as enum directly? `GetPropertyValue<TestIdGenerationStrategy>`?

src/Adapter/MSTest.TestAdapter/ObjectModel/UnitTestElement.cs
```diff
@@ -249,9 +272,38 @@ internal TestCase ToTestCase()
             idProvider.AppendString(testCase.DisplayName);
```
Evangelink, line 293: Any reason not to ditch the `display name`? It makes no sense to me to have "display name" as part of an ID. I would expect our ID to be simply similar to what roslyn does.

src/Adapter/MSTest.TestAdapter/Extensions/TestCaseExtensions.cs
Haplois: That type doesn't exist in TP it might cause problems when deserializing the attribute, also adds unnecessary metadata gloat.

src/Adapter/MSTest.TestAdapter/ObjectModel/UnitTestElement.cs
Haplois: It's there just to keep behavior consistent with 2.2.4+

Evangelink: 2.2.4 to 2.2.10 is using "display name" strategy, right? The "data" strategy is the new one you introduced to fix open bugs, right?

Haplois: Yep. You are correct. I will add some tests this weekend from my new flat hopefully.

Evangelink: FYI I am actually writing tests now.

Haplois (commented):
How is this going? Would you like any help @Evangelink?

Evangelink:
Hey thanks for asking @Haplois. Looks like I just need to fix CI tests (they are passing locally for me). There are still some tickets open (not fixed by this change) that I need to investigate but I feel confident we will be able to merge this one soon(TM).

src/TestFramework/TestFramework/Attributes/DataSource/TestIdGenerationStrategy.cs
```diff
@@ -0,0 +1,39 @@
...
+{
+    /// <summary>
+    /// Uses legacy test ID generation. <see cref="ITestDataSource"/> tests will not be discovered and instead they will be collapsed into one parent test.
+    /// </summary>
+    /// <remarks>
+    /// This option is incompatible with <see cref="TestDataSourceDiscoveryOption.DuringDiscovery"/> option of <see cref="TestDataSourceDiscoveryAttribute"/> and will be ignored.
+    /// This was the default option until version 2.2.3.
+    /// </remarks>
+    [Obsolete("This strategy is provided to reduce impact on existing users. It will be removed in a future release.")]
+    Legacy = 0,
```
Evangelink, line 21: Marked as obsolete so that we can drop it in a near future because we mainly only want to ease migration but not support all these ID generators.

src/TestFramework/TestFramework/Attributes/DataSource/TestIdGenerationStrategy.cs
```diff
@@ -0,0 +1,39 @@
...
+
+    /// <summary>
+    /// Uses a combination of executor ID, file name, fully qualified name and display name to generate the test ID.
+    /// </summary>
+    /// <remarks>
+    /// This is the default behavior between versions 2.2.4 and 3.0.0.
+    /// </remarks>
+    [Obsolete("This strategy is provided to reduce impact on existing users. It will be removed in a future release.")]
+    DisplayName = 1,
```
Evangelink, line 30: Marked as obsolete so that we can drop it in a near future because we mainly only want to ease migration but not support all these ID generators.

src/Adapter/MSTest.TestAdapter/Discovery/AssemblyEnumerator.cs
```diff
@@ -92,17 +93,25 @@ internal ICollection<UnitTestElement> EnumerateAssembly(string assemblyFileName,
...
-        var discoverInternals = assembly.GetCustomAttribute<UTF.DiscoverInternalsAttribute>() != null;
-        var testDataSourceDiscovery = assembly.GetCustomAttribute<UTF.TestDataSourceDiscoveryAttribute>()?.DiscoveryOption ?? UTF.TestDataSourceDiscoveryOption.DuringDiscovery;
-
+        var discoverInternals = assembly.GetCustomAttribute<DiscoverInternalsAttribute>() != null;
+        var testIdGenerationStrategy = assembly.GetCustomAttribute<TestIdGenerationStrategyAttribute>()?.Strategy
+            ?? TestIdGenerationStrategy.FullyQualifiedTest;
+
+        var testDataSourceDiscovery = assembly.GetCustomAttribute<TestDataSourceDiscoveryAttribute>()?.DiscoveryOption
+#pragma warning disable CS0618 // Type or member is obsolete
+            ?? (testIdGenerationStrategy == TestIdGenerationStrategy.Legacy
```
Evangelink, line 105: When using legacy strategy, there is no point in trying to "read" data during discovery as the ID will do nothing with them.

src/Adapter/MSTest.TestAdapter/Discovery/AssemblyEnumerator.cs
```diff
@@ -354,60 +366,38 @@ private static bool TryProcessTestDataSourceTests(UnitTestElement test, TestMeth
...
-            var discoveredIndex = new Dictionary<string, int>();
-            var serializationFailed = false;
             var index = 0;
 
             foreach (var d in data)
             {
                 var discoveredTest = test.Clone();
                 discoveredTest.DisplayName = dataSource.GetDisplayName(methodInfo, d) ?? discoveredTest.DisplayName;
 
-                // if we have a duplicate test name don't expand the test, bail out.
```
Evangelink, line 372: Maybe I should keep this logic for the "DisplayName" strategy.

src/Adapter/MSTest.TestAdapter/Execution/TestClassInfo.cs
```diff
@@ -303,7 +303,6 @@ public void RunClassInitialize(TestContext testContext)
                     finally
                     {
                         IsClassInitializeExecuted = true;
-                        Debug.Assert(testContext is ITestContext, "Test context doesn't implement ITestContext.");
```
Evangelink, line 306: Remove assertion because in some tests we are mocking the test context leading to this being false.

src/Adapter/MSTest.TestAdapter/Extensions/TestCaseExtensions.cs
```diff
@@ -74,17 +75,13 @@ internal static UnitTestElement ToUnitTestElement(this TestCase testCase, string
         var isAsync = (testCase.GetPropertyValue(Constants.AsyncTestProperty) as bool?) ?? false;
         var testClassName = testCase.GetPropertyValue(Constants.TestClassNameProperty) as string;
         var name = testCase.GetTestName(testClassName);
+        var testIdGenerationStrategy = (TestIdGenerationStrategy)testCase.GetPropertyValue(
+            Constants.TestIdGenerationStrategyProperty,
+            (int)TestIdGenerationStrategy.FullyQualifiedTest);
```
Evangelink: Default value is latest strategy.

test/E2ETests/Automation.CLI/DiscoveryEventsHandler.cs
```diff
@@ -13,12 +13,12 @@ public class DiscoveryEventsHandler : ITestDiscoveryEventsHandler
...
+    public List<string> Tests { get; } = new List<string>();
 
-    public DiscoveryEventsHandler()
-    {
-        Tests = new List<string>();
-    }
+    /// <summary>
+    /// Gets the list of messages received from the discovery process.
+    /// </summary>
+    public List<string> Messages { get; } = new List<string>();
```
Evangelink: Useful when debugging so I decided to keep it.

Evangelink (commented):
@Haplois please feel free to do a review when you have some free time.

src/Adapter/MSTest.TestAdapter/Discovery/AssemblyEnumerator.cs
```diff
@@ -354,60 +366,38 @@ private static bool TryProcessTestDataSourceTests(UnitTestElement test, TestMeth
...
-                    break;
-                }
-
                 try
                 {
                     discoveredTest.TestMethod.SerializedData = DataSerializationHelper.Serialize(d);
                     discoveredTest.TestMethod.DataType = DynamicDataType.ITestDataSource;
                 }
-                catch (SerializationException)
+                catch (SerializationException ex)
```
MarcoRossignoli, line 406: why don't log the message somewhere?

src/Adapter/MSTest.TestAdapter/ObjectModel/UnitTestElement.cs
```diff
@@ -183,57 +184,88 @@ internal TestCase ToTestCase()
...
+                // Legacy Id generation is to rely on default ID generation of TestCase from TestPlatform.
+                break;
+
+            case TestIdGenerationStrategy.DisplayName:
+                string filePath = testCase.Source;
+                try
+                {
+                    filePath = Path.GetFileName(filePath);
+                }
+                catch
```
MarcoRossignoli: can you add a comment?

src/Adapter/MSTest.TestAdapter/ObjectModel/UnitTestElement.cs
```diff
@@ -183,57 +184,88 @@ internal TestCase ToTestCase()
...
+                {
+                }
+
+                testCase.Id = GenerateDisplayNameStrategyTestId(testCase, filePath);
+                break;
+#pragma warning restore CS0618 // Type or member is obsolete
+
+            case TestIdGenerationStrategy.FullyQualifiedTest:
```
MarcoRossignoli:
```suggestion
            case TestIdGenerationStrategy.FullyQualified:
```

src/Adapter/MSTest.TestAdapter/PublicAPI/PublicAPI.Unshipped.txt
```diff
@@ -1 +1 @@
-﻿
\ No newline at end of file
+﻿Microsoft.VisualStudio.TestPlatform.MSTest.TestAdapter.ObjectModel.TestMethod.TestIdGenerationStrategy.get -> Microsoft.VisualStudio.TestTools.UnitTesting.TestIdGenerationStrategy
```
MarcoRossignoli, line 1: move to shipped?

src/Adapter/MSTest.TestAdapter/ObjectModel/UnitTestElement.cs
```diff
@@ -249,9 +281,26 @@ internal TestCase ToTestCase()
...
-        return testCase;
+    private Guid GenerateSerializedDataStrategyTestId()
+    {
+        var idProvider = new TestIdProvider();
+        idProvider.AppendString(Constants.ExecutorUriString);
+        idProvider.AppendString(TestMethod.AssemblyName);
+        idProvider.AppendString(TestMethod.FullClassName);
+        idProvider.AppendString(TestMethod.Name);
+
+        if (TestMethod.SerializedData != null)
```
MarcoRossignoli, line 328: are these parameters values to make the generation unique?

src/Adapter/MSTestAdapter.PlatformServices/PublicAPI/PublicAPI.Unshipped.txt
```diff
@@ -1 +1 @@
-﻿
\ No newline at end of file
+﻿Microsoft.VisualStudio.TestPlatform.MSTestAdapter.PlatformServices.Interface.ObjectModel.ITestMethod.TestIdGenerationStrategy.get -> Microsoft.VisualStudio.TestTools.UnitTesting.TestIdGenerationStrategy
```
MarcoRossignoli, line 1: move to shipped?

src/TestFramework/TestFramework/PublicAPI/PublicAPI.Unshipped.txt
```diff
@@ -1,2 +1,9 @@
 ﻿Microsoft.VisualStudio.TestTools.UnitTesting.DataRowAttribute.DataRowAttribute(object data) -> void
```
MarcoRossignoli, line 1: move to shipped?

test/E2ETests/Automation.CLI/DiscoveryEventsHandler.cs
MarcoRossignoli: List?It was IList I know that's not so different

test/E2ETests/DiscoveryAndExecutionTests/TestId.DefaultStrategy.cs
```diff
@@ -0,0 +1,190 @@
...
+
+using Microsoft.MSTestV2.CLIAutomation;
+using Microsoft.MSTestV2.Smoke.DiscoveryAndExecutionTests;
+
+namespace DiscoveryAndExecutionTests;
+public partial class TestId : CLITestBase
+{
+    private const string DefaultStrategyDll = "TestIdProject.DefaultStrategy.dll";
+
+    #region DataRow tests
```
MarcoRossignoli: region?

MarcoRossignoli (dismissed):

test/E2ETests/Automation.CLI/DiscoveryEventsHandler.cs
Evangelink: Ideally I would expose some immutable but as it's only for testing purposes I went with simplest solution. Regarding `IList` vs `List`, I tend to prefer to return type as specific as possible and accept as generic as possible.

src/Adapter/MSTest.TestAdapter/PublicAPI/PublicAPI.Unshipped.txt
Evangelink: I was thinking about moving all APIs to shipped either right before release or right after but in this way it's easier (IMO) to understand if that's an API I can still change because it's not yet published.

src/Adapter/MSTest.TestAdapter/Discovery/AssemblyEnumerator.cs
Haplois: If we remove this, it will cause error in some scenarios. If we're removing bail-out, we should hard fail in that scenario.

src/Adapter/MSTest.TestAdapter/Extensions/TestCaseExtensions.cs
Haplois: That might be breaking change for some users, in that case we should increment the major version IMO.

src/Adapter/MSTest.TestAdapter/ObjectModel/UnitTestElement.cs
Haplois: Yes. It should, unless we missed something.

Haplois (approved):
I think removing bail-out strategy might cause issues for some users, we need to rethink that, or at lease do a version increment.

test/E2ETests/Automation.CLI/DiscoveryEventsHandler.cs
MarcoRossignoli: ah it's test harness it's fine.

src/Adapter/MSTest.TestAdapter/PublicAPI/PublicAPI.Unshipped.txt
MarcoRossignoli: By experience we forgot to do that, but it's fine the same.

Evangelink:
Thanks for the review guys!

@Haplois What do you mean by the bail-out strategy?

src/Adapter/MSTest.TestAdapter/Discovery/AssemblyEnumerator.cs
Haplois: This is the bail-out strategy.

Haplois:
See my comment above on the code.

src/Adapter/MSTest.TestAdapter/Extensions/TestCaseExtensions.cs
Evangelink: Yes! This is part of release v3, we need major bump anyways due to changes on supported TFMs and also because we did a lot of refactoring I feel more comfortable with announcing major bump as we could have break something (uncaught by our tests).

Thanks for pointing it out!

src/TestFramework/TestFramework/PublicAPI/PublicAPI.Unshipped.txt
Evangelink: Will keep it as unshipped for now and if I see the same problem of "we forgot to move it", I will revise and use only shipped file. Thanks!

src/Adapter/MSTestAdapter.PlatformServices/PublicAPI/PublicAPI.Unshipped.txt
Evangelink: Will keep it as unshipped for now and if I see the same problem of "we forgot to move it", I will revise and use only shipped file. Thanks!

src/Adapter/MSTest.TestAdapter/PublicAPI/PublicAPI.Unshipped.txt
Evangelink: Will keep it as unshipped for now and if I see the same problem of "we forgot to move it", I will revise and use only shipped file. Thanks!

src/TestFramework/TestFramework/Attributes/DataSource/TestIdGenerationStrategyAttribute.cs
```diff
@@ -0,0 +1,29 @@
...
+
+namespace Microsoft.VisualStudio.TestTools.UnitTesting;
+
+/// <summary>
+/// Specifies how to generate test ID.
+/// </summary>
+[AttributeUsage(AttributeTargets.Assembly, AllowMultiple = false)]
+public class TestIdGenerationStrategyAttribute : Attribute
+{
+    /// <summary>
```
Haplois, line 14: Should we add a `DefaultStrategy` const here?

src/TestFramework/TestFramework/Attributes/DataSource/TestIdGenerationStrategy.cs
Haplois: Setting this to 0 makes it default value when it's not set. Do we want that?

src/Adapter/MSTest.TestAdapter/ObjectModel/UnitTestElement.cs
```diff
@@ -249,9 +281,26 @@ internal TestCase ToTestCase()
...
-        testCase.Id = idProvider.GetId();
+        return idProvider.GetId();
+    }
 
-        return testCase;
+    private Guid GenerateSerializedDataStrategyTestId()
+    {
+        var idProvider = new TestIdProvider();
+        idProvider.AppendString(Constants.ExecutorUriString);
+        idProvider.AppendString(TestMethod.AssemblyName);
```
Evangelink: It's not fully clear to me if this is important to keep or not. I was told that we are supposed to consider 2 assemblies in different locations as different (this would help to disambiguate metadata for example) but we are not doing that in legacy nor display name (we only look at assembly name not assembly path).

@Haplois Do you know if that's a requirement?

src/Adapter/MSTest.TestAdapter/Discovery/AssemblyEnumerator.cs
Evangelink: Reverting this part.

src/TestFramework/TestFramework/Attributes/DataSource/TestIdGenerationStrategyAttribute.cs
Evangelink: What would it bring for you?

src/Adapter/MSTest.TestAdapter/ObjectModel/UnitTestElement.cs
Haplois: There is a bug there, trying to fix it :)

src/TestFramework/TestFramework/Attributes/DataSource/TestIdGenerationStrategyAttribute.cs
Haplois: Good question, probably nothing - it might let user to know what is the current default.

src/Adapter/MSTest.TestAdapter/ObjectModel/UnitTestElement.cs
Evangelink: Do you mean this is a bug on older versions and this is something we want to fix with new strategy?

src/TestFramework/TestFramework/Attributes/DataSource/TestIdGenerationStrategy.cs
Evangelink: I have run some manual checks and added some integration tests and I confirm that having `Legacy` as `0` is not a problem. When nothing is setup we correctly fallback to default being `FullyQualified`.

src/TestFramework/TestFramework/Attributes/DataSource/TestIdGenerationStrategyAttribute.cs
Evangelink: Will keep as-is for now then. I expect that the comments are providing enough information and I don't expect people to use explicitly `FullyQualified`.

test/E2ETests/Automation.CLI/DiscoveryEventsHandler.cs
Evangelink: Went with immutable anyways because it's not much more work and it's cleaner :)

src/Adapter/MSTest.TestAdapter/Discovery/AssemblyEnumerator.cs
Evangelink: Logged.

src/Adapter/MSTest.TestAdapter/ObjectModel/UnitTestElement.cs
Evangelink: Done

test/E2ETests/DiscoveryAndExecutionTests/TestId.DefaultStrategy.cs
Evangelink: Removed :)

Evangelink:
@MarcoRossignoli @Haplois I have addressed all comments, if you want to have another look. @Haplois There is one open question remaining for you.

src/Adapter/MSTest.TestAdapter/ObjectModel/UnitTestElement.cs
Haplois: Fix is attached.

[0001-Fixed-testId-generation-bug.zip](https://github.com/microsoft/testfx/files/9869774/0001-Fixed-testId-generation-bug.zip)

Haplois:
@evangelink uploaded a patch file for the broken ID gen. Can't push to this branch anymore.

src/Adapter/MSTest.TestAdapter/ObjectModel/UnitTestElement.cs
Evangelink: Few questions:
- What's the problem with using `AssemblyName` for path?
- You are using `Path.GetFileName` which gives only assembly name and not full path, is it expected?
- What's the problem with relying only on `FullClassName`? And not the ManagedXXX?

Haplois: We want to migrate away from FullyQualifedName and stop using it since it's not standardized. 

First two questions have the same answer: backwards compatibility. This is how TP handles ID gen.

Evangelink: Ok but if that's backward compatibility it's ok to not use that in the new strategy because we don't have to care about past. As for `FullyQualifiedName`, I haven't seen any case where I have a value that differ so I am not sure to really understand the benefit. Do you have examples I could add for my tests?

The fact that it is not standard might be bothering for display name, UI positioning or something else but for test ID I don't see any problem. Also, looking at how it's done, we use runtime reflection info so I don't see how it could be bad/wrong.

Evangelink: Also, what about the path? We currently use only file name and not path, is it something that needs to be fixed or is it ok?

Evangelink:
@Haplois I have applied and updated your patch. Let me know if you have any other comment or if I can move forward with merging the PR.

src/Adapter/MSTest.TestAdapter/ObjectModel/UnitTestElement.cs
```diff
@@ -243,15 +243,16 @@ private void SetTestCaseId(TestCase testCase)
                 {
                     filePath = Path.GetFileName(filePath);
                 }
-                catch
+                catch (ArgumentException)
                 {
+                    // In case path contains invalid characters.
```
MarcoRossignoli: I think you know about https://learn.microsoft.com/en-us/dotnet/api/system.io.path.getinvalidpathchars?view=netstandard-2.0

is it on purpose this catch?

MarcoRossignoli (approved):

src/Adapter/MSTest.TestAdapter/ObjectModel/UnitTestElement.cs
Evangelink: As discussed offline, keeping as-is as we expect failure to be rare.

PR #1307, opened 2022-10-03
Ensure exit in case of timeout/cancellation

### LINKED ISSUE #522: Timeout attribute does not call test cleanup method
## Steps to reproduce
```c#
[TestClass] 
public class TestCleanup 
{ 
[TestInitialize()] 
public void MyTestInitialize() 
{ 
Console.WriteLine("Calling initialize"); 
} 


// Use TestCleanup to run code after each test has run 

[TestCleanup()] 
public void MyTestCleanup() 
{ 
MessageBox.Show("Cleanup"); 
Console.WriteLine( ng cleanup"); 
} 

[Timeout(5000)] 
[TestMethod] 
public void TestMethod1() 
{ 
Console.WriteLine("Sleeping for 10 seconds"); 
System.Threading.Thread.Sleep(10000); 
Console.WriteLine("Waking up"); 
} 

} 
```

## Expected behavior
It is essential all tests execute their TestCleanup code, as this code may have important steps to recover from an error such as a time out. 

## Actual behavior
The result of executing this code is the test times out but the message box is not displayed and the Calling cleanup message is not written to the console.

[AB#1634637](https://devdiv.visualstudio.com/0bdbc590-a062-4c3f-b0f6-9383f67865ee/_workitems/edit/1634637)

Evangelink:
Ensure sync and async calls to constructor, test initialize, test method, test cleanup and dispose awaiting can be cancelled to respect timeout/cancellation feature.

Ensure test cleanup and dispose are correctly called even in case of timeout/cancellation during test initialize or test method.

Also introduces a new property onto the TimeoutAttribute to control the time given to the cleanup in case of a timeout/cancellation of the test run.

Fixes #522

PS: There are more refactoring I would like to introduce (e.g., reducing number of try/catch) but I wanted to keep this change as minimal as possible. Once this is merged, I will work on refactoring, simplification and cleanup of this class.

### DIFF
--- a/src/Adapter/MSTest.TestAdapter/Execution/TestMethodInfo.cs
+++ b/src/Adapter/MSTest.TestAdapter/Execution/TestMethodInfo.cs
@@ -688,7 +688,8 @@ public class TestMethodInfo : ITestMethod
         object classInstance = null;
         try
         {
-            classInstance = await Task.Run(() => Parent.Constructor.Invoke(null), cancellationToken).WithCancellation(cancellationToken);
+            classInstance = await Task.Run(() => Parent.Constructor.Invoke(null), cancellationToken)
+                .WithCancellation(cancellationToken);
         }
         catch (Exception ex)
         {
--- a/src/Adapter/MSTest.TestAdapter/Execution/TestMethodInfo.cs
+++ b/src/Adapter/MSTest.TestAdapter/Execution/TestMethodInfo.cs
@@ -120,6 +120,7 @@ public class TestMethodInfo : ITestMethod
             {
                 result = IsTimeoutSet
                     ? ExecuteInternalWithTimeout(arguments)
+                    // TODO: Break public API and flow up async/await call. In the meantime, let's use blocking call.
                     : ExecuteInternal(arguments, CancellationToken.None, CancellationToken.None).GetAwaiter().GetResult();
             }
             finally
@@ -785,6 +786,7 @@ public class TestMethodInfo : ITestMethod
         {
             try
             {
+                // TODO: Rewrite ThreadOperations to accept async/await pattern. In the meantime, we have to use blocking call.
                 result = ExecuteInternal(arguments, testRunCancellationToken, cleanupTokenSource.Token).GetAwaiter().GetResult();
             }
             catch (Exception ex)
--- a/src/Adapter/MSTest.TestAdapter/Execution/TestMethodInfo.cs
+++ b/src/Adapter/MSTest.TestAdapter/Execution/TestMethodInfo.cs
@@ -11,6 +11,7 @@ using System.Globalization;
 using System.Reflection;
 using System.Text;
 using System.Threading;
+using System.Threading.Tasks;
 
 using Microsoft.VisualStudio.TestPlatform.MSTest.TestAdapter.Extensions;
 using Microsoft.VisualStudio.TestPlatform.MSTest.TestAdapter.Helpers;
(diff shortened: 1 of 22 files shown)


src/TestFramework/TestFramework/Attributes/TestMethod/TimeoutAttribute.cs
```diff
@@ -19,9 +19,15 @@ public sealed class TimeoutAttribute : Attribute
     /// <param name="timeout">
     /// The timeout in milliseconds.
     /// </param>
-    public TimeoutAttribute(int timeout)
+    /// <param name="cleanupTimeout">
+    /// The cleanup timeout in milliseconds.
+    /// This represents the time allowed for the cleanup (TestCleanup, Dispose) to complete when the test times out.
+    /// The default value is infinite, meaning that all cleanup will be called and left infinite time to process.
+    /// </param>
+    public TimeoutAttribute(int timeout, int cleanupTimeout = int.MaxValue)
```
Evangelink, line 27: I have made multiple tests and I think it's better to have the default behavior being "call cleanup".

src/Adapter/MSTest.TestAdapter/Execution/TestMethodInfo.cs
```diff
@@ -117,14 +118,9 @@ public virtual TestResult Invoke(object[] arguments)
...
-                {
-                    result = ExecuteInternalWithTimeout(arguments);
-                }
-                else
-                {
-                    result = ExecuteInternal(arguments);
-                }
+                result = IsTimeoutSet
+                    ? ExecuteInternalWithTimeout(arguments)
+                    : ExecuteInternal(arguments, CancellationToken.None, CancellationToken.None).GetAwaiter().GetResult();
```
MarcoRossignoli, line 124: Pls add a comment to explain why you need this blocking call

MarcoRossignoli: Also are we sure we want block here? It means that starting from now we're pushing the thread pool in this hot loop(test invoke is our host loop I suppose) are the perf good? Cannot we move this wait above in the chain a bit?

src/Adapter/MSTest.TestAdapter/Execution/TestMethodInfo.cs
```diff
@@ -762,9 +769,31 @@ void executeAsyncAction()
...
             TestResult timeoutResult = new() { Outcome = UTF.UnitTestOutcome.Timeout, TestFailureException = new TestFailedException(ObjectModelUnitTestOutcome.Timeout, errorMessage) };
             return timeoutResult;
         }
+
+        // Local functions
+        void ExecuteAsyncAction()
+        {
+            try
+            {
+                result = ExecuteInternal(arguments, testRunCancellationToken, cleanupTokenSource.Token).GetAwaiter().GetResult();
```
MarcoRossignoli, line 790: don't we have a sync way here too?

src/TestFramework/TestFramework/Attributes/TestMethod/TimeoutAttribute.cs
MarcoRossignoli: agree

MarcoRossignoli (approved):

src/TestFramework/TestFramework/Attributes/TestMethod/TimeoutAttribute.cs
nohwnd: I have the opposite feeling, people add the Timeout to make sure that the tests actually finish. So we are balancing two opposite things, ensure it finishes, and have the cleanup run to completion.

Personally I would add 10% of the timeout time to allow cleanup to complete. And let people define their "grace period" in milliseconds or as infinite.

That way they can have both.

Issue #1180, opened 2022-08-25
Cannot provide 2 arrays to DataRow


nohwnd:
## Description

DataRow does not allow providing 2 array parameters.. 

## Steps to reproduce

Use `[DataRow(new[] { "a" }, new[] {"b"} )]` in a test.

## Expected behavior

I can provide any reasonable number of parameters in the same form. Possibly there are constructor overloads that take 1 - 7 parameters for DataRow, and DataRow is not sealed so I can add more.

## Actual behavior

You get error: 

CS0182	An attribute argument must be a constant expression, typeof expression or array creation expression of an attribute parameter type.

This is confusing because we do provide types that fit this criteria, and because using three arrays works just fine.  
`[DataRow(new[] { "a" }, new[] { "b" }, new[] { "c" })]`

In reality you are running into the implementation detail of DataRow which uses params and that makes array mismatch between the inferred `string[]` and the destination `object[]` types.

You can work around the issue by wrapping your parameters into additional array, or by subclassing DataRow:

```cs 
[TestClass]
public class UnitTest1
{
    [TestMethod]
    // [DataRow(new[] { "a" }, new[] { "b" })] // Does not compile
    [DataRow(new[] { "a" }, new object[] { new[] { "b" } })]
    [DataRow2(new[] { "c" }, new[] { "d" })]
    public void TestMethod1(string[] _, string[] __)
    {
    }

    [TestMethod]
    [DataRow(new[] { "a" }, new[] { "b" }, new[] { "c" })]
    [DataRow2(new[] { "d" }, new[] { "e" }, new[] { "f" })]
    public void TestMethod2(string[] _, string[] __, string[] ___)
    {
    }
}

public class DataRow2Attribute : DataRowAttribute
{
    public DataRow2Attribute(object data1) : base(data1) { }
    public DataRow2Attribute(object data1, object data2) : base(data1, new[] { data2 }) { }
    public DataRow2Attribute(object data1, object data2, object data3) : base(data1, new[] { data2, data3 }) { }
    public DataRow2Attribute(object data1, object data2, object data3, object data4) : base(data1, new[] { data2, data3, data4 }) { }
}
```

This still does not fix the issue fully, because the name of the method is translated to  `TestMethod2 (System.String[],System.String[],System.String[])`, which makes only one of the row to run per method, not both. 

Get display name is not overridable, so unless I want to implement the whole attribute myself I don't have an easy way to override it.

[...]

PR #1332, opened 2022-10-12
Refactor available ctors for DataRowAttribute

Evangelink:
Fixes the issue #1180 that was preventing to have exactly 2 arrays as argument of DataRowAttribute ctor because the compiler was considering the second array as the params array and not as a first item of it.

### DIFF
--- a/src/TestFramework/TestFramework/Attributes/DataSource/DataRowAttribute.cs
+++ b/src/TestFramework/TestFramework/Attributes/DataSource/DataRowAttribute.cs
@@ -4,6 +4,7 @@
 using System;
 using System.Collections.Generic;
 using System.Globalization;
+using System.Linq;
 using System.Reflection;
 
 namespace Microsoft.VisualStudio.TestTools.UnitTesting;
--- a/src/TestFramework/TestFramework/Attributes/DataSource/DataRowAttribute.cs
+++ b/src/TestFramework/TestFramework/Attributes/DataSource/DataRowAttribute.cs
@@ -37,7 +37,7 @@ public class DataRowAttribute : Attribute, ITestDataSource
     /// <summary>
     /// Gets data for calling test method.
     /// </summary>
-    public object[] Data { get; }
+    public object[] Data { get; private set; }
 
     /// <summary>
     /// Gets or sets display name in test results for customization.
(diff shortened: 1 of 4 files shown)


test/UnitTests/TestFramework.UnitTests/Attributes/DataRowAttributeTests.cs
```diff
@@ -84,12 +100,68 @@ public void GetDisplayNameShouldReturnSpecifiedDisplayName()
...
+        // Arrange
+        var dataRow = new DataRowAttribute(new[] { "a" }, new[] { "1" });
+        var methodInfoMock = new Mock<MethodInfo>();
+        methodInfoMock.SetupGet(x => x.Name).Returns("MyMethod");
+
+        // Act
+        var displayName = dataRow.GetDisplayName(methodInfoMock.Object, dataRow.Data);
+
+        // Assert
+        Verify(displayName == "MyMethod (System.String[],System.String[])");
```
Evangelink, line 151: This display name and the next one are not ideal but we will address it in a different PR (see issue https://github.com/microsoft/testfx/issues/1333).

engyebrahim (approved):

src/TestFramework/TestFramework/Attributes/DataSource/DataRowAttribute.cs
```diff
@@ -18,34 +18,21 @@ public class DataRowAttribute : Attribute, ITestDataSource
...
     }
 
     /// <summary>
     /// Initializes a new instance of the <see cref="DataRowAttribute"/> class which takes in an array of arguments.
     /// </summary>
-    /// <param name="data1"> A data object. </param>
-    /// <param name="moreData"> More data. </param>
-    public DataRowAttribute(object data1, params object[] moreData)
+    /// <param name="data"> More data. </param>
+    public DataRowAttribute(params object[] data)
```
MarcoRossignoli, line 32: it's a breaking right?

Evangelink: Yes that's a breaking change but we are targeting MSTest v3 so it "should be ok". Outside of compilation, I also expect little to no change from users.

Issue #1328, opened 2022-10-10
Different behavior with TestContext.WriteLine() in [ClassInitialize] and [ClassCleanup] function. 


engyebrahim:
## Describe the bug

Now if we add `TestContext.WriteLine(message)` in the `ClassInitialize` function, the "message" will appear with one (even if we have multiple tests) of the test function result. And if we have `TestContext.WriteLine(message)` in the `ClassCleanup` function, the "message" won't appear at all.

## Steps To Reproduce

Provide the steps to reproduce the behavior:

1. Add a test class.
2. Add `[ClassInitialize]`, `[ClassCleanup]` and `[TestMethod] ` methods.
3. Add `TestContext.WriteLine()` to all of them with different messages.
4. Run the test and will see that you have the initialize message and the testmethod message but not the cleanup message.

## Expected behavior

We have the messages from both `ClassInitialize` and `ClassCleanup`, or don't have the messages from any of them.

## Actual behavior
We have the messages from `ClassInitialize` but not from `ClassCleanup`.

[AB#1642314](https://devdiv.visualstudio.com/0bdbc590-a062-4c3f-b0f6-9383f67865ee/_workitems/edit/1642314)

PR #1339, opened 2022-10-14
Remove class/assembly initialization messages from logs.

engyebrahim:
Fix: https://github.com/microsoft/testfx/issues/1328

### DIFF
--- a/src/Adapter/MSTest.TestAdapter/Execution/TestClassInfo.cs
+++ b/src/Adapter/MSTest.TestAdapter/Execution/TestClassInfo.cs
@@ -303,7 +303,7 @@ public class TestClassInfo
                     finally
                     {
                         IsClassInitializeExecuted = true;
-                        Debug.Assert(testContext is ITestContext, "TestContext is null.");
+                        Debug.Assert(testContext is ITestContext, "Test context doesn't implement ITestContext.");
                         (testContext as ITestContext)?.ClearDiagnosticMessages();
                     }
                 }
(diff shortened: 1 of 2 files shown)


src/Adapter/MSTest.TestAdapter/Execution/TestClassInfo.cs
```diff
@@ -302,6 +303,7 @@ public void RunClassInitialize(TestContext testContext)
                     finally
                     {
                         IsClassInitializeExecuted = true;
+                        (testContext as ITestContext).ClearDiagnosticMessages();
```
Evangelink: You are doing a `as` cast which means that it's possible the value returns `null` which would cause a `NullReferenceException` when calling `ClearDiagnosticMessages`. It's ok to play safe by doing `as` cast but then you should handle the null case. Besides, we could probably add a `Debug.Assert` so that we get some warning if this assumption suddenly becomes invalid.

```suggestion
                        Debug.Assert(testContext is ITestContext);
                        (testContext as ITestContext)?.ClearDiagnosticMessages();
```

Evangelink (commented):
PR title is a little long and could probably be simplified to `Remove class/assembly initialization messages from logs`.

I see that the PR is marked as fixing the ticket which means that we will be left with no ticket tracking the fact that we ultimately want to have assembly/class initialize/cleanup messages to be printed somehow/somewhere.

src/Adapter/MSTest.TestAdapter/Execution/TestClassInfo.cs
```diff
@@ -302,6 +303,8 @@ public void RunClassInitialize(TestContext testContext)
                     finally
                     {
                         IsClassInitializeExecuted = true;
+                        Debug.Assert(testContext is ITestContext, "TestContext is null.");
```
Evangelink: Test context is not null if the assertion fails, it means that test context doesn't implement `ITestContext` :)

Evangelink (approved):

Issue #1340, opened 2022-10-14
Assembly and class cleanup and init TestContext.WriteLine() don't appear in any place.


engyebrahim:
## Describe the bug

if you added TestContext.WriteLine() in class/assembly cleanup or init  won;t appear in any results/logs.
unlike the ones we added in the test cleanup/init.

## Steps To Reproduce

1. add TestContext.WriteLine() in classinit
2. add TestContext.WriteLine() in testinit
3. you will see the message from testinit in the test logs but will not see the one from classinit

## Expected behavior
get the logs from class/assemblt init/cleanup functions

## Actual behavior
we don't get the logs from class/assemblt init/cleanup functions

[AB#1643045](https://devdiv.visualstudio.com/0bdbc590-a062-4c3f-b0f6-9383f67865ee/_workitems/edit/1643045)

PR #1390, opened 2022-11-16
Add class/assembly cleanup/init messages to first/last test

engyebrahim:
fix: https://github.com/microsoft/testfx/issues/1340


src/Adapter/MSTest.TestAdapter/Execution/UnitTestRunner.cs
```diff
@@ -134,16 +136,41 @@ internal UnitTestResult[] RunSingleTest(TestMethod testMethod, IDictionary<strin
...
+                bool shouldRunCleanup = false;
                 if (_classCleanupManager is null)
                 {
                     return notRunnableResult;
                 }
 
                 DebugEx.Assert(testMethodInfo is not null, "testMethodInfo should not be null.");
-                _classCleanupManager.MarkTestComplete(testMethodInfo, testMethod, out bool shouldRunClassCleanup);
-                if (shouldRunClassCleanup)
+                _classCleanupManager?.MarkTestComplete(testMethodInfo, testMethod, out shouldRunClassCleanup, out shouldRunCleanup);
```
Evangelink: This doesn't need to be nullable, null check is done on line 141.

src/Adapter/MSTest.TestAdapter/Execution/UnitTestRunner.cs
```diff
@@ -134,16 +136,41 @@ internal UnitTestResult[] RunSingleTest(TestMethod testMethod, IDictionary<strin
...
+                try
```
Evangelink: Can you explain why this try/finally was added?

src/Adapter/MSTest.TestAdapter/Execution/UnitTestRunner.cs
```diff
@@ -134,16 +136,41 @@ internal UnitTestResult[] RunSingleTest(TestMethod testMethod, IDictionary<strin
...
-                    testMethodInfo.Parent.RunClassCleanup(ClassCleanupBehavior.EndOfClass);
+                    // Class cleanup can throw exceptions in which case we need to ensure that we fail the test.
+                    if (shouldRunClassCleanup)
+                    {
+                        testMethodInfo.Parent.RunClassCleanup(ClassCleanupBehavior.EndOfClass);
+                    }
+
+                    if (shouldRunCleanup)
+                    {
+                        RunCleanup();
```
Evangelink: What is this cleanup doing? Class.EndOfAssembly + Assembly cleanup? If so, could you rename the method and also the 
 `shouldRunCleanup` variable?

src/Adapter/MSTest.TestAdapter/Execution/UnitTestRunner.cs
```diff
@@ -134,16 +136,41 @@ internal UnitTestResult[] RunSingleTest(TestMethod testMethod, IDictionary<strin
...
+                }
+                finally
+                {
+                    using LogMessageListener logListener = new(MSTestSettings.CurrentSettings.CaptureDebugTraces);
+                    var cleanupTestContextMessages = testContext.GetAndClearDiagnosticMessages();
+                    string cleanupLogs = logListener.StandardOutput;
+                    string? cleanupTrace = logListener.DebugTrace;
+                    string cleanupErrorLogs = logListener.StandardError;
+
+                    var lastResult = notRunnableResult[notRunnableResult.Length - 1];
```
Evangelink: Are we sure we have at least 1 run? Otherwise we would have index out of range exception.

src/Adapter/MSTest.TestAdapter/Execution/UnitTestRunner.cs
```diff
@@ -200,33 +226,57 @@ internal UnitTestResult[] RunSingleTest(TestMethod testMethod, IDictionary<strin
         return result;
     }
 
-    private void RunClassCleanupIfEndOfClass(TestMethodInfo testMethodInfo, TestMethod testMethod, UnitTestResult[] results)
+    private void TryToRunCleanups(ITestContext testContext, TestMethodInfo testMethodInfo, TestMethod testMethod, UnitTestResult[] results)
```
Evangelink: I usually prefer to keep `TryXXX` prefix for method returning a bool (i.e. using the try pattern). If your idea is to suggest that it's not always running cleanup you could go with `RunCleanupsIfNeeded`.

src/Adapter/MSTest.TestAdapter/Execution/UnitTestRunner.cs
```diff
@@ -363,9 +413,9 @@ private class ClassCleanupManager
             _reflectHelper = reflectHelper ?? new ReflectHelper();
         }
 
-        public void MarkTestComplete(TestMethodInfo testMethodInfo, TestMethod testMethod, out bool shouldCleanup)
+        public void MarkTestComplete(TestMethodInfo testMethodInfo, TestMethod testMethod, out bool shouldClassCleanup, out bool shouldAssimplyCleanup)
```
Evangelink: Rename `shouldClassCleanup` with `shouldRunClassCleanup` and `shouldAssimplyCleanup` with `shouldRunAssemblyCleanup`.

test/E2ETests/Smoke.E2E.Tests/SuiteLifeCycleTests.cs
```diff
@@ -27,32 +27,16 @@ private void ValidateTestRunLifecycle(string targetFramework)
...
-        // Assembly and Class init/cleanup logs don't appear in the tests' results.
-        classCleanupCalledCount.Should().Be(0);
-        assemblyCleanupCalledCount.Should().Be(0);
-        assemblyInitCalledCount.Should().Be(0);
-        classInitCalledCount.Should().Be(0);
-
+        int numberOfLines;
         var caseClassCleanupWithCleanupBehaviorEndOfAssembly = RunEventsHandler.PassedTests.Single(x => x.TestCase.FullyQualifiedName.Contains("SuiteLifeCycleTestClass_ClassCleanupWithCleanupBehaviorEndOfAssembly"));
+        numberOfLines = caseClassCleanupWithCleanupBehaviorEndOfAssembly.Messages.Single().Text.Split('\n').Length;
+        Verify(numberOfLines == (targetFramework == "net6.0" ? 12 : 11)); // The number of the logs + 3 empety lines + 1 for the logs' header.
```
Evangelink:
```suggestion
        Verify(numberOfLines == (targetFramework == "net6.0" ? 12 : 11)); // The number of the logs + 3 empty lines + 1 for the logs' header.
```

Evangelink: Please fix the typo on the other instances too.

test/UnitTests/MSTestAdapter.UnitTests/Execution/TestExecutionManagerTests.cs
```diff
@@ -177,42 +177,6 @@ public void RunTestsForCancellationTokenCanceledSetToTrueShouldSendZeroResults()
         Verify(_frameworkHandle.TestCaseEndList.Count == 0);
     }
 
-    public void RunTestsShouldLogResultCleanupWarningsAsErrorsWhenTreatClassCleanupWarningsAsErrorsIsTrue()
```
Evangelink, line 180: Why is it removed?

test/E2ETests/Smoke.E2E.Tests/SuiteLifeCycleTests.cs
```diff
@@ -70,74 +56,96 @@ private void ValidateTestRunLifecycle(string targetFramework)
...
         caseClassInitializeWithInheritanceBehaviorBeforeEachDerivedClassAndClassCleanupWithInheritanceBehaviorNone.Messages.Single().Text.Should().Contain(
             targetFramework == "net6.0" ?
             """
-            Ctor was called
-            TestInitialize was called
-            TestMethod was called
-            TestCleanup was called
-            DisposeAsync was called
-            Dispose was called
+            Base ClassInitialize.InheritanceBehaviorBeforeEachDerivedClass was called
```
Evangelink: Having a full picture now, I think it'd be better to have a prefix property being string empty by default and can be overridden/set by children. So that when you are running project itself, there is no prefix (e.g. "Ctor was called") but when it's called in the context of a child class, the prefix "Base " is added.

test/E2ETests/Smoke.E2E.Tests/SuiteLifeCycleTests.cs
```diff
@@ -262,17 +262,17 @@ private void ValidateTestRunLifecycle(string targetFramework)
             Base TestCleanup was called
             Base DisposeAsync was called
             Base Dispose was called
-            Base ClassCleanup.InheritanceBehaviorBeforeEachDerivedClass was called
-            Current ClassCleanup was called
-            ClassCleanup.EndOfAssembly was called
+            Base ClassCleanup.InheritanceBehaviorNone was called
```
Evangelink: It seems that the order of execution changed, is it expect?

Evangelink: It looks like this is the only test for which calls order changed.

Evangelink (commented):
Overall LGTM. Could you please review the test `caseInheritClassWithCleanupInheritanceBehaviorNone_ParentTestMethod ` where the order of calls changed?

test/E2ETests/Smoke.E2E.Tests/SuiteLifeCycleTests.cs
```diff
@@ -27,348 +27,279 @@ private void ValidateTestRunLifecycle(string targetFramework)
...
-            TestInitialize was called
-            Derived class TestInitialize was called
-            TestMethod was called
-            Derived class TestCleanup was called
-            TestCleanup was called
-            Dispose was called
-            """);
+        caseInheritClassWithCleanupInheritanceBehaviorNone_ParentTestMethod.Messages.Single().Text.Should().Contain(
+            targetFramework == "net6.0"
+            ? """
```
Evangelink: So this test case is failing for net6.0. As far as I can see, messages are not produced in the same order.

Potential causes could be:
- Parallel execution working differently for netcore and netfx causing different methods to be called at different time
- Methods are not discovered in the same order
- Implementation is buggy

Evangelink (changes requested):
I think there is something fishy with the implementation (maybe how you find the test to attach cleanup results to) because it seems that the `ClassCleanup.InheritanceBehaviorBeforeEachDerivedClass` messages are not attached to the correct test.

src/Adapter/MSTest.TestAdapter/ObjectModel/UnitTestResult.cs
```diff
@@ -183,14 +183,14 @@ internal TestResult ToTestResult(TestCase testCase, DateTimeOffset startTime, Da
 
         if (!StringEx.IsNullOrEmpty(DebugTrace))
         {
-            string debugTraceMessagesInStdOut = string.Format(CultureInfo.InvariantCulture, "\n\n{0}\n{1}", Resource.DebugTraceBanner, DebugTrace);
+            string debugTraceMessagesInStdOut = string.Format(CultureInfo.CurrentCulture, "{2}{2}{0}{2}{1}", Resource.DebugTraceBanner, DebugTrace, Environment.NewLine);
```
Evangelink, line 186: Use OS line return to avoid mixing line return kinds.

test/E2ETests/Smoke.E2E.Tests/SuiteLifeCycleTests.cs
```diff
@@ -27,348 +27,319 @@ private void ValidateTestRunLifecycle(string targetFramework)
...
+            LifeCycleClassCleanup.ClassCleanup was called
+            LifeCycleClassInitializeNoneAndClassCleanupBeforeEachDerivedClass.ClassCleanup was called
+            LifeCycleDerivedClassInitializeAndCleanupNone.ClassCleanup was called
+            LifeCycleClassInitializeAndCleanupNone.ClassCleanup was called
+            LifeCycleDerivedClassInitializeAndCleanupBeforeEachDerivedClass.ClassCleanup was called
+            LifeCycleClassInitializeAndCleanupBeforeEachDerivedClass.ClassCleanup was called
+            LifeCycleClassInitializeAndCleanupBeforeEachDerivedClass.ClassCleanup was called
+            LifeCycleDerivedClassInitializeBeforeEachDerivedClassAndClassCleanupNone.ClassCleanup was called
+            LifeCycleClassInitializeBeforeEachDerivedClassAndClassCleanupNone.ClassCleanup was called
+            LifeCycleClassCleanupEndOfAssembly.ClassCleanup was called
```
Evangelink: @engyebrahim I have reworked the messages so that it's easier to validate that all tests are functionally correct. It also helped me to see that on netfx the order is the same while in netcore the order is different. We now need to identify if this is a bug of implementation or something linked to resolution order.

src/Adapter/MSTest.TestAdapter/Execution/UnitTestRunner.cs
```diff
@@ -134,16 +136,55 @@ internal UnitTestResult[] RunSingleTest(TestMethod testMethod, IDictionary<strin
...
+                bool shouldRunClassCleanup = false;
+                bool shouldRunClassAndAssemblyCleanup = false;
                 if (_classCleanupManager is null)
                 {
                     return notRunnableResult;
                 }
 
                 DebugEx.Assert(testMethodInfo is not null, "testMethodInfo should not be null.");
-                _classCleanupManager.MarkTestComplete(testMethodInfo, testMethod, out bool shouldRunClassCleanup);
+                _classCleanupManager.MarkTestComplete(testMethodInfo, testMethod, out shouldRunClassCleanup, out shouldRunClassAndAssemblyCleanup);
```
Evangelink: It seems that most of the code below is extremely similar to the one of `RunCleanupsIfNeeded`, could we merge both?

Evangelink (dismissed):

Evangelink (approved):

PR #1399, opened 2022-11-18
DeploymentItemAttribute only works for a test class with test method

### LINKED ISSUE #634: DeploymentItem doesn't work for pure test class without test methods
## Description
According to the documentation the DeploymentItem can 

> be specified on **test class** or test method ([src](https://docs.microsoft.com/en-us/dotnet/api/microsoft.visualstudio.testtools.unittesting.deploymentitemattribute?view=mstest-net-1.3.2))

Anyways, when applying the DeploymentItem to a TestClass which only contains a (valid) AssemblyInitialize- or AssemblyCleanup-Method, the specified resource won't be deployed.

This seems due to the fact that currently the deployment item is only evaluated for TestCase-items (see [src/Adapter/PlatformServices.Shared/netstandard1.3/Utilities/ns13DeploymentUtilityBase.cs](https://github.com/microsoft/testfx/blob/master/src/Adapter/PlatformServices.Shared/netstandard1.3/Utilities/ns13DeploymentUtilityBase.cs) line 52).

Either the documentation should be updated or the deployment item should be evaluated for any class regarding the existence of a test method.
Another solution could be enabling the DeploymentItem attribute to target assemblies.

## Steps to reproduce
```
    [DeploymentItem("Resource.bin")]
    [TestClass] //Note the TestClass-attribute
    public class Class1
    {
        [AssemblyInitialize]
        public static void AssemblyInit(TestContext context)
        {
            Assert.IsTrue(File.Exists("Resource.bin")); //Fails
        }
    }
```
## Expected behavior
Assert.IsTrue succeeds

## Actual behavior
Assert.IsTrue fails

##  Workaround
As a workaround  an empty bodied test method can be added. Then it works as expected:

```
    [DeploymentItem("Resource.bin")]
    [TestClass] //Note the TestClass-attribute
    public class Class1
    {
        [AssemblyInitialize]
        public static void AssemblyInit(TestContext context)
        {
            Assert.IsTrue(File.Exists("Resource.bin")); //Fails
        }
        [TestMethod]
        public void ForceDeployment() { }
    }
```

[AB#1681020](https://devdiv.visualstudio.com/0bdbc590-a062-4c3f-b0f6-9383f67865ee/_workitems/edit/1681020)

Evangelink:
Fixes #634

### DIFF
--- a/src/TestFramework/TestFramework.Extensions/Attributes/DeploymentItemAttribute.cs
+++ b/src/TestFramework/TestFramework.Extensions/Attributes/DeploymentItemAttribute.cs
@@ -11,6 +11,11 @@ namespace Microsoft.VisualStudio.TestTools.UnitTesting;
 /// Can have multiple instances of the attribute to specify more than one item.
 /// The item path can be absolute or relative, if relative, it is relative to RunConfig.RelativePathRoot.
 /// </summary>
+/// <remarks>
+/// If specified on a test class, the class needs to contain at least one test method. This means that the
+/// attribute cannot be combined with a test class that would contain only a AssemblyInitialize or ClassInitialize
+/// method.
+/// </remarks>
 /// <example>
 /// [DeploymentItem("file1.xml")]
 /// [DeploymentItem("file2.xml", "DataFiles")]


Evangelink:
@AbhitejJohn Do you know what's the process to have the official doc [learn.microsoft.com](https://learn.microsoft.com/en-us/dotnet/api/microsoft.visualstudio.testtools.unittesting.deploymentitemattribute?view=visualstudiosdk-2022&viewFallbackFrom=mstest-net-1.3.2) updated?

MarcoRossignoli (approved):

PR #1401, opened 2022-11-21
Propagate UI culture to appdomain

Evangelink:


drognanar (dismissed):

Evangelink:
@MarcoRossignoli I am having some issue with the setup of the test we discussed about so I will move forward with the PR as-is and add the extra test later on.

Issue #1336, opened 2022-10-13
Allow to override GetDisplayName method from DataRowAttribute


Evangelink:
> As far as I can see, there is a [DisplayName](https://github.com/microsoft/testfx/blob/main/src/TestFramework/TestFramework/Attributes/DataSource/DataRowAttribute.cs?rgh-link-date=2022-10-12T14%3A17%3A06Z#L59) property that is settable so you should be able to manually override the name.

DisplayName is settable, but that does not solve the problem imho. To generate DisplayName I need to get the input data (method and params) and I get those in GetDisplayName. 

I also most likely want to rely on the `base` implementation of GetDisplayName, and only add some additional info, not re-implement the whole method from scratch or hide it. So imho GetDisplayName should be virtual.

_Originally posted by @nohwnd in https://github.com/microsoft/testfx/issues/1180#issuecomment-1277155353_
      

[AB#1642313](https://devdiv.visualstudio.com/0bdbc590-a062-4c3f-b0f6-9383f67865ee/_workitems/edit/1642313)

inikulshin:
@Evangelink thanks for this improvement.

But did you managed to make overridden `GetDisplayName() `to be used by Visual Studio? I thought it was the main reason for making it `virtual`, but with latest (**17.7.4**) Visual Studio I still see all the lines the same:

![image](https://github.com/microsoft/testfx/assets/17042537/4e2af11b-1203-47f6-b4ab-2e2af898d835)

MSTest packages are **3.0.2**, custom DataRow implementation is the following:
```
    [AttributeUsage(AttributeTargets.Method, AllowMultiple = true)]
    public class DataRowAttribute : Microsoft.VisualStudio.TestTools.UnitTesting.DataRowAttribute
    {
        public DataRowAttribute() : base() { }
        public DataRowAttribute(object data1) : base(data1) { }
        public DataRowAttribute(object data1, params object[] moreData) : base(data1, moreData) { }

        public override string? GetDisplayName(MethodInfo methodInfo, object?[]? data)
        {
            return base.GetDisplayName(methodInfo,
                Array.ConvertAll(data, obj => (obj is Array ? "[" + string.Join(",", obj as Array) + "]" : obj)));
        }
    }
```

PR #1413, opened 2022-11-28
Allow to override GetDisplayName method from DataRowAttribute

engyebrahim:
fix: https://github.com/microsoft/testfx/issues/1336

### DIFF
--- a/test/E2ETests/TestAssets/DataRowTestProject/DataRowTests_OverrideGetDisplayName.cs
+++ b/test/E2ETests/TestAssets/DataRowTestProject/DataRowTests_OverrideGetDisplayName.cs
@@ -1,0 +1,34 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System.Diagnostics;
+using System.Reflection;
+
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+
+namespace DataRowAttributeTestProject;
+
+[TestClass]
+public class DataRowTests_OverrideGetDisplayName : DataRowAttribute
+{
+    [TestCategory("OverridedGetDiplayName")]
+    [DummyDataRowAttribute]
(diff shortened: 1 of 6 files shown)


Evangelink (commented):
As discussed offline, please add tests to this PR.

Also, could you please create another ticket to suggest implementing a feature where user could specify the name (and type) of a method doing the generation of the display name? Similarly to `DynamicDataDisplayName`.

test/UnitTests/TestFramework.UnitTests/Attributes/DataRowAttributeTests.cs
```diff
@@ -164,4 +178,22 @@ public void GetDisplayNameForMultipleArraysOfMultipleItems()
...
+        }
+
+        public override string GetDisplayName(MethodInfo methodInfo, object[] data)
+        {
+            if (!string.IsNullOrWhiteSpace(DisplayName))
+            {
+                return DisplayName;
+            }
+
+            return methodInfo.Name;
```
Evangelink: Please return a constant string instead (and update assertion) so that we don't get confused with a change of behavior that would still use method name.

Evangelink (commented):
See my previous comment.

test/E2ETests/TestAssets/DataRowTestProject/DataRowTests_OverrideGetDisplayName.cs
```diff
@@ -0,0 +1,34 @@
...
+
+using System.Diagnostics;
+using System.Reflection;
+
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+
+namespace DataRowAttributeTestProject;
+
+[TestClass]
+public class DataRowTests_OverrideGetDisplayName : DataRowAttribute
```
Evangelink:
```suggestion
public class DataRowTests_OverrideGetDisplayName
```

Evangelink (approved):

PR #1414, opened 2022-11-28
Unify DeploymentDirectory location across target frameworks

### LINKED ISSUE #1097: DeploymentDirectory is 'Out' folder using .NET Framework, but build folder using .NET 6
## Description
When running a unit test with MSTest under .NET Framework, the `TestContext.DeploymentDirectory` path is the 'Out' folder inside the test run directory, e.g:
```
C:\Repositories\MySolution\TestResults\Deploy_User yyyy-MM-dd HH_mm_ss\Out
```
When running the same test under .NET 6, the `TestContext.DeploymentDirectory` path is the build directory, e.g:
```
C:\Repositories\MySolution\MyUnitTestProject\bin\Debug\net6.0
```
I've looked around but this change in behavior between Framework & Core doesn't seem to be documented anywhere. In fact, [the documentation for the DeploymentDirectory property](https://docs.microsoft.com/en-us/dotnet/api/microsoft.visualstudio.testtools.unittesting.testcontext.deploymentdirectory?view=visualstudiosdk-2022) says it's "typically a subdirectory of TestRunDirectory" - which isn't the case under .NET 6.

I've looked at existing issues; #653 seems similar at first but I think that's actually related to `DeploymentItem` rather than `DeploymentDirectory` despite the title - whereas the issue I'm describing is present whether or not `DeploymentItem` is used.

Is this a bug or intended behavior?

The reason I'm asking is that I have existing tests which are dependent on this directory being isolated for each run.

## Steps to reproduce
Create a new unit test project in .NET 6. Add the following unit test and run it.
```csharp
[TestMethod]
public void TestDeploymentDirectory()
{
    Assert.IsTrue(TestContext.DeploymentDirectory.StartsWith(TestContext.TestRunDirectory));
}
```

## Expected behavior
The test passes, since the `DeploymentDirectory` is inside the `TestRunDirectory`. This is the current behavior if you change the target framework to `net480`.

## Actual behavior
The test fails.

## Environment
Windows 10, Visual Studio 2022 v17.1.1
MSTest.TestAdapter v2.2.7
MSTest.TestFramework v2.2.7
Microsoft.NET.Test.Sdk 16.11.0
coverlet.collector 3.1.0

Evangelink:
Fixes #1097


nohwnd:
This change is setting the deploymentDirectory on testContext to be the same on .NET as it was on .NET Framework.

Issue #1149, opened 2022-07-14
TestId no longer constant across test runs


Cjewett:
Problem:

Trx specification seems to have changed significantly from 2.2.3 to 2.2.4+. As a result it makes it difficult (impossible?) to track historic results, map child tests to parents, and muddies the water for all UnitTestResults in the Trx file when using DataRow and DynamicData attributes. 

In 2.2.3 all DataRow and DynamicData tests would show up as an InnerResult for a parent. The result would contain a testId that mapped to the parent's testId that it is nested under. This made sense because you could see the overall test result by viewing the parent and then look at each individual test result when looking at the child tests. After upgrading to 2.2.4 or later it doesn't seem like any of the consistency that Trx files had are valid anymore. 

This has negative side effects depending on the CI/CD platform you use. In our example we use Bamboo which has always displayed parent results. Now that we have upgraded it shows each individual DataRow and DynamicData test result now too. While that might be okay for teams using default tooling that doesn't require any custom plugins or features to maintain a code base these new changes have made it difficult for those of us who need extra feature sets to keep our teams sailing smoothly.

Our team has a monolithic repository. Unit, integration, and functional tests can all be found in our C# test projects that run as part of our continuous integration process. As a result we have had to build tooling to manage and maintain tests. To do this we process Trx files, store them in our own data store, and run analysis on them. Some reasons why are listed below. 
- Does the test have the traits of a flaky test? If so, quarantine it in Bamboo so developers are not blocked from merging.
- Is this test too slow consistently? If so, notify users that there is a reproducible performance issue. 
- Is this test randomly slow? If so, notify users that there is a hard to find performance issue. 
- Does this test fail the first time but pass on the second attempt? If so, notify users that their test is unreliable. 
- Does the test only fail between certain hours of the day? If so, notify the user.

You can see where that is going.

To create such a system we have had to rely on constants between test runs that lets us map results historically. For the last 8 years the test id has always been that constant. You couldn't even open a Trx file in Visual Studio if there were duplicate test ids being used. I haven't tested that in a while though.

If we want to quarantine a test we could quarantine the parent and that would impact the children. Under this Trx structure in 2.2.4+ that is no longer the case. While it would be fine to quarantine based on the ClassName+TestName we are noticing that the ClassName+TestName is not unique across all test results either. Example of that below:

```
<?xml version="1.0" encoding="utf-8"?>
[...]

PR #1418, opened 2022-11-29
Ensure test ID strategy is saved as property

Evangelink:
In addition to the fix on Test Platform side, I have realized that we don't store the ID strategy property so we always restore it as the new strategy which is causing some issue for inner results (in case of Legacy strategy).

Relates to #1149

### DIFF
--- a/src/Adapter/MSTest.TestAdapter/ObjectModel/UnitTestElement.cs
+++ b/src/Adapter/MSTest.TestAdapter/ObjectModel/UnitTestElement.cs
@@ -230,6 +230,8 @@ internal class UnitTestElement
 
     private void SetTestCaseId(TestCase testCase, string testFullName)
     {
+        testCase.SetPropertyValue(Constants.TestIdGenerationStrategyProperty, (int)TestMethod.TestIdGenerationStrategy);
+
         switch (TestMethod.TestIdGenerationStrategy)
         {
 #pragma warning disable CS0618 // Type or member is obsolete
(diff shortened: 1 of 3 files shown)


src/Adapter/MSTest.TestAdapter/ObjectModel/UnitTestElement.cs
```diff
@@ -230,6 +230,8 @@ internal TestCase ToTestCase()
 
     private void SetTestCaseId(TestCase testCase, string testFullName)
     {
+        testCase.SetPropertyValue(Constants.TestIdGenerationStrategyProperty, (int)TestMethod.TestIdGenerationStrategy);
```
Evangelink, line 233: Saved and restored as int. I could use string which would be more human readable but int feels safer for renames.

Note that we are simply doing round-trip between discovery and run so it's probably as safe to be using string.

MarcoRossignoli: can be saved somewhere and reused in the future? You know what I mean...could this break other usages? Like breaking change in the enum?

MarcoRossignoli (approved):

src/Adapter/MSTest.TestAdapter/ObjectModel/UnitTestElement.cs
Evangelink: After some investigation, I will keep this as `int` for now as it matches what's done for the other enums. Also, it seems that there are some issues with using `string`, I will try to investigate them.

PR #1446, opened 2022-12-09
Fix regressions with DataRow supported arguments

### LINKED ISSUE #1445: Regression in MSTest v3 DataRow
## Describe the bug

The example below was working up to MSTest v3 and now is raising some compiler warning `CS0182: An attribute argument must be a constant expression, typeof expression or array creation expression of an attribute parameter type`.

## Steps To Reproduce

```csharp
[TestMethod]
[DataRow(new string[] { "" })] // Compiler warning
public void StringArray(string[] lines)
{
}
```

[AB#1697642](https://devdiv.visualstudio.com/0bdbc590-a062-4c3f-b0f6-9383f67865ee/_workitems/edit/1697642)

Evangelink:
Fixes #1445


src/TestFramework/TestFramework/Attributes/DataSource/DataRowAttribute.cs
```diff
@@ -15,25 +15,279 @@ namespace Microsoft.VisualStudio.TestTools.UnitTesting;
 [AttributeUsage(AttributeTargets.Method, AllowMultiple = true)]
 public class DataRowAttribute : Attribute, ITestDataSource
 {
+    /// <summary>
+    /// Initializes a new instance of the <see cref="DataRowAttribute"/> class.
+    /// </summary>
+    public DataRowAttribute()
```
Evangelink, line 21: IMO we should **not** keep this one because that's equivalent to not having any data (we expect parameterless method). WDYT?

jakubch1 (dismissed):

src/TestFramework/TestFramework/Attributes/DataSource/DataRowAttribute.cs
jakubch1: Are you confident enough that this will not break people?

Evangelink: If they were using this case, it would break them for sure but the case makes no functional sense.

I mean if you write:

```csharp
[TestMethod]
[DataRow()]
public void A() {}
```

Why not simply doing:

```csharp
[TestMethod]
public void A() {}
```

jakubch1: Any such case in github?

Evangelink: Looking at GH, it seems that some people are using it (mainly us but also some others):
https://grep.app/search?q=%5BDataRow%28%29%5D

Evangelink: Otherwise we keep and mark as `Obsolete` so we can remove in v4. Sounds like best course of action.

jakubch1 (approved):

MarcoRossignoli (approved):

PR #1452, opened 2022-12-12
Allow most APIs to accept nullable values or arguments

### LINKED ISSUE #1451: MSTest v3 nullability on assertions should be more permissive
## Describe the bug

Most of the assertions APIs in MSTest v2 were supporting null values and would fail in case of null without forcing `Assert.IsNotNull` before. In MSTest v3, most of the APIs are expecting non-nullable entries.

## Steps To Reproduce

```csharp
StringAssert.StartsWith(nullableString, "something"); // potential null warning
```

Evangelink:
Fix #1451


Evangelink:
Closed as won't ship as it's bending too much the nullability.

PR #1467, opened 2022-12-14
Allow most APIs to accept nullable values or arguments

### LINKED ISSUE #1451: MSTest v3 nullability on assertions should be more permissive

Evangelink:
Fixes #1451


src/TestFramework/TestFramework/Assertions/Assert.AreEqual.cs
```diff
@@ -31,7 +32,8 @@ public sealed partial class Assert
     /// <exception cref="AssertFailedException">
     /// Thrown if <paramref name="expected"/> is not equal to <paramref name="actual"/>.
     /// </exception>
-    public static void AreEqual<T>(T? expected, T? actual)
+    public static void AreEqual<T>(
+        [NotNullIfNotNull(nameof(actual))] T? expected, [NotNullIfNotNull(nameof(expected))] T? actual)
```
Evangelink: `NotNullIfNotNull` is just a compiler hint to say the post condition of this method is that if one of the 2 params is not null then the other isn't null. 

This is valid for `AreEqual` methods as two objects are equal only if they are both `null` or both not null and equal. For all `AreNotEqual` we cannot make the same guarantee .

src/TestFramework/TestFramework/Assertions/Assert.AreEqual.cs
```diff
@@ -400,9 +412,7 @@ public static void AreNotEqual<T>(T? notExpected, T? actual, IEqualityComparer<T
     /// <paramref name="actual"/>.
     /// </exception>
     public static void AreEqual(float expected, float actual, float delta)
-    {
-        AreEqual(expected, actual, delta, string.Empty, null);
-    }
+        => AreEqual(expected, actual, delta, string.Empty, null);
```
Evangelink, line 406: Cosmetic change but as I was reviewing ALL assert APIs, I thought I could go for it too.

src/TestFramework/TestFramework/Assertions/Assert.AreEqual.cs
```diff
@@ -1266,15 +1241,13 @@ public static void AreEqual(string? expected, string? actual, bool ignoreCase, s
     /// indicates a case-insensitive comparison.)
     /// </param>
     /// <param name="culture">
-    /// A CultureInfo object that supplies culture-specific comparison information.
+    /// A CultureInfo object that supplies culture-specific comparison information. If culture is null, the current culture is used.
```
Evangelink, line 1243: This is now matching the runtime behaviour and it's made explicit to users.

src/TestFramework/TestFramework/Assertions/Assert.AreEqual.cs
```diff
@@ -1190,9 +1170,7 @@ public static void AreNotEqual(double notExpected, double actual, double delta,
     /// Thrown if <paramref name="expected"/> is not equal to <paramref name="actual"/>.
     /// </exception>
     public static void AreEqual(string? expected, string? actual, bool ignoreCase)
```
Evangelink, line 1171: I missed these `string` APIs where I could also put the `NotNullIfNotNull` attribute.

src/TestFramework/TestFramework/Assertions/Assert.AreSame.cs
```diff
@@ -30,10 +31,8 @@ public sealed partial class Assert
     /// Thrown if <paramref name="expected"/> does not refer to the same object
     /// as <paramref name="actual"/>.
     /// </exception>
-    public static void AreSame<T>(T? expected, T? actual)
-    {
-        AreSame(expected, actual, string.Empty, null);
-    }
+    public static void AreSame<T>([NotNullIfNotNull(nameof(actual))] T? expected, [NotNullIfNotNull(nameof(expected))] T? actual)
```
Evangelink: `AreSame` is like "are references equal" so it's ok to say that if one parameter is not null the other will always be non-null.

src/TestFramework/TestFramework/Assertions/Assert.cs
```diff
@@ -131,6 +129,7 @@ internal static void CheckParameterNotNull(object? param, string assertionName,
     /// The converted string.
     /// </returns>
     [SuppressMessage("ReSharper", "RedundantToStringCall", Justification = "We are ensuring ToString() isn't overloaded in a way to misbehave")]
+    [return: NotNull]
```
Evangelink, line 132: This means that the returned value is always non-null

src/TestFramework/TestFramework/Assertions/CollectionAssert.cs
```diff
@@ -49,13 +50,10 @@ private CollectionAssert()
     /// The element that is expected to be in the collection.
     /// </param>
     /// <exception cref="AssertFailedException">
-    /// Thrown if <paramref name="element"/> is not found in
-    /// <paramref name="collection"/>.
+    /// <paramref name="collection"/> is null, or <paramref name="collection"/> does not contain element <paramref name="element"/>.
```
Evangelink: Update explanation message about cases leading to exception being thrown.

MarcoRossignoli (approved):

MarcoRossignoli:
:shipit:

jakubch1 (approved):

src/TestFramework/TestFramework/Assertions/Assert.AreEqual.cs
Evangelink: Reverting as compiler team implemented this feature only for "output" parameters (`out` and `ref`) as opposed to some other post conditions.

jakubch1 (approved):

MarcoRossignoli (approved):

PR #1470, opened 2022-12-15
Fix class/assembly cleanups log collect and attachment

### LINKED ISSUE #1448: Trace.WriteLine in AssemblyCleanup is not printed in MSTest v3.0.0
## Describe the bug

Methods having the **AssemblyCleanupAttribute** are not called.

Example:
```
[TestClass]
public sealed class DummyUnitTest
{
    [AssemblyInitialize]
    public static void AssemblyInit(TestContext testContext)
    {
        Trace.WriteLine(nameof(AssemblyInit));
    }

    [AssemblyCleanup]
    public static void AssemblblyCleanup()
    {
        Trace.WriteLine(nameof(AssemblblyCleanup));
    }

    [TestMethod]
    public void DummyTest()
    {
        Trace.WriteLine(nameof(this.DummyTest));
    }
}
```

Expected output (by MSTest 2.2.10):
```
...
AssemblyInit
DummyTest
AssemblblyCleanup
...
```

Actual output (by MSTest 3.0.0)
```
...
AssemblyInit
DummyTest
...
```

## Steps To Reproduce

Run the unit test as described above with MSTest v3.0.0.

## Expected behavior

The AssemblblyCleanup method should be called.

## Actual behavior

The AssemblblyCleanup method is not called.

[AB#1699540](https://devdiv.visualstudio.com/0bdbc590-a062-4c3f-b0f6-9383f67865ee/_workitems/edit/1699540)

Evangelink:
- Ensures Trace/Debug/Console logs calls inside `ClassCleanup `and `AssemblyCleanup `methods are correctly collected.
- Ensures that failed `ClassCleanup `or `AssemblyCleanup` is attached to last test run

Fixes #1448

### DIFF
--- a/src/Adapter/MSTest.TestAdapter/Execution/TestAssemblyInfo.cs
+++ b/src/Adapter/MSTest.TestAdapter/Execution/TestAssemblyInfo.cs
@@ -234,4 +234,58 @@ public class TestAssemblyInfo
             }
         }
     }
+
+    /// <summary>
+    /// Calls the assembly cleanup method in a thread-safe.
+    /// </summary>
+    /// <remarks>
+    /// It is a replacement for RunAssemblyCleanup but as we are in a bug-fix version, we do not want to touch
+    /// public API and so we introduced this method.
+    /// </remarks>
+    internal void ExecuteAssemblyCleanup()
+    {
+        if (AssemblyCleanupMethod == null)
+        {
+            return;
+        }
+
+        lock (_assemblyInfoExecuteSyncObject)
+        {
+            try
+            {
+                AssemblyCleanupMethod.InvokeAsSynchronousTask(null);
+            }
+            catch (Exception ex)
+            {
+                var realException = ex.InnerException ?? ex;
+
+                string errorMessage;
+
+                // special case AssertFailedException to trim off part of the stack trace
+                if (realException is AssertFailedException or AssertInconclusiveException)
+                {
+                    errorMessage = realException.Message;
+                }
+                else
+                {
+                    errorMessage = StackTraceHelper.GetExceptionMessage(realException);
+                }
+
+                var exceptionStackTraceInfo = StackTraceHelper.GetStackTraceInformation(realException);
+                DebugEx.Assert(AssemblyCleanupMethod.DeclaringType?.Name is not null, "AssemblyCleanupMethod.DeclaringType.Name is null");
+
+                throw new TestFailedException(
+                    UnitTestOutcome.Failed,
+                    string.Format(
+                        CultureInfo.CurrentCulture,
+                        Resource.UTA_AssemblyCleanupMethodWasUnsuccesful,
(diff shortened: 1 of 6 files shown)


src/Adapter/MSTest.TestAdapter/Execution/TestAssemblyInfo.cs
```diff
@@ -234,4 +234,58 @@ public void RunAssemblyInitialize(TestContext testContext)
...
     }
+
+    /// <summary>
+    /// Calls the assembly cleanup method in a thread-safe.
+    /// </summary>
+    /// <remarks>
+    /// It is a replacement for RunAssemblyCleanup but as we are in a bug-fix version, we do not want to touch
+    /// public API and so we introduced this method.
+    /// </remarks>
+    internal void ExecuteAssemblyCleanup()
```
Evangelink, line 245: Duplicates of the method above expect that we throw instead of returning string message.

src/Adapter/MSTest.TestAdapter/Execution/TestClassInfo.cs
```diff
@@ -410,4 +410,74 @@ public void RunClassInitialize(TestContext testContext)
...
     }
+
+    /// <summary>
+    /// Execute current and base class cleanups.
+    /// </summary>
+    /// <remarks>
+    /// This is a replacement for RunClassCleanup but as we are on a bug fix version, we do not want to change
+    /// the public API, hence this method.
+    /// </remarks>
+    internal void ExecuteClassCleanup()
```
Evangelink, line 421: Duplicates of the method above except that we throw exception instead of either throwing or returning error message.

src/Adapter/MSTest.TestAdapter/Execution/TestAssemblyInfo.cs
```diff
@@ -234,4 +234,58 @@ public void RunAssemblyInitialize(TestContext testContext)
...
+    {
+        if (AssemblyCleanupMethod == null)
```
MarcoRossignoli, line 247:
```suggestion
        if (AssemblyCleanupMethod is null)
```

MarcoRossignoli, line 247: should you double check inside the lock?

MarcoRossignoli (approved):
a comment but lgtm

src/Adapter/MSTest.TestAdapter/Execution/TestAssemblyInfo.cs
MarcoRossignoli: I see it's not the style....

Evangelink: I will keep as-is as I am only duplicating the existing method but I will keep a note to try to improve concurrency on these classes.

Issue #1377, opened 2022-11-11
Update TestContext.Properties return type to IDictionary<string, object>


Evangelink:
## Background and Motivation

Currently, the `TestContext.Properties` property returns `IDictionary` for compatibility reason. This is forcing users to rely on old-aged API where they could be using API coming from `Dictionary<string, object>`.

## Proposed Feature

Replace `IDictionary` with `IDictionary<string, object>` .

[AB#1707845](https://devdiv.visualstudio.com/0bdbc590-a062-4c3f-b0f6-9383f67865ee/_workitems/edit/1707845)

Youssef1313:
@Evangelink I'm re-opening as the PR that closed this issue was reverted

PR #1481, opened 2022-12-19
Update TestContext.Properties return type to IDictionary<string, object?>

engyebrahim:
fix: https://github.com/microsoft/testfx/issues/1377


src/Adapter/MSTestAdapter.PlatformServices/Services/TestContextImplementation.cs
```diff
@@ -116,7 +116,7 @@ public TestContextImplementation(ITestMethod testMethod, StringWriter stringWrit
 #endif
 
     /// <inheritdoc/>
-    public override IDictionary Properties => (IDictionary)_properties;
+    public override Dictionary<string, object?> Properties => _properties;
```
Evangelink: I would get rid of the field and replace this property with a readonly property.

Evangelink (commented):
You can simplify furthermore the `GetProperty` method of `TestContext`.

@MarcoRossignoli Do you think we should go with `IDictionary` over `Dictionary`?

MarcoRossignoli:
> @MarcoRossignoli Do you think we should go with IDictionary over Dictionary?

I would use always abstraction if possible so yes.

src/TestFramework/TestFramework.Extensions/TestContext.cs
```diff
@@ -197,7 +197,7 @@ public abstract class TestContext
     {
         DebugEx.Assert(Properties is not null, "Properties is null");
 #if WINDOWS_UWP || WIN_UI
-        if (!((IDictionary<string, object>)Properties).TryGetValue(name, out object? propertyValue))
+        if (!Properties.TryGetValue(name, out object? propertyValue))
```
Evangelink, line 199: You can merge all the #IF/#else part to keep only the `TryGetValue` part.

Evangelink (approved):

Issue #1284, opened 2022-09-27
Convert MSTest to async pattern


Evangelink:
## Description

Currently MSTest is based on sync pattern on all the code and we fake async behavior through `.Wait()`, `.GetAwaiter().GetResult()` or `Task.Run`. It would be better to have MSTest fully async.

Note that on the adapter to Test Platform level, our entry points are not async and it won't be possible to convert these. It's ok as we can consider these methods as our main.

SimonCropp:
@Evangelink i dont think this is possible to properly fix without changes to `ITestExecutor`. since we would still need to use `.GetAwaiter().GetResult()` in `MSTestExecutor`

Evangelink:
We can't "fully" fix but my idea is that everything under the test platform interface is done "async". We know that the implementation detail of Test Platform allows us to have a "single thread" so that's ok.

SimonCropp:
wont that still have the possibility of deadlocks?

Evangelink:
It will help solve most of them but for sure it won't solve everything because of some implementation choices that were made. I am still ramping up in the knowledge of the codebase and I am often surprised by some things that were done (e.g. using `TaskScheduler.Default` and expecting to have a different thread for each task while there is no guarantee from runtime - yes most of the time that's quite true but that's not the contract).

SimonCropp:
forgive my ignorance. but why cant ITestExecutor be changed in a new major? or obsoleted (with warning) and IAsyncTestExecutor added?

Evangelink:
Technically it's possible but [Test Platform](https://github.com/microsoft/vstest) is similar to mstest and far away from supporting async/await. The platform is in stabilization phase at the moment so it'd be hard to push for this change there.

Evangelink:
> The test runner will then just look for IAsyncTestMethod and call ExecuteAsync without the 'wait' call. This allows anyone to switch to it, and also any custom TestMethodAttributes can implement the interface. This could then be achieved in 3.x without breaking.

@dotMorten thanks for the suggestion! It would definitely help with some part but I feel there might still be some issues as we would need to flow the async/await up to the "entry point" to be sure to have a correct behavior. To do so, we would need to change some public API (honestly I don't know if they are used outside or not) hence the need to wait for v4.

I'll try to make a test to list exactly what would be breaking because maybe we can apply similar pattern in multiple places.

dotMorten:
[...]

PR #1527, opened 2023-01-02
try fix some deadlocks

SimonCropp:
help to contribute to https://github.com/microsoft/testfx/issues/1284.


Evangelink:
FYI this will help to contribute to #1284.

SimonCropp:
@Evangelink is there a way to update the api test result to match the api changes?

SimonCropp:
@Evangelink ok this is ready for a review

Evangelink (commented):
First pass LGTM. 

I will need to review the changes to public API to see if it's used on GitHub, and I also want to have a more in-depth look at some of the semaphore changes.

Side question, do you have any example of scenario where you hit the deadlock? Just trying to see if we could improve our tests (which isn't great at the moment).

SimonCropp:
> Side question, do you have any example of scenario where you hit the deadlock? Just trying to see if we could improve our tests (which isn't great at the moment). 

a consumer of Verify.MsTest hit it here https://github.com/VerifyTests/Verify/issues/751 and my guess was it is related to MSTests use of getawaiter, but i cant be sure. 

So no i dont think i could produce a test which would reliably reproduce a deadlock

src/Adapter/MSTest.TestAdapter/Execution/TestClassInfo.cs
```diff
@@ -21,7 +23,7 @@ namespace Microsoft.VisualStudio.TestPlatform.MSTest.TestAdapter.Execution;
 /// </summary>
 public class TestClassInfo
 {
-    private readonly object _testClassExecuteSyncObject;
+    private SemaphoreSlim _testClassExecuteSemaphore;
```
Evangelink: Can be made `readonly`

```suggestion
    private readonly SemaphoreSlim _testClassExecuteSemaphore;
```

src/TestFramework/TestFramework/Attributes/TestMethod/TestMethodAttribute.cs
```diff
@@ -41,8 +42,8 @@ public TestMethodAttribute(string? displayName)
     /// <param name="testMethod">The test method to execute.</param>
     /// <returns>An array of TestResult objects that represent the outcome(s) of the test.</returns>
     /// <remarks>Extensions can override this method to customize running a TestMethod.</remarks>
-    public virtual TestResult[] Execute(ITestMethod testMethod)
+    public virtual async Task<TestResult[]> Execute(ITestMethod testMethod)
```
Evangelink, line 45: I need to think about this one, as far as I can see it's quite often overridden. I am not sure if that's better to break or offer the two methods `Execute` and `ExecuteAsync` and obsolete the former.

src/TestFramework/TestFramework/Interfaces/ITestMethod.cs
```diff
@@ -57,7 +58,7 @@ public interface ITestMethod
     /// <remarks>
     /// This call handles asynchronous test methods as well.
     /// </remarks>
-    TestResult Invoke(object[]? arguments);
+    Task<TestResult> Invoke(object[]? arguments);
```
Evangelink, line 61: Same remark here, it's also used in a few places.

src/Adapter/MSTest.TestAdapter/Execution/TestClassInfo.cs
```diff
@@ -21,7 +23,7 @@ namespace Microsoft.VisualStudio.TestPlatform.MSTest.TestAdapter.Execution;
 /// </summary>
 public class TestClassInfo
```
Evangelink, line 24: As far as I can see on GitHub and internal repo this is not used outside of MSTest so changes to public API are fine (I will do a follow-up PR changing accessibility where possible to reduce un-needed/wanted public surface).

src/Adapter/MSTest.TestAdapter/Execution/TestExecutionManager.cs
```diff
@@ -320,36 +313,33 @@ private void ExecuteTestsInSource(IEnumerable<TestCase> tests, IRunContext? runC
 
                 for (int i = 0; i < parallelWorkers; i++)
                 {
-                    tasks.Add(_taskFactory(() =>
```
Evangelink, line 323: I need to look more into this part because it's changing the logic (I am not sure if that's good or not).

SimonCropp:
rebased

src/Adapter/MSTest.TestAdapter/Execution/TestExecutionManager.cs
SimonCropp: yeah given i cant compile (and hence cant run locally) i was guessing as to the correct new code.

happy to take direction from you

Evangelink:
@SimonCropp I am curious as to what issues you are facing with building, would you mind sharing the errors?

src/Adapter/MSTest.TestAdapter/Execution/TestExecutionManager.cs
Evangelink: Probably better to keep the old behaviour in here, we are missing tests to be 100% sure of what was the real intent here.

Evangelink:
@SimonCropp I am starting to think it's probably better to go full breaking change for v4. It's likely to break users extending MSTest but it feels like lots of extra work and increased likelihood of deadlock to keep the `GetAwaiter` option.

Issue #1320, opened 2017-12-11
Add documentation for [DeploymentItem] attribute.


jayaranigarg:
Bunch of people facing issues while using this attribute. https://github.com/Microsoft/testfx/issues/295

[AB#1735369](https://devdiv.visualstudio.com/0bdbc590-a062-4c3f-b0f6-9383f67865ee/_workitems/edit/1735369)

Issue #676, opened 2020-01-14
Deployment Items don't work using runsettings file


NetanelMosheCohen:
I'm trying to migrate our automation project - Selenium + C# +MSTest V1 to MSTest V2.
Until now I worked with testsettings file and used DeploymentItem to deploy few files to the output folder. 
But now I have to use runsettings file and the deployment items option is not working.
I tried to use legacy settings but it still doesn't work.
I searched all over the internet for something which describes how to do this but nothing has worked.

Attached testsettings (that managed to deploy items) and runsettings files (that did NOT manage to deploy items).

Is there something missing in my runsettings file? Can someone please provide me with a runsettings file which was tested and managed to actually deploy items?

Thanks

## Environment
.NET Framewok - 4.6.1
Programming language - C#
Test Adapter and version - MSTest V2 (latest)
OS - Win 10

[RapNet.zip](https://github.com/microsoft/testfx/files/4058268/RapNet.zip)

N-Olbert:
LegacySettingsd arent supported in MSTestV2..
https://github.com/Microsoft/testfx/issues/557#issuecomment-460155882

Maybe checkout the *DeployTestSourceDependencies*-Attribute or use plain DeploymentItem-Attributes.

I stumbled accross this myself a year ago. The documentation is very poor here.

kavyashreer23:
Hi,
Does changing to .runsettings from .testsettings file, support remote deployment in Test Controller and Agents with MSTestV2 ?
How to use the below properties in .runsettings file. Can anyone please help on this?

_<RemoteController name="Test Controller name"> 
<AgentRule name="All agents">
 </AgentRule>_

engyebrahim:
it's related to this PR too https://github.com/microsoft/testfx/pull/1571 as useing legacy settings aren't supported anymore inside the testsettings

PR #1581, opened 2023-02-06
Add documentation for [DeploymentItem] attribute

engyebrahim:
fix: https://github.com/microsoft/testfx/issues/1320 and https://github.com/microsoft/testfx/issues/676


Evangelink:
@engyebrahim I don't understand how this PR solves #676. My understanding is that people where using `.testsettings` (e.g. legacy configuration file) which isn't supported by MSTest. I feel like the issue should be simply closed.

src/TestFramework/TestFramework.Extensions/Attributes/DeploymentItemAttribute.cs
```diff
@@ -6,7 +6,8 @@
 namespace Microsoft.VisualStudio.TestTools.UnitTesting;
 
 /// <summary>
-/// Used to specify deployment item (file or directory) for per-test deployment.
```
Evangelink, line 9: It's important to keep the `per-test deployment` mention.

src/TestFramework/TestFramework.Extensions/Attributes/DeploymentItemAttribute.cs
```diff
@@ -6,7 +6,8 @@
...
+/// Used to specify deployment item (file or directory) for copying files or folders specified as deployment items to the deployment directory.
+/// Deployment directory is where all the deployment items are present along with TestSource dll.
```
Evangelink, line 10: We should probably add a link to the corresponding property of the testcontext object.

Evangelink (approved):

Evangelink (approved):

Issue #727, opened 2020-08-11
No 'AddResultFile' implementation in TestContext on UWP test project


dotMorten:
## Description

UWP doesn't have the option to add a result file. I get that the storage APIs can be limited here and are sandboxed, but the test runner is more than capable of getting the files from within the application and app data folder, so this really shouldn't be an issue. It's also possible to allow UWP apps to read/write folders outside the application folder by giving the the user `ALL APPLICATION PACKAGES` read/write access to said folder, so all of this should be possible to use provided a proper setup in UWP - as an alternative we can just create our own post-build scripts to go pick up the files, but without the ability to add the test result file entry we wouldn't know which file belongs to which test.

## Steps to reproduce
Create UWP Test project. Inspect TestContext and the missing APIs

![image](https://user-images.githubusercontent.com/1378165/89929742-84698b00-dbbe-11ea-9a2b-32f0bec15941.png)

## Expected behavior
Ability to add result files to the generated TRX report.

## Actual behavior
> What is the behavior observed?

## Environment
VS2019.7

[AB#1743231](https://devdiv.visualstudio.com/0bdbc590-a062-4c3f-b0f6-9383f67865ee/_workitems/edit/1743231)

nohwnd:
@MSLukeWest Luke are you still working on the UWP runner, should this be possible?

MSLukeWest:
@nohwnd It looks like that wouldn't be done in the UWP provider, it would be done here in https://github.com/microsoft/testfx/tree/main/src/Adapter/PlatformServices.Universal.  It looks like right now it's context implementation comes from https://github.com/microsoft/testfx/blob/main/src/Adapter/PlatformServices.Shared/netstandard1.0/Services/ns10TestContextImplementation.cs

Evangelink:
I don't see any constraint on MSTest side to implement this feature but I am not sure if it would be handled by the specific Test Platform of UWP. My understanding is that there is a specific "freezed" version of TP for UWP.

@MSLukeWest could you confirm that if this is added to the `UnitTestResult` for UWP, it would be handled correctly on TP side?

MSLukeWest:
@Evangelink For the most part UWP will run against the latest test platform binaries, so I think this would work.  If you tell me the name of the assembly that would contain the fix I can double check this for you.

Evangelink:
@engyebrahim Could you please create a PR for this issue and then provide the MSTest package to @MSLukeWest so he can confirm that's working fine?

[...]

PR #1593, opened 2023-02-14
Enable TestContext.AddResultFile API for WinUI

engyebrahim:
fix: https://github.com/microsoft/testfx/issues/727 and https://github.com/microsoft/testfx/issues/1594


engyebrahim:
Hi @MSLukeWest, Can you please check the change using UWP?

MSLukeWest (approved):
Please make sure to let me know when this gets inserted into VS so I can do some additional testing.

MSLukeWest (approved):

Evangelink:
Ping @engyebrahim

Evangelink (approved):
As per my tests this seems to be working only for WinUI. I suspect that UWP provider is ignoring this info we are providing.

Please double check if you see the same behaviour. Approving now so you can merge when you are done with testing.

PR #1604, opened 2023-02-22
Call verify-nupkgs.ps1 at the end of the pack step

Evangelink:
cc @nohwnd


eng/AfterSolutionBuild.targets
```diff
@@ -0,0 +1,7 @@
+<Project>
+
+  <Target Name="_VerifyNuGetPackages" AfterTargets="Pack" Condition="'$(DotNetBuildFromSource)' != 'true'">
```
Evangelink:
```suggestion
  <Target Name="_VerifyNuGetPackages" AfterTargets="Pack" Condition=" '$(OS)' == 'Windows_NT' ">
```

Issue #1645, opened 2023-04-28
[MSTest][Discovery] Unable to load types from the test source ... Could not load file or assembly


eleshoe:
- Windows 10 Enterprise x64 
- Visual Studio Enterprise 2022 (64 bit) Version 17.5.4
- DemoProject.dll: Project under test targeting .Net standard 2.0 
- DemoTestBase.dll: Test helper project targeting .Net standard 2.0
- DemoUnitTestProject.dll: Unit test project targeting .Net 6 and .Net framework 4.7.2. Having 2 unit test classes: UnitTest1.cs, UnitTest2.cs, with UnitTest1 inhering from DemoTestBase, but not UnitTest2.
- These 3 projects are in the same solution.
- The unit test project is referencing the other 2 projects using project reference.
- Using a .runsettings file to resolve assembly directory path.

Assembly output structure
```
C:\MyDev\Sources\bin\DemoProject.dll
C:\MyDev\Sources\bin\DemoTestBase.dll
C:\MyDev\Sources\unittest\DemoUnitTestProject.dll (this is .Net framework 4.7.2)
C:\MyDev\Sources\unittest\net6.0\DemoUnitTestProject.dll (this is .Net 6)
```

When I do a Clean and Rebuild on the unit test project, I see below error message in the Tests pane (with logging level = Trace).
```
[Warning] [MSTest][Discovery][C:\MyDev\Sources\unittest\net6.0\DemoUnitTestProject.dll] Unable to load types from the test source 'C:\MyDev\Sources\unittest\net6.0\DemoUnitTestProject.dll'. Some or all of the tests in this source may not be discovered.
Error: System.IO.FileNotFoundException: Could not load file or assembly 'DemoTestBase, Version=1.0.0.0, Culture=neutral, PublicKeyToken=null'. The system cannot find the file specified.
```

This is the complete trace
```
[28/04/2023 11:56:07.910 AM]  Interrupt: Enqueueing WaitForBuildOperation
[28/04/2023 11:56:07.910 AM]  Enqueue operation 'WaitForBuildOperation', hashcode:42412516 
[28/04/2023 11:56:07.910 AM]  Operation left in the the queue: 1
[28/04/2023 11:56:07.910 AM]  	'WaitForBuildOperation', hashcode:42412516
[28/04/2023 11:56:07.910 AM]  

[28/04/2023 11:56:07.910 AM]  Operation Dequeue : 'WaitForBuildOperation'
[...]

PR #1669, opened 2023-05-17
Adding warning that AssemblyResolution works only with netframworks

engyebrahim:
fix: https://github.com/microsoft/testfx/issues/1645


src/Adapter/MSTest.TestAdapter/Resources/Resource.resx
```diff
@@ -384,4 +384,7 @@ Error: {1}</value>
     <comment>{0}: Zero based index if an element inside of an array,
 {1}: Test name</comment>
   </data>
+  <data name="AssemblyResolutionIsOnlyWorkingWithNetFramework" xml:space="preserve">
+    <value>Note that AssemblyResolution is only being applied when using .NET Framework target.</value>
```
Evangelink: Maybe we should mention that this is a runsettings configuration entry.

Maybe something like `The runsettings 'AssemblyResolution' entries are only supported in .NET Framework. If your tests are working consider removing this entry otherwise you will need to make sure all dependent assemblies are found in the ouput directory of your test`.

@eleshoe Do you think this warning would give you enough information to understand what to do and how to fix your test?

src/Adapter/MSTest.TestAdapter/MSTestDiscoverer.cs
```diff
@@ -51,6 +52,14 @@ public class MSTestDiscoverer : ITestDiscoverer
...
         {
             MSTestSettings.PopulateSettings(discoveryContext);
+
+#if !NETFRAMEWORK
+            MSTestAdapterSettings settings = MSTestSettingsProvider.Settings;
+            if (settings.SearchDirectories.Count > 1)
+            {
+                logger.SendMessage(TestMessageLevel.Warning, Resource.AssemblyResolutionIsOnlyWorkingWithNetFramework);
+            }
+#endif
```
Evangelink: I think this needs to be done also in `MSTestExecutor`. I would probably also hide this details under the `MSTestSettings` class by adding a `ValidateSettings` method so that it's easier to reuse and it's grouping settings validation under the same umbrella (in case we need to validate more entries in the future).

MarcoRossignoli: A `bool MSTestAdapterSettings.Validate(Action<string> validationError)` maybe could help

or

`MSTestAdapterSettings.Validate(ILogger)` if there're different logger levels

src/Adapter/MSTestAdapter.PlatformServices/Services/MSTestAdapterSettings.cs
```diff
@@ -44,7 +45,7 @@ public MSTestAdapterSettings()
     /// <summary>
     ///  Gets list of paths recursive or non recursive paths.
     /// </summary>
-    protected List<RecursiveDirectoryPath> SearchDirectories { get; private set; }
+    public List<RecursiveDirectoryPath> SearchDirectories { get; private set; }
```
Evangelink: This no longer needs to be public.

src/Adapter/MSTestAdapter.PlatformServices/Services/MSTestAdapterSettings.cs
```diff
@@ -127,6 +128,17 @@ public static MSTestAdapterSettings ToSettings(XmlReader reader)
         return settings;
     }
 
+    public static void ValidateSettings(IMessageLogger logger)
+    {
+#if !NETFRAMEWORK
+        MSTestAdapterSettings settings = MSTestSettingsProvider.Settings;
+        if (settings.SearchDirectories.Count > 1)
```
Evangelink: This property can be retrieved by subclasses and it's possible that some extending user would add some path to this property so I am not sure this is the best way to check that.

I would suggest setting some boolean field if the `AssemblyResolution` setting is found and check this field here.

src/Adapter/MSTestAdapter.PlatformServices/Services/MSTestAdapterSettings.cs
```diff
@@ -127,6 +128,17 @@ public static MSTestAdapterSettings ToSettings(XmlReader reader)
         return settings;
     }
 
+    public static void ValidateSettings(IMessageLogger logger)
+    {
+#if !NETFRAMEWORK
+        MSTestAdapterSettings settings = MSTestSettingsProvider.Settings;
```
Evangelink: This doesn't seem needed, you are on the class containing the property you want.

src/Adapter/MSTest.TestAdapter/MSTestDiscoverer.cs
```diff
@@ -51,6 +52,7 @@ public class MSTestDiscoverer : ITestDiscoverer
         try
         {
             MSTestSettings.PopulateSettings(discoveryContext);
+            MSTestAdapterSettings.ValidateSettings(logger);
```
Evangelink: It's ok as-is but I would recommend to have the `ValidateSettings` method on `MSTestSettings` that would then call the one on `MSTestAdapterSettings`. This way when reading code here you have something more logical (I load settings and validate from the same object).

src/Adapter/MSTestAdapter.PlatformServices/Services/MSTestAdapterSettings.cs
```diff
@@ -127,6 +128,17 @@ public static MSTestAdapterSettings ToSettings(XmlReader reader)
         return settings;
     }
 
+    public static void ValidateSettings(IMessageLogger logger)
```
Evangelink: Does it need to be `public`? Can it be `internal`?

engyebrahim: when I do it protected I can't call it from the `MSTestSettings` class

Evangelink: I said internal not protected :P

engyebrahim: sorry my bad

src/Adapter/MSTest.TestAdapter/MSTestExecutor.cs
```diff
@@ -86,6 +86,7 @@ public void RunTests(IEnumerable<string>? sources, IRunContext? runContext, IFra
         try
         {
             MSTestSettings.PopulateSettings(runContext);
+            MSTestSettings.ValidateSettings(frameworkHandle);
```
Evangelink, line 91: Might be worth adding a comment in the other RunTests method to say why we don't validate there.

src/Adapter/MSTestAdapter.PlatformServices/Services/MSTestAdapterSettings.cs
```diff
@@ -127,6 +138,17 @@ public static MSTestAdapterSettings ToSettings(XmlReader reader)
...
+    {
+#if !NETFRAMEWORK
+        MSTestAdapterSettings settings = MSTestSettingsProvider.Settings;
```
Evangelink: As per my previous comment, this is not needed and can be removed :)

src/Adapter/MSTestAdapter.PlatformServices/Services/MSTestAdapterSettings.cs
```diff
@@ -46,6 +47,10 @@ public MSTestAdapterSettings()
     /// </summary>
     protected List<RecursiveDirectoryPath> SearchDirectories { get; private set; }
 
+#if !NETFRAMEWORK
```
Evangelink: I think (might be wrong) that you could remove the compiler directive everywhere but in the Validate method.

src/Adapter/MSTestAdapter.PlatformServices/Services/MSTestAdapterSettings.cs
Evangelink: It's still public

Evangelink (commented):
A few more changes and we will be good to merge

src/Adapter/MSTestAdapter.PlatformServices/Services/MSTestAdapterSettings.cs
engyebrahim: ValidateSettings func and the class we are using it inside are in different projects can't make it internal

src/Adapter/MSTestAdapter.PlatformServices/Services/MSTestAdapterSettings.cs
engyebrahim: gives me runtime errors

Evangelink (approved):

PR #1715, opened 2023-07-11
Return T from the generic IsInstanceOfType

gao-artur:
@Evangelink, I have a few questions:
1. How to specify `T` in the `<returns>` section? `<see cref="T"/>` causes an error.
2. Any suggestions for better tests?

### DIFF
--- a/test/UnitTests/TestFramework.UnitTests/Assertions/AssertTests.IsInstanceOfTypeTests.cs
+++ b/test/UnitTests/TestFramework.UnitTests/Assertions/AssertTests.IsInstanceOfTypeTests.cs
@@ -67,13 +67,15 @@ public partial class AssertTests
     [TestMethod]
     public void IsInstanceOfTypeUsingGenericType_OnSameInstance_DoesNotThrow()
     {
-        Assert.IsInstanceOfType<int>(5);
+        var result = Assert.IsInstanceOfType<int>(5);
+        Verify(result == 5);
     }
 
     [TestMethod]
     public void IsInstanceOfTypeUsingGenericType_OnHigherInstance_DoesNotThrow()
     {
-        Assert.IsInstanceOfType<object>(5);
+        var result = Assert.IsInstanceOfType<object>(5);
+        Verify(((int)result) == 5);
     }
 
     [TestMethod]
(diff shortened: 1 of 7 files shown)


gao-artur:

Evangelink:
Hey thanks for the draft @gao-artur!



> 1. How to specify `T` in the `<returns>` section? `<see cref="T"/>` causes an error.

You should be using `<typeparamref name="T"/>`.

> 2. Any suggestions for better tests?

I think what you have done is enough. I don't anticipate any corner case with casting as we ensure it's castable before.

test/UnitTests/TestFramework.UnitTests/Assertions/AssertTests.IsInstanceOfTypeTests.cs
```diff
@@ -67,13 +67,15 @@ public void IsInstanceOfTypeUsingGenericType_WhenValueIsNull_Fails()
     [TestMethod]
     public void IsInstanceOfTypeUsingGenericType_OnSameInstance_DoesNotThrow()
     {
-        Assert.IsInstanceOfType<int>(5);
+        var result = Assert.IsInstanceOfType<int>(5);
```
Evangelink: Nit: I would make the type explicit instead of `var`

```suggestion
        int result = Assert.IsInstanceOfType<int>(5);
```

test/UnitTests/TestFramework.UnitTests/Assertions/AssertTests.IsInstanceOfTypeTests.cs
```diff
@@ -67,13 +67,15 @@ public void IsInstanceOfTypeUsingGenericType_WhenValueIsNull_Fails()
...
+        Verify(result == 5);
     }
 
     [TestMethod]
     public void IsInstanceOfTypeUsingGenericType_OnHigherInstance_DoesNotThrow()
     {
-        Assert.IsInstanceOfType<object>(5);
+        var result = Assert.IsInstanceOfType<object>(5);
```
Evangelink: Same nit here

```suggestion
        object result = Assert.IsInstanceOfType<object>(5);
```

Evangelink:
After discussion with PMs we will have to target this change for v4 of MSTest.

gao-artur:
I see you already changed it. No action from my side needed, right?

Evangelink:
Yep it's all good!

PR #1737, opened 2023-09-07
Don't stop

nohwnd:
Avoid unlikely bug where test is marked as inconclusive but then fails, and the duration is captured from the first stop of the stopwatch that happened in the inconclusive branch.

There is no issue to link to, I just noticed this when digging in code.


Evangelink (approved):
Rejecting ;)

PR #1757, opened 2023-10-30
Code clean-up 

ViktorHofer:
See commits for description of changes. **Review by commits**.

### DIFF
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -1,4 +1,3 @@
-<?xml version="1.0" encoding="utf-8"?>
 <Project>
 
   <Import Project="Sdk.props" Sdk="Microsoft.DotNet.Arcade.Sdk" />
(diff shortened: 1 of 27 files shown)


Directory.Build.props
```diff
@@ -73,9 +70,4 @@
     <StrongNameKeyId>Microsoft</StrongNameKeyId>
   </PropertyGroup>
 
-  <!-- Test config -->
-  <PropertyGroup>
-    <UseVSTestRunner>true</UseVSTestRunner>
```
Evangelink, line 78: I recall I had some issue with xunit when this entry was not around but let's see.

Evangelink (approved):

Directory.Build.props
ViktorHofer: This was just moved into Versions.props where Arcade feature flags are expected to be located.

PR #1794, opened 2023-11-20
MSTest deps on Test SDK is superior or equal

Evangelink:


fhnaseer (approved):

src/Package/MSTest/MSTest.nuspec
```diff
@@ -6,13 +6,13 @@
       <group targetFramework="netstandard2.0">
         <dependency id="MSTest.TestAdapter" version="[$version$]" />
         <dependency id="MSTest.TestFramework" version="[$version$]" />
-        <dependency id="Microsoft.NET.Test.Sdk" version="[$TestPlatformVersion$]" />
+        <dependency id="Microsoft.NET.Test.Sdk" version="$TestPlatformVersion$" />
```
fhnaseer, line 9: Not sure with the symbols convention here. Can it be $TestPlatformVersion$ or higher?

Evangelink: This used to be = version (syntax is version between brackets), this is now >=.

PR #1803, opened 2023-11-21
Fix test project

nohwnd:


Directory.Build.Local.props
```diff
@@ -76,7 +76,7 @@
 
   <!-- Test config -->
   <PropertyGroup>
-    <TestRunnerName>RepoCustomRunner</TestRunnerName>
+    <TestRunnerName>TestingPlatformRunner</TestRunnerName>
```
Evangelink, line 79: What's the reasoning here? This seems wrong. The current target was already able to run either the vstest + custom adapter for regular MSTest or TA tests for the runner. This change is breaking this.

nohwnd: I was just trying to make it work in the parent repo where this name is used and was missing. It now works in the parent repo (or at least does not throw as before), so if this change breaks the flow here, we will need something even more complicated.

test/UnitTests/Microsoft.Testing.Platform.UnitTests/Microsoft.Testing.Platform.UnitTests.csproj
```diff
@@ -27,18 +27,15 @@
...
-    <PackageReference Include="Microsoft.Build.Utilities.Core" Version="$(MicrosoftBuildUtilitiesCore)" PrivateAssets="All" />
     <PackageReference Include="Moq" Version="$(MoqVersion)" />
     <PackageReference Include="Microsoft.Testing.Framework" Version="$(MicrosoftTestingPlatformVersion)" />
     <PackageReference Include="Microsoft.Testing.Framework.SourceGeneration" Version="$(MicrosoftTestingPlatformVersion)" />
-    <PackageReference Include="Microsoft.Testing.Platform.Extensions.CodeCoverage" Version="$(MicrosoftTestingPlatformExtensionsCodeCoverageVersion)" Condition=" '$(EnableCodeCoverage)' == 'true' " />
-    <PackageReference Include="Microsoft.Testing.Platform.Extensions.HangDump" Version="$(MicrosoftTestingPlatformVersion)" />
-    <PackageReference Include="Microsoft.Testing.Platform.Extensions.Telemetry" Version="$(MicrosoftTestingPlatformVersion)" />
-    <PackageReference Include="Microsoft.Testing.Platform.MSBuild" Version="$(MicrosoftTestingPlatformVersion)" />
+      <PackageReference Include="Microsoft.Testing.Platform.MSBuild" Version="$(MicrosoftTestingPlatformVersion)" />
   </ItemGroup>
```
Evangelink: Nit: Indentation is too deep
```suggestion
    <PackageReference Include="Microsoft.Testing.Platform.MSBuild" Version="$(MicrosoftTestingPlatformVersion)" />
  </ItemGroup>
```

test/UnitTests/Microsoft.Testing.Platform.UnitTests/Program.cs
```diff
@@ -66,3 +66,41 @@ public async Task AfterRunAsync(int returnValue, CancellationToken cancellationT
 
     public Task BeforeRunAsync(CancellationToken cancellationToken) => Task.CompletedTask;
 }
+
+internal sealed class FakeTrxReportGeneratorCommandLine : ICommandLineOptionsProvider
```
Evangelink, line 70: Was it only for debugging purposes?

Directory.Build.Local.props
Evangelink: Oh ok, I misunderstood the change you actually did. That's all good as-is, thanks!

test/UnitTests/Microsoft.Testing.Platform.UnitTests/.editorconfig
```diff
@@ -0,0 +1,29 @@
...
+# SA1009: Closing parenthesis should be spaced correctly
+dotnet_diagnostic.SA1009.severity = none
+
+# SA1111: Closing parenthesis should be on line of last parameter
+dotnet_diagnostic.SA1111.severity = none
+
+# SA1414: Tuple types in signatures should have element names
+dotnet_diagnostic.SA1414.severity = none
+
+dotnet_diagnostic.CA1861.severity = none
```
Evangelink, line 28:
```suggestion
# CA1861: Avoid constant arrays as arguments
dotnet_diagnostic.CA1861.severity = none
```

test/UnitTests/Microsoft.Testing.Platform.UnitTests/.editorconfig
```diff
@@ -0,0 +1,29 @@
...
+
+dotnet_diagnostic.CA1001.severity = none
```
Evangelink, line 31:
```suggestion
# CA1001: Types that own disposable fields should be disposable
dotnet_diagnostic.CA1001.severity = none
```

Evangelink, line 31: I feel like this one should not be disabled but instead violations should be fixed.

test/UnitTests/Microsoft.Testing.Platform.UnitTests/Microsoft.Testing.Platform.UnitTests.csproj
```diff
@@ -27,18 +27,15 @@
   <!-- end netstandard2.0 polyfill -->
 
   <ItemGroup>
-    <PackageReference Include="Microsoft.Build.Utilities.Core" Version="$(MicrosoftBuildUtilitiesCore)" PrivateAssets="All" />
     <PackageReference Include="Moq" Version="$(MoqVersion)" />
     <PackageReference Include="Microsoft.Testing.Framework" Version="$(MicrosoftTestingPlatformVersion)" />
     <PackageReference Include="Microsoft.Testing.Framework.SourceGeneration" Version="$(MicrosoftTestingPlatformVersion)" />
-    <PackageReference Include="Microsoft.Testing.Platform.Extensions.CodeCoverage" Version="$(MicrosoftTestingPlatformExtensionsCodeCoverageVersion)" Condition=" '$(EnableCodeCoverage)' == 'true' " />
-    <PackageReference Include="Microsoft.Testing.Platform.Extensions.HangDump" Version="$(MicrosoftTestingPlatformVersion)" />
-    <PackageReference Include="Microsoft.Testing.Platform.Extensions.Telemetry" Version="$(MicrosoftTestingPlatformVersion)" />
```
Evangelink, line 36: Could you explain why they are disabled?

test/UnitTests/Microsoft.Testing.Platform.UnitTests/Program.cs
nohwnd: The runner in parent repo provides those parameters, but we cannot use the reporter because there are breaking changes in the platform. If the provider is missing the run exits with "unknown parameter". So I removed the real provider and added this fake one for now. 

I described the exact issue in our internal chat.

test/UnitTests/Microsoft.Testing.Platform.UnitTests/.editorconfig
nohwnd: 🤷‍♀️ it's the exact same code as in the original repo, so either we are not reporting it there or we also ignore them, or there were changes that I did not mean to do.

test/UnitTests/Microsoft.Testing.Platform.UnitTests/Microsoft.Testing.Platform.UnitTests.csproj
nohwnd: These "external" extensions depend on platform, but if you make a breaking change in platform they prevent the tests from running. So the change cannot be merged, and these plugins cannot be updated.

test/UnitTests/Microsoft.Testing.Platform.UnitTests/.editorconfig
nohwnd: This is what it is complaining about: 

test\UnitTests\Microsoft.Testing.Platform.UnitTests\ServerMode\ServerDataConsumerServiceTests.cs(19,21): error CA1001: Type 'ServerDataConsumerServiceTests' owns disposable field(s) '_service' but is not disposabl
e (https://learn.microsoft.com/dotnet/fundamentals/code-analysis/quality-rules/ca1001) [S:\p\testanywhere\mstest\test\UnitTests\Microsoft.Testing.Platform.UnitTests\Microsoft.Testing.Platform.UnitTests.csproj::TargetFramework=net462]

can we merge this now, and discuss later?

PR #1819, opened 2023-11-23
Add configuration tests to boost coverage

cvpoienaru:
Added configuration tests to boost coverage


src/Platform/Microsoft.Testing.Platform/Configurations/ConfigurationManager.cs
```diff
@@ -16,7 +16,7 @@ internal sealed class ConfigurationManager : IConfigurationManager
 
     public void AddConfigurationSource(Func<IConfigurationSource> source) => _configurationSources.Add(source);
 
-    internal async Task<IConfiguration> BuildAsync(IServiceProvider serviceProvider, FileLoggerProvider? syncFileLoggerProvider)
+    internal async Task<IConfiguration> BuildAsync(IServiceProvider serviceProvider, ILoggerProvider? syncFileLoggerProvider)
```
Evangelink: @MarcoRossignoli Please review this change. From the field name, I think this is a case where we DO NEED to have a file logger instance.

src/Platform/Microsoft.Testing.Platform/Configurations/AggregatedConfiguration.cs
```diff
@@ -64,7 +70,11 @@ public void SetTestHostWorkingDirectory(string workingDirectory)
         _testHostWorkingDirectory = workingDirectory;
     }
 
-    public async Task CheckTestResultsDirectoryOverrideAndCreateItAsync(ICommandLineOptions commandLineOptions, IRuntime runtime, ServiceProvider serviceProvider)
+    public async Task CheckTestResultsDirectoryOverrideAndCreateItAsync(
+        ICommandLineOptions commandLineOptions,
+        IRuntime runtime,
+        IFileSystem fileSystem,
+        ServiceProvider serviceProvider)
```
Evangelink, line 72: @cvpoienaru @MarcoRossignoli Can't we have these parameters as ctor instead? It feels weird to have technical object on a public API.

src/Platform/Microsoft.Testing.Platform/Logging/FileLoggerProvider.cs
```diff
@@ -5,7 +5,9 @@
 
 namespace Microsoft.Testing.Platform.Logging;
 
-internal sealed class FileLoggerProvider : ILoggerProvider, IDisposable
+#pragma warning disable CA1852
```
Evangelink: You probably crafted this disable manually, could you add the `// rule message` part either manually or let the lightbulb add the suppression.

src/Platform/Microsoft.Testing.Platform/Logging/FileLoggerProvider.cs
```diff
@@ -16,6 +18,14 @@ internal sealed class FileLoggerProvider : ILoggerProvider, IDisposable
     private readonly string _logPrefixName;
     private readonly bool _customDirectory;
 
+    /* for testing */
+    internal FileLoggerProvider()
+    {
+        _clock = new SystemClock();
+        _logPrefixName = string.Empty;
+        _customDirectory = false;
+    }
```
Evangelink: It's not super clear, do you want to avoid the `FileLogger` creation? Otherwise, I don't see why you are not simply providing arguments to the other ctor.

src/Platform/Microsoft.Testing.Platform/Logging/FileLoggerProvider.cs
```diff
@@ -16,6 +18,14 @@ internal sealed class FileLoggerProvider : ILoggerProvider, IDisposable
     private readonly string _logPrefixName;
     private readonly bool _customDirectory;
 
+    /* for testing */
+    internal FileLoggerProvider()
```
Evangelink: Nit: suggested syntax avoid having the comment being moved around and ending up not attached.

```suggestion    
    internal /* for testing */ FileLoggerProvider()
```

test/UnitTests/Microsoft.Testing.Platform.UnitTests/Configuration/ConfigurationManagerTests.cs
```diff
@@ -0,0 +1,179 @@
...
+                mockServiceProvider.Object,
+                null));
+
+        fakeConfigurationSource.Verify(x => x.IsEnabledAsync(), Times.Once);
+        fakeConfigurationSource.Verify(x => x.InitializeAsync(), Times.Once);
+    }
+}
+
+#pragma warning disable CA1852
+internal class FakeConfigurationSource : IConfigurationSource, IAsyncInitializableExtension
```
Evangelink, line 157: Why not seal the class instead of disabling the rule?

src/Platform/Microsoft.Testing.Platform/Services/ServiceProvider.cs
```diff
@@ -73,7 +73,7 @@ public bool TryAddService(object service)
         return true;
     }
 
-    public IEnumerable<object> GetServicesInternal(
+    public /* for testing purposes */ virtual IEnumerable<object> GetServicesInternal(
```
Evangelink: I don't feel super comfortable to allow virtual just for tests.

src/Platform/Microsoft.Testing.Platform/Logging/FileLoggerProvider.cs
```diff
@@ -29,11 +39,11 @@ public FileLoggerProvider(string logFolder, IClock clock, LogLevel logLevel, str
 
     public LogLevel LogLevel { get; }
 
-    public FileLogger FileLogger { get; private set; }
+    public FileLogger? FileLogger { get; private set; }
 
     public bool SyncFlush { get; }
 
-    public async Task CheckLogFolderAndMoveToTheNewIfNeededAsync(string testResultDirectory)
+    public /* for testing */ virtual async Task CheckLogFolderAndMoveToTheNewIfNeededAsync(string testResultDirectory)
```
Evangelink: I am not super comfortable to make something virtual for testing.

src/Platform/Microsoft.Testing.Platform/Logging/FileLoggerProvider.cs
```diff
@@ -29,11 +39,11 @@ public FileLoggerProvider(string logFolder, IClock clock, LogLevel logLevel, str
 
     public LogLevel LogLevel { get; }
 
-    public FileLogger FileLogger { get; private set; }
+    public FileLogger? FileLogger { get; private set; }
```
Evangelink: Is it nullable for testing?

src/Platform/Microsoft.Testing.Platform/Configurations/AggregatedConfiguration.cs
MarcoRossignoli: api is not public this object is internal or? it's needed because if I recall well we have these object late and not when we create the type, if are available before fine to me.
Internal is not needed, the type is internal.

src/Platform/Microsoft.Testing.Platform/Configurations/ConfigurationManager.cs
MarcoRossignoli: Yep we use it to log during the infra building, so we call create and write in a sync way to the `FileLogger` on purpose.

src/Platform/Microsoft.Testing.Platform/Configurations/AggregatedConfiguration.cs
```diff
@@ -52,7 +52,13 @@ public AggregatedConfiguration(IConfigurationProvider[] configurationProviders)
         }
     }
 
-    public /* for testing */ void SetCurrentWorkingDirectory(string workingDirectory)
+    internal /* for testing */ void SetResultDirectory(string resultDirectory)
```
MarcoRossignoli: why move to internal?

src/Platform/Microsoft.Testing.Platform/Configurations/AggregatedConfiguration.cs
Evangelink: Ah yeah sorry I thought the class was public. Still if method isn't private or static, I would expect technical objects to be provided at ctor level.

src/Platform/Microsoft.Testing.Platform/Configurations/AggregatedConfiguration.cs
cvpoienaru: `SetResultDirectory` is new, so made it internal by default. If you mean why I moved `SetCurrentWorkingDirectory` to internal, that is because it's only used for testing so I see no purpose in keeping it public for now. Should it be the case in the future to make it public, we can deal with that later. WDYT ?

src/Platform/Microsoft.Testing.Platform/Configurations/AggregatedConfiguration.cs
cvpoienaru: So what should we do about this method ? Since the type is not public, should we address it later ?

src/Platform/Microsoft.Testing.Platform/Logging/FileLoggerProvider.cs
cvpoienaru: Indeed, I've manually added this suppression. It's strange because nothing is wrong and VS doesn't warn of it as a consequence, but when I build in CLI I get an error. The error I get is because the analyzer thinks the class can be sealed, however it cannot because it has a virtual member. Will manually add the rule description too.

src/Platform/Microsoft.Testing.Platform/Logging/FileLoggerProvider.cs
cvpoienaru: Yes, made it nullable to avoid having to instantiate it. Should I revert it ?

src/Platform/Microsoft.Testing.Platform/Configurations/AggregatedConfiguration.cs
MarcoRossignoli: If I'm not mistaken the class is already internal so by design all methods can stay public, this class will be never exposed so it's fine keep all methods public for now.

test/UnitTests/Microsoft.Testing.Platform.UnitTests/Configuration/ConfigurationManagerTests.cs
cvpoienaru: We cannot seal it because we have virtual members.

src/Platform/Microsoft.Testing.Platform/Configurations/AggregatedConfiguration.cs
MarcoRossignoli: We discussed off-line with @Evangelink and for now it can stays as-is, we will refactor if we find a good way to not confuse the "state" and the references mutation/aliasing.

src/Platform/Microsoft.Testing.Platform/Services/ServiceProvider.cs
cvpoienaru: If we don't mark this `virtual` then we cannot make the `ServiceProvider` return a `Mock<FileLoggerProvider>`. This in turn makes testing `CheckTestResultsDirectoryOverrideAndCreateItAsync` more complex to fully cover as well as slower, because we'll have to move a file to a new log folder and the paths will all have to be legit.

src/Platform/Microsoft.Testing.Platform/Logging/FileLoggerProvider.cs
cvpoienaru: Same as the comment below, if we don't mark this `virtual` then we cannot mock it and this means the scenarios to test the aggregated configuration will be more complex since we'll have to pass legit paths around and account for the work this method does too. WDYT ?

src/Platform/Microsoft.Testing.Platform/Configurations/ConfigurationManager.cs
cvpoienaru: Ok, changed it back to `FileLoggerProvider`, but in this case I'll have to mark `CreateLogger` as `virtual` in order to mock it. Do you agree ?

src/Platform/Microsoft.Testing.Platform/Configurations/AggregatedConfiguration.cs
cvpoienaru: Great, thanks ! Will resolve the conversation then.

src/Platform/Microsoft.Testing.Platform/Logging/FileLoggerProvider.cs
cvpoienaru: Indeed, I wanted to avoid the creation of the `FileLogger`. Should I instead create it for real ?

src/Platform/Microsoft.Testing.Platform/Configurations/AggregatedConfiguration.cs
Evangelink: +1, let's keep public

src/Platform/Microsoft.Testing.Platform/Configurations/AggregatedConfiguration.cs
```diff
@@ -64,7 +70,11 @@ public void SetTestHostWorkingDirectory(string workingDirectory)
         _testHostWorkingDirectory = workingDirectory;
     }
 
-    public async Task CheckTestResultsDirectoryOverrideAndCreateItAsync(ICommandLineOptions commandLineOptions, IRuntime runtime, ServiceProvider serviceProvider)
+    public async Task CheckTestResultsDirectoryOverrideAndCreateItAsync(
+        ICommandLineOptions commandLineOptions,
+        IRuntime runtime,
+        IFileSystem fileSystem,
```
Evangelink, line 71: Just double checked and file system and runtime are both system API wrappers and so are available from the start. I would prefer to have them moved to the ctor.

src/Platform/Microsoft.Testing.Platform/Logging/FileLoggerProvider.cs
Evangelink: Well the problem is that with the change, all callers of this property now have to handle null check when we know this is not null.

Depending on whether the `FileLogger` ctor does something or not, I would either make the setter internal with the comment "for testing" or take `FileLogger` in the "for testing" ctor.

src/Platform/Microsoft.Testing.Platform/Configurations/ConfigurationManager.cs
Evangelink: Maybe the easiest is to introduce a `IFileLoggerProvider` and replace this use by that. This is a specialized interface and can be easily mocked.

src/Platform/Microsoft.Testing.Platform/Services/ServiceProvider.cs
cvpoienaru: Reverted, thanks !

src/Platform/Microsoft.Testing.Platform/Logging/FileLoggerProvider.cs
cvpoienaru: Reverted.

src/Platform/Microsoft.Testing.Platform/Logging/FileLoggerProvider.cs
cvpoienaru: Reverted.

src/Platform/Microsoft.Testing.Platform/Configurations/AggregatedConfiguration.cs
cvpoienaru: Reverted, thanks.

src/Platform/Microsoft.Testing.Platform/Logging/FileLoggerProvider.cs
cvpoienaru: Reverted.

src/Platform/Microsoft.Testing.Platform/Configurations/AggregatedConfiguration.cs
cvpoienaru: Closing this as you already did a PR for this, thanks !

src/Platform/Microsoft.Testing.Platform/Configurations/ConfigurationManager.cs
cvpoienaru: Fixed, thanks for the suggestion.

src/Platform/Microsoft.Testing.Platform/Logging/FileLoggerProvider.cs
cvpoienaru: Removed the warning suppression.

src/Platform/Microsoft.Testing.Platform/Services/ServiceProvider.cs
```diff
@@ -15,7 +15,7 @@
 
 namespace Microsoft.Testing.Platform.Services;
 
-internal sealed class ServiceProvider : IServiceProvider, ICloneable
+internal class ServiceProvider : IServiceProvider, ICloneable
```
Evangelink, line 18: Can be kept sealed

Evangelink (approved):
Let's merge as-is, I will do a post-merge refactoring

src/Platform/Microsoft.Testing.Platform/Services/ServiceProvider.cs
cvpoienaru: Fixed by: https://github.com/microsoft/testfx/pull/1833#pullrequestreview-1748080483

PR #1836, opened 2023-11-24
Enable localization for Telemetry

cvpoienaru:
Enabled localization for Telemetry.


src/Platform/Microsoft.Testing.Platform/Telemetry/TelemetryManager.cs
```diff
@@ -38,16 +41,16 @@ public async Task<ITelemetryCollector> BuildAsync(ServiceProvider serviceProvide
         bool isTelemetryOptedOut = !testApplicationOptions.EnableTelemetry;
 
         ILogger<TelemetryManager> logger = loggerFactory.CreateLogger<TelemetryManager>();
-        await logger.LogInformationAsync($"TestApplicationOptions.EnableTelemetry: {testApplicationOptions.EnableTelemetry}");
+        await logger.LogInformationAsync(string.Format(CultureInfo.InvariantCulture, PlatformResources.TelemetryManagerTestApplicationOptionsTelemetryEnabled, testApplicationOptions.EnableTelemetry));
```
Evangelink, line 44: Let's not translate logger messages, otherwise that will be a nightmare for us when we will have to debug logs coming from different cultures.

Adding this comment only once, please revert all locations.

Evangelink (changes requested):

PR #1864, opened 2023-12-01
Add experimental `MSBuild` sdk project

MarcoRossignoli:
This PR is adding the first version of the MSBuild sdk for MSTest, it could be interesting mainly for:

* add or remove dependencies without the needs to ask to the user to do it, we know that's not easy to flow new features that needs new dependency to the users, they've to read the documentation but it's hard to reach everyone.
* we can provide simple msbuild props knob to disable/enable features.
* we can have better "testing coverage" because we can always test "our" configurations and guarantee that everything is good, having less "moving parts", users will add dependencies only for their business code.
* it's battle tested and used by other well known .NET platform to simplify the "default usage" https://learn.microsoft.com/en-us/dotnet/core/project-sdk/overview#available-sdks

The outcome will be a project file like:
```xml
<Project Sdk="MSTest.Sdk/3.2.0-dev">
  <PropertyGroup>
    <TargetFramework>net8.0</TargetFramework>
    <!--<UseMSTestRunner>false</UseMSTestRunner>-->
    <!--<TestingPlatformDonetTestSupport>true</TestingPlatformDonetTestSupport>-->
  </PropertyGroup>
</Project>
```

The sample above is using the version inside the `Sdk` attribute self but it can be done "globally" using the `global.json` too like described here https://learn.microsoft.com/en-us/visualstudio/msbuild/how-to-use-project-sdk

### DIFF
--- a/src/Package/MSTest.Sdk/Sdk/Sdk.props
+++ b/src/Package/MSTest.Sdk/Sdk/Sdk.props
@@ -1,0 +1,21 @@
+﻿<?xml version="1.0" encoding="utf-8" ?>
+<Project>
+  <!-- Implicit top import -->
+  <Import Project="Sdk.props" Sdk="Microsoft.NET.Sdk" />
+  
+  <PropertyGroup>
+    <UseMSTestRunner>true</UseMSTestRunner>
+    <TestingPlatformDonetTestSupport>false</TestingPlatformDonetTestSupport>
+    <IsPackable>false</IsPackable>
+    <IsTestProject>true</IsTestProject>
+    <LangVersion>preview</LangVersion>
+    <ImplicitUsings>enable</ImplicitUsings>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <PackageReference Include="Microsoft.NET.Test.Sdk" Version="17.9.0-preview-23580-01" />
+    <PackageReference Include="MSTest.TestAdapter" Version="3.2.0-preview.23601.5" />
+    <PackageReference Include="MSTest.TestFramework" Version="3.2.0-preview.23601.5" />
+    <PackageReference Include="Microsoft.Testing.Platform.Extensions.VSTestBridge" Version="1.0.0-alpha.23601.2" />
+  </ItemGroup>
+</Project>
--- a/src/Package/MSTest.Sdk/Sdk/Sdk.targets
+++ b/src/Package/MSTest.Sdk/Sdk/Sdk.targets
@@ -1,0 +1,40 @@
+﻿<?xml version="1.0" encoding="utf-8" ?>
+<Project xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+  <Target Name="_SetupEntryPoint" BeforeTargets="GenerateProgramFile" Condition=" '$(UseMSTestRunner)' == 'false' " >
+    <PropertyGroup>
+      <!-- Disable testing platform entry point generation -->
+      <GenerateTestingPlatformEntryPoint>false</GenerateTestingPlatformEntryPoint>
+    </PropertyGroup>
+    <!--
+          We need to re-enable the VSTest sdk program.* generation by it's by default disabled when we reference Microsoft.Testing.Platform.MSBuild.
+          We cannot use the GenerateProgramFile property as Microsoft.NET.Test.Sdk.targets it's an 'InitialTargets' and we cannot plug any target before it.
+    -->
+    <CallTarget Targets="_GenerateProgramFileTakenFromVSTest" />
+  </Target>
+
+  <!--
+    ============================================================
+    GenerateProgramFile taken from Microsoft.NET.Test.Sdk.targets 
+    Generates Program file which contains the Main entry point
+    ============================================================
+  -->
+  <Target Name="_GenerateProgramFileTakenFromVSTest" >
+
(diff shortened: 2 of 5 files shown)


src/Package/MSTest.Sdk/Sdk/Sdk.props
```diff
@@ -0,0 +1,21 @@
...
+    <UseMSTestRunner>true</UseMSTestRunner>
+    <TestingPlatformDonetTestSupport>false</TestingPlatformDonetTestSupport>
+    <IsPackable>false</IsPackable>
+    <IsTestProject>true</IsTestProject>
+    <LangVersion>preview</LangVersion>
+    <ImplicitUsings>enable</ImplicitUsings>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <PackageReference Include="Microsoft.NET.Test.Sdk" Version="17.9.0-preview-23580-01" />
```
MarcoRossignoli: this can be refactored in a prop, it's a starting point

src/Package/MSTest.Sdk/Sdk/Sdk.props
```diff
@@ -0,0 +1,21 @@
...
+  <!-- Implicit top import -->
+  <Import Project="Sdk.props" Sdk="Microsoft.NET.Sdk" />
+  
+  <PropertyGroup>
+    <UseMSTestRunner>true</UseMSTestRunner>
+    <TestingPlatformDonetTestSupport>false</TestingPlatformDonetTestSupport>
+    <IsPackable>false</IsPackable>
+    <IsTestProject>true</IsTestProject>
+    <LangVersion>preview</LangVersion>
+    <ImplicitUsings>enable</ImplicitUsings>
```
MarcoRossignoli: this one maybe should be in the project self like asp.net template
```xml
<Project Sdk="Microsoft.NET.Sdk.Web">

  <PropertyGroup>
    <TargetFramework>net8.0</TargetFramework>
    <Nullable>enable</Nullable>
  </PropertyGroup>

</Project>
```

Evangelink: Yep I would prefer to have these coming from user project (`LangVersion` + `ImplicitUsings`)

src/Package/MSTest.Sdk/Sdk/Sdk.props
Evangelink: Let's start with stable version

```suggestion
    <PackageReference Include="Microsoft.NET.Test.Sdk" Version="17.8.0" />
```

src/Package/MSTest.Sdk/Sdk/Sdk.targets
```diff
@@ -0,0 +1,40 @@
+﻿<?xml version="1.0" encoding="utf-8" ?>
+<Project xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+  <Target Name="_SetupEntryPoint" BeforeTargets="GenerateProgramFile" Condition=" '$(UseMSTestRunner)' == 'false' " >
+    <PropertyGroup>
+      <!-- Disable testing platform entry point generation -->
+      <GenerateTestingPlatformEntryPoint>false</GenerateTestingPlatformEntryPoint>
+    </PropertyGroup>
+    <!--
+          We need to re-enable the VSTest sdk program.* generation by it's by default disabled when we reference Microsoft.Testing.Platform.MSBuild.
+          We cannot use the GenerateProgramFile property as Microsoft.NET.Test.Sdk.targets it's an 'InitialTargets' and we cannot plug any target before it.
```
Evangelink, line 10: This is not entirely clear to me. Maybe we can have some offline sync.

MarcoRossignoli: sure, the fact that we enable by default the runner in the sdk means that the user can disable it using `<UseMSTestRunner>false</UseMSTestRunner>` so we set it 2 time and the set to false doesn't re-enable the insertion of the old sdk program.* . This will fail the build...so we need to "restore the file" if user disable it. The issue is that the SDK part that insert the program.* is an init target https://github.com/microsoft/vstest/blob/6e825a1dcdb03ce95cff9c740e3f616d43d96f3a/src/package/Microsoft.NET.Test.Sdk/netcoreapp/Microsoft.NET.Test.Sdk.targets#L13 and so we cannot run some task before it to set back the `$(GeneratedProgramFile)` to true.

src/Package/MSTest.Sdk/Sdk/Sdk.props
```diff
@@ -0,0 +1,19 @@
...
+  <PropertyGroup>
+    <UseMSTestRunner>true</UseMSTestRunner>
+    <TestingPlatformDonetTestSupport>false</TestingPlatformDonetTestSupport>
+    <IsPackable>false</IsPackable>
+    <IsTestProject>true</IsTestProject>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <PackageReference Include="Microsoft.NET.Test.Sdk" Version="17.8.0" />
+    <PackageReference Include="MSTest.TestAdapter" Version="3.2.0-preview.23601.5" />
```
MarcoRossignoli: @Evangelink is it possible have this version updated at build time from versions.prop using the "template update" strategy?

MarcoRossignoli: done

PR #1884, opened 2023-12-05
Fix command line message

Evangelink:


MarcoRossignoli:
Going to close the bug so thanks

PR #1917, opened 2023-12-11
Add net7.0 and net8.0 target frameworks

Evangelink:


Romfos:
@Evangelink Hi, could you please add motivation for this change?
What is the value to target directly .net 7 and .net 8 if we already have .net 6 and no special optimizations for .net 7+.
As the result we have veeery long list of target frameworks.
No blaming, no action required, just curious for me, maybe i miss something =)

Evangelink:
Performance gain that are pretty worth it (especially for net8.0) :)

Romfos:
but, if i right understand, tfm of dependency library do not have impact on it? perf is depends on runtime?

Evangelink:
@Romfos indeed you are right. The NuGet packages we have do not provide any net7 or net8 improvements and we don't use any special API so this change is for now pointless. I will revert when I am back from holidays.

PR #1935, opened 2023-12-13
Improve test coverage for loggers

cvpoienaru:
Improve test coverage for loggers


src/Platform/Microsoft.Testing.Platform/Helpers/System/IBlockingCollection.cs
```diff
@@ -0,0 +1,15 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+#if !NETCOREAPP
+namespace Microsoft.Testing.Platform.Helpers;
+
+internal interface IBlockingCollection<T> : IDisposable
```
MarcoRossignoli: why we need to mock this types? I would not do it to not pay virtual dispatch on these core object? 
They're well tested runtime types too.

src/Platform/Microsoft.Testing.Platform/Helpers/System/IChannel.cs
```diff
@@ -0,0 +1,17 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+#if NETCOREAPP
+namespace Microsoft.Testing.Platform.Helpers;
+
+internal interface IChannel<T>
```
MarcoRossignoli: why we need to mock this types? I would not do it to not pay virtual dispatch on these core hot loop objects? 
Also this one is well tested inside the runtime.

src/Platform/Microsoft.Testing.Platform/Helpers/System/ISemaphore.cs
```diff
@@ -0,0 +1,13 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+namespace Microsoft.Testing.Platform.Helpers;
+
+internal interface ISemaphore : IDisposable
```
MarcoRossignoli: Ditto

cvpoienaru: I would argue virtual dispatch cost is insignificant, after all we're dealing with I/O when logging which should be a few orders of magnitude slower. I know they are well tested in the runtime, but it's easier and more efficient to test all code paths by replacing them with mocks rather than wait for the semaphore implementation to timeout, for example.

src/Platform/Microsoft.Testing.Platform/Helpers/System/IChannel.cs
cvpoienaru: Same as below.

src/Platform/Microsoft.Testing.Platform/Helpers/System/IBlockingCollection.cs
cvpoienaru: Same as below.

test/UnitTests/Microsoft.Testing.Platform.UnitTests/Logging/ConsoleLoggerTests.cs
```diff
@@ -0,0 +1,74 @@
...
+    private readonly Mock<IClock> _mockClock = new();
+    private readonly ConsoleLogger _consoleLogger;
+
+    public ConsoleLoggerTests(ITestExecutionContext testExecutionContext)
+        : base(testExecutionContext)
+    {
+        _mockConsole.Setup(x => x.WriteLine(It.IsAny<string>()));
+        _mockClock.Setup(x => x.UtcNow).Returns(new DateTimeOffset(new(2023, 5, 29, 3, 42, 13)));
+
+        _consoleLogger = (new ConsoleLoggerProvider(MinimumLogLevel, _mockConsole.Object, _mockClock.Object).CreateLogger(Category) as ConsoleLogger)!;
```
Evangelink: Why are you not using `ConsoleLogger` ctor?

test/UnitTests/Microsoft.Testing.Platform.UnitTests/Logging/ConsoleLoggerTests.cs
```diff
@@ -0,0 +1,74 @@
...
+        Assert.IsFalse(_consoleLogger.IsEnabled(LogLevel.Trace));
+    }
+
+    public void ConsoleLogger_Log_FormattedStringIsCorrect()
+    {
+        _consoleLogger.Log(LogLevel.Error, Message, _exception, Formatter);
+        _consoleLogger.Log(LogLevel.Information, Message, null, Formatter);
+
+        _mockConsole.Verify(
+            x => x.WriteLine($"{string.Format(CultureInfo.InvariantCulture, _expectedPrefix, LogLevel.Information)} {Formatter(Message, null)}"),
```
Evangelink: I understand reusability but it would be so much easier to read the full string to understand the expectation.

test/UnitTests/Microsoft.Testing.Platform.UnitTests/Logging/ConsoleLoggerTests.cs
```diff
@@ -0,0 +1,74 @@
...
+    }
+
+    public void ConsoleLogger_CheckEnabled()
+    {
+        Assert.IsTrue(_consoleLogger.IsEnabled(LogLevel.Information));
+        Assert.IsTrue(_consoleLogger.IsEnabled(LogLevel.Error));
+        Assert.IsFalse(_consoleLogger.IsEnabled(LogLevel.Trace));
+    }
+
+    public void ConsoleLogger_Log_FormattedStringIsCorrect()
```
Evangelink: I am not a big fan of multiple calls in the same methods because then you have no guarantee of no mix. Assume that `exception` prints `null` and vice-versa, this test would still pass.

test/UnitTests/Microsoft.Testing.Platform.UnitTests/Logging/ServerLoggerForwarderTests.cs
```diff
@@ -0,0 +1,112 @@
...
+
+#if NETCOREAPP
+    private readonly Mock<IChannel<ServerLogMessage>> _mockProducerConsumer = new();
+#else
+    private readonly Mock<IBlockingCollection<ServerLogMessage>> _mockProducerConsumer = new();
+#endif
+    private readonly Mock<IServerTestHost> _mockServerTestHost = new();
+    private readonly Mock<IProducerConsumerFactory<ServerLogMessage>> _mockProducerConsumerFactory = new();
+
+    public ServerLoggerForwarderTests(ITestExecutionContext testExecutionContext)
```
Evangelink, line 28: I think only the mocked `ServerTestHost` is needed here. We don't "care" much about the internal implementation, we just want to ensure that calling Log does push a message in the server.

test/UnitTests/Microsoft.Testing.Platform.UnitTests/Logging/ServerLoggerForwarderTests.cs
```diff
@@ -0,0 +1,112 @@
...
+
+        Assert.Throws<InvalidOperationException>(() =>
+            serverLoggerForwarder.Log(LogLevel.Information, Result2, null, Formatter));
+        _mockProducerConsumer.Verify(x => x.TryWrite(It.IsAny<ServerLogMessage>()), Times.Exactly(2));
+#else
+        _mockProducerConsumer.Verify(x => x.Add(It.IsAny<ServerLogMessage>()), Times.Once);
+#endif
+    }
+
+    public async Task ServerLoggerForwarder_LogAsync()
```
Evangelink: This should be a parameterized test with each log level.

test/UnitTests/Microsoft.Testing.Platform.UnitTests/Logging/ServerLoggerForwarderTests.cs
```diff
@@ -0,0 +1,112 @@
...
+                new(LogLevel.Information, Result2),
+            });
+        _mockProducerConsumer.Setup(x => x.Add(It.IsAny<ServerLogMessage>()));
+#endif
+
+        _mockServerTestHost.Setup(x => x.IsInitialized).Returns(true);
+        _mockServerTestHost.Setup(x => x.PushDataAsync(It.IsAny<IData>()));
+    }
+
+    public void ServerLoggerForwarder_Log()
```
Evangelink: This should be a parameterized test with each log level.

test/UnitTests/Microsoft.Testing.Platform.UnitTests/Logging/ServerLoggerForwarderTests.cs
```diff
@@ -0,0 +1,112 @@
...
+                _mockProducerConsumerFactory.Object)
+            .CreateLogger("Test") as ServerLoggerForwarder)!)
+        {
+            await serverLoggerForwarder.LogAsync(LogLevel.Trace, Result1, null, Formatter);
+            await serverLoggerForwarder.LogAsync(LogLevel.Information, Result1, null, Formatter);
+        }
+
+        _mockServerTestHost.Verify(x => x.PushDataAsync(It.IsAny<ServerLogMessage>()), Times.Exactly(2 + 1));
+    }
+}
```
Evangelink, line 82: I think you also need some test that ensure that when the server is not initialized, the logs are just lost (nothing is pushed to server).

test/UnitTests/Microsoft.Testing.Platform.UnitTests/Logging/NopLoggerTests.cs
```diff
@@ -0,0 +1,54 @@
...
+
+    private static int s_formatterCalls;
+
+    public NopLoggerTests(ITestExecutionContext testExecutionContext)
+        : base(testExecutionContext)
+    {
+        _nopLogger = new NopLogger();
+    }
+
+    public void NopLogger_CheckDisabled()
```
Evangelink: This should be a parameterized test with each log level.

test/UnitTests/Microsoft.Testing.Platform.UnitTests/Logging/NopLoggerTests.cs
```diff
@@ -0,0 +1,54 @@
...
+
+        Assert.AreEqual(0, s_formatterCalls);
+    }
+
+    public async ValueTask NopLogger_LogAsync_NoFormatterCalls()
+    {
+        await _nopLogger.LogAsync(LogLevel.Error, Message, _exception, Formatter);
+        await _nopLogger.LogAsync(LogLevel.Information, Message, null, Formatter);
+
+        Assert.AreEqual(0, s_formatterCalls);
```
Evangelink: These should also be parameterized.

test/UnitTests/Microsoft.Testing.Platform.UnitTests/Logging/LoggingExtensionsTests.cs
```diff
@@ -0,0 +1,83 @@
...
+        _mockLogger.Setup(x => x.LogAsync(LogLevel.Debug, Message, null, LoggingExtensions.Formatter));
+        _mockLogger.Setup(x => x.LogAsync(LogLevel.Information, Message, null, LoggingExtensions.Formatter));
+        _mockLogger.Setup(x => x.LogAsync(LogLevel.Warning, Message, null, LoggingExtensions.Formatter));
+        _mockLogger.Setup(x => x.LogAsync(LogLevel.Error, Message, null, LoggingExtensions.Formatter));
+        _mockLogger.Setup(x => x.LogAsync(LogLevel.Error, Message, _exception, LoggingExtensions.Formatter));
+        _mockLogger.Setup(x => x.LogAsync(LogLevel.Error, _exception.ToString(), null, LoggingExtensions.Formatter));
+        _mockLogger.Setup(x => x.LogAsync(LogLevel.Critical, Message, null, LoggingExtensions.Formatter));
+    }
+
+    public void LoggerExtensions_LogTest()
```
Evangelink: There should be one test per level. With this way of testing, if there was an inversion (let's say `Trace` calls `Debug` and `Debug` calls `Trace`) the test would pass and so won't offer safetynet.

test/UnitTests/Microsoft.Testing.Platform.UnitTests/Logging/LoggingExtensionsTests.cs
```diff
@@ -0,0 +1,83 @@
...
+        _mockLogger.Verify(x => x.Log(LogLevel.Debug, Message, null, LoggingExtensions.Formatter), Times.Once);
+        _mockLogger.Verify(x => x.Log(LogLevel.Information, Message, null, LoggingExtensions.Formatter), Times.Once);
+        _mockLogger.Verify(x => x.Log(LogLevel.Warning, Message, null, LoggingExtensions.Formatter), Times.Once);
+        _mockLogger.Verify(x => x.Log(LogLevel.Error, Message, null, LoggingExtensions.Formatter), Times.Once);
+        _mockLogger.Verify(x => x.Log(LogLevel.Error, _exception.ToString(), null, LoggingExtensions.Formatter), Times.Once);
+        _mockLogger.Verify(x => x.Log(LogLevel.Error, Message, _exception, LoggingExtensions.Formatter), Times.Once);
+        _mockLogger.Verify(x => x.Log(LogLevel.Critical, Message, null, LoggingExtensions.Formatter), Times.Once);
+    }
+
+    public async ValueTask LoggerExtensions_LogAsyncTest()
```
Evangelink: Same comment as per previous test.

test/UnitTests/Microsoft.Testing.Platform.UnitTests/Logging/LoggingExtensionsTests.cs
Evangelink: I would also suggest a different name for the test so we can understand easily what's the goal (e.g. `LogTraceAsync_CallsLogAsyncWithLogLevelTrace`).

test/UnitTests/Microsoft.Testing.Platform.UnitTests/Logging/LoggerTests.cs
```diff
@@ -0,0 +1,85 @@
...
+                new ConsoleLoggerProvider(MinimumLogLevel, _mockConsole.Object, _mockClock.Object).CreateLogger(Category),
+                new NopLogger(),
+            },
+            MinimumLogLevel);
+
+        _mockLoggerFactory.Setup(x => x.CreateLogger(It.IsAny<string>())).Returns(_logger);
+        _genericLogger = new Logger<string>(_mockLoggerFactory.Object);
+    }
+
+    public void Logger_CheckEnabled()
```
Evangelink: Use parameterized tests

Evangelink: Instead of that, I would have multiple tests that checks the >= part. For example, given `Trace` all log level return true. Given `Debug`, `TRace` returns false and the other true....

test/UnitTests/Microsoft.Testing.Platform.UnitTests/Logging/LoggerTests.cs
```diff
@@ -0,0 +1,85 @@
...
+        await _genericLogger.LogAsync(LogLevel.Error, Message, _exception, Formatter);
+        await _genericLogger.LogAsync(LogLevel.Information, Message, null, Formatter);
+
+        _mockConsole.Verify(
+            x => x.WriteLine($"{string.Format(CultureInfo.InvariantCulture, _expectedPrefix, LogLevel.Information)} {Formatter(Message, null)}"),
+            Times.Once);
+        _mockConsole.Verify(
+            x => x.WriteLine($"{string.Format(CultureInfo.InvariantCulture, _expectedPrefix, LogLevel.Error)} {Formatter(Message, _exception)}"),
+            Times.Once);
+    }
```
Evangelink: Given the name of the class, I'd expect to see tests for Logger.cs but this is not what we have here.

What we should test is that when calling `Log` or `LogAsync` we call the same api on each logger passed by ctor if and only if it is enabled for the given level.

test/UnitTests/Microsoft.Testing.Platform.UnitTests/Logging/LoggerFactoryTests.cs
```diff
@@ -0,0 +1,50 @@
...
+        _mockMonitor.Setup(x => x.Lock(It.IsAny<object>())).Returns(new Mock<IDisposable>().Object);
+        _mockLoggerProvider.Setup(x => x.CreateLogger(It.IsAny<string>())).Returns(_mockLogger.Object);
+
+        _loggerProviders = new[]
+        {
+            _mockLoggerProvider.Object,
+        };
+    }
+
+    public void LoggerFactory_WriteLogFromFactory()
```
Evangelink: This test doesn't test public API of the `LoggerFactory`. Instead, I would expect some test that checks that a logger is created only once per category (by having the dic internals for testing purpose).

Evangelink (changes requested):
A couple of global remarks:

- I think there is a bit too many types mocked (e.g. collections) that don't bring huge value.
- I would focus on testing either a few specific important internal things (singleton, lock...) or the public api functionality but not all the internal details.
- Every test with loglevel should be a parameterized test that checks all log levels.
- Avoid, as much as possible, testing multiple calls in the same test method or if you do, try to act and assert and not do many act and then multiple asserts. This is having high chance of hidding potential bugs.

test/UnitTests/Microsoft.Testing.Platform.UnitTests/Logging/LoggerTests.cs
cvpoienaru: Fixed.

cvpoienaru: Fixed, thanks for the suggestion !

test/UnitTests/Microsoft.Testing.Platform.UnitTests/Logging/ConsoleLoggerTests.cs
cvpoienaru: ConsoleLogger removed, closing.

test/UnitTests/Microsoft.Testing.Platform.UnitTests/Logging/ConsoleLoggerTests.cs
cvpoienaru: ConsoleLogger removed, closing.

test/UnitTests/Microsoft.Testing.Platform.UnitTests/Logging/ConsoleLoggerTests.cs
cvpoienaru: ConsoleLogger removed, closing.

test/UnitTests/Microsoft.Testing.Platform.UnitTests/Logging/LoggingExtensionsTests.cs
cvpoienaru: Fixed, thanks !

test/UnitTests/Microsoft.Testing.Platform.UnitTests/Logging/LoggingExtensionsTests.cs
cvpoienaru: Fixed, thanks !

cvpoienaru: Fixed, thanks !

test/UnitTests/Microsoft.Testing.Platform.UnitTests/Logging/NopLoggerTests.cs
cvpoienaru: Fixed.

test/UnitTests/Microsoft.Testing.Platform.UnitTests/Logging/NopLoggerTests.cs
cvpoienaru: Fixed.

test/UnitTests/Microsoft.Testing.Platform.UnitTests/Logging/LoggerTests.cs
cvpoienaru: Fixed, thanks.

test/UnitTests/Microsoft.Testing.Platform.UnitTests/Logging/ServerLoggerForwarderTests.cs
cvpoienaru: Fixed, thanks.

test/UnitTests/Microsoft.Testing.Platform.UnitTests/Logging/ServerLoggerForwarderTests.cs
cvpoienaru: Fixed.

test/UnitTests/Microsoft.Testing.Platform.UnitTests/Logging/ServerLoggerForwarderTests.cs
cvpoienaru: Fixed.

test/UnitTests/Microsoft.Testing.Platform.UnitTests/Logging/ServerLoggerForwarderTests.cs
cvpoienaru: Fixed.

test/UnitTests/Microsoft.Testing.Platform.UnitTests/Logging/LoggerFactoryTests.cs
cvpoienaru: Fixed, thanks.

src/Platform/Microsoft.Testing.Platform/Helpers/System/IFileSystem.cs
```diff
@@ -9,6 +9,8 @@ internal interface IFileSystem
 
     string CreateDirectory(string path);
 
+    void Move(string src, string dst);
```
Evangelink: Let's match real API naming

```suggestion
    void Move(string sourceFileName, string destFileName);
```

src/Platform/Microsoft.Testing.Platform/Helpers/System/SystemFileSystem.cs
```diff
@@ -9,6 +9,8 @@ internal sealed class SystemFileSystem : IFileSystem
 
     public string CreateDirectory(string path) => Directory.CreateDirectory(path).FullName;
 
+    public void Move(string src, string dst) => File.Move(src, dst);
```
Evangelink: Matching real API

```suggestion
    public void Move(string sourceFileName, string destFileName) => File.Move(src, dst);
```

src/Platform/Microsoft.Testing.Platform/Logging/ApplicationLoggingState.cs
```diff
@@ -8,12 +8,11 @@ namespace Microsoft.Testing.Platform.Logging;
...
 {
     public ApplicationLoggingState(LogLevel logLevel, CommandLineParseResult commandLineParseResult)
+        : this(logLevel, commandLineParseResult, null, false)
     {
-        LogLevel = logLevel;
-        CommandLineParseResult = commandLineParseResult;
     }
 
-    public ApplicationLoggingState(LogLevel logLevel, CommandLineParseResult commandLineParseResult, FileLoggerProvider fileLoggerProvider, bool isSynchronousWrite)
+    public ApplicationLoggingState(LogLevel logLevel, CommandLineParseResult commandLineParseResult, FileLoggerProvider? fileLoggerProvider, bool isSynchronousWrite)
```
Evangelink: It's changing a bit the semantic but maybe we don't mind.

src/Platform/Microsoft.Testing.Platform/Logging/FileLogger.cs
```diff
@@ -215,57 +225,51 @@ private void EnqueueLog<TState>(LogLevel logLevel, TState state, Exception? exce
...
         try
         {
+#if NETCOREAPP
             // We don't need cancellation token because the task will be stopped when the Channel is completed thanks to the call to Complete() inside the Dispose method.
             while (await _channel.Reader.WaitToReadAsync())
             {
                 await _writer.WriteLineAsync(await _channel.Reader.ReadAsync());
             }
-        }
-        catch (Exception ex)
```
Evangelink, line 230: Can you comment on why this is removed?

src/Platform/Microsoft.Testing.Platform/Logging/FileLoggerOptions.cs
```diff
@@ -0,0 +1,19 @@
...
+    string? fileName = null,
+    bool syncFlush = true)
+{
+    public string LogFolder { get; private set; } = logFolder;
+
+    public string LogPrefixName { get; private set; } = logPrefixName;
+
+    public string? FileName { get; private set; } = fileName;
+
+    public bool SyncFlush { get; private set; } = syncFlush;
```
Evangelink: I think you should be able to use readonly props:

```suggestion
    public string LogFolder { get; } = logFolder;

    public string LogPrefixName { get; } = logPrefixName;

    public string? FileName { get; } = fileName;

    public bool SyncFlush { get; } = syncFlush;
```

src/Platform/Microsoft.Testing.Platform/ServerMode/IServerTestHost.cs
```diff
@@ -1,11 +1,15 @@
...
+
 namespace Microsoft.Testing.Platform.ServerMode;
 
 internal interface IServerTestHost
 {
     bool IsInitialized { get; }
 
     Task SendTelemetryEventUpdateAsync(TelemetryEventArgs args);
+
+    Task PushDataAsync(IData data);
```
Evangelink, line 14: Can you comment on why you are changing API?

test/UnitTests/Microsoft.Testing.Platform.UnitTests/Logging/ServerLoggerForwarderTests.cs
```diff
@@ -0,0 +1,262 @@
...
+    [Arguments(LogLevel.None, LogLevel.Critical)]
+    public async Task ServerLoggerForwarder_ServerLogNotInitialized_NoLogAsyncForwarded(LogLevel defaultLevel, LogLevel currentLevel)
+    {
+        _mockServerTestHost.Setup(x => x.IsInitialized).Returns(false);
+
+        using (ServerLoggerForwarder serverLoggerForwarder = (new ServerLoggerForwarderProvider(
+                defaultLevel,
+                new SystemTask(),
+                _mockServerTestHost.Object)
+            .CreateLogger("Test") as ServerLoggerForwarder)!)
```
Evangelink: Avoid as cast with bang, if you are sure you can cast, you should force cast.

test/UnitTests/Microsoft.Testing.Platform.UnitTests/Logging/ServerLoggerForwarderTests.cs
```diff
@@ -0,0 +1,262 @@
...
+        {
+            await serverLoggerForwarder.LogAsync(currentLevel, Message, null, Formatter);
+        }
+
+        _mockServerTestHost.Verify(x => x.PushDataAsync(It.IsAny<ServerLogMessage>()), shouldLog ? Times.Once : Times.Never);
+    }
+
+    [Arguments(LogLevel.Trace, LogLevel.Trace)]
```
Evangelink: It's probably worth introducing some member that returns this combination and share it across all tests needing it. It will be a lot easier in future.

test/UnitTests/Microsoft.Testing.Platform.UnitTests/Logging/ServerLoggerForwarderTests.cs
Evangelink: Please apply to the various tests

test/UnitTests/Microsoft.Testing.Platform.UnitTests/Logging/NopLoggerTests.cs
```diff
@@ -0,0 +1,66 @@
...
+
+    private static int s_formatterCalls;
+
+    public NopLoggerTests(ITestExecutionContext testExecutionContext)
+        : base(testExecutionContext)
+    {
+        _nopLogger = new NopLogger();
+    }
+
+    [Arguments(LogLevel.Trace)]
```
Evangelink: Might be worth creating a member and using `ArgumentsProvider` instead.

test/UnitTests/Microsoft.Testing.Platform.UnitTests/Logging/NopLoggerTests.cs
```diff
@@ -0,0 +1,66 @@
...
+    private static readonly Func<string, Exception?, string> Formatter =
+        (state, exception) =>
+        {
+            ++s_formatterCalls;
+            return string.Empty;
+        };
+
+    private const string Message = "DummyMessage";
+    private readonly Exception _exception = new("TestException");
+    private readonly NopLogger _nopLogger;
```
Evangelink: Nit

```suggestion
    private readonly NopLogger _nopLogger = new();
```

test/UnitTests/Microsoft.Testing.Platform.UnitTests/Logging/NopLoggerTests.cs
```diff
@@ -0,0 +1,66 @@
...
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using Microsoft.Testing.Framework;
+using Microsoft.Testing.Platform.Logging;
+using Microsoft.Testing.TestInfrastructure;
+
+namespace Microsoft.Testing.Platform.UnitTests;
+
+[TestGroup]
+public class NopLoggerTests : TestBase
```
Evangelink: Nit:

```suggestion
public class NopLoggerTests(ITestExecutionContext testExecutionContext) : TestBase(testExecutionContext)
```

test/UnitTests/Microsoft.Testing.Platform.UnitTests/Logging/NopLoggerTests.cs
```diff
@@ -0,0 +1,66 @@
...
+    private readonly NopLogger _nopLogger;
+
+    private static int s_formatterCalls;
+
+    public NopLoggerTests(ITestExecutionContext testExecutionContext)
+        : base(testExecutionContext)
+    {
+        _nopLogger = new NopLogger();
+    }
+
```
Evangelink: Nit

```suggestion
```

test/UnitTests/Microsoft.Testing.Platform.UnitTests/Logging/LoggingExtensionsTests.cs
```diff
@@ -0,0 +1,136 @@
...
+    public LoggingExtensionsTests(ITestExecutionContext testExecutionContext)
+        : base(testExecutionContext)
+    {
+    }
+
+    public void LoggerExtensions_LogTrace_CallsLogWithLogLevelTrace()
+    {
+        _mockLogger.Setup(x => x.Log(LogLevel.Trace, Message, null, LoggingExtensions.Formatter));
+        _mockLogger.Object.LogTrace(Message);
+        _mockLogger.Verify(x => x.Log(LogLevel.Trace, Message, null, LoggingExtensions.Formatter), Times.Once);
```
Evangelink, line 25: Probably good as-is but it would be even better to check that the other are not called.

To do so you could use the strict mode of Moq (when creating the mock object) or create an helper that takes a log level and asserts all but given were not called.

test/UnitTests/Microsoft.Testing.Platform.UnitTests/Logging/LoggingExtensionsTests.cs
```diff
@@ -0,0 +1,136 @@
...
+using Microsoft.Testing.Framework;
+using Microsoft.Testing.Platform.Logging;
+using Microsoft.Testing.TestInfrastructure;
+
+using Moq;
+
+namespace Microsoft.Testing.Platform.UnitTests;
+
+[TestGroup]
+public class LoggingExtensionsTests : TestBase
```
Evangelink: Nit

```suggestion
public class LoggingExtensionsTests(ITestExecutionContext testExecutionContext) : TestBase(testExecutionContext)
```

test/UnitTests/Microsoft.Testing.Platform.UnitTests/Logging/LoggingExtensionsTests.cs
```diff
@@ -0,0 +1,136 @@
...
+{
+    private const string Message = "Test";
+    private readonly Exception _exception = new("TestException");
+
+    private readonly Mock<ILogger> _mockLogger = new();
+
+    public LoggingExtensionsTests(ITestExecutionContext testExecutionContext)
+        : base(testExecutionContext)
+    {
+    }
```
Evangelink: Nit

```suggestion
```

src/Platform/Microsoft.Testing.Platform/ServerMode/IServerTestHost.cs
Evangelink: @MarcoRossignoli are we fine with this change? I don't know if there was some plan to open this interface.

test/UnitTests/Microsoft.Testing.Platform.UnitTests/Logging/LoggerTests.cs
```diff
@@ -0,0 +1,194 @@
...
+        _mockLogger.Setup(x => x.IsEnabled(It.IsAny<LogLevel>())).Returns<LogLevel>(currentLogLevel => currentLogLevel >= logLevel);
+
+        Logger logger = new(new[] { _mockLogger.Object }, logLevel);
+
+        Mock<ILoggerFactory> mockLoggerFactory = new();
+        mockLoggerFactory.Setup(x => x.CreateLogger(It.IsAny<string>())).Returns(logger);
+        return new Logger<string>(mockLoggerFactory.Object);
+    }
+
+    [Arguments(LogLevel.Trace, LogLevel.Trace, true)]
```
Evangelink: Use a member data to ease sharing.

Evangelink (commented):
Please prefer ArgumentsProvider with members to ease reusing the same entries over the 40 lines of arguments on top of the various methods.

test/UnitTests/Microsoft.Testing.Platform.UnitTests/Logging/NopLoggerTests.cs
cvpoienaru: Fixed, thanks !

test/UnitTests/Microsoft.Testing.Platform.UnitTests/Logging/ServerLoggerForwarderTests.cs
cvpoienaru: Fixed, thanks !

src/Platform/Microsoft.Testing.Platform/ServerMode/IServerTestHost.cs
cvpoienaru: I've added this method in order to assert on the mock that it has been called, meaning a message was indeed pushed to the server.

test/UnitTests/Microsoft.Testing.Platform.UnitTests/Logging/LoggerTests.cs
cvpoienaru: Fixed, thanks.

test/UnitTests/Microsoft.Testing.Platform.UnitTests/Logging/LoggingExtensionsTests.cs
cvpoienaru: Fixed, thanks !

src/Platform/Microsoft.Testing.Platform/Logging/FileLogger.cs
cvpoienaru: It's not really removed, but there were 2 almost identical `try-catch` blocks, one for `#if NETCOREAPP` branch and one for `#else` branch. I merged the 2 blocks into one and kept conditional compilation inside the `try` block only.

test/UnitTests/Microsoft.Testing.Platform.UnitTests/Logging/ServerLoggerForwarderTests.cs
cvpoienaru: Fixed, thanks.

src/Platform/Microsoft.Testing.Platform/Logging/ApplicationLoggingState.cs
cvpoienaru: Fixed, thanks !

Evangelink (approved):

PR #1985, opened 2023-12-21
Rework MSTEST0002 and MSTEST0003

Evangelink:
MSTEST0002 now handles all valid layout for test classes 
MSTEST0003 now handles all valid layout for test methods


src/Analyzers/MSTest.Analyzers/Resources.resx
```diff
@@ -117,23 +117,57 @@
...
-    <value>Methods marked with '[TestMethod]' should be public</value>
+  <data name="TestMethodShouldBeValidDescription" xml:space="preserve">
+    <value>Test methods, methods marked with the '[TestMethod]' attribute, should respect the following layout to be considered valid by MSTest:
+- it should be 'public'
+- it should not be 'static'
+- it should not be generic
+- it should not be 'abstract'
+- return type should be 'void' or 'Task'
+- it should not be 'async void'
+- it should be a method of kind 'ordinary'.</value>
```
nohwnd: Is there a more user friendly way to put this? I am sure it is accurate, but it is not helpful (to me). Is it supposed to mean that this is not a local function or something like that?

MarcoRossignoli (dismissed):

src/Analyzers/MSTest.Analyzers/Resources.resx
nohwnd: Looking at the rule and the enum https://learn.microsoft.com/en-us/dotnet/api/microsoft.codeanalysis.methodkind?view=roslyn-dotnet-4.6.0 I can see there are a lot of options of what the method should not be (e.g. a destructor), I think it would be worth it to put this, and ignore the rest of possible, but obscure placements, like a property getter, or delegate:

- it should be an ordinary method on a class
- it should not be constructor, destructor or finalizer

Evangelink: Yeah I would need to check all places where it's valid to put the attribute (e.g. ctor doesn't work) and then lsit them. I'll do it.

src/Analyzers/MSTest.Analyzers/Resources.resx
```diff
@@ -117,23 +117,57 @@
...
-  <data name="TestMethodShouldBePublicMessageFormat" xml:space="preserve">
-    <value>Test method '{0}' should be public</value>
+  <data name="TestClassShouldBeValidTitle" xml:space="preserve">
+    <value>Test classes should have valid layout</value>
   </data>
-  <data name="TestMethodShouldBePublicTitle" xml:space="preserve">
-    <value>Methods marked with '[TestMethod]' should be public</value>
+  <data name="TestMethodShouldBeValidDescription" xml:space="preserve">
+    <value>Test methods, methods marked with the '[TestMethod]' attribute, should respect the following layout to be considered valid by MSTest:
+- it should be 'public'
```
nohwnd: It can be internal if you enable it: https://github.com/microsoft/testfx/pull/944

Evangelink: Handled now! Thanks!

MarcoRossignoli (approved):

PR #2036, opened 2024-01-02
Revert adding net7.0 and net8.0 to packages

Evangelink:
Relates to https://github.com/microsoft/testfx/pull/1917#issuecomment-1871211763


MarcoRossignoli:
why? As soon as we add new analyzers to improve or use new api we will start to benefit from the multi-tfm. What's the issue to have multi-tfm?

MarcoRossignoli (changes requested):
I don't see problem with multi-tfm here if not the size.

Evangelink:
@Romfos the runner `Microsoft.Testing.Platform` is actually having improved APIs for both net7.0 and net8.0 so it's providing quite some perf boost to have the extra additions. We could avoid the extra TFMs for the framework only but it's going to be easier to know this is already packaging everything needed so we can start using the new features.

Romfos:
it make sense if we have specific optimizations for .net 8.0. thank you for clarification

Evangelink:
Thanks for caring :)

PR #2079, opened 2024-01-10
Avoid some allocations

MarcoRossignoli:


src/Platform/Microsoft.Testing.Platform/Helpers/IEnumerableExtensions.cs
```diff
@@ -0,0 +1,23 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System.Collections;
+
+namespace Microsoft.Testing.Platform.Helpers;
+
+internal static class IEnumerableExtensions
+{
+    public static IReadOnlyCollection<T> ToReadOnlyCollectionEnumerator<T>(this IEnumerable<T> enumerable)
```
Evangelink, line 10: The method suggests that it is returning an enumerator but that's not the case. I would prefer `ToReadOnlyCollection`.

src/Platform/Microsoft.Testing.Platform/Helpers/IEnumerableExtensions.cs
```diff
@@ -0,0 +1,23 @@
...
+        => new ReadOnlyCollectionIEnumerableWrapper<T>(enumerable);
+
+    private readonly struct ReadOnlyCollectionIEnumerableWrapper<T>(IEnumerable<T> enumerable) : IReadOnlyCollection<T>
+    {
+        private readonly IEnumerable<T> _enumerable = enumerable;
+
+        public int Count => _enumerable.Count();
```
Evangelink, line 17: Isn't it better to persist the count so it's not performing the enumeration foreach everytime?

PR #2080, opened 2024-01-10
Fix display name for data row

### LINKED ISSUE #1635: TestMethod displayName is being ignored if DataRowAttribute is specified
## Describe the bug

When specifying a displayName in a TestMethod, it works as expected and the display name is picked up. 
However, if a DataRowAttribute is used on a TestMethod, the displayName of the TestMethod is being ignored.

Versions:
```
    <PackageReference Include="MSTest.TestAdapter" Version="3.0.2" />
    <PackageReference Include="MSTest.TestFramework" Version="3.0.2" />
```

## Steps To Reproduce
Provided 2 tests, both with custom displayNames, one with a DataRowAttribute, and another one without:
```
namespace UnitTesting
{
    [TestClass]
    public class UnitTests
    {
        [TestMethod("SomeCustomDisplayName")]
        public void TestMethod1()
        {
        }

        [TestMethod("SomeCustomDisplayName2")]
        [DataRow("SomeDataRow")]
        public void TestMethod2(string dataRowObject)
        {
        }
    }
}
```

## Expected behavior
Expected behavior is that the custom display name is picked up for both tests, like:

```
UnitTests (2)
--SomeCustomDisplayName
--SomeCustomDisplayName2 (SomeDataRow)
```

## Actual behavior
TestMethod2 does not pick up the custom name 

Screenshot 2023-04-18 101905

nohwnd:
Fix #1635

### DIFF
--- a/src/TestFramework/TestFramework/Attributes/DataSource/DataRowAttribute.cs
+++ b/src/TestFramework/TestFramework/Attributes/DataSource/DataRowAttribute.cs
@@ -86,6 +86,34 @@ public class DataRowAttribute : Attribute, ITestDataSource
             ? new object[] { data.AsEnumerable() }
             : data.AsEnumerable();
 
+        // On the most basic level, if we only used TestMethodAttribute the problem could be fixed here,
+        // but user can inherit from the attribute type (not a big problem we could construct it), or (worse) they can provide
+        // an override for every attribute on the class that is determined by GetTestMethodAttribute
+        // (as shown in https://github.com/microsoft/testfx/issues/1635#issuecomment-1532714683)
+        // This means that we cannot naively pick up DisplayName from the attribute and use it,
+        // because we don't hold that final instance.
+        //
+        // This whole replacement is implemented on the Adapter level (incorrectly?), and so we don't have access to the
+        // TestMethodInfo type, which holds all of the information of how the adapter resolved the type. It probably is also not
+        // our job to do some difficult logic here.
+        //
+        // So maybe the simplest solution would be to replace the return type with a result type, and return a multi part name,
+        // which can describe DisplayNameOverride and parameters. That way the upper logic knows if it should pick up the name,
+        // or use the method name, and also has the parameters.
+        //
+        // Problem with this is that it is not universal, and that this interface is public.
+        //
+        // We could also hot patch based on something else, like returning just name starting with ( from the data source,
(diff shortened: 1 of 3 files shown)


src/TestFramework/TestFramework/Attributes/DataSource/DataRowAttribute.cs
```diff
@@ -86,6 +86,34 @@ public DataRowAttribute(params object?[]? data)
...
+        // an override for every attribute on the class that is determined by GetTestMethodAttribute
+        // (as shown in https://github.com/microsoft/testfx/issues/1635#issuecomment-1532714683)
+        // This means that we cannot naively pick up DisplayName from the attribute and use it,
+        // because we don't hold that final instance.
+        //
+        // This whole replacement is implemented on the Adapter level (incorrectly?), and so we don't have access to the
+        // TestMethodInfo type, which holds all of the information of how the adapter resolved the type. It probably is also not
+        // our job to do some difficult logic here.
+        //
+        // So maybe the simplest solution would be to replace the return type with a result type, and return a multi part name,
```
nohwnd, line 100: this would be breaking change.

src/TestFramework/TestFramework/Attributes/DataSource/DataRowAttribute.cs
```diff
@@ -86,6 +86,34 @@ public DataRowAttribute(params object?[]? data)
...
             : data.AsEnumerable();
 
+        // On the most basic level, if we only used TestMethodAttribute the problem could be fixed here,
+        // but user can inherit from the attribute type (not a big problem we could construct it), or (worse) they can provide
+        // an override for every attribute on the class that is determined by GetTestMethodAttribute
+        // (as shown in https://github.com/microsoft/testfx/issues/1635#issuecomment-1532714683)
+        // This means that we cannot naively pick up DisplayName from the attribute and use it,
+        // because we don't hold that final instance.
+        //
+        // This whole replacement is implemented on the Adapter level (incorrectly?), and so we don't have access to the
```
nohwnd, line 96: incorrectly  in sense that the test framework cannot work without this completely, so it should be in the framework, not in the adapter (imho).

nohwnd:
Partially implemented the ugly fix that does not require interface change. To fully implement it would need to be used also in the other places which call get display name (assembly enumerator for sure).

PR #2084, opened 2024-01-10
Update name of extensions on InternalsVisibleTo

Evangelink:


Evangelink:
Force merging here too to unblock extensions.

Issue #704, opened 2020-04-03
AssemblyInitialize, ClassInitialize Timeout


fl3pp:
We are trying to migrate from MSTestV1 to MSTestV2. While doing so we noticed the behavior change documented in [DeltaWithMSTestV1 - Behavior change 12](https://github.com/microsoft/testfx-docs/blob/master/docs/deltaWithMSTestV1.md) which leads to some trouble. 

Unfortunately, we did not find any way to limit the execution time of the `ClassInitialize` and `AssemblyInitialize` methods.

In a comment in issue #496 there was a talk about a timeout attribute on these methods. Is there such a feature available or is there another way to set the timeouts for these methods?

As far as we've seen in the code, there is no way to limit the execution time of these methods. Because this is a feature we need (and we consider it as essential) we would be willing to go ahead and implement it by ourselves.

To avoid doing work that will have no chance to get merged into MSTestV2, we would like to know if such a feature would fit into your concept/vision for MSTestV2.

We imagine providing the following additional settings in `.runsettings` files: 

~~~~ XML
<RunSettings>
   <MSTest>
      <AssemblyInitializeTimeout>5000</AssemblyInitializeTimeout>
      <ClassInitializeTimeout>5000</ClassInitializeTimeout>
   </MSTest>
</RunSettings>
~~~~

[AB#1934903](https://devdiv.visualstudio.com/0bdbc590-a062-4c3f-b0f6-9383f67865ee/_workitems/edit/1934903)

HeroMaxPower:
I quite like this feature from a CI/CD point of view, for those who are unlucky enough to load large amounts of test data before the testing begins this would help narrow down the cause of a Test Execution failure. Since i imagine that 90% of the use for Assembly/Class Initialize is based for loading in test data or establishing connections to test resources. It seems like a good idea to limit the time this if a dependency fails. 

There has been some cases where a build or a deployment with tests simply ran forever and gives no information of the cause. However if i know a good estimate of the time test data should be loaded in then i could use this feature to make sure it doesn't exceed the time and free up resources for another build.

dennis-yemelyanov:
Totally agree with all the points here. This is a must have feature in most real life scenarios. It's great that MsTestV2 has a way to globally set a timeout for all tests, and we rely on it a lot to avoid blocking our CI/CD pipelines. However, not covering ClassInitialize/AssemblyInitialize methods limits the usefulness of this timeout setting and makes it almost useful for those who make some network calls inside these initialization methods.

[...]

PR #2085, opened 2024-01-10
Implement `AssemblyInitialize`, `ClassInitialize` timeout 

MarcoRossignoli:
Closes https://github.com/microsoft/testfx/issues/704


src/Adapter/MSTest.TestAdapter/Execution/TestAssemblyInfo.cs
```diff
@@ -134,7 +139,21 @@ public void RunAssemblyInitialize(TestContext testContext)
                 {
                     try
                     {
-                        AssemblyInitializeMethod.InvokeAsSynchronousTask(null, testContext);
+                        CancellationTokenSource? timeoutTokenSource = null;
```
Evangelink, line 142: Could you please add a comment in code that says we don't replace the CTS of the `TestContext` because it can be overrided by sub-classes which would break the behavior.

src/Adapter/MSTest.TestAdapter/Execution/TestClassInfo.cs
```diff
@@ -318,6 +326,30 @@ public void RunClassInitialize(TestContext testContext)
...
+    {
+        if (methodInfo is null)
+        {
+            return;
+        }
+
+        CancellationTokenSource? timeout = null;
+        try
+        {
+            if (ClassInitializeMethodTimeoutMilliseconds.TryGetValue(methodInfo, out int timeoutMilliseconds))
```
Evangelink, line 341: Add a similar comment here (about the override).

src/Adapter/MSTest.TestAdapter/Execution/TypeCache.cs
```diff
@@ -427,6 +427,19 @@ private TestAssemblyInfo GetAssemblyInfo(Type type)
                 if (IsAssemblyOrClassInitializeMethod<AssemblyInitializeAttribute>(methodInfo))
                 {
                     assemblyInfo.AssemblyInitializeMethod = methodInfo;
+
+                    if (_reflectionHelper.IsAttributeDefined<TimeoutAttribute>(methodInfo, false))
```
Evangelink, line 431: Is it doing any optimization to use `IsAttributeDefined` instead of doing null check on `GetAttribute`? I am asking because if you are using the `if (GetXX is {} timeoutAttribute)` then you can skip the debug assert.

test/IntegrationTests/Microsoft.Testing.Platform.Acceptance.IntegrationTests/AssemblyAndClassInitializeTimeout.cs
```diff
@@ -0,0 +1,130 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using Microsoft.Testing.Platform.Acceptance.IntegrationTests.Helpers;
```
Evangelink: Today this project contains only acceptance of the platform not MSTest. I am fine to start adding tests there but it'd be nice to rename the project into `MSTest.Acceptance.IntegrationTests`. and add a folder (no need to update namespace) for MSTest related tests and another for platform.

src/Adapter/MSTest.TestAdapter/Execution/TypeCache.cs
MarcoRossignoli: `IsAttributeDefined` goes to the cache, also the getattributes could return more values allocating more, the isdefined is a quicker contains string check.

test/IntegrationTests/Microsoft.Testing.Platform.Acceptance.IntegrationTests/AssemblyAndClassInitializeTimeout.cs
MarcoRossignoli: done

src/Adapter/MSTest.TestAdapter/Execution/TestAssemblyInfo.cs
```diff
@@ -134,7 +139,23 @@ public void RunAssemblyInitialize(TestContext testContext)
...
+                        try
+                        {
+                            // TestContext is public and the CancellationTokenSource property has got protected set accessor.
+                            // So we don't substitute the curre CTS instance but we signal the current one.
```
Evangelink:
```suggestion
                            // So we don't substitute the current CTS instance but instead we signal the current one.
```

src/Adapter/MSTest.TestAdapter/Execution/TestClassInfo.cs
```diff
@@ -318,6 +326,32 @@ public void RunClassInitialize(TestContext testContext)
...
+        if (methodInfo is null)
+        {
+            return;
+        }
+
+        CancellationTokenSource? timeout = null;
+        try
+        {
+            // TestContext is public and the CancellationTokenSource property has got protected set accessor.
+            // So we don't substitute the curre CTS instance but we signal the current one.
```
Evangelink:
```suggestion
            // So we don't substitute the current CTS instance but instead, we signal the current one.
```

test/IntegrationTests/MSTest.Acceptance.IntegrationTests/MSTest.Acceptance.IntegrationTests.csproj
```diff
@@ -0,0 +1,37 @@
...
+    <GenerateProgramFile>false</GenerateProgramFile>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <Compile Include="$(RepoRoot)test\IntegrationTests\Microsoft.Testing.Platform.Acceptance.IntegrationTests\Helpers\AcceptanceAssert.cs" Link="Helpers\AcceptanceAssert.cs" />
+    <Compile Include="$(RepoRoot)test\IntegrationTests\Microsoft.Testing.Platform.Acceptance.IntegrationTests\Helpers\AcceptanceFixture.cs" Link="Helpers\AcceptanceFixture.cs" />
+    <Compile Include="$(RepoRoot)test\IntegrationTests\Microsoft.Testing.Platform.Acceptance.IntegrationTests\Helpers\AcceptanceTestBase.cs" Link="Helpers\AcceptanceTestBase.cs" />
+  </ItemGroup>
+
+  <ItemGroup>
```
Evangelink: You should add the coverage package

test/IntegrationTests/MSTest.Acceptance.IntegrationTests/Program.cs
```diff
@@ -0,0 +1,77 @@
...
+
+using System.Diagnostics;
+using System.Runtime.InteropServices;
+
+using Microsoft.Testing.Extensions;
+using Microsoft.Testing.Framework.Configurations;
+using Microsoft.Testing.Platform.CommandLine;
+#if ENABLE_CODECOVERAGE
+using Microsoft.Testing.Platform.Extensions.CodeCoverage;
+#endif
```
Evangelink: We don't have the flag in this repo

```suggestion
using Microsoft.Testing.Extensions.CodeCoverage;
```

test/IntegrationTests/MSTest.Acceptance.IntegrationTests/Program.cs
```diff
@@ -0,0 +1,77 @@
...
+using Microsoft.Testing.Platform.Extensions.TestHost;
+
+ITestApplicationBuilder builder = await TestApplication.CreateBuilderAsync(args);
+builder.TestHost.AddTestApplicationLifecycleCallbacks(sp => new GlobalTasks(sp.GetCommandLineOptions()));
+
+builder.AddTestFramework(new TestFrameworkConfiguration(Debugger.IsAttached ? 1 : Environment.ProcessorCount), new SourceGeneratedTestNodesBuilder());
+#if ENABLE_CODECOVERAGE
+builder.AddCodeCoverage();
+#endif
```
Evangelink:
```suggestion
builder.AddCodeCoverageProvider();
```

test/IntegrationTests/MSTest.Acceptance.IntegrationTests/AssemblyAndClassInitializeTimeout.cs
```diff
@@ -0,0 +1,131 @@
...
+    {
+        _acceptanceFixture = acceptanceFixture;
+    }
+
+    [ArgumentsProvider(nameof(GetBuildMatrixTfmBuildConfiguration))]
+    public async Task AssemblyInitialize_Timeout_Should_Stop_Execution(string tfm, BuildConfiguration buildConfiguration)
+    {
+        using TestAsset testAsset = await TestAsset.GenerateAssetAsync(_assetName, SourceCode
+            .PatchCodeWithReplace("$TargetFramework$", tfm)
+            .PatchCodeWithReplace("$MSTestVersion$", MSTestVersion), addPublicFeeds: true);
```
Evangelink: I don't think this is needed

```suggestion
            .PatchCodeWithReplace("$MSTestVersion$", MSTestVersion));
```

test/IntegrationTests/MSTest.Acceptance.IntegrationTests/AssemblyAndClassInitializeTimeout.cs
```diff
@@ -0,0 +1,131 @@
...
+    }
+
+    protected const string SourceCode = """
+#file AssemblyAndClassInitializeTimeout.csproj
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <OutputType>Exe</OutputType>
+    <EnableMSTestRunner>true</EnableMSTestRunner>
+    <PlatformTarget>x64</PlatformTarget>
```
Evangelink: Is this needed?

test/IntegrationTests/MSTest.Acceptance.IntegrationTests/AssemblyAndClassInitializeTimeout.cs
```diff
@@ -0,0 +1,131 @@
...
+        DotnetMuxerResult result = await DotnetCli.RunAsync($"build -nodeReuse:false {testAsset.TargetAssetPath} -c {buildConfiguration}", _acceptanceFixture.NuGetGlobalPackagesFolder.Path);
+        var testHost = TestHost.LocateFrom(testAsset.TargetAssetPath, _assetName, tfm, buildConfiguration: buildConfiguration);
+
+        foreach (string initMethodToTimeout in new[] { "TIMEOUT_CLASSINIT", "TIMEOUT_ASSEMBLYINIT", "TIMEOUT_BASE_CLASSINIT" })
+        {
+            var testhostResult = await testHost.ExecuteAsync(environmentVariables: new Dictionary<string, string>() { { initMethodToTimeout, "1" } });
+
+            if (initMethodToTimeout == "TIMEOUT_CLASSINIT")
+            {
+                testhostResult.AssertOutputContains("Class Initialization method TestClass.ClassInit threw exception. System.Threading.Tasks.TaskCanceledException: A task was canceled..");
```
Evangelink: Could you double check if we don't print a different custom message for the case of timeout in a test method? If we do, we may want to do something similar on the cases here.

Evangelink: I don't recall if `AssertOutputContains` allows for custom message. If it does, it might be worth specifying the timeout mode in which this happened.

test/IntegrationTests/MSTest.Acceptance.IntegrationTests/AssemblyAndClassInitializeTimeout.cs
```diff
@@ -0,0 +1,131 @@
...
+    [ArgumentsProvider(nameof(GetBuildMatrixTfmBuildConfiguration))]
+    public async Task AssemblyInitialize_Timeout_Should_Stop_Execution(string tfm, BuildConfiguration buildConfiguration)
+    {
+        using TestAsset testAsset = await TestAsset.GenerateAssetAsync(_assetName, SourceCode
+            .PatchCodeWithReplace("$TargetFramework$", tfm)
+            .PatchCodeWithReplace("$MSTestVersion$", MSTestVersion), addPublicFeeds: true);
+        DotnetMuxerResult result = await DotnetCli.RunAsync($"build -nodeReuse:false {testAsset.TargetAssetPath} -c {buildConfiguration}", _acceptanceFixture.NuGetGlobalPackagesFolder.Path);
+        var testHost = TestHost.LocateFrom(testAsset.TargetAssetPath, _assetName, tfm, buildConfiguration: buildConfiguration);
+
+        foreach (string initMethodToTimeout in new[] { "TIMEOUT_CLASSINIT", "TIMEOUT_ASSEMBLYINIT", "TIMEOUT_BASE_CLASSINIT" })
```
Evangelink: I am wondering if the resource creation should be done once only in some init and then you would have 3 tests that simply executes the assembly with different env var set. This would simplify understand failed case and make test reading easier.

test/IntegrationTests/MSTest.Acceptance.IntegrationTests/MSTest.Acceptance.IntegrationTests.csproj
MarcoRossignoli: I'll follow-up in next PR

test/IntegrationTests/MSTest.Acceptance.IntegrationTests/Program.cs
MarcoRossignoli: I'll follow-up in next PR

test/IntegrationTests/MSTest.Acceptance.IntegrationTests/Program.cs
MarcoRossignoli: ditto

test/IntegrationTests/MSTest.Acceptance.IntegrationTests/AssemblyAndClassInitializeTimeout.cs
MarcoRossignoli: nope copy paste from past tests where on netfx we got x86 but looks like no more the case.

test/IntegrationTests/MSTest.Acceptance.IntegrationTests/MSTest.Acceptance.IntegrationTests.csproj
```diff
@@ -0,0 +1,37 @@
...
+    <PackageReference Include="Microsoft.Testing.Extensions.Trx" Version="$(MicrosoftTestingPlatformVersion)" />
```
Evangelink: This should be `.TrxReport` once rebased.

test/IntegrationTests/MSTest.Acceptance.IntegrationTests/AssemblyAndClassInitializeTimeout.cs
MarcoRossignoli: discussed offline we will keep as-is for now

MarcoRossignoli: We will split in different tests so argument or test name will help us to understand the type of timeout failing

eng/Build.props
```diff
@@ -10,6 +10,7 @@
     <ProjectToBuild Include="$(RepoRoot)test/UnitTests/Microsoft.Testing.Platform.UnitTests/Microsoft.Testing.Platform.UnitTests.csproj" />
     <ProjectToBuild Include="$(RepoRoot)test/IntegrationTests/MSTest.Acceptance.IntegrationTests/MSTest.Acceptance.IntegrationTests.csproj" />
     <ProjectToBuild Include="$(RepoRoot)test/IntegrationTests/Microsoft.Testing.Platform.Acceptance.IntegrationTests/Microsoft.Testing.Platform.Acceptance.IntegrationTests.csproj" />
+    <ProjectToBuild Include="$(RepoRoot)test/IntegrationTests/MSTest.Acceptance.IntegrationTests/MSTest.Acceptance.IntegrationTests.csproj" />
```
Evangelink: Not needed, it's present on line 11

```suggestion
```

test/IntegrationTests/MSTest.Acceptance.IntegrationTests/AssemblyAndClassInitializeTimeout.cs
```diff
@@ -0,0 +1,141 @@
...
+    // There's a bug in TAFX where we need to use it at least one time somewhere to use it inside the fixture self (AcceptanceFixture).
+    public AssemblyAndClassInitializeTimeout(ITestExecutionContext testExecutionContext, TestAssetFixture testAssetFixture,
+        AcceptanceFixture globalFixture)
+        : base(testExecutionContext)
+    {
+        _testAssetFixture = testAssetFixture;
+    }
+
+    [ArgumentsProvider(nameof(TargetFrameworks.All), typeof(TargetFrameworks))]
+    public async Task AssemblyInitialize_Timeout_For_ClassInitialize_Should_Fail(string tfm)
```
Evangelink:
```suggestion
    public async Task ClassInitialize_WhenTimeoutExpires_ClassInitializeIsCancelled(string tfm)
```

test/IntegrationTests/MSTest.Acceptance.IntegrationTests/AssemblyAndClassInitializeTimeout.cs
```diff
@@ -0,0 +1,141 @@
...
+    {
+        var testHost = TestHost.LocateFrom(_testAssetFixture.MSTestTargetAssetPath, TestAssetFixture.MSTestAssetName, tfm, buildConfiguration: BuildConfiguration.Release);
+        var testhostResult = await testHost.ExecuteAsync(environmentVariables: new Dictionary<string, string>() { { "TIMEOUT_CLASSINIT", "1" } });
+        testhostResult.AssertOutputContains("Class Initialization method TestClass.ClassInit threw exception. System.Threading.Tasks.TaskCanceledException: A task was canceled..");
+    }
+
+    [ArgumentsProvider(nameof(TargetFrameworks.All), typeof(TargetFrameworks))]
+    public async Task AssemblyInitialize_Timeout_For_ClassInitialize_With_BaseClass_Should_Fail(string tfm)
```
Evangelink:
```suggestion
    public async Task BaseClassInitialize_WhenTimeoutExpires_ClassInitializeIsCancelled(string tfm)
```

test/IntegrationTests/MSTest.Acceptance.IntegrationTests/AssemblyAndClassInitializeTimeout.cs
```diff
@@ -0,0 +1,141 @@
...
+    {
+        var testHost = TestHost.LocateFrom(_testAssetFixture.MSTestTargetAssetPath, TestAssetFixture.MSTestAssetName, tfm, buildConfiguration: BuildConfiguration.Release);
+        var testhostResult = await testHost.ExecuteAsync(environmentVariables: new Dictionary<string, string>() { { "TIMEOUT_BASE_CLASSINIT", "1" } });
+        testhostResult.AssertOutputContains("Class Initialization method TestClass.ClassInitBase threw exception. System.Threading.Tasks.TaskCanceledException: A task was canceled..");
+    }
+
+    [ArgumentsProvider(nameof(TargetFrameworks.All), typeof(TargetFrameworks))]
+    public async Task AssemblyInitialize_Timeout_For_AssemblyInitialize_With_BaseClass_Should_Fail(string tfm)
```
Evangelink:
```suggestion
    public async Task AssemblyInitialize_WhenTimeoutExpires_AssemblyInitializeIsCancelled(string tfm)
```

test/IntegrationTests/MSTest.Acceptance.IntegrationTests/AssemblyAndClassInitializeTimeout.cs
```diff
@@ -0,0 +1,141 @@
...
+
+public class TestClassBase
+{
+    [ClassInitialize(inheritanceBehavior: InheritanceBehavior.BeforeEachDerivedClass)]
+    [Timeout(2000)]
+    public static async Task ClassInitBase(TestContext testContext)
+    {
+        if (Environment.GetEnvironmentVariable("TIMEOUT_BASE_CLASSINIT") == "1")
+        {
+            await Task.Delay(10_000, testContext.CancellationTokenSource.Token);
```
Evangelink, line 95: We probably want to fail test if cancellation doesn't happen

```suggestion
            await Task.Delay(10_000, testContext.CancellationTokenSource.Token);
            Assert.Fail("This should not be called");
```

test/IntegrationTests/MSTest.Acceptance.IntegrationTests/AssemblyAndClassInitializeTimeout.cs
```diff
@@ -0,0 +1,141 @@
...
+[TestClass]
+public class TestClass : TestClassBase
+{
+    [AssemblyInitialize]
+    [Timeout(2000)]
+    public static async Task AssemblyInit(TestContext testContext)
+    {
+        if (Environment.GetEnvironmentVariable("TIMEOUT_ASSEMBLYINIT") == "1")
+        {
+            await Task.Delay(60_000, testContext.CancellationTokenSource.Token);
```
Evangelink, line 114: We probably want to fail test if cancellation doesn't happen

```suggestion
            await Task.Delay(60_000, testContext.CancellationTokenSource.Token);
            Assert.Fail("This should not be called");
```

test/IntegrationTests/MSTest.Acceptance.IntegrationTests/AssemblyAndClassInitializeTimeout.cs
```diff
@@ -0,0 +1,141 @@
...
+        }
+    }
+
+    [ClassInitialize]
+    [Timeout(2000)]
+    public static async Task ClassInit(TestContext testContext)
+    {
+        if (Environment.GetEnvironmentVariable("TIMEOUT_CLASSINIT") == "1")
+        {
+            await Task.Delay(60_000, testContext.CancellationTokenSource.Token);
```
Evangelink, line 128: We probably want to fail test if cancellation doesn't happen

```suggestion
            await Task.Delay(60_000, testContext.CancellationTokenSource.Token);
            Assert.Fail("This should not be called");
```

Evangelink (commented):
Last round and then we can merge :)

Evangelink (approved):

PR #2112, opened 2024-01-16
Add code coverage

MarcoRossignoli:


eng/CodeCoverage.props
```diff
@@ -0,0 +1,10 @@
+<Project>
+  <PropertyGroup>
+    <EnableCodeCoverage>true</EnableCodeCoverage>
```
Evangelink: I think there should be a condition otherwise this is always true and overriding project config.

eng/CodeCoverage.props
```diff
@@ -0,0 +1,10 @@
+<Project>
```
Evangelink: Could we have some auto-import in Directory.build.props or target with some condition to filter out vstest like projects(e.g. `UseVsTestRunner`).

eng/CodeCoverage.props
MarcoRossignoli: I think that we disable bulk for all project if we disable or? Like we did for breaking changes
Do you want the chance to override per test runner project?

eng/CodeCoverage.props
MarcoRossignoli: I was wondering that this props make sense only in the context of test runner empowered tests and to avoid to have mixed props or possible collision till we're mixed runner for arcade, but if you want we can  have shared, looks to me for now more "clear" and explicit that it's "for this project" only.

`UseVsTestRunner` is at the moment specified at test project level so or we add some import of some file from tests folder but looks to me that we're in mixed mode so some knob at project level is needed(or move from eng to tests with some "different name" to avoid the porting everywhere).

eng/CodeCoverage.props
Evangelink: Fine for me to disable globally only.

eng/CodeCoverage.props
Evangelink: We currently default to saying all tests use vstest runner and we specify on case by case when we want to use TA runner (see https://github.com/microsoft/testfx/blob/main/Directory.Build.Local.props#L65-L66 for default config). If you import the code coverage props in the directory build local targets with the condition on vstest runner it should work, or am I missing something?

MarcoRossignoli: I think it will work correctly it's fine we should not clash any msbuild important properties, I'll import in the main dir props with the `UseVsTestRunner = false`

Issue #2121, opened 2024-01-18
DataTestMethodAttribute doesn't have a constructor to set display name


HannoZ:
## Summary
The TestMethodAttribute has a constructor that allows you to set the DisplayName property.  DataTestMethodAttribute inherits from this class, but doesn't have the constructor to provide a display name, so it's impossible to provide an alternative display name for 

## Background and Motivation
Providing a display name for unit tests is a nice way to have a more descriptive display for your tests. However it is now not possible to do this for DataTestMethod unit tests because the required constructor is not present.  
DataRowAttribute does have the ability to set the DisplayName which is nice, but it would be great to also be able to set it for the test itself.

## Proposed Feature
I guess it's just a matter of adding 
`    public DataTestMethodAttribute(string? displayName)
        : base(displayName)
    {
    }`
(I'm still trying to build the repository without success so I cannot verify this myself yet)

Evangelink:
Hi @HannoZ,

Thank you for the report. This indeeds looks like the feature missed some cases, we will work on it (except if you are willing to take care of it).

What's the issue you are facing with building the repository?

HannoZ:
I would be glad to fix it myself, if I can manage to build the solution. There's a lot of advanced stuff in there so it's not very easy to troubleshoot. First I was having some issues with the .NET 9 alpha SDK, it seemed to be installed by running Build.cmd but it wasn't. So (after some searching) I found the installer on https://github.com/dotnet/installer. That helped to solve a lot of build errors, but there is one left in project FxExtensibility: _"The analyzer assembly 'C:\Program Files\dotnet\sdk\9.0.100-alpha.1.24067.14\Sdks\Microsoft.NET.Sdk\codestyle\cs\Microsoft.CodeAnalysis.CSharp.CodeStyle.dll' references version '4.10.0.0' of the compiler, which is newer than the currently running version '4.8.0.0'."_
And there is also an issue with uap10.0.16299  which causes a compile error in UWP_UITestMethodAttribute.cs

Evangelink:
Thanks! I'll try to improve the build doc page.

[...]

PR #2125, opened 2024-01-18
Fix issue #2121 DataTestMethodAttribute is missing a constructor

HannoZ:
A fix for issue #2121 DataTestMethodAttribute doesn't have a constructor to set display name 

I added the constructor and extended the TypeEnumerator logic to get the DisplayName for DataTestMethodAttribute (only if no regular TestMethodAttribute was found) 
Renamed an existing unit test for TestMethodAttribute  and added a new test for DataTestMethodAttribute that verifies that the DisplayName is set.

### DIFF
--- a/src/Adapter/MSTest.TestAdapter/Discovery/TypeEnumerator.cs
+++ b/src/Adapter/MSTest.TestAdapter/Discovery/TypeEnumerator.cs
@@ -199,6 +199,13 @@ internal class TypeEnumerator
         var testMethodAttribute = _reflectHelper.GetCustomAttribute<TestMethodAttribute>(method);
         testElement.DisplayName = testMethodAttribute?.DisplayName ?? method.Name;
 
+        // OR get DisplayName from DataTestMethodAttribute
+        if (testMethodAttribute == null)
+        {
+            var dataTestMethodAttribute = _reflectHelper.GetCustomAttribute<DataTestMethodAttribute>(method);
+            testElement.DisplayName = dataTestMethodAttribute?.DisplayName ?? method.Name;
+        }
+
         return testElement;
     }
 }
--- a/src/TestFramework/TestFramework/Attributes/DataSource/DataTestMethodAttribute.cs
+++ b/src/TestFramework/TestFramework/Attributes/DataSource/DataTestMethodAttribute.cs
@@ -9,4 +9,21 @@ namespace Microsoft.VisualStudio.TestTools.UnitTesting;
 [AttributeUsage(AttributeTargets.Method, AllowMultiple = false)]
 public class DataTestMethodAttribute : TestMethodAttribute
 {
+    /// <summary>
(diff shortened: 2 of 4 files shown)


HannoZ:
@microsoft-github-policy-service agree [company="Nextens"]

src/Adapter/MSTest.TestAdapter/Discovery/TypeEnumerator.cs
```diff
@@ -199,6 +199,13 @@ internal UnitTestElement GetTestFromMethod(MethodInfo method, bool isDeclaredInT
         var testMethodAttribute = _reflectHelper.GetCustomAttribute<TestMethodAttribute>(method);
         testElement.DisplayName = testMethodAttribute?.DisplayName ?? method.Name;
 
+        // OR get DisplayName from DataTestMethodAttribute
+        if (testMethodAttribute == null)
+        {
+            var dataTestMethodAttribute = _reflectHelper.GetCustomAttribute<DataTestMethodAttribute>(method);
+            testElement.DisplayName = dataTestMethodAttribute?.DisplayName ?? method.Name;
+        }
```
Evangelink: I will run some test locally but this part should not be required. As far as I remember `GetCustomAttribute` is looking for inherited attributes so should take `DataTestMethod` or any custom child you would create.

HannoZ: You're probably right. It's actually a flag that you can set: https://learn.microsoft.com/en-us/dotnet/api/system.reflection.memberinfo.getcustomattributes?view=net-8.0#system-reflection-memberinfo-getcustomattributes(system-type-system-boolean)
The issue here is the mocked ReflectHelper, you can only Setup the call for a certain attribute, it isn't actually doing the reflection call. So in a real scenario it will work, but it would not be possible to write a test to verify this, or at least not at the point where I added the test. (I'm not familiar with all the code so you may have a better suggestion)

Evangelink: Sorry got busy with quite a few things. I was able to confirm that this part is not needed and that a manual test confirms this works. I also saw that we are using mocks who are causing issue in the unit test. I'll have a look during the weekend or early next week.

Evangelink (approved):

src/Adapter/MSTest.TestAdapter/Discovery/TypeEnumerator.cs
Evangelink: Ok I did a quick (dirty change) this is not ideal test but I don't think the mocking is really best test we can do. As we are super close to release date, I will ship as-is and later I will work on adding some integration test for that.

Evangelink:
Thanks for the contribution @HannoZ :shipit:

Issue #2134, opened 2024-01-19
Move `--diagnostic` default verbosity to `Trace`


MarcoRossignoli:
Normal is not so useful...and if Trace it's heavy(shouldn't it's async) we can suggest to update the verbosity.

PR #2135, opened 2024-01-20
Set default verbosity to `Trace`

MarcoRossignoli:
closes https://github.com/microsoft/testfx/issues/2134


Evangelink:
@MarcoRossignoli it would be good if you could also submit a PR on docs website.

Issue #2145, opened 2024-01-22
Improve design of ServerLoggerForwarder and ServerTestHost


Evangelink:
The `ServerLoggerForwarder` object needs to be instantiated early to be passed in the various logger locations. But it can be only truely effective when the `ServerTestHost` is up and running. Today we rely on each logging call to try to request the testhost and until it exists all messages are lost.

We should improve this design by ensuring all logs are stored until they can be sent to the client. We should also explicitely call some initialize method on the forwarder once the testhost is up and running.

PR #2149, opened 2024-01-23
Improve design of ServerLoggerForwarder and ServerTestHost

MarcoRossignoli:
contributes to https://github.com/microsoft/testfx/issues/2145

will followup with tests for the `ServerLoggerForwarderProvider`


src/Platform/Microsoft.Testing.Platform/Logging/InMemoryStoreLogger.cs
```diff
@@ -0,0 +1,25 @@
...
+    where T : class
+{
+    private readonly LogLevel _logLevel;
+
+    public InMemoryStoreLogger(LogLevel logLevel)
+    {
+        _logLevel = logLevel;
+    }
+
+    public ConcurrentBag<T> Values { get; private set; } = new();
```
Evangelink:
```suggestion
    public ConcurrentBag<T> Values { get; } = new();
```

Evangelink: Shall it be only protected?

src/Platform/Microsoft.Testing.Platform/Logging/ServerLogMessageInMemoryStore.cs
```diff
@@ -0,0 +1,40 @@
...
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+namespace Microsoft.Testing.Platform.Logging;
+
+internal sealed class ServerLogMessageInMemoryStore : InMemoryStoreLogger<ServerLogMessage>
+{
+    public ServerLogMessageInMemoryStore(LogLevel logLevel)
+        : base(logLevel)
+    {
+    }
```
Evangelink: Nit (not mandatory)

```suggestion
internal sealed class ServerLogMessageInMemoryStore(LogLevel logLevel) : InMemoryStoreLogger<ServerLogMessage>(logLevel)
{
```

Evangelink (dismissed):

src/Platform/Microsoft.Testing.Platform/Logging/InMemoryStoreLogger.cs
MarcoRossignoli: mmm needs to be accessed from outside, what you mean?

MarcoRossignoli: like `IEnum<T>` on the type and make it protected?

MarcoRossignoli: We can also remove this type and implement it only for the server mode...maybe it's too much have this abstraction.

Evangelink (approved):

PR #2167, opened 2024-01-25
Fix server mode `PerRequestServerDataConsumer`

MarcoRossignoli:


src/Platform/Microsoft.Testing.Platform/Hosts/IServerTestHost.cs
```diff
@@ -12,5 +12,7 @@ internal interface IServerTestHost
 
     Task SendTelemetryEventUpdateAsync(TelemetryEventArgs args);
 
+    Task SendTestUpdateAsync(TestNodeStateChangedEventArgs update);
```
drognanar, line 13: do we still need the IsInitialized property above? or is it the case that if the class has access to ServerTestHost it will always be initialized?

src/Platform/Microsoft.Testing.Platform/ServerMode/PerRequestServerDataConsumerService.cs
```diff
@@ -184,10 +186,10 @@ private async Task SendTestNodeUpdatesIfNecessaryAsync(Guid runId, CancellationT
                 _nodeAggregatorSemaphore.Release();
             }
 
-            ServerTestHost? server = _serviceProvider.GetService<ServerTestHost>();
-            if (change is not null && (server?.IsInitialized == true))
+            if (change is not null)
             {
-                await server.SendTestUpdateAsync(change);
+                ArgumentGuard.IsNotNull(_serverTestHost);
```
drognanar: do we need argument guards if the field is part of the constructor and not nullable?

drognanar: if it's set to null on dispose shouldn't we still null check `_serverTestHost.SendTestUpdateAsync`, in case the argument guard call is made right before the dispose sets the _serverTestHost to null?

src/Platform/Microsoft.Testing.Platform/Hosts/IServerTestHost.cs
MarcoRossignoli: let me check

src/Platform/Microsoft.Testing.Platform/ServerMode/PerRequestServerDataConsumerService.cs
MarcoRossignoli: Checking...imo this class is too complex...we should simply enqueue to the host and this class should have simple lifecycle of the request...here we have internal async to sync with root object passed...it's not needed.

src/Platform/Microsoft.Testing.Platform/Hosts/IServerTestHost.cs
MarcoRossignoli: removed

PR #2190, opened 2024-01-29
MSTEST0011: [TestProperty] can only be set on methods marked with [TestMethod] 

mariam-abdulla:


Evangelink:
@mariam-abdulla Can you please rebase to use the helper from #2176 (and use naming similar to what I did)

Evangelink:
As discussed, we will close this rule and provide a single rule that would check all the attributes that should only be set on test methods.

Thanks for the contribution @mariam-abdulla

PR #2191, opened 2024-01-29
Ensure proper usage of Guard helpers

Evangelink:
`ArgumentGuard` is made to check only arguments, for internal state, we should use `ApplicationStateGuard` instead.


Evangelink:
Force merging.

PR #2208, opened 2024-01-30
Add support for ValueTask return type

### LINKED ISSUE #2106: Allow `ValueTask` as return type for test methods and others
## Summary

Most of the test methods/init/cleanup allow multiple variantes (`void` or `Task`), let's extend this to support also `ValueTask`.

## Background and Motivation

It's possible to reduce furthermore allocations (for specific scenario) by having `ValueTask` instead of `Task`.

## Proposed Feature

Test method, init and cleanup methods should allow `ValueTask` as return type. Analyzers should be updated to match this new pattern.

## Alternative Designs

None

Evangelink:
Assembly/Class/Test Initialize/Cleanup and TestMethods now support `ValueTask` as return type.

Fixes #2106


cvpoienaru (dismissed):
LGTM

Evangelink:
I just realized that I need to update the analyzers to support htis case too.

PR #2221, opened 2024-01-31
MSTEST0007: [<Test Attributes>] can only be set on methods marked with [TestMethod]

cvpoienaru:
MSTEST0007: [<Test Attributes>] can only be set on methods marked with [TestMethod]


src/Analyzers/MSTest.Analyzers/Helpers/DiagnosticIds.cs
```diff
@@ -11,4 +11,5 @@ internal static class DiagnosticIds
     public const string PublicTypeShouldBeTestClassRuleId = "MSTEST0004";
     public const string TestContextShouldBeValidRuleId = "MSTEST0005";
     public const string AvoidExpectedExceptionAttributeRuleId = "MSTEST0006";
+    public const string AttributeOnTestMethodRuleId = "MSTEST0007";
```
Evangelink: It should be `UseAttributeOnTestMethodRuleId`.

src/Analyzers/MSTest.Analyzers/Helpers/WellKnownTypeNames.cs
```diff
@@ -8,15 +8,20 @@ internal static class WellKnownTypeNames
...
     public const string MicrosoftVisualStudioTestToolsUnitTestingTestClassAttribute = "Microsoft.VisualStudio.TestTools.UnitTesting.TestClassAttribute";
     public const string MicrosoftVisualStudioTestToolsUnitTestingTestCleanupAttribute = "Microsoft.VisualStudio.TestTools.UnitTesting.TestCleanupAttribute";
     public const string MicrosoftVisualStudioTestToolsUnitTestingTestContext = "Microsoft.VisualStudio.TestTools.UnitTesting.TestContext";
     public const string MicrosoftVisualStudioTestToolsUnitTestingTestInitializeAttribute = "Microsoft.VisualStudio.TestTools.UnitTesting.TestInitializeAttribute";
     public const string MicrosoftVisualStudioTestToolsUnitTestingTestMethodAttribute = "Microsoft.VisualStudio.TestTools.UnitTesting.TestMethodAttribute";
+    public const string MicrosoftVisualStudioTestToolsUnitTestingTestPropertyAttribute = "Microsoft.VisualStudio.TestTools.UnitTesting.TestPropertyAttribute";
+    public const string MicrosoftVisualStudioTestToolsUnitTestingWorkItemAttribute = "Microsoft.VisualStudio.TestTools.UnitTesting.WorkItemAttribute";
```
Evangelink, line 27: This is missing `CssIterationAttribute` and `CssProjectStructureAttribute`.

src/Analyzers/MSTest.Analyzers/UseAttributeOnTestMethodAnalyzer.cs
```diff
@@ -0,0 +1,162 @@
...
+
+namespace MSTest.Analyzers;
+
+[DiagnosticAnalyzer(LanguageNames.CSharp, LanguageNames.VisualBasic)]
+public sealed class UseAttributeOnTestMethodAnalyzer : DiagnosticAnalyzer
+{
+    internal static readonly DiagnosticDescriptor OwnerRule = DiagnosticDescriptorHelper.Create(
+        DiagnosticIds.AttributeOnTestMethodRuleId,
+        title: new LocalizableResourceString(
+            nameof(Resources.UseAttributeOnTestMethodAnalyzerTitle), Resources.ResourceManager, typeof(Resources), GetShortAttributeName(WellKnownTypeNames.MicrosoftVisualStudioTestToolsUnitTestingOwnerAttribute)),
```
Evangelink: Oh interesting, I wasn't aware we could do that!

src/Analyzers/MSTest.Analyzers/UseAttributeOnTestMethodAnalyzer.cs
```diff
@@ -0,0 +1,162 @@
...
+        messageFormat: new LocalizableResourceString(
+            nameof(Resources.UseAttributeOnTestMethodAnalyzerMessageFormat), Resources.ResourceManager, typeof(Resources), GetShortAttributeName(WellKnownTypeNames.MicrosoftVisualStudioTestToolsUnitTestingOwnerAttribute)),
```
Evangelink: It's quite some extra complexity to compute the same split over and over again. I maintain my initial though that the sharing part is adding more complexity that it brings simplicity.

src/Analyzers/MSTest.Analyzers/UseAttributeOnTestMethodAnalyzer.cs
```diff
@@ -0,0 +1,162 @@
...
+        (WellKnownTypeNames.MicrosoftVisualStudioTestToolsUnitTestingOwnerAttribute, OwnerRule),
+        (WellKnownTypeNames.MicrosoftVisualStudioTestToolsUnitTestingPriorityAttribute, PriorityRule),
+        (WellKnownTypeNames.MicrosoftVisualStudioTestToolsUnitTestingTestPropertyAttribute, TestPropertyRule),
+        (WellKnownTypeNames.MicrosoftVisualStudioTestToolsUnitTestingWorkItemAttribute, WorkItemRule),
+        (WellKnownTypeNames.MicrosoftVisualStudioTestToolsUnitTestingDescriptionAttribute, DescriptionRule),
+        (WellKnownTypeNames.MicrosoftVisualStudioTestToolsUnitTestingExpectedExceptionAttribute, ExpectedExceptionRule),
+    };
+
+    public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics { get; }
+        = RuleTuples.Select(tuple => tuple.Rule).ToImmutableArray();
```
Evangelink: They all have the same Id so I would go with only declaring one rule here, it's far more perfomant efficient.

```suggestion
        = ImmutableArray.Create(OwnerRule);
```

src/Analyzers/MSTest.Analyzers/Resources.resx
```diff
@@ -218,6 +218,12 @@
   <data name="TestMethodShouldBeValidTitle" xml:space="preserve">
     <value>Test methods should have valid layout</value>
   </data>
+  <data name="UseAttributeOnTestMethodAnalyzerMessageFormat" xml:space="preserve">
+    <value>[{0}] should be set on test method</value>
```
nohwnd: Use the new wording please: [<Test Attributes>] can only be set on methods marked with [TestMethod]

src/Analyzers/MSTest.Analyzers/UseAttributeOnTestMethodAnalyzer.cs
nohwnd: I agree that this is pretty complex for a rule that most users will never see fail, but will always run in the background. Please either ensure that it does no allocations (e.g. the .Split to figure out a known string), or simplify this rule. I would be totally happy with reporting a single line that lists all the attributes that are wrong. Or say "Attribute can only be used on [TestMethod]." and link to help that lists all the attributes.

I would lean on Amaury's expertise here to optimize this :)

src/Analyzers/MSTest.Analyzers/Helpers/DiagnosticIds.cs
cvpoienaru: Fixed, thanks.

src/Analyzers/MSTest.Analyzers/Helpers/WellKnownTypeNames.cs
cvpoienaru: Good catch, fixed now ! Thanks !

src/Analyzers/MSTest.Analyzers/Resources.resx
cvpoienaru: Fixed, thanks.

src/Analyzers/MSTest.Analyzers/UseAttributeOnTestMethodAnalyzer.cs
cvpoienaru: Removed the split and instead hardcoded the attribute short name string, as per my offline discussion with Amaury.

Evangelink:
@cvpoienaru Could you please submit a PR to dotnet/docs to add the doc related to this analyzer: https://learn.microsoft.com/en-us/dotnet/core/testing/unit-testing-mstest-analyzers.

PR #2226, opened 2024-02-01
Allow IEnumerable of tuple for DynamicData source

### LINKED ISSUE #1335: Allow IEnumerable of Tuples for DynamicData
## Summary

Allow code like this:
```csharp
public (int a, int b int sum)[] TestData => new []
{
    (a: 1, b: 2, sum: 3),
};

[TestMethod]
[DynamicData(nameof(TestData))]
public void SumIsCalculatedCorrectly(int a, int b, int sum)
{
   ...
}
```

## Background and Motivation

Commonly when using `DynamicDataAttribute` the test data is relatively complex.
The input for it needs to be passed as an `IEnumerable<object[]>`.
It's easy to mix up some elements, so they don't match the signature of the test function. 
With big data rows it can also sometimes be difficult to read which element matches which parameter of the test function.

## Proposed Feature

Allow `IEnumerable<ITuple>` for the data of `DynamicDataAttribute`. This could be solved very easily in `DynamicDataAttribute.GetData()`: If `obj is IEnumerable<ITuple>` then convert it to an `IEnumerable<ITuple>`.
For each `ITuple` create a new array and copy the elements of the tuple over.

A further step could be an analyzer that warns if types or names don't match up.

## Alternative Designs

Offer an API to run tests with given parameters in a type-safe way.

### LINKED ISSUE #1624: Allow not only object[] but also Tuple and ValueTuple for DynamicData
## Summary

The existing `Microsoft.VisualStudio.TestTools.UnitTesting.DynamicDataAttribute` allows us to supply the data to a test method. Currently it only understands the data sources that return `IEnumerable<object[]>`. 

## Background and Motivation

The existing loosely-typed approach based on `object[]` works, but sometimes a strongly-typed approach is more desirable.

The problem with the loosely-typed `object[]` approach is that a compiler does not validate the types returned from a particular data source. This may lead to surprise errors in run time.

## Proposed Feature

I propose to add the support for more strongly-typed approaches in addition to the existing `object[]`.

Before we start, let's take a look at the existing loosely-typed code snippet that unfortunately will blow up in run time:

```c#
[TestClass]
public class FooSemanticsTests
{
    static IEnumerable<object[]> TestData
    {
        get
        {
            yield return new object[] { new FooName("123"), true };
            yield return new object[] { new FooName("10"), 10 };
        }
    }

    [TestMethod]
    [DynamicData(nameof(TestData))]
    public void Test(FooName fooName, bool isSimpleName)
    {
    }
}
```

The snipped above uses existing loosely-typed `object[]` approach which does not provide the benefits of compile-time type validations. The lack of compile-time validation leads to wasted time and requires additional development resources.

However, `DynamicData` attribute could, in theory, allow the consumption of strongly-typed tuples.

Let's take a look at the corresponding code snippet. It uses a strongly-typed approach based on `System.ValueType`. This time, the existing error will be caught in compile time:

```c#
[TestClass]
public class FooSemanticsTests
{
    static IEnumerable<(FooName, bool)> TestData
    {
        get
        {
            yield return (new FooName("123"), true);
            yield return (new FooName("10"), 10);  // <-- C# compiler will generate a type cast error at this line 
        }
    }

    [TestMethod]
    [DynamicData(nameof(TestData))]
    public void Test(FooName fooName, bool isSimpleName)
    {
    }
}
```

Currently, the above code snippet cannot be understood by `MSTest.TestFramework` (v3.0.2) in run time: 
```
Parameter 'Property or method TestData on FooSemanticsTests does not return IEnumerable<object[]>.'
```

The same limitation applies to `System.Tuple`: it is not understood either.

## Alternative Designs

As an alternative, I considered using adapter properties to convert from `IEnumerable<System.ValueTuple>` to `IEnumerable<object[]>`. However, that approach is hardly practical due to boilerplate distractions nobody wants to invest in.

[AB#1950023](https://devdiv.visualstudio.com/0bdbc590-a062-4c3f-b0f6-9383f67865ee/_workitems/edit/1950023)

Evangelink:
To avoid breaking public API, we decided to convert the tuple in `object[]` to preserve it. The current solution relies on `ITuple` which is only available in NET471+. Given we currently target net462 for full framework, that means that the feature will only be available for netcore. I'll weigh the benefit/cost ratio of adding net471 build
 in the package so this feature can be available more widely.

Fixes #1335 and fixes #1624

### DIFF
--- a/src/TestFramework/TestFramework/Attributes/DataSource/DynamicDataAttribute.cs
+++ b/src/TestFramework/TestFramework/Attributes/DataSource/DynamicDataAttribute.cs
@@ -1,6 +1,11 @@
 ﻿// Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
+#if NET471_OR_GREATER || NETCOREAPP
+using System.Collections;
+using System.Runtime.CompilerServices;
+#endif
+using System.Diagnostics.CodeAnalysis;
 using System.Globalization;
 using System.Reflection;
 
@@ -114,7 +119,7 @@ public sealed class DynamicDataAttribute : Attribute, ITestDataSource
                     _dynamicDataDeclaringType.FullName));
         }
 
-        if (obj is not IEnumerable<object[]> enumerable)
+        if (!TryGetData(obj, out IEnumerable<object[]>? data))
         {
             throw new ArgumentNullException(
                 string.Format(
@@ -123,7 +128,8 @@ public sealed class DynamicDataAttribute : Attribute, ITestDataSource
                     _dynamicDataSourceName,
                     _dynamicDataDeclaringType.FullName));
         }
-        else if (!enumerable.Any())
+
+        if (!data.Any())
         {
             throw new ArgumentException(
                 string.Format(
@@ -133,7 +139,7 @@ public sealed class DynamicDataAttribute : Attribute, ITestDataSource
                     _dynamicDataDeclaringType.FullName));
         }
 
-        return enumerable;
+        return data;
     }
 
     /// <inheritdoc />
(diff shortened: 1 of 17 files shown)


Evangelink:
> 1. how do you guard that data that currently contain tuple are not expanded in unwanted way? Maybe there should be a flag that keeps it working as is?

Not entirely sure to get the issue you are describing. Do you have some example?

> 2. Do you think it would make sense to add a similar "container" that is generic as the Typed Theory Data. Because ensuring that the output of the data source fits into the parameters of the test method needs to be done via an analyzer. This allows users to define their own types that have simpler constructors, but also might trap us into having to analyze all output types with inheritance. So probably no, and we just stick with tuples?

Yes, I think it makes sense as alternative, see #2227. We did something similar in TATF, it's also making it easier to provide custom name for each entry if we want.

> 3. also please plan for analyzer that connects those pieces together. Because that is imho half of the value. I know the linked issue says that with tuples the error will be caught on compile time, but I don't see how that would be possible without an analyzer. I don't think there is any way to directly link attribute to be "spread" over method signature.  But I might be wrong :))

#2237

nohwnd:
1. And existing method that is emulating what we do and outputs single item array with a tuple in it, and then deconstructs that in the test method. This change will break my code because my tuple will try to bind to 2 parameters instead of 1. Or no?

```cs
namespace TestProject74
{
    [TestClass]
    public class UnitTest1
    {
        [DynamicData(nameof(GetData))]
        [TestMethod]
        public void TestMethod1((string file, string mode) t)
        {
            Console.WriteLine(t.file);
        }

        public IEnumerable<object[]> GetData()
        {
            return new[]
            {
                new object[] { (file: "aaa.txt", mode: "write") }
            };
        }
    }
}
```

Evangelink:
I expect that this would still work from what I have seen but I will add a test to ensure that's the case. Thanks!

PR #2265, opened 2024-02-06
Use Task.Run for Assembly/Class init to allow exiting

Evangelink:
If user doesn't cooperatively cancel, we will just exit observing the method execution so we have a more reactive experience.

cc @MarcoRossignoli


src/Adapter/MSTest.TestAdapter/Execution/TestClassInfo.cs
```diff
@@ -268,17 +269,23 @@ public void RunClassInitialize(TestContext testContext)
...
                             if (baseInitCleanupMethods.Item2 != null)
                             {
                                 BaseClassCleanupMethodsStack.Push(baseInitCleanupMethods.Item2);
                             }
                         }
 
-                        initializeMethod = null;
-
-                        InvokeInitializeMethod(ClassInitializeMethod, testContext);
+                        if (ClassInitializationException is not null)
```
MarcoRossignoli: should it be is null?

src/Adapter/MSTest.TestAdapter/Helpers/MethodRunner.cs
```diff
@@ -0,0 +1,54 @@
...
+
+internal static class MethodRunner
+{
+    internal static TestFailedException? RunWithTimeoutAndCancellation(
+        Action action, CancellationTokenSource cancellationTokenSource, int? timeout, MethodInfo methodInfo,
+        string methodCancelledMessageFormat, string methodTimedOutMessageFormat)
+    {
+        try
+        {
+            var executionTask = Task.Run(action, cancellationTokenSource.Token);
```
MarcoRossignoli: this can throw `TaskCanceledException` that's `OperationCanceledException` but not a timeout or do we fold every `OperationCanceledException` of this method in a "timeout"?

src/Adapter/MSTest.TestAdapter/Helpers/MethodRunner.cs
```diff
@@ -0,0 +1,54 @@
...
+
+            if (!timeout.HasValue)
+            {
+                executionTask.Wait(cancellationTokenSource.Token);
```
MarcoRossignoli, line 35: also this one can throw OCE but it's not a real timeout

src/Adapter/MSTest.TestAdapter/Helpers/MethodRunner.cs
```diff
@@ -0,0 +1,54 @@
...
+                return cancellationTokenSource.IsCancellationRequested
```
MarcoRossignoli: Wonder if this make sense...if we reach this point means that anyway the test ran...and we have a race with a cancellation but "we have the result" actually, so we could return null, something else in-progress eventually fire some OCE

src/Adapter/MSTest.TestAdapter/Helpers/MethodRunner.cs
```diff
@@ -0,0 +1,54 @@
...
+                        string.Format(CultureInfo.InvariantCulture, methodCancelledMessageFormat, methodInfo.DeclaringType!.FullName, methodInfo.Name))
+                    : null;
+            }
+
+            if (executionTask.Wait(timeout.Value, cancellationTokenSource.Token))
+            {
+                return null;
+            }
+
+            // Timed out or cancelled.
```
MarcoRossignoli: timed out only I think..cancelled by api will throw and finish inside the catch

Evangelink: Correct.

src/Adapter/MSTest.TestAdapter/Helpers/MethodRunner.cs
Evangelink: OperationCanceledException
The cancellationToken was canceled.

src/Adapter/MSTest.TestAdapter/Helpers/MethodRunner.cs
Evangelink: Actually this one doesn't throw OCE according to doc

src/Adapter/MSTest.TestAdapter/Helpers/MethodRunner.cs
Evangelink: Yep looks like it always throws so I should return only `null`.

src/Adapter/MSTest.TestAdapter/Helpers/MethodRunner.cs
Evangelink: Now it's better.

PR #2275, opened 2024-02-09
Fix script after directory.build.props changing bin path

nohwnd:


Evangelink (approved):
I confirm this is now working properly locally.

PR #2292, opened 2024-02-09
TestInitialize should be valid

engyebrahim:
fix: https://github.com/microsoft/testfx/issues/2233


src/Analyzers/MSTest.Analyzers/TestInitializeShouldBeValidAnalyzer.cs
```diff
@@ -0,0 +1,93 @@
...
+        {
+            context.ReportDiagnostic(methodSymbol.CreateDiagnostic(NotStaticRule, methodSymbol.Name));
+        }
+
+        if (methodSymbol.ReturnsVoid && methodSymbol.IsAsync)
+        {
+            context.ReportDiagnostic(methodSymbol.CreateDiagnostic(NotAsyncVoidRule, methodSymbol.Name));
+        }
+
+        if (methodSymbol.GetResultantVisibility() != SymbolVisibility.Public || methodSymbol.DeclaredAccessibility != Accessibility.Public)
```
Evangelink: `GetResultantVisibility` already checks the right side of your or condition so it can be dropped.

src/Analyzers/MSTest.Analyzers/AnalyzerReleases.Unshipped.md
```diff
@@ -4,4 +4,5 @@
 
 Rule ID | Category | Severity | Notes
 --------|----------|----------|-------
-MSTEST0007 | Usage | Info | UseAttributeOnTestMethodAnalyzer, [Documentation](https://learn.microsoft.com/dotnet/core/testing/unit-testing-mstest-analyzers-MSTEST0007)
\ No newline at end of file
+MSTEST0007 | Usage | Info | UseAttributeOnTestMethodAnalyzer, [Documentation](https://learn.microsoft.com/dotnet/core/testing/unit-testing-mstest-analyzers-MSTEST0007)
+MSTEST0003 | Usage | Warning | TestInitializeShouldBeValidAnalyzer, [Documentation](https://learn.microsoft.com/dotnet/core/testing/unit-testing-mstest-analyzers-MSTEST0003)
```
Evangelink: The ID seems wrong, I would expect it to be `MSTEST0008`.

src/Analyzers/MSTest.Analyzers/TestInitializeShouldBeValidAnalyzer.cs
```diff
@@ -0,0 +1,93 @@
...
+
+[DiagnosticAnalyzer(LanguageNames.CSharp, LanguageNames.VisualBasic)]
+public sealed class TestInitializeShouldBeValidAnalyzer : DiagnosticAnalyzer
+{
+    private static readonly LocalizableResourceString Title = new(nameof(Resources.TestInitializeShouldBeValidTitle), Resources.ResourceManager, typeof(Resources));
+    private static readonly LocalizableResourceString Description = new(nameof(Resources.TestInitializeShouldBeValidDescription), Resources.ResourceManager, typeof(Resources));
+    private static readonly LocalizableResourceString MessageFormat = new(nameof(Resources.TestInitializeShouldBeValidMessageFormat_Public), Resources.ResourceManager, typeof(Resources));
+
+    internal static readonly DiagnosticDescriptor PublicRule = DiagnosticDescriptorHelper.Create(
+        DiagnosticIds.TestClassShouldBeValidRuleId,
```
Evangelink: This is reusing an existing diagnostic ID. This should be a new entry.

src/Analyzers/MSTest.Analyzers/TestInitializeShouldBeValidAnalyzer.cs
```diff
@@ -0,0 +1,93 @@
...
+namespace MSTest.Analyzers;
+
+[DiagnosticAnalyzer(LanguageNames.CSharp, LanguageNames.VisualBasic)]
+public sealed class TestInitializeShouldBeValidAnalyzer : DiagnosticAnalyzer
+{
+    private static readonly LocalizableResourceString Title = new(nameof(Resources.TestInitializeShouldBeValidTitle), Resources.ResourceManager, typeof(Resources));
+    private static readonly LocalizableResourceString Description = new(nameof(Resources.TestInitializeShouldBeValidDescription), Resources.ResourceManager, typeof(Resources));
+    private static readonly LocalizableResourceString MessageFormat = new(nameof(Resources.TestInitializeShouldBeValidMessageFormat_Public), Resources.ResourceManager, typeof(Resources));
+
+    internal static readonly DiagnosticDescriptor PublicRule = DiagnosticDescriptorHelper.Create(
```
Evangelink, line 22: Could you check if when we set `[assembly: DiscoverInternals]` the `TestInitialize` allows internal? I would expect the behavior to be similar to `TestMethod` attribute. If that's the case, you can mimic the logic from the `TestMethodShouldBeValidAnalyzer` class.

src/Analyzers/MSTest.Analyzers/TestInitializeShouldBeValidAnalyzer.cs
```diff
@@ -0,0 +1,93 @@
...
+        Description,
+        Category.Usage,
+        DiagnosticSeverity.Warning,
+        isEnabledByDefault: true);
+
+    internal static readonly DiagnosticDescriptor NotStaticRule = PublicRule.WithMessage(new(nameof(Resources.TestInitializeShouldBeValidMessageFormat_NotStatic), Resources.ResourceManager, typeof(Resources)));
+    internal static readonly DiagnosticDescriptor NoParametersRule = PublicRule.WithMessage(new(nameof(Resources.TestInitializeShouldBeValidMessageFormat_NoParameters), Resources.ResourceManager, typeof(Resources)));
+    internal static readonly DiagnosticDescriptor ReturnTypeRule = PublicRule.WithMessage(new(nameof(Resources.TestInitializeShouldBeValidMessageFormat_ReturnType), Resources.ResourceManager, typeof(Resources)));
+    internal static readonly DiagnosticDescriptor NotAsyncVoidRule = PublicRule.WithMessage(new(nameof(Resources.TestInitializeShouldBeValidMessageFormat_NotAsyncVoid), Resources.ResourceManager, typeof(Resources)));
+
```
Evangelink, line 38: Could you check if the `TestInitialize` is called if the containing type is not marked as `[TestClass]`? If that's not the case, we should probably raise for this case.

src/Analyzers/MSTest.Analyzers/Resources.resx
```diff
@@ -181,6 +181,31 @@
   <data name="TestContextShouldBeValidTitle" xml:space="preserve">
     <value>Test context property should have valid layout</value>
   </data>
+  <data name="TestInitializeShouldBeValidDescription" xml:space="preserve">
+    <value>TestInitialize attribute should follow the following layout to be valid:
+- it should be 'public' 
+- it should not be 'static'
+- it should not take any parameter
+- it should not return a value (if you are using async-await in method then return-type must be Task).</value>
```
Evangelink: We also allow ValueTask.

test/UnitTests/MSTest.Analyzers.UnitTests/TestInitializeShouldBeValidAnalyzerTests.cs
```diff
@@ -0,0 +1,209 @@
...
+            [TestClass]
+            public class MyTestClass
+            {
+                [TestInitialize]
+                public int {|#0:TestInitialize0|}()
+                {
+                    return 0;
+                }
+
+                [TestInitialize]
```
Evangelink, line 183: Is it actually working to have multiple valid methods marked with `[TestInitialize]`?

test/UnitTests/MSTest.Analyzers.UnitTests/TestInitializeShouldBeValidAnalyzerTests.cs
```diff
@@ -0,0 +1,209 @@
...
+    public async Task WhenTestInitializeHasParameters_Diagnostic()
+    {
+        var code = """
+            using Microsoft.VisualStudio.TestTools.UnitTesting;
+
+            [TestClass]
+            public class MyTestClass
+            {
+                [TestInitialize]
+                public void {|#0:TestInitialize|}(int x)
```
Evangelink: This is good! I would add another method with `TestContext testContext` because this is a common pattern for `AssemblyInitialize` and `ClassInitialize` methods so people could get confused and willing to do something similar. I already know this is correctly handled by your code, this is just to explicitely document the case :)

Evangelink (commented):
Awesome first version! I am super happy with the quality of the work on this PR, GG!

src/Analyzers/MSTest.Analyzers/Resources.resx
```diff
@@ -181,6 +181,44 @@
...
     <value>Test context property should have valid layout</value>
   </data>
+  <data name="TestInitializeShouldBeValidDescription" xml:space="preserve">
+    <value>TestInitialize attribute should follow the following layout to be valid:
+- it should be 'public' 
+- it should not be 'static'
+- it should not be generic
+- it should not be 'abstract'
+- it should not take any parameter
+- return type should be 'void' or 'Task'
```
nohwnd:
```suggestion
- return type should be 'void', 'Task' or 'ValueTask'
```

nohwnd: Looks like you have it in the other string, just not here.

test/UnitTests/MSTest.Analyzers.UnitTests/TestInitializeShouldBeValidAnalyzerTests.cs
nohwnd: It does not, it throws here. https://github.com/microsoft/testfx/blob/main/src/Adapter/MSTest.TestAdapter/Execution/TestClassInfo.cs#L189

But it should not matter for this test.

src/Analyzers/MSTest.Analyzers/TestInitializeShouldBeValidAnalyzer.cs
nohwnd: it does not allow internal even with [assembly: DiscoverInternals]

test/UnitTests/MSTest.Analyzers.UnitTests/TestInitializeShouldBeValidAnalyzerTests.cs
Evangelink: Not for this test but we may want to either tweak the analyzer or add a new analyzer to ensure people don't define multiple time this analyzer.

src/Analyzers/MSTest.Analyzers/TestInitializeShouldBeValidAnalyzer.cs
Evangelink: I really love this discover internals that only work halfway and doesn't allow you to have it all internal xD

src/Analyzers/MSTest.Analyzers/AnalyzerReleases.Unshipped.md
```diff
@@ -4,4 +4,5 @@
 
 Rule ID | Category | Severity | Notes
 --------|----------|----------|-------
-MSTEST0007 | Usage | Info | UseAttributeOnTestMethodAnalyzer, [Documentation](https://learn.microsoft.com/dotnet/core/testing/unit-testing-mstest-analyzers-MSTEST0007)
\ No newline at end of file
+MSTEST0007 | Usage | Warning | TestInitializeShouldBeValidAnalyzer, [Documentation](https://learn.microsoft.com/dotnet/core/testing/unit-testing-mstest-analyzers-MSTEST0008)
```
Evangelink: MSTEST0007 should not have changed, only MSTEST0008

src/Analyzers/MSTest.Analyzers/Helpers/DiagnosticIds.cs
```diff
@@ -12,4 +12,5 @@ internal static class DiagnosticIds
     public const string TestContextShouldBeValidRuleId = "MSTEST0005";
     public const string AvoidExpectedExceptionAttributeRuleId = "MSTEST0006";
     public const string UseAttributeOnTestMethodRuleId = "MSTEST0007";
+    public const string TestInitializeShouldBeValidRuleId = "MSTEST0007";
```
Evangelink: Look at the line above, it's already defining the same ID :)

src/Analyzers/MSTest.Analyzers/Resources.resx
```diff
@@ -181,15 +181,53 @@
...
     <value>Test methods, methods marked with the '[TestMethod]' attribute, should respect the following layout to be considered valid by MSTest:
 - it should be 'public' (or 'internal' if '[assembly: DiscoverInternals]' attribute is set)
 - it should not be 'static'
 - it should not be generic
 - it should not be 'abstract'
-- return type should be 'void' or 'Task'
+- return type should be 'void', 'Task' or 'ValueTask'
 - it should not be 'async void'
-- it should be a special method (finalizer, operator...).</value>
+- it should not be a special method (finalizer, operator...).</value>
```
Evangelink: Good catch but I would prefer to have these changes in a different PR please.

engyebrahim: here we go: https://github.com/microsoft/testfx/pull/2295

test/UnitTests/MSTest.Analyzers.UnitTests/TestInitializeShouldBeValidAnalyzerTests.cs
engyebrahim: I think we can add another analyzer checking duplication for all attributes which can't exist more than one time.
@nohwnd @Evangelink  what do you think?

src/Analyzers/MSTest.Analyzers/TestInitializeShouldBeValidAnalyzer.cs
engyebrahim: I tried it does not get called

engyebrahim: @Evangelink should I add something in the analyzer for that?

src/Analyzers/MSTest.Analyzers/TestInitializeShouldBeValidAnalyzer.cs
engyebrahim: yeah i tested that and it's not working with internal

src/Analyzers/MSTest.Analyzers/AnalyzerReleases.Unshipped.md
```diff
@@ -4,4 +4,5 @@
 
 Rule ID | Category | Severity | Notes
 --------|----------|----------|-------
-MSTEST0007 | Usage | Info | UseAttributeOnTestMethodAnalyzer, [Documentation](https://learn.microsoft.com/dotnet/core/testing/unit-testing-mstest-analyzers-MSTEST0007)
\ No newline at end of file
+MSTEST0007 | Usage | Warning | TestInitializeShouldBeValidAnalyzer, [Documentation](https://learn.microsoft.com/dotnet/core/testing/unit-testing-mstest-analyzers-MSTEST0007)
```
Evangelink: Still not good :) Let me help you:

```suggestion
MSTEST0007 | Usage | Info | UseAttributeOnTestMethodAnalyzer, [Documentation](https://learn.microsoft.com/dotnet/core/testing/unit-testing-mstest-analyzers-MSTEST0007)
```

src/Analyzers/MSTest.Analyzers/TestInitializeShouldBeValidAnalyzer.cs
Evangelink: I am not sure what's best. I am wondering if it's not better to simply add one more condition as part of this rule as it matches the setup of having a valid layout for `TestInitialize`.

If I recall correctly, `TestClass` can be inherited so you would need to check that the containing type inherits from `TestClass` and not simply is equal to it.

src/Analyzers/MSTest.Analyzers/TestInitializeShouldBeValidAnalyzer.cs
```diff
@@ -0,0 +1,114 @@
...
+        {
+            context.ReportDiagnostic(methodSymbol.CreateDiagnostic(NotStaticRule, methodSymbol.Name));
+        }
+
+        if (methodSymbol.ReturnsVoid && methodSymbol.IsAsync)
+        {
+            context.ReportDiagnostic(methodSymbol.CreateDiagnostic(NotAsyncVoidRule, methodSymbol.Name));
+        }
+
+        if (methodSymbol.DeclaredAccessibility != Accessibility.Public)
```
Evangelink: It's better to use `GetResultantVisbility` because it's taking into account class (and outer class) visbility.

For example, with what you did

```cs
internal class C
{
  public void M() {} // Declared is public but resultant is internal because class is internal
}

File: test/UnitTests/MSTest.Analyzers.UnitTests/TestInitializeShouldBeValidAnalyzerTests.cs (hunk 7 above)
Comment (Evangelink) [reply test/UnitTests/MSTest.Analyzers.UnitTests/TestInitializeShouldBeValidAnalyzerTests.cs:183]: Let's leave that case out for now, I am not sure.

**Review by Evangelink — COMMENTED** [review]
Nearly ready to be merged.

File: src/Analyzers/MSTest.Analyzers/TestInitializeShouldBeValidAnalyzer.cs (hunk 14 above)
Comment (engyebrahim) [reply src/Analyzers/MSTest.Analyzers/TestInitializeShouldBeValidAnalyzer.cs]: I tried using it but the test fails, I think it's not covering all needed checks

Comment (engyebrahim) [reply src/Analyzers/MSTest.Analyzers/TestInitializeShouldBeValidAnalyzer.cs]: should i go back to checking both conditions?

Comment (Evangelink) [reply src/Analyzers/MSTest.Analyzers/TestInitializeShouldBeValidAnalyzer.cs]: Right, I totally forgot that `protected` was considered as public resultant visiblity by the tool. So yes please, check for declared visiblity = public first and then add the check to get resultant visibility.

File: src/Analyzers/MSTest.Analyzers/TestInitializeShouldBeValidAnalyzer.cs (hunk 5 above)
Comment (Evangelink) [reply src/Analyzers/MSTest.Analyzers/TestInitializeShouldBeValidAnalyzer.cs:38]: Let's ignore this case for now, I have added a comment here https://github.com/microsoft/testfx/issues/2051#issuecomment-1938645968

File: src/Analyzers/MSTest.Analyzers/TestInitializeShouldBeValidAnalyzer.cs [hunk 15]
```diff
@@ -0,0 +1,114 @@
...
+        {
+            context.ReportDiagnostic(methodSymbol.CreateDiagnostic(NotStaticRule, methodSymbol.Name));
+        }
+
+        if (methodSymbol.ReturnsVoid && methodSymbol.IsAsync)
+        {
+            context.ReportDiagnostic(methodSymbol.CreateDiagnostic(NotAsyncVoidRule, methodSymbol.Name));
+        }
+
+        if (methodSymbol.GetResultantVisibility() != SymbolVisibility.Public || methodSymbol.DeclaredAccessibility != Accessibility.Public)
```
Evangelink: Nit for perf

```suggestion
        if (methodSymbol.DeclaredAccessibility != Accessibility.Public || methodSymbol.GetResultantVisibility() != SymbolVisibility.Public)
```

Evangelink (approved):

Issue #2219, opened 2024-01-31
An exception occurred while invoking executor executor://mstestadapter/v2. System.IO.DirectoryNotFoundException


cbersch:
## Describe the bug
Our Azure DevOps pipeline, which runs tests from 60 assemblies in parallel via `dotnet test`, randomly fails with the exception
```
An exception occurred while invoking executor executor://mstestadapter/v2. System.IO.DirectoryNotFoundException: Could not find a part of the path 'F:\agent\_work\1\s\TestResults\Deploy_user 20240131T155218[1]\In\BLDSRV01'.
```
The .NET8 projects reference packages
- `MSTest.TestAdapter` 3.1.1
- `MSTest.TestFramework` 3.1.1
- `Microsoft.NET.Test.Sdk` 17.8.0

## Steps To Reproduce
Not really, but the Task runs the following `dotnet test` command:
```
dotnet test ./MySolution.sln --no-build --no-restore --configuration=Release --logger=trx --results-directory=TestResults --settings=.runsettings --collect="Code Coverage;Format=cobertura" --diag=TestResults\dotnet-test.diag
```
## Expected behavior
The tests are executed and `dotnet test` exits with code 0

## Actual behavior
The tests are not executed and `dotnet test` exits with code 1

## Additional context
The relevant part from the `diag` file
```
TpTrace Information: 0 : 23220, 4, 2024/01/31, 15:52:18.408, 2190101548319, testhost.dll, MSTest - MSTestDiscoverer: Found 8 tests from source F:\agent\_work\1\s\_Build\Release\Assemblies\ipipe.DataMapping.Test\ipipe.DataMapping.Test.dll
TpTrace Verbose: 0 : 23220, 9, 2024/01/31, 15:52:18.512, 2190102589919, testhost.dll, TcpClientExtensions.MessageLoopAsync: Polling on remoteEndPoint: [::ffff:127.0.0.1]:55794 localEndPoint: [::ffff:127.0.0.1]:55853
TpTrace Error: 0 : 23220, 4, 2024/01/31, 15:52:19.234, 2190109808184, testhost.dll, BaseRunTests.RunTestInternalWithExecutors: An exception occurred while invoking executor executor://mstestadapter/v2. System.IO.DirectoryNotFoundException: Could not find a part of the path 'F:\agent\_work\1\s\TestResults\Deploy_user 20240131T155218[1]\In\BLDSRV01'.
[...]

PR #2294, opened 2024-02-09
Add retry mechanism for creating directory

Evangelink:
Relates to #2219

### DIFF
--- a/src/Adapter/MSTestAdapter.PlatformServices/Utilities/FileUtility.cs
+++ b/src/Adapter/MSTestAdapter.PlatformServices/Utilities/FileUtility.cs
@@ -24,10 +24,33 @@ internal class FileUtility
     {
         DebugEx.Assert(!StringEx.IsNullOrEmpty(directory), "directory");
 
-        if (!Directory.Exists(directory))
+        if (Directory.Exists(directory))
         {
-            Directory.CreateDirectory(directory);    // Creates subdir chain if necessary.
+            return;
         }
+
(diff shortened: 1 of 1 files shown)


src/Adapter/MSTestAdapter.PlatformServices/Utilities/FileUtility.cs
```diff
@@ -24,10 +24,33 @@ public virtual void CreateDirectoryIfNotExists(string directory)
...
+            catch (DirectoryNotFoundException)
+            {
+                attempt++;
+                if (attempt == 5)
+                {
+                    throw;
+                }
+                else
+                {
+                    Thread.Sleep(attempt * 200);
```
nohwnd, line 49: This sleep seems very long, IMHO you need maybe Thread.Sleep(1) to get the smallest amount of sleep, or use none.

src/Adapter/MSTestAdapter.PlatformServices/Utilities/FileUtility.cs
```diff
@@ -24,10 +24,33 @@ public virtual void CreateDirectoryIfNotExists(string directory)
...
+            return;
         }
+
+        int attempt = 0;
+        do
+        {
+            try
+            {
+                // Creates subdir chain if necessary.
+                Directory.CreateDirectory(directory);
```
nohwnd, line 38: Add `break` here to not loop forever?

PR #2328, opened 2024-02-13
AssemblyInitialize should be valid

engyebrahim:
fix: https://github.com/microsoft/testfx/issues/2229


src/Analyzers/MSTest.Analyzers/AssemblyInitializeShouldBeValidAnalyzer.cs
```diff
@@ -0,0 +1,110 @@
...
+        context.ConfigureGeneratedCodeAnalysis(GeneratedCodeAnalysisFlags.None);
+        context.EnableConcurrentExecution();
+
+        context.RegisterCompilationStartAction(context =>
+        {
+            if (context.Compilation.TryGetOrCreateTypeByMetadataName(WellKnownTypeNames.MicrosoftVisualStudioTestToolsUnitTestingAssemblyInitializeAttribute, out var assemblyInitializeAttributeSymbol))
+            {
+                var taskSymbol = context.Compilation.GetOrCreateTypeByMetadataName(WellKnownTypeNames.SystemThreadingTasksTask);
+                var valueTaskSymbol = context.Compilation.GetOrCreateTypeByMetadataName(WellKnownTypeNames.SystemThreadingTasksValueTask);
+                var testContextSymbol = context.Compilation.GetOrCreateTypeByMetadataName(WellKnownTypeNames.MicrosoftVisualStudioTestToolsUnitTestingTestContext);
```
Evangelink: I would move this one to the if on line 48 because if we cannot resolve this type, there is no way for us to validate correctly the method layout so there's no need to register. But what you did makes sense :)

src/Analyzers/MSTest.Analyzers/AssemblyInitializeShouldBeValidAnalyzer.cs
```diff
@@ -0,0 +1,110 @@
...
+        {
+            context.ReportDiagnostic(methodSymbol.CreateDiagnostic(StaticRule, methodSymbol.Name));
+        }
+
+        if (methodSymbol.ReturnsVoid && methodSymbol.IsAsync)
+        {
+            context.ReportDiagnostic(methodSymbol.CreateDiagnostic(NotAsyncVoidRule, methodSymbol.Name));
+        }
+
+        if (methodSymbol.DeclaredAccessibility != Accessibility.Public || methodSymbol.GetResultantVisibility() != SymbolVisibility.Public)
```
Evangelink: Have you checked if when setting `DiscoverInternals` we could have the assembly init internal? I know you checked for `TestInitialize` but I am wondering if this attribute behaves differently.

src/Analyzers/MSTest.Analyzers/Helpers/WellKnownTypeNames.cs
```diff
@@ -21,6 +21,7 @@ internal static class WellKnownTypeNames
     public const string MicrosoftVisualStudioTestToolsUnitTestingTestCleanupAttribute = "Microsoft.VisualStudio.TestTools.UnitTesting.TestCleanupAttribute";
     public const string MicrosoftVisualStudioTestToolsUnitTestingTestContext = "Microsoft.VisualStudio.TestTools.UnitTesting.TestContext";
     public const string MicrosoftVisualStudioTestToolsUnitTestingTestInitializeAttribute = "Microsoft.VisualStudio.TestTools.UnitTesting.TestInitializeAttribute";
+    public const string MicrosoftVisualStudioTestToolsUnitTestingAssemblyInitializeAttribute = "Microsoft.VisualStudio.TestTools.UnitTesting.AssemblyInitializeAttribute";
```
Evangelink: Can you please keep this file sorted by name?

src/Analyzers/MSTest.Analyzers/Resources.resx
```diff
@@ -117,6 +117,40 @@
...
+    <value>Assembly Initialize '{0}' should return 'void', 'Task' or 'ValueTask'</value>
+  </data>
+  <data name="AssemblyInitializeShouldBeValidMessageFormat_SingleContextParameter" xml:space="preserve">
+    <value>Assembly Initialize '{0}' should take a single parameter of type TestContext</value>
+  </data>
+  <data name="AssemblyInitializeShouldBeValidMessageFormat_Static" xml:space="preserve">
+    <value>Assembly Initialize '{0}' should be 'static'</value>
+  </data>
+  <data name="AssemblyInitializeShouldBeValidTitle" xml:space="preserve">
+    <value>Assembly Initializes method should have valid layout</value>
```
Evangelink:
```suggestion
    <value>Assembly initialize methods should have valid layout</value>
```

src/Analyzers/MSTest.Analyzers/Resources.resx
```diff
@@ -117,6 +117,40 @@
...
+- it should be 'public' 
+- it should be 'static'
+- it should not be generic
+- it should take one parameter of type TestContext
+- return type should be 'void', 'Task' or 'ValueTask'
+- it should not be 'async void'
+- it should not be a special method (finalizer, operator...).</value>
+  </data>
+  <data name="AssemblyInitializeShouldBeValidMessageFormat_NotAsyncVoid" xml:space="preserve">
+    <value>Assembly Initialize '{0}' should return 'void', 'Task' or 'ValueTask'</value>
```
Evangelink:
```suggestion
    <value>Assembly initialize method '{0}' should return 'void', 'Task' or 'ValueTask'</value>
```

src/Analyzers/MSTest.Analyzers/Resources.resx
```diff
@@ -117,6 +117,40 @@
...
+  </data>
+  <data name="AssemblyInitializeShouldBeValidMessageFormat_NotGeneric" xml:space="preserve">
+    <value>Assembly Initialize '{0}' should not be generic</value>
```
Evangelink:
```suggestion
    <value>Assembly initialize method '{0}' should not be generic</value>
```

src/Analyzers/MSTest.Analyzers/Resources.resx
```diff
@@ -117,6 +117,40 @@
...
+  </data>
+  <data name="AssemblyInitializeShouldBeValidMessageFormat_Ordinary" xml:space="preserve">
+    <value>Assembly Initialize '{0}' should be an 'ordinary' method</value>
```
Evangelink:
```suggestion
    <value>Assembly initialize method '{0}' should be an 'ordinary' method</value>
```

src/Analyzers/MSTest.Analyzers/Resources.resx
```diff
@@ -117,6 +117,40 @@
...
+  </data>
+  <data name="AssemblyInitializeShouldBeValidMessageFormat_Public" xml:space="preserve">
+    <value>Assembly Initialize '{0}' should be 'public'</value>
```
Evangelink:
```suggestion
    <value>Assembly initialize method '{0}' should be 'public'</value>
```

src/Analyzers/MSTest.Analyzers/Resources.resx
```diff
@@ -117,6 +117,40 @@
...
+  </data>
+  <data name="AssemblyInitializeShouldBeValidMessageFormat_ReturnType" xml:space="preserve">
+    <value>Assembly Initialize '{0}' should return 'void', 'Task' or 'ValueTask'</value>
```
Evangelink:
```suggestion
    <value>Assembly initialize method '{0}' should return 'void', 'Task' or 'ValueTask'</value>
```

src/Analyzers/MSTest.Analyzers/Resources.resx
```diff
@@ -117,6 +117,40 @@
...
+  </data>
+  <data name="AssemblyInitializeShouldBeValidMessageFormat_SingleContextParameter" xml:space="preserve">
+    <value>Assembly Initialize '{0}' should take a single parameter of type TestContext</value>
```
Evangelink:
```suggestion
    <value>Assembly initialize method '{0}' should take a single parameter of type 'TestContext'</value>
```

src/Analyzers/MSTest.Analyzers/Resources.resx
```diff
@@ -117,6 +117,40 @@
...
+  </data>
+  <data name="AssemblyInitializeShouldBeValidMessageFormat_Static" xml:space="preserve">
+    <value>Assembly Initialize '{0}' should be 'static'</value>
```
Evangelink:
```suggestion
    <value>Assembly initialize method '{0}' should be 'static'</value>
```

src/Analyzers/MSTest.Analyzers/Resources.resx
```diff
@@ -255,7 +289,7 @@
     <value>Test Initialize '{0}' should return 'void', 'Task' or 'ValueTask'</value>
   </data>
   <data name="TestInitializeShouldBeValidTitle" xml:space="preserve">
-    <value>Test methods should have valid layout</value>
+    <value>Test method method should have valid layout</value>
```
Evangelink: Please revert this change (and related change in all the impacted xlf files).

src/Analyzers/MSTest.Analyzers/Resources.resx
```diff
@@ -117,6 +117,40 @@
   <resheader name="writer">
     <value>System.Resources.ResXResourceWriter, System.Windows.Forms, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089</value>
   </resheader>
+  <data name="AssemblyInitializeShouldBeValidDescription" xml:space="preserve">
+    <value>Assembly Initialize attribute should follow the following layout to be valid:
```
Evangelink:
```suggestion
    <value>Assembly Initialize methods should follow the following layout to be valid:
```

Evangelink: or maybe even better as 

```suggestion
    <value>Methods marked with [AssemblyInitialize] should follow the following layout to be valid:
```

Evangelink: Oooh and I just noticed but `should follow the following` doesn't read so well. It's possible I didn't notice on the previous PRs... sorry.

What about `should adhere to the following layout`?

Evangelink: @nohwnd I need some English writting check, my brain is confused.

test/UnitTests/MSTest.Analyzers.UnitTests/AssemblyInitializeShouldBeValidAnalyzerTests.cs
```diff
@@ -0,0 +1,253 @@
...
+            """;
+
+        await VerifyCS.VerifyAnalyzerAsync(
+            code,
+            VerifyCS.Diagnostic(AssemblyInitializeShouldBeValidAnalyzer.StaticRule)
+                .WithLocation(0)
+                .WithArguments("AssemblyInitialize"));
+    }
+
+    public async Task WhenAssemblyInitializeHasParameters_Diagnostic()
```
Evangelink: Test talks about parameters but the code is parameterless. We should have both cases.

src/Analyzers/MSTest.Analyzers/Resources.resx
nohwnd: I don't think we used '' around ordinary in other rules.

src/Analyzers/MSTest.Analyzers/Resources.resx
nohwnd: LGTM

src/Analyzers/MSTest.Analyzers/AssemblyInitializeShouldBeValidAnalyzer.cs
engyebrahim: yup I checked it, It doesn't work with `DiscoverInternals `

src/Analyzers/MSTest.Analyzers/Resources.resx
engyebrahim: @nohwnd we used '' it the other rules, is it better to remove it from them all?
I will resolve conversation here and if we agreed to remove the '' I'll remove it in another PR for all

src/Analyzers/MSTest.Analyzers/AssemblyInitializeShouldBeValidAnalyzer.cs
engyebrahim: yup make scene, I've thought about it :)

src/Analyzers/MSTest.Analyzers/Resources.resx
```diff
@@ -117,6 +117,40 @@
...
+- it should be 'public' 
+- it should be 'static'
+- it should not be generic
+- it should take one parameter of type 'TestContext'
+- return type should be 'void', 'Task' or 'ValueTask'
+- it should not be 'async void'
+- it should not be a special method (finalizer, operator...).</value>
+  </data>
+  <data name="AssemblyInitializeShouldBeValidMessageFormat_NotAsyncVoid" xml:space="preserve">
+    <value>AssemblyInitialize Method '{0}' should return 'void', 'Task' or 'ValueTask'</value>
```
Evangelink: You've been using lower cases "method" in the TestInit/Cleanup messages, let's do something similar here:

```suggestion
    <value>AssemblyInitialize method '{0}' should return 'void', 'Task' or 'ValueTask'</value>
```

Please update the other entries.

src/Package/MSTest.Sdk/Sdk/Runner/Runner.targets
```diff
@@ -0,0 +1,72 @@
+<?xml version="1.0" encoding="utf-8" ?>
```
Evangelink: This should not be part of this PR.

test/UnitTests/MSTest.Analyzers.UnitTests/AssemblyInitializeShouldBeValidAnalyzerTests.cs
```diff
@@ -0,0 +1,253 @@
...
+            VerifyCS.Diagnostic(AssemblyInitializeShouldBeValidAnalyzer.NotGenericRule)
+                .WithLocation(0)
+                .WithArguments("AssemblyInitialize"));
+    }
+
+    [Arguments("protected")]
+    [Arguments("internal")]
+    [Arguments("internal protected")]
+    [Arguments("private")]
+    public async Task WhenAssemblyInitializeIsNotPublic_Diagnostic(string accessibility)
```
Evangelink, line 125: Please add either another class in this test or a different test that checks the resultant visiblity. Something like `internal class` and `public static void` method.

This way if we ever do a change and drop resultant visibility we would have a test failing.

In case I missed it in `TestInit` and `TestCleanup` PRs, would it be ok for you to do another PR also adding the check?!

Evangelink (commented):
Few more nits and all good to be merged.

test/UnitTests/MSTest.Analyzers.UnitTests/AssemblyInitializeShouldBeValidAnalyzerTests.cs
engyebrahim: that one helps to fix an issue! thanks! 
I'll update it in both TestInit and TestCleanup in another PR

engyebrahim: here we go: https://github.com/microsoft/testfx/pull/2346

src/Analyzers/MSTest.Analyzers/AssemblyInitializeShouldBeValidAnalyzer.cs
```diff
@@ -0,0 +1,113 @@
...
+        }
+
+        if (methodSymbol.ReturnsVoid && methodSymbol.IsAsync)
+        {
+            context.ReportDiagnostic(methodSymbol.CreateDiagnostic(NotAsyncVoidRule, methodSymbol.Name));
+        }
+
+        if (methodSymbol.DeclaredAccessibility != Accessibility.Public
+            || (!canDiscoverInternals && methodSymbol.GetResultantVisibility() != SymbolVisibility.Public)
+                 || (canDiscoverInternals && methodSymbol.GetResultantVisibility() == SymbolVisibility.Private))
```
Evangelink: Indent nit

```suggestion
            || (!canDiscoverInternals && methodSymbol.GetResultantVisibility() != SymbolVisibility.Public)
            || (canDiscoverInternals && methodSymbol.GetResultantVisibility() == SymbolVisibility.Private))
```

test/UnitTests/MSTest.Analyzers.UnitTests/AssemblyInitializeShouldBeValidAnalyzerTests.cs
```diff
@@ -0,0 +1,273 @@
...
+                [AssemblyInitialize]
+                public static void AssemblyInitialize(TestContext context)
+                {
+                }
+            }
+            """;
+
+        await VerifyCS.VerifyAnalyzerAsync(code);
+    }
+
```
Evangelink, line 53: Adding another test with public class, internal method and can discover so we don't forget about this weird case.

```suggestion
    }

    public async Task WhenAssemblyInitializeIsInternal_InsidePublicClassWithDiscoverInternals_Diagnostic()
    {
        var code = """
            using Microsoft.VisualStudio.TestTools.UnitTesting;

            [assembly: DiscoverInternals]

            [TestClass]
            public class MyTestClass
            {
                [AssemblyInitialize]
                internal static void {|#0:AssemblyInitialize|}(TestContext context)
                {
                }
            }
            """;

        await VerifyCS.VerifyAnalyzerAsync(
            code,
            VerifyCS.Diagnostic(AssemblyInitializeShouldBeValidAnalyzer.PublicRule)
                .WithLocation(0)
                .WithArguments("AssemblyInitialize"));
    }
```

Evangelink (dismissed):

Evangelink (approved):

PR #2329, opened 2024-02-13
Fix TestInitialize and TestCleanup analyzers resources

engyebrahim:


MarcoRossignoli:
why not TestInitialize?

MarcoRossignoli (commented):
why not TestInitialize?

engyebrahim:
@MarcoRossignoli its's like that in all messages in testInitialize analyzer, should i change it all?

src/Analyzers/MSTest.Analyzers/Resources.resx
```diff
@@ -255,7 +255,7 @@
     <value>Test Initialize '{0}' should return 'void', 'Task' or 'ValueTask'</value>
   </data>
   <data name="TestInitializeShouldBeValidTitle" xml:space="preserve">
-    <value>Test methods should have valid layout</value>
+    <value>Test Initialize should have valid layout</value>
```
Evangelink: Thanks @engyebrahim, I totally missed that during review. I think what reads best is probably `Test initialize methods should have valid layout`. For the messages, I would go with `Test initialize method '....`.

@cvpoienaru I might have done the same mistake for cleanup, could you check please?

engyebrahim: @Evangelink I fixed it for both analyzers

Evangelink (approved):
Awesome, thanks @engyebrahim!

I have updated the PR title to better reflect the new scope of the changes.

PR #2341, opened 2024-02-14
Update SDK after the follow-up

MarcoRossignoli:


src/Package/MSTest.Sdk/Sdk/Sdk.props.template
```diff
@@ -1,43 +1,13 @@
 <?xml version="1.0" encoding="utf-8" ?>
 <Project>
+
   <!-- Implicit top import -->
   <Import Project="Sdk.props" Sdk="Microsoft.NET.Sdk" />
 
-  <!-- Core -->
   <PropertyGroup>
-    <IsPackable>false</IsPackable>
+    <UseVSTest Condition=" '$(UseVSTest)' == '' " >false</UseVSTest>
```
MarcoRossignoli, line 8: @Evangelink are you fine with this one? We can avoid to advertise but could be a useful plan B

Issue #2238, opened 2024-02-01
Analyzer: DataRow should be valid


Evangelink:
## Summary

Title: DataRow should be valid
Category: Usage
Description: <to be defined based on all constraints>
Default severity: Warning

Please refer to `MSTEST0003: Test methods should have valid layout`.

Non exhaustive list of constraints:
- number of parameters should match method parameters count
- parameter type should match method parameter type

[AB#1950784](https://devdiv.visualstudio.com/0bdbc590-a062-4c3f-b0f6-9383f67865ee/_workitems/edit/1950784)

Evangelink:
Fixed by #2352

PR #2352, opened 2024-02-16
MSTEST0014: DataRow should be valid

cvpoienaru:
Fixes issue number: https://github.com/microsoft/testfx/issues/2238


src/Analyzers/MSTest.Analyzers/DataRowShouldBeValidAnalyzer.cs
```diff
@@ -0,0 +1,172 @@
...
+            return;
+        }
+
+        // DataRow constructors have either zero or one argument(s). If we get here, we are
+        // on the one argument case. Check if we match either of the array argument constructors
+        // and expand the array argument if we do.
+        ImmutableArray<TypedConstant> constructorArguments = attribute.ConstructorArguments;
+        if (constructorArguments[0].IsNull)
+        {
+            return;
```
Evangelink: I think that this early return would be causing the following test case to not work

```cs
[DataRow(null)]
public void M()
{
}
```

src/Analyzers/MSTest.Analyzers/DataRowShouldBeValidAnalyzer.cs
```diff
@@ -0,0 +1,172 @@
...
+        if (constructorArguments[0].Kind is TypedConstantKind.Array)
+        {
+            constructorArguments = constructorArguments[0].Values;
+        }
+
+        // 1. Diagnostic on mismatch of constructor argument count and method argument count if method
+        //    doesn't accept params.
+        // 2. Diagnostic on lower constructor argument count than method argument count if method
+        //    accepts params. Discard params argument itself because it can contain 0 params.
+        int lastMethodArgumentIndex = methodSymbol.Parameters.Length - 1;
```
Evangelink: Let's be consistent with wording this is not last argument index but last parameter idnex.

src/Analyzers/MSTest.Analyzers/DataRowShouldBeValidAnalyzer.cs
```diff
@@ -0,0 +1,172 @@
...
+        bool methodHasParams = methodSymbol.Parameters[lastMethodArgumentIndex].Type.Kind == SymbolKind.ArrayType;
```
Evangelink: This doesn't mean that the last parameter is params, only this is an array.

src/Analyzers/MSTest.Analyzers/DataRowShouldBeValidAnalyzer.cs
```diff
@@ -0,0 +1,172 @@
...
+        MessageFormat,
+        Description,
+        Category.Usage,
+        DiagnosticSeverity.Warning,
+        isEnabledByDefault: true);
+
+    internal static readonly DiagnosticDescriptor AtLeastOneArgumentRule = DataRowOnTestMethodRule
+        .WithMessage(new(nameof(Resources.DataRowShouldBeValidMessageFormat_AtLeastOneArgument), Resources.ResourceManager, typeof(Resources)));
+
+    internal static readonly DiagnosticDescriptor ArgumentNumberRule = DataRowOnTestMethodRule
```
Evangelink: Maybe `Count` instead of `Number`? Maybe even better as `ArgumentCountMismatchRule`.

src/Analyzers/MSTest.Analyzers/DataRowShouldBeValidAnalyzer.cs
```diff
@@ -0,0 +1,172 @@
...
+        .WithMessage(new(nameof(Resources.DataRowShouldBeValidMessageFormat_ArgumentNumber), Resources.ResourceManager, typeof(Resources)));
+
+    internal static readonly DiagnosticDescriptor ArgumentTypeRule = DataRowOnTestMethodRule
```
Evangelink: Maybe `ArgumentTypeMismatchRule`.

src/Analyzers/MSTest.Analyzers/DataRowShouldBeValidAnalyzer.cs
```diff
@@ -0,0 +1,172 @@
...
+            return;
+        }
+
+        int i = 0;
+        ITypeSymbol? paramsElementType = methodHasParams
+            ? ((IArrayTypeSymbol)methodSymbol.Parameters[lastMethodArgumentIndex].Type).ElementType
+            : null;
+        foreach (TypedConstant constructorArgument in constructorArguments)
+        {
+            ITypeSymbol argumentType = constructorArgument.Type!;
```
Evangelink: Not fan of seeing bangs in the code. I'd prefer a null check on 166.

src/Analyzers/MSTest.Analyzers/Resources.resx
```diff
@@ -160,6 +160,28 @@
...
+- should only be set on a test method;
+- should have at least one argument;
+- argument count should match method argument count;
+- argument type should match method argument type.</value>
+  </data>
+  <data name="DataRowShouldBeValidMessageFormat_ArgumentNumber" xml:space="preserve">
+    <value>DataRow argument count should match method argument count</value>
+  </data>
+  <data name="DataRowShouldBeValidMessageFormat_ArgumentType" xml:space="preserve">
+    <value>DataRow argument type should match method argument type</value>
```
Evangelink: It'd be best to gather all the indices where the type mismatch and report it here.

src/Analyzers/MSTest.Analyzers/Resources.resx
```diff
@@ -160,6 +160,28 @@
...
   </data>
+  <data name="DataRowShouldBeValidDescription" xml:space="preserve">
+    <value>DataRow attribute should have the following layout to be valid:
+- should only be set on a test method;
+- should have at least one argument;
+- argument count should match method argument count;
+- argument type should match method argument type.</value>
+  </data>
+  <data name="DataRowShouldBeValidMessageFormat_ArgumentNumber" xml:space="preserve">
+    <value>DataRow argument count should match method argument count</value>
```
Evangelink: Would be good to report arguments count and parameters count, just so in case we have a bug it's easier for users to grasp what the analyzer found/considered.

src/Analyzers/MSTest.Analyzers/DataRowShouldBeValidAnalyzer.cs
```diff
@@ -0,0 +1,172 @@
...
+    }
+
+    private static void AnalyzeAttribute(SymbolAnalysisContext context, AttributeData attribute, IMethodSymbol methodSymbol)
+    {
+        if (attribute.ApplicationSyntaxReference?.GetSyntax() is not { } syntax)
+        {
+            return;
+        }
+
+        if (attribute.ConstructorArguments.Length == 0)
```
Evangelink, line 140: Looking at the code this seems to be a valid and supported scenario that's equivalent to passing an empty array: https://github.com/microsoft/testfx/blob/dd4ceaf4b1d8a9b3d89484e5d72357da393d8a70/src/TestFramework/TestFramework/Attributes/DataSource/DataRowAttribute.cs#L19

I don't find it super nice design and maybe we should report on it but that cannot be through this rule as it's set as warning and so could break people although it's valid (not preferred but still valid).

src/Analyzers/MSTest.Analyzers/Resources.resx
```diff
@@ -160,6 +160,28 @@
   <data name="AvoidExpectedExceptionAttributeTitle" xml:space="preserve">
     <value>Avoid '[ExpectedException]'</value>
   </data>
+  <data name="DataRowShouldBeValidDescription" xml:space="preserve">
+    <value>DataRow attribute should have the following layout to be valid:
```
Evangelink: Shall we say `DataRow entry` instead of `DataRow attribute`?

test/UnitTests/MSTest.Analyzers.UnitTests/DataRowShouldBeValidAnalyzerTests.cs
```diff
@@ -0,0 +1,325 @@
...
+                public void TestMethod1(int i, int j, string s)
+                {
+                }
+            }
+            """
+        ;
+
+        await VerifyCS.VerifyAnalyzerAsync(code, VerifyCS.Diagnostic(DataRowShouldBeValidAnalyzer.ArgumentTypeRule).WithLocation(0));
+    }
+}
```
Evangelink, line 453: It'd be great to have one test with `[DataTestMethod]` instead of `[TestMethod]` and another with a custom local attribute that inherits from `TestMethodAttribute`.

Evangelink:
Also, please use the regular "fixes issue numer" or "closes issue number" text in your description so that the issue is correctly linked to that PR.

src/Analyzers/MSTest.Analyzers/DataRowShouldBeValidAnalyzer.cs
cvpoienaru: Fixed.

src/Analyzers/MSTest.Analyzers/DataRowShouldBeValidAnalyzer.cs
cvpoienaru: Fixed, thanks ! Originally I was only checking if it is params but the meaning changed without changing the name :)

src/Analyzers/MSTest.Analyzers/DataRowShouldBeValidAnalyzer.cs
cvpoienaru: Fixed, thanks.

src/Analyzers/MSTest.Analyzers/DataRowShouldBeValidAnalyzer.cs
cvpoienaru: Fixed.

src/Analyzers/MSTest.Analyzers/Resources.resx
cvpoienaru: Fixed.

src/Analyzers/MSTest.Analyzers/Resources.resx
cvpoienaru: Fixed, thanks.

src/Analyzers/MSTest.Analyzers/Resources.resx
cvpoienaru: Fixed, thanks.

src/Analyzers/MSTest.Analyzers/DataRowShouldBeValidAnalyzer.cs
cvpoienaru: Fixed.

src/Analyzers/MSTest.Analyzers/DataRowShouldBeValidAnalyzer.cs
cvpoienaru: Tested it and it passes, as in it doesn't generate any diagnostics.

test/UnitTests/MSTest.Analyzers.UnitTests/DataRowShouldBeValidAnalyzerTests.cs
cvpoienaru: Done, thanks for the suggestions !

src/Analyzers/MSTest.Analyzers/DataRowShouldBeValidAnalyzer.cs
cvpoienaru: I would argue that breaking is desired in this case. The fix is very low effort and we already agreed that using an empty `DataRow` makes no sense so this will declutter the test code. This will also ease the way to deprecating the no args constructor.

Evangelink: We are not releasing a breaking change version so we SHOULD NOT do breaking changes. This case should be dropped for now.

> This will also ease the way to deprecating the no args constructor.

We cannot deprecate it until we have removed all .NET Framework users as this is required for .NET Framework users enabling the CLS compliant functionality.

test/UnitTests/MSTest.Analyzers.UnitTests/DataRowShouldBeValidAnalyzerTests.cs
```diff
@@ -0,0 +1,398 @@
...
+            using Microsoft.VisualStudio.TestTools.UnitTesting;
+
+            [TestClass]
+            public class MyTestClass
+            {
+                [DataRow(null)]
+                [TestMethod]
+                public void TestMethod1()
+                {
+                }
```
Evangelink: As I was saying previously, this case is not supported and does throw an error at runtime:

image

so we need arguments count mismatch rule here.

test/UnitTests/MSTest.Analyzers.UnitTests/DataRowShouldBeValidAnalyzerTests.cs
```diff
@@ -0,0 +1,398 @@
...
+    {
+        var code = """
+            using Microsoft.VisualStudio.TestTools.UnitTesting;
+
+            [TestClass]
+            public class MyTestClass
+            {
+                [{|#0:DataRow()|}]
+                [TestMethod]
+                public void TestMethod1(int i)
```
Evangelink, line 344: This case throws at runtime and should be handled through the count mismatch rule.

src/Analyzers/MSTest.Analyzers/DataRowShouldBeValidAnalyzer.cs
Evangelink: It doesn't

test/UnitTests/MSTest.Analyzers.UnitTests/DataRowShouldBeValidAnalyzerTests.cs
cvpoienaru: Thanks for the demo, I removed the testcase.

test/UnitTests/MSTest.Analyzers.UnitTests/DataRowShouldBeValidAnalyzerTests.cs
cvpoienaru: Fixed, thanks.

test/UnitTests/MSTest.Analyzers.UnitTests/DataRowShouldBeValidAnalyzerTests.cs
cvpoienaru: Sorry, I misread the comment. I put the test case back in but we're handling it with count mismatch.

src/Analyzers/MSTest.Analyzers/DataRowShouldBeValidAnalyzer.cs
cvpoienaru: This is now fixed, thanks.

src/Analyzers/MSTest.Analyzers/DataRowShouldBeValidAnalyzer.cs
cvpoienaru: Removed this case, thanks.

src/Analyzers/MSTest.Analyzers/DataRowShouldBeValidAnalyzer.cs
```diff
@@ -0,0 +1,214 @@
...
+                ArgumentTypeMismatchRule,
+                FormatTypeMismatchIndexList(typeMismatchIndices)));
+        }
+    }
+
+    private static bool IsArgumentCountMismatch(int constructorArgumentsLength, int methodParametersLength, bool lastMethodParameterIsArray)
+    {
+        // 1. If last method parameter is not an array the lengths must be the same.
+        // 2. If last method parameter is an array the argument count check is relaxed and we only
+        //    need to make sure we don't have less constructor arguments than actual method paramters.
```
Evangelink: I am surprised by the explanation. I would expect the following cases:

1. last parameter is params, then we need arguments count >= parameters count - 1
2. single parameter as array, arguments count >= 0
3. otherwise both lengths should match.

src/Analyzers/MSTest.Analyzers/DataRowShouldBeValidAnalyzer.cs
```diff
@@ -0,0 +1,214 @@
...
+        return lastMethodParameterIsArray
+            ? constructorArgumentsLength < methodParametersLength - 1
+            : constructorArgumentsLength != methodParametersLength;
+    }
+
+    private static string FormatTypeMismatchIndexList(List<(int ConstructorArgumentIndex, int MethodParameterIndex)> typeMismatchIndices)
+    {
+        return string.Join(", ", typeMismatchIndices.ToArray());
```
Evangelink: Join has an overload with `IEnumerable` so there is no need to pay for the `ToArray`.

src/Analyzers/MSTest.Analyzers/DataRowShouldBeValidAnalyzer.cs
Evangelink: I don't see a test case with something like `int a, object b` and `[DataRow(1, 2, 3)]`.

src/Analyzers/MSTest.Analyzers/DataRowShouldBeValidAnalyzer.cs
```diff
@@ -0,0 +1,214 @@
...
+                constructorArguments.Length,
+                methodSymbol.Parameters.Length));
+            return;
+        }
+
+        // Check constructor argument types match method parameter types.
+        List<(int ConstructorArgumentIndex, int MethodParameterIndex)> typeMismatchIndices = new();
+        for (int i = 0, j = 0; i < constructorArguments.Length; ++i, ++j)
+        {
+            if (constructorArguments[i].IsNull)
```
Evangelink, line 176: I'd add a comment here in code to say that `null` is considered as `default` for non-nullable types.

src/Analyzers/MSTest.Analyzers/DataRowShouldBeValidAnalyzer.cs
```diff
@@ -0,0 +1,214 @@
...
+            context.ReportDiagnostic(syntax.CreateDiagnostic(
+                ArgumentCountMismatchRule,
+                constructorArguments.Length,
+                methodSymbol.Parameters.Length));
+            return;
+        }
+
+        // Check constructor argument types match method parameter types.
+        List<(int ConstructorArgumentIndex, int MethodParameterIndex)> typeMismatchIndices = new();
+        for (int i = 0, j = 0; i < constructorArguments.Length; ++i, ++j)
```
Evangelink: It's not clear to me why you need `j`?

src/Analyzers/MSTest.Analyzers/DataRowShouldBeValidAnalyzer.cs
```diff
@@ -0,0 +1,214 @@
...
+            }
+
+            ITypeSymbol? argumentType = constructorArguments[i].Type;
+            ITypeSymbol paramType = (lastMethodParameterIsArray && j >= methodSymbol.Parameters.Length - 1)
+                ? ((IArrayTypeSymbol)lastMethodParameter.Type).ElementType
+                : methodSymbol.Parameters[j].Type;
+
+            if (argumentType is not null && !argumentType.IsAssignableTo(paramType, context.Compilation))
+            {
+                typeMismatchIndices.Add((i, j));
```
Evangelink: These 2 indices are always the same in your alg so there isn't much point in having both, or am I missing something?

src/Analyzers/MSTest.Analyzers/DataRowShouldBeValidAnalyzer.cs
cvpoienaru: Cases 1 and 3 are already handled by this condition. Case 2 is directly handled by the algorithm in the sense that this is a supported scenario. We iterate through each constructor argument and compare against method parameter type. If there's a type mismatch then we report diagnostics, but the argument count rule doesn't apply here.

Tests for 2nd case from your list:

image

cvpoienaru: > I don't see a test case with something like `int a, object b` and `[DataRow(1, 2, 3)]`.

There was something similar with 2 int args:

image

I've added the test case you requested too though in case the 2nd param being an object is relevant:

image

src/Analyzers/MSTest.Analyzers/DataRowShouldBeValidAnalyzer.cs
cvpoienaru: You're right, fixed it ! Thanks !

src/Analyzers/MSTest.Analyzers/DataRowShouldBeValidAnalyzer.cs
cvpoienaru: Fixed, thanks.

src/Analyzers/MSTest.Analyzers/DataRowShouldBeValidAnalyzer.cs
cvpoienaru: They are not necessarily the same, the lists could still have different lengths if the last method parameter is an array and there are more ctor args (e.g.: 4 ctor args, 2 method params and last is an actual params array). However, in that case I should be able to use the actual index of the last param.

src/Analyzers/MSTest.Analyzers/DataRowShouldBeValidAnalyzer.cs
cvpoienaru: Replied below why `i` and `j` are not always equal.

src/Analyzers/MSTest.Analyzers/DataRowShouldBeValidAnalyzer.cs
Evangelink: Sorry I gave a wrong example, I mean `int i, object[] o` the test no longer report count mismatch.

What you want is not checking last parameter is array but last parameter is params.

cvpoienaru: This is now fixed as well, thanks !

src/Analyzers/MSTest.Analyzers/DataRowShouldBeValidAnalyzer.cs
```diff
@@ -0,0 +1,231 @@
...
+            || (lastMethodParameterIsArray
+                && !lastMethodParameterIsParams
+                && !uniqueMethodParameter
+                && !hasDefaultValue);
+
+        // 1. If strict matching is required then the constructor arguments count and method
+        //    parameter count must be the same.
+        // 2. If strict matching is not required then the argument count check is relaxed and we
+        //    only need to make sure we don't have less constructor arguments than actual method
+        //    paramters.
```
Evangelink:
```suggestion
        //    parameters.
```

Evangelink (approved):
Nice work @cvpoienaru :shipit: 

Could you please follow-up with the related analyzer doc PR?!

PR #2353, opened 2024-02-16
MSTEST0013: AssemblyCleanup should be valid

engyebrahim:
fix: https://github.com/microsoft/testfx/issues/2230


src/Analyzers/MSTest.Analyzers/AssemblyCleanupShouldBeValidAnalyzer.cs
```diff
@@ -0,0 +1,117 @@
...
+        }
+
+        if (methodSymbol.ReturnsVoid && methodSymbol.IsAsync)
+        {
+            context.ReportDiagnostic(methodSymbol.CreateDiagnostic(NotAsyncVoidRule, methodSymbol.Name));
+        }
+
+        if (methodSymbol.DeclaredAccessibility != Accessibility.Public
+            || (!canDiscoverInternals && methodSymbol.GetResultantVisibility() != SymbolVisibility.Public)
+            || (canDiscoverInternals && methodSymbol.GetResultantVisibility() == SymbolVisibility.Private))
```
nohwnd, line 99: This will not run. 

The declared accessibility of the method should always be public.  The DiscoverInternals should only apply to the class (when disabled only public classes  with [TestClass] should be scanned, if enabled public and internal classes should be scanned).

nohwnd: The assembly initialize rule has the same problem:

https://github.com/microsoft/testfx/pull/2328/files#diff-8a52ce5241463aca67cecbf9557a84903d08cf5e63c7889953064fdfa27ae28dR99

You will run into this:

https://github.com/microsoft/testfx/blob/main/src/Adapter/MSTest.TestAdapter/Extensions/MethodInfoExtensions.cs#L39-L48

@Evangelink  this resource, and other resources don't reflect the changes made for Task and ValueTask https://github.com/microsoft/testfx/blob/main/src/Adapter/MSTest.TestAdapter/Resources/Resource.resx#L212

engyebrahim: @nohwnd  Yeah that's the correct case here, I added tests that test all those scenarios correctly

Evangelink: @nohwnd `DeclaredAccesibility` is what is declared on the method and is the first check Engy's doing here. The other checks are using `GetResultantVisibility` that walks up tree (class + outer classes) to get the final accessibility. We had offline chat with Engy and the code she wrotes LGTM.

Evangelink: > @Evangelink this resource, and other resources don't reflect the changes made for Task and ValueTask https://github.com/microsoft/testfx/blob/main/src/Adapter/MSTest.TestAdapter/Resources/Resource.resx#L212

Thanks! I see some other resources are missing edit. I'll do a follow-up

Evangelink (approved):
Perfect!

src/Analyzers/MSTest.Analyzers/AssemblyCleanupShouldBeValidAnalyzer.cs
nohwnd: You are right, and I am wrong.  :) (For others: The first check makes sure that the class always has `public` on the member, the other two check the visibility due to placement in a internal or private class (or public class in internal class etc.)).

Evangelink: @engyebrahim Would you mind doing a follow-up PR where you add a comment on top of the check with what Jakub wrotes for the various analyzers doing the same trick? It will be helpful for other devs.

PR #2431, opened 2024-02-23
Split Microsoft.Testing.Framework version

nohwnd:
Microsoft.Testing.Framework will rename the nugets and namespaces to Microsoft.Testing.Internal.Framework, split the version in dependencies, so we don't break you when we stop producing it. 

Fix for the new package names and namespaces will revert this.

### DIFF
--- a/eng/Version.Details.xml
+++ b/eng/Version.Details.xml
@@ -17,6 +17,10 @@
       <Uri>https://github.com/microsoft/testanywhere</Uri>
       <Sha>3c6c97befaea627d16aa2444443ae31472dc2278</Sha>
     </Dependency>
+    <Dependency Name="Microsoft.Testing.Framework" Version="1.1.0-preview.24120.3">
+      <Uri>https://github.com/microsoft/testanywhere</Uri>
+      <Sha>3c6c97befaea627d16aa2444443ae31472dc2278</Sha>
+    </Dependency>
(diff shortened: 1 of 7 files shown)


Evangelink:
Thanks @nohwnd. Not sure how the rename is going to affect the version. We would still ship TATF as the same version of the platform. Only MSTest new packages would have a different version.

nohwnd:
When the internal repo will stop producing Microsoft.Testing.Framework* nugets, only Microsoft.Testing.Platform will get updated by maestro, and the version of Microsoft.Testing.Framework will stay "pinned". So build will still be able to resolve all nugets and won't fail. 

A next PR will then unpin the version, and update the name of the consumed package, and the namespaces.

Evangelink:
I think it's really time for me to go on holidays...

> When the internal repo will stop producing Microsoft.Testing.Framework* nuget

Not going to happen soon. We said we rename current package and produce new one so there is no change at all in term of versioning. Am I missing something?

nohwnd:
Yes, the Microsoft.Testing.Framework nuget will rename to Microsoft.Testing.Internal.Framework. So either I can rename it and all consuming repos will get maestro updates that won't be able to merge until me or them fix all their code. Or I can do this change, rename the packages. And then go around the repos one by one updating them to new version, without unannounced broken PR that someone will spend time investigating.

PR #2446, opened 2024-02-26
Add example for simple framework without VSTest bridge

nohwnd:
Add example for simple test framework called Contoso.EnsureTestFramework, that does the bare minimum to wire together the infrastructure needed.


samples/mstest-runner/EnsureTestFramework/EnsureTestFramework/EnsureTestFramework.cs
```diff
@@ -0,0 +1,119 @@
...
+                default:
+                    throw new NotSupportedException();
+            }
+        }
+        finally
+        {
+            context.Complete();
+        }
+
+        return;
```
MarcoRossignoli:
```suggestion
```

PR #2454, opened 2024-02-26
Remove the `Microsoft.Testing.Extensions.Telemetry` from the package

MarcoRossignoli:
Remove the `Microsoft.Testing.Extensions.Telemetry` from the package it will come from the VSBridge


MarcoRossignoli:
@Evangelink  FYI

PR #2457, opened 2024-02-26
MSTEST0010: Flag cases where classinit is set on generic test class

engyebrahim:
fix: https://github.com/microsoft/testfx/issues/2423


src/Analyzers/MSTest.Analyzers/ClassInitializeShouldBeValidAnalyzer.cs
```diff
@@ -62,7 +63,20 @@ public override void Initialize(AnalysisContext context)
...
+        var namedTypeSymbol = context.Symbol.ContainingType;
+
+        bool isClassInitAttrExist = false;
+        bool isInheritanceModeSet = false;
+        foreach (AttributeData attr in methodSymbol.GetAttributes())
+        {
+            if (SymbolEqualityComparer.Default.Equals(attr.AttributeClass, classInitializeAttributeSymbol))
+            {
+                isClassInitAttrExist = true;
+                isInheritanceModeSet = attr.ConstructorArguments.Length == 1 ? true : false;
```
MarcoRossignoli: what if 
`[ClassInitialize(InheritanceBehavior.None)]`

and can you try with named args?

`[ClassInitialize(InheritanceBehavior = InheritanceBehavior.BeforeEachDerivedClass)]`

src/Analyzers/MSTest.Analyzers/ClassInitializeShouldBeValidAnalyzer.cs
```diff
@@ -62,7 +63,20 @@ public override void Initialize(AnalysisContext context)
         INamedTypeSymbol? valueTaskSymbol, INamedTypeSymbol? testContextSymbol, bool canDiscoverInternals)
     {
         var methodSymbol = (IMethodSymbol)context.Symbol;
-        if (!methodSymbol.IsClassInitializeMethod(classInitializeAttributeSymbol))
+        var namedTypeSymbol = context.Symbol.ContainingType;
+
+        bool isClassInitAttrExist = false;
```
nohwnd: rename to `hasClassInitAttribute` ?

src/Analyzers/MSTest.Analyzers/ClassInitializeShouldBeValidAnalyzer.cs
nohwnd: Yeah, the current implementation is quite brittle if the attribute evolves.

src/Analyzers/MSTest.Analyzers/ClassInitializeShouldBeValidAnalyzer.cs
```diff
@@ -75,6 +89,11 @@ public override void Initialize(AnalysisContext context)
             return;
         }
 
+        if (namedTypeSymbol.IsGenericType && !isInheritanceModeSet)
```
nohwnd: perf: This rule is for generic types only. Moving the first part of this check  namedTypeSymbol.IsGenericType to the start of this method, and returning when type is not generic, would save us allocating all the attributes in GetAttributes() for every method on non-generic type.

src/Analyzers/MSTest.Analyzers/ClassInitializeShouldBeValidAnalyzer.cs
```diff
@@ -75,6 +79,36 @@ public override void Initialize(AnalysisContext context)
...
         }
 
+        if (namedTypeSymbol.IsGenericType)
+        {
+            bool isInheritanceModeSet = false;
+            foreach (AttributeData attr in methodSymbol.GetAttributes())
+            {
+                ImmutableArray<TypedConstant> constructorArguments = attr.ConstructorArguments;
+
+                for (int i = 0; i < constructorArguments.Length; ++i)
```
MarcoRossignoli: Why not a `foreach`? Looks easy right?

src/Analyzers/MSTest.Analyzers/ClassInitializeShouldBeValidAnalyzer.cs
```diff
@@ -75,6 +79,36 @@ public override void Initialize(AnalysisContext context)
...
+                {
+                    // Null is considered as default for non-nullable types.
+                    if (constructorArgument.IsNull)
+                    {
+                        continue;
+                    }
+
+                    // We need to check that the inheritanceBehavior is not set to none and it's value inside the enum is zero
+                    if (SymbolEqualityComparer.Default.Equals(constructorArgument.Type, inheritanceBehaviorSymbol)
+                        && constructorArgument.Value?.ToString() != "0")
```
MarcoRossignoli: can this one be null after the above check?

src/Analyzers/MSTest.Analyzers/ClassInitializeShouldBeValidAnalyzer.cs
```diff
@@ -75,6 +79,36 @@ public override void Initialize(AnalysisContext context)
...
+                    {
+                        isInheritanceModeSet = true;
```
MarcoRossignoli, line 107: can we "break" as soon as we find the value? Or it's possible that another one will set it false?

src/Analyzers/MSTest.Analyzers/ClassInitializeShouldBeValidAnalyzer.cs
engyebrahim: yup the first check is for the object itself  not the value

src/Analyzers/MSTest.Analyzers/ClassInitializeShouldBeValidAnalyzer.cs
```diff
@@ -75,6 +80,38 @@ public override void Initialize(AnalysisContext context)
             return;
         }
 
+        if (namedTypeSymbol.IsGenericType)
```
nohwnd: My previous feedback was closed, but I still think this can move up higher.  IsClassInitializeMethod checks all the attributes, but if the type is not generic we don't care about any method on it. (at least that is how I understand the rule.

engyebrahim: the test class doesn't have to be generic to have a classInit method but the rule is
in case the test class is generic and has a classinit method it should have the inheritance behaviour 

so based on that I think we don't need to change 
@nohwnd what do you think?

nohwnd: Yes that is how I understand it as well. 

if class is NOT generic, this rule does not apply at all, and can return immediately. No matter if the method we are inspecting is classInit.

So to me this makes sense to change, because we can return immediately with a cheaper check (class.isGeneric), rather than later with a more expensive check (method.GetAttributes().Any(IsClassInit) && class.IsGeneric)

engyebrahim: the problem is if it's a generic class, I can't Report Diagnostic before making sure first that's a classInit method

nohwnd: I did not notice this rule is implementing other checks to make sure a class initialize method is valid. In that case returning right away would not make sense.

src/Analyzers/MSTest.Analyzers/ClassInitializeShouldBeValidAnalyzer.cs
```diff
@@ -49,19 +51,22 @@ public override void Initialize(AnalysisContext context)
...
             }
         });
     }
 
     private static void AnalyzeSymbol(SymbolAnalysisContext context, INamedTypeSymbol classInitializeAttributeSymbol, INamedTypeSymbol? taskSymbol,
-        INamedTypeSymbol? valueTaskSymbol, INamedTypeSymbol? testContextSymbol, bool canDiscoverInternals)
+        INamedTypeSymbol? valueTaskSymbol, INamedTypeSymbol? testContextSymbol, INamedTypeSymbol? inheritanceBehaviorSymbol, bool canDiscoverInternals)
     {
         var methodSymbol = (IMethodSymbol)context.Symbol;
+        var namedTypeSymbol = context.Symbol.ContainingType;
```
MarcoRossignoli:
```suggestion
        var containingTypeSymbol = context.Symbol.ContainingType;
```

src/Analyzers/MSTest.Analyzers/Resources.resx
```diff
@@ -238,6 +238,9 @@
 - it should not be 'async void'
 - it should not be a special method (finalizer, operator...).</value>
   </data>
+  <data name="ClassInitializeShouldBeValidMessageFormat_NotAGenericClassUnlessInheritanceModeSet" xml:space="preserve">
+    <value>ClassInitialize method '{0}' can't be inside a generic class without the Inheritance mode is set on it</value>
```
MarcoRossignoli:
```suggestion
    <value>ClassInitialize method '{0}' can't be inside a generic class without the InheritanceBehavior mode is set on it</value>
```

MarcoRossignoli (dismissed):

MarcoRossignoli (approved):

PR #2478, opened 2024-02-28
Add .NET Latest data source

nohwnd:
Add .NET latest data source that can be used to decorate a test (unlike NetCurrent).

### DIFF
--- a/test/Utilities/Microsoft.Testing.TestInfrastructure/TargetFrameworks.cs
+++ b/test/Utilities/Microsoft.Testing.TestInfrastructure/TargetFrameworks.cs
@@ -18,6 +18,8 @@ public static class TargetFrameworks
 
     public static TestArgumentsEntry<string> NetCurrent { get; } = Net[0];
 
+    public static TestArgumentsEntry<string>[] NetLatest { get; } = new[] { Net[0] };
+
     public static TestArgumentsEntry<string>[] NetFramework { get; } = new TestArgumentsEntry<string>[] { new("net462", "net462") };
 
     public static TestArgumentsEntry<string>[] All { get; } = RuntimeInformation.IsOSPlatform(OSPlatform.Windows)


test/Utilities/Microsoft.Testing.TestInfrastructure/TargetFrameworks.cs
```diff
@@ -18,6 +18,8 @@ public static class TargetFrameworks
 
     public static TestArgumentsEntry<string> NetCurrent { get; } = Net[0];
 
+    public static TestArgumentsEntry<string>[] NetLatest { get; } = new[] { Net[0] };
```
MarcoRossignoli, line 21: why you need this one?

nohwnd: I am using it for NativeAot tests, the other entry is not usable with ArguementsProvider because it is not an array so then the source gen of internal framework does not work. And I don't want to run native aot tests x times, or having to go and update them to run on latest when latest .net version changes.

MarcoRossignoli: > And I don't want to run native aot tests x times

I don't get why you don't simply use `NetCurrent` so?

nohwnd: NetCurrent does not work work with the ArgumentsProvider because it is not an array. If I change it I would have to fix all the other tests using it, and I don't want to mix that into my PR.

MarcoRossignoli: But why you need an array if you test only 1 tfm? this is not clear..

nohwnd: The current source gen will not generate correct code if this is not an array. Even for 1 item.

PR #2479, opened 2024-02-29
Automatic `MSTestExtension` versioning

MarcoRossignoli:
@Evangelink I had to add the source file because Compile items cannot be added if we don't update the project for "manual" Includes and I don't like the idea to miss some defaults. We can decide to do it...but.

So what will happen is that on CI we will mutate the file before the build directly there for now. We do similar for the SDK the only difference is that the file is "excluded" in the .gitignore.

Another solution can be add a resource...but it's less performant. 

If version is not changed the task won't touch the file so in local we won't see any difference.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -139,7 +139,6 @@ project.fragment.lock.json
 # ===========================
 /src/Package/MSTest.Sdk/Sdk/Sdk.props
 /src/Package/MSTest.Sdk/Sdk/Runner/Runner.targets
-/src/Adapter/MSTest.TestAdapter/Versions.cs
 
 # Playground project (use force add if a change is really worth it)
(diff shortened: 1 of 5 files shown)


nohwnd:
I've seen source generators being used for this. e.g. this: https://github.com/PowerShell/PowerShell/blob/master/src/System.Management.Automation/SourceGenerators/PSVersionInfoGenerator/PSVersionInfoGenerator.cs

or the one that xunit uses: 

https://github.com/devlooped/ThisAssembly

MarcoRossignoli:
Interesting that's another good approach and maybe cleaner too. I think we can invest there too, for now I'll add this way to have good telemetry for release but we should come back this one cc: @Evangelink

Issue #2421, opened 2024-02-22
MSTEST0013: Flag cases where assemblycleanup is set on generic test class 


Evangelink:
## Summary

With the change that will be introduced through the fix of #2419, assembly cleanup should be improved to also raise an issue when it is defined on a generic test class.

[AB#1976481](https://devdiv.visualstudio.com/0bdbc590-a062-4c3f-b0f6-9383f67865ee/_workitems/edit/1976481)

engyebrahim:
`AssemblyCleanup` does't have an inheritance param

PR #2487, opened 2024-02-29
MSTEST0013: Flag cases where assemblycleanup is set on generic test class

engyebrahim:
fix: https://github.com/microsoft/testfx/issues/2421


MarcoRossignoli (approved):

src/Analyzers/MSTest.Analyzers/AssemblyCleanupShouldBeValidAnalyzer.cs
```diff
@@ -61,6 +62,8 @@ public override void Initialize(AnalysisContext context)
         INamedTypeSymbol? valueTaskSymbol, bool canDiscoverInternals)
     {
         var methodSymbol = (IMethodSymbol)context.Symbol;
+        var containingTypeSymbol = context.Symbol.ContainingType;
```
Evangelink, line 65: I would move this variable close to where it's used. Also, I would probably just drop it and use directly `context.Symbol.ContainingType.IsGenericType` on line 80.

src/Analyzers/MSTest.Analyzers/Resources.resx
```diff
@@ -130,6 +130,9 @@
   <data name="AssemblyCleanupShouldBeValidMessageFormat_NoParameters" xml:space="preserve">
     <value>AssemblyCleanup method '{0}' should not take any parameter</value>
   </data>
+  <data name="AssemblyCleanupShouldBeValidMessageFormat_NotAGenericClass" xml:space="preserve">
+    <value>AssemblyCleanup method '{0}' can't be inside a generic class</value>
```
Evangelink, line 134: Maybe `AssemblyCleanup method '{0}' can't be declared on a generic class`

Evangelink (commented):
A few nits, that would be good to apply them to the other rules as I expect they are done similarly.

cc @MarcoRossignoli

MarcoRossignoli:
@engyebrahim pls follow up the @Evangelink comments

PR #2511, opened 2024-03-05
Exclude template file `RepositoryVersion.cs` from git

MarcoRossignoli:
There's some difference between linux and windows Compile item we need a task to fix the build.


src/Platform/Microsoft.Testing.Platform/Microsoft.Testing.Platform.csproj
```diff
@@ -8,6 +8,19 @@
     <SuppressLicenseValidation>true</SuppressLicenseValidation>
   </PropertyGroup>
 
+  <!-- Insert the RepositoryVersion.cs -->
+  <PropertyGroup>
+    <EnableDefaultCompileItems>false</EnableDefaultCompileItems>
+  </PropertyGroup>
+  <ItemGroup>
+    <Compile Include="**/*.cs" Exclude="**/*.user; **/*.*proj; **/*.sln; **/*.vssscc" />
```
Evangelink, line 16: The exclude section looks weird to me. If you include only `.cs` you should not have to exclude other extensions.

MarcoRossignoli: I've copied what is done by default https://learn.microsoft.com/en-us/dotnet/core/project-sdk/overview#default-includes-and-excludes

Issue #2509, opened 2024-03-04
Add `Microsoft.CodeCoverage.MSBuild` to the aot MSTest.SDK for static code coverage


MarcoRossignoli:
[AB#1986978](https://devdiv.visualstudio.com/0bdbc590-a062-4c3f-b0f6-9383f67865ee/_workitems/edit/1986978)

Issue #2508, opened 2024-03-04
Rename MSBuild knob to enable aot engine to `TestEngineName`


MarcoRossignoli:
https://github.com/microsoft/testfx/blob/main/src/Package/MSTest.Sdk/Sdk/Sdk.props.template#L9

thanks @baronfel for the suggestion

[AB#1986977](https://devdiv.visualstudio.com/0bdbc590-a062-4c3f-b0f6-9383f67865ee/_workitems/edit/1986977)

Evangelink:
Different issue: MSTestEngineVersion is defined twice in the linked file (line 11 and 15)

MarcoRossignoli:
> Different issue: MSTestEngineVersion is defined twice in the linked file (line 11 and 15)

I wanted to be sure to not forget it

PR #2516, opened 2024-03-05
Apply `MSTest.Sdk` feedbacks

MarcoRossignoli:
closes https://github.com/microsoft/testfx/issues/2509
closes https://github.com/microsoft/testfx/issues/2508

### DIFF
--- a/src/Package/MSTest.Sdk/Sdk/Runner/Runner.targets.template
+++ b/src/Package/MSTest.Sdk/Sdk/Runner/Runner.targets.template
@@ -3,6 +3,6 @@
 
   <Import Project="$(MSBuildThisFileDirectory)ClassicEngine.targets" Condition=" '$(EngineName)' == 'Classic' "/>
 
-  <Import Project="$(MSBuildThisFileDirectory)TestAnywhereEngine.targets" Condition=" '$(EngineName)' == 'TestAnywhere' "/>
+  <Import Project="$(MSBuildThisFileDirectory)TestingPlatform.targets" Condition=" '$(EngineName)' == 'TestAnywhere' "/>
 
 </Project>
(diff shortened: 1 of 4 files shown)


src/Package/MSTest.Sdk/Sdk/Runner/Runner.targets.template
```diff
@@ -1,8 +1,8 @@
 <?xml version="1.0" encoding="utf-8" ?>
 <Project xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
 
-  <Import Project="$(MSBuildThisFileDirectory)ClassicEngine.targets" Condition=" '$(EngineName)' == 'Classic' "/>
+  <Import Project="$(MSBuildThisFileDirectory)ClassicEngine.targets" Condition=" '$(TestEngineName)' == 'Classic' "/>
 
-  <Import Project="$(MSBuildThisFileDirectory)TestAnywhereEngine.targets" Condition=" '$(EngineName)' == 'TestAnywhere' "/>
+  <Import Project="$(MSBuildThisFileDirectory)TestingPlatform.targets" Condition=" '$(TestEngineName)' == 'TestAnywhere' "/>
```
Evangelink, line 6: I would avoid leaking `TestAnywhere` to public. Can we use something else? I am not sure if this engine relates to the new one supporting reflection + reflection-free or linked to the runner.

MarcoRossignoli: This engine should be the one allow us to run anywhere if it's using or not reflection should not be a detail users should know imo.

I'm fine with whatever name I had already some complain about it, we can have some generic name or specific as "Reflectionless" but I recall wasn't liked during the discussion.
Here testanywhere is only a word it's not a product or something else, I know we're biased by internal usage but outside means "run tests anywhere" and wasn't meant to link with our internal codename.

@nohwnd @pavelhorak do you have some idea to suggest

PR #2521, opened 2024-03-06
Rename to azure-pipelines

nohwnd:
Rename there was a typo in what amaury posted, and I copied it. Now it is missing `s` in the name, and that does not align with Roslyn.


nohwnd:
Merging, it won't have any effect here.

PR #2523, opened 2024-03-06
First pass on making this compliant with the new templates

nohwnd:
The pipeline is compliant, but uses auto baselining, which is okay. First pass because we should also use outputs instead of publishing through inline tasks, and clean up the non-prod pipeline because it does not need all the conditions.


nohwnd:
Next steps are here: https://github.com/microsoft/testfx/issues/2524

MarcoRossignoli (approved):

PR #2556, opened 2024-03-13
MSTEST0017: Assertion arguments should be passed in the correct order

### LINKED ISSUE #2371: Assertion arguments should be passed in the correct order
## Summary

The standard assertions library methods such as `AreEqual` and `AreSame` expect the first argument to be the expected value and the second argument to be the actual value.

Having the expected value and the actual value in the wrong order will not alter the outcome of tests, (succeed/fail when it should) but the error messages will contain misleading information.

This rule raises an issue when the actual argument to an assertions library method is a hard-coded value and the expected argument is not. We assume that a literal, constant or variable/field/property containing `expected` in its name is the `expected` value and that any variable containing `actual` is the `actual` value.

Category: Assertion
Enabled by default: True
Default severity: Info

Evangelink:
Fixes #2371


MarcoRossignoli (dismissed):

src/Analyzers/MSTest.Analyzers/AssertionArgsShouldBePassedInCorrectOrder.cs
```diff
@@ -0,0 +1,121 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System.Collections.Immutable;
+using System.Runtime.CompilerServices;
```
engyebrahim: unused using pls remove it

engyebrahim (commented):
for the tests we don't have any tests that raise the analyzer rule

src/Analyzers/MSTest.Analyzers/AssertionArgsShouldBePassedInCorrectOrder.cs
```diff
@@ -0,0 +1,121 @@
...
+    private static readonly LocalizableResourceString Title = new(nameof(Resources.AssertionArgsShouldBePassedInCorrectOrderTitle), Resources.ResourceManager, typeof(Resources));
+    private static readonly LocalizableResourceString Description = new(nameof(Resources.AssertionArgsShouldBePassedInCorrectOrderDescription), Resources.ResourceManager, typeof(Resources));
+    private static readonly LocalizableResourceString MessageFormat = new(nameof(Resources.AssertionArgsShouldBePassedInCorrectOrderMessageFormat), Resources.ResourceManager, typeof(Resources));
+
+    internal static readonly DiagnosticDescriptor Rule = DiagnosticDescriptorHelper.Create(
+        DiagnosticIds.AssertionArgsShouldBePassedInCorrectOrderRuleId,
+        Title,
+        MessageFormat,
+        Description,
+        Category.Usage,
```
engyebrahim, line 37: category in the task is `Assertion`

PR #2564, opened 2024-03-14
Use System.Text.Json instead of Jsonite

mariam-abdulla:
Currently, we are falling back to Jsonite in case of reference types, so we replaced System.Text.Json with Jsonite by adding serializers and deserializers.

### DIFF
--- a/src/Platform/Microsoft.Testing.Platform/ServerMode/Json/Json.cs
+++ b/src/Platform/Microsoft.Testing.Platform/ServerMode/Json/Json.cs
@@ -491,26 +491,23 @@ internal sealed class Json
                 int id = json.OptionalPropertyBind<int>(jsonElement, JsonRpcStrings.Id);
 
                 object? @params = null;
-                if (method != JsonRpcMethods.Exit)
+                if (jsonElement.TryGetProperty(JsonRpcStrings.Params, out JsonElement value))
                 {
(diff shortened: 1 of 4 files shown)


src/Platform/Microsoft.Testing.Platform/ServerMode/Json/Json.cs
```diff
@@ -155,6 +746,23 @@ internal T Bind<T>(JsonElement element, string? property = null)
             }
         }
 
+        return Deserialize<T>(element);
+    }
+
+    internal bool TryBind<T>(JsonElement element, out T? value, string? property = null)
+    {
+        if (property != null && !element.TryGetProperty(property, out element))
```
MarcoRossignoli:
```suggestion
        if (property is not null && !element.TryGetProperty(property, out element))
```

Use the `is` and `is not` operator for nullable check `==` and `!=` works 99.9% of cases, we can sync on it.

MarcoRossignoli (commented):
great work

MarcoRossignoli (approved):

src/Platform/Microsoft.Testing.Platform/ServerMode/Json/Json.cs
mariam-abdulla: Sure.

MarcoRossignoli (dismissed):

MarcoRossignoli (approved):

Issue #2414, opened 2024-02-22
Timeout support for ClassCleanup, AssemblyCleanup, TestCleanup and TestInitalize


Evangelink:
## Summary

Follow-up from implementation of #704 and #496. Some internal customers are requesting the timeout support to be extended of `ClassCleanup`, `AssemblyCleanup`, `TestCleanup` and `TestInitalize`.

[AB#1987531](https://devdiv.visualstudio.com/0bdbc590-a062-4c3f-b0f6-9383f67865ee/_workitems/edit/1987531)

engyebrahim:
fixed: https://github.com/microsoft/testfx/pull/2570fixed

PR #2570, opened 2024-03-15
Init and cleanup timeout

engyebrahim:
related: https://github.com/microsoft/testfx/issues/2414


src/Adapter/MSTest.TestAdapter/Execution/TestAssemblyInfo.cs
```diff
@@ -59,6 +59,8 @@ internal set
     /// </summary>
     internal int? AssemblyInitializeMethodTimeoutMilliseconds { get; set; }
 
+    internal int? AssemblyCleanupMethodTimeoutMilliseconds { get; set; }
```
MarcoRossignoli, line 65: pls add the comment for coherency with the others member of this class

src/Adapter/MSTest.TestAdapter/Helpers/MethodRunner.cs
```diff
@@ -12,6 +12,72 @@ namespace Microsoft.VisualStudio.TestPlatform.MSTest.TestAdapter.Helpers;
 
 internal static class MethodRunner
 {
+    internal static TestFailedException? RunWithTimeout(
```
MarcoRossignoli: Why don't reuse the below one and simply pass a `CancellationToken.None`? Should it work?

src/Adapter/MSTest.TestAdapter/Execution/TypeCache.cs
```diff
@@ -449,6 +449,22 @@ private TestAssemblyInfo GetAssemblyInfo(Type type)
...
+                        {
+                            var message = string.Format(CultureInfo.CurrentCulture, Resource.UTA_ErrorInvalidTimeout, methodInfo.DeclaringType!.FullName, methodInfo.Name);
+                            throw new TypeInspectionException(message);
+                        }
+
+                        var timeoutAttribute = _reflectionHelper.GetAttribute<TimeoutAttribute>(methodInfo);
+                        DebugEx.Assert(timeoutAttribute != null, "TimeoutAttribute cannot be null");
+                        assemblyInfo.AssemblyCleanupMethodTimeoutMilliseconds = timeoutAttribute.Timeout;
+                    }
+                    else if (MSTestSettings.CurrentSettings.AssemblyInitializeTimeout > 0)
```
MarcoRossignoli: isn't wrong here to use the `AssemblyInitializeTimeout`? I suppose you wanted `MSTestSettings.CurrentSettings.AssemblyCleanupTimeout` here.

src/Adapter/MSTest.TestAdapter/Resources/Resource.Designer.cs
```diff
@@ -61,6 +61,24 @@ internal class Resource {
             }
         }
         
+        /// <summary>
+        ///   Looks up a localized string similar to Assembly cleanup method &apos;{0}.{1}&apos; timed out.
+        /// </summary>
+        internal static string AssemblyCleanupTimedOut {
```
MarcoRossignoli, line 67: I suppose you need to rebuild in local to generate all the translation files.

src/Adapter/MSTest.TestAdapter/Helpers/MethodRunner.cs
engyebrahim: the parameter type below is `CancellationTokenSource` which does not have `none` property.

test/UnitTests/MSTestAdapter.UnitTests/Execution/TestAssemblyInfoTests.cs
```diff
@@ -319,7 +319,7 @@ public void RunAssemblyCleanupShouldThrowTheInnerMostExceptionWhenThereAreMultip
 
         var actualErrorMessage = _testAssemblyInfo.RunAssemblyCleanup();
 
-        Verify(actualErrorMessage.StartsWith("Assembly Cleanup method DummyTestClass.AssemblyCleanupMethod failed. Error Message: System.InvalidOperationException: I fail.. StackTrace:", StringComparison.Ordinal));
+        Verify(actualErrorMessage.StartsWith("Assembly Cleanup method DummyTestClass.AssemblyCleanupMethod failed. Error Message: System.TypeInitializationException: The type initializer for 'FailingStaticHelper' threw an exception. ---> System.InvalidOperationException: I fail.. StackTrace:", StringComparison.Ordinal));
```
MarcoRossignoli: This is strange this stack should not be affected or?

src/Adapter/MSTest.TestAdapter/Execution/TestAssemblyInfo.cs
```diff
@@ -208,33 +213,46 @@ public void RunAssemblyInitialize(TestContext testContext)
...
+                     () => AssemblyCleanupMethod.InvokeAsSynchronousTask(null),
+                     new CancellationTokenSource(),
+                     AssemblyCleanupMethodTimeoutMilliseconds,
+                     AssemblyCleanupMethod,
+                     Resource.AssemblyCleanupWasCancelled,
+                     Resource.AssemblyCleanupTimedOut);
             }
             catch (Exception ex)
             {
-                var realException = ex.GetRealException();
```
MarcoRossignoli, line 221: I think that this one should be preserved right?

You've changed with https://github.com/microsoft/testfx/pull/2570/files#diff-2f90496895f16574f24e69119cf4002a51f31248fa2e1535c57495f4bb725837R241

test/UnitTests/MSTestAdapter.UnitTests/Execution/TestAssemblyInfoTests.cs
engyebrahim: AssemblyCleanup is using also the helper methods and that changing the stack

src/Adapter/MSTest.TestAdapter/Execution/TestAssemblyInfo.cs
engyebrahim: no I have changed it with that: https://github.com/microsoft/testfx/pull/2570/files#diff-2f90496895f16574f24e69119cf4002a51f31248fa2e1535c57495f4bb725837R231

so I don't need to preserve it

src/Adapter/MSTest.TestAdapter/Execution/TestAssemblyInfo.cs
```diff
@@ -208,33 +213,46 @@ public void RunAssemblyInitialize(TestContext testContext)
...
             }
         }
+
+        // If assemblyCleanup was successful, then don't do anything
+        if (assemblyCleanupException is null)
+        {
+            return null;
+        }
+
+        var realException = assemblyCleanupException.InnerException ?? assemblyCleanupException;
```
Evangelink: This change is removing the helper I added to avoid the type initialization exception.

```suggestion
        var realException = ex.GetRealException();
```

src/Adapter/MSTest.TestAdapter/Execution/TestAssemblyInfo.cs
```diff
@@ -251,36 +269,57 @@ internal void ExecuteAssemblyCleanup()
...
         }
+
+        // If assemblyCleanup was successful, then don't do anything
+        if (assemblyCleanupException is null)
+        {
+            return;
+        }
+
+        // Cache and return an already created TestFailedException.
+        if (assemblyCleanupException is TestFailedException)
```
Evangelink, line 298: I am not sure to understand the logic behind this caching, could you please add a comment in the code.

src/Adapter/MSTest.TestAdapter/Execution/TestClassInfo.cs
```diff
@@ -108,6 +111,24 @@ internal set
...
+    /// Gets the timeout for the test initialize methods.
+    /// We can use a dictionary because the MethodInfo is unique in an inheritance hierarchy.
+    /// </summary>
+    internal Dictionary<MethodInfo, int> TestInitializeMethodTimeoutMilliseconds { get; private set; }
+
+    /// <summary>
+    /// Gets the timeout for the test cleanup methods.
+    /// We can use a dictionary because the MethodInfo is unique in an inheritance hierarchy.
+    /// </summary>
+    internal Dictionary<MethodInfo, int> TestCleanupMethodTimeoutMilliseconds { get; private set; }
```
Evangelink: These 2 dictionaries seem to be initialized only from ctor so the properties should be readonly instead.

src/Adapter/MSTest.TestAdapter/Execution/TestClassInfo.cs
```diff
@@ -383,56 +405,57 @@ public void RunClassInitialize(TestContext testContext)
...
+        }
 
-                    // special case AssertFailedException to trim off part of the stack trace
-                    string errorMessage = realException is AssertFailedException or AssertInconclusiveException
-                        ? realException.Message
-                        : realException.GetFormattedExceptionMessage();
+        // If ClassCleanup was successful, then don't do anything
+        if (ClassCleanupException == null)
+        {
+            // We didn't but it in a finally function above to not change the old behavior as it was only in the the try
```
Evangelink: I don't really get this message, could you give me a little more insights?

test/UnitTests/MSTestAdapter.UnitTests/Execution/TestAssemblyInfoTests.cs
Evangelink: This is changed because of the change you did on src/Adapter/MSTest.TestAdapter/Execution/TestAssemblyInfo.cs line 241.

src/Adapter/MSTest.TestAdapter/Execution/TestAssemblyInfo.cs
engyebrahim: in the rest of the function we are creating a `TestFailedException` but in case the helper function returned this error I don't have to take time creating it again

src/Adapter/MSTest.TestAdapter/Execution/TestClassInfo.cs
engyebrahim: comment is related to that line
https://github.com/microsoft/testfx/pull/2570/files/57d6b5a709c2220520229da2b541f22028e07a82#diff-894e540fd1b1ac62982e36ac99212fcf3c5d0123db60951b3809b5c4f569717cR430

as old code has this line inside the try so if exception happens that flag would stay false 
so I didn't add it in `finally`

src/Adapter/MSTest.TestAdapter/Execution/TestAssemblyInfo.cs
MarcoRossignoli: I think that the comment is confusing it's not a cache we simply throw it as is.

src/Adapter/MSTest.TestAdapter/Execution/TestAssemblyInfo.cs
```diff
@@ -251,36 +269,57 @@ internal void ExecuteAssemblyCleanup()
...
             }
         }
+
+        // If assemblyCleanup was successful, then don't do anything
+        if (assemblyCleanupException is null)
+        {
+            return;
+        }
+
+        // Cache and return an already created TestFailedException.
```
MarcoRossignoli:
```suggestion
        // If the exception is already  a `TestFailedException` we throw it as-is
```

src/Adapter/MSTest.TestAdapter/Execution/TestClassInfo.cs
MarcoRossignoli: This is confusing also to me, I would keep the set at line 399 so the semantic is clear, if we don't have exception we run all the init, move down here "far" for the invocation could confuse the reader. The fact that we evaluate the exception here is only because we can set it in 2 place inside try/catch(411) and inside the block catch(421).

src/Adapter/MSTest.TestAdapter/Execution/TestClassInfo.cs
```diff
@@ -458,50 +483,80 @@ internal void ExecuteClassCleanup()
...
-                throw testFailedException;
+                ClassCleanupException = exception;
             }
         }
+
+        // If ClassCleanup was successful, then don't do anything
+        if (ClassCleanupException == null)
+        {
+            // We didn't but it in a finally function above to not change the old behavior as it was only in the the try
+            IsClassCleanupExecuted = true;
```
MarcoRossignoli: same comment as above

src/Adapter/MSTest.TestAdapter/Execution/TestClassInfo.cs
```diff
@@ -458,50 +483,80 @@ internal void ExecuteClassCleanup()
...
+            exceptionStackTraceInfo,
+            realException);
+        ClassCleanupException = testFailedException;
+
+        throw testFailedException;
+    }
+
+    private TestFailedException? InvokeCleanupMethod(MethodInfo? methodInfo)
+    {
+        if (methodInfo is null)
```
MarcoRossignoli: do we expect a null method? Isn't confusing...we return null in 2 case but the semantic is completely different one is a "wrong" parameter the other one is and invocation without errrors.

I would throw if null or?

src/Adapter/MSTest.TestAdapter/Execution/TestMethodInfo.cs
```diff
@@ -467,65 +468,84 @@ private void RunTestCleanupMethod(object classInstance, TestResult result)
...
-            }
+        // If testCleanup was successful, then don't do anything
+        if (testCleanupException == null)
+        {
+            return;
+        }
 
-            Exception realException = ex.GetRealException();
-            string formattedExceptionMessage = realException.GetFormattedExceptionMessage();
+        if (testCleanupException is TestFailedException)
```
MarcoRossignoli, line 481: we could add the comment also here to be aligned with the other place where we take this shortcut

src/Adapter/MSTest.TestAdapter/Execution/TestMethodInfo.cs
```diff
@@ -467,65 +468,84 @@ private void RunTestCleanupMethod(object classInstance, TestResult result)
...
-            var finalStackTraceInfo = cleanupStackTraceInfo != null
+        StackTraceInformation? cleanupStackTraceInfo = null;
+        var realExceptionStackTraceInfo = realException.TryGetStackTraceInformation();
+        if (realExceptionStackTraceInfo != null)
+        {
+            cleanupStackTrace.Append(realExceptionStackTraceInfo.ErrorStackTrace);
+            cleanupStackTraceInfo ??= realExceptionStackTraceInfo;
+        }
+
+        StackTraceInformation? finalStackTraceInfo = null;
```
MarcoRossignoli, line 536: this can be now null, in past wasn't possible are we sure that's fine?

src/Adapter/MSTest.TestAdapter/Execution/TestMethodInfo.cs
```diff
@@ -549,41 +571,105 @@ private bool RunTestInitializeMethod(object classInstance, TestResult result)
...
+            return true;
+        }
 
-            result.Outcome = realException is AssertInconclusiveException ? UTF.UnitTestOutcome.Inconclusive : UTF.UnitTestOutcome.Failed;
-            result.TestFailureException = new TestFailedException(
-                result.Outcome.ToUnitTestOutcome(),
-                errorMessage,
-                stackTrace,
-                realException);
+        if (testInitializeException is TestFailedException)
```
MarcoRossignoli, line 600: let's add the comment here too

src/Adapter/MSTest.TestAdapter/Execution/TestMethodInfo.cs
```diff
@@ -549,41 +571,105 @@ private bool RunTestInitializeMethod(object classInstance, TestResult result)
...
+            errorMessage,
+            stackTrace,
+            realException);
+
         return false;
     }
 
+    private TestFailedException? InvokeInitializeMethod(MethodInfo? methodInfo, object classInstance)
+    {
+        if (methodInfo is null)
```
MarcoRossignoli: same question as above

src/Adapter/MSTest.TestAdapter/Execution/TestMethodInfo.cs
```diff
@@ -549,41 +571,105 @@ private bool RunTestInitializeMethod(object classInstance, TestResult result)
...
+            new CancellationTokenSource(),
+            timeout,
+            methodInfo,
+            Resource.TestInitializeWasCancelled,
+            Resource.TestInitializeTimedOut);
+    }
+
+    private TestFailedException? InvokeCleanupMethod(MethodInfo? methodInfo, object classInstance)
+    {
+        if (methodInfo is null)
```
MarcoRossignoli: same question as above

test/IntegrationTests/MSTest.Acceptance.IntegrationTests/InitializeAndCleanupTimeoutTests.cs
```diff
@@ -286,6 +370,62 @@ public static async Task ClassInit(TestContext testContext)
...
+        }
+        else
+        {
+            await Task.CompletedTask;
+        }
+    }
+
+    $TimeoutAttribute$
+    [ClassCleanup]
+    public static async Task ClassCleanupMethod()
```
MarcoRossignoli: can you move the cleanup method after the relative init for clarity?

I got the reason for the inner structure but it's harder to review imo

MarcoRossignoli (commented):
Some question/nit, the most important is that ref that now can be null that scare me a bit but generally good

src/Adapter/MSTest.TestAdapter/Execution/TestAssemblyInfo.cs
engyebrahim: updated the comment

src/Adapter/MSTest.TestAdapter/Execution/TestClassInfo.cs
engyebrahim: I could have exception inside the try assigned to the exception var, so I can't set inside the try until i check the var
what do you think?

src/Adapter/MSTest.TestAdapter/Execution/TestMethodInfo.cs
engyebrahim: we have tests with the stack empty and no exception happend

MarcoRossignoli (dismissed):

MarcoRossignoli (approved):

PR #2571, opened 2024-03-15
Get real exception in case of error in AssemblyInitialize

### LINKED ISSUE #2307: Print init and other errors recursively
## Summary

Assembly initialize shows only the top level exception which might not have all the info.

## Background and Motivation

AssemblyIniti is static method, so it lends itself to initializing the behavior in static constructor, when there are additional helpers, they might also initialize in static ctor, and if that throws we don't see the error. 

```cs
namespace TestProject80
{
    [TestClass]
    public class UnitTest1
    {
        [AssemblyInitialize]
        public static void Init(TestContext context)
        {
            // This will throw in ctor.
            InitHelper.A();
        }
        [TestMethod]
        public void TestMethod1()
        {
        }
    }

    public static class InitHelper
    {
        static InitHelper()
        {
            throw new Exception("AAAA");
        }

        public static void A() {  }
    }
}
```

Will fail with: 

```
  Message: 
Assembly Initialization method TestProject80.UnitTest1.Init threw exception. System.TypeInitializationException: The type initializer for 'TestProject80.InitHelper' threw an exception.. Aborting test execution.

  Stack Trace: 
InitHelper.cctor() line 22
--- End of inner exception stack trace ---
InitHelper.A() line 25
UnitTest1.Init(TestContext context) line 10

## Proposed Feature

<!--
Please provide a sketch of the feature you are proposing. Be as specific as you can: the more specific the proposal, the easier the process will be. Including screenshots of some of the existing problems can also help a lot here.
-->

## Alternative Designs

<!--
Were there other options you considered?
How does this compare to features in other editors?
-->

### PR DESCRIPTION
**Pull request by nohwnd**
When my test fails with TargetInvocationException or TypeInitializationException, I get no useful info from the exception. Find the first inner exception that is not this common wrapper exception so user has more information about the fail.

Fix #2307

### DIFF
--- a/src/Adapter/MSTest.TestAdapter/Extensions/ExceptionExtensions.cs
+++ b/src/Adapter/MSTest.TestAdapter/Extensions/ExceptionExtensions.cs
@@ -27,18 +27,10 @@ internal static class ExceptionExtensions
         // TargetInvocationException: Because .NET Framework wraps method.Invoke() into TargetInvocationException.
         // TypeInitializationException: Because AssemblyInitialize is static, and often helpers that are also static
         // are used to implement it, and they fail in constructor.
-        while (exception is TargetInvocationException or TypeInitializationException)
+        while (exception is TargetInvocationException or TypeInitializationException 
+            && exception.InnerException is not null)
         {
-            if (exception.InnerException is not null)
-            {
-                exception = exception.InnerException;
-            }
-            else
-            {
-                // Break because we want to return TargetInvocationException (or TypeInitializationException)
-                // when there is no inner exception, and don't want this loop to run forever.
-                break;
-            }
+            exception = exception.InnerException;
         }
 
         return exception;
--- a/src/Adapter/MSTest.TestAdapter/Extensions/ExceptionExtensions.cs
+++ b/src/Adapter/MSTest.TestAdapter/Extensions/ExceptionExtensions.cs
@@ -27,7 +27,7 @@ internal static class ExceptionExtensions
         // TargetInvocationException: Because .NET Framework wraps method.Invoke() into TargetInvocationException.
         // TypeInitializationException: Because AssemblyInitialize is static, and often helpers that are also static
         // are used to implement it, and they fail in constructor.
-        while (exception is TargetInvocationException or TypeInitializationException 
+        while (exception is TargetInvocationException or TypeInitializationException
             && exception.InnerException is not null)
         {
             exception = exception.InnerException;
--- a/src/Adapter/MSTest.TestAdapter/Extensions/ExceptionExtensions.cs
+++ b/src/Adapter/MSTest.TestAdapter/Extensions/ExceptionExtensions.cs
@@ -19,15 +19,30 @@ namespace Microsoft.VisualStudio.TestPlatform.MSTest.TestAdapter.Extensions;
 internal static class ExceptionExtensions
 {
     /// <summary>
-    /// In .NET framework, the exception thrown by the test method invocation is wrapped in a TargetInvocationException, so we
-    /// need to unwrap it to get the real exception.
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: src/Adapter/MSTest.TestAdapter/Extensions/ExceptionExtensions.cs [hunk 1]
```diff
@@ -19,15 +19,30 @@ namespace Microsoft.VisualStudio.TestPlatform.MSTest.TestAdapter.Extensions;
...
 {
     /// <summary>
-    /// In .NET framework, the exception thrown by the test method invocation is wrapped in a TargetInvocationException, so we
-    /// need to unwrap it to get the real exception.
+    /// TargetInvocationException and TypeInitializationException do not carry any useful information
+    /// to the user. Find the first inner exception that has useful information.
     /// </summary>
     internal static Exception GetRealException(this Exception exception)
-        => exception.GetType() == typeof(TargetInvocationException)
-            && exception.Source is "mscorlib" or "System.Private.CoreLib"
```
nohwnd, line 27: @evangelink You added this code here: https://github.com/microsoft/testfx/commit/1e2abd5ff1ab79e8e423dfbaeb8a7254e017256d#diff-6cc57f94b56e0b9b613d4fc0e280124e5b9aa206c13345d3e641716af76e47dfR23-R31 

I think you are trying to make sure that the exception is thrown from methodInfo.Invoke(), but this seems unnecessary, because (in my humble opinion) if there is inner exception, it will almost always have more useful info, even if user throws it themselves.  

I also cannot imagine a scenario where we would be breaking user tests in a way that makes the experience for them worse.

Evangelink: I was trying to be safest possible because I only know how runtime behaves for these exceptions and not what kind of info would be available. But I agree that for these cases it seems logical to do so.

@MarcoRossignoli do you have any case in mind where this wouldn't be a good idea?

src/Adapter/MSTest.TestAdapter/Extensions/ExceptionExtensions.cs
```diff
@@ -19,15 +19,30 @@ namespace Microsoft.VisualStudio.TestPlatform.MSTest.TestAdapter.Extensions;
...
+            {
+                exception = exception.InnerException;
+            }
+            else
+            {
+                // Break because we want to return TargetInvocationException (or TypeInitializationException)
+                // when there is no inner exception, and don't want this loop to run forever.
+                break;
+            }
+        }
```
Evangelink: Nit:

```suggestion
        while (exception is TargetInvocationException or TypeInitializationException 
            && exception.InnerException is not null)
        {
            exception = exception.InnerException;
        }
```

Evangelink (approved):

Evangelink:
@nohwnd Not for this PR but could you have a quick look if there is a need for a similar fix for:
- AssemblyCleanup
- ClassInit
- ClassCleanup

nohwnd:
https://github.com/microsoft/testfx/blob/main/src/Adapter/MSTest.TestAdapter/Execution/TestAssemblyInfo.cs
https://github.com/microsoft/testfx/blob/main/src/Adapter/MSTest.TestAdapter/Execution/TestClassInfo.cs

Yup they need fix. Both use the same ex.InnerException ??  ex pattern that was used here before.

src/Adapter/MSTest.TestAdapter/Extensions/ExceptionExtensions.cs
MarcoRossignoli: I would not expect usage of this exception if not by reflection infra. I feel good to remove till feedback will say opposite.

Evangelink (dismissed):

Evangelink (approved):

PR #2579, opened 2024-03-15
Rename Microsoft.Testing.Framework and related projects to Microsoft.Testing.Internal.Framework

nohwnd:
Microsoft.Testing.Framework namespace is used by the new mstest engine. To avoid confusion rename "TATF" to Microsoft.Testing **.Internal.** Framework


test/UnitTests/Microsoft.Testing.Platform.UnitTests/Messages/TestNodeUidTests.cs
```diff
@@ -1,9 +1,11 @@
 ﻿// Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
-using Microsoft.Testing.Framework;
+using Microsoft.Testing.Internal.Framework;
 using Microsoft.Testing.TestInfrastructure;
 
+using FrameworkTestNodeUid = Microsoft.Testing.Internal.Framework.TestNodeUid;
+
```
Evangelink, line 8:
```suggestion
using FrameworkTestNodeUid = Microsoft.Testing.Internal.Framework.TestNodeUid;
```

Evangelink (commented):
I would expect that you have to update the source gen tooling before moving this PR forward.

nohwnd:
Should work now, Windows tests passed, but not linux, because I screwed up the name casing in csproj, and linux file system is case sensitive.

PR #2602, opened 2024-03-19
Fix parent package path

nohwnd:
Set the parent directory closer to packages to avoid scanning of all artifacts.


nohwnd:
https://dev.azure.com/dnceng/internal/_build/results?buildId=2408011&view=results only trx warnings are there after this change.

PR #2619, opened 2024-03-20
Move IsTestingPlatformApplication from .props to .targets

Evangelink:
Leave a chance to user to override `EnableMSTestRunner` property before we evaluate it for the `IsTestingPlatformApplication` property.


src/Adapter/Build/Common/MSTest.TestAdapter.props
```diff
@@ -2,8 +2,7 @@
 
   <PropertyGroup>
     <EnableMSTestV2CopyResources Condition=" '$(EnableMSTestV2CopyResources)' == '' ">true</EnableMSTestV2CopyResources>
-    <EnableMSTestRunner Condition=" '$(EnableMSTestRunner)' == '' ">false</EnableMSTestRunner>
-    <IsTestingPlatformApplication>$(EnableMSTestRunner)</IsTestingPlatformApplication>
```
Evangelink, line 6: @MarcoRossignoli I don't see this property in the SDK configuration. I think this is a bug but I would like confirmation.

Evangelink: Not needed because it will be set by this props when the adpater will be imported by the MSTest sdk.

Evangelink:
This is not possible.

PR #2634, opened 2024-03-27
Update NativeAOT runner example

nohwnd:


samples/mstest-runner/NativeAotRunner/nuget.config
```diff
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="utf-8"?>
 <configuration>
   <packageSources>
-    <!-- MSTest early access packages. See: https://aka.ms/mstest/preview uncomment this line -->
-    <!-- <add key="test-tools" value="https://pkgs.dev.azure.com/dnceng/public/_packaging/test-tools/nuget/v3/index.json" /> -->
+     <clear />
+     <add key="nuget" value="https://api.nuget.org/v3/index.json" />
```
Evangelink, line 5: We should not do that, this is causing security alerts. We should instead use the `dotnet-public` feed. I think just dropping this `nuget.config` to use the one from parent folder would be enough.

nohwnd:
Using that feed made it fail the restore. Do you have a build that has the compliance errors please? As long as there is a single public feed preceeded by <clear> that should be okay.

Odesláno z aplikace Outlook pro iOS<https://aka.ms/o0ukef>
________________________________
Od: Amaury Levé ***@***.***>
Odesláno: Friday, March 29, 2024 9:22:33 AM
Komu: microsoft/testfx ***@***.***>
Kopie: Jakub Jares ***@***.***>; State change ***@***.***>
Předmět: Re: [microsoft/testfx] Update NativeAOT runner example (PR #2634)


@Evangelink commented on this pull request.

________________________________

In samples/mstest-runner/NativeAotRunner/nuget.config<https://github.com/microsoft/testfx/pull/2634#discussion_r1544225776>:

> @@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="utf-8"?>
 <configuration>
   <packageSources>
-    
-    
+     <clear />
+     <add key="nuget" value="https://api.nuget.org/v3/index.json" />


We should not do that, this is causing security alerts. We should instead use the dotnet-public feed. I think just dropping this nuget.config to use the one from parent folder would be enough.

Evangelink:
> Using that feed made it fail the restore.

Tested and I can repro. The error is coming from the fact that these packages were not mirrored, I have triggered a mirror request so we should be able to drop the extra `nuget.config` file soon.

> Do you have a build that has the compliance errors please? 

It wasn't a CG but it was a new security alert. I am not sure how to find back the closed/fixed alerts.

PR #2672, opened 2024-04-07
Cleanup `TestApplicationBuilder`

MarcoRossignoli:
@Evangelink  this is not more needed right?
I don't see any usage in TA repo I think that now we use the bridge for console and the "default VSTest integration" for compat mode right?


Evangelink (approved):
I'd guess not. I also cannot find any reference.

PR #2674, opened 2024-04-08
Wording suggestions

nohwnd:
Added few details and simplified some sentences.
Rewrote language to use present time, as is the usual case on learn microsoft.


docs/testingplatform/architecture.md
```diff
@@ -32,23 +37,23 @@ Copyright(c) Microsoft Corporation.  All rights reserved.
 Passed! - Failed: 0, Passed: 1, Skipped: 0, Total: 1, Duration: 5ms - Contoso.UnitTests.dll (win-x64 - .NET 9.0.0-preview.1.24080.9)
 ```
 
-When working in the testing space, it is common to collect information from test sessions to evaluate code quality.
```
nohwnd, line 35: This is interesting detail, but it seemed to distract from what you were trying to highlight, which was starting one more process.

docs/testingplatform/architecture.md
```diff
@@ -64,19 +69,19 @@ public static class TestApplicationBuilderExtensions
...
+The process layout looks like this:
 
 ```mermaid
 graph TD;
     TestHostController:'Contoso.UnitTests.exe'-->TestHost:'Contoso.UnitTests.exe;
 ```
 
 > [!NOTE]
-> The provided example assumes a console application layout, which handles the restart process correctly and propagates all command line arguments to the child process.
+> The provided example assumes a console application layout, which handles the start process correctly and propagates all command line arguments to the child process.
```
nohwnd, line 82: start, because restart was suggesting that the currently running process will exit and start again, which is not the case.

docs/testingplatform/pillars.md
```diff
@@ -11,5 +11,5 @@ The main driving factors for the evolution of the new testing platform are:
...
 * **Support all .NET form factors**: Support current and future .NET form factors, including Native AOT.
 * **Performant**: Finding the right balance between features and extension points to avoid bloating the runtime with non-fundamental code. The new test platform is designed to "orchestrate" a test run, rather than providing implementation details on how to do it.
-* **Enough extensible**: The new platform provides essential extensibility points to allow for maximum customization of runtime execution. It allows you to configure the test process host, observe the test process, and consume information from the test framework within the test host process.
+* **Extensible enough**: The new platform is built on extensibility points to allow for maximum customization of runtime execution. It allows you to configure the test process host, observe the test process, and consume information from the test framework within the test host process.
```
nohwnd, line 14: essential evokes to me that you cannot be without them, fundamental is probably more appropriate here, but imho `is built on`, communicates the point the best without any confusion.

docs/testingplatform/registertestframework.md
```diff
@@ -63,4 +64,4 @@ return await testApplication.RunAsync();
 ```
 
 > [!NOTE]
-> Return empty [ITestFrameworkCapabilities](capabilities.md) should not hinder the execution of the test session. The fundamental features of discovering and running tests should always be ensured. The impact should be limited to extensions that may opt out if the test framework lacks a certain feature.
+> Returning empty [ITestFrameworkCapabilities](capabilities.md) should not prevent the execution of the test session. All test frameworks should be capable of discovering and running tests. The impact should be limited to extensions that may opt out if the test framework lacks a certain feature.
```
nohwnd, line 67: I am actually not fully sure what you meant here. 

> The impact should be limited to extensions that may opt out if the test framework lacks a certain feature.

Do you mean that test framework announces capabilities that are additional to discover, and run. And other extensions can use those capabilities to disable themselves based on that? E.g. if my framework can (non-conditionally) provide source code file paths (e.g. via source gen), and I have extension that relies on source code file paths, I can use the capability to enable or disable my extension.

MarcoRossignoli: For instance if your adapter doesn't support the trx capability your tests will run or be discovered and simply if the trx extension is involved will be a trx extension duty decide what to do. Today for instance we print in console something like "trx will be subotimal"

MarcoRossignoli: > And other extensions can use those capabilities to disable themselves based on that?

Not only disable but if implemented get the needed information to work correctly.

MarcoRossignoli (approved):

PR #2675, opened 2024-04-08
Add project samples for MSTest.Sdk

Evangelink:


MarcoRossignoli (approved):

cremor:
@Evangelink You've added 3 projects in this PR. If you compare their "equivalent project configuration when not using MSTest.Sdk" parts, one of them says it references the MSTest meta package while the other two say that they reference individual MSTest packages. And the native AOT sample doesn't even reference MSTest.Analyzers. Is this really correct?

Evangelink:
`MSTest` package is a metapackage that only simplified referencing other packages (one part of the SDK functionality) but is equivalent to referencing directly the packages. Do you have a preference? Would you expect to see a comment saying that?

For Native AOT, we definitely can add the analyzers if that's the point of your question.

cremor:
I know that, I was just wondering if there was a reason for the shown code in the samples. E.g. for AOT it might really not reference the meta package if MSTest.TestAdapter doesn't work with AOT.

I prefer referencing the meta package, but I think the samples should show how the SDK actually works.

Issue #1674, opened 2023-05-19
Do not rely on testhost to define thread apartment


Evangelink:
## Summary

Provide a way for users to specify that the given test(s) should be run on STA. This is particularly helpful for some UI testing with WPF, WinForms or some other client technologies.

## Background and Motivation

Today, MSTest relies on the threading apartment of the hosting testhost to be STA or MTA. As shown in this issue https://github.com/microsoft/vstest/issues/1865 vstest creates a STA testhost for .NET Framework but not for the other frameworks. 

This means that today, users who are targeting .NET and want to run their tests with STA have no mean to do so.

## Proposed Feature

We could implement something similar to what is done by Andrew Arnott for xUnit (see https://github.com/AArnott/Xunit.StaFact) or try to find a way to only ask for users to define the `[STAThread]` attribute on their test and detect the available synchronisation context.

## Alternative Designs

N/A

[AB#2020496](https://devdiv.visualstudio.com/0bdbc590-a062-4c3f-b0f6-9383f67865ee/_workitems/edit/2020496)
```[tasklist]
### Tasks
- [x] MSTest runner should support running tests in STA
- [x] MSTest .NET Core through VSTest should support running tests in STA
```

fforjan:
@Evangelink in such context, i'm always asking the question... what am i doing wrong because I wasnt expecting to be the first one to run 'UI' tests in .NET ? 
does it mean either no does testing in .NET  or no one uses mstest for this, including teams at Microsoft ?

Evangelink:
I'd love to have a good answer to provide you but I don't... 

I am inheriting MSTest codebase and have to admit I wasn't using it prior to joining Microsoft. I am guessing many users are still using MSTest with .NET Framework and so this problem hasn't yet unveiled. I am pretty surprised at the number of differences (in features or behavior) between MSTest for .NET Framework and .NETCore

fforjan:
Dear @Evangelink  any update ?

Evangelink:
No news on that front. We are having discussions about roadmap and future of some of the testing tools lately. I will keep you posted when I have some news.

fforjan:
@Evangelink  Désolé but I'm back to bother you again. 

We are a bit blocked in our migration to .net core.
This is a long journey and I just want to explain why this issue - with no workaround- is blocking us.
We are looking to have a long .net core migration (multi year) due to the different technologies and size of the applications we are using. 
But overall, for each components, we do as such:
[...]

PR #2682, opened 2024-04-09
Add support for STA thread

Evangelink:
When run with MSTest runner, when on Windows and the user specifies `ExecutionApartmentState` entry in the runsettings and the current thread is not already in the right state, we now create a global thread that switches to the requested apartment state.

When run on VSTest, we rely on VSTest setting the right global state.

The test method execution is also fixed for .NET Core to rely on thread when we need to be on STA thread.

Contributes to #1674


src/Adapter/MSTest.TestAdapter/RunConfigurationSettings.cs
```diff
@@ -29,6 +30,11 @@ public RunConfigurationSettings()
     /// </summary>
     public bool CollectSourceInformation { get; private set; }
 
+    /// <summary>
+    /// Gets a value indicating the requested platform apartment state.
+    /// </summary>
+    public ApartmentState? ExecutionApartmentState { get; private set; }
```
Evangelink: I could use `Unknown` instead of `null`.

src/Adapter/MSTestAdapter.PlatformServices/Services/ThreadOperations.cs
```diff
@@ -20,7 +24,38 @@ public class ThreadOperations : IThreadOperations
...
+            // Task execution canceled.
+            return false;
+        }
+    }
+#endif
+
+#if NET6_0_OR_GREATER
+    [System.Runtime.Versioning.SupportedOSPlatform("windows")]
+#endif
+    private static bool ExecuteWithThread(Action action, int timeout, CancellationToken cancellationToken)
```
MarcoRossignoli:
```suggestion
    private static bool ExecuteWithCustomThread(Action action, int timeout, CancellationToken cancellationToken)
```

src/Adapter/MSTestAdapter.PlatformServices/Services/ThreadOperations.cs
```diff
@@ -20,7 +24,38 @@ public class ThreadOperations : IThreadOperations
...
+    private static bool ExecuteWithThreadPool(Action action, int timeout, CancellationToken cancellationToken)
+    {
+        try
+        {
+            var executionTask = Task.Run(action, cancellationToken);
+
+            // False means execution timed out.
+            return executionTask.Wait(timeout, cancellationToken);
+        }
+        catch (OperationCanceledException)
```
MarcoRossignoli: I would verify the ct in the catch == cancellationToken.Token

Evangelink: I was thinking about it but was afraid to do some breaking change as side-effect. WDYT?

src/Adapter/MSTest.TestAdapter/TestingPlatformAdapter/MSTestBridgedTestFramework.cs
```diff
@@ -52,13 +54,50 @@ internal sealed class MSTestBridgedTestFramework : SynchronizedSingleSessionVSTe
...
-#pragma warning disable CA1859 // Use concrete types when possible for improved performance
-        ITestExecutor testExecutor = new MSTestExecutor();
-#pragma warning restore CA1859 // Use concrete types when possible for improved performance
+        MSTestExecutor testExecutor = new();
         using (cancellationToken.Register(testExecutor.Cancel))
+        {
+            if (RuntimeInformation.IsOSPlatform(OSPlatform.Windows))
+            {
+                ApartmentState currentApartmentState = Thread.CurrentThread.GetApartmentState();
+                ApartmentState? requestedApartmentState = MSTestSettings.RunConfigurationSettings.ExecutionApartmentState;
```
Evangelink: If this is set on Linux, I may want to emit some warning (both for runner and vstest).

src/Adapter/MSTestAdapter.PlatformServices/Services/ThreadOperations.cs
MarcoRossignoli: ok, wonder if in future a log that report the "wrong" token caould be useful in case of some but let's move on as-is

src/Adapter/MSTestAdapter.PlatformServices/Services/ThreadOperations.cs
```diff
@@ -31,16 +66,13 @@ public bool Execute(Action action, int timeout, CancellationToken cancelToken)
...
-        // (based on user configuration). We are currently relying on testhost to be STA or MTA and will run ALL tests
-        // in this mode which is not ideal for different hosting mechanism not for performance.
         executionThread.SetApartmentState(Thread.CurrentThread.GetApartmentState());
         executionThread.Start();
 
         try
         {
-            var executionTask = Task.Run(() => executionThread.Join(timeout), cancelToken);
-            if (executionTask.Wait(timeout, cancelToken))
+            var executionTask = Task.Run(() => executionThread.Join(timeout), cancellationToken);
```
MarcoRossignoli, line 74: I think that here we don't need anymore the timeout in the join we will catch it with the line 75, anyway we won't stop the thread so less "timeout source" place are easier to deal with

MarcoRossignoli, line 75: Pls comment the alg

Join(timeout) returns false if the custom thread doesn't return in time

src/Adapter/MSTestAdapter.PlatformServices/Services/ThreadOperations.cs
```diff
@@ -31,16 +66,13 @@ public bool Execute(Action action, int timeout, CancellationToken cancelToken)
...
+            if (executionTask.Wait(timeout, cancellationToken))
```
MarcoRossignoli: I think we can use only the `cancellationToken` here and remove the timeout, the wait will:
* throw for exception
* return true and the result can be true/false(expired)
* throw cancelled for the token

src/Adapter/MSTest.TestAdapter/TestingPlatformAdapter/MSTestBridgedTestFramework.cs
```diff
@@ -52,13 +54,50 @@ internal sealed class MSTestBridgedTestFramework : SynchronizedSingleSessionVSTe
...
+                        var threadTask = Task.Run(entryPointThread.Join, cancellationToken);
+#if NET6_0_OR_GREATER
+                        await threadTask.WaitAsync(cancellationToken);
+#else
+#pragma warning disable VSTHRD103 // Call async methods when in an async method
+                        threadTask.Wait(cancellationToken);
+#pragma warning restore VSTHRD103 // Call async methods when in an async method
+#endif
+                    }
+                    catch
```
Evangelink: @MarcoRossignoli I am wondering if that wouldn't be better to let the exception bubble up to the bridge/platform, WDYT?

Evangelink: Or at least maybe print either to the output device or using the VSTest IMessageLogger

src/Adapter/MSTest.TestAdapter/TestingPlatformAdapter/MSTestBridgedTestFramework.cs
```diff
@@ -52,13 +55,51 @@ internal sealed class MSTestBridgedTestFramework : SynchronizedSingleSessionVSTe
...
+                        await threadTask.WaitAsync(cancellationToken);
+#else
+#pragma warning disable VSTHRD103 // Call async methods when in an async method
+                        threadTask.Wait(cancellationToken);
+#pragma warning restore VSTHRD103 // Call async methods when in an async method
+#endif
+                    }
+                    catch (Exception ex)
+                    {
+                        request.FrameworkHandle.SendMessage(TestMessageLevel.Error, ex.ToString());
```
Evangelink: Printing using VSTest interface, I am wondering if that's better to use `IOutputDevice` here instead. WDYT @MarcoRossignoli?

MarcoRossignoli: Is the bridge plugged into it? If so I would keep as it...otherwise mixed stuff could be confusing...if we want use directly the new api we need to take into consideration how much time to move everything, I prefer 1 api than mixed ones.

MarcoRossignoli (dismissed):

src/Adapter/MSTest.TestAdapter/PublicAPI/PublicAPI.Unshipped.txt
```diff
@@ -1 +1,2 @@
 ﻿#nullable enable
+Microsoft.VisualStudio.TestPlatform.MSTest.TestAdapter.RunConfigurationSettings.ExecutionApartmentState.get -> System.Threading.ApartmentState?
```
MarcoRossignoli: do we?

Evangelink: I have followed the pattern but that's true we could make it internal. Let me change.

MarcoRossignoli (approved):

eXpl0it3r:
With #2688 / #3278 there are also attributes to enable STA per test class now

PR #2704, opened 2024-04-13
Introduce `RequestNotSupportedException`

MarcoRossignoli:
This exception should be used by the test framework author in case of unknown request.

@Evangelink feel free to refactor as you want. As soon as will be merged I'll it to the documentation.

This allow to evolve the testing platform with new request without side effect, for instance if we want to know the total amount of tests the will run introducing in future a `CountTestsRequest` to have a better interface(progress bar) we can add it and in case of `RequestNotSupportedException` exception  we simply bail out that feature.

This doesn't mean that capabilities won't be used for this reason, but recognize if the adapter supports or not a request is useful to try to carry on the test session to the end "if possible".


MarcoRossignoli:
cc: @Evangelink

src/Platform/Microsoft.Testing.Platform/Requests/RequestNotSupportedException.cs
```diff
@@ -0,0 +1,15 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+namespace Microsoft.Testing.Platform.Requests;
+
+[Serializable]
```
Evangelink, line 6: I think we should avoid this one based on docs updates: https://github.com/dotnet/docs/issues/34893

src/Platform/Microsoft.Testing.Platform/Requests/RequestNotSupportedException.cs
```diff
@@ -0,0 +1,15 @@
...
+public class RequestNotSupportedException : Exception
```
Evangelink, line 7: Shall we have the usual ctors? Message and inner exception?

Evangelink (commented):
Could it make sense to inherit from `NotSupportException` instead?

src/Platform/Microsoft.Testing.Platform/Requests/RequestNotSupportedException.cs
MarcoRossignoli: I was wondering if make sense...the platform doesn't needs any message, I think we will take the exception and we will do something like
* print in console
* crash
* move on if we're good with it...

So I wonder what the test framework author could put in that message.

But I won't oppose if you want it to "standardize" or have some inheritance in future, it's not sealed atm.

Evangelink: No you are right, we won't do anything about it.

Evangelink (commented):
I have been checking the code and I think we did a fail on our design and we should have had a return to the `ExecuteRequest` api https://github.com/microsoft/testfx/blob/main/src/Platform/Microsoft.Testing.Platform/TestFramework/ITestFramework.cs#L21

MarcoRossignoli:
> I have been checking the code and I think we did a fail on our design and we should have had a return to the ExecuteRequest api https://github.com/microsoft/testfx/blob/main/src/Platform/Microsoft.Testing.Platform/TestFramework/ITestFramework.cs#L21

Discussed a bit off line, fine to me if we want to use NotSupportedException but I would not change the return value of the api, the only 2 scenario come to my mind to return prematurely from this method is for a unhandled exception or missing request support.

MarcoRossignoli:
Close as not needed atm, we can use `NotSupportedException`

Issue #2656, opened 2024-04-05
False positive in TestClassShouldHaveTestMethodAnalyzer (MSTEST0016) for derived class


KrisVandermotten:
## Describe the bug

MSTEST0016 incorrectly reports a diagnostic on a non-static test class that derives from a class containing test methods.

## Steps To Reproduce

```cs
public abstract class BaseClass(string name)
{
    [TestMethod]
    public void TestMethod1UsingName() { }

    [TestMethod]
    public void TestMethod2UsingName() { }
}

[TestClass]
public class Derived1 : BaseClass
{
    public Derived1() : base("foo") { }
}

[TestClass]
public class Derived2 : BaseClass
{
    public Derived2() : base("bar") { }
}
```

## Expected behavior

No MSTEST0016 diagnostics are reported.

## Actual behavior

Two diagnostics are reported:

MSTEST0016	Test class 'Derived1' should have at least one test method or be 'static' with method(s) marked by '[AssemblyInitialization]' and/or '[AssemblyCleanup]'

MSTEST0016	Test class 'Derived2' should have at least one test method or be 'static' with method(s) marked by '[AssemblyInitialization]' and/or '[AssemblyCleanup]'

## Additional context

It goes without saying that the steps to reproduce are a minimal example. Our actual code passes more interesting objects into the base class constructor than a `string name`, which are used by many more than two test methods located in the base class.

[AB#2018952](https://devdiv.visualstudio.com/0bdbc590-a062-4c3f-b0f6-9383f67865ee/_workitems/edit/2018952)

PR #2715, opened 2024-04-15
Fix false positive in TestClassShouldHaveTestMethodAnalyzer (MSTEST0016) for derived class

engyebrahim:
fix: https://github.com/microsoft/testfx/issues/2656


src/Analyzers/MSTest.Analyzers/TestClassShouldHaveTestMethodAnalyzer.cs
```diff
@@ -73,22 +73,29 @@ public override void Initialize(AnalysisContext context)
         bool hasAssemblyAttribute = false;
         bool hasTestMethod = false;
 
-        foreach (var classMember in classSymbol.GetMembers())
+        var baseType = classSymbol;
```
Evangelink: Nit: It's more current type than base type.

test/UnitTests/MSTest.Analyzers.UnitTests/TestClassShouldHaveTestMethodAnalyzerTests.cs
```diff
@@ -101,4 +101,44 @@ public static class {|#0:MyTestClass|}
                 .WithLocation(0)
                 .WithArguments("MyTestClass"));
     }
+
+    public async Task WhenTestClassWithoutAssemblyAttributes_InheritsFromClassHasTestMethod_NoDiagnostic()
```
Evangelink: I would add these test cases:
- Base is not abstract
- Base is not abstract and is marked with `[TestClass]`
- Base is abstract and is marked with `[TestClass]`

And maybe one more where you have base with no test method with a base that has test method. This test will ensure we would get failure if we were changing the main alg to look only at first level base.

test/UnitTests/MSTest.Analyzers.UnitTests/TestClassShouldHaveTestMethodAnalyzerTests.cs
```diff
@@ -101,4 +101,44 @@ public static class {|#0:MyTestClass|}
...
+
+            [TestClass]
+            public class Derived : BaseClass
+            {
+            }
+            """;
+        await VerifyCS.VerifyAnalyzerAsync(code);
+    }
+
+    public async Task WhenTestClassWithoutAssemblyAttributes_InheritsFromClassDoesNotHaveTestMethod_NoDiagnostic()
```
Evangelink: Please add another test where the base (and another with base base) contains only assembly init or cleanup to provde we still raise.

Evangelink (commented):
Nice work @engyebrahim

Evangelink (approved):

PR #2717, opened 2024-04-15
Added AssertInstanceOf overloads with out parameter

### LINKED ISSUE #2702: AssertInstanceOf overloads with out parameter
## Summary

I would like new generic AssertInstanceOf overloads that have an additional out parameter with the casted instance.

## Background and Motivation

I always missed and out parameter for the casted instance that was validated with AssertInstanceOf.
Often I first checked with AssertInstanceOf and in following asserts I want to assert properites on that object as instance of that just validated type, so I always have to cast. But that always felt redundant, as a type check was done with AssertInstanceOf, it just didn't provide a casted instance back.

## Proposed Feature

New overloads of all (3) generic AssertInstanceOf overloads but with and additional out parameter which cast the value to the generic type parameter.

```csharp
public static void IsInstanceOfType<T>([NotNull] object? value, out T instance);
public static void IsInstanceOfType<T>([NotNull] object? value, out T instance, string? message);
public static void IsInstanceOfType<T>([NotNull] object? value, out T instance, string? message, params object?[]? parameters);
```

I have a branch ready for a PR.

## Alternative Designs

~~I don't know of any.~~
Returning T instead, see discussion below

Mrxx99:
This adds Assert.IsInstanceOfType overloads with an out parameter as discussed in #2702 

fixes #2702


Mrxx99:

src/TestFramework/TestFramework/Assertions/Assert.IsInstanceOfType.cs
```diff
@@ -131,6 +149,18 @@ public static void IsInstanceOfType<T>([NotNull] object? value, string? message)
...
+    /// <summary>
+    /// Tests whether the specified object is an instance of the generic
+    /// type and throws an exception if the generic type is not in the
+    /// inheritance hierarchy of the object.
+    /// </summary>
+    /// <typeparam name="T">The expected type of <paramref name="value"/>.</typeparam>
+    public static void IsInstanceOfType<T>([NotNull] object? value, out T instance, string? message, params object?[]? parameters)
+    {
+        IsInstanceOfType(value, typeof(T), message, parameters);
+        instance = (T)value!;
```
Evangelink: Value is validated as not-null so you can do

```suggestion
        instance = (T)value;
```

Evangelink (approved):
A small nit that I will force merge but otherwise LGTM.

Thank you so much for the contribution @Mrxx99 :shipit:

PR #2727, opened 2024-04-17
Update PackageReferences to be properly defined

dansiegel:
# Description

Updates PackageReference's so that they are marked as Implicitly Defined. This provides better compatibility with CentralPackageManagement, and with the added `Sdk` meta property it makes it easier to filter Packages added by the SDK.

Fixes #2726

PR #2732, opened 2024-04-18
Remove SDK part of the PackageReference

Evangelink:
I cannot find any documentation about this and I don't see any project using this pattern.

Follow-up from #2727


dansiegel:
https://github.com/dotnet/maui/blob/8e4450cbc14932a6c74aeb8b7bfee9c94eca18b0/src/Controls/src/Build.Tasks/nuget/buildTransitive/netstandard2.0/Microsoft.Maui.Controls.Globs.props#L20

@evangelink you'll see the pattern used here... it's a good way to tag items when they are described from an SDK or Workload as they provide additional metadata that makes it easier to filter from custom targets if you needed to.

Evangelink:
Oh cool! I was checking in .NET SDK and online and couldn't find it or any associated doc. Let me bring it back and thanks for the follow-up.

Issue #2733, opened 2024-04-19
UnauthorizedAccessException thrown as MSTest.Sdk tries to create TestResults in .NET installation directory


martincostello:
## Describe the bug

Running `dotnet test` in a folder containing a solution that has a project using the MSTest.Sdk SDK fails due to the SDK trying to create the `TestResults` folder in the .NET installation folder (e.g. `C:\Program Files\dotnet`).

## Steps To Reproduce

1. Clone https://github.com/martincostello/alexa-london-travel/commit/f8ab76f6a15535f0a33528147f9599984ab3c474
2. Run `dotnet test` in the root of the repository

## Expected behavior

The command succeeds, with `TestResults` being created in the project directory if it does not exist.

## Actual behavior

An error is logged to the console when `dotnet test` tries to run the `LondonTravel.Skill.NativeAotTests` project as it tries to create the directory in the wrong location, which is in a protected system folder:

```text
C:\Users\Martin.Costello\.nuget\packages\microsoft.testing.platform.msbuild\1.1.0\buildMultiTargeting\Microsoft.Testing.Platform.MSBuild.targets(228,5): error : Tests failed: Unhandled exception. System.UnauthorizedAccessException: Access to the path 'C:\Program Files\dotnet\TestResults' is denied. [C:\Coding\martincostello\alexa-london-travel\test\LondonTravel.Skill.NativeAotTests\LondonTravel.Skill.NativeAotTests.csproj]
C:\Users\Martin.Costello\.nuget\packages\microsoft.testing.platform.msbuild\1.1.0\buildMultiTargeting\Microsoft.Testing.Platform.MSBuild.targets(228,5): error :    at System.IO.FileSystem.CreateDirectory(String fullPath, Byte[] securityDescriptor) [C:\Coding\martincostello\alexa-london-travel\test\LondonTravel.Skill.NativeAotTests\LondonTravel.Skill.NativeAotTests.csproj]
C:\Users\Martin.Costello\.nuget\packages\microsoft.testing.platform.msbuild\1.1.0\buildMultiTargeting\Microsoft.Testing.Platform.MSBuild.targets(228,5): error :    at System.IO.Directory.CreateDirectory(String path) [C:\Coding\martincostello\alexa-london-travel\test\LondonTravel.Skill.NativeAotTests\LondonTravel.Skill.NativeAotTests.csproj]
[...]

PR #2735, opened 2024-04-19
Take path to dll from Reflection when running in PublishAOT but not native

nohwnd:
Fix https://github.com/microsoft/testfx/issues/2733


src/Platform/Microsoft.Testing.Platform/Services/CurrentTestApplicationModuleInfo.cs
```diff
@@ -52,11 +52,27 @@ public string GetCurrentTestApplicationFullPath()
...
         {
+            // Application is running non-native, this is safe to call.
 #pragma warning disable IL3000
             moduleName = Assembly.GetEntryAssembly()?.Location;
 #pragma warning restore IL3000
         }
+        else
+        {
+            bool a = false;
+            if (IsCurrentTestApplicationHostDotnetMuxer && a)
```
MarcoRossignoli: I think this is fine, I wonder if there's some attribute added to check too, I think that the runtime needs to understand the "mode", I would expected or an attribute or something in the runtime config file, if it's in the attribute we could use that path instead to check the current mutex, but this is a way.

MarcoRossignoli: @MichalStrehovsky @vitek-karas is there an attribute emitted to recognize this scenario?

src/Platform/Microsoft.Testing.Platform/Services/CurrentTestApplicationModuleInfo.cs
```diff
@@ -52,11 +52,27 @@ public string GetCurrentTestApplicationFullPath()
 
         if (_runtimeFeature.IsDynamicCodeSupported)
         {
+            // Application is running non-native, this is safe to call.
```
MichalStrehovsky: If you support single file, this is not safe to call. This could be a single file app (`PublishSingleFile`) that returns the same thing from `Assembly.GetEntryAssembly()?.Location` as native AOT.

The better way to check for single file is to compare the `moduleName` you got here with an empty string instead of trying to approximate with IsDynamicCodeSupported.

I think this would not just fix PublishSingleFile, but also fix your bug.

Pretty much every time you think you want to detect Native AOT, you need to step back.

src/Platform/Microsoft.Testing.Platform/Services/CurrentTestApplicationModuleInfo.cs
vitek-karas: In general I would be comfortable with the assumption that the entry point assembly has a location. It will have it in most cases, but we should not assume so (There's a proposal to allow setting the entry assembly to pretty much anything, and then in theory it could happen even in normal CoreCLR non-single file that the assembly doesn't have a location).

src/Platform/Microsoft.Testing.Platform/Services/CurrentTestApplicationModuleInfo.cs
nohwnd: Used this approach, and it works. Thank you!

Evangelink:
I would move the acceptance test in this repo

Evangelink (approved):

PR #2769, opened 2024-04-25
Ensure that fixtures also support sta threading

### LINKED ISSUE #2700: Adding Init/Cleanup timeouts change the threading of these methods
If you have a [TestInitialize] method, previously it would run on an STA thread. Now it runs on an MTA thread.
This seems related to the timeout change, as it runs the initializer finalizer on a separate Task, thus breaking the tests that expected an STA thread.

Even when using the new runsetting `<ExecutionThreadApartmentState>STA</ExecutionThreadApartmentState>`, the thread of the initialized method is still MTA.

Evangelink:
Fixes #2700

### DIFF
--- a/src/Adapter/MSTest.TestAdapter/Helpers/MethodRunner.cs
+++ b/src/Adapter/MSTest.TestAdapter/Helpers/MethodRunner.cs
@@ -3,6 +3,7 @@
 
 using System.Globalization;
 using System.Reflection;
+using System.Runtime.InteropServices;
 
 using Microsoft.VisualStudio.TestPlatform.MSTest.TestAdapter.ObjectModel;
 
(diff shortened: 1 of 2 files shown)


src/Adapter/MSTest.TestAdapter/Helpers/MethodRunner.cs
```diff
@@ -83,6 +91,91 @@ internal static class MethodRunner
...
             }
 
+            throw;
+        }
+    }
+
+#if NET6_0_OR_GREATER
+    [System.Runtime.Versioning.SupportedOSPlatform("windows")]
+#endif
+    private static TestFailedException? RunWithTimeoutAndCancellationWithCustomThread(
```
drognanar: consider: add STA to the name

nohwnd: yes please, `RunWithTimeoutAndCancellationWithSTAThread` is definitely a better name. **Custom** suggests that we are not even using the `System.Threading.Thread`.

drognanar (dismissed):

PR #2825, opened 2024-05-06
Don't start thread/task when not using timeout for fixture methods

### LINKED ISSUE #2802: Unable to set thread specific values in `AssemblyInitialize` since MSTest 3.3
## Describe the bug
After updating my solution from MSTest 3.2.2 to 3.3.1 a few of my tests are failing on the build agent (but not on my local machine).
The failing tests seem to have one thing in common: They rely on a specific `CultureInfo.CurrentCulture`. The required culture is the one I'm using on my machine, but the build agent uses a different culture. But the `CurrentCulture` is set in my `AssemblyInitialize` method, and this worked previously.

## Steps To Reproduce
```cs
    [TestClass]
    public static class Init
    {
        [AssemblyInitialize]
        public static void AssemblyInitialize(TestContext ctx)
        {
            CultureInfo.CurrentCulture = new CultureInfo("th-TH");
        }
    }

    [TestClass]
    public class UnitTest1
    {
        [TestMethod]
        public void TestMethod1()
        {
            Assert.AreEqual("th-TH", CultureInfo.CurrentCulture.Name);
        }
    }
```

## Expected behavior
The test succeeds (it did with MSTest 3.2.2).

## Actual behavior
The test fails. `CurrentCulture` is the culture of the machine which is running the test.

## Additional context
A quote from @MarcoRossignoli in https://github.com/microsoft/testfx/issues/2653#issuecomment-2050338823
> I would not consider a breaking change too, MSTest doesn't give any guarantee for what I know(I don't see documentation around it) on the nature of the threads that are running the "tests method", we only guarantee that the execution order is the expected one. Tools/Tests should not take "dependency" on the runtime non documented/guarantee context.
> 
> We need the freedom to change it for more than one reason, timeout is one...but there're others that could come in future.

I see why you need this. And I also know that relying on a thread specific value is not good in tests. But now that I have such tests (because that worked previously), how can I solve this? Is there another place where I can initialize thread specific values for all tests?

And it would really have helped debugging this if the behaviour change was documented in the release notes!

AB#2051796

Evangelink:
Fixes #2802


src/Adapter/MSTest.TestAdapter/Helpers/MethodRunner.cs
```diff
@@ -16,9 +16,18 @@ internal static class MethodRunner
     internal static TestFailedException? RunWithTimeoutAndCancellation(
         Action action, CancellationTokenSource cancellationTokenSource, int? timeout, MethodInfo methodInfo,
         string methodCancelledMessageFormat, string methodTimedOutMessageFormat)
-        => RuntimeInformation.IsOSPlatform(OSPlatform.Windows) && Thread.CurrentThread.GetApartmentState() == ApartmentState.STA
-            ? RunWithTimeoutAndCancellationWithSTAThread(action, cancellationTokenSource, timeout, methodInfo, methodCancelledMessageFormat, methodTimedOutMessageFormat)
-            : RunWithTimeoutAndCancellationWithThreadPool(action, cancellationTokenSource, timeout, methodInfo, methodCancelledMessageFormat, methodTimedOutMessageFormat);
+    {
+        if (timeout is null)
+        {
+            action();
```
MarcoRossignoli, line 22: is it the cancellation verified before after somewhere out of this method?

Evangelink: Same behavior as before introducing the timeout feature.

MarcoRossignoli (approved):

PR #2829, opened 2024-05-07
Fix parameters/arguments check for data driven tests

### LINKED ISSUE #2828: [3.4 preview] Test parameter type check is very strict
## Describe the bug
After updating MSTest from 3.3.1 to 3.4.0-preview.24257.1 I noticed that one of my data driven tests now fails because of a method parameter type mismatch. But the test ran fine in all versions before 3.4.

## Steps To Reproduce
Example test:
```cs
      [TestMethod]
      [DataRow("Field", "Customer", 1, null)]
      public void TestMethod(string field, string entityType, long entityId, string richTextValue)
      {
      }
```

## Expected behavior
Test passes.

## Actual behavior
Test fails with this message:

```
Test data doesn't match method parameters. Either the count or types are different.
Test expected 4 parameter(s), with types 'String, String, Int64, String',
but received 4 argument(s), with types 'String, String, Int32, null'.
```

## Additional context
I assume the problem is only Int32 vs. Int64.
Is this type check so strict on purpose? Isn't the implicit conversion (which worked in older MSTest versions) enough?

I think this is caused by #2691

nohwnd:
This fixes regression, where we validate types of provided data (e.g. in `[DataRow]`) to the expected parameters of the test method exactly by their types. This is too strict and will fail for any types that can implicitly convert. 

This PR simply invokes the method, and relies on runtime to do the right checks, and we just handle and improve the exceptions that it throws. We only handle the two that are documented here: https://learn.microsoft.com/en-us/dotnet/api/system.reflection.methodbase.invoke?view=net-8.0 as happening when parameters don't match the target method. 

As a result, user can provide int32 number to int64 parameter and it will work. As mentioned in the linked issue.

Fix #2828

### DIFF
--- a/src/Adapter/MSTest.TestAdapter/Extensions/MethodInfoExtensions.cs
+++ b/src/Adapter/MSTest.TestAdapter/Extensions/MethodInfoExtensions.cs
@@ -205,5 +205,5 @@ internal static class MethodInfoExtensions
                 methodParametersLengthOrZero,
                 string.Join(", ", methodParameters?.Select(p => p.ParameterType.Name) ?? Array.Empty<string>()),
                 argumentsLengthOrZero,
-                string.Join(", ", arguments?.Select(a => a?.GetType().Name ?? "null") ?? Array.Empty<string>())));
+                string.Join(", ", arguments?.Select(a => a?.GetType().Name ?? "null") ?? Array.Empty<string>())), innerException);
 }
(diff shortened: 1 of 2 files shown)


src/Adapter/MSTest.TestAdapter/Extensions/MethodInfoExtensions.cs
```diff
@@ -171,55 +171,39 @@ internal static void InvokeAsSynchronousTask(this MethodInfo methodInfo, object?
...
-            task = methodInfo.Invoke(classInstance, arguments) as Task;
+            try
+            {
+                task = methodInfo.Invoke(classInstance, arguments) as Task;
+            }
+            catch (TargetParameterCountException ex)
+            {
+                throw GetParameterCountMismatchException(methodInfo, arguments, methodParameters, methodParametersLengthOrZero, argumentsLengthOrZero, ex);
+            }
+            catch (ArgumentException ex)
```
Evangelink: Nit using the catch/when clause.

```suggestion
            catch (Exception ex) when (ex is TargetParameterCountException or ArgumentException)
```

Evangelink (dismissed):

Evangelink (dismissed):

Evangelink (approved):

PR #2834, opened 2024-05-07
Add sample of MSTest runner for WinUI

Evangelink:
I need to finish work on this sample, I am having some restore issues. Moving the project outside of MSTest makes it work so I am not sure what's interferring.


Evangelink:
It seems to be trying to resolve .NET runtime 6.0.30 on my machine, not sure why but this isn't the case for all machines so I'll merge this sample.

PR #2842, opened 2024-05-10
Cleaning command line validations and adding unit tests

fhnaseer:
- Removed extra validations
- Added command line arguments validation tests


src/Platform/Microsoft.Testing.Platform/CommandLine/PlatformCommandLineProvider.cs
```diff
@@ -34,6 +34,13 @@ internal sealed class PlatformCommandLineProvider : ICommandLineOptionsProvider
     public const string TestHostControllerPIDOptionKey = "internal-testhostcontroller-pid";
     public const string ExitOnProcessExitOptionKey = "exit-on-process-exit";
 
+    public const string VerbosityTrace = "Trace";
+    public const string VerbosityDebug = "Debug";
+    public const string VerbosityInformation = "Information";
+    public const string VerbosityWarning = "Warning";
+    public const string VerbosityError = "Error";
+    public const string VerbosityCritical = "Critical";
```
nohwnd: Nit: Would it be possible to use an array of strings here? The check below seems easier that way, and it seems easier to later print that to help etc.

Nit, because you are just moving the code that was there before, and the set of options here seems pretty final.

src/Platform/Microsoft.Testing.Platform/CommandLine/PlatformCommandLineProvider.cs
```diff
@@ -187,19 +163,19 @@ public Task<ValidationResult> ValidateCommandLineOptionsAsync(ICommandLineOption
...
             RoslynDebug.Assert(pid.Length == 1);
             int parentProcessPid = int.Parse(pid[0], CultureInfo.InvariantCulture);
             try
             {
                 // We let the api to do the validity check before to go down the subscription path.
                 // If we don't fail here but we fail below means that the parent process is not there anymore and we can take it as exited.
-                Process.GetProcessById(parentProcessPid);
+                _ = Process.GetProcessById(parentProcessPid);
             }
             catch (Exception ex)
```
nohwnd: not your code, but we can be more specific and catch ArgumentException.
```suggestion
            catch (ArgumentException ex)
```

test/UnitTests/Microsoft.Testing.Platform.UnitTests/CommandLine/PlatformCommandLineProviderTests.cs
```diff
@@ -0,0 +1,169 @@
...
+        if (!isValid)
+        {
+            Assert.AreEqual(PlatformResources.PlatformCommandLineDiagnosticOptionExpectsSingleArgumentErrorMessage, validateOptionsResult.ErrorMessage);
+        }
+    }
+
+    [Arguments("32", true)]
+    [Arguments("32.32", false)]
+    [Arguments("invalid", false)]
+    public async Task IsValid_If_Port_Is_Integer(string timeout, bool isValid)
```
nohwnd: the parameter is called `timeout` but it should be `port` or `value`

test/UnitTests/Microsoft.Testing.Platform.UnitTests/CommandLine/PlatformCommandLineProviderTests.cs
```diff
@@ -0,0 +1,169 @@
...
+        if (!isValid)
+        {
+            Assert.AreEqual(string.Format(CultureInfo.InvariantCulture, PlatformResources.PlatformCommandLinePortOptionSingleArgument, PlatformCommandLineProvider.PortOptionKey), validateOptionsResult.ErrorMessage);
+        }
+    }
+
+    [Arguments("32", true)]
+    [Arguments("32.32", false)]
+    [Arguments("invalid", false)]
+    public async Task IsValid_If_ClientPort_Is_Integer(string timeout, bool isValid)
```
nohwnd: the parameter is called `timeout` but it should be `port` or `value`

test/UnitTests/Microsoft.Testing.Platform.UnitTests/CommandLine/PlatformCommandLineProviderTests.cs
```diff
@@ -0,0 +1,169 @@
...
+        if (!isValid)
+        {
+            Assert.AreEqual(string.Format(CultureInfo.InvariantCulture, PlatformResources.PlatformCommandLinePortOptionSingleArgument, PlatformCommandLineProvider.ClientPortOptionKey), validateOptionsResult.ErrorMessage);
+        }
+    }
+
+    [Arguments("32", true)]
+    [Arguments("32.32", false)]
+    [Arguments("invalid", false)]
+    public async Task IsValid_If_ExitOnProcessExit_Is_Integer(string timeout, bool isValid)
```
nohwnd: the parameter is called `timeout` but it should be `pid` or `value`

test/UnitTests/Microsoft.Testing.Platform.UnitTests/CommandLine/PlatformCommandLineProviderTests.cs
```diff
@@ -0,0 +1,169 @@
...
+        }
+    }
+
+    public async Task IsValid_If_Diagnostics_Provided_With_Other_Diagnostics_Provided()
+    {
+        var provider = new PlatformCommandLineProvider();
+        var options = new Mock<ICommandLineOptions>();
+        _ = options.Setup(o => o.IsOptionSet(PlatformCommandLineProvider.DiagnosticOptionKey)).Returns(true);
+        _ = options.Setup(o => o.IsOptionSet(PlatformCommandLineProvider.DiagnosticOutputDirectoryOptionKey)).Returns(true);
+        _ = options.Setup(o => o.IsOptionSet(PlatformCommandLineProvider.DiagnosticOutputFilePrefixOptionKey)).Returns(true);
```
nohwnd: consider: testing this without Moq. If you stick with moq, consider running validate on the mock.

(It might be just the horror story I am living through right now on testfx to fix moq based tests).

test/UnitTests/Microsoft.Testing.Platform.UnitTests/CommandLine/TreeNodeFilterCommandLineOptionsProviderTests.cs
```diff
@@ -0,0 +1,48 @@
...
+public class TreeNodeFilterCommandLineOptionsProviderTests : TestBase
+{
+    public TreeNodeFilterCommandLineOptionsProviderTests(ITestExecutionContext testExecutionContext)
+    : base(testExecutionContext)
+    {
+    }
+
+    private static TreeNodeFilterCommandLineOptionsProvider GetProvider()
+    {
+        var extension = new Mock<IExtension>();
```
nohwnd: consider: testing this without moq.

test/UnitTests/Microsoft.Testing.Platform.UnitTests/CommandLine/PlatformCommandLineProviderTests.cs
```diff
@@ -0,0 +1,169 @@
...
+    [Arguments("invalid", false)]
+    public async Task IsValid_If_Verbosity_Has_CorrectValue(string dumpType, bool isValid)
+    {
+        var provider = new PlatformCommandLineProvider();
+        CommandLineOption option = provider.GetCommandLineOptions().First(x => x.Name == PlatformCommandLineProvider.DiagnosticVerbosityOptionKey);
+
+        ValidationResult validateOptionsResult = await provider.ValidateOptionArgumentsAsync(option, [dumpType]).ConfigureAwait(false);
+        Assert.AreEqual(isValid, validateOptionsResult.IsValid);
+
+        if (!isValid)
```
Evangelink: Could be worth validating there is no error message when it's valid.

Also, but that's up to you, I would split the valid and invalid cases as we assert differently.

test/UnitTests/Microsoft.Testing.Platform.UnitTests/CommandLine/PlatformCommandLineProviderTests.cs
Evangelink: You can also use the strict mode by default (in the ctor of the mock) where it will fail if a call is made to some non-mocked API.

test/UnitTests/Microsoft.Testing.Platform.UnitTests/CommandLine/PlatformCommandLineProviderTests.cs
```diff
@@ -0,0 +1,169 @@
...
+        var options = new Mock<ICommandLineOptions>();
+        _ = options.Setup(o => o.IsOptionSet(PlatformCommandLineProvider.DiagnosticOptionKey)).Returns(true);
+        _ = options.Setup(o => o.IsOptionSet(PlatformCommandLineProvider.DiagnosticOutputDirectoryOptionKey)).Returns(true);
+        _ = options.Setup(o => o.IsOptionSet(PlatformCommandLineProvider.DiagnosticOutputFilePrefixOptionKey)).Returns(true);
+
+        ValidationResult validateOptionsResult = await provider.ValidateCommandLineOptionsAsync(options.Object).ConfigureAwait(false);
+        Assert.IsTrue(validateOptionsResult.IsValid);
+    }
+
+    public async Task IsValid_If_AllDiagnostics_Missing()
```
Evangelink: Missing is having a "negative" connotation to me (e.g. something expected is missing). I'd prefer a naming like `WhenNoOptionSpecified`.

Evangelink (commented):
Could you please add assertion that the error message is empty/null when the option is valid?

test/UnitTests/Microsoft.Testing.Platform.UnitTests/CommandLine/PlatformCommandLineProviderTests.cs
fhnaseer: removed mock,

test/UnitTests/Microsoft.Testing.Platform.UnitTests/CommandLine/TreeNodeFilterCommandLineOptionsProviderTests.cs
fhnaseer: removed,

test/UnitTests/Microsoft.Testing.Platform.UnitTests/CommandLine/PlatformCommandLineProviderTests.cs
fhnaseer: split into valid and invalid,

test/UnitTests/Microsoft.Testing.Platform.UnitTests/CommandLine/PlatformCommandLineProviderTests.cs
fhnaseer: updated,

PR #2843, opened 2024-05-10
Flow execution context across fixture methods when using timeout

Evangelink:
## Problem

Before the timeout PR, or when no timeout is being set (runsettings or on fixture/test methods), then all methods are being called synchronously and so the state is "shared" and can be mutated by all steps. For example, when changing the culture in `AssemblyInitialize` this culture will be set for the `ClassInitialize`, all the way down to the test method then back to the `AssemblyCleanup`.

When setting up timeouts, we are starting new thread so we can stop observing the execution when the timeout expires (to overcome non-cooperative cancellation). Because of this thread creation, each thread works in isolation and its state is not preserved and flown to the other methods which is unexpected by users.

## Solution

To overcome this issue, we need to manually capture and restore the right [`ExecutionContext`](thttps://learn.microsoft.com/dotnet/api/system.threading.executioncontext) at the right time. The logic I decided to go with is to flow top-to-bottom meaning that the `AssemblyInitialize` context is flown down to the `ClassInitialize`... This is allowing to have a behavior similar to the one when no timeout is set. 

## Design decision

But there is a question happening for `ClassCleanup` and `AssemblyCleanup` methods where the state when there is no timeout is undeterministic as it depends on the last executed test (for `ClassCleanup`) and last executed `ClassCleanup` for `AssemblyCleanup`. Although, technically, we could mimic this behavior, I think it's better to have the `ClassCleanup` reusing the state of left by the corresponding `ClassInitialize` and have the `AssemblyCleanup` reusing the state left the by corresponding `AssemblyInitialize` as it fixes the undeterminism and it is also closer to what happens for awaited async methods.

## Examples

When timeout is set:

```mermaid
flowchart TD

A[AssemblyInitialize\nCulture is English\nSet culture to Japanese] --> B(ClassInitialize\nCulture is Japanese\nSet culture to Italian)
B --> C(TestInitialize\nCulture is Italian\nSet culture to French)
C --> D(TestMethod\nCulture is French\nSet culture to Czech)
D --> E(TestCleanup\nCulture is Czech\nSet culture to Spanish)
E --> F(ClassCleanup\nCulture is Italian - result of ClassInitialize state and not Spanish result of TestCleanup\nSet culture to Hungarian)
F --> G(AssemblyCleanup\nCulture is Japanese - result of AssemblyInitialize state and not Hungarian of ClassCleanup)
```

When timeout is not set:

```mermaid
flowchart TD

A[AssemblyInitialize\nCulture is English\nSet culture to Japanese] --> B(ClassInitialize\nCulture is Japanese\nSet culture to Italian)
B --> C(TestInitialize\nCulture is Italian\nSet culture to French)
C --> D(TestMethod\nCulture is French\nSet culture to Czech)
D --> E(TestCleanup\nCulture is Czech\nSet culture to Spanish)
E --> F(ClassCleanup\nCulture is Spanish result of TestCleanup\nSet culture to Hungarian)
F --> G(AssemblyCleanup\nCulture is Hungarian result of ClassCleanup)
```

Fixes AB#2051896


test/IntegrationTests/MSTest.Acceptance.IntegrationTests/ThreadContextTests.cs
```diff
@@ -117,12 +209,562 @@ public void TestMethod1()
...
+    public void TestCleanup()
+    {
+        Assert.AreEqual(TestMethodCultureCodeName, CultureInfo.CurrentCulture.Name, "TestCleanup culture check failed");
+        CultureInfo.CurrentCulture = new CultureInfo(TestCleanupCultureCodeName);
+    }
+
+    [ClassCleanup]
+    public static void ClassCleanup()
+    {
+        if (Environment.GetEnvironmentVariable("MSTEST_TEST_FLOW_CONTEXT") == "true")
```
Evangelink, line 335: All the cases where you will see the usage of this env var are cases where I would need input from reviewers. Basically we can either make the code to work like the "sync" case with the same "inconsistencies" to ensure everything is simple. Or we can try to consider that we want to bring some kind of functional logic to the way we save and restore the execution context (what is currently done).

src/Adapter/MSTestAdapter.PlatformServices/Services/ExecutionContextService.cs
```diff
@@ -0,0 +1,175 @@
...
+
+using System.Collections.Concurrent;
+using System.Diagnostics;
+
+namespace Microsoft.VisualStudio.TestPlatform.MSTestAdapter.PlatformServices;
+
+internal static class ExecutionContextService
+{
+    /// <summary>
+    /// The execution context to use by class. This context will be derived from the assembly level context and shared down
```
nohwnd: When you say derived here, what do you mean? That it will by a copy of the thread context that was used to run Assembly setup?

src/Adapter/MSTestAdapter.PlatformServices/Services/ExecutionContextService.cs
```diff
@@ -0,0 +1,175 @@
...
+    /// When we execute the action, we need to ensure we are restoring the execution context that was captured in the logical flow of execution.
+    /// After the action is executed we capture the current execution context and save it for the next action to use based on the current execution context scope.
+    ///
+    /// The logical flow of execution is:
+    /// - AssemblyInitialize execution context is saved at "assembly level" and a copy is flown to each ClassInitialize.
+    /// - ClassInitialize execution context is saved at "class level" and a copy is flown to each TestInitialize/TestMethod/TestCleanup.
+    /// - TestInitialize/TestMethod/TestCleanup execution context is mutating the "instance level"
+    /// - ClassCleanup reuses the "class level" execution context.
+    /// - AssemblyCleanup reuses the "assembly level" execution context.
+    /// </summary>
```
nohwnd, line 52: This comment above would be really nice to have as a mermaid graph on this PR, or as ascii with a single concrete example. 

e.g. I set culture at class level, where it will be visible, and if I mutate it in test, where will the change propagate.

src/Adapter/MSTestAdapter.PlatformServices/Services/ExecutionContextService.cs
```diff
@@ -0,0 +1,175 @@
...
+    internal static void RunActionOnContext(Action action, IExecutionContextScope executionContextScope)
+    {
+        if (GetScopedExecutionContext(executionContextScope) is not { } executionContext)
+        {
```
nohwnd, line 59: Would it be worth it to add trace logging? Are we able to see the values in the context, or this is better solved by debugging?

test/IntegrationTests/MSTest.Acceptance.IntegrationTests/ThreadContextTests.cs
```diff
@@ -117,12 +209,562 @@ public void TestMethod1()
...
+        {
+            Assert.AreEqual(ClassInitCultureCodeName, CultureInfo.CurrentCulture.Name, "ClassCleanup culture check failed");
```
nohwnd: Once the behaviors are complete, it would be helpful if the error was more concrete about the intent. E.g. ClassCleanup should have the same culture as ClassInit, because ....

test/IntegrationTests/MSTest.Acceptance.IntegrationTests/ThreadContextTests.cs
nohwnd: In this case, is the meaning of the flag: MSTEST restore class init context in class cleanup? 

I think that is reasonable behavior in some cases, and not in others. We don't know how many tests will run from the current class, and we don't know what they do or if they even complete, so ensuring that the state is the same as it was when ClassInit completed seems reasonable. 

On the other hand, if I think about storing other data, e.g. in thread local storage, or if we added new functionality that would have explicit ClassContext where users would be able to add their data, then I would like the data from tests to aggregate in the class context, e.g. adding all the database table names my Ordered tests created, and then cleaning them up in ClassCleanup, so I don't have to have a separate DB for each test.

Sounds kinda wild, but if this happens it would be nice to have consistent behavior.

Evangelink: > In this case, is the meaning of the flag: MSTEST restore class init context in class cleanup?

For this exact case of class cleanup, the current non-threaded behavior is to use whatever state left by the last executing test. My implementation is trying to work similarly to async/await with a `using` (try/finally) principle, so the class cleanup will reuse the state left by the init but not by what's under it.

> I think that is reasonable behavior in some cases, and not in others.

Agreed, that's why I am calling it out :) I am on the fence about keeping current logic that seems logical (to me and seems to you) or to ensure we have always the same (or as close as possible) to not having the timeout. My main cons about having the same behavior is that anyways when you have sync vs async you do have differences by runtime and you have to know it so it seems reasonable to also know that having non-cooperative timeout changes the execution model. Obviously that means documenting it out correctly.

As you, and other reviewers can see, the problem only happen for cleanup states. And the reported bug was only complaining about inits not flowing down to the test method.

src/Adapter/MSTestAdapter.PlatformServices/Services/ExecutionContextService.cs
Evangelink: Yes sorry poor wording. Also the alg changed a bit, we are trying to retrieve the cotnext left by `AssemblyInitialize` method, if it doesn't exist (suppressed or not such method) we will capture current state (i.e. main execution thread).

src/Adapter/MSTestAdapter.PlatformServices/Services/ExecutionContextService.cs
Evangelink: Good suggestion, I will do it in the PR and add it in the code (will use a different "graph" for the code comment).

src/Adapter/MSTestAdapter.PlatformServices/Services/ExecutionContextService.cs
Evangelink: What's your idea here? Seeing the state of the classes and instances contexts? Or displaying info about the current scope?

test/IntegrationTests/MSTest.Acceptance.IntegrationTests/ThreadContextTests.cs
Evangelink: Agreed, I'll refactor once we all agree on the direction.

src/Adapter/MSTestAdapter.PlatformServices/Services/ExecutionContextService.cs
nohwnd: I am thinking that someone will report problem that this stuff does not work as expected, and if we could easily dump the context to text, then they could run with trace and see what is happening and which values we move where. But maybe that is just too complicated to put here.

MarcoRossignoli (approved):

src/Adapter/MSTestAdapter.PlatformServices/Services/ExecutionContextService.cs
Evangelink: I think that's a good idea, but as far as I can tell in the code, we only have `IAdapterTraceLogger` with `Warning`, `Error` and `Info`. So I will add a todo in the code for when we will be flowing a proper logger interface.

test/IntegrationTests/MSTest.Acceptance.IntegrationTests/ThreadContextTests.cs
Evangelink: It seems that we have more votes for the current implementation: ClassCleanup and AssemblyCleanup reusing the state of their respective Initialize rather than the state of the previous undeterministic cleanup.

PR #2858, opened 2024-05-16
Cancel tests with CancellationToken without Timeout attribute

### LINKED ISSUE #1305: Cancelling TestContext on non [Timeout] [TestMethod] should still cancel test
## Description

With current implementation, the `CancellationTokenSource` property of the `TestContext` always allows you to cancel a test but if the `[TestMethod]` is not marked with `[Timeout]` nothing will happen. This seems like a confusing behavior and as a user I would expect that  my manual cancellation request is honored.

AB#2050315

fhnaseer:
Fixes #1305


Evangelink:
Thanks for the investigation and detailled explanations in the call @fhnaseer, as discussed let's close this as won't fix.

Issue #1748, opened 2023-10-10
Possibility to treat missing dynamic data as warning?


fx999dev:
Hello,

we dynamically get our data for our unit test from different databases. Sometimes there is no data for specific cases on specific databases. This is not a bug and intended. However these tests do not run because of missing data. This is expected.
The following message is displayed:

> Property or method TestData on XyViewModelTests returns empty  IEnumerable<object[]>

 But these tests are then marked as failed --- is there a possibility to treat such cases (when there is not data) as warning?

Thank you!

[AB#2046260](https://devdiv.visualstudio.com/0bdbc590-a062-4c3f-b0f6-9383f67865ee/_workitems/edit/2046260)

engyebrahim:
Hi @fx999dev,
thanks for the suggestion!
is it gonna be okay if marked as failed and attached a warning with the output?
or you suggested to mark it as ignored with the output with the reason of no driven data received?

Evangelink:
We will add a runsettings MSTest entry that will allow to consider this case as a single test returning "inconclusive" result.

PR #2865, opened 2024-05-17
Run and fail tests when ITestDataSource.GetData returns empty IEnumerable

### LINKED ISSUE #1052: When data source provides no data, it is impossible to see why tests does not run
## Description
When data source provides no data it is very hard to understand why test does not run. There is no warning that can be easily seen. And debugger breakpoint in the data source class is not hit (because it only attaches in run). 

## Steps to reproduce
```cs
using FluentAssertions;
using System;
using System.Linq;
using System.Reflection;

namespace mstest44
{
    [TestClass]
    public class UnitTest1
    {
        [TestMethod]
        [MyDataSource(1)]

        public void TestMethod1(int i)
        {
            
        }

        [TestMethod]
        [MyDataSource(0)]

        public void TestMethod2(int i)
        {

        }
    }

    public class MyDataSource : Attribute, ITestDataSource
    {
        private readonly int _count;

        public MyDataSource(int count)
        {
            _count = count;
        }

        {
            var items = new List<object[]>();
            foreach (var i in Enumerable.Range(100, _count))
            {
                items.Add(new object[] { i });
            }

            return items;
        }

        {
            return null;
        }
    }
}
```

## Expected behavior
I expect to see some warning, but have the ability to force not generating any data. Probably as an additional property on the data source interface?

## Actual behavior

One of my data sources are is not running any tests, because I did not add the data to the collection, and the test stays blue in TE, and I can't figure it out. 

image

## Environment
Latest TP, with latest mstest.

AB#2051847

fhnaseer:
Fixes #1052 and follows #1748


test/IntegrationTests/MSTest.Acceptance.IntegrationTests/DynamicDataTests.cs
```diff
@@ -11,7 +11,8 @@ namespace MSTest.Acceptance.IntegrationTests;
 public class DynamicDataTests : AcceptanceTestBase
```
Evangelink: Can you please rename the file/class into `ParameterizedTestTests` or ` DataTestTests`? You are now testing more than just `DynamicData`.

src/Adapter/MSTest.TestAdapter/Discovery/AssemblyEnumerator.cs
```diff
@@ -302,6 +302,11 @@ private static bool TryProcessTestDataSourceTests(UnitTestElement test, TestMeth
             try
             {
                 data = dataSource.GetData(methodInfo);
+
+                if (!data.Any())
```
Evangelink, line 309: Could you please add a comment to mention the other place to look at/change? And mention this one on the `TstMethodRunner`.

Evangelink (approved):
Good job @fhnaseer, I have added 3 small nits and we can merge after.

test/IntegrationTests/MSTest.Acceptance.IntegrationTests/DynamicDataTests.cs
fhnaseer: Updated,

src/Adapter/MSTest.TestAdapter/Discovery/AssemblyEnumerator.cs
fhnaseer: Added,

Evangelink (approved):

Issue #1698, opened 2023-06-29
Separate Test Results and Logs for AssemblyInitialize, AssemblyCleanup, ClassInitialize and ClassCleanup


Gracie-Work:
## Summary

Nowadays, MSTest only returns test results and appends the logs to the first executed test case for AssemblyInitialize and ClassInitialize, and appends the logs to the last executed test case for ClassCleanup logs and AssemblyCleanup.
 
## Background and Motivation
Substrate engineers highly depend on test logs and results for debugging failures. When the logs are appended to the first executed or the last one, it makes it difficult for users to debug it or find the logs. We also have test failure analysis based on different types of failures. Especially when there are failures in Initialize, we encourage users to dig more into the failures before checking in. But since there are no AssemblyInitialize/ClassInitialize results, it's impossible for us to provide the failure type. 

## Proposed Feature
MSTest framework to return results as follows:
•    AssemblyInitialize Failed/Succeeded – Output logs
•    ClassInitialize Failed/Succeeded – Output logs
•    TestMethod1 Failed/Succeeded – Output logs
•    TestMethod2 Failed/Succeeded – Output logs
•    ClassCleanup Failed/Succeeded -Output logs
•    AssemblyCleanup Failed/Succeeded -Output logs

Evangelink:
Hi @Gracie-Work,

When you are talking about separate logs, do you mean a separate log file? Or do you mean a separate entry in the trx (or VS) or some other tool?

Gracie-Work:
Similar to the issue [#1497 ](https://github.com/microsoft/testfx/issues/1497), we need treat these results as "test case" but different type of test cases. Slightly different from the former issue, we'd like to get the separated result entry in below APIs first. Which I assume it leverages vstest.console.exe in backend.

• Microsoft.VisualStudio.TestPlatform.ObjectModel 
• Microsoft.TestPlatform.TranslationLayer 

In the ideal world, the experience should be unified in vstest.console.exe, TranslationLayer API, VS test explorer.

Evangelink:
Fixed by #2904

Vincezhaoc:
Hi @Evangelink,

I have some questions based on [change log 3.5.0](https://github.com/microsoft/testfx/blob/main/docs/Changelog.md#350---2024-07-15) and [PR2904](https://github.com/microsoft/testfx/pull/2904).

### Purpose for finding this feature.
[...]

PR #2904, opened 2024-05-22
Display fixture (Assembly/Class Initialize/Cleanup) methods as "test" entries

### LINKED ISSUE #1497: (Assembly/Class)(Initialize/Cleanup) methods as independent entries in Test Explorer
## Summary

Consider `AssemblyInitialize`, `AssemblyCleanup`, `ClassInitialize` and `ClassCleanup` as specific "test case" so that they are displayed as specific items in Visual Studio Test Explorer. This would ease understanding issues on these steps.

## Background and Motivation

Before MSTest v3, errors or warnings during the (Assembly/Class)(Initialize/Cleanup) were shown only in the logs (except for specific cases). Starting with MSTest v3, `AssemblyInitialize` messages/errors are associated with 1st test run for given assembly, `AssemblyCleanup` to last test run for the assembly, and similarly, `ClassInitialize` to 1st test run for given class and `ClassCleanup` for last  test run for given class.

This new behaviour allows to not miss any issue but is not ideal, especially in a parallel run context where 1st/last test run could differ, causing different test to be marked as failed.

## Proposed Feature

Consider these methods as specific, non-runnable, entries.

## Alternative Designs

None

fhnaseer:
Fixes #1497, #1698

Adds AssemblyInitialize, AssemblyCleanup, ClassInitialize, ClassCleanup tests methods.

Run all tests
image

Run only one test => Class and Assembly test methods are also executed
image

Run fixture tests only => Ignored
image


src/Adapter/MSTest.TestAdapter/MSTestSettings.cs
```diff
@@ -191,6 +191,11 @@ public static RunConfigurationSettings RunConfigurationSettings
     /// </summary>
     public bool TreatClassAndAssemblyCleanupWarningsAsErrors { get; private set; }
 
+    /// <summary>
+    /// Gets a value indicating whether initialize and cleanup methods are added in test results or not.
+    /// </summary>
+    public bool FixturesEnabled { get; private set; }
```
Evangelink: What about `ConsiderFixturesAreSeparateEntries`? I would also update the comment to make it more explicit this will create a different test case for the fixture and so it will be reported in UI, TRX and discovered tests (can you check that part?)

src/Adapter/MSTest.TestAdapter/PublicAPI/PublicAPI.Unshipped.txt
```diff
@@ -1 +1,3 @@
 ﻿#nullable enable
+Microsoft.VisualStudio.TestPlatform.MSTest.TestAdapter.Execution.TestAssemblyInfo.AssemblyCleanupException.get -> System.Exception?
```
Evangelink: Can we avoid having these public?

src/Adapter/MSTest.TestAdapter/Discovery/AssemblyEnumerator.cs
```diff
@@ -265,7 +265,94 @@ private bool DynamicDataAttached(IDictionary<string, object?> sourceLevelParamet
...
+        if (testMethodInfo == null)
+        {
+            return false;
+        }
+
+        // Add non-runnable tests like AssemblyInitialize, AssemblyCleanup, ClassInitialize, ClassCleanup.
+        if (MSTestSettings.CurrentSettings.FixturesEnabled)
+        {
+            AddNonRunnableTests(testMethodInfo, tests, nonRunnableTests);
+        }
```
Evangelink: I think this deserves to be called directly as it's not linked to dynamic data feature.

src/Adapter/MSTest.TestAdapter/Discovery/AssemblyEnumerator.cs
```diff
@@ -228,7 +228,7 @@ internal virtual TypeEnumerator GetTypeEnumerator(Type type, string assemblyFile
                 {
                     if (discoveryOption == TestDataSourceDiscoveryOption.DuringDiscovery)
                     {
-                        if (DynamicDataAttached(sourceLevelParameters, test, tests))
+                        if (DynamicDataAttached(sourceLevelParameters, test, tests, nonRunnableTests))
```
Evangelink: Because of the line above, if users ask for discovery of data soruces at execution, we won't see the extra entries although this is not linked to data source. The call should be extracted.

src/Adapter/MSTest.TestAdapter/Constants.cs
```diff
@@ -10,6 +10,36 @@ namespace Microsoft.VisualStudio.TestPlatform.MSTest.TestAdapter;
 /// </summary>
 internal static class Constants
 {
+    /// <summary>
+    /// Assembly.
+    /// </summary>
+    internal const string AssemblyFixtures = "Assembly";
```
Evangelink: For all these new entries, please use a naming that conveys the meaning. Is it a prefix? suffix? Category name? Entry display name?

Evangelink (commented):
I'll clone the code and deep dive more tomorrow.

src/Adapter/MSTest.TestAdapter/Discovery/AssemblyEnumerator.cs
fhnaseer: I need testMethodInfo object. That's I am populating here. If we can get this object in a higher level, we can poopulate over there as well

src/Adapter/MSTest.TestAdapter/PublicAPI/PublicAPI.Unshipped.txt
fhnaseer: It can be internal. `AssemblyInitalizeException` is public. So, I followed the same pattern.

src/Adapter/MSTest.TestAdapter/Constants.cs
fhnaseer: Updated,

src/Adapter/MSTest.TestAdapter/MSTestSettings.cs
nohwnd: ReportFixturesToClient? ReportFixtureResultsExplicitly? I think you had one more good word for these somewhere @Evangelink  that was not fixtures. So it is not confused with the Fixtures functionality we wanted for mstest and had in TATF.

nohwnd:
I'd like to review this more thoroughly, but I am not sure when I will be able to do that. Please don't close this too fast.

src/Adapter/MSTest.TestAdapter/MSTestSettings.cs
fhnaseer: I am thinking to have this as enum rather than boolean.

```
<FixturesVisibility>Hidden</FixturesVisibility>
<FixturesVisibility>Visible</FixturesVisibility>
<FixturesVisibility>VisibleOnlyOnFailure</FixturesVisibility>
....
```

src/Adapter/MSTest.TestAdapter/Discovery/AssemblyEnumerator.cs
fhnaseer: updated,

src/Adapter/MSTest.TestAdapter/Discovery/AssemblyEnumerator.cs
fhnaseer: updated,

test/IntegrationTests/TestAssets/TestAssets.sln
```diff
@@ -0,0 +1,181 @@
+﻿
```
Evangelink: Any need for this file?

src/Adapter/MSTest.TestAdapter/PublicAPI/PublicAPI.Unshipped.txt
Evangelink: Not sure why the other is public, let's avoid making things public when there is no strong need for it.

src/Adapter/MSTest.TestAdapter/MSTestSettings.cs
Evangelink: If that's not too complex to handle, then the enum makes sense as it offers some flexibility.

fhnaseer: I checked the possibility of adding enum. It is not complex, but it will require adding new resources. I would suggest to keep it as it is. There might be some changes related to logging for fixture methods and this can be handled at that time. Maybe it will require to add more settings.

src/Adapter/MSTest.TestAdapter/PublicAPI/PublicAPI.Unshipped.txt
fhnaseer: updated,

test/IntegrationTests/TestAssets/TestAssets.sln
fhnaseer: removed,

src/Adapter/MSTest.TestAdapter/MSTestSettings.cs
Evangelink: I agree, let's go minimal for now.

src/Adapter/MSTest.TestAdapter/ObjectModel/FixtureTestResult.cs
```diff
@@ -0,0 +1,30 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+namespace Microsoft.VisualStudio.TestPlatform.MSTest.TestAdapter.ObjectModel;
+
+[Serializable]
+internal class FixtureTestResult
```
Evangelink: I think it would be valuable to also have:
- duration
- stack trace

we also probably need to work on the captured messages.

Evangelink: Could be `sealed`.

src/Adapter/MSTest.TestAdapter/ObjectModel/FixtureTestResult.cs
```diff
@@ -0,0 +1,30 @@
...
+    {
+        IsExecuted = isExecuted;
+        Outcome = outcome;
+        ExceptionMessage = exceptionMessage;
+    }
+
+    /// <summary>
+    /// Gets or sets a value indicating whether the test is executed or not.
+    /// </summary>
+    public bool IsExecuted { get; set; }
```
Evangelink: All the properties could be readonly

src/Adapter/MSTest.TestAdapter/Discovery/AssemblyEnumerator.cs
```diff
@@ -77,9 +76,9 @@ public AssemblyEnumerator(MSTestSettings settings)
     internal ICollection<UnitTestElement> EnumerateAssembly(string assemblyFileName, out ICollection<string> warnings)
     {
         DebugEx.Assert(!StringEx.IsNullOrWhiteSpace(assemblyFileName), "Invalid assembly file name.");
-
         var warningMessages = new List<string>();
         var tests = new List<UnitTestElement>();
+        var fixturesTests = new HashSet<string>();
```
Evangelink, line 82: We are not using this, can it be moved down?

src/Adapter/MSTest.TestAdapter/Discovery/AssemblyEnumerator.cs
```diff
@@ -278,13 +297,82 @@ private bool DynamicDataAttached(IDictionary<string, object?> sourceLevelParamet
...
+    }
+
+    private static void AddFixtureTests(TestMethodInfo testMethodInfo, List<UnitTestElement> tests, HashSet<string> fixtureTests)
+    {
+        string assemblyName = testMethodInfo.Parent.Parent.Assembly.GetName().Name!;
+        string assemblyLocation = testMethodInfo.Parent.Parent.Assembly.Location;
+        string className = testMethodInfo.Parent.ClassType.Name;
+        string classFullName = testMethodInfo.Parent.ClassType.FullName!;
+
+        if (!fixtureTests.Contains(assemblyLocation))
```
Evangelink, line 312: Please add some comments to this method, it's not easy to grasp what you are doing.

src/Adapter/MSTest.TestAdapter/Discovery/AssemblyEnumerator.cs
```diff
@@ -278,13 +297,82 @@ private bool DynamicDataAttached(IDictionary<string, object?> sourceLevelParamet
...
+            {
+                tests.Add(GetClassFixtureTest(testMethodInfo.Parent.ClassCleanupMethod, assemblyName, className, classFullName, assemblyLocation, Constants.ClassCleanupFixtureTrait));
+            }
+        }
+
+        static UnitTestElement GetAssemblyFixtureTest(MethodInfo methodInfo, string assemblyName, string className, string classFullName,
+            string assemblyLocation, string fixtureType)
+        {
+            string methodName = GetMethodName(methodInfo);
+            string[] hierarchy = [null!, assemblyName, $"[{Constants.AssemblyFixturesHierarchyName}]", methodName];
```
Evangelink: Do we need the parameters? cc @drognanar

src/Adapter/MSTest.TestAdapter/Discovery/AssemblyEnumerator.cs
```diff
@@ -278,13 +297,82 @@ private bool DynamicDataAttached(IDictionary<string, object?> sourceLevelParamet
...
+        }
+
+        static UnitTestElement GetFixtureTest(string classFullName, string assemblyLocation, string methodType, string methodName, string[] hierarchy)
+        {
+            var method = new TestMethod(classFullName, methodName,
+                hierarchy, methodName, classFullName, assemblyLocation, false,
+                TestIdGenerationStrategy.FullyQualified);
+            return new UnitTestElement(method)
+            {
+                DisplayName = $"[{methodType}].{methodName}",
```
Evangelink: I would prefer not to have the parameters for the display name

src/Adapter/MSTest.TestAdapter/Discovery/AssemblyEnumerator.cs
```diff
@@ -278,13 +297,82 @@ private bool DynamicDataAttached(IDictionary<string, object?> sourceLevelParamet
...
+        string className = testMethodInfo.Parent.ClassType.Name;
+        string classFullName = testMethodInfo.Parent.ClassType.FullName!;
+
+        if (!fixtureTests.Contains(assemblyLocation))
+        {
+            _ = fixtureTests.Add(assemblyLocation);
+
+            if (testMethodInfo.Parent.Parent.AssemblyInitializeMethod is not null)
+            {
+                tests.Add(GetAssemblyFixtureTest(testMethodInfo.Parent.Parent.AssemblyInitializeMethod, assemblyName, className, classFullName, assemblyLocation, Constants.AssemblyInitializeFixtureTrait));
```
Evangelink: Might be better to use an enum than strings for the fixture type.

src/Adapter/MSTest.TestAdapter/Discovery/AssemblyEnumerator.cs
```diff
@@ -235,7 +235,15 @@ internal virtual TypeEnumerator GetTypeEnumerator(Type type, string assemblyFile
                 {
                     if (discoveryOption == TestDataSourceDiscoveryOption.DuringDiscovery)
                     {
-                        if (DynamicDataAttached(sourceLevelParameters, test, tests))
+                        Lazy<TestMethodInfo?> testMethodInfo = GetTestMethodInfo(sourceLevelParameters, test);
+
+                        // Add fixture tests like AssemblyInitialize, AssemblyCleanup, ClassInitialize, ClassCleanup.
+                        if (MSTestSettings.CurrentSettings.FixturesEnabled && testMethodInfo.Value is not null)
```
Evangelink: This is called foreach type, we may want to have assembly fixtures filled only once.

Evangelink: Also this is called within the foreach of each test case we have found which is not good.

src/Adapter/MSTest.TestAdapter/Discovery/UnitTestDiscoverer.cs
```diff
@@ -165,6 +180,15 @@ internal void SendTestCases(string source, IEnumerable<UnitTestElement> testElem
 
                 discoverySink.SendTestCase(testCase);
             }
+
+            // If there are runnable tests, then add all fixture tests to the discovery sink.
+            if (hasAnyRunnableTests)
+            {
+                foreach (TestCase testCase in fixtureTests)
+                {
+                    discoverySink.SendTestCase(testCase);
```
Evangelink, line 192: I think we are sending the entries twice (here and in the loop above) because you add to the list but don't skip the rest of the logic.

src/Adapter/MSTest.TestAdapter/Execution/TestAssemblyInfo.cs
```diff
@@ -95,6 +95,11 @@ internal set
     /// </summary>
     public Exception? AssemblyInitializationException { get; internal set; }
 
+    /// <summary>
+    /// Gets or sets the assembly cleanup exception.
+    /// </summary>
+    internal Exception? AssemblyCleanupException { get; set; }
```
Evangelink: As far as I can see, it's only set from this class so the setter should be `private`.

src/Adapter/MSTest.TestAdapter/Discovery/AssemblyEnumerator.cs
fhnaseer: I kept the declaration here with the other lists. It is passed to every type so that we do not add duplicate fixtures for the same type.

src/Adapter/MSTest.TestAdapter/Discovery/AssemblyEnumerator.cs
fhnaseer: Added,

src/Adapter/MSTest.TestAdapter/Discovery/UnitTestDiscoverer.cs
fhnaseer: Added code comments. In the above loop we put fixture methods in a separate list and do not add. They are added here.

src/Adapter/MSTest.TestAdapter/Execution/TestAssemblyInfo.cs
fhnaseer: Updated,

src/Adapter/MSTest.TestAdapter/ObjectModel/FixtureTestResult.cs
fhnaseer: updated,

src/Adapter/MSTest.TestAdapter/ObjectModel/FixtureTestResult.cs
fhnaseer: updated,

src/Adapter/MSTest.TestAdapter/Discovery/UnitTestDiscoverer.cs
```diff
@@ -109,13 +112,25 @@ internal void SendTestCases(string source, IEnumerable<UnitTestElement> testElem
             foreach (UnitTestElement testElement in testElements)
             {
                 var testCase = testElement.ToTestCase();
+                bool hasFixtureTraits = testCase.Traits.Any(t => t.Name == Constants.FixturesTestTrait);
 
                 // Filter tests based on test case filters
                 if (filterExpression != null && !filterExpression.MatchTestCase(testCase, (p) => TestMethodFilter.PropertyValueProvider(testCase, p)))
                 {
+                    // If test is a fixture test, add it to the list of fixture tests.
+                    if (hasFixtureTraits)
```
Evangelink, line 121: I think this should be moved out of the filter check but maybe I am missing something in the logic of the flow. For example, if filter is null, you probably still want to show the fixture. Also, we don't really need to process the filtering to add the fixture node.

fhnaseer: This is not for discovering test for the test explorer. It is filtering out tests which needs to be executed.

Scenarios from test explorer
1. Run all tests
1.1. Filter returns everything (tests + fixtures)
1.2. Tests are reported as before
1.3. All fixtures are moved to a separate list and added at the end

2. Run only some tests
2.1. Filter returns all fixtures + selected tests
2.2. Selected tests are report as before
2.3. All fixtures are moved to a separate list and added at the end

3. Run only fixture test
3.1. Filter returns selected fixture test and is reported (status update for this as user explicitly ran this fixture)
3.2. other fixtures are not added to the list (no status update for these)

src/Adapter/MSTest.TestAdapter/Discovery/AssemblyEnumerator.cs
fhnaseer: Yes. These parameters form the tree node

src/Adapter/MSTest.TestAdapter/Discovery/AssemblyEnumerator.cs
fhnaseer: Inner method will populate it only once for Assembly fixtures. There is a check for it.

fhnaseer: Again there is check that skips adding fixtures if already present. I need testMethodInfo object to get fixture methods so need to do this in loop (for each test). If we can store all fixtures in some separate data structure then we can move adding fixtures one step up.

src/Adapter/MSTest.TestAdapter/Discovery/AssemblyEnumerator.cs
fhnaseer: I was thinking about it. Strings are more flexible for the display purpose. [Assembly Fixture @ Microsoft].AssemblyInitialize. This can be done using [Description] attribute but it adds some complexity to the code.
Also, in hierarchy we need to pass string so there will be enum to string conversion over there.

src/Adapter/MSTest.TestAdapter/Discovery/AssemblyEnumerator.cs
fhnaseer: We need to populate full methodName for navigation from TE to code. So there will be one more string to remove parameters.

PR #2907, opened 2024-05-23
Rework fixture analyzers and codefixes

Evangelink:


Evangelink:
I'll add some tests that shows we do multiple fixes in a single iteration.

nohwnd (approved):

PR #2909, opened 2024-05-23
use primary constructor params

SimonCropp:


Evangelink:
Thanks @SimonCropp! We haven't done this change yet because when discussing it within the team we have a couple of people thinking it's making it harder to differentiate fields from local variables. I was tinking to switch the setting to be forcing `this.` along with this change but didn't manage to discuss and agree on anything with the team (yet).

SimonCropp:
yep. no worries. will close for now

PR #2912, opened 2024-05-23
remove redundant variable assignment

SimonCropp:


Evangelink (approved):
Yeah JetBrains is definitely better than VS for all these default analyzers :)

Thanks again @SimonCropp

:shipit:

PR #2919, opened 2024-05-23
Allow test framework to take ownership of the banner

### LINKED ISSUE #2657: Display MSTest banner in runner mode
We should display MSTest banner (version, parallelism mode...)

AB#2050327

Evangelink:
Fixes #2657


src/Platform/Microsoft.Testing.Platform/Capabilities/TestFramework/IBannerMessageOwnerCapability.cs
```diff
@@ -0,0 +1,9 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+namespace Microsoft.Testing.Platform.Capabilities.TestFramework;
+
+public interface IBannerMessageOwnerCapability : ITestFrameworkCapability
+{
+    string GetBannerMessage();
```
Evangelink: Shall we allow for async call? I am wondering if we should provide the platform info here or through the service provider?

MarcoRossignoli: Async could be good...some scanning of files here could be done by the adapters, also I think that when we register we offer the service provider so in case the informations could be taken from there when we decide how to expose these informations, otherwise today means have a simple placehoder IContextInformation or something similar

MarcoRossignoli (commented):
In ServerTestHost.cs line 393 we should not create the capability again but reuse the one inside the service provider. To avoid to lose states between instances and have a singleton one for all the executions.

src/Platform/Microsoft.Testing.Platform/Capabilities/TestFramework/IBannerMessageOwnerCapability.cs
MarcoRossignoli: I think that we should implement it for MSTest so we can shape the most interesting information for it.

src/Platform/Microsoft.Testing.Platform/Hosts/TestHostBuilder.cs
```diff
@@ -32,7 +32,7 @@
 
 namespace Microsoft.Testing.Platform.Hosts;
 
-internal class TestHostBuilder(IFileSystem fileSystem, IRuntimeFeature runtimeFeature, IEnvironment environment, IProcessHandler processHandler, ITestApplicationModuleInfo testApplicationModuleInfo) : ITestHostBuilder
+internal partial class TestHostBuilder(IFileSystem fileSystem, IRuntimeFeature runtimeFeature, IEnvironment environment, IProcessHandler processHandler, ITestApplicationModuleInfo testApplicationModuleInfo) : ITestHostBuilder
```
MarcoRossignoli: do we need partial?

Evangelink: Nope some code refactoring messed up

src/Platform/Microsoft.Testing.Platform/Capabilities/TestFramework/IBannerMessageOwnerCapability.cs
```diff
@@ -0,0 +1,22 @@
...
+/// </summary>
+public interface IBannerMessageOwnerCapability : ITestFrameworkCapability
+{
+    /// <summary>
+    /// Process the banner message and return the message to be displayed.
+    /// </summary>
+    /// <returns>
+    /// The banner message to be displayed or <c>null</c> to use the default banner message.
+    /// </returns>
+    Task<string?> GetBannerMessageAsync();
```
Evangelink, line 24: Allow to return null so that users could have a way to fallback to default banner if some logic isn't successful on framework side.

src/Platform/Microsoft.Testing.Platform/Services/IPlatformInformation.cs
```diff
@@ -0,0 +1,73 @@
...
+
+    /// <summary>
+    /// Gets the platform's version.
+    /// </summary>
+    string? Version { get; }
+
+    /// <summary>
+    /// Gets the platform's commit hash.
+    /// </summary>
+    string? CommitHash { get; }
```
Evangelink, line 31: All nullable in case the info is not available - for example with Native AOT.

src/Platform/Microsoft.Testing.Platform/Services/IPlatformInformation.cs
```diff
@@ -0,0 +1,73 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System.Globalization;
+using System.Reflection;
+
+namespace Microsoft.Testing.Platform.Services;
+
+public interface IPlatformInformation
```
Evangelink, line 11: Shall we expose the other runtime info through this interface? Or as the logic is directly the one of runtime, we let framework authors doing the job?

MarcoRossignoli (approved):

MarcoRossignoli (dismissed):
It would be good add the list of public capabilities here https://github.com/microsoft/testfx/blob/main/docs/testingplatform/Index.md under the 4 point

src/Platform/Microsoft.Testing.Platform/.editorconfig
```diff
@@ -0,0 +1,3 @@
+[*.{cs,vb}]
+# TAEXP001: Type is for evaluation purposes only and is subject to change or removal in future updates. Suppress this diagnostic to proceed.
+dotnet_diagnostic.TAEXP001.severity = none
```
Evangelink:
```suggestion
# TPEXP: Type is for evaluation purposes only and is subject to change or removal in future updates. Suppress this diagnostic to proceed.
dotnet_diagnostic.TPEXP.severity = none
```

src/Platform/Microsoft.Testing.Platform/Capabilities/TestFramework/IBannerMessageOwnerCapability.cs
```diff
@@ -0,0 +1,25 @@
...
+namespace Microsoft.Testing.Platform.Capabilities.TestFramework;
+
+/// <summary>
+/// An optional test framework capability that allows the test framework to provide the banner message to the platform.
+/// If the message is null or if the capability is not present, the platform will use its default banner message.
+///
+/// This capability implementation allows to abstract away the various conditions that the test framework may need to consider
+/// to decide whether or not the banner message should be displayed.
+/// </summary>
+[Experimental("TAEXP001")]
```
Evangelink:
```suggestion
[Experimental("TPEXP", "https://aka.ms/testingplatform/diagnostics#{0}")]
```

src/Platform/Microsoft.Testing.Platform/Services/IPlatformInformation.cs
```diff
@@ -0,0 +1,75 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System.Diagnostics.CodeAnalysis;
+using System.Globalization;
+using System.Reflection;
+
+namespace Microsoft.Testing.Platform.Services;
+
+[Experimental("TAEXP001")]
```
Evangelink:
```suggestion
[Experimental("TPEXP", "https://aka.ms/testingplatform/diagnostics#{0}")]
```

PR #2923, opened 2024-05-23
Support dotnet test Help Option Through Pipes

mariam-abdulla:


src/Platform/Microsoft.Testing.Platform/Hosts/TestHostBuilder.cs
```diff
@@ -283,10 +286,76 @@ internal class TestHostBuilder(IFileSystem fileSystem, IRuntimeFeature runtimeFe
...
+
+                        if (commandLineOption.Arity == ArgumentArity.ExactlyOne)
+                        {
+                            arity = "ExactlyOne";
+                        }
+
+                        commandLineHelpOptions.Add(new CommandLineOptionMessage(
+                            commandLineOption.Name,
+                            commandLineOption.Description,
+                            arity,
```
nohwnd: The values seem the same why not commandLineOption.Arity.ToString() ? Or am I overlooking something?

src/Platform/Microsoft.Testing.Platform/CommandLine/PlaltformCommandLineProvider.cs
```diff
@@ -53,14 +54,15 @@ internal sealed class PlatformCommandLineProvider : ICommandLineOptionsProvider
         new(ExitOnProcessExitOptionKey, PlatformResources.PlatformCommandLineExitOnProcessExitOptionDescription, ArgumentArity.ExactlyOne, false, isBuiltIn: true),
 
         // Hidden options
-        new(ServerOptionKey, PlatformResources.PlatformCommandLineServerOptionDescription, ArgumentArity.Zero, true, isBuiltIn: true),
+        new(ServerOptionKey, PlatformResources.PlatformCommandLineServerOptionDescription, ArgumentArity.ZeroOrMore, true, isBuiltIn: true),
```
MarcoRossignoli: ZeroOrOne?

src/Platform/Microsoft.Testing.Platform/Hosts/TestHostBuilder.cs
mariam-abdulla: Because we can't have this value from arity property as we have:
public static readonly ArgumentArity ZeroOrMore = new(0, int.MaxValue);

src/Platform/Microsoft.Testing.Platform/CommandLine/PlaltformCommandLineProvider.cs
mariam-abdulla: It should be ZeroOrOne, I will fix it.

src/Platform/Microsoft.Testing.Platform/CommandLine/PlaltformCommandLineProvider.cs
```diff
@@ -33,6 +33,7 @@ internal sealed class PlatformCommandLineProvider : ICommandLineOptionsProvider
     public const string MinimumExpectedTestsOptionKey = "minimum-expected-tests";
     public const string TestHostControllerPIDOptionKey = "internal-testhostcontroller-pid";
     public const string ExitOnProcessExitOptionKey = "exit-on-process-exit";
+    public const string DotNetTestPipeOptionKey = "dotnet-test-pipe";
```
MarcoRossignoli: Can you pls refactor a bit this const list and split in 2 group to have server ones clear?

...
// Server mode
ServerOptionKey
PortOptionKey
ClientPortOptionKey
ClientHostOptionKey
DotNetTestPipeOptionKey

src/Platform/Microsoft.Testing.Platform/Hosts/TestHostBuilder.cs
```diff
@@ -283,10 +286,76 @@ internal class TestHostBuilder(IFileSystem fileSystem, IRuntimeFeature runtimeFe
             return toolsTestHost;
         }
 
+        NamedPipeClient? namedPipeClient = null;
```
MarcoRossignoli: Can you refactor all this logic in a method like `ConnectToTestHostProcessMonitorIfAvailableAsync`

something like `NamedPipeClient? ConnectToDotnetTestPipeIfAvailableAsync(...);`

```
// If we are in server mode and we the pipe name is provided
// then, we need to connect to the pipe server.
NamedPipeClient? dotnetTestPipeClient = ConnectToDotnetTestPipeIfAvailableAsync(...);
```

src/Platform/Microsoft.Testing.Platform/Hosts/TestHostBuilder.cs
```diff
@@ -283,10 +286,76 @@ internal class TestHostBuilder(IFileSystem fileSystem, IRuntimeFeature runtimeFe
...
+            namedPipeClient.RegisterSerializer<VoidResponse>(new VoidResponseSerializer());
+
+            await namedPipeClient.ConnectAsync(testApplicationCancellationTokenSource.CancellationToken);
+        }
+
         // If --help is invoked we return
         if (commandLineHandler.IsHelpInvoked())
         {
-            await commandLineHandler.PrintHelpAsync(toolsInformation.Tools);
+            if (namedPipeClient is not null)
```
MarcoRossignoli: Also here can we refactor in a method like `await SendCommandLineOptionsToDotnetTestPipeAsync(NamedPipeClient dotnetTestPipeClient);`

src/Platform/Microsoft.Testing.Platform/Hosts/TestHostBuilder.cs
```diff
@@ -283,10 +286,76 @@ internal class TestHostBuilder(IFileSystem fileSystem, IRuntimeFeature runtimeFe
...
+            {
+                List<CommandLineOptionMessage> commandLineHelpOptions = new();
+                foreach (ICommandLineOptionsProvider commandLineOptionProvider in commandLineHandler.CommandLineOptionsProviders)
+                {
+                    foreach (CommandLineOption commandLineOption in commandLineOptionProvider.GetCommandLineOptions())
+                    {
+                        string arity = string.Empty;
+
+                        if (commandLineOption.Arity == ArgumentArity.Zero)
```
MarcoRossignoli: can't we use a compact pattern matching for this one? 

https://learn.microsoft.com/en-us/dotnet/csharp/fundamentals/functional/pattern-matching

src/Platform/Microsoft.Testing.Platform/Hosts/TestHostBuilder.cs
```diff
@@ -283,10 +286,76 @@ internal class TestHostBuilder(IFileSystem fileSystem, IRuntimeFeature runtimeFe
...
+                        {
+                            arity = "Zero";
+                        }
+
+                        if (commandLineOption.Arity == ArgumentArity.ZeroOrOne)
+                        {
+                            arity = "ZeroOrOne";
```
MarcoRossignoli: wonder if we could use `Enum.GetName` https://learn.microsoft.com/en-us/dotnet/api/system.enum.getname?view=net-8.0 also if I think that this is faster, so we can keep it for now.

src/Platform/Microsoft.Testing.Platform/IPC/Serializers/BaseSerializer.cs
```diff
@@ -3,14 +3,15 @@
 
 #if NETCOREAPP
 using System.Buffers;
-#endif
-using System.Text;
```
MarcoRossignoli: nit: extra line

src/Platform/Microsoft.Testing.Platform/IPC/Serializers/BaseSerializer.cs
```diff
@@ -52,6 +53,29 @@ protected static void WriteString(Stream stream, string str)
         }
     }
 
+    protected static void WriteSize(Stream stream, string str)
```
MarcoRossignoli: nit: I don't know if we should call it WriteStringSize() to have clearer code during the read, up to you if you feel good with that I'm fine.

src/Platform/Microsoft.Testing.Platform/IPC/Serializers/BaseSerializer.cs
```diff
@@ -68,6 +92,22 @@ protected static void WriteLong(Stream stream, long value)
         stream.Write(bytes);
     }
 
+    protected static void WriteShort(Stream stream, short value)
+    {
+        Span<byte> bytes = stackalloc byte[sizeof(short)];
```
MarcoRossignoli: for coherency or we use numbers everywhere of `sizeof(...)` sizeof is expanded at compile time https://sharplab.io/#v2:EYLgHgbALANALiAlgGwD4AEBMBGAsAKHQGYACLEgYRIG8CT6zTEA7OEgWQAoBKGuhgegDsJAM6IAXgFMA9gDNOLONwDc/egF8CGoA===

pls substitute all places in this file where we have the number in favour of `sizeof(...)` that looks to me more explicit

src/Platform/Microsoft.Testing.Platform/IPC/Serializers/BaseSerializer.cs
```diff
@@ -68,6 +92,22 @@ protected static void WriteLong(Stream stream, long value)
...
+        BitConverter.TryWriteBytes(bytes, value);
```
MarcoRossignoli: The guard here is missing `ApplicationStateGuard`

src/Platform/Microsoft.Testing.Platform/IPC/Serializers/BaseSerializer.cs
```diff
@@ -120,11 +189,94 @@ protected static void WriteLong(Stream stream, long value)
...
+    protected static void WriteAtPosition(Stream stream, int value, long position)
+    {
+        var currentPosition = stream.Position;
+        SetPosition(stream, position);
+        WriteInt(stream, value);
+        SetPosition(stream, currentPosition);
+    }
+
+    private static int GetSize<T>()
+        where T : struct
```
MarcoRossignoli: you can use compact pattern matching also here and return directly the sizeof

src/Platform/Microsoft.Testing.Platform/IPC/Constants.cs
```diff
@@ -0,0 +1,19 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
```
MarcoRossignoli: I would rename the file in ObjectFieldId

src/Platform/Microsoft.Testing.Platform/IPC/Constants.cs
```diff
@@ -0,0 +1,19 @@
...
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+namespace Microsoft.Testing.Platform.IPC;
```
MarcoRossignoli: pls add a !!!HUGE WARNING MESSAGE!!! here where we explain that this file needs to be kept aligned to the SDK one and that the ids !!!CANNOT BE!!! absolutely exchange

Explain a bit the protocol logic/reason for future devs

src/Platform/Microsoft.Testing.Platform/IPC/Serializers/CommandLineOptionMessagesSerializer.cs
```diff
@@ -0,0 +1,157 @@
...
+        return (IsNull(commandLineOptionMessage.Name) ? 1 : 0) +
+            (IsNull(commandLineOptionMessage.Description) ? 1 : 0) +
+            (IsNull(commandLineOptionMessage.Arity) ? 1 : 0) +
+            (IsNull(commandLineOptionMessage.IsHidden) ? 1 : 0) +
+            (IsNull(commandLineOptionMessage.IsBuiltIn) ? 1 : 0);
+    }
+
+    private static bool IsNull<T>(T field)
+    {
+        return typeof(T) == typeof(string) ? field is not null : true;
```
MarcoRossignoli: think that this check can be removed and used directly inside the GetFieldCount.

 A reference type can be null, like a string, a value type cannot, like boolean if it's a nullable Null<T> it's a ref type anyway and the is null check will work correctly.

```
     private static int GetFieldCount(CommandLineOptionMessage commandLineOptionMessage)
    {
        return (commandLineOptionMessage.Name is not null ? 1 : 0) +
        ...            
    }
```    
this will also take advantage of the compiler that will help to fail if we're comparing a value type to null, a value type cannot be null so the check is not needed at all, better parf.
(Also the name IsNull looks like you wanted to return the opposite so if it's not null.)

src/Platform/Microsoft.Testing.Platform/IPC/Serializers/CommandLineOptionMessagesSerializer.cs
```diff
@@ -0,0 +1,157 @@
...
+        return commandLineOptionMessages;
+    }
+
+    public void Serialize(object objectToSerialize, Stream stream)
+    {
+        RoslynDebug.Assert(stream.CanSeek, "We expect a seekable stream.");
+
+        CommandLineOptionMessages commandLineOptionMessages = (CommandLineOptionMessages)objectToSerialize;
+
+        WriteShort(stream, (short)GetFieldCount(commandLineOptionMessages));
```
MarcoRossignoli: "short" it in the GetFieldCount if we expect that, also we should use an `ushort` that's wider.

src/Platform/Microsoft.Testing.Platform/IPC/Constants.cs
```diff
@@ -0,0 +1,19 @@
...
+
+internal class CommandLineOptionMessagesFields
```
MarcoRossignoli:
```suggestion
internal class CommandLineOptionMessagesFieldsId
```

src/Platform/Microsoft.Testing.Platform/IPC/Constants.cs
```diff
@@ -0,0 +1,19 @@
...
+{
+    internal const int ModuleName = 1;
+    internal const int CommandLineOptionMessageList = 2;
+}
+
+internal class CommandLineOptionMessageFields
```
MarcoRossignoli:
```suggestion
internal class CommandLineOptionMessageFieldsId
```

src/Platform/Microsoft.Testing.Platform/IPC/Serializers/CommandLineOptionMessagesSerializer.cs
```diff
@@ -0,0 +1,157 @@
...
+
+        WriteField(stream, CommandLineOptionMessagesFields.ModuleName, commandLineOptionMessages.ModuleName);
+        WriteCommandLineOptionMessagesPayload(stream, CommandLineOptionMessagesFields.CommandLineOptionMessageList, commandLineOptionMessages.CommandLineOptionMessageList);
+    }
+
+    private static void WriteCommandLineOptionMessagesPayload(Stream stream, short id, CommandLineOptionMessage[] commandLineOptionMessageList)
+    {
+        if (commandLineOptionMessageList is null)
```
MarcoRossignoli: null or empty? Or do you encode empty?

src/Platform/Microsoft.Testing.Platform/IPC/Serializers/CommandLineOptionMessagesSerializer.cs
```diff
@@ -0,0 +1,157 @@
...
+
+        CommandLineOptionMessages commandLineOptionMessages = (CommandLineOptionMessages)objectToSerialize;
+
+        WriteShort(stream, (short)GetFieldCount(commandLineOptionMessages));
+
+        WriteField(stream, CommandLineOptionMessagesFields.ModuleName, commandLineOptionMessages.ModuleName);
+        WriteCommandLineOptionMessagesPayload(stream, CommandLineOptionMessagesFields.CommandLineOptionMessageList, commandLineOptionMessages.CommandLineOptionMessageList);
+    }
+
+    private static void WriteCommandLineOptionMessagesPayload(Stream stream, short id, CommandLineOptionMessage[] commandLineOptionMessageList)
```
MarcoRossignoli: you don't need `short id` param...is needed only inside the call right?

src/Platform/Microsoft.Testing.Platform/IPC/Serializers/CommandLineOptionMessagesSerializer.cs
```diff
@@ -0,0 +1,157 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using Microsoft.Testing.Platform.IPC.Models;
+
+namespace Microsoft.Testing.Platform.IPC.Serializers;
+
+internal sealed class CommandLineOptionMessagesSerializer : BaseSerializer, INamedPipeSerializer
```
MarcoRossignoli, line 39: Can you pls above every serializer illustrate the binary shape? something like

```
/*

|---FieldCount----| 4bytes
|---ModuleName----| string id 1
....

*/
```

src/Platform/Microsoft.Testing.Platform/Hosts/TestHostBuilder.cs
MarcoRossignoli: Never mind me it's not an enum.

src/Platform/Microsoft.Testing.Platform/Hosts/TestHostBuilder.cs
```diff
@@ -283,10 +286,20 @@ internal class TestHostBuilder(IFileSystem fileSystem, IRuntimeFeature runtimeFe
             return toolsTestHost;
         }
 
+        NamedPipeClient? namedPipeClient = await ConnectToDotnetTestPipeIfAvailableAsync(commandLineHandler, testApplicationCancellationTokenSource);
```
MarcoRossignoli: we need to dispose this guy at some time...we need to find out the correct lifecycle for it

MarcoRossignoli (commented):
Aside that client dispose LGTM

src/Platform/Microsoft.Testing.Platform/CommandLine/InformativeCommandLineTestHost.cs
```diff
@@ -2,12 +2,31 @@
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
 using Microsoft.Testing.Platform.Hosts;
+using Microsoft.Testing.Platform.IPC;
 
 namespace Microsoft.Testing.Platform.CommandLine;
 
-internal sealed class InformativeCommandLineTestHost(int returnValue) : ITestHost
+internal sealed class InformativeCommandLineTestHost(int returnValue, NamedPipeClient? namedPipeClient = null) : ITestHost, IDisposable
```
MarcoRossignoli: can we have more specific name instead of `namedPipeClient`?

src/Platform/Microsoft.Testing.Platform/CommandLine/InformativeCommandLineTestHost.cs
```diff
@@ -2,12 +2,31 @@
...
+    , IAsyncDisposable
+#pragma warning restore SA1001 // Commas should be spaced correctly
+#endif
 {
     private readonly int _returnValue = returnValue;
+    private readonly NamedPipeClient? _namedPipeClient = namedPipeClient;
 
     public Task<int> RunAsync() => Task.FromResult(_returnValue);
+
+    public void Dispose() => _namedPipeClient?.Dispose();
```
MarcoRossignoli: include this in the else like above if would be good do not compile at all the non async version for core so we do only async disposal where we have support

MarcoRossignoli: never mind me keep as-is maybe we will have some context where we cannot go async...the disposal helper will take the correct decision https://github.com/microsoft/testfx/blob/main/src/Platform/Microsoft.Testing.Platform/Helpers/DisposeHelper.cs

MarcoRossignoli (approved):
LGTM when tests are green

MarcoRossignoli (approved):
HUGE!

src/Platform/Microsoft.Testing.Platform/Hosts/TestHostBuilder.cs
```diff
@@ -446,6 +459,77 @@ internal class TestHostBuilder(IFileSystem fileSystem, IRuntimeFeature runtimeFe
...
+                    GetArity(commandLineOption.Arity),
+                    commandLineOption.IsHidden,
+                    commandLineOption.IsBuiltIn));
+            }
+        }
+
+        await namedPipeClient.RequestReplyAsync<CommandLineOptionMessages, VoidResponse>(new CommandLineOptionMessages(Path.GetFileName(_testApplicationModuleInfo.GetCurrentTestApplicationFullPath()), commandLineHelpOptions.OrderBy(option => option.Name).ToArray()), cancellationToken);
+    }
+
+    private static string GetArity(ArgumentArity argumentArity)
```
Evangelink: Isn't `ToString` enough?

I would rename this method to be something like `GetArityName` or at least something suggesting it's giving a string representation of the enum.

src/Platform/Microsoft.Testing.Platform/Hosts/TestHostBuilder.cs
```diff
@@ -446,6 +459,77 @@ internal class TestHostBuilder(IFileSystem fileSystem, IRuntimeFeature runtimeFe
...
+    }
+
+    private static async Task<NamedPipeClient?> ConnectToDotnetTestPipeIfAvailableAsync(CommandLineHandler commandLineHandler, CTRLPlusCCancellationTokenSource cancellationTokenSource)
+    {
+        NamedPipeClient? namedPipeClient = null;
+
+        // If we are in server mode and the pipe name is provided
+        // then, we need to connect to the pipe server.
+        if (commandLineHandler.TryGetOptionArgumentList(PlatformCommandLineProvider.ServerOptionKey, out string[]? serverArgs) &&
+            serverArgs?.Length > 0 &&
```
Evangelink: Shall it be equals 1?

src/Platform/Microsoft.Testing.Platform/IPC/Models/CommandLineOptionMessages.cs
```diff
@@ -0,0 +1,8 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+namespace Microsoft.Testing.Platform.IPC.Models;
+
+internal sealed record class CommandLineOptionMessage(string Name, string Description, string Arity, bool IsHidden, bool IsBuiltIn) : IRequest;
+
+internal sealed record class CommandLineOptionMessages(string ModuleName, CommandLineOptionMessage[] CommandLineOptionMessageList) : IRequest;
```
Evangelink: Nit: class by default

```suggestion
internal sealed record CommandLineOptionMessage(string Name, string Description, string Arity, bool IsHidden, bool IsBuiltIn) : IRequest;

internal sealed record CommandLineOptionMessages(string ModuleName, CommandLineOptionMessage[] CommandLineOptionMessageList) : IRequest;
```

src/Platform/Microsoft.Testing.Platform/IPC/ObjectFieldIds.cs
```diff
@@ -0,0 +1,23 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+namespace Microsoft.Testing.Platform.IPC;
+
+// WARNING: Please note this file needs to be kept aligned with the one in the dotnet sdk.
+// The protocol follows the concept of optional properties.
+// The id is used to identify the property in the stream and it will be skipped if it's not recognized.
+// We can add new properties with new ids, but we CANNOT change the existing ids (to support backwards compatibility).
+internal class CommandLineOptionMessagesFieldsId
```
Evangelink: Static?

src/Platform/Microsoft.Testing.Platform/IPC/ObjectFieldIds.cs
```diff
@@ -0,0 +1,23 @@
...
+{
+    internal const int ModuleName = 1;
+    internal const int CommandLineOptionMessageList = 2;
+}
+
+internal class CommandLineOptionMessageFieldsId
```
Evangelink: Static?

src/Platform/Microsoft.Testing.Platform/IPC/Serializers/BaseSerializer.cs
```diff
@@ -120,11 +187,73 @@ protected static void WriteLong(Stream stream, long value)
...
+    protected static void WriteAtPosition(Stream stream, int value, long position)
+    {
+        long currentPosition = stream.Position;
+        SetPosition(stream, position);
+        WriteInt(stream, value);
+        SetPosition(stream, currentPosition);
+    }
+
+    private static int GetSize<T>()
+        where T : struct => typeof(T) == typeof(int)
```
Evangelink: I would switch instead of this chained ternary

src/Platform/Microsoft.Testing.Platform/Hosts/TestHostBuilder.cs
MarcoRossignoli: it was more to avoid client issue the day we start to use it, empty is a null in the protocol, because null is missing field.

MarcoRossignoli: @mariam-abdulla what you think? can we have always a string also if not recognized? I was wondering also if we should for now avoid this info at all, I don't know when if we will use it, anyway if we have not aligned(it should not happen we won't add new arity) we will have some issue. 

I think that we can remove this field for the good, code/perf for now.

src/Platform/Microsoft.Testing.Platform/Hosts/TestHostBuilder.cs
MarcoRossignoli: Today should be 1, anyway we were thinking to have in future a list of protocol to try if we add new ones and the client can understand more something like 

`--server protocol1 protocol2 protocol3`

having a way for the server to connect back with what it knows.

We can anyway in that case have always 1 with separator  `--server protocol1|protocol2|protocol3`

@mariam-abdulla if the Arity is `ZeroOrOne` (zero jsonrpc by default) the check is not needed here but I don't recall.

src/Platform/Microsoft.Testing.Platform/Hosts/TestHostBuilder.cs
mariam-abdulla: Yes, we can remove it as we are currently not using it.

src/Platform/Microsoft.Testing.Platform/Hosts/TestHostBuilder.cs
mariam-abdulla: Yes.

MarcoRossignoli (approved):

MarcoRossignoli (approved):

PR #2924, opened 2024-05-23
Add suppressor for CA1822

### LINKED ISSUE #1225: Create an analyzer suppression for CA1822
## Description

When using [CA1822](https://docs.microsoft.com/dotnet/fundamentals/code-analysis/quality-rules/ca1822), a lot of false positives are raised on test members.

We should provide a suppressor to avoid all these false-positives.

AB#2050316

Evangelink:
Fixes #1225


Evangelink:
As explained in the ticket, it seems `CA1822` was changed to not report on methods marked with attributes.

PR #2933, opened 2024-05-24
remove redundant null suppression and cache args in LogMessage

SimonCropp:


Evangelink (dismissed):

SimonCropp:
hmm looks like the suppression was needed for some frameworks. i will fix

Evangelink (approved):

PR #2944, opened 2024-05-24
use some pattern matching

SimonCropp:
IMO it is more readable


Evangelink (approved):
I know some people who will complain but I like the new syntax so good for me. Do you know if there is an analyzer associated we could use?

SimonCropp:
>  Do you know if there is an analyzer associated we could use?

i only know of how to flag them as errors in rider/resharper

PR #2960, opened 2024-05-25
remove redundant null check in RunTestCleanupMethod

SimonCropp:


MarcoRossignoli (approved):

MarcoRossignoli:
Looks like some tests are failing @Evangelink  are we strict with MSTest tests assertions or do we accept to update it if make sense?

Evangelink (approved):

PR #2962, opened 2024-05-25
elide simple asyncs

SimonCropp:


MarcoRossignoli:
Thanks Simon!

despite some perf improvement we prefer to keep the clarity in case of issues that could be confusing to debug https://github.com/davidfowl/AspNetCoreDiagnosticScenarios/blob/master/AsyncGuidance.md#prefer-asyncawait-over-directly-returning-task

PR #2963, opened 2024-05-25
move to file scoped namespaces

SimonCropp:


Evangelink (commented):
Hi @SimonCropp, 

Are you seeing issues when compiling the analyzers and code-fixes projects? I have disabled most of the rules we use in our repo on the `RoslynAnalyzerHelpers` folder to make it easier for me to diff if there is any change in these helpers that I simply copy-pasted from `dotnet/roslyn-analyzers`.

The namespace change doesn't bother me more than that but I am wondering if it would impact negatively our experience of reflecting changes. I am not sure... Will be thinking about it and let you know.

SimonCropp:
> Are you seeing issues when compiling the analyzers and code-fixes projects

nope

what are you seeing?

Evangelink:
I don't have issues, I was wondering the reasons for these changes. As I said, I am/was trying to make it easier to diff the files from roslyn-analyzers repo hence the difference in style and the set of disabled analyzers for these folders.

I don't know if that's really a good idea or not.

SimonCropp:
i was just making the approach to namespaces consistent. happy to close if u dont like the idea

SimonCropp:
but yeah. let kill it

PR #2966, opened 2024-05-26
use come collection expressions

SimonCropp:


Evangelink (dismissed):

SimonCropp:
@Evangelink i must admit i am not a fan of the formatting rules only failing in CI, and only failing for one file per build

Evangelink:
@SimonCropp is it not raising warnings locally? Because the only difference is that building from VS should give warnings and giving from CLI (build.cmd/build.sh) should give errors (warnAsErrors). This is setup to allow for more flexibility while prototyping in VS.

Evangelink (approved):

PR #2948, opened 2024-05-24
Fix assembly resolution error

Evangelink:
Fixes #2922.

PR #2969, opened 2024-05-26
simplify GetResolutionPaths

SimonCropp:
the resolutionPathsDictionary is only used for a key check. values are not used and there is only one key


Evangelink (changes requested):
Let's wait for the PR #2948 to be merged so we can double check (re-run CI) this is still ok when the resolver is fixed.

Evangelink:
Alright, let's see if CI is still passing.

Evangelink (approved):
Seems to be good

PR #2971, opened 2024-05-26
use a HashSet for cultures

SimonCropp:


Evangelink (approved):
Everyday I am discovering another wtf in this codebase :) Thanks @SimonCropp

PR #2974, opened 2024-05-27
use some StringSyntax Xml

SimonCropp:


Evangelink (commented):
You will probably need to enrich the polyfill file to add this new attribute.

SimonCropp:
thanks. will take a look at that

have u considered using some pre-packaged polyfills? eg https://github.com/SimonCropp/Polyfill ?

Evangelink:
> have u considered using some pre-packaged polyfills? eg [SimonCropp/Polyfill](https://github.com/SimonCropp/Polyfill) ?

Happy to set it up, but lets make sure it doesn't flow as dependency to our package.

Evangelink:
Hopefully with the rebase we should be good now.

Evangelink (approved):

PR #2985, opened 2024-05-27
remove CommandLineInfo

SimonCropp:
it adds no value. it takes a local copy of variables, that are already in scope, and exposes them to be used in the same scope


Evangelink (approved):
Yeah I guess that was refactored away and this part wasn't removed. Thanks!

Evangelink (dismissed):

Evangelink (approved):

PR #2986, opened 2024-05-27
move InitializeAsync check to an extension method

SimonCropp:


src/Platform/Microsoft.Testing.Platform/Extensions/IAsyncInitializableExtension.cs
```diff
@@ -13,3 +13,14 @@ public interface IAsyncInitializableExtension
     /// </summary>
     Task InitializeAsync();
 }
+
+internal static class InitializableExtension
+{
+    public static async Task TryInitialize(this object target)
```
Evangelink: Let's use the `Async` suffix.

Evangelink (commented):
Please use the `Async` suffix for the extension method.

Evangelink (approved):

PR #2994, opened 2024-05-27
Remove ITask

SimonCropp:
it is an abstraction that is not used


nohwnd:
Unhandled exception. System.TypeLoadException: Could not load type 'Microsoft.Testing.Platform.Helpers.ITask' from assembly 'Microsoft.Testing.Platform, Version=42.42.42.42, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a'.
   at Microsoft.Testing.Extensions.TrxReportExtensions.<>c__DisplayClass0_0.<AddTrxReportProvider>b__6(IServiceProvider serviceProvider)
   at Microsoft.Testing.Platform.Tools.ToolsManager.BuildAsync(IServiceProvider serviceProvider) in /_/src/Platform/Microsoft.Testing.Platform/Tools/ToolsManager.cs:line 24
   at Microsoft.Testing.Platform.Hosts.TestHostBuilder.BuildAsync(String[] args, ApplicationLoggingState loggingState, TestApplicationOptions testApplicationOptions, IUnhandledExceptionsHandler unhandledExceptionsHandler, DateTimeOffset createBuilderStart) in /_/src/Platform/Microsoft.Testing.Platform/Hosts/TestHostBuilder.cs:line 261
   at Microsoft.Testing.Platform.Builder.TestApplicationBuilder.BuildAsync() in /_/src/Platform/Microsoft.Testing.Platform/Builder/TestApplicationBuilder.cs:line 113
   at Program.<Main>$(String[] args) in /_/test/UnitTests/Microsoft.Testing.Platform.UnitTests/Program.cs:line 34
   
This is used by code that is extending the testing.platform and that is in our non-public repo.  I only see few usages, and I don't see any usages in Tests. 

Is this something that seemed like a good idea and did not catch on, or it is a pattern that will propagate to MSTest eventually?

Evangelink:
We had to take some shortcuts and so sadly we didn't test code as extensively as we wanted. The main idea of this helper is for us to be able to better test some race conditions as we could control the task.

I would prefer to keep it there for now and revisit dropping it if we didn't manage to invest into some of the tests in the upcoming period.

Evangelink (changes requested):
We had to take some shortcuts and so sadly we didn't test code as extensively as we wanted. The main idea of this helper is for us to be able to better test some race conditions as we could control the task.

I would prefer to keep it there for now and revisit dropping it if we didn't manage to invest into some of the tests in the upcoming period.

PR #3014, opened 2024-05-28
Use polyfill package

Evangelink:
cc @SimonCropp


src/Analyzers/MSTest.Analyzers/PublicAPI.Unshipped.txt
```diff
@@ -0,0 +1,114 @@
+﻿#nullable enable
```
Evangelink, line 1: Not sure why but the analyzer suddenly triggered issues on the project although the configuration to not raise when the file doesn't exist is setup.

src/Platform/Microsoft.Testing.Platform/GlobalSuppressions.cs
```diff
@@ -0,0 +1,8 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System.Diagnostics.CodeAnalysis;
+
+// We usually do not want to suppress issues through GlobalSuppressions file but in this case we have to do it because
+// we are not able to suppress the issue differently.
+[assembly: SuppressMessage("ApiDesign", "RS0030:Do not use banned APIs", Justification = "Source-generated file that cannot match analyzers requirements", Scope = "member", Target = "~M:System.Reflection.NullabilityInfoContext.CheckParameterMetadataType(System.Reflection.ParameterInfo,System.Reflection.NullabilityInfo)")]
```
Evangelink, line 8: We use a custom class for `Debug.Assert` (`RoslynDebug`) and `string.IsNullOrXXX` (`RoslynString`) that add annotation to the APIs. We are using banned apis to ensure they are used and it's triggering on the generated class although the generated code comment is setup.

PR #3028, opened 2024-05-29
remove set for toolname on CommandLineParseResult

SimonCropp:
since it looks like the intent is for it to be a readonly type


Evangelink (commented):
@MarcoRossignoli for review here

MarcoRossignoli (approved):

PR #3041, opened 2024-05-30
move to TimeProvider

SimonCropp:


Evangelink:
I haven't yet checked if the produced package is impacted but I am reluctant moving this forward. We want to ensure that the core platform is not having any dependency because this is one of the main issues of VSTest.

SimonCropp:
@Evangelink i can respect that. will close for now.

although note that Microsoft.Bcl.TimeProvider is only a polyfill package for older runtimes. it is not required in net8 and higher.

PR #3053, opened 2024-05-31
Improve display name when DataRow contains arrays

### LINKED ISSUE #1333: Improve display name when DataRow contains arrays
## Summary

When the `DataRowAttribute` contains multiple items including at least one array, then the display name will display the type instead of unwrapping the values.

## Example

```csharp
public void GetDisplayNameForMultipleArraysOfOneItem()
{
	// Arrange
	var dataRow = new DataRowAttribute(new[] { "a" }, new[] { "1" });
	var methodInfoMock = new Mock<MethodInfo>();
	methodInfoMock.SetupGet(x => x.Name).Returns("MyMethod");

	// Act
	var displayName = dataRow.GetDisplayName(methodInfoMock.Object, dataRow.Data);

	// Assert
	Verify(displayName == "MyMethod (System.String[],System.String[])"); // Expecting `MyMethod ([a],[1])`
}

public void GetDisplayNameForMultipleArraysOfMultipleItems()
{
	// Arrange
	var dataRow = new DataRowAttribute(new[] { "a", "b", "c" }, new[] { "1", "2", "3" });
	var methodInfoMock = new Mock<MethodInfo>();
	methodInfoMock.SetupGet(x => x.Name).Returns("MyMethod");

	// Act
	var displayName = dataRow.GetDisplayName(methodInfoMock.Object, dataRow.Data);

	// Assert
	Verify(displayName == "MyMethod (System.String[],System.String[])"); // Expecting `MyMethod ([a,b,c],[1,2,3])`
}
```

MichelZ:
Fixes #1333 

Improves the display name for DataRow tests for Arrays from:
`MyMethod (System.String[])`
to
`MyMethod ([a,b,c])`


src/TestFramework/TestFramework/Attributes/DataSource/DataRowAttribute.cs
```diff
@@ -84,6 +85,19 @@ public DataRowAttribute(params object?[]? data)
...
+    /// </summary>
+    private static string GetObjectString(IEnumerable<object?> enumerable) =>
+        string.Join(
+            ",",
+            enumerable.Select(x =>
+                x == null
+                ? null
+                : x.GetType().IsArray
+                    ? $"[{GetObjectString(((IEnumerable)x).Cast<object>())}]"
+                    : x.ToString()));
```
nohwnd: I guess it was the analyzers pushing you this way, but this code looks super dense. Can you please unwrap it a little bit? 

Do you think it would make sense to limit the number of recursive calls we do here by passing down a counter?

Evangelink (commented):
I wasn't able to review but just wanted to point out that this change would be a breaking change for people still using the display name for the tests id generation. Because it's obsolete it might not be as important but wanted to point it out so we don't merge too fast without thinking about consequences.

MichelZ:
> I wasn't able to review but just wanted to point out that this change would be a breaking change for people still using the display name for the tests id generation. Because it's obsolete it might not be as important but wanted to point it out so we don't merge too fast without thinking about consequences.

True. Do you think it would be helpful if the feature was an opt-in feature? I.e. a Flag on the Attribute for the Array Display Name?

src/TestFramework/TestFramework/Attributes/DataSource/DataRowAttribute.cs
MichelZ: Thanks. I have refactored it, hopefully it looks cleaner and easier to understand now

Could work with a counter, sure. Would this need to be configurable in the Attribute with a default limit so it's flexible?

nohwnd: Maybe it is not needed, this is for the data row only, so the data will be provided by a constant array expression directly to the attribute, and I cannot see a way to make that array recursive. Which is the only case I am worried about.

nohwnd:
@evangelink Good to know I was thinking the same, but there was no breaking label, or future milestone.  Can this be linked to id generation strategy? We have that as assembly level attribute (but imho also as a runsettings option). And if user is not using `Legacy` then we could generate the new display name? 

https://grep.app/search?q=idgenerationstrategy&filter[repo.pattern][0]=testfx

nohwnd:
The build is crashing on StackOverflow, I've added steps to install procdump so vstest can collect dumps. and pushed into your main Michel, for your next PR I would recommend creating a separate branch that is not main, so we don't block you if we take some time to merge this PR.

nohwnd:
Correction I cannot push to your main, because that is not a feature branch. So I created this PR, once that is merged you should be able to update and see why is the code is failing with Stack Overflow. https://github.com/microsoft/testfx/pull/3059

MichelZ:
Thanks, that should not be necessary just for me though, I fixed the StackOverflow issue now

nohwnd:
We will merge this next week at earliest, once evangelink and other folks are back from vacation and we can discuss how to avoid the breaking change. 

If you are okay with this PR hanging here and blocking your `main` branch then no action is needed. If you want to dig into more stuff, I would suggest checking out another branch, and making a new PR from it.

Evangelink:
We will expose a `protected` property on the `DataRowAttribute` that let subclasses know what's the selected `TestIdGenerationStrategy` and we will use this property to condition the change and avoid any breaking change.

@MichelZ I will push the change to your PR.

Evangelink (approved):
Thank you for your contribution @MichelZ :shipit:

Evangelink:
Flaky test, force merging.

PR #3059, opened 2024-06-03
Install procdump

nohwnd:
PR #3053  is failing with stack overflow and tests are running under VSTEST, but we don't install procdump so there is no dump collected.


install-procdump.ps1
MichelZ, line 1: You should move this to the /eng/ directory (or adjust the path in the pipeline)

nohwnd: thanks, I am doing too many things at the same time :/

Issue #3042, opened 2024-05-30
Use `ConcurrentDictionary` for attribute cache


nohwnd:
[Evangelink](https://github.com/Evangelink) [2 days ago](https://github.com/microsoft/testfx/pull/2839#discussion_r1617268740)
What about using ConcurrentDictionary instead?

Member
Author
@[nohwnd](https://github.com/nohwnd) nohwnd [2 days ago](https://github.com/microsoft/testfx/pull/2839#discussion_r1617613275)
I was thinking about it I wanted to measure the impact once this is merged, I saw time spent in locks in PerfView, but it also was not significantly faster without any locks. So please merge this approach first.

Member
Author
@[nohwnd](https://github.com/nohwnd) nohwnd [26 minutes ago](https://github.com/microsoft/testfx/pull/2839#discussion_r1619963687)
I was also thinking about writer only lock, but I am not sure if concurrent dictionary does not do that for me automatically already.

PR #3061, opened 2024-06-03
Use concurrent dictionary for attribute cache

nohwnd:
Use concurrent dictionary to store attribute cache.

Fix #3042

PR #3062, opened 2024-06-04
Use ConcurrentDictionary for attribute cache, and avoid func

nohwnd:
Fix https://github.com/microsoft/testfx/issues/3042

Better implementation of https://github.com/microsoft/testfx/pull/3061 so we can avoid creating reflection accessor, and so we avoid creating 700k Funcs (it is almost 100ms faster on 10k tests).


test/UnitTests/MSTestAdapter.UnitTests/Discovery/TypeEnumeratorTests.cs
```diff
@@ -98,6 +98,23 @@ public void GetTestsShouldReturnBaseTestMethodsInSameAssembly()
 
     public void GetTestsShouldReturnBaseTestMethodsFromAnotherAssemblyByDefault()
     {
+        string runSettingsXml =
+            """
+            <RunSettings>
+              <MSTestV2>
+                <CaptureTraceOutput>true</CaptureTraceOutput>
```
nohwnd, line 105: this test was failing when all tests run and passing when run by itself, it inherited the correct setup from somewhere previously. These are the exact settings it needs, copied from a test that is just below it.

MarcoRossignoli (approved):

PR #3063, opened 2024-06-04
Dont rebuild test properties

nohwnd:
This avoids re-creating test properties multiple times. 

It does not have a huge impact on run time, we save about 20ms.


nohwnd:
closing it did not bring much speedup.

Issue #2752, opened 2024-04-24
Public methods should be test methods


Evangelink:
Category: Design
Severity: Info
Enabled by default: false

Use cases: a public method is not maked with `[TestMethod]` or a derived attribute.

AB#2078797

engyebrahim:
@Evangelink should it work only if that public method is inside a `TestClass`?

PR #3065, opened 2024-06-04
MSTEST0029: Public methods should be test methods

engyebrahim:
fix: https://github.com/microsoft/testfx/issues/2752


test/UnitTests/MSTest.Analyzers.UnitTests/PublicMethodShouldBeTestMethodAnalyzerTests.cs
```diff
@@ -0,0 +1,92 @@
...
+            {
+                [DerivedTestMethod]
+                public void MyTestMethod()
+                {
+                }
+            }
+            """;
+        await VerifyCS.VerifyAnalyzerAsync(code);
+    }
+}
```
MarcoRossignoli, line 146: We could add a test for intheritance of test classes like

```cs
public class Base
{
    public void MyTestMethod()
    {
    }
}

[TestClass]
public class MyTestClass : Base
{
    [TestMethod]
    public void MyTestMethod()
    {
    }
}
```

What's the expected behavior here?

MarcoRossignoli (approved):

MarcoRossignoli (approved):

PR #3072, opened 2024-06-07
Update CODEOWNERS

nohwnd:
Remove Evangelink as owner of all code, so every PR does not require his approval, and we can automerge when he is not working (which is rare, but happens).

image


Evangelink (approved):
I didn't think of that, thanks!

Issue #3069, opened 2024-06-06
Remove Usings.cs from samples


cremor:
The global usings should not be needed any more with #2701

But they now need to be added to the "Below is the equivalent project configuration when not using MSTest.Sdk" sections in the csproj files as `<Using Include="Microsoft.VisualStudio.TestTools.UnitTesting" />`.

nohwnd:
Thanks, please see the linked PR, I was not sure if I am getting your request correctly.

PR #3073, opened 2024-06-07
Replace Usings.cs with <Using> in SDK samples

nohwnd:
Fix https://github.com/microsoft/testfx/issues/3069


cremor:
(Answering your comment in the issue) Yes, that's what I meant, thanks.

But there are more samples in the 3.4 branch, see https://github.com/microsoft/testfx/tree/rel/3.4/samples/public/DemoMSTestSdk
Those additional samples are also linked in the latest blog post. Is it by design that those are not in the main branch?

nohwnd:
I doubt it, @evangelink was finishing his blogpost, and probably forgot to backport to main.

cremor:
Then it would maybe be a good idea to target the 3.4 branch with this PR, change all samples and only then merge it back to main?

nohwnd:
retargeted, the subsequent PR will have to checkout the whole samples/public folder from here, rather than cherry-picking just the final comment. Or take all the relevant commits to samples folder.

PR #3092, opened 2024-06-12
Refactor command line validation to ease testing

Evangelink:
Refator `CommandLineHandler` to reduce responsibilities of the class. The new design is not yet optimal but allows to more easily do validation tests of command line providers without having to do a full end-to-end.

This is the first step of my plan to better validate our current providers as we keep discovering "stupid" mistakes around arity and missing validations.

This will also help to validate the test run parameter option by providing extensive test cases.


nohwnd:
Can you please add info to the PR why this is better approach than what was there before? It would make it easier to review if I know the motivation for the change :)

PR #3098, opened 2024-06-13
Discriminate MissingMethodException from hot reload

### LINKED ISSUE #3076: A test method using Activator.CreateInstance that raises an exception is not failing and is showing a "Not Run" status.
## Describe the bug

I'm testing a piece of code that uses Activator.CreateInstance. Although it fails, the test method status is showing as "Not Run".

## Steps To Reproduce

```
[TestMethod]
public void CleanData_AnonymousObject()
{
    var input = new
    {
        Field1 = "StringWith\0Null",
        Field2 = "NormalString",
    };
    var cleanedObject = Activator.CreateInstance(input.GetType());
}
```

## Expected behavior

The test fail and the status of the test is fail

## Actual behavior

The test status is Not Run

## Additional context

With XUnit the test fail because there is the following exception

Cannot dynamically create an instance of type '<>f__AnonymousType2`3[System.String,System.String,<>f__AnonymousType3`1[System.String]]'. Reason: No parameterless constructor defined.

Evangelink:
We added a special case when `MissingMethodException` is throw inside the execution of a test when we added the support for hot reload in MSTest. The handling wasn't precise enough leading to mixing up between the test method not being found and the test method content failing with `MissingMethodException`.

Fixes #3076


src/Adapter/MSTest.TestAdapter/Helpers/RuntimeContext.cs
```diff
@@ -0,0 +1,12 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+namespace Microsoft.VisualStudio.TestPlatform.MSTest.TestAdapter.Helpers;
+
+internal static class RuntimeContext
+{
+    // TODO: We should be using a capability from the runner instead of looking at environment variables.
+    public static bool IsInHotReloadContext()
```
nohwnd: Do we expect this value to change over the lifetime of the process? Can you cache it?

Evangelink: It shouldn't change so yeah I can cache it. Will do.

PR #3106, opened 2024-06-14
MSTest runner: allow overriding TestRunParameters

Evangelink:
When using MSTest runner it wasn't possible to override or provide `TestRunParameters` entries while with VSTest we could override any entry from runsettings. We have decided to not allow to change any runsettings entry but to provide first-class level support for test run parameters.

The syntax is `--test-parameter Key=Value` and we can provide multiple pairs.

### DIFF
--- a/test/IntegrationTests/MSTest.Acceptance.IntegrationTests/TestRunParametersTests.cs
+++ b/test/IntegrationTests/MSTest.Acceptance.IntegrationTests/TestRunParametersTests.cs
@@ -18,10 +18,21 @@ public sealed class TestRunParametersTests : AcceptanceTestBase
     }
 
     [ArgumentsProvider(nameof(TargetFrameworks.All), typeof(TargetFrameworks))]
-    public async Task CanProvideTestRunParameters(string tfm)
+    public async Task TestRunParameters_WhenProvidingMultipleArgumentsToTheOption(string tfm)
     {
         var testHost = TestHost.LocateFrom(_testAssetFixture.ProjectPath, TestAssetFixture.ProjectName, tfm);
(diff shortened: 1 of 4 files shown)


nohwnd (approved):

nohwnd (changes requested):
Wait no, didn't we say that multiple will be provided by repeating the parameters because we don't consume multiple values into a parameter?

test/IntegrationTests/MSTest.Acceptance.IntegrationTests/TestRunParametersTests.cs
```diff
@@ -0,0 +1,98 @@
...
+        : base(testExecutionContext)
+    {
+        _testAssetFixture = testAssetFixture;
+    }
+
+    [ArgumentsProvider(nameof(TargetFrameworks.All), typeof(TargetFrameworks))]
+    public async Task CanProvideTestRunParameters(string tfm)
+    {
+        var testHost = TestHost.LocateFrom(_testAssetFixture.ProjectPath, TestAssetFixture.ProjectName, tfm);
+        TestHostResult testHostResult = await testHost.ExecuteAsync("--settings my.runsettings --test-parameter MyParameter2=MyValue2 MyParameter3=MyValue3");
```
nohwnd, line 24: I am talking about this line, I thought that parameters don't consume arrays, and so the `MyParameter3=MyValue3` should trigger an error, even though the parameter is oneOrMany, because it requires `--test-parameter MyParameter2=MyValue2 --test-parameter MyParameter3=MyValue3` syntax.

Maybe I just remember it wrong?

Evangelink: Both syntaxes are supported (see https://learn.microsoft.com/dotnet/standard/commandline/syntax#argument-arity). I'll add a variant with the other syntax so we have both documented.

nohwnd (approved):

PR #2915, opened 2024-05-23
remove usages of DefinedTypes

SimonCropp:
DefinedTypes is essentially an inefficient wrapper around GetTypes().
```
   public virtual IEnumerable<TypeInfo> DefinedTypes
        {
            [RequiresUnreferencedCode("Types might be removed")]
            get
            {
                Type[] types = GetTypes();
                TypeInfo[] typeinfos = new TypeInfo[types.Length];
                for (int i = 0; i < types.Length; i++)
                {
                    TypeInfo typeinfo = types[i].GetTypeInfo();
                    if (typeinfo == null)
                        throw new NotSupportedException(SR.Format(SR.NotSupported_NoTypeInfo, types[i].FullName));

                    typeinfos[i] = typeinfo;
                }
                return typeinfos;
            }
        }
```

better to just use GetTypes() directly and not bother messing with TypeInfo

PR #3113, opened 2024-06-17
Fix some typos and use method group

Evangelink:
Extracted from #2915


Evangelink:
Do not merge once this is green, I need to amend author to mention @SimonCropp

PR #3114, opened 2024-06-17
Favor GetXXX() over DefinedXXX in internal test framework

Evangelink:
Extracted from #2915 (cc @SimonCropp)


nohwnd (approved):

Evangelink:
Ok so some change in this PR is causing the tests to hang forever

Issue #3102, opened 2024-06-14
CollectionAssert.AreEquivalent missing overload with `IEqualityComparer`


Evangelink:
## Summary

CollectionAssert.AreEquivalent missing overload with `IComparer`

stan-sz:
A use case could be asserting two collections of shuffled strings of various casing are equivalent.

PR #3130, opened 2024-06-19
Add overloads to CollectionAssert.AreEquivalent/AreNotEquivalent with IEqualityComparer

engyebrahim:
fix: https://github.com/microsoft/testfx/issues/3102
fix: https://github.com/microsoft/testfx/issues/3131


src/TestFramework/TestFramework/Assertions/CollectionAssert.cs
```diff
@@ -590,7 +590,177 @@ public static void IsNotSubsetOf([NotNull] ICollection? subset, [NotNull] IColle
...
+    /// the code under test.
+    /// </param>
+    /// <param name="comparer">
+    /// The compare implementation to use when comparing elements of the collection.
+    /// </param>
+    /// <exception cref="AssertFailedException">
+    /// <paramref name="expected"/> and <paramref name="actual"/> nullabilities don't match,
+    /// or if any element was found in one of the collections but not the other.
+    /// </exception>
+    public static void AreEquivalent(
```
Evangelink: Let's keep all `AreEquivalent` and `AreNotEquivalent` groupped together.

src/TestFramework/TestFramework/Assertions/CollectionAssert.cs
```diff
@@ -590,7 +590,177 @@ public static void IsNotSubsetOf([NotNull] ICollection? subset, [NotNull] IColle
...
+    /// <param name="actual">
+    /// The second collection to compare. This is the collection produced by
+    /// the code under test.
+    /// </param>
+    /// <exception cref="AssertFailedException">
+    /// <paramref name="expected"/> and <paramref name="actual"/> nullabilities don't match,
+    /// or if collections contain the same elements, including the same number of duplicate
+    /// occurrences of each element.
+    /// </exception>
+    public static void AreNotEquivalent(
```
Evangelink: I suppose you are adding these APIs for parity? Could we do it in a separate PR please?

src/TestFramework/TestFramework/Assertions/CollectionAssert.cs
```diff
@@ -1444,5 +1625,12 @@ private class ObjectComparer : IComparer
         int IComparer.Compare(object? x, object? y) => Equals(x, y) ? 0 : -1;
     }
 
+    private class ObjectEqualityComparer : IEqualityComparer<object>
```
Evangelink: It's probably better to use `EqualityComparer.Default` instead of this custom class.

src/TestFramework/TestFramework/Assertions/CollectionAssert.cs
```diff
@@ -590,7 +590,108 @@ public static void IsNotSubsetOf([NotNull] ICollection? subset, [NotNull] IColle
...
+    /// <param name="comparer">
+    /// The compare implementation to use when comparing elements of the collection.
+    /// </param>
+    /// <exception cref="AssertFailedException">
+    /// <paramref name="expected"/> and <paramref name="actual"/> nullabilities don't match,
+    /// or if any element was found in one of the collections but not the other.
+    /// </exception>
+    public static void AreEquivalent<T>(
+        [NotNullIfNotNull(nameof(actual))] ICollection? expected, [NotNullIfNotNull(nameof(expected))] ICollection? actual, [NotNull] IEqualityComparer<T>? comparer)
+        where T : notnull
```
Evangelink: Please update all methods to remove the `where T : notnull` because it prevents nullable type for no good reason.  And add some test with a nullable type please.

src/TestFramework/TestFramework/Assertions/CollectionAssert.cs
```diff
@@ -590,7 +590,108 @@ public static void IsNotSubsetOf([NotNull] ICollection? subset, [NotNull] IColle
...
+    /// </param>
+    /// <param name="comparer">
+    /// The compare implementation to use when comparing elements of the collection.
+    /// </param>
+    /// <exception cref="AssertFailedException">
+    /// <paramref name="expected"/> and <paramref name="actual"/> nullabilities don't match,
+    /// or if any element was found in one of the collections but not the other.
+    /// </exception>
+    public static void AreEquivalent<T>(
+        [NotNullIfNotNull(nameof(actual))] ICollection? expected, [NotNullIfNotNull(nameof(expected))] ICollection? actual, [NotNull] IEqualityComparer<T>? comparer)
```
Evangelink: As we introduce generics, I would be tempted to change to `IEnumerable<T>` to allow more types.

src/TestFramework/TestFramework/Assertions/CollectionAssert.cs
```diff
@@ -1280,6 +1487,7 @@ internal static bool IsSubsetOfHelper(ICollection subset, ICollection superset)
         return true;
     }
 
+#pragma warning disable CS8714
```
Evangelink, line 1490: Can't we have narrower scope pragma disable?

src/TestFramework/TestFramework/Assertions/CollectionAssert.cs
```diff
@@ -590,7 +590,105 @@ public static void IsNotSubsetOf([NotNull] ICollection? subset, [NotNull] IColle
...
+    /// </param>
+    /// <param name="comparer">
+    /// The compare implementation to use when comparing elements of the collection.
+    /// </param>
+    /// <exception cref="AssertFailedException">
+    /// <paramref name="expected"/> and <paramref name="actual"/> nullabilities don't match,
+    /// or if any element was found in one of the collections but not the other.
+    /// </exception>
+    public static void AreEquivalent<T>(
+        [NotNullIfNotNull(nameof(actual))] IEnumerable<T>? expected, [NotNullIfNotNull(nameof(expected))] IEnumerable<T>? actual, [NotNull] IEqualityComparer<T>? comparer)
```
Evangelink: You probably need `IEnumerable<T?>?` in the various places. As far as I can tell, you don't have a test that would contain some null items, it'd be good to have one.

src/TestFramework/TestFramework/Assertions/CollectionAssert.cs
```diff
@@ -603,27 +701,32 @@ public static void IsNotSubsetOf([NotNull] ICollection? subset, [NotNull] IColle
...
         // If both collections are empty, they are equivalent.
-        if (expected.Count == 0)
+        if (!expected.Any())
         {
             return;
         }
 
         // Search for a mismatched element.
-        if (FindMismatchedElement(expected, actual, out int expectedCount, out int actualCount, out object? mismatchedElement))
+        if (FindMismatchedElement<T>(expected, actual, comparer, out int expectedCount, out int actualCount, out object? mismatchedElement))
```
Evangelink: You probably don't need the explicit `T`

```suggestion
        if (FindMismatchedElement(expected, actual, comparer, out int expectedCount, out int actualCount, out object? mismatchedElement))
```

src/TestFramework/TestFramework/Assertions/CollectionAssert.cs
```diff
@@ -1255,8 +1462,8 @@ internal static bool IsSubsetOfHelper(ICollection subset, ICollection superset)
         // $ CONSIDER: comparison, which should result in ~n*log(n) + m*log(m) + n.
 
         // Count the occurrences of each object in both collections.
-        Dictionary<object, int> subsetElements = GetElementCounts(subset, out int subsetNulls);
-        Dictionary<object, int> supersetElements = GetElementCounts(superset, out int supersetNulls);
+        Dictionary<object, int> subsetElements = GetElementCounts(subset.Cast<object>(), EqualityComparer<object>.Default, out int subsetNulls);
```
Evangelink, line 1465: Could we have the 2 collections as `IEnumerable<T>` on the signature or is it having too many impacts?

src/TestFramework/TestFramework/Assertions/CollectionAssert.cs
engyebrahim: no, i need it for both functions

src/TestFramework/TestFramework/Assertions/CollectionAssert.cs
engyebrahim: this function is used only inside an API which is not generic what is the point of making it generic?

Evangelink: None :)

Evangelink (dismissed):

Evangelink (approved):

PR #3150, opened 2024-06-21
remove un-used variables

SimonCropp:


Evangelink (approved):
Thank you @SimonCropp :shipit:

PR #3151, opened 2024-06-21
remove redundant null checks

SimonCropp:


SimonCropp:
i would argue TryGetStackTraceInformationReturnsNullIfExceptionIsNull is an invalid scenario

Evangelink (approved):

Evangelink:
> i would argue TryGetStackTraceInformationReturnsNullIfExceptionIsNull is an invalid scenario

`StackTrace` property is defined as nullable https://learn.microsoft.com/dotnet/api/system.exception.stacktrace.

PR #3153, opened 2024-06-21
remove redundant null suppressions

SimonCropp:


Evangelink (commented):
Looks like there are some false positives in Resharper. Could you please update the entries failing?

SimonCropp:
ahhh. that would be because sometimes it get confused between the nullability of different runtimes

Evangelink (approved):

PR #3157, opened 2024-06-21
remove use of GetTypeInfo().DeclaredProperties

SimonCropp:
functionally equivalent. but less allocation

### DIFF
--- a/src/Adapter/MSTest.TestAdapter/Discovery/TypeValidator.cs
+++ b/src/Adapter/MSTest.TestAdapter/Discovery/TypeValidator.cs
@@ -106,7 +106,7 @@ internal class TypeValidator
     {
         DebugEx.Assert(type != null, "HasCorrectTestContextSignature type is null");
 
-        IEnumerable<PropertyInfo> propertyInfoEnumerable = type.GetTypeInfo().DeclaredProperties;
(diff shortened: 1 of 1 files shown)


src/Adapter/MSTest.TestAdapter/Discovery/TypeValidator.cs
```diff
@@ -106,7 +106,7 @@ internal static bool HasCorrectTestContextSignature(Type type)
     {
         DebugEx.Assert(type != null, "HasCorrectTestContextSignature type is null");
 
-        IEnumerable<PropertyInfo> propertyInfoEnumerable = type.GetTypeInfo().DeclaredProperties;
+        IEnumerable<PropertyInfo> propertyInfoEnumerable = type.GetProperties(BindingFlags.DeclaredOnly);
```
Evangelink, line 109: This is not equivalent. According to documentation `DeclaredProperties` returns properties from hierarchy while `BindingFlags.DeclaredOnly` returns only properties of current type.

I would need to make some tests to see if `static` are considerd by `DeclaredProperties` or not to see if that's equivalent to `GetProperties()`.

Evangelink (changes requested):
Test `IsValidTestClassShouldReturnFalseForTestClassesWithInvalidTestContextSignature` is failing as it's showing a functional difference between the 2 calls.

PR #3158, opened 2024-06-21
Remove typeinfo in adapter to test platform

SimonCropp:


SimonCropp:
let skill this. i will try and keep future PRs smaller

Issue #2688, opened 2024-04-10
Introduce `[STATestMethod]` and `[STATestClass]`


Evangelink:
Instead of forcing all tests to be on STA, we could offer the ability to only create a custom thread for the few methods that needs to be STA. This should provide a performance boost.

AB#2078830

Evangelink:
Given some requests we have received, I am wondering if we should introduce either a different attribute or have a "property" on the various places where we want to add this support. 

My main concern with `[STATestMethod]` and `[STATestClass]` is that there will be no way to enable STA threading for Assembly/Class Initialize/Cleanup methods outside of turning the full application to STA.

I would be tempted to vote for a "property" rather than a new attribute because the new attribute will also put us on the corner of having users trying to put the attribute in locations not expected/supported.

As for the design, I think we should support the following scenarios:

1. Full test cycle (ctor, test initialize, test method, test cleanup, dispose) are STA
2. All test cycles from current class are STA
3. Pair AssemblyInitialize/Cleanup and/or ClassInitialize/Cleanup are STA

I don't think there is a strong need to support something more fine-grained (like only ctor, only test method but not init/cleanups...).

nohwnd:
`[STATestClass]` seems like a good enough granularity, given that we don't have the ability to run multiple assembly level setups. It also seems reasonably easy for users to simulate assembly level setup with a lock and static. And also to split tests into needs STA classes and does not need STA classes.

maybe the requirements are different, but no matter how I look at the scenarios I can come up with in my head, STATestClass is practical for all of them, as long as there is also a way to run everything in assembly with STA mode (which there is because it motivates this feature request).

Evangelink:
That's a pretty clever suggestion! I think this is the right next increment, we can adjust based on users/customers feedback.

@nohwnd We consider `[ClassInitialize]` and `[ClassCleanup]` to be part of the STA thread execution, right?

Nirmal4G:
Why not `[TestClass(Threading = ApartmentState.STA)]` or `[TestMethod(STA = true)]` or `[TestMethod, STAThread]`?

These gives better visibility and resuse of existing knowledge with types than creating an entirety new attributes just for specifying threading model. Also, there's an existing type in the BCL for threading model that we can use it here.

BTW, I vote for the middle one. It's simple. The existing attributes can be declared only on methods. The first two allow parameterization that'll allow changing at Runtime but the enum one is too longer to type.

Evangelink:
[...]

PR #3188, opened 2024-06-27
Introduce `[STATestClass]`

engyebrahim:
fix: https://github.com/microsoft/testfx/issues/2688


src/Adapter/MSTest.TestAdapter/Execution/TestMethodRunner.cs
```diff
@@ -72,7 +75,7 @@ public TestMethodRunner(TestMethodInfo testMethodInfo, TestMethod testMethod, IT
     /// Executes a test.
     /// </summary>
     /// <returns>The test results.</returns>
-    internal UnitTestResult[] Execute()
+    internal UnitTestResult[] Execute(ClassCleanupManager? _classCleanupManager, TypeCache _typeCache)
```
Evangelink, line 142: This method is getting super complex. I would suggest that you extract the content into sub-methods.

src/Adapter/MSTest.TestAdapter/Execution/TestMethodRunner.cs
```diff
@@ -81,36 +84,34 @@ internal UnitTestResult[] Execute()
...
         }
         catch (TestFailedException ex)
         {
+            runFailed = true;
             result = [new UnitTestResult(ex)];
         }
         catch (Exception ex)
         {
+            runFailed = true;
             if (result == null || result.Length == 0)
```
Evangelink: This logic is now obsolete. Result is never set from this path.

src/Adapter/MSTest.TestAdapter/Execution/TestMethodRunner.cs
```diff
@@ -128,16 +129,187 @@ internal UnitTestResult[] Execute()
             result[result.Length - 1] = newResult;
 #pragma warning restore IDE0056 // Use index operator
         }
+
+        if (_testMethodInfo.Parent.ClassAttribute is STATestClassAttribute
```
Evangelink: This is changing the logic of the method. Before, when a problem was thrown from assembly init, the rest of the code wasn't called while now it is.

Evangelink (commented):
I'll wait for the refactoring to be done before doing review because it will change drastrically the look of the code.

src/Adapter/MSTest.TestAdapter/Execution/TestMethodRunner.cs
Evangelink: You may want to check that it inherits from `STATestClass` to leave the opportunity to users to extend the behavior (this will need a specific test).

engyebrahim: no,i have a boolean to check if assinit failed it won't enter the code only the cleanups as it was out this func

src/Adapter/MSTest.TestAdapter/Execution/TestMethodRunner.cs
engyebrahim: no it comes here if assinit throws

src/Adapter/MSTest.TestAdapter/Execution/TestMethodRunner.cs
```diff
@@ -81,63 +148,159 @@ internal UnitTestResult[] Execute()
...
+                initializationTrace = logListener.GetAndClearDebugTrace();
+                initializationErrorLogs = logListener.GetAndClearStandardError();
+                initializationTestContextMessages = _testContext.GetAndClearDiagnosticMessages();
+            }
+        }
+        catch (Exception ex)
+        {
+            runFailed = true;
+            RunCatch(ex, ref result);
+            RunFinally(ref result, ref initializationLogs, ref initializationTrace, ref initializationErrorLogs, ref initializationTestContextMessages);
```
Evangelink, line 175: The semantic changed, this will now be executed only in case of failure and not for success cases.

src/Adapter/MSTest.TestAdapter/Execution/TestMethodRunner.cs
```diff
@@ -81,63 +148,159 @@ internal UnitTestResult[] Execute()
...
+            try
+            {
+                var threadTask = Task.Run(entryPointThread.Join, _testContext.Context.CancellationTokenSource.Token);
+                threadTask.Wait(_testContext.Context.CancellationTokenSource.Token);
+            }
+            catch (Exception ex)
+            {
+                runFailed = true;
+                RunCatch(ex, ref result);
+                RunFinally(ref result, ref initializationLogs, ref initializationTrace, ref initializationErrorLogs, ref initializationTestContextMessages);
```
Evangelink, line 201: The semantic changed, this will now be executed only in case of failure and not for success cases.

src/Adapter/MSTest.TestAdapter/Execution/TestMethodRunner.cs
```diff
@@ -81,63 +148,159 @@ internal UnitTestResult[] Execute()
...
+        }
+        catch (Exception ex)
+        {
+            runFailed = true;
+            RunCatch(ex, ref result);
+            RunFinally(ref result, ref initializationLogs, ref initializationTrace, ref initializationErrorLogs, ref initializationTestContextMessages);
+        }
+
+        Type attributeType = _testMethodInfo.Parent.ClassAttribute.GetType();
+        if ((_testMethodInfo.Parent.ClassAttribute is STATestClassAttribute || attributeType.IsSubclassOf(typeof(STATestClassAttribute)))
```
Evangelink, line 179: As a rule of thumb, always try to have simpler checks done first (e.g. `runFailed` is less expensive than others so should be checked first).

src/Adapter/MSTest.TestAdapter/Execution/TestMethodRunner.cs
```diff
@@ -81,63 +148,159 @@ internal UnitTestResult[] Execute()
 
         UnitTestResult[]? result = null;
 
+        bool shouldRunClassCleanup = false;
```
Evangelink, line 151: Prefer initialization of variables as close as possible to their usage.

Evangelink (commented):
First level review. Don't start the changes now, I need to spend more time on the logic.

src/Adapter/MSTest.TestAdapter/Execution/TestMethodRunner.cs
engyebrahim: in the success case i'm calling finally function again

src/Adapter/MSTest.TestAdapter/Execution/TestMethodRunner.cs
engyebrahim: in the success case i'm calling finally function again

fforjan:
@Evangelink by introduce a new subclass, does it mean any code analyzer will have issue because it will not find a TestClass but a  STATestClass and need to be made aware of it ?

Evangelink:
> @Evangelink by introduce a new subclass, does it mean any code analyzer will have issue because it will not find a TestClass but a STATestClass and need to be made aware of it ?

For our analyzers we are checking derived from `TestClassAttribute` so they would remain functional. For third-party analyzers for MSTest, I cannot comment without knowing their logic.

Issue #2862, opened 2024-05-17
MSTESTXXXX: Type contaning `[TestMethod]` must be a `[TestClass]`


nohwnd:
## Summary

Add **usage** analyzer that ensures that a non abstract class with `[TestMethod]` is marked with `[TestClass]`.

## Background and Motivation

Discoverer looks for classes that have attribute derived from `[TestClass]`, and does not look through inheritance chain (parent classes).

https://github.com/microsoft/testfx/blob/main/src/Adapter/MSTest.TestAdapter/Discovery/TypeEnumerator.cs#L55
https://github.com/microsoft/testfx/blob/main/src/Adapter/MSTest.TestAdapter/Discovery/TypeValidator.cs#L54

Not marking a class as `[TestClass]` is an easy way to lose tests, and currently there is no build time, or runtime warning. 

## Proposed Feature

Ensure that when a method has `[TestMethod]` attribute, it is defined by a class that is marked as `[TestClass]`. Skip this rule for abstract classes, because they are used to inherit tests to concrete classes.

q: do we see those inherited methods on a type? we should ensure that type is testClass as well (most likely, unless the user does some complex inheritance chain).

Examples: 

```
namespace TestProject97
{


namespace TestProject97
{
    // SCENARIO1
    // No test class, this loses tests. Would show WARNING.
    // User can either add [TestClass] or mark the class as abstract, if they only intend to inherit the tests.

    public class WithTestMethods_WithoutTestClass
    {
        [TestMethod]
        public void TestMethod1()
        {
        }

        [TestMethod]
        public void TestMethod2()
        {
        }
    }

    // SCENARIO2
    // This test class will inherit parent tests. 2 tests will run from here.
    [TestClass]
    public class WithoutTestMethods_WithTestClass : WithTestMethods_WithoutTestClass
    {

    }

    // This test class will inherit parent tests. Should it show WARNING? IMHO yes.
    public class WithoutTestMethods_WithoutTestClass : WithTestMethods_WithoutTestClass
    {

    }

    // This test class has no tests,
    // it is in examples only to prove that TestClass
    // is not looked up with inheritance (by checking parent class).
    [TestClass]
    public class ParentWithTestTestClass
    {

    }

    // SCENARIO3:
    // Those tests won't run, WARNING should show, the parent class [TestClass] won't be respected, because we don't search
    // inheritance.
    public class ChildWithoutTestClass : ParentWithTestTestClass
    {
        [TestMethod]
        public void TestMethod1()
[...]

PR #3199, opened 2024-07-01
MSTEST0030: Type contaning [TestMethod] must be a [TestClass]

engyebrahim:
fix: https://github.com/microsoft/testfx/issues/2862


src/Analyzers/MSTest.Analyzers/xlf/Resources.ja.xlf
```diff
@@ -520,6 +520,21 @@
         <target state="translated">テスト メソッドを無視することはできません</target>
         <note />
       </trans-unit>
+      <trans-unit id="TypeContaningTestMethodShouldBeATestClassDescription">
```
nohwnd: Conta **i** ning. 

This typo is in many places. Both in text and in type name.

test/UnitTests/MSTest.Analyzers.UnitTests/TypeContainingTestMethodShouldBeATestClassAnalyzer.cs
```diff
@@ -0,0 +1,148 @@
...
+                [DerivedTestMethod]
+                public void MyTestMethod()
+                {
+                }
+            }
+            """;
+
+        await VerifyCS.VerifyAnalyzerAsync(code);
+    }
+}
```
nohwnd, line 165: Worth adding another test case for the abstract class not having a diagnostic?

nohwnd (commented):
Great job, this analyzer will be so helpful :)

src/Analyzers/MSTest.Analyzers/AnalyzerReleases.Unshipped.md
```diff
@@ -7,3 +7,4 @@ Rule ID | Category | Severity | Notes
 --------|----------|----------|-------
 MSTEST0026 | Usage | Info | AssertionArgsShouldAvoidConditionalAccessRuleId, [Documentation](https://learn.microsoft.com/dotnet/core/testing/mstest-analyzers/mstest0026)
 MSTEST0029 | Design | Disabled | PublicMethodShouldBeTestMethodAnalyzer, [Documentation](https://learn.microsoft.com/dotnet/core/testing/mstest-analyzers/mstest0029)
+MSTEST0030 | Usage |Info | TypeContainingTestMethodShouldBeATestClassAnalyzer, [Documentation](https://learn.microsoft.com/dotnet/core/testing/mstest-analyzers/mstest0030)
```
Evangelink: Nit:

```suggestion
MSTEST0030 | Usage | Info | TypeContainingTestMethodShouldBeATestClassAnalyzer, [Documentation](https://learn.microsoft.com/dotnet/core/testing/mstest-analyzers/mstest0030)
```

src/Analyzers/MSTest.Analyzers/TypeContainingTestMethodShouldBeATestClassAnalyzer.cs
```diff
@@ -0,0 +1,104 @@
...
+        context.EnableConcurrentExecution();
+
+        context.RegisterCompilationStartAction(context =>
+        {
+            if (context.Compilation.TryGetOrCreateTypeByMetadataName(WellKnownTypeNames.MicrosoftVisualStudioTestToolsUnitTestingTestMethodAttribute, out INamedTypeSymbol? testMethodAttributeSymbol)
+                && context.Compilation.TryGetOrCreateTypeByMetadataName(WellKnownTypeNames.MicrosoftVisualStudioTestToolsUnitTestingTestClassAttribute, out INamedTypeSymbol? testClassAttributeSymbol))
+            {
+                INamedTypeSymbol? taskSymbol = context.Compilation.GetOrCreateTypeByMetadataName(WellKnownTypeNames.SystemThreadingTasksTask);
+                INamedTypeSymbol? valueTaskSymbol = context.Compilation.GetOrCreateTypeByMetadataName(WellKnownTypeNames.SystemThreadingTasksValueTask);
+                bool canDiscoverInternals = context.Compilation.CanDiscoverInternals();
```
Evangelink: All these are unused so you could remove them.

src/Analyzers/MSTest.Analyzers/TypeContainingTestMethodShouldBeATestClassAnalyzer.cs
```diff
@@ -0,0 +1,104 @@
...
+        if (isTestClass)
+        {
+            return;
+        }
+
+        bool hasTestMethod = false;
+        INamedTypeSymbol? currentType = namedTypeSymbol;
+        do
+        {
+            foreach (ISymbol classMember in currentType.GetMembers())
```
Evangelink, line 79: Only methods can be marked with `[TestMethod]` so you should filter out to ensure you don't call `GetAttributes` on all members.

src/Analyzers/MSTest.Analyzers/TypeContainingTestMethodShouldBeATestClassAnalyzer.cs
```diff
@@ -0,0 +1,104 @@
...
+            {
+                foreach (AttributeData attribute in classMember.GetAttributes())
+                {
+                    if (attribute.AttributeClass.Inherits(testMethodAttributeSymbol))
+                    {
+                        hasTestMethod = true;
```
Evangelink, line 90: There is no need to continue the 2 foreach loops when you have found the attribute.

src/Analyzers/MSTest.Analyzers/Resources.resx
```diff
@@ -436,4 +436,13 @@
   <data name="UseParallelizeAttributeAnalyzerTitle" xml:space="preserve">
     <value>Explicitly enable or disable tests parallelization</value>
   </data>
+  <data name="TypeContaningTestMethodShouldBeATestClassTitle" xml:space="preserve">
+    <value>Type containing '[TestMethod]' must be a '[TestClass]'</value>
```
Evangelink:
```suggestion
    <value>Type containing '[TestMethod]' must be marked with '[TestClass]'</value>
```

Evangelink: I am wondering if we should be not using the attribute name explicitely as we allow sub-classing, something like:

`Types containing test methods should be test classes`.

src/Analyzers/MSTest.Analyzers/Resources.resx
```diff
@@ -436,4 +436,13 @@
...
+  </data>
+  <data name="TypeContaningTestMethodShouldBeATestClassDescription" xml:space="preserve">
+    <value>Type contaning '[TestMethod]' must be a '[TestClass]'.</value>
```
Evangelink: Typo:

```suggestion
    <value>Type containing '[TestMethod]' must be a '[TestClass]'.</value>
```

Evangelink: I would add some details here to help users understand why.

src/Analyzers/MSTest.Analyzers/Resources.resx
```diff
@@ -436,4 +436,13 @@
...
+  </data>
+  <data name="TypeContaningTestMethodShouldBeATestClassMessageFormat" xml:space="preserve">
+    <value>Type '{0}' contaning '[TestMethod]' must be a '[TestClass]'</value>
```
Evangelink:
```suggestion
    <value>Class '{0}' contains test methods and should be marked with '[TestClass]'</value>
```

test/UnitTests/MSTest.Analyzers.UnitTests/TypeContainingTestMethodShouldBeATestClassAnalyzer.cs
Evangelink: Yes please, this is one of the condition.

Evangelink (commented):
@nohwnd Can you please review the messages

src/Analyzers/MSTest.Analyzers/Resources.resx
nohwnd: Imho using the name `[TestClass]` is clearer for 90% of use cases. So I would keep it.

src/Analyzers/MSTest.Analyzers/Resources.resx
nohwnd: maybe: ", otherwise the test method will be silently ignored." ?

src/Analyzers/MSTest.Analyzers/Resources.resx
Evangelink: Good for me. @engyebrahim please only apply the suggested change.

test/UnitTests/MSTest.Analyzers.UnitTests/TypeContainingTestMethodShouldBeATestClassAnalyzer.cs
```diff
@@ -0,0 +1,161 @@
...
+                public void MyTestMethod()
+                {
+                }
+            }
+            """;
+
+        await VerifyCS.VerifyAnalyzerAsync(code);
+    }
+
+    public async Task WhenAbstractClass_DoesNotHaveTestMethod_NoDiagnostic()
```
Evangelink: I think you should test that there is no warning when the class **has** test method and is abstract but not marked with `[TestClass]`.

Evangelink (approved):

PR #3206, opened 2024-07-04
Downgrade some log levels from info to debug

Evangelink:
Downgrade some log levels to debug so that they don't show up in VS output logs by default.

Fixes https://devdiv.visualstudio.com/DevDiv/_workitems/edit/2112664

cc @MarcoRossignoli @drognanar


MarcoRossignoli (dismissed):

drognanar (approved):

Evangelink:
Will try to work on updating the failing tests during the weekend

Issue #3107, opened 2024-06-15
Suggestion: Issue diagnostic for IsNull/IsNotNull when called with non-nullable type


bjornhellander:
Would be nice with a diagnostic when Assert.IsNotNull and Assert.IsNull is called with a value type other than Nullable<T>, since those asserts will either always pass or always fail.

AB#2112219

Evangelink:
Hi @bjornhellander! Thanks for the suggestion, this would be valuable, adding for our next milestone.

bjornhellander:
I see that a pull request has been created, but I wonder if we could discuss the way forward briefly first, unless you have already discussed this internally?

Since I created this issue, I have discovered MSTEST0025 which triggers on various calls which will always fail. One way of handling this issue could be to 1) update MSTEST0025 to also check for calls to IsNull with a value type other than Nullable<T> and 2) create a new analyzer that checks various calls that will always pass, including calls to IsNotNull with a value type other than Nullable<T>.

Another way could be to promote MSTEST0025 to detect **any** assert call with a pre-determined outcome, regardless if they always pass or always fail.

Or do you think that creating a brand new analyzer specifically for IsNull and IsNotNull is the way to go for this?

I created the issue after refactoring some code, probably changing some field or property from Nullable<T> to T. I see this as the main situation that these diagnostics would be helpful in detecting. That might not be the case for the things that MSTEST0025 currently checks. Feels more like it's detecting noob errors :-) So I understand completely if you want to create a new analyzer, but just wanted to check if you had discussed any alternative options.

Evangelink:
Hi @bjornhellander,

Thanks for the comment! Indeed, I went too fast!

Based on your suggestions and after a quick chat with the team, I think we will move forward with:

1. Update [MSTEST0025](https://learn.microsoft.com/dotnet/core/testing/mstest-analyzers/mstest0025) to handle the always failing part of `Assert.IsNull` and `Assert.IsNotNull`. We may want to change the suggestion because replacing with `Assert.Fail` seems to be only a small fraction of the reason.

2. Create a new rule for handling cases where we have assertions that are always passing (e.g. `Assert.IsTrue(true)`, `Assert.AreEqual(myVar, myVar)`, `Assert.IsNull(myNonNullableStruct)`...)

PR #3210, opened 2024-07-04
MARES0032: Issue diagnostic for IsNull/IsNotNull when called with non-nullable type

engyebrahim:
fix: https://github.com/microsoft/testfx/issues/3107


engyebrahim:
It's not working as i can't find any thing to check if variable is nullable

bjornhellander:
Can you clarify what problems you are having? I will add some comments that might help.

src/Analyzers/MSTest.Analyzers/Helpers/WellKnownTypeNames.cs
```diff
@@ -36,6 +36,7 @@ internal static class WellKnownTypeNames
     public const string SystemDescriptionAttribute = "System.ComponentModel.DescriptionAttribute";
     public const string SystemIAsyncDisposable = "System.IAsyncDisposable";
     public const string SystemIDisposable = "System.IDisposable";
+    public const string SystemNullable = "System.Nullable";
```
bjornhellander, line 39: I suspect this should be like this since this is a generic type:
```suggestion
    public const string SystemNullable = "System.Nullable`1";
```

src/Analyzers/MSTest.Analyzers/UseNullableForIsNullAndIsNotNullAssertionsAnalyzer.cs
```diff
@@ -0,0 +1,69 @@
...
+
+        if ((invocationOperation.TargetMethod.Name != "IsNull" && invocationOperation.TargetMethod.Name != "IsNotNull")
+            || !SymbolEqualityComparer.Default.Equals(assertSymbol, invocationOperation.TargetMethod.ContainingType))
+        {
+            return;
+        }
+
+        foreach (IArgumentOperation argument in invocationOperation.Arguments)
+        {
+            if (SymbolEqualityComparer.Default.Equals(argument.Type, nullableSymbol) && argument.Parameter?.Name == "value") // not working.
```
bjornhellander, line 62: Did it not work because `nullableSymbol` was `null`? See my first comment in that case.

Some additional hints, in case any of them help:

I suspect that you also need to skip implicit conversion operations in the argument to get the type of the actual argument. This is probably true if `argument.Type` is always `object` when you debug.

Another problem is probably that the type of the argument is typically `Nullable<int>` etc, not the generic `Nullable<T>` that you retrieve in the compilation start action. There are properties on the type symbol that you can use to retrieve the original generic type that the type was created from.

The idea was also to detect value types other than `Nullable<T>`, so it seems that your logic is a bit off.

test/UnitTests/MSTest.Analyzers.UnitTests/UseNullableForIsNullAndIsNotNullAssertionsAnalyzerTests.cs
```diff
@@ -0,0 +1,52 @@
...
+        string code = """
+            using Microsoft.VisualStudio.TestTools.UnitTesting;
+
+            [TestClass]
+            public class MyTestClass
+            {
+                [TestMethod]
+                public void MyTestMethod()
+                {
+                    string? var = null;
```
bjornhellander, line 24: Since `string` is a reference type, this will not result in `Nullable<string>`. `Nullable<T>` can only be used with value types. To be clear, maybe it is better to actually write `Nullable<...>` in the tests, instead of the short-hand `?`. You could test with for example `Nullable<int>` and also a generic `Nullable<T>`.

test/UnitTests/MSTest.Analyzers.UnitTests/UseNullableForIsNullAndIsNotNullAssertionsAnalyzerTests.cs
```diff
@@ -0,0 +1,52 @@
...
+        string code = """
+            using Microsoft.VisualStudio.TestTools.UnitTesting;
+
+            [TestClass]
+            public class MyTestClass
+            {
+                [TestMethod]
+                public void MyTestMethod()
+                {
+                    string var = "value";
```
bjornhellander, line 44: This is correct, but not the main concern, for the same reason as above. Also test for example `int` and a generic `T` constrained to be a value type. And maybe even `string?` to be clear.

Issue #3211, opened 2024-07-04
Expose the PID int the server initialization response


drognanar:
If the `initialize` response included a PID, then the client's could grab the process handle, even in scenarios where the client does not directly create the server process.

The request response could look as follows:

```csharp
internal record InitializeRequestArgs(int ProcessId, ClientInfo ClientInfo, ClientCapabilities Capabilities);
internal record InitializeResponseArgs(int? ProcessId, ServerInfo ServerInfo, ServerCapabilities Capabilities);
```

This for instance could allow the client to auto attach during debugging to the RPC process, without the need for an explicit `client/attachDebugger` RPC call.

PR #3212, opened 2024-07-04
Return the pid during the handshake inside the InitializeResponseArgs

MarcoRossignoli:
closes https://github.com/microsoft/testfx/issues/3211

### DIFF
--- a/src/Platform/Microsoft.Testing.Platform/ServerMode/Json/Json.cs
+++ b/src/Platform/Microsoft.Testing.Platform/ServerMode/Json/Json.cs
@@ -80,8 +80,9 @@ internal sealed class Json
 
         _serializers[typeof(InitializeResponseArgs)] = new JsonObjectSerializer<InitializeResponseArgs>(response =>
         {
-            var list = new (string, object?)[2]
+            var list = new (string, object?)[3]
             {
+                (JsonRpcStrings.ProcessId, response.ProcessId),
                 (JsonRpcStrings.ServerInfo, response.ServerInfo),
                 (JsonRpcStrings.Capabilities, response.Capabilities),
             };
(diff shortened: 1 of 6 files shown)


src/Platform/Microsoft.Testing.Platform/ServerMode/SerializerUtilities.cs
```diff
@@ -515,10 +516,11 @@ static SerializerUtilities()
 
         Deserializers[typeof(InitializeResponseArgs)] = new ObjectDeserializer<InitializeResponseArgs>(properties =>
         {
+            int processId = GetRequiredPropertyFromJson<int>(properties, JsonRpcStrings.ProcessId);
```
drognanar, line 519: is it a required property? or should we handle the case that it's missing or null?

src/Platform/Microsoft.Testing.Platform/ServerMode/Json/Json.cs
```diff
@@ -509,6 +510,7 @@ public Json(Dictionary<Type, JsonSerializer>? serializers = null, Dictionary<Typ
 
         _deserializers[typeof(InitializeResponseArgs)] = new JsonElementDeserializer<InitializeResponseArgs>(
           (json, jsonElement) => new InitializeResponseArgs(
+                  ProcessId: json.Bind<int>(jsonElement, JsonRpcStrings.ProcessId),
```
drognanar, line 513: similarly, should we handle the null case?

src/Platform/Microsoft.Testing.Platform/ServerMode/SerializerUtilities.cs
MarcoRossignoli: We force that property in the implementation so we have it always. From the protocol perspective is fine from our implementation I don't have scenario where we cannot get the PID, normal mode(jitter) and AOT let us get it.

src/Platform/Microsoft.Testing.Platform/ServerMode/Json/Json.cs
MarcoRossignoli: ditto

drognanar (approved):

PR #3231, opened 2024-07-09
Refactor STA pool code

MarcoRossignoli:


Evangelink (approved):

src/Adapter/MSTest.TestAdapter/Execution/TestExecutionManager.cs
```diff
@@ -43,34 +43,35 @@ internal TestExecutionManager(IEnvironment environment, Func<Action, Task>? task
...
-    internal static class ThreadingHelper
+    private static Task DefaultFactoryAsync(Action action)
     {
-        public static Task RunAsync(Action action)
+        if (MSTestSettings.RunConfigurationSettings.ExecutionApartmentState == ApartmentState.STA
+            && RuntimeInformation.IsOSPlatform(OSPlatform.Windows))
         {
-            if (MSTestSettings.RunConfigurationSettings.ExecutionApartmentState == ApartmentState.STA
-                && RuntimeInformation.IsOSPlatform(OSPlatform.Windows))
+            TaskCompletionSource<int> tcs = new();
```
danielmarbach, line 54: Is this deliberately not using the asynchronous continuation option? If so would it be worthwhile to add a comment of why?

https://github.com/davidfowl/AspNetCoreDiagnosticScenarios/blob/master/AsyncGuidance.md#always-create-taskcompletionsourcet-with-taskcreationoptionsruncontinuationsasynchronously

src/Adapter/MSTest.TestAdapter/Execution/TestExecutionManager.cs
```diff
@@ -43,34 +43,35 @@ internal TestExecutionManager(IEnvironment environment, Func<Action, Task>? task
...
+                }
+            });
+
+            entryPointThread.SetApartmentState(ApartmentState.STA);
+            entryPointThread.Start();
+            return tcs.Task;
+        }
+        else
+        {
+            return Task.Run(action);
```
danielmarbach, line 74: Depending on the assumptions of the code base this change might have undesirable consequences. If the action delegate can execute arbitrary long running work and has to be assumed to do so the previous way was superior

See https://github.com/davidfowl/AspNetCoreDiagnosticScenarios/blob/master/AsyncGuidance.md#avoid-using-taskrun-for-long-running-work-that-blocks-the-thread

src/Adapter/MSTest.TestAdapter/Execution/TestExecutionManager.cs
MarcoRossignoli: We don't have any continuation on this task but a wait all when the queue is empty https://github.com/microsoft/testfx/blob/main/src/Adapter/MSTest.TestAdapter/Execution/TestExecutionManager.cs#L364

src/Adapter/MSTest.TestAdapter/Execution/TestExecutionManager.cs
MarcoRossignoli: The rationale is that we expect usually quick tests like unit tests with a lot of data sets. In that case we save the creation of custom expensive threads and we will have better reuse (long running will create new thread, we see it often in the profilers and we wait the creation of all of those before to start the tests, so for few quick tests we're paying for no reason).

If tests are "slow" anyway we/you will have to wait...so follow the thread pool injection policy that is handling also the blocking(Wait() vs await) api won't make a lot of difference compared to custom threads from timing/performance perspective.

On an internal custom adapter implementation based on thread pool we saw better perf.

If user wants to override the injection policy it can always use inside the assembly init the min thread pool threads.

danielmarbach: Thanks for the explanation. Sorry for the drive by comments. I read through the change list and saw this change. Got curious about the reasons behind the nature of the change and since the PR didn't contain much more information I figured I ask.

MarcoRossignoli: Some more thoughts on it, here we're not in an asp.net application where you should free the thread pool soon as possible to have a good throughput. There you don't want that your "background" workers steal resources and decrease the RPS, the background work can be done in async, handle a request needs to be done soon as possible. 

In testing world/engine, tests are our "requests" so the thread pool is our "scheduler", the thread pool solves a lot of issues related to how to scale w.t.r. the amount of available resources.

src/Adapter/MSTest.TestAdapter/Execution/TestExecutionManager.cs
MarcoRossignoli: Some more thoughts on it, MSTest code is pretty old and we're trying to move it to be fully async, we're not yet there so despite you can define a `Task` return test, it is handled in a suboptimal way.

cc: @Evangelink

src/Adapter/MSTest.TestAdapter/Execution/TestExecutionManager.cs
MarcoRossignoli: > Thanks for the explanation. Sorry for the drive by comments. I read through the change list and saw this change. Got curious about the reasons behind the nature of the change and since the PR didn't contain much more information I figured I ask.

No worries and my bad I apologize, I did the change after a quick internal sync and completely forgot  to properly describe.

danielmarbach: FYI I linked to those guidelines not because I wanted to necessarily compare to asp.net. Many of those guidelines apply to broader scenarios and it was the quickest available source to link while I was typing my thoughts.

What I see on CI/CD environments is that they often run on shared runners with limited CPU resources available. But given the added explanation I see the tradeoffs make sense

MarcoRossignoli: > What I see on CI/CD environments is that they often run on shared runners with limited CPU resources available. But given the added explanation I see the tradeoffs make sense

True, anyway we're limiting the number of workers by default to the numbers of CPU and you can decide the degree with https://learn.microsoft.com/en-us/dotnet/api/microsoft.visualstudio.testtools.unittesting.parallelizeattribute

PR #3235, opened 2024-07-09
Customize MSTest runner banner

Evangelink:
Change default banner from being testing platform banner to MSTest banner.

Before `.NET Testing Platform v1.3.0 XXX`
After `MSTest v3.5.0 XXXX`


MarcoRossignoli (approved):
:shipit:

PR #3240, opened 2024-07-10
Workaround the UWP release mode issue with DataRow data

### LINKED ISSUE #3071: DataRows in UWP Release builds no longer works
## Describe the bug

If you use data rows in a UWP app and run the tests in release builds, you're met with the following error:
```
  Failed TestMethod1 (System.Object[]) [< 1 ms]
  Error Message:
   Cannot run test method 'UnitTestProject1.UnitTest1.TestMethod1': Test data doesn't match method parameters. Either the count or types are different.
Test expected 2 parameter(s), with types 'String, Int32',
but received 1 argument(s), with types 'Object[]'.
```

## Steps To Reproduce

1. Create a new UWP Unit Test project.
2. Upgrade to v3.4.3
3. Add a datarow test
```cs
        [TestMethod]
        [DataRow("Name", 1)]
        [DataRow("Other Name", 2)]
        public void TestMethod1(string val1, int val2)
        {
            Assert.IsFalse(string.IsNullOrEmpty(val1));
            Assert.IsTrue(val2 > 0);
        }
```
4. Build in release mode.
5. From commandline (testexplorer doesn't seem to work with release builds) run the following command (Replace `UnitTestProject1` with your appname):
   `vstest.console.exe bin\x86\Release\UnitTestProject1.build.appxrecipe /InIsolation /Platform:x86 /framework:FrameworkUap10`

## Expected behavior
Test passes

## Actual behavior
Error:
```
Test expected 2 parameter(s), with types 'String, Int32',
but received 1 argument(s), with types 'Object[]'.
```
## Additional context

Also reproduced the problem with 3.4.0.
With 3.3.0 and 3.2.0 you get a different error, but along the same lines: `System.ArgumentException: Object of type 'System.Object[]' cannot be converted to type 'System.String'.`
v3.1.1 won't run due to adapter issue.
v3.0.4 works and can even launch the tests in release mode from test explorer

Evangelink:
There seems to be a bug in UWP release mode when ".NET Native tool chain" is enabled causing the data to be wrapped in an `object[]`. While waiting for a proper fix, we are adding a logic to unwrap the data.

Manually tested.

Fixes #3071

### DIFF
--- a/src/Adapter/MSTest.TestAdapter/Extensions/MethodInfoExtensions.cs
+++ b/src/Adapter/MSTest.TestAdapter/Extensions/MethodInfoExtensions.cs
@@ -162,10 +162,18 @@ internal static class MethodInfoExtensions
         {
             int methodParametersLengthOrZero = methodParameters?.Length ?? 0;
             int argumentsLengthOrZero = arguments?.Length ?? 0;
-            if (methodParametersLengthOrZero != argumentsLengthOrZero)
+
+#if WINDOWS_UWP
+            // There is a bug with UWP in release mode where the arguments are wrapped in an object[], so we need to unwrap it.
+            // See https://github.com/microsoft/testfx/issues/3071
+            if (argumentsLengthOrZero == 1
+                && argumentsLengthOrZero < methodParametersLengthOrZero
+                && arguments![0] is object[] args)
             {
-                throw GetParameterCountMismatchException(methodInfo, arguments, methodParameters, methodParametersLengthOrZero, argumentsLengthOrZero, innerException: null);
+                arguments = args;
+                argumentsLengthOrZero = args.Length;
             }
+#endif
 
             try
             {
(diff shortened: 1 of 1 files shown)


src/Adapter/MSTest.TestAdapter/Extensions/MethodInfoExtensions.cs
```diff
@@ -162,18 +162,36 @@ internal static void InvokeAsSynchronousTask(this MethodInfo methodInfo, object?
...
             int methodParametersLengthOrZero = methodParameters?.Length ?? 0;
             int argumentsLengthOrZero = arguments?.Length ?? 0;
-            if (methodParametersLengthOrZero != argumentsLengthOrZero)
+
+#if WINDOWS_UWP
+            // There is a bug with UWP in release mode where the arguments are wrapped in an object[], so we need to unwrap it.
+            // See https://github.com/microsoft/testfx/issues/3071
+            if (argumentsLengthOrZero == 1
+                && argumentsLengthOrZero < methodParametersLengthOrZero
+                && arguments![0] is object[] args)
```
Evangelink, line 171: We could be even more restrictive and only unwrap if `args.Length == methodParametersLengthOrZero`, WDYT?

PR #3243, opened 2024-07-10
Skipped tests count as not run

### LINKED ISSUE #3216: If all tests are skipped, exit code is 0 not 8
Unsure whether you guys would consider this a bug, but it doesn't make sense in my head to try to execute a test suite and then skip everything. That would imply to me something has gone wrong. 

And if you really did want to do that, you could just pass in --ignore-exit-code 8

So I'd want an error exit code to be returned to make me aware of that.

Repro:

```csharp
{

    {
        {
            Uid = "2",
            DisplayName = "Blah",
            Properties = new PropertyBag(new SkippedTestNodeStateProperty())
        }));
     
        context.Complete();
        
        return Task.CompletedTask;
    }

}

internal class Capabilities : ITestFrameworkCapabilities
{
    IReadOnlyCollection<ITestFrameworkCapability> ICapabilities<ITestFrameworkCapability>.Capabilities
        => Array.Empty<ITestFrameworkCapability>();
}
```

Evangelink:
Update the logic to consider skipped tests as not run:

- If all tests are skipped, exit code is 8
- If all tests are skipped, summary considers run as failed

Fixes #3216


src/Platform/Microsoft.Testing.Platform/Messages/TestNodeProperties.Categories.cs
```diff
@@ -15,7 +15,6 @@ internal static class TestNodePropertiesCategories
     public static Type[] WellKnownTestNodeTestRunOutcomeProperties { get; } =
         [
             typeof(PassedTestNodeStateProperty),
-            typeof(SkippedTestNodeStateProperty),
```
MarcoRossignoli, line 18: this could break some extension I suppose, is it needed?

Evangelink: Type is `internal` so it should be fine. Also, yes the array is used directly to count test that run. As we said, skipped should not be considered as run.

MarcoRossignoli: It's internal but used around by built-in extensions too, I would take a look around before and in case duplicate it

Evangelink: Not used anywhere else in this repo (all OSS extensions: https://github.com/search?q=repo%3Amicrosoft%2Ftestfx%20WellKnownTestNodeTestRunOutcomeProperties%20&type=code) but indeed retry is impacted

Issue #989, opened 2021-09-29
Generic overloads for CollectionAssert


daniel-lerch:
### Motivation
I have been using `CollectionAssert` for quite some time but when I want to check whether the contents of a `ReadOnlyCollection<T>` match my expectation a cannot use `CollectionAssert` because many generic collection interfaces do not inherit from `ICollection`.

### Proposal
Add new generic overloads of all methods to take `IReadOnlyCollection<T>` or `IEnumerable<T>` instead of `ICollection`.

Haplois:
Hi, thanks for the proposal. 

If you're up for it, we are open for community PR's.

daniel-lerch:
Because of some difficulties described in #990, I will not continue to work on this issue and migrate my unit tests to Xunit instead.

jbcutting:
Note: Another advantage of this will be the ability to use collection expressions with `CollectionAssert`, e.g. `CollectionAssert.AreEqual([1, 2, 3], [4, 5, 6])` as opposed to having to specify the type with the non-generic `ICollection`, e.g., `CollectionAssert.AreEqual(new[] { 1, 2, 3 }, (int[])[4, 5, 6])`

Evangelink:
Hi all! Just a quick note to say that we are all convinced about this, this sadly just lacks prioritization at the moment.

RioMcBoo:
Hi, how can i see the code and do changes? I created fork, but it is only headers inside =(

What the problem with Indentation? I wanted to help

nohwnd:
@RioMcBoo which file are you looking at where there are only headers?

RioMcBoo:
> [@RioMcBoo](https://github.com/RioMcBoo) which file are you looking at where there are only headers?

https://github.com/microsoft/testfx/blob/main/src/TestFramework/TestFramework/Assertions/CollectionAssert.Equality.cs - this one

The methods nested in head methods highlighted by white color and any attempts to open them to look inside follows with "you can not move caret to it" or something like this

nohwnd:
In those files I see regions.

<img width="700" height="353" alt="Image" src="https://github.com/user-attachments/assets/6be19ed4-12e1-4b41-96f4-2b427ceb3234" />

Once expanded the code is there, and in the code I can navigate to other parts of the class with ctrl+click.

<img width="758" height="192" alt="Image" src="https://github.com/user-attachments/assets/085736df-db60-4109-a7b6-e29b285944a5" />

I see no issue on my side.

[...]

PR #3258, opened 2024-07-12
Generic overloads for CollectionAssert

engyebrahim:
fix: https://github.com/microsoft/testfx/issues/989


src/TestFramework/TestFramework/Assertions/CollectionAssert.cs
```diff
@@ -176,6 +250,80 @@ public static void DoesNotContain([NotNull] ICollection? collection, object? ele
     /// </exception>
     public static void DoesNotContain([NotNull] ICollection? collection, object? element, [StringSyntax(StringSyntaxAttribute.CompositeFormat)] string? message,
         params object?[]? parameters)
+    => DoesNotContain(collection?.Cast<object>(), element, message, parameters);
```
Evangelink: Indentation is broken :)

src/TestFramework/TestFramework/Assertions/CollectionAssert.cs
```diff
@@ -302,45 +510,110 @@ public static void AllItemsAreUnique([NotNull] ICollection? collection, string?
...
+    /// An array of parameters to use when formatting <paramref name="message"/>.
+    /// </param>
+    /// <exception cref="AssertFailedException">
+    /// <paramref name="collection"/> is null, or <paramref name="collection"/> contains at least one duplicate
+    /// element.
+    /// </exception>
+    public static void AllItemsAreUnique<T>([NotNull] IEnumerable<T?>? collection, [StringSyntax(StringSyntaxAttribute.CompositeFormat)] string? message,
+        params object?[]? parameters)
+    {
+        Assert.CheckParameterNotNull(collection, "CollectionAssert.AllItemsAreUnique", "collection", string.Empty);
```
Evangelink: Please do call out if the method was only moved or if the content was changed, that's easing reviewer's job.

src/TestFramework/TestFramework/Assertions/CollectionAssert.cs
```diff
@@ -302,45 +510,110 @@ public static void AllItemsAreUnique([NotNull] ICollection? collection, string?
...
+
+        message = Assert.ReplaceNulls(message);
+
+        bool foundNull = false;
+        Dictionary<object, bool> table = [];
+        foreach (object? current in collection)
```
Evangelink: Current is not `object?` but `T?`

src/TestFramework/TestFramework/Assertions/CollectionAssert.cs
```diff
@@ -1652,5 +2701,23 @@ private class ObjectComparer : IComparer
...
+        {
+            Assert.CheckParameterNotNull(comparer, "Assert", "comparer", string.Empty);
+            _comparer = comparer;
+        }
+
+        bool IEqualityComparer<T>.Equals(T? x, T? y) =>
+            // Use the comparer to check if the two items are equal
+            _comparer?.Compare(x, y) == 0;
+
+        public int GetHashCode(T obj) =>
```
Evangelink: If I do recall well, the `GetHashCode` is sometimes called instead of `Equals` (e.g. key of `Dictionary`) so the logic should be updated to return the same hashcode in case of equality.

@MarcoRossignoli could you confirm or infirm?

MarcoRossignoli: Equals and GetHashCode have to be equivalent...GetHashCode is used for instance when the object is a key in an hashtable...but in general if A.Equals(B) = true => A.GetHashCode() == B.GetHashCode()

src/TestFramework/TestFramework/Assertions/Assert.LegacyCollection .cs
```diff
@@ -0,0 +1,966 @@
...
+    /// The collection in which to search for the element.
+    /// </param>
+    /// <param name="element">
+    /// The element that is expected to be in the collection.
+    /// </param>
+    /// <exception cref="AssertFailedException">
+    /// <paramref name="collection"/> is null, or <paramref name="collection"/> does not contain
+    /// element <paramref name="element"/>.
+    /// </exception>
+    public static void ContainsLegacy([NotNull] ICollection? collection, object? element)
```
Evangelink: As discussed all `ICollection` should be `IEnumerable`

src/TestFramework/TestFramework/Assertions/Assert.GenericCollection.cs
```diff
@@ -0,0 +1,1782 @@
...
+    /// The collection in which to search for the element.
+    /// </param>
+    /// <param name="element">
+    /// The element that is expected to be in the collection.
+    /// </param>
+    /// <exception cref="AssertFailedException">
+    /// <paramref name="collection"/> is null, or <paramref name="collection"/> does not contain
+    /// element <paramref name="element"/>.
+    /// </exception>
+    public static void Contains<T1, T2>([NotNull] IEnumerable<T1?>? collection, T2? element)
```
Evangelink, line 40: Do we need the 2 generic types?

src/TestFramework/TestFramework/Assertions/CollectionAssert.cs
```diff
@@ -13,6 +13,7 @@ namespace Microsoft.VisualStudio.TestTools.UnitTesting;
 /// with collections within unit tests. If the condition being tested is not
 /// met, an exception is thrown.
 /// </summary>
+[Obsolete("We have now Assert.XXX with all generic CollectionAsserts and Assert.XXXLegacy for old on")]
```
Evangelink, line 16:
```suggestion
[Obsolete("Replace calls to CollectionAssert.XXX with Assert.XXX for generic collections or Assert.XXXLegacy for non-generic ones")]
```

PR #3264, opened 2024-07-15
Fix ResultFiles placement in TRX report

### LINKED ISSUE #3260: TRX file: <ResultFiles> not under node <UnitTestResult>, but under <Output> node
## Describe the bug

During my testing for our migration from MSTest meta nuget package to the new MSTest SDK, I noticed that Azure DevOps doesn't upload files anymore, which were added via MSTests `TestContext.AddResultFile(myPath)` method. It works when not using the new MSTest SDK. I found that the resulting `.trx` file has a slight change in where the node `<ResultFiles>` is placed.

## Expected behavior
- Old behavior (without MSTest SDK)
- `<ResultFiles>` is placed under `<UnitTestResult>` on the same level as `Output`
image

## Actual behavior
- New behavior (with MSTest SDK)
- `<ResultFiles>` is placed under `<Output>`
image

## Additional context

I tried different settings and combinations in my `Directory.Build.props`. But they have no impact on the described bug.
I also tried to change (and shorten) the local folder which is used to save the files. But also no success.
-   https://learn.microsoft.com/en-us/dotnet/core/testing/unit-testing-mstest-sdk
-   https://learn.microsoft.com/en-us/dotnet/core/testing/unit-testing-platform-integration-dotnet-test
-   https://learn.microsoft.com/en-us/dotnet/core/testing/unit-testing-mstest-configure

```xml
  <PropertyGroup Condition="$(MSBuildProjectName.EndsWith('Tests'))">
    <TestingPlatformDotnetTestSupport>true</TestingPlatformDotnetTestSupport>
    <TestingExtensionsProfile>Default</TestingExtensionsProfile>
    <TestingPlatformShowTestsFailure>true</TestingPlatformShowTestsFailure>
    <TestingPlatformCaptureOutput>false</TestingPlatformCaptureOutput>
  </PropertyGroup>
  ```
 I noticed another odd behavior, but I'm not sure if it is related. As you can see in the attached trx-files: The same test got way more <ResultFile> entries now. The test itself didn't change. But each test (one test per datarow) gets all the resultfiles from the other datarows attached. But before examining this, I want to wait until the above bug is resolved

nohwnd:
ResultFiles tag was incorrectly placed in the Output. Move it out under TestResults where it belongs.

Fix #3260

### DIFF
--- a/src/Platform/Microsoft.Testing.Extensions.TrxReport/TrxReportEngine.cs
+++ b/src/Platform/Microsoft.Testing.Extensions.TrxReport/TrxReportEngine.cs
@@ -525,6 +525,12 @@ internal sealed partial class TrxReportEngine
                 output.Add(errorInfoElement);
             }
 
+            // add collectorDataEntries details
+            if (output.HasElements && outcome != "NotExecuted")
+            {
+                unitTestResult.Add(output);
+            }
+
             if (_artifactsByTestNode.TryGetValue(testNode.Uid, out List<SessionFileArtifact>? fileArtifacts))
             {
                 var resultFiles = new XElement("ResultFiles");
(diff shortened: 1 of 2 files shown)


Evangelink:
Thank you for the fast fix @nohwnd <3

PR #3275, opened 2024-07-17
Fix collectionAssert.AreEqual fails for list of lists

engyebrahim:
fix: https://github.com/microsoft/testfx/issues/711

### DIFF
--- a/src/TestFramework/TestFramework/Assertions/CollectionAssert.cs
+++ b/src/TestFramework/TestFramework/Assertions/CollectionAssert.cs
@@ -1611,48 +1611,60 @@ public sealed class CollectionAssert
             return true;
         }
 
-        return CompareCollections(expected, actual, comparer, ref reason);
+        return CompareIEnumerable(expected, actual, comparer, ref reason);
     }
 
-    private static bool CompareCollections(ICollection? expected, ICollection? actual, IComparer comparer, ref string reason)
+    private static bool CompareIEnumerable(IEnumerable? expected, IEnumerable? actual, IComparer comparer, ref string reason)
     {
         if ((expected == null) || (actual == null))
         {
             return false;
         }
 
-        if (expected.Count != actual.Count)
+        var stack = new Stack<Tuple<IEnumerator, IEnumerator, int>>();
+        stack.Push(new(expected.GetEnumerator(), actual.GetEnumerator(), 0));
+
+        while (stack.Count > 0)
         {
-            reason = FrameworkMessages.NumberOfElementsDiff;
-            return false;
-        }
+            Tuple<IEnumerator, IEnumerator, int> cur = stack.Pop();
+            IEnumerator expectedEnum = cur.Item1;
+            IEnumerator actualEnum = cur.Item2;
+            int position = cur.Item3;
 
-        IEnumerator expectedEnum = expected.GetEnumerator();
-        IEnumerator actualEnum = actual.GetEnumerator();
-
-        int position = 0;
-        while (expectedEnum.MoveNext() && actualEnum.MoveNext())
-        {
-            object? curExpected = expectedEnum.Current;
-            object? curActual = actualEnum.Current;
-
-            if (curExpected is ICollection curExpectedCollection && curActual is ICollection curActualCollection)
+            while (expectedEnum.MoveNext())
             {
-                if (!CompareCollections(curExpectedCollection, curActualCollection, comparer, ref reason))
+                if (!actualEnum.MoveNext())
                 {
+                    reason = FrameworkMessages.NumberOfElementsDiff;
                     return false;
                 }
-            }
-            else if (comparer.Compare(curExpected, curActual) != 0)
-            {
-                reason = string.Format(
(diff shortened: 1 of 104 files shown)


test/UnitTests/TestFramework.UnitTests/Assertions/CollectionAssertTests.cs
```diff
@@ -215,6 +215,38 @@ public void CollectionAssertAreEqualComparerMessageParametersNullabilityPostCond
         comparer.ToString(); // no warning
     }
 
+    public void CollectionAssertAreEqual_EqualNestedLists_Passes()
```
Evangelink, line 219: Can you please add a negative test for both AreEqual and AreNotEqual (a test that fails)

src/TestFramework/TestFramework/Assertions/CollectionAssert.cs
```diff
@@ -1605,43 +1605,57 @@ internal static bool IsSubsetOfHelper(ICollection subset, ICollection superset)
...
+        while (expectedEnum.MoveNext() && actualEnum.MoveNext())
+        {
+            object? curExpected = expectedEnum.Current;
+            object? curActual = actualEnum.Current;
+
+            if (curExpected is ICollection curExpectedCollection && curActual is ICollection curActualCollection)
             {
-                bool areEqual = comparer.Compare(expectedEnum.Current, actualEnum.Current) == 0;
-                if (!areEqual)
+                if (!CompareCollections(curExpectedCollection, curActualCollection, comparer, ref reason))
```
Evangelink: It's usually a good idea to avoid recursion and to use a stack instead (push/pop) to avoid StackOverflow if you have a really deep nested collection.

src/TestFramework/TestFramework/Assertions/CollectionAssert.cs
```diff
@@ -1605,43 +1605,57 @@ internal static bool IsSubsetOfHelper(ICollection subset, ICollection superset)
...
+        IEnumerator expectedEnum = expected.GetEnumerator();
+        IEnumerator actualEnum = actual.GetEnumerator();
+
+        int position = 0;
+        while (expectedEnum.MoveNext() && actualEnum.MoveNext())
+        {
+            object? curExpected = expectedEnum.Current;
+            object? curActual = actualEnum.Current;
+
+            if (curExpected is ICollection curExpectedCollection && curActual is ICollection curActualCollection)
```
Evangelink: We may want to allow IEnumerable because I think your fix wouldn't work for a non-ICollection inner collection. You could try with `List<ReadOnlyCollection<int>>`

src/TestFramework/TestFramework/Assertions/CollectionAssert.cs
```diff
@@ -1605,43 +1605,69 @@ internal static bool IsSubsetOfHelper(ICollection subset, ICollection superset)
...
+                if (!actualEnum.MoveNext())
+                {
+                    reason = FrameworkMessages.NumberOfElementsDiff;
+                    return false;
+                }
+
+                object? curExpected = expectedEnum.Current;
+                object? curActual = actualEnum.Current;
+
+                if (curExpected is IEnumerable curExpectedEnum && curActual is IEnumerable curActualEnum)
```
Evangelink, line 1645: Do we have a test with a collection type that isn't `ICollection` to avoid regression?

src/TestFramework/TestFramework/Assertions/CollectionAssert.cs
```diff
@@ -1605,43 +1605,69 @@ internal static bool IsSubsetOfHelper(ICollection subset, ICollection superset)
...
+        }
 
-            IEnumerator expectedEnum = expected.GetEnumerator();
-            IEnumerator actualEnum = actual.GetEnumerator();
-            int i = 0;
-            while (expectedEnum.MoveNext() && actualEnum.MoveNext())
+        var stack = new Stack<Tuple<IEnumerator, IEnumerator, int>>();
+        stack.Push(new(expected.GetEnumerator(), actual.GetEnumerator(), 0));
+
+        while (stack.Count > 0)
```
Evangelink, line 1627: What would be ideal is to do a test that generates a huge collection

Evangelink (commented):
If you can add the 2 more tests so we have a strong coverage, that'd be awesome

src/TestFramework/TestFramework/Assertions/CollectionAssert.cs
engyebrahim: I did a test with List<ReadOnlyCollection<int>>

Evangelink: Yes sorry I am blind...

Evangelink (dismissed):

Evangelink (approved):

Evangelink (dismissed):

Evangelink (dismissed):

Evangelink (approved):

PR #3286, opened 2024-07-18
Feat: Add [STATestMethod]

Evangelink:
Relates to #2688


MarcoRossignoli (approved):

MWinhard:
[STATestClass] and [STATestMethod] attributes are still not recognized. I'm using latest stable MSTest 3.5.2 and Visual Studio 2022 17.11.0. Do I need something else?

Evangelink:
They will be included in 3.6.0, we can release a public preview version if you are interested.

MWinhard:
Thank you for the quick reply. I can wait for 3.6.0. But please document in [https://learn.microsoft.com/en-us/dotnet/core/testing/unit-testing-mstest-writing-tests-attributes](url) that `[STATestClass]` and `[STATestMethod]` are available "starting with MSTest 3.6", as with the `CooperativeCancellation` property.

Evangelink:
@MWinhard that's valuable feedback, thank you. I am addressing it here https://github.com/dotnet/docs/pull/42330

PR #3292, opened 2024-07-22
Interactive display for terminals

nohwnd:
Adds output device that writes tests to ansi or non-ansi terminal, with or without progress indicator.

Features: 

A "progress" bar that shows counts of passed, failed and skipped tests. In ANSI mode the progress is updated live on the bottom of the screen, in non-ansi mode the progress is output repeatedly every 3 seconds. 

image

Colored stack traces, with links to files, and (optional) relative paths: 

image

Test run summary: 

image

artifacts report: 

image

Optionally showing passed tests: 

image

Optionally showing link to assembly the test is coming from: 

image

image

Optional per assembly summary: 
image

- [x] implement TestHostControllerInfo.CurrentProcessIsTestHostController to disable progress reporting in controller.
- [x] fix output tests

Split:
-  implement Console.Out that will respect when we are writing or not writing progress so we don't end up keeping progress on screen. Also same problem when the controller outputs to screen. - moving to https://github.com/microsoft/testfx/issues/3491
- surface options to show passed tests etc.  - moved to https://github.com/microsoft/testfx/issues/3492


src/Platform/Microsoft.Testing.Platform/OutputDevice/OutputFormatter.cs
```diff
@@ -0,0 +1,29 @@
...
+        }
+
+#pragma warning disable IDE0046 // Convert to conditional expression
+        if (RoslynString.IsNullOrWhiteSpace(text))
+        {
+            return text.Replace("\r", "\\r").Replace("\n", "\\n").Replace("\t", "\\t");
+        }
+#pragma warning restore IDE0046 // Convert to conditional expression
+
+        return text;
```
Evangelink:
```suggestion
        if (RoslynString.IsNullOrWhiteSpace(text))
        {
            return text.Replace("\r", "\\r").Replace("\n", "\\n").Replace("\t", "\\t");
        }

        // Fallback to returning original string.
        return text;
```

src/Platform/Microsoft.Testing.Platform/UI/AssemblyRun.cs
```diff
@@ -0,0 +1,48 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+namespace Microsoft.Testing.Platform.UI;
+
+internal sealed class AssemblyRun
```
MarcoRossignoli: I would call it TestModule, we support aot and we could have more asm per module(one to many with custom entry point).

src/Adapter/MSTestAdapter.PlatformServices/Services/ThreadSafeStringWriter.cs
```diff
@@ -28,7 +28,7 @@ public class ThreadSafeStringWriter : StringWriter
     /// we initiate this AsyncLocal in a place that is a separate Task for each test, and so any output that is written into a common stream
     /// (e.g. via Console.WriteLine - which is static, and hence common), will them be multiplexed by the appropriate AsyncLocal, and in effect
     /// we will get outputs splits for each test even if two tests run at the same time and write into console.
-    /// See https://github.com/microsoft/testfx/pull/1705 for a fix of a related bug, in that bug we intialized the state too early, and it then inherited
+    /// See https://github.com/microsoft/testfx/pull/1705 for a fix of a related bug, in that bug we initialized the state too early, and it then inherited
```
Evangelink: Different PR (or push to main) please.

src/Platform/Microsoft.Testing.Platform/OutputDevice/OutputFormatter.cs
nohwnd: 👍

nohwnd:
failed test, failed summary:
image

progress indicator: 

image

optional output of passed tests, and optional source of the message (will be useful for dotnet test): 

image

optional assembly summary: 

image

attachements: 

image


busy indicator: 

image

src/Platform/Microsoft.Testing.Platform/UI/AssemblyRun.cs
nohwnd: renamed.

src/Platform/Microsoft.Testing.Platform/UI/AnsiCodes.cs
```diff
@@ -0,0 +1,144 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+namespace Microsoft.Testing.Platform.UI;
```
Evangelink: I would move all the `UI` folders/namespaces under `OutputDevice` (I am fine to create a `Console` folder/namespace under it).

src/Platform/Microsoft.Testing.Platform/UI/ITerminal.cs
```diff
@@ -0,0 +1,44 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+namespace Microsoft.Testing.Platform.UI;
+
+/// <summary>
+/// An ANSI or non-ANSI terminal that is capable of rendering the messages from <see cref="ConsoleWithProgress"/>.
+/// </summary>
+internal interface ITerminal
```
Evangelink: I don't have a good suggestion for now but it seems weird to me reading the code to have `ITerminal` and `IConsole`

src/Platform/Microsoft.Testing.Platform/UI/AnsiTerminal.cs
```diff
@@ -0,0 +1,307 @@
...
+    private readonly string? _baseDirectory;
+    private readonly bool _useBusyIndicator;
+    private readonly StringBuilder _stringBuilder = new();
+    private bool _isBatching;
+    private TestWorkerFrame _currentFrame = new(Array.Empty<TestWorker>(), 0, 0);
+
+    public AnsiTerminal(IConsole console, string? baseDirectory)
+    {
+        _console = console;
+        // TODO: is this unsafe?
```
Evangelink: I would avoid such todo, either we have a doubt and we should investigate or have a ticket to investigate or we just drop the todo.

src/Platform/Microsoft.Testing.Platform/UI/AnsiTerminal.cs
```diff
@@ -0,0 +1,307 @@
...
+        {
+            _console.Write(AnsiCodes.HideCursor);
+        }
+    }
+
+    public void StartUpdate()
+    {
+        if (_isBatching)
+        {
+            throw new InvalidOperationException("Console is already in batching mode.");
```
Evangelink: Message should be in a resx.

src/Platform/Microsoft.Testing.Platform/UI/AnsiTerminal.cs
```diff
@@ -0,0 +1,307 @@
...
+    public void AppendLink(string? path, int? lineNumber)
+    {
+        if (RoslynString.IsNullOrWhiteSpace(path))
+        {
+            return;
+        }
+
+        // For non code files, point to the directory, so we don't end up running the
+        // exe by clicking at the link.
+        bool linkToFile = path.EndsWith(".cs", ignoreCase: true, CultureInfo.CurrentCulture)
```
Evangelink: It might be a little limiting, what about script files?

src/Platform/Microsoft.Testing.Platform/UI/ConsoleLogger.cs
```diff
@@ -0,0 +1,701 @@
...
+
+using Microsoft.Testing.Platform.Helpers;
+using Microsoft.Testing.Platform.Resources;
+
+namespace Microsoft.Testing.Platform.UI;
+
+/// <summary>
+/// Console logger that outputs test progress and is capable of writing ANSI or non-ANSI output via the given terminal.
+/// </summary>
+internal partial class ConsoleLogger : IDisposable
```
Evangelink: Why partial? It doesn't seem there are other partial parts.

src/Platform/Microsoft.Testing.Platform/UI/ConsoleLogger.cs
```diff
@@ -0,0 +1,701 @@
...
+    private readonly ConsoleLoggerOptions _options;
+
+    private readonly ConsoleWithProgress _consoleWithProgress;
+
+    private readonly uint? _originalConsoleMode;
+
+    /// <summary>
+    /// Time of the oldest observed test target start.
+    /// </summary>
+    private DateTimeOffset? _testStartTime;
```
Evangelink: I would rename the variable to be more precise and avoid the comment (e.g. `_oldestTestStartTime`).

src/Platform/Microsoft.Testing.Platform/UI/ConsoleLogger.cs
```diff
@@ -0,0 +1,701 @@
...
+
+    /// <summary>
+    /// Time of the most recently observed test target finished.
+    /// </summary>
+    private DateTimeOffset? _testEndTime;
```
Evangelink: Same comment here about having a more precise name.

src/Platform/Microsoft.Testing.Platform/UI/ConsoleLogger.cs
```diff
@@ -0,0 +1,701 @@
...
+    /// </summary>
+    internal ConsoleLogger(IConsole console, ConsoleLoggerOptions options)
+    {
+        _options = options;
+
+        bool showProgress = _options.ShowProgress;
+        ConsoleWithProgress consoleWithProgress;
+        if (!_options.UseAnsi)
+        {
+            consoleWithProgress = new ConsoleWithProgress(new AnsiTerminal(console, _options.BaseDirectory), showProgress, updateEvery: 33);
```
Evangelink: I would create a constant for these values and add a comment about how you picked the value.

src/Platform/Microsoft.Testing.Platform/UI/ConsoleLogger.cs
```diff
@@ -0,0 +1,701 @@
...
+        }
+        else
+        {
+            // Autodetect.
+            (bool consoleAcceptsAnsiCodes, bool _, uint? originalConsoleMode) = NativeMethods.QueryIsScreenAndTryEnableAnsiColorCodes();
+            _originalConsoleMode = originalConsoleMode;
+            consoleWithProgress = consoleAcceptsAnsiCodes
+                ? new ConsoleWithProgress(new AnsiTerminal(console, _options.BaseDirectory), showProgress, updateEvery: 33)
```
Evangelink: Same comment about constants and comment.

src/Platform/Microsoft.Testing.Platform/UI/ConsoleLogger.cs
```diff
@@ -0,0 +1,701 @@
...
+        _consoleWithProgress.StartShowingProgress(workerCount);
+    }
+
+    public void AssemblyRunStarted(string assembly, string? targetFramework, string? architecture)
+    {
+        if (_options.ShowAssembly && _options.ShowAssemblyStartAndComplete)
+        {
+            _consoleWithProgress.WriteToTerminal(terminal =>
+            {
+                terminal.Append("Running tests from ");
```
Evangelink: I suppose this needs some translation and so should be in a resx?

src/Platform/Microsoft.Testing.Platform/UI/ConsoleLogger.cs
```diff
@@ -0,0 +1,701 @@
...
+        }
+
+        int totalTests = _assemblies.Values.Sum(a => a.TotalTests);
+        int totalFailedTests = _assemblies.Values.Sum(a => a.FailedTests);
+        int totalSkippedTests = _assemblies.Values.Sum(a => a.SkippedTests);
+
+        bool notEnoughTests = totalTests < _options.MinimumExpectedTests;
+        bool allTestsWereSkipped = totalTests == 0 || totalTests == totalSkippedTests;
+        bool anyTestFailed = totalFailedTests > 0;
+        bool runFailed = anyTestFailed || notEnoughTests || allTestsWereSkipped || _wasCancelled;
```
Evangelink: Could we reuse the logic from the console class? I'm pretty sure the logic will evolve differently because it's split.

src/Platform/Microsoft.Testing.Platform/UI/ConsoleLogger.cs
```diff
@@ -0,0 +1,701 @@
...
+        terminal.SetColor(runFailed ? TerminalColor.Red : TerminalColor.Green);
+
+        terminal.AppendLine();
+        terminal.Append($"Test run summary: ");
```
Evangelink: Nit: for perf

```suggestion
        terminal.Append("Test run summary: ");
```

src/Platform/Microsoft.Testing.Platform/UI/ConsoleLogger.cs
```diff
@@ -0,0 +1,701 @@
...
+    /// </summary>
+    private static void AppendAssemblyResult(ITerminal terminal, bool succeeded, int countErrors, int countWarnings)
+    {
+        if (!succeeded)
+        {
+            terminal.SetColor(TerminalColor.DarkRed);
+            // If the build failed, we print one of three red strings.
+            string text = (countErrors > 0, countWarnings > 0) switch
+            {
+                (true, true) => $"failed with {countErrors} error(s) and {countWarnings} warning(s)",
```
Evangelink: Resx?

src/Platform/Microsoft.Testing.Platform/UI/TestWorker.cs
```diff
@@ -0,0 +1,35 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+namespace Microsoft.Testing.Platform.UI;
+
+internal class TestWorker
```
Evangelink: It's not super clear what this is representing. Is this a `TestProgressState` object?

src/Platform/Microsoft.Testing.Platform/UI/StopwatchAbstraction.cs
```diff
@@ -0,0 +1,13 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+namespace Microsoft.Testing.Platform.UI;
```
Evangelink: It seems it belongs to the `Helpers/System` folder, isn't it?

src/Platform/Microsoft.Testing.Platform/UI/NonAnsiTerminal.cs
```diff
@@ -0,0 +1,240 @@
...
+                string durationString = $" ({p.Stopwatch.Elapsed.TotalSeconds:F1}s)";
+
+                int passed = p.Passed;
+                int failed = p.Failed;
+                int skipped = p.Skipped;
+
+                string? detail = !RoslynString.IsNullOrWhiteSpace(p.Detail) ? $"- {p.Detail}" : null;
+                Append('[');
+                SetColor(TerminalColor.DarkGreen);
+                Append("✔️");
```
Evangelink: Could we have a green mark?

src/Platform/Microsoft.Testing.Platform/UI/NonAnsiTerminal.cs
```diff
@@ -0,0 +1,240 @@
...
+                Append(passed.ToString(CultureInfo.CurrentCulture));
```
Evangelink: From your screenshot, I find the icon too big and too close to the number

image

src/Platform/Microsoft.Testing.Platform/UI/LoggerOutcome.cs
```diff
@@ -0,0 +1,43 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+#if !NET7_0_OR_GREATER
+#endif
+#if !NET7_0_OR_GREATER
+#endif
+namespace Microsoft.Testing.Platform.UI;
```
Evangelink:
```suggestion

namespace Microsoft.Testing.Platform.UI;
```

src/Platform/Microsoft.Testing.Platform/UI/IMessage.cs
```diff
@@ -0,0 +1,11 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+namespace Microsoft.Testing.Platform.UI;
+
+/// <summary>
+/// Error or warning message that was sent to screen during the build.
+/// </summary>
+internal interface IMessage
```
Evangelink: Is it only for build? If that's the case maybe `IBuildMessage` would be more accurate.

src/Platform/Microsoft.Testing.Platform/UI/FileUtilities.cs
```diff
@@ -0,0 +1,37 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System.Diagnostics;
+
+namespace Microsoft.Testing.Platform.UI;
```
Evangelink: helpers namespace/folder.

Evangelink (commented):
I need more time but sending my first comments

src/Platform/Microsoft.Testing.Platform/UI/ConsoleLogger.cs
nohwnd: It is using regex source generator.

src/Platform/Microsoft.Testing.Platform/UI/AnsiTerminal.cs
nohwnd: It would also be useful for trx, logs, etc. I originally had it to check .dll and .exe  but that does not work on linux where files don't have to have any extension, including executables. 

And any way of figuring out if file is executable seemed too expensive, so I went with (currently pretty limited) whitelist.

src/Platform/Microsoft.Testing.Platform/UI/StopwatchAbstraction.cs
nohwnd: The abstraction is only useful for testing this logger. I would not move it among System, unless we want to use it else where / everywhere.

src/Platform/Microsoft.Testing.Platform/UI/NonAnsiTerminal.cs
nohwnd: ✅ yes, this with a box. This one is more consistently green across systems, but also the only one that is outlined. 

https://emojipedia.org/check-mark-button#designs

src/Platform/Microsoft.Testing.Platform/UI/NonAnsiTerminal.cs
nohwnd: This is why I originally used a simple `+-?` rather than emojis. We don't have any power over formatting the emojis, nor they are consistent across systems, or easily searchable in text (e.g. you have a long list of tests and you want to find failing, it is easier to write `/ -` than `/ ❌` especially when the emoji does not survive some glyph conversion in azdo raw output. 😄

src/Platform/Microsoft.Testing.Platform/UI/ConsoleLogger.cs
```diff
@@ -0,0 +1,701 @@
...
+            _originalConsoleMode = originalConsoleMode;
+            consoleWithProgress = consoleAcceptsAnsiCodes
+                ? new ConsoleWithProgress(new AnsiTerminal(console, _options.BaseDirectory), showProgress, updateEvery: 33)
+                : new ConsoleWithProgress(new NonAnsiTerminal(console), showProgress, updateEvery: 3_000);
+        }
+
+        _consoleWithProgress = consoleWithProgress;
+    }
+
+    public void TestExecutionStarted(DateTimeOffset testStartTime, int workerCount)
```
Evangelink: It's called from only one place with `1` for worker, do we need the parameter? What does it represent?

src/Platform/Microsoft.Testing.Platform/UI/IMessage.cs
nohwnd: It is not, it is during test run. The idea is that we can preserve failures in case we want to repeat them at the end of the output in an overview.

src/Platform/Microsoft.Testing.Platform/UI/ConsoleLoggerOutputDevice.cs
```diff
@@ -0,0 +1,511 @@
...
+                         assembly: _assemblyName,
+                         targetFramework: _targetFramework,
+                         architecture: _architecture,
+                         displayName: testNodeStateChanged.TestNode.DisplayName,
+                         outcome: LoggerOutcome.Error,
+                         duration: duration,
+                         errorMessage: errorState.Exception?.Message ?? errorState.Explanation,
+                         errorStackTrace: errorState.Exception?.StackTrace,
+                         expected: null,
+                         actual: null);
```
Evangelink: Arguments are not indented correctly (that's the case for all the other entries of the switch).

Evangelink: Also, I don't think the argument prefix is bringing anything (`duration: duration`, `targetFramework: _targetFramework`...).

src/Platform/Microsoft.Testing.Platform/UI/AnsiTerminal.cs
nohwnd: removed.

src/Platform/Microsoft.Testing.Platform/UI/AnsiCodes.cs
nohwnd: done

src/Platform/Microsoft.Testing.Platform/UI/AnsiTerminal.cs
nohwnd: image

src/Platform/Microsoft.Testing.Platform/UI/ConsoleLogger.cs
nohwnd: yes.

src/Adapter/MSTestAdapter.PlatformServices/Services/ThreadSafeStringWriter.cs
nohwnd: done

src/Platform/Microsoft.Testing.Platform/UI/AnsiTerminal.cs
nohwnd: done

src/Platform/Microsoft.Testing.Platform/UI/ConsoleLogger.cs
nohwnd: also fixed.

src/Platform/Microsoft.Testing.Platform/UI/ConsoleLogger.cs
nohwnd: done.

src/Platform/Microsoft.Testing.Platform/UI/ConsoleLogger.cs
nohwnd: yes.

src/Platform/Microsoft.Testing.Platform/UI/ConsoleLogger.cs
nohwnd: yes.

src/Platform/Microsoft.Testing.Platform/UI/ConsoleLogger.cs
nohwnd: moved to resource.

src/Platform/Microsoft.Testing.Platform/UI/NonAnsiTerminal.cs
nohwnd: replaced.

src/Platform/Microsoft.Testing.Platform/UI/ConsoleLogger.cs
nohwnd: This will be used by dotnet test when we have multiple projects running via this same logger, when there is the new protocol between dotnet test and test `.exe`.

src/Platform/Microsoft.Testing.Platform/UI/ConsoleLoggerOutputDevice.cs
nohwnd: formated removed in places where obvious value was provided.

src/Platform/Microsoft.Testing.Platform/UI/FileUtilities.cs
nohwnd: can this please wait once we connect it to dotnet test?

src/Platform/Microsoft.Testing.Platform/UI/ITerminal.cs
nohwnd: same here, but it was even weirder when there were x classes named "console". Keep as is for now?

src/Platform/Microsoft.Testing.Platform/UI/ConsoleLogger.cs
nohwnd: I am not sure which "console class" you mean, was it the original ConsoleOutputDevice?

src/Platform/Microsoft.Testing.Platform/UI/TestWorker.cs
nohwnd: Yes this is collecting state for a single "test worker" that is a process that is running tests and reporting them.

test/UnitTests/Microsoft.Testing.Platform.UnitTests/Microsoft.Testing.Platform.UnitTests.csproj
```diff
@@ -23,4 +23,8 @@
     <ProjectReference Include="$(RepoRoot)src\Platform\Microsoft.Testing.Platform\Microsoft.Testing.Platform.csproj" />
   </ItemGroup>
 
+  <ItemGroup>
+    <Folder Include="ConsoleOutputDevice\" />
+  </ItemGroup>
+
```
Evangelink: Un-needed?

```suggestion
```

src/Platform/Microsoft.Testing.Platform/Microsoft.Testing.Platform.csproj
```diff
@@ -129,5 +129,8 @@ This package provides the core platform and the .NET implementation of the proto
       <PackagePath>build/$(TargetFramework)</PackagePath>
     </TfmSpecificPackageFile>
   </ItemGroup>
+  <ItemGroup>
+    <Folder Include="UI\" />
+  </ItemGroup>
```
Evangelink: Needed?

```suggestion
```

nohwnd:
Moved back to draft: 

- [ ] implement TestHostControllerInfo.CurrentProcessIsTestHostController to disable progress reporting in controller.
- [ ] implement Console.Out that will respect when we are writing or not writing progress so we don't end up keeping progress on screen.

src/Platform/Microsoft.Testing.Platform/UI/ConsoleLogger.cs
Evangelink: Yes it was that (I couldn't remember the name). This is now merged so marking this as resovled.

src/Platform/Microsoft.Testing.Platform/OutputDevice/Console/TestWorker.cs
```diff
@@ -0,0 +1,35 @@
...
+        Failed = failed;
+        Skipped = skipped;
+        AssemblyName = assemblyName ?? throw new ArgumentNullException(nameof(assemblyName));
+        TargetFramework = targetFramework ?? throw new ArgumentNullException(nameof(targetFramework));
+        Architecture = architecture ?? throw new ArgumentNullException(nameof(architecture));
+        Stopwatch = stopwatch ?? throw new ArgumentNullException(nameof(stopwatch));
+        Detail = detail;
+    }
+
+    public int Passed { get; internal set; }
```
Evangelink: From what I see all these properties should be readonly.

src/Platform/Microsoft.Testing.Platform/OutputDevice/Console/TestWorker.cs
```diff
@@ -0,0 +1,35 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+namespace Microsoft.Testing.Platform.OutputDevice.Console;
+
+internal class TestWorker
```
Evangelink: It's not doing any "work" so I'd prefer something like `TestProgressState`.

src/Platform/Microsoft.Testing.Platform/OutputDevice/Console/TestWorkerFrame.cs
```diff
@@ -0,0 +1,137 @@
...
+
+        terminal.SetColor(TerminalColor.DarkRed);
+        terminal.Append("❌");
+        terminal.Append(failed.ToString(CultureInfo.CurrentCulture));
+        terminal.ResetColor();
+
+        terminal.Append("/");
+
+        terminal.SetColor(TerminalColor.DarkYellow);
+        terminal.Append("❔");
```
Evangelink: Not sure where you took the icons from, do we have something that would better represent skipped? Question mark is more of something unsure to me. Maybe something like https://glebbahmutov.com/blog/images/rt/skip1.png

src/Platform/Microsoft.Testing.Platform/OutputDevice/Console/TestWorkerFrame.cs
```diff
@@ -0,0 +1,137 @@
...
+        terminal.Append(p.Architecture);
+        terminal.Append(')');
+        if (!RoslynString.IsNullOrWhiteSpace(detail))
+        {
+            terminal.Append(" - ");
+            terminal.Append(detail);
+        }
+
+        terminal.SetCursorHorizontal(Width - durationString.Length);
+        terminal.Append(durationString);
```
Evangelink: Maybe a clock icon for duration could be good too.

src/Platform/Microsoft.Testing.Platform/UI/TestWorker.cs
Evangelink: I don't know what the others think but this is really "just" a state object so `TestProgressState` seems best to me.

src/Platform/Microsoft.Testing.Platform/UI/StopwatchAbstraction.cs
Evangelink: Could be globally useful but you are right, this is only used here for now.

src/Platform/Microsoft.Testing.Platform/UI/NonAnsiTerminal.cs
Evangelink: I am not sure what JS is doing but most of their frameworks have a nice CLI output:

https://designmodo.com/wp-content/uploads/2016/06/01-javascript-console-code.jpg
https://th.bing.com/th/id/R.8663b3da0350d2e851969f90aa850bf4?rik=Z19u9Vv%2brzeIcg&riu=http%3a%2f%2fdesignmodo.com%2fwp-content%2fuploads%2f2016%2f06%2f01-javascript-console-code.jpg&ehk=z7XaC0qbkaxKTg%2bqcp64ueGR25lfM4k8oMvwjF%2bCyCc%3d&risl=&pid=ImgRaw&r=0

src/Platform/Microsoft.Testing.Platform/UI/IMessage.cs
Evangelink: In that case, the comment should be updated, I would maybe still go with `IProgressMessage` or `IResultMessage`.

src/Platform/Microsoft.Testing.Platform/UI/FileUtilities.cs
Evangelink: As far as I can see, you did it, so resolving.

src/Platform/Microsoft.Testing.Platform/UI/ConsoleLogger.cs
Evangelink: Is it responsibility of the caller or this class to know how many state object should handle the progress?

src/Platform/Microsoft.Testing.Platform/OutputDevice/Console/NonAnsiTerminal.cs
```diff
@@ -0,0 +1,240 @@
...
+
+    public void Append(char value)
+    {
+        if (_isBatching)
+        {
+            _stringBuilder.Append(value);
+        }
+        else
+        {
+            _console.Write(value.ToString());
```
Evangelink: The system console has an overload with `char`, so it's better to add it to `IConsole` and to use it here instead of forcing call to `ToString()`.

src/Platform/Microsoft.Testing.Platform/OutputDevice/Console/AnsiTerminal.cs
```diff
@@ -0,0 +1,321 @@
...
+
+    public void Append(char value)
+    {
+        if (_isBatching)
+        {
+            _stringBuilder.Append(value);
+        }
+        else
+        {
+            _console.Write(value.ToString());
```
Evangelink: The system console has an overload with `char`, so it's better to add it to `IConsole` and to use it here instead of forcing call to `ToString()`.

src/Platform/Microsoft.Testing.Platform/OutputDevice/Console/LoggerArtifact.cs
```diff
@@ -0,0 +1,9 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+namespace Microsoft.Testing.Platform.OutputDevice.Console;
+
+/// <summary>
+/// An artifact / attachment that was reported during run.
+/// </summary>
+internal record class LoggerArtifact(bool OutOfProcess, string? Assembly, string? TargetFramework, string? Architecture, string? TestName, string Path);
```
Evangelink: What about `TestRunArtifact`? In the core platform, we make the difference between an artifact attached to a test and a "global" artifact, it might be interesting to keep the same logic.

src/Platform/Microsoft.Testing.Platform/OutputDevice/Console/LoggerOutcome.cs
```diff
@@ -0,0 +1,40 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+namespace Microsoft.Testing.Platform.OutputDevice.Console;
+
+/// <summary>
+/// Outcome of a test.
+/// </summary>
+internal enum LoggerOutcome
```
Evangelink, line 9: Given what we are doing with it, I think we need only the 3 states (passed, failed, skipped).

src/Platform/Microsoft.Testing.Platform/OutputDevice/Console/ConsoleLogger.cs
```diff
@@ -0,0 +1,702 @@
...
+
+        if (duration.TotalMilliseconds < 1000)
+        {
+            terminal.Append(duration.TotalMilliseconds.ToString("F0", CultureInfo.CurrentCulture));
+            terminal.Append("ms");
+        }
+        else
+        {
+            terminal.Append(duration.TotalSeconds.ToString("F1", CultureInfo.CurrentCulture));
+            terminal.Append('s');
```
Evangelink: What about minutes and hours? We know it's possible to have such "slow" tests

src/Platform/Microsoft.Testing.Platform/OutputDevice/Console/ConsoleLogger.cs
```diff
@@ -0,0 +1,702 @@
...
+        {
+            return;
+        }
+
+        TerminalColor color = outcome switch
+        {
+            LoggerOutcome.Error or LoggerOutcome.Fail or LoggerOutcome.Cancelled or LoggerOutcome.Timeout => TerminalColor.DarkRed,
+            LoggerOutcome.Skipped => TerminalColor.DarkYellow,
+            LoggerOutcome.Passed => TerminalColor.DarkGreen,
+            _ => throw new NotImplementedException(),
```
Evangelink: We are using `NotSupportedException` in the rest of the codebase.

src/Platform/Microsoft.Testing.Platform/OutputDevice/Console/ConsoleLogger.cs
```diff
@@ -0,0 +1,702 @@
...
+        };
+        string outcomeText = outcome switch
+        {
+            LoggerOutcome.Fail or LoggerOutcome.Error => PlatformResources.FailedLowercase,
+            LoggerOutcome.Skipped => PlatformResources.SkippedLowercase,
+            LoggerOutcome.Cancelled or LoggerOutcome.Timeout => $"{PlatformResources.FailedLowercase} ({PlatformResources.CancelledLowercase})",
+            LoggerOutcome.Passed => PlatformResources.PassedLowercase,
+            _ => throw new NotImplementedException(),
```
Evangelink: We are using `NotSupportedException `in the rest of the codebase.

src/Platform/Microsoft.Testing.Platform/OutputDevice/Console/ConsoleLogger.cs
```diff
@@ -0,0 +1,702 @@
...
+        };
+
+        terminal.SetColor(color);
+        terminal.Append(outcomeText);
+        terminal.ResetColor();
+        terminal.Append(' ');
+        terminal.Append(displayName);
+        terminal.SetColor(TerminalColor.Gray);
+        terminal.Append(' ');
+        terminal.Append($"({duration.TotalSeconds:F1}s)");
```
Evangelink: We may want to reuse the helper here (that takes care of ms, min...)

src/Platform/Microsoft.Testing.Platform/OutputDevice/Console/ConsoleLogger.cs
```diff
@@ -0,0 +1,702 @@
...
+        if (_options.ShowAssembly)
+        {
+            terminal.AppendLine();
+            terminal.Append(SingleIndentation);
+            terminal.Append("from ");
```
Evangelink: Resx?

src/Platform/Microsoft.Testing.Platform/OutputDevice/Console/ConsoleLogger.cs
```diff
@@ -0,0 +1,702 @@
...
+    }
+
+    private static void AppendStackFrame(ITerminal terminal, string stackTraceLine)
+    {
+        terminal.Append(DoubleIndentation);
+        Match match = GetFrameRegex().Match(stackTraceLine);
+        if (match.Success)
+        {
+            terminal.SetColor(TerminalColor.Gray);
+            terminal.Append("at ");
```
Evangelink: Resx?

src/Platform/Microsoft.Testing.Platform/OutputDevice/Console/ConsoleLogger.cs
```diff
@@ -0,0 +1,702 @@
...
+            terminal.ResetColor();
+            terminal.SetColor(TerminalColor.Red);
+            terminal.Append(match.Groups["code"].Value);
+            terminal.SetColor(TerminalColor.Gray);
+            terminal.Append(" in ");
```
Evangelink: Resx?

src/Platform/Microsoft.Testing.Platform/OutputDevice/Console/ConsoleLogger.cs
```diff
@@ -0,0 +1,702 @@
...
+        _consoleWithProgress.WriteToTerminal(terminal =>
+        {
+            terminal.SetColor(TerminalColor.Yellow);
+            terminal.AppendLine(text);
+            terminal.ResetColor();
+        });
+    }
+
+    internal void WriteErrorMessage(string assembly, string? targetFramework, string? architecture, Exception exception) =>
+        WriteErrorMessage(assembly, targetFramework, architecture, exception.ToString());
```
Evangelink: Nit: for consistency with the other methods

```suggestion
    internal void WriteErrorMessage(string assembly, string? targetFramework, string? architecture, Exception exception)
        => WriteErrorMessage(assembly, targetFramework, architecture, exception.ToString());
```

src/Platform/Microsoft.Testing.Platform/UI/NonAnsiTerminal.cs
nohwnd: They are using x and a different glyph: 

image

src/Platform/Microsoft.Testing.Platform/UI/ConsoleLogger.cs
nohwnd: The caller tells us what is the level of parallelism, so we can know how many projects will run at the same time at max, and make UI choices based on that. e.g. we could write: "Starting run in parallel on 10 workers".

Actually managing the states is the responsibility of this console with progress.

src/Platform/Microsoft.Testing.Platform/OutputDevice/Console/TestWorkerFrame.cs
nohwnd: reverted the glyphs to not be emojis, maybe an arrow would work, but  the similar arrow I found has this issue filed: https://github.com/microsoft/vscode-python/issues/7705 for not being easily visible.

src/Platform/Microsoft.Testing.Platform/OutputDevice/Console/TestWorkerFrame.cs
nohwnd: If you want to go back to emojis fully I can add it. But to me the emojis steal too much focus in the output.

src/Platform/Microsoft.Testing.Platform/OutputDevice/Console/NonAnsiTerminal.cs
nohwnd: done

src/Platform/Microsoft.Testing.Platform/OutputDevice/Console/LoggerArtifact.cs
nohwnd: Renamed.

src/Platform/Microsoft.Testing.Platform/OutputDevice/Console/ConsoleLogger.cs
nohwnd: I used the standard human readable rendering we use in other places.

src/Platform/Microsoft.Testing.Platform/OutputDevice/Console/ConsoleLogger.cs
nohwnd: done.

src/Platform/Microsoft.Testing.Platform/OutputDevice/Console/ConsoleLogger.cs
nohwnd: yes. even though stack traces are not localized in .net "core"

src/Platform/Microsoft.Testing.Platform/OutputDevice/Console/ConsoleLogger.cs
nohwnd: yes. even though stack traces are not localized in .net "core"

src/Platform/Microsoft.Testing.Platform/OutputDevice/Console/ConsoleLogger.cs
nohwnd: did the change in place.

src/Platform/Microsoft.Testing.Platform/UI/IMessage.cs
nohwnd: renamed to IProgressMessage

src/Platform/Microsoft.Testing.Platform/OutputDevice/Console/TestWorkerFrame.cs
nohwnd: image

src/Platform/Microsoft.Testing.Platform/OutputDevice/Console/TestProgressState.cs
```diff
@@ -0,0 +1,35 @@
...
+namespace Microsoft.Testing.Platform.OutputDevice.Console;
+
+internal class TestProgressState
+{
+    public TestProgressState(int passed, int failed, int skipped, string assemblyName, string? targetFramework, string? architecture, StopwatchAbstraction stopwatch, string? detail)
+    {
+        Passed = passed;
+        Failed = failed;
+        Skipped = skipped;
+        AssemblyName = assemblyName ?? throw new ArgumentNullException(nameof(assemblyName));
```
Evangelink: `assemblyName` is declared non-nullable. If you are doing validation because this can be called by something/someone "external", we should also validate the ints to be postiive.

src/Platform/Microsoft.Testing.Platform/OutputDevice/Console/TestProgressState.cs
```diff
@@ -0,0 +1,35 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+namespace Microsoft.Testing.Platform.OutputDevice.Console;
+
+internal class TestProgressState
+{
+    public TestProgressState(int passed, int failed, int skipped, string assemblyName, string? targetFramework, string? architecture, StopwatchAbstraction stopwatch, string? detail)
```
Evangelink, line 8: Any reason to have tfm and architecture nullable and to throw when null? It's better to mark as non-nullable.

src/Platform/Microsoft.Testing.Platform/Resources/PlatformResources.resx
```diff
@@ -553,4 +553,35 @@ Read more about Microsoft Testing Platform telemetry: https://aka.ms/testingplat
...
+    <comment>from followed by a file name to point to the file from which test is originating</comment>
+  </data>
+  <data name="StackFrameAt" xml:space="preserve">
+    <value>at</value>
+    <comment>at that is used for a stack frame location in a stack trace, is followed by a class and method name</comment>
+  </data>
+  <data name="StackFrameIn" xml:space="preserve">
+    <value>in</value>
+    <comment>in that is used in stack frame it is followed by file name</comment>
+  </data>
```
Evangelink, line 586: Not sure what are the localization rule here at MS but I recall that in past, I got feedback from localization teams that we should not split sentences as we do in English as the setup/order of the full sentence can be different based on languages. It was recommended to have full sentence localized.

src/Platform/Microsoft.Testing.Platform/TestHostControllers/ITestHostControllerInfo.cs
```diff
@@ -7,5 +7,11 @@ internal interface ITestHostControllerInfo
 {
     bool HasTestHostController { get; }
 
+    /// <summary>
+    /// Gets information whether the current process is a test controller or not.
+    /// When null the value has not been set yet, it is figured out while we are building the test application.
+    /// </summary>
+    bool? CurrentProcessIsTestHostController { get; }
```
Evangelink: I like the design of having boolean properties prefixed with `Is`, `Can`, `Has`, maybe here we should go with `IsCurrentProcessTestHostController`?

test/IntegrationTests/Microsoft.Testing.Platform.Acceptance.IntegrationTests/Helpers/AcceptanceAssert.cs
```diff
@@ -69,6 +69,30 @@ public static void AssertOutputRegEx(this DotnetMuxerResult dotnetMuxerResult, s
     public static void AssertOutputDoesNotContain(this TestHostResult testHostResult, string value, [CallerMemberName] string? callerMemberName = null, [CallerFilePath] string? callerFilePath = null, [CallerLineNumber] int callerLineNumber = 0)
         => Assert.That(!testHostResult.StandardOutput.Contains(value, StringComparison.Ordinal), GenerateFailedAssertionMessage(testHostResult), callerMemberName: callerMemberName, callerFilePath: callerFilePath, callerLineNumber: callerLineNumber);
 
+    public static void AssertOutputContainsSummary(this TestHostResult testHostResult, int failed, int passed, int skipped, bool? aborted = false, int? minimumNumberOfTests = null, [CallerMemberName] string? callerMemberName = null, [CallerFilePath] string? callerFilePath = null, [CallerLineNumber] int callerLineNumber = 0)
+    {
+        int totalTests = failed + passed + skipped;
+        string result = minimumNumberOfTests != null && totalTests < minimumNumberOfTests
+            ? $"Minimum expected tests policy violation, tests ran {totalTests}, minimum expected {minimumNumberOfTests}"
+            : aborted is not null and true
```
Evangelink: Isn't this enough?

```suggestion
            : aborted is true
```

test/Utilities/Microsoft.Testing.TestInfrastructure/CommandLine.cs
```diff
@@ -112,5 +112,17 @@ public static int MaxOutstandingCommands
         }
     }
 
+    /// <summary>
+    /// Depending on command line settings, e.g. when using Windows Terminal
+    /// .NET Framework app might have BOM at the beginning of the captured output, which breaks output comparisons
+    /// while no visible difference is seen between the outputs.
+    /// </summary>
+    private string ClearBOM(string outputLine)
```
Evangelink: Nit:

```suggestion
    private static string ClearBOM(string outputLine)
```

src/Platform/Microsoft.Testing.Platform/OutputDevice/Console/TestProgressFrame.cs
```diff
@@ -0,0 +1,137 @@
...
+        }
+
+        // clear no longer used lines
+        if (i < previousFrame.ProgressCount)
+        {
+            terminal.Append($"{AnsiCodes.CSI}{AnsiCodes.EraseInDisplay}");
+        }
+    }
+
+    public void Clear() => ProgressCount = 0;
```
Evangelink, line 154: Shall we also reset `_progressItems`?

src/Platform/Microsoft.Testing.Platform/OutputDevice/Console/TestModule.cs
```diff
@@ -0,0 +1,48 @@
...
+
+    public string? TargetFramework { get; }
+
+    public string? Architecture { get; }
+
+    public StopwatchAbstraction Stopwatch { get; }
+
+    public List<string> Attachments { get; } = new();
+
+    public List<IProgressMessage> Messages { get; } = new();
```
Evangelink, line 29: Could we expose as "readonly"/"immutable" to avoid having anything that can mutate this?

src/Platform/Microsoft.Testing.Platform/OutputDevice/Console/ConsoleLogger.cs
```diff
@@ -0,0 +1,742 @@
...
+    internal const string DoubleIndentation = $"{SingleIndentation}{SingleIndentation}";
+
+    internal const string TripleIndentation = $"{SingleIndentation}{SingleIndentation}{SingleIndentation}";
+
+    internal Func<StopwatchAbstraction> CreateStopwatch { get; set; } = SystemStopwatch.StartNew;
+
+    /// <summary>
+    /// The two directory separator characters to be passed to methods like <see cref="string.IndexOfAny(char[])"/>.
+    /// </summary>
+    private readonly Dictionary<string, TestModule> _assemblies = new();
```
Evangelink, line 38: Shall we handle parallelism?

Evangelink (commented):
We still need to get rid of all the logger wording

src/Platform/Microsoft.Testing.Platform/Resources/PlatformResources.resx
nohwnd: Yes that is true, but the stack frames are normally not localized at all unless you are on .net framework. And these words are on one line either as 

**at** code line **in** file

**from** file

So they are not part of a bigger sentence. and .net framework has the same kind of resources for those two words, Word_At and Word_In.

src/Platform/Microsoft.Testing.Platform/TestHostControllers/ITestHostControllerInfo.cs
nohwnd: Renamed.

src/Platform/Microsoft.Testing.Platform/Resources/PlatformResources.resx
Evangelink: Perfect then! Thanks for checking @nohwnd <3

src/Platform/Microsoft.Testing.Platform/OutputDevice/Console/TestProgressState.cs
nohwnd: No, it is autogenerated code oversight. Fixed.

src/Platform/Microsoft.Testing.Platform/OutputDevice/Console/TestProgressState.cs
nohwnd: also code that was generated and later fixed to be non nullable. Fixed.

src/Platform/Microsoft.Testing.Platform/OutputDevice/Console/TestProgressFrame.cs
nohwnd: We could, but we will just spend more time cleaning up the array even though we will never touch it again. Keeping as is.

src/Platform/Microsoft.Testing.Platform/OutputDevice/Console/ConsoleLogger.cs
nohwnd: No the invocations are coming from the message bus which is serial.

src/Platform/Microsoft.Testing.Platform/OutputDevice/Console/TestModule.cs
nohwnd: It can be done, but it feels like too much ceremony for this simple internal object that just holds the state and is not passed outside of the logger.

Evangelink (approved):
Merging as-is. We will do a follow-up for the last few nits

PR #3299, opened 2024-07-22
move BuiltInTypeNames to frozen dictionary

SimonCropp:


SimonCropp:
yep understood. and i suspected that might be the answer.

r u happy for me to resubmit with conditional compilation so it uses frozon in net 8 and higher?

Evangelink:
I think that would be fine (@MarcoRossignoli?)

PR #3300, opened 2024-07-22
Fix MessageBusProxy.InitAsync to be proxying _messageBus.InitAsync

SimonCropp:


Evangelink:
@MarcoRossignoli can you check this one please?

MarcoRossignoli (approved):
Yep, we're not using it at the moment but it should.
We init the concrete message bus where we create the inner one so we never found this potential bug.

Evangelink (approved):

SimonCropp:
so it would have been a stack overflow?

PR #3306, opened 2024-07-22
Fix shutdown order for server mode

MarcoRossignoli:
In server mode we cannot send back logging information to VS after the "exit". 

cc: @drognanar


Evangelink (approved):
I am assuming that would be hard to safe guard with some test?

MarcoRossignoli:
> I am assuming that would be hard to safe guard with some test?

We cannot I did it manually simulating/reproing a throw new timeout.

PR #3310, opened 2024-07-23
 _factories.Count cant be 0

SimonCropp:


Evangelink (commented):
I am not sure this is functionally wanted, cc @MarcoRossignoli

SimonCropp:
my point is, based on the code, it cant be empty. so it is dead code

PR #3311, opened 2024-07-23
remove list duplication to arrays

SimonCropp:


SimonCropp:
will try again in smaller iterations

PR #3314, opened 2024-07-23
Feat: allow cooperative timeout

### LINKED ISSUE #2432: Timeout features should allow for a cooperative only mode
In the current design of the timeout feature, the methods marked with, and supporting, the timeout  attribute when timeout expires will produce the following behaviors:
1. the cancellation token linked to the related `TestContext` will be triggered
2. the activity (thread or task for current implementation) of running the method will be unobserved resulting in a dangling activity.

The first part, cooperative cancellation, is virtuous but the second part, dangling activity, is not recommended and could lead to more problems (from CPU/Memory usage for nothing to global test app/suite being in some unknown state).

We cannot simply break the second behavior has it would required many changes from users but we should encourage the new behavior. To do so, I suggest the following course of action:

1. In release v3.4 of MSTest, we will:
    - introduce a new property on timeout attribute to allow/disallow non-cooperative mode on the given method
    - introduce a runsettings entry to allow/disallow non-cooperative mode globally (the timeout property would override the global setting)
    - introduce an analyzer that flags cases where the non-cooperative mode is allowed

1. In release v4 of MSTest, we will do a breaking change disallowing the non-cooperative mode by default.

Evangelink:
The current implementation of the timeout will create a "thread" that will observe the execution of the step (fixture or test method) for a given time (timeout value) and will stop observing after that. This means that we are producing dangling tasks where code continues to execute and possibly mutate state when we move to the next step which can have many unwanted consequences.

With this PR, we introduce a virtuous mode that ressemble what user has to do when implementing cooperative cancellation in their production code. 

/!\ It's important to note that when opting-in to this mode, if the method (fixture or test) is ignoring the token then the method will continue.

Fixes #2432

### DIFF
--- a/src/Adapter/MSTest.TestAdapter/Helpers/FixtureMethodRunner.cs
+++ b/src/Adapter/MSTest.TestAdapter/Helpers/FixtureMethodRunner.cs
@@ -5,6 +5,7 @@ using System.Globalization;
 using System.Reflection;
 using System.Runtime.InteropServices;
 
+using Microsoft.VisualStudio.TestPlatform.MSTest.TestAdapter.Execution;
 using Microsoft.VisualStudio.TestPlatform.MSTest.TestAdapter.ObjectModel;
 using Microsoft.VisualStudio.TestPlatform.MSTestAdapter.PlatformServices;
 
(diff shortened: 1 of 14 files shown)


test/IntegrationTests/MSTest.Acceptance.IntegrationTests/Properties/launchSettings.json
```diff
@@ -2,7 +2,7 @@
   "profiles": {
     "MSTest.Acceptance.IntegrationTests": {
       "commandName": "Project",
-      "commandLineArgs": "--treenode-filter /*/*/*/**"
+      "commandLineArgs": "--treenode-filter /*/*/InitializeAndCleanupTimeout/CooperativeCancellation*/**"
```
MarcoRossignoli: cleanup

src/Adapter/MSTest.TestAdapter/Helpers/FixtureMethodRunner.cs
```diff
@@ -15,21 +16,76 @@ namespace Microsoft.VisualStudio.TestPlatform.MSTest.TestAdapter.Helpers;
...
-        if (timeout is null)
+        if (timeoutInfo is null)
         {
             action();
             return null;
         }
 
+        if (timeoutInfo.Value.CooperativeCancellation)
+        {
+            return RunWithCooperativeCancellation(action, cancellationTokenSource, timeoutInfo.Value, methodInfo, methodCancelledMessageFormat, methodTimedOutMessageFormat);
```
MarcoRossignoli, line 32: should this wrap in some way RunWithTimeoutAndCancellationWithSTAThread and RunWithTimeoutAndCancellationWithThreadPool ?

Evangelink: WDYM? I don't need to start a task/thread because I'll run the action synchronously as I don't need to "unobserve" it.

test/IntegrationTests/MSTest.Acceptance.IntegrationTests/Properties/launchSettings.json
Evangelink:
```suggestion
      "commandLineArgs": "--treenode-filter /*/*/*/**"
```

src/TestFramework/TestFramework/Attributes/TestMethod/TimeoutAttribute.cs
```diff
@@ -35,4 +37,21 @@ public TimeoutAttribute(TestTimeout timeout)
...
+    /// Gets or sets a value indicating whether the method should be cooperatively cancelled on timeout.
+    /// When set to <see langword="true"/>, the method should be designed to cooperatively cancel itself when the timeout is reached.
+    /// Otherwise, when set to <see langword="false"/>, the task method will be unobserved.
+    /// </summary>
+    public bool CooperativeCancellation
+    {
+        get => _isCooperativeCancellation;
+        set
+        {
+            IsCooperativeCancellationSet = true;
```
Evangelink, line 56: Nullable boolean is not supported on attributes hence the trick with this other property to know this was set.

nohwnd:
```suggestion
            // Attributes don't allow nullable boolean, so we need this to know that the value was set explicitly.
            IsCooperativeCancellationSet = true;
```

src/TestFramework/TestFramework/Attributes/TestMethod/TimeoutAttribute.cs
```diff
@@ -35,4 +37,21 @@ public TimeoutAttribute(TestTimeout timeout)
     /// Gets the timeout in milliseconds.
     /// </summary>
     public int Timeout { get; }
+
+    /// <summary>
+    /// Gets or sets a value indicating whether the method should be cooperatively cancelled on timeout.
+    /// When set to <see langword="true"/>, the method should be designed to cooperatively cancel itself when the timeout is reached.
+    /// Otherwise, when set to <see langword="false"/>, the task method will be unobserved.
+    /// </summary>
```
nohwnd:
```suggestion
    /// Gets or sets a value indicating whether the test method should be cooperatively cancelled on timeout. 
    /// When set to <see langword="true"/>, the cancellation token is cancelled on timeout, and the method completion is awaited. The test method and all the code it calls, must be designed in a way that it observes the cancellation and cancels cooperatively. If the test method does not complete, the timeout does not force it to complete.
    /// When set to <see langword="false"/>, the cancellation token is cancelled on timeout, timeout result is reported and the method task will continue running on background. This may lead to conflicts in file access on test cleanup, unobserved exceptions, and memory leaks.
    /// </summary>
```

src/Adapter/MSTest.TestAdapter/Helpers/FixtureMethodRunner.cs
MarcoRossignoli: so we're already in the correct thread context at this point? Like if the thread is STA are we running cooperatively in a STA context?

Evangelink: Yes! We are always in the right context before joining this method. We needed to do a lot of extra work (starting a different thread/task + moving the execution context only because we want to be able to stop observing the execution).

Evangelink: But that's a good point, let me update the STA tests to confirm that!

MarcoRossignoli: kk

nohwnd (approved):

PR #3330, opened 2024-07-24
Fix spelling issues

Evangelink:
I have installed on my VS https://marketplace.visualstudio.com/items?itemName=EWoodruff.VisualStudioSpellCheckerVS2022andLater, configured it and fixed some of the issues found.


src/Adapter/MSTest.TestAdapter/Resources/Resource.resx
```diff
@@ -364,13 +364,13 @@ Error: {1}</value>
     <value>Assembly initialize method '{0}.{1}' timed out after {2}ms</value>
   </data>
   <data name="AssemblyInitializeWasCancelled" xml:space="preserve">
-    <value>Assembly initialize method '{0}.{1}' was cancelled</value>
+    <value>Assembly initialize method '{0}.{1}' was canceled</value>
```
Evangelink, line 367: It seems that the version with 2 L is British and that American prefer the one L

Issue #3020, opened 2024-05-29
should settings parsing ignore invalid values?


SimonCropp:
should settings parsing ignore invalid values?

![image](https://github.com/microsoft/testfx/assets/122666/28c6b354-57bc-47f9-bbb9-2e8f76ab46ac)

AB#2112231

Evangelink:
It would definitely be better if we were printing a warning in case of invalid entries. 

We could also have another setting (disabled by default for now) to make it fail/stop test execution.

nohwnd:
[like]  Jakub Jares reacted to your message:
________________________________

PR #3338, opened 2024-07-25
Fix settings parsing ignore invalid values

engyebrahim:
fix: https://github.com/microsoft/testfx/issues/3020

### DIFF
--- a/src/Adapter/MSTest.TestAdapter/MSTestSettings.cs
+++ b/src/Adapter/MSTest.TestAdapter/MSTestSettings.cs
@@ -246,33 +246,7 @@ public class MSTestSettings
     /// The discovery context that contains the runsettings.
     /// </param>
     [Obsolete("this function will be removed in v4.0.0")]
-    public static void PopulateSettings(IDiscoveryContext? context)
-    {
-        RunConfigurationSettings = RunConfigurationSettings.PopulateSettings(context);
-
-        if (context?.RunSettings == null || StringEx.IsNullOrEmpty(context.RunSettings.SettingsXml))
-        {
-            // This will contain default adapter settings
-            CurrentSettings = new MSTestSettings();
-            return;
-        }
-
-        MSTestSettings? aliasSettings = GetSettings(context.RunSettings.SettingsXml, SettingsNameAlias, null);
-
-        // If a user specifies MSTestV2 in the runsettings, then prefer that over the v1 settings.
-        if (aliasSettings != null)
-        {
-            CurrentSettings = aliasSettings;
-        }
-        else
-        {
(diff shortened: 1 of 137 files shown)


src/Adapter/MSTest.TestAdapter/MSTestSettings.cs
```diff
@@ -245,6 +245,7 @@ public static void PopulateSettings(MSTestSettings settings)
     /// <param name="context">
     /// The discovery context that contains the runsettings.
     /// </param>
+    [Obsolete("this function will be removed in v4.0.0")]
     public static void PopulateSettings(IDiscoveryContext? context)
```
Evangelink: You could reduce duplicate code by having this method calling `PopulateSettings(context ,null)`

src/Adapter/MSTest.TestAdapter/Resources/Resource.resx
```diff
@@ -407,4 +407,7 @@ Error: {1}</value>
 Test expected {2} parameter(s), with types '{3}',
 but received {4} argument(s), with types '{5}'.</value>
   </data>
+  <data name="InvalidValue" xml:space="preserve">
+    <value>Invalid Value for '{0}', couldn't parse it.</value>
```
Evangelink: Maybe something like `Invalid value '{0}' for runsettings entry '{1}', setting will be ignored`

Evangelink (commented):
I would expect either to edit/add tests for invalid values (1 invalid value per entry is enough if you are adding them) for the various entries.

test/UnitTests/MSTestAdapter.UnitTests/MSTestSettingsTests.cs
```diff
@@ -85,11 +85,44 @@ public void MapInconclusiveToFailedShouldBeConsumedFromRunSettingsWhenSpecified(
...
+                    <TestTimeout>timeout</TestTimeout>
+                    <TreatClassAndAssemblyCleanupWarningsAsErrors>3</TreatClassAndAssemblyCleanupWarningsAsErrors>
+                    <ForcedLegacyMode>3</ForcedLegacyMode>
+               </MSTestV2>
+            </RunSettings>
+            """;
+
+        var adapterSettings = MSTestSettings.GetSettings(runSettingxml, MSTestSettings.SettingsNameAlias, _mockMessageLogger.Object);
+
+        _mockMessageLogger.Verify(lm => lm.SendMessage(TestMessageLevel.Warning, It.IsAny<string>()), Times.Exactly(18));
```
Evangelink, line 123: Please do check the messages

Evangelink (approved):

PR #3340, opened 2024-07-25
Add  the `DotnetTestDataConsumer` for the test host controller

MarcoRossignoli:
Followup of https://github.com/microsoft/testfx/pull/3317

cc: @mariam-abdulla

### DIFF
--- a/src/Platform/Microsoft.Testing.Platform/Hosts/TestHostControllersTestHost.cs
+++ b/src/Platform/Microsoft.Testing.Platform/Hosts/TestHostControllersTestHost.cs
@@ -30,6 +30,7 @@ internal sealed class TestHostControllersTestHost : CommonTestHost, ITestHost, I
     private readonly TestHostControllerConfiguration _testHostsInformation;
     private readonly IEnvironment _environment;
     private readonly IClock _clock;
+    private readonly NamedPipeClient? _dotnetTestPipeClient;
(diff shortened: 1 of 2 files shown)


src/Platform/Microsoft.Testing.Platform/Hosts/TestHostControllersTestHost.cs
```diff
@@ -129,6 +131,12 @@ protected override async Task<int> InternalRunAsync()
                 dataConsumersBuilder.Add(dataConsumerDisplay);
             }
 
+            // We register the DotnetTestDataConsumer as last to ensure that it will be the last one to consume the data.
```
Evangelink, line 134: Do we ensure that registration order is matching reception order?

MarcoRossignoli: yep the message bus build in order.

mariam-abdulla (approved):

src/Platform/Microsoft.Testing.Platform/Hosts/TestHostControllersTestHost.cs
MarcoRossignoli: here anyway is less than important anyway we wait that all queue are empty before to move on drain.

PR #3342, opened 2024-07-25
Simplify pipelines

Evangelink:


azure-pipelines-official.yml
```diff
@@ -147,7 +132,14 @@ extends:
             - script: eng\common\CIBuild.cmd
                 -configuration $(_BuildConfig)
                 -prepareMachine
-                $(_InternalBuildArgs)
+                /p:DotNetSignType=real
```
ViktorHofer: That is already the default setting

azure-pipelines-official.yml
```diff
@@ -147,7 +132,14 @@ extends:
...
+                /p:TeamName=$(_TeamName)
+                /p:DotNetFinalVersionKind=$(_ReleaseVersionKind)
+                /p:DotNetPublishUsingPipelines=true
+                /p:DotNetSymbolServerTokenMsdl=$(microsoft-symbol-server-pat)
+                /p:DotNetSymbolServerTokenSymWeb=$(symweb-symbol-server-pat)
```
ViktorHofer, line 139: @mmitche are these two still required?

azure-pipelines-official.yml
Evangelink:
```suggestion
```

Issue #3223, opened 2024-07-08
MSTEST0010: report when class is `abstract` and inheritance is not set


Evangelink:
## Summary

Report when an `abstract` class declares a member marked with `[ClassInitialize]` without specifying the inheritance behavior.

## Background and Motivation

When an `abstract` class is declaring a `ClassInitialize` member without specifying the inheritance behavior, then this member will simply be ignored so we should flag the case.

## Proposed Feature

Scenarios to flag:

```cs
[TestClass]
public abstract class BaseClass
{
    [ClassInitialize]
    public static void ClassInit(TestContext context)
    {
    }
}
```

```cs
public abstract class BaseClass
{
    [ClassInitialize]
    public static void ClassInit(TestContext context)
    {
    }
}
```

Valid scenario:

```cs
[TestClass]
public abstract class BaseClass
{
    [ClassInitialize(Microsoft.VisualStudio.TestTools.UnitTesting.InheritanceBehavior.BeforeEachDerivedClass)]
    public static void ClassInit(TestContext context)
    {
    }
}
```
or 

```cs
public abstract class BaseClass
{
    [ClassInitialize(Microsoft.VisualStudio.TestTools.UnitTesting.InheritanceBehavior.BeforeEachDerivedClass)]
    public static void ClassInit(TestContext context)
    {
    }
}
```

PR #3347, opened 2024-07-26
MSTEST0010: report when class is abstract and inheritance is not set

engyebrahim:
fix: https://github.com/microsoft/testfx/issues/3223

### DIFF
--- a/src/Analyzers/MSTest.Analyzers/Resources.resx
+++ b/src/Analyzers/MSTest.Analyzers/Resources.resx
@@ -197,8 +197,8 @@
 - take a single parameter of type 'TestContext'
 - return type should be 'void', 'Task' or 'ValueTask'
 - not be 'async void'
-- not be a special method (finalizer, operator...).
-- the attribute should have an inheritance behavior parameter if the containing class is abstract</value>
+- not be a special method (finalizer, operator...)
+- the attribute should have an inheritance behavior parameter if the containing class is abstract.</value>
   </data>
   <data name="ClassInitializeShouldBeValidMessageFormat" xml:space="preserve">
     <value>ClassInitialize method '{0}' signature is invalid</value>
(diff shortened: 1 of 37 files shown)


src/Analyzers/MSTest.Analyzers/Resources.resx
```diff
@@ -197,7 +197,8 @@
 - take a single parameter of type 'TestContext'
 - return type should be 'void', 'Task' or 'ValueTask'
 - not be 'async void'
-- not be a special method (finalizer, operator...).</value>
+- not be a special method (finalizer, operator...)
+- the attribute should have an inheritance behavior parameter if the containing class is abstract.</value>
```
Evangelink: It's more than simply value is set, it should be set to `BeforeEachDerivedClass`.

`- 'InheritanceBehavior.BeforeEachDerivedClass' attribute parameter should be specified if the class is 'abstract'.`

test/UnitTests/MSTest.Analyzers.UnitTests/ClassInitializeShouldBeValidAnalyzerTests.cs
```diff
@@ -541,4 +541,74 @@ public static void ClassInitialize(TestContext testContext)
...
+                public static void [|ClassInitialize|](TestContext testContext)
+                {
+                }
+            }
+            """;
+
+        await VerifyCS.VerifyAnalyzerAsync(code);
+    }
+
+    public async Task WhenClassInitializeIsOnAbstractClassMarkedWithTestClass_AndWithoutInheritanceBehavior_Diagnostic()
```
Evangelink, line 597: Let's add a test with `InheritanceBehavior.None` set

Evangelink (commented):
Thinking about it, could you also create a ticket to improve ClassInit/Cleanup to report if `InheritanceBehavior.BeforeEachDerivedClass` is set on a `sealed` class?

Evangelink (approved):

PR #3362, opened 2024-07-29
Fix class initialization/cleanup calls

### LINKED ISSUE #1889: ClassCleanup not called in base class
## Describe the bug

MSTest.TestFramework 3.0.4

ClassCleanup method in base class not called

    [TestClass]
    public class BaseClass
    {
        [ClassCleanup]
        public static void ClassCleanupBase()
        {
            Console.WriteLine(nameof(BaseClass) + nameof(ClassCleanupBase));
        }
    }

    [TestClass]
    public class UnitTest1 : BaseClass
    {
        [TestMethod]
        public void TestMethod1()
        {
            Console.WriteLine(nameof(TestMethod1));
        }
    /*
        [ClassCleanup]
        {
            Console.WriteLine(nameof(UnitTest1) + nameof(ClassCleanup));
        }
    */
    }

## Additional context

Verified in debuger.
Also mentioned in [Class Initialize inheritance #143](https://github.com/microsoft/testfx/issues/143#issuecomment-747357059)
The commented method works.

Evangelink:
Fixes #1889

### DIFF
--- a/src/Adapter/MSTest.TestAdapter/Execution/TestClassInfo.cs
+++ b/src/Adapter/MSTest.TestAdapter/Execution/TestClassInfo.cs
@@ -22,7 +22,7 @@ namespace Microsoft.VisualStudio.TestPlatform.MSTest.TestAdapter.Execution;
 /// </summary>
 public class TestClassInfo
 {
-    private readonly object _testClassExecuteSyncObject;
+    private readonly object _testClassExecuteSyncObject = new();
     private MethodInfo? _classCleanupMethod;
     private MethodInfo? _classInitializeMethod;
     private MethodInfo? _testCleanupMethod;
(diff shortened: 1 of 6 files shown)


src/Adapter/MSTest.TestAdapter/Execution/TestClassInfo.cs
```diff
@@ -144,7 +135,12 @@ internal set
     /// <summary>
     /// Gets a stack of class cleanup methods to be executed.
     /// </summary>
-    public Stack<MethodInfo> BaseClassCleanupMethodsStack { get; internal set; }
+    [Obsolete("API will be dropped in v4")]
```
fhnaseer, line 138: Should we mention alternative (BaseClassCleanupMethods) here? Same comment for all other obsolete attributes as well.

fhnaseer (approved):

src/Adapter/MSTest.TestAdapter/Execution/TestClassInfo.cs
Evangelink: Good remark! The replacement is internal but there is no good reason to have the class public so I'll keep as-is as users have no alternative.

PR #3381, opened 2024-08-01
Send Tests information to Dotnet Test

mariam-abdulla:
We flow the below data:

1. Test states
2. Attachments
3. Test session start/end events
4. Handshake info (pid, architecture, tfm, os, host type, module path)


src/Platform/Microsoft.Testing.Platform/Hosts/CommonTestHost.cs
```diff
@@ -26,25 +33,40 @@ public async Task<int> RunAsync()
...
+                    foreach (ITestApplicationLifecycleCallbacks testApplicationLifecycleCallbacks in ServiceProvider.GetServicesInternal<ITestApplicationLifecycleCallbacks>())
+                    {
+                        await testApplicationLifecycleCallbacks.AfterRunAsync(exitCode, testApplicationCancellationToken);
+                        await DisposeHelper.DisposeAsync(testApplicationLifecycleCallbacks);
+                    }
                 }
             }
+            else
+            {
+                exitCode = ExitCodes.IncompatibleProtocolVersion;
```
Evangelink: You could have a first return directly at the start of the method when the pipe is null, this would simplify the rest of this method

src/Platform/Microsoft.Testing.Platform/Hosts/CommonTestHost.cs
```diff
@@ -53,6 +75,15 @@ public async Task<int> RunAsync()
...
         finally
         {
+#if NETCOREAPP
+            if (dotnetTestPipeClient is not null)
+            {
+                await dotnetTestPipeClient.DisposeAsync();
+            }
+#else
+            dotnetTestPipeClient?.Dispose();
+#endif
```
Evangelink: You can use the helper for this case, it simplifies reading

```suggestion
            await DisposeHelper.DisposeAsync(dotnetTestPipeClient);
```

src/Platform/Microsoft.Testing.Platform/Hosts/CommonTestHost.cs
```diff
@@ -61,6 +92,25 @@ public async Task<int> RunAsync()
         return exitCode;
     }
 
+    private async Task<bool> IsDotnetTestHandshakeSuccessfulAsync(NamedPipeClient dotnetTestPipeClient, ITestApplicationModuleInfo? testApplicationModuleInfo, CancellationToken testApplicationCancellationToken)
+    {
+        HandshakeInfo handshakeInfo = new(new Dictionary<string, string>()
+        {
+            { HandshakeInfoPropertyNames.PID, ServiceProvider.GetProcessHandler().GetCurrentProcess().Id.ToString(CultureInfo.InvariantCulture) },
```
Evangelink, line 94: Getting the process ID can throw if the process is disposed, we had a couple of crashes because of that (e.g. https://github.com/microsoft/testfx/pull/3193)

src/Platform/Microsoft.Testing.Platform/Hosts/CommonTestHost.cs
```diff
@@ -61,6 +92,25 @@ public async Task<int> RunAsync()
...
+            { HandshakeInfoPropertyNames.Architecture, RuntimeInformation.OSArchitecture.ToString() },
+            { HandshakeInfoPropertyNames.Framework, RuntimeInformation.FrameworkDescription },
+            { HandshakeInfoPropertyNames.OS, RuntimeInformation.OSDescription },
+            { HandshakeInfoPropertyNames.ProtocolVersion, ProtocolConstants.Version },
+            { HandshakeInfoPropertyNames.HostType, this is ConsoleTestHost ? nameof(ConsoleTestHost) : nameof(TestHostControllersTestHost) },
```
Evangelink: This doesn't take into account potential other sub-classes. Maybe `this.GetType().Name` would be better.

src/Platform/Microsoft.Testing.Platform/Hosts/CommonTestHost.cs
```diff
@@ -61,6 +92,25 @@ public async Task<int> RunAsync()
         return exitCode;
     }
 
+    private async Task<bool> IsDotnetTestHandshakeSuccessfulAsync(NamedPipeClient dotnetTestPipeClient, ITestApplicationModuleInfo? testApplicationModuleInfo, CancellationToken testApplicationCancellationToken)
```
Evangelink: When reading this method name, I would not expect it to actually do the handshake. Maybe `DoHandshakeAsync`? Having the `bool` return is indicating if that's successful or not.

src/Platform/Microsoft.Testing.Platform/IPC/DotnetTestDataConsumer.cs
```diff
@@ -32,13 +36,144 @@ public DotnetTestDataConsumer(NamedPipeClient dotnetTestPipeClient)
 
     public string DisplayName => nameof(DotnetTestDataConsumer);
 
-    public string Description => "Send back to the dotnet test informations";
+    public string Description => "Send back information to the dotnet test";
 
-    public Task ConsumeAsync(IDataProducer dataProducer, IData value, CancellationToken cancellationToken) => Task.CompletedTask;
+    public async Task ConsumeAsync(IDataProducer dataProducer, IData value, CancellationToken cancellationToken)
+    {
+        if (value is TestNodeUpdateMessage testNodeUpdateMessage)
```
Evangelink: Use switch instead of if/elseif as you check only types, that's easier for cognitive load.

src/Platform/Microsoft.Testing.Platform/IPC/Models/FileArtifactInfo.cs
```diff
@@ -0,0 +1,6 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+namespace Microsoft.Testing.Platform.IPC.Models;
+
+internal sealed record FileArtifactInfo(string FullPath, string DisplayName, string Description, string TestUid, string TestDisplayName, string SessionUid, string ModulePath) : IRequest;
```
Evangelink: I would expect all requests to end with `Request` but I don't know if the current code is respecting that.

Evangelink (approved):

Evangelink (dismissed):

src/Platform/Microsoft.Testing.Platform/Hosts/CommonTestHost.cs
MarcoRossignoli: like in the sdk current process cannot be null. I think it's fine here.

src/Platform/Microsoft.Testing.Platform/Hosts/CommonTestHost.cs
```diff
@@ -56,11 +56,56 @@ public async Task<int> RunAsync()
...
+            { HandshakeInfoPropertyNames.Framework, RuntimeInformation.FrameworkDescription },
+            { HandshakeInfoPropertyNames.OS, RuntimeInformation.OSDescription },
+            { HandshakeInfoPropertyNames.ProtocolVersion, ProtocolConstants.Version },
+            { HandshakeInfoPropertyNames.HostType, this.GetType().Name },
+            { HandshakeInfoPropertyNames.ModulePath, testApplicationModuleInfo?.GetCurrentTestApplicationFullPath() ?? string.Empty },
+        });
+
+        HandshakeInfo response = await dotnetTestPipeClient.RequestReplyAsync<HandshakeInfo, HandshakeInfo>(handshakeInfo, testApplicationCancellationToken);
+
+        return response.Properties.TryGetValue(HandshakeInfoPropertyNames.ProtocolVersion, out string? protocolVersion) &&
```
MarcoRossignoli: Check should follow the SemVer if major is compatible we're fine. We can for now go on with this but we should fix it soon.

src/Platform/Microsoft.Testing.Platform/Hosts/TestHostBuilder.cs
```diff
@@ -494,6 +493,11 @@ private async Task SendCommandLineOptionsToDotnetTestPipeAsync(NamedPipeClient n
         return namedPipeClient;
     }
 
+    private static bool HasDotnetTestServerOption(CommandLineHandler commandLineHandler) =>
+        commandLineHandler.TryGetOptionArgumentList(PlatformCommandLineProvider.ServerOptionKey, out string[]? serverArgs) &&
+        serverArgs.Length == 1 &&
+        serverArgs[0].Equals(ServerOptionValue, StringComparison.Ordinal);
```
MarcoRossignoli: Can we rename `ServerOptionValue` in `DotnetTestCliProtocol` or something more clear?

src/Platform/Microsoft.Testing.Platform/Hosts/TestHostBuilder.cs
```diff
@@ -380,7 +380,7 @@ internal class TestHostBuilder(IFileSystem fileSystem, IRuntimeFeature runtimeFe
         serviceProvider.AddServices(testApplicationLifecycleCallback);
 
         // ServerMode and Console mode uses different host
-        if (commandLineHandler.IsOptionSet(PlatformCommandLineProvider.ServerOptionKey))
+        if (commandLineHandler.IsOptionSet(PlatformCommandLineProvider.ServerOptionKey) && !TestHostBuilder.HasDotnetTestServerOption(commandLineHandler))
```
MarcoRossignoli:
```suggestion
        if (commandLineHandler.IsOptionSet(PlatformCommandLineProvider.ServerOptionKey) && !HasDotnetTestServerOption(commandLineHandler))
```

src/Platform/Microsoft.Testing.Platform/IPC/Serializers/FileArtifactInfoSerializer.cs
```diff
@@ -0,0 +1,126 @@
...
+    |---File ModulePath Value---| (n bytes)
+*/
+
+internal sealed class FileArtifactInfoSerializer : BaseSerializer, INamedPipeSerializer
+{
+    public int Id => 7;
+
+    public object Deserialize(Stream stream)
+    {
+        string fullPath = string.Empty;
```
MarcoRossignoli: empty is a good value I think it should be null or?

MarcoRossignoli (dismissed):
some nit generally good, we should fix the nullable on deserialize to ensure clear "missing data" for reference type like string, empty one could be a good value.

src/Platform/Microsoft.Testing.Platform/Hosts/CommonTestHost.cs
mariam-abdulla: Okay, I'll do it in the next PR.

MarcoRossignoli (approved):

Issue #3323, opened 2024-07-24
undocumented/implicit `[TestClass]`/`abstract` requirement for `MSTEST0008` (an maybe possible others)


LukasGelke:
## Describe the bug

Description states:
```txt
Warning (active)	MSTEST0008	Methods marked with [TestInitialize] should follow the following layout to be valid:
- be 'public' 
- not be 'static'
- not be generic nor defined on a generic class
- not be 'abstract'
- not take any parameter
- return type should be 'void', 'Task' or 'ValueTask'
- not be 'async void'
- not be a special method (finalizer, operator...)
```

## Steps To Reproduce
```cs
public sealed class MyExample1Test
{
  [TestInitialize]
  public void Init() { } // MSTEST0008
}

[TestClass]
public sealed class MyExample2Test
{
  [TestInitialize]
  public void Init() { } // allowed, because [TestClass]
}

public abstract class MyExample3Test
{
  [TestInitialize]
  public void Init() { } // allowed, because abstract
}
```

## Expected behavior
Diagnostic Message should include (a) line(s) like:
```
- be defined in a `abstract` class or a class with a `TestClassAttribute`
```
## Actual behavior
no mention of `abstract`/ `TestClassAttribute` leads to confusion

Evangelink:
Hi @LukasGelke, 

Thanks for the feedback! We did a couple of fixes and it seems we haven't always done the follow-up for the docs.

engyebrahim:
Hello, the rule updated rule should be: 
- be defined in any class type or a sealed class marked with `TestClassAttribute`
-

PR #3391, opened 2024-08-02
Update description for (init/cleanup) analyzers

engyebrahim:
fix: https://github.com/microsoft/testfx/issues/3323


Evangelink (commented):
This is still incomplete. Maybe the best is to copy what appears on the doc page (e.g. https://learn.microsoft.com/dotnet/core/testing/mstest-analyzers/mstest0012)

engyebrahim:
@Evangelink https://github.com/microsoft/testfx/blob/main/src/Analyzers/MSTest.Analyzers/Helpers/FixtureUtils.cs#L50
the code is different for int/cleanup assembly

Evangelink (dismissed):
Small nit, could you have keywords and types surrounded by single quotes please. Otherwise LGTM

Evangelink (approved):
Thanks @engyebrahim!

PR #3405, opened 2024-08-03
remove GetNonDerivedAttributes

SimonCropp:
i dont think it is used?


Evangelink (commented):
@nohwnd I am not sure if you were planning on using it or not

nohwnd:
I am not planning on using it, it was probably factored out of the spaghetti code that was there before, that grabbed all the  attributes and filtered them in variety of ways, and later I've used IsNonDerivedAttributeDefined instead of it.

nohwnd (approved):

PR #3414, opened 2024-08-03
remove redundant catch around GetConnectionProperties

SimonCropp:
paging @Haplois


Haplois:
> paging @Haplois 

Yes, removing that should be OK; I couldn't find a good reason for that catch.

Evangelink (approved):

PR #3415, opened 2024-08-03
use GetExceptionMessage instead of ex.Message in TestDataSource

SimonCropp:
given i assume the result of GetExceptionMessage should be used


Evangelink (approved):
Sadly this legacy part is not well tested so I don't know much the impact but I also assume this was the intent.

PR #3422, opened 2024-08-03
remove ServerModeManagerExtensions

SimonCropp:


Evangelink (commented):
@MarcoRossignoli is it used somewhere else (externally)?

SimonCropp:
how can it be used externally if it is internal?

Evangelink:
> how can it be used externally if it is internal?

The wonders of `InternalsVisibleTo`!

MarcoRossignoli (approved):
It's fine they're helpers that we wanted to ship in future but we can cleanup for now.

PR #3424, opened 2024-08-03
remove ToolsInformation

SimonCropp:
has tools is not used. and to have a wrapper for an array seems pointless.

also moved tools instances to IReadOnlyList to avoid array duplications


Evangelink:
Thanks @SimonCropp, this is a concept we developped as prototype but didn't manage to finish it.

@MarcoRossignoli I'm not sure if we will do it soon enough to justify keeping it in the code or if we should rather drop it for now.

Evangelink (approved):

PR #3452, opened 2024-08-04
cache EmptyGuidUid

SimonCropp:


src/Platform/Microsoft.Testing.Extensions.TrxReport/TrxProcessLifetimeHandler.cs
```diff
@@ -148,7 +150,7 @@ public async Task OnTestHostProcessExitedAsync(ITestHostProcessInformation testH
             var extensionInfo = new ExtensionInfo(prodArtifacts.Key.Uid, prodArtifacts.Key.Version, prodArtifacts.Key.DisplayName, prodArtifacts.Key.Description);
             foreach (FileArtifact fileArtifact in prodArtifacts.Value)
             {
-                perProducerArtifact.Add(new SessionFileArtifact(new SessionUid(Guid.Empty.ToString()), fileArtifact.FileInfo, fileArtifact.DisplayName, fileArtifact.Description));
```
Evangelink, line 151: @MarcoRossignoli is this expected? If so it would be worth adding a comment

SimonCropp: TBH reusing Guid.Empty did seem a little weird

MarcoRossignoli: We will remove it soon, at the moment we don't have any usage, but with the new dotnet test we will need it so it will be refactored soon, we can keep as is for now, I'm going to work on it this sprint.

PR #3484, opened 2024-08-06
use async logging in HangDumpProcessLifetimeHandler

SimonCropp:


src/Platform/Microsoft.Testing.Extensions.HangDump/HangDumpProcessLifetimeHandler.cs
```diff
@@ -222,7 +222,7 @@ public async Task OnTestHostProcessExitedAsync(ITestHostProcessInformation testH
 
     private async Task ActivityTimerAsync()
```
MarcoRossignoli, line 223: In this method we should be sync, Mutex works with thread affinity and await is changing the thread.

SimonCropp: fair enough. will close

PR #3506, opened 2024-08-09
remove PrivateObject

SimonCropp:
does not seem to be used


Evangelink (changes requested):
Sadly it is used externally, this is a **weird** feature of MSTest for .NET Framework (see https://learn.microsoft.com/dotnet/core/testing/unit-testing-mstest-writing-tests#testing-private-members).

SimonCropp:
then shouldn’t it at least have some tests in this repo

Evangelink:
It definitely should :) As many of the inherited codebases, it's globally untested. I am trying my best to stabilize things but don't have time to just add tests for everyhting so I am focusing on ensuring new code is tested and bugs have some safetynet.

PR #3507, opened 2024-08-09
use new \e escape sequence

SimonCropp:


SimonCropp:
hmmm. i think this only worked for me since i have a newer sdk

i will close and resubmit when .net 9 is out

PR #3508, opened 2024-08-09
use Guard from Polyfill 

SimonCropp:


src/Adapter/MSTest.TestAdapter/ObjectModel/TestMethod.cs
```diff
@@ -43,10 +43,7 @@ public TestMethod(string name, string fullClassName, string assemblyName, bool i
         string fullClassName, string assemblyName, bool isAsync, string? displayName,
         TestIdGenerationStrategy testIdGenerationStrategy)
     {
-        if (StringEx.IsNullOrEmpty(assemblyName))
```
SimonCropp, line 46: i assume whitespace is also invalid?

src/Adapter/MSTest.TestAdapter/GlobalUsings.cs
```diff
@@ -0,0 +1,4 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+global using Polyfills;
```
Evangelink: Could you use global usings from the csproj please?

src/Adapter/MSTest.TestAdapter/ObjectModel/TestMethod.cs
Evangelink: I would say so!

Evangelink (dismissed):
Can you please define global using in the csproj?

Evangelink (approved):

Evangelink:
I have triggered sync of the package, I'll update the PR once it's done

Evangelink:
Alright so tests are failing because we force the locally built version of the platform that contains a breaking change (ArgumentGuard is gone). So I'll force merge and skip tests and do the roundtrip insertion.

Issue #3494, opened 2024-08-08
Platform.MSBuild should allow generating a helper for registration of extensions


Evangelink:
## Summary

Generate some test application builder extension that registers all extensions exposing a "hook".

## Background and Motivation

`Microsoft.Testing.Platform.MSBuild` can auto-generate an entry point (`Main`) in which it will add the registration of all the extensions that define a "hook". We should offer more flexibility by allowing to generate either the entry point or an extension method. 

Having this separate extension method will help with:

- xUnit v3 (cc @bradwilson)
- explicit main
- WinUI and UWP

## Proposed Feature

In addition to the flag for generating the platform entry point, let's introduce another flag that allows to generate an extension method for registering the extensions exposing a "hook". I suggest to call this method `AddAutoRegisteredExtensions`.

## Alternative Designs

- None

AB#2201333

PR #3524, opened 2024-08-12
Remove Condition=" '$(GenerateTestingPlatformEntryPoint)' == 'True' " from extensions 

engyebrahim:
fix: https://github.com/microsoft/testfx/issues/3502

PR #3522, opened 2024-08-12
Add --? alias for --help

engyebrahim:
fix: https://github.com/microsoft/testfx/issues/3501

PR #3525, opened 2024-08-12
Platform.MSBuild should allow generating a helper for registration of extensions

MarcoRossignoli:
closes https://github.com/microsoft/testfx/issues/3494

### DIFF
--- a/src/Platform/Microsoft.Testing.Platform.MSBuild/Tasks/TestingPlatformAutoRegisteredExtensions.cs
+++ b/src/Platform/Microsoft.Testing.Platform.MSBuild/Tasks/TestingPlatformAutoRegisteredExtensions.cs
@@ -3,6 +3,7 @@
 
 #pragma warning disable CS8618 // Properties below are set by MSBuild.
 
+using System.Diagnostics;
 using System.Globalization;
 using System.Text;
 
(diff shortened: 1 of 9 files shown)


src/Platform/Microsoft.Testing.Platform.MSBuild/Tasks/TestingPlatformAutoRegisteredExtensions.cs
```diff
@@ -0,0 +1,231 @@
...
+namespace Microsoft.TestingPlatform.Extensions
+
+open System.Runtime.CompilerServices
+
+[<System.Diagnostics.CodeAnalysis.ExcludeFromCodeCoverage>]
+[<Extension>]
+type AutoRegisteredExtensions() =
+
+    [<Extension>]
+    static member AddAutoRegisteredExtensions (builder: Microsoft.Testing.Platform.Builder.ITestApplicationBuilder, args: string[])=
```
Evangelink:
```suggestion
    static member AddAutoRegisteredExtensions (builder: Microsoft.Testing.Platform.Builder.ITestApplicationBuilder, args: string[]) =
```

Evangelink (dismissed):
Looks like internal poll was more in favor of `AddSelfRegisteredExtensions`. One small nit and otherwise LGTM!

Thanks for taking over @MarcoRossignoli

bradwilson:
I assume, to go along with this, all the existing extensions will be updated so that their ItemGroup registrations will be gated on `GenerateAutoRegisteredExtensions`? I would set `GenerateAutoRegisteredExtensions` to `true` and `GenerateTestingPlatformEntryPoint` to `false` and I'd get just the registration function without `Main`, right?

Sounds good to me. 😄

MarcoRossignoli:
> I would set GenerateAutoRegisteredExtensions to true and GenerateTestingPlatformEntryPoint to false and I'd get just the registration function without Main, right?

True

@engyebrahim helper with the other two issue too  https://github.com/microsoft/testfx/pull/3524 https://github.com/microsoft/testfx/pull/3522

PR #3537, opened 2024-08-13
Reduce verbosity of internal test framework

Evangelink:
cc @MarcoRossignoli and @nohwnd


nohwnd:
🤷‍♀️okay I guess, I don't know how the logging in testfx is wired, last time we talked about it I was under the impression that Eqttrace is replaced by your internal logging implementation, that is eventually wired to eqttrace or a different logger.

MarcoRossignoli:
Also to me it's not super clear what we want to achieve here.

Evangelink:
That's not MSTest, just the dummy framework that runs tests in mstest

PR #3539, opened 2024-08-13
Save all project files with UTF8 (NOBOM)

nohwnd:
Make sure all files are saved with utf8 without BOM so we don't see changes in files made by VS when we touch file that was previously saved with utf8 bom, because VS saves project files without BOM, as dictated by this setting: 

```
[*.{csproj,vbproj,vcxproj,vcxproj.filters,proj,projitems,shproj}]
indent_size = 2
charset = utf-8
```


nohwnd:
@evangelink, do you prefer to continue having the project files with or without BOM? By default VS sets BOM, but we override it in .editorconfig.

Evangelink:
> @Evangelink, do you prefer to continue having the project files with or without BOM? By default VS sets BOM, but we override it in .editorconfig.
> 
> Which every you choose it is about the same amount of changed files, to make them all consistent.

I have copied the configuration that was active on either `runtime` or `roslyn` (I don't recall) but I don't have any strong preference.

PR #3545, opened 2024-08-14
Fix warnings on newest VS

nohwnd:
Fix warnings that show up when you use VS 17.12-p1 and newer.


Evangelink (approved):

Evangelink:
Hopefully that's unblocking you @nohwnd

PR #3561, opened 2024-08-15
Fix `DisableParallelization` shows invalid value warning always

engyebrahim:
fix: https://github.com/microsoft/testfx/issues/3560


test/UnitTests/MSTestAdapter.UnitTests/MSTestSettingsTests.cs
```diff
@@ -649,6 +649,7 @@ public void DisableParallelizationShouldBeFalseByDefault()
         MSTestSettings.PopulateSettings(_mockDiscoveryContext.Object, _mockMessageLogger.Object);
 
         Verify(!MSTestSettings.CurrentSettings.DisableParallelization);
+        _mockMessageLogger.Verify(lm => lm.SendMessage(TestMessageLevel.Warning, It.IsAny<string>()), Times.Never);
```
Evangelink, line 652: Not for this PR - we should consider having strict mocks so we would catch if some warning is produced when not expecting it.

Or at least, maybe ensure we have some integration tests where we don't have any warning for the defualt use-cases.

Evangelink (approved):

PR #3565, opened 2024-08-15
Add --output option

nohwnd:
Add option to allow outputting normal, or detailed, so we can output without or with passed tests.

### DIFF
--- a/src/Platform/Microsoft.Testing.Platform/OutputDevice/Terminal/TerminalTestReporterCommandLineOptionsProvider.cs
+++ b/src/Platform/Microsoft.Testing.Platform/OutputDevice/Terminal/TerminalTestReporterCommandLineOptionsProvider.cs
@@ -14,6 +14,8 @@ internal sealed class TerminalTestReporterCommandLineOptionsProvider : ICommandL
     public const string NoProgressOption = "no-progress";
     public const string NoAnsiOption = "no-ansi";
     public const string OutputOption = "output";
+    public const string OutputOptionNormalArgument = "normal";
+    public const string OutputOptionDetailedArgument = "detailed"
 
     /// <inheritdoc />
     public string Uid { get; } = nameof(TerminalTestReporterCommandLineOptionsProvider);
@@ -43,10 +45,10 @@ internal sealed class TerminalTestReporterCommandLineOptionsProvider : ICommandL
         {
             NoProgressOption => ValidationResult.ValidTask,
             NoAnsiOption => ValidationResult.ValidTask,
-            OutputOption => "normal".Equals(arguments[0], StringComparison.OrdinalIgnoreCase) || "detailed".Equals(arguments[0], StringComparison.OrdinalIgnoreCase)
+            OutputOption => OutputOptionNormalArgument.Equals(arguments[0], StringComparison.OrdinalIgnoreCase) || OutputOptionDetailedArgument.Equals(arguments[0], StringComparison.OrdinalIgnoreCase)
                 ? ValidationResult.ValidTask
                 : ValidationResult.InvalidTask(PlatformResources.TerminalOutputOptionInvalidArgument),
-            _ => throw new NotSupportedException(),
+            _ => throw ApplicationStateGuard.Unreachable(),
         };
 
     public Task<ValidationResult> ValidateCommandLineOptionsAsync(ICommandLineOptions commandLineOptions)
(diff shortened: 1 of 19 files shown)


src/Platform/Microsoft.Testing.Platform/OutputDevice/Terminal/TerminalTestReporterCommandLineOptionsProvider.cs
```diff
@@ -32,12 +33,21 @@ internal sealed class TerminalTestReporterCommandLineOptionsProvider : ICommandL
...
     public Task<ValidationResult> ValidateOptionArgumentsAsync(CommandLineOption commandOption, string[] arguments)
-        => ValidationResult.ValidTask;
+        => commandOption.Name switch
+        {
+            NoProgressOption => ValidationResult.ValidTask,
+            NoAnsiOption => ValidationResult.ValidTask,
+            OutputOption => "normal".Equals(arguments[0], StringComparison.OrdinalIgnoreCase) || "detailed".Equals(arguments[0], StringComparison.OrdinalIgnoreCase)
+                ? ValidationResult.ValidTask
+                : ValidationResult.InvalidTask(PlatformResources.TerminalOutputOptionInvalidArgument),
+            _ => throw new NotSupportedException(),
```
Evangelink: I think we should use the unreachable state exception here (ApplicationStateGuard type)

src/Platform/Microsoft.Testing.Platform/OutputDevice/TerminalOutputDevice.cs
```diff
@@ -98,6 +98,13 @@ public TerminalOutputDevice(ITestApplicationCancellationTokenSource testApplicat
         bool noAnsi = parseResult.IsOptionSet(TerminalTestReporterCommandLineOptionsProvider.NoAnsiOption);
         bool noProgress = parseResult.IsOptionSet(TerminalTestReporterCommandLineOptionsProvider.NoProgressOption);
 
+        bool showPassed = false;
+        OptionRecord? output = parseResult.Options.FirstOrDefault(o => o.Option == TerminalTestReporterCommandLineOptionsProvider.OutputOption);
+        if (output != null && output.Arguments.Length > 0 && "Detailed".Equals(output.Arguments[0], StringComparison.OrdinalIgnoreCase))
```
Evangelink: Could we use some constants or some enum instead of duplicating the string?

src/Platform/Microsoft.Testing.Platform/Resources/PlatformResources.resx
```diff
@@ -596,4 +596,10 @@ Read more about Microsoft Testing Platform telemetry: https://aka.ms/testingplat
   <data name="TerminalTestReporterDisplayName" xml:space="preserve">
     <value>Terminal test reporter</value>
   </data>
+  <data name="TerminalOutputOptionDescription" xml:space="preserve">
+    <value>Output verbosity when reporting tests. Valid values are 'Normal', 'Detailed'. Default is 'Normal'.</value>
```
Evangelink, line 600: Shall we mention case-insensitive?

Evangelink (approved):
A few comments but LGTM!

I forgot to write it down on the previous PRs but could you please also do the doc updates related to these new options?

src/Platform/Microsoft.Testing.Platform/OutputDevice/TerminalOutputDevice.cs
nohwnd: used const.

src/Platform/Microsoft.Testing.Platform/Resources/PlatformResources.resx
nohwnd: I thought that is assumed by default, imho we should strive for all values to be case insensitive unless they are paths on a case sensitive filesystem.

src/Platform/Microsoft.Testing.Platform/OutputDevice/Terminal/TerminalTestReporterCommandLineOptionsProvider.cs
nohwnd: done.

src/Platform/Microsoft.Testing.Platform/Resources/PlatformResources.resx
Evangelink: Let me check our doc and update it if not mentioned.

Evangelink (dismissed):

nohwnd:
> I forgot to write it down on the previous PRs but could you please also do the doc updates related to these new options?

Sure.

PR #3566, opened 2024-08-15
Remove unused parameter

nohwnd:


src/Analyzers/MSTest.Analyzers.CodeFixes/TestClassShouldBeValidFixer.cs
```diff
@@ -52,7 +52,7 @@ public override async Task RegisterCodeFixesAsync(CodeFixContext context)
             diagnostic);
     }
 
-    public static async Task<Document> FixClassDeclarationAsync(Document document, SyntaxNode root, ClassDeclarationSyntax classDeclaration, CancellationToken cancellationToken)
+    public static async Task<Document> FixClassDeclarationAsync(Document document, SyntaxNode _, ClassDeclarationSyntax classDeclaration, CancellationToken cancellationToken)
```
Evangelink: Let's remove the parameter instead please

nohwnd: I tried that, but the testing fixture that consumes this method requires this method signature.

Evangelink: Huh? We should not have any test fixture using that (only line 50).

nohwnd: yeah you are right, I've misread what is happening.

Evangelink (approved):

Issue #3398, opened 2024-08-02
MSTEST0005: Add code fix


engyebrahim:
related to: https://github.com/microsoft/testfx/issues/3375

AB#2200924

PR #3571, opened 2024-08-15
MSTEST0005: Add code fix

engyebrahim:
fix: https://github.com/microsoft/testfx/issues/3398


src/Analyzers/MSTest.Analyzers.CodeFixes/Helpers/Utils.cs
```diff
@@ -0,0 +1,28 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using Analyzer.Utilities;
+
+using Microsoft.CodeAnalysis;
+
+namespace MSTest.Analyzers.Helpers;
+
+internal class Utils
```
Evangelink: Class could be `static`.

Evangelink: Let's rename it to `CompilationExtensions`.

src/Analyzers/MSTest.Analyzers.CodeFixes/Helpers/Utils.cs
```diff
@@ -0,0 +1,28 @@
...
+{
+    public static bool IsDiscoverInternalsAttributePresent(SemanticModel semanticModel)
```
Evangelink: This would be better as some extension method on the Compilation, maybe renamed as `HasDiscoverInternalAttribute`

src/Analyzers/MSTest.Analyzers.CodeFixes/TestContextShouldBeValidFixer.cs
```diff
@@ -0,0 +1,128 @@
...
+
+        MemberDeclarationSyntax newMemberDeclaration = memberDeclaration.WithModifiers(modifiers);
+
+        if (newMemberDeclaration is FieldDeclarationSyntax fieldDeclarationSyntax)
+        {
+            newMemberDeclaration = ConvertFieldToProperty(fieldDeclarationSyntax);
+        }
+        else
+        {
+            // ensure that the property have seter and getter
```
Evangelink:
```suggestion
            // ensure that the property has setter and getter
```

Evangelink (approved):
Small nit but LGTM

src/Analyzers/MSTest.Analyzers.CodeFixes/Helpers/Utils.cs
engyebrahim: we have the same naming in the analyzer project and cause conflict

Evangelink: In that case the helper should be drop and you should use `semanticModel.Compilation.CanDiscoverInternals`

Evangelink (approved):

PR #3574, opened 2024-08-15
Add empty line between output and progress indicator

nohwnd:
Added empty line because it looks better when you have passed enabled, and there is constant stream of finished tests. 

image

### DIFF
--- a/src/Platform/Microsoft.Testing.Platform/OutputDevice/TerminalOutputDevice.cs
+++ b/src/Platform/Microsoft.Testing.Platform/OutputDevice/TerminalOutputDevice.cs
@@ -5,6 +5,7 @@ using System.Globalization;
 using System.Runtime.InteropServices;
 using System.Text;
 
+using Microsoft.Testing.Platform.CommandLine;
 using Microsoft.Testing.Platform.Extensions.Messages;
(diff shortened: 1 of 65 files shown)


Evangelink (dismissed):

src/Platform/Microsoft.Testing.Platform/OutputDevice/TerminalOutputDevice.cs
```diff
@@ -115,7 +115,13 @@ public TerminalOutputDevice(ITestApplicationCancellationTokenSource testApplicat
             // The test host controller info is not setup and populated until after this constructor, because it writes banner and then after it figures out if
             // the runner is a testHost controller, so we would always have it as null if we capture it directly. Instead we need to check it via
             // func.
-            : () => testHostControllerInfo.IsCurrentProcessTestHostController == null ? null : !testHostControllerInfo.IsCurrentProcessTestHostController;
+            : () => _isServerMode
+                ? false
+                : _isListTests
```
Evangelink, line 120: Approving as-is, and maybe it's my F# experience talking here but a switch pattern would probably be easier to read here.

Evangelink (approved):

PR #3582, opened 2024-08-16
Fix writing colors to terminal

nohwnd:
Allow any system.color to be written by formatted text message, instead of just supporting red and yellow for error and warning.

### DIFF
--- a/src/Platform/Microsoft.Testing.Platform/OutputDevice/Terminal/TerminalTestReporter.cs
+++ b/src/Platform/Microsoft.Testing.Platform/OutputDevice/Terminal/TerminalTestReporter.cs
@@ -580,7 +580,7 @@ internal sealed partial class TerminalTestReporter : IDisposable
             return;
         }
 
-        terminal.SetColor(TerminalColor.Red);
+        terminal.SetColor(TerminalColor.Blue);
         AppendIndentedMessage(terminal, errorMessage, SingleIndentation);
         terminal.ResetColor();
         terminal.AppendLine();
@@ -683,7 +683,7 @@ internal sealed partial class TerminalTestReporter : IDisposable
 
         _terminalWithProgress.WriteToTerminal(terminal =>
         {
-            terminal.SetColor(TerminalColor.Red);
+            terminal.SetColor(TerminalColor.Blue);
             terminal.AppendLine(text);
             terminal.ResetColor();
         });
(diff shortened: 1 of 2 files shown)


src/Platform/Microsoft.Testing.Platform/OutputDevice/Terminal/TerminalTestReporter.cs
```diff
@@ -704,6 +704,61 @@ internal void WriteWarningMessage(string assembly, string? targetFramework, stri
...
+        }
+        else
+        {
+            _terminalWithProgress.WriteToTerminal(terminal => terminal.AppendLine(text));
+        }
+    }
+
+    private TerminalColor ToTerminalColor(ConsoleColor consoleColor)
+    {
+        switch (consoleColor)
```
Evangelink: Nit: switch expression would be more compact

src/Platform/Microsoft.Testing.Platform/OutputDevice/Terminal/TerminalTestReporter.cs
```diff
@@ -704,6 +704,61 @@ internal void WriteWarningMessage(string assembly, string? targetFramework, stri
...
+                terminal.ResetColor();
+            });
+        }
+        else
+        {
+            _terminalWithProgress.WriteToTerminal(terminal => terminal.AppendLine(text));
+        }
+    }
+
+    private TerminalColor ToTerminalColor(ConsoleColor consoleColor)
```
Evangelink:
```suggestion
    private static TerminalColor ToTerminalColor(ConsoleColor consoleColor)
```

src/Platform/Microsoft.Testing.Platform/OutputDevice/Terminal/TerminalTestReporter.cs
```diff
@@ -704,6 +704,61 @@ internal void WriteWarningMessage(string assembly, string? targetFramework, stri
...
-        => _terminalWithProgress.WriteToTerminal(terminal => terminal.AppendLine(text));
+    internal void WriteMessage(string text, SystemConsoleColor? color = null)
+    {
+        if (color != null)
+        {
+            _terminalWithProgress.WriteToTerminal(terminal =>
+            {
+                terminal.SetColor(ToTerminalColor(color.ConsoleColor));
+                terminal.AppendLine(text);
+                terminal.ResetColor();
```
Evangelink, line 715: Shall we be using some kind of try/finally pattern here?

nohwnd: I don't think there is much opportunity for error. It is probably much easier to just reset the color at the end of the run, if you are afraid of keeping the color set to non-default. But so far I haven't seen it happen.

Evangelink: Ok trusting you here.

Evangelink (approved):

Evangelink:
Restarting failed pipeline as the error seems unrelated to your change.

PR #3583, opened 2024-08-16
Fix shortening TFM and architecture

### LINKED ISSUE #3579: Strange runtime identifier printed in the banner
image

Code here https://github.com/microsoft/testfx/blob/main/src/Platform/Microsoft.Testing.Platform/OutputDevice/TerminalOutputDevice.cs#L141

Why don't use the api https://learn.microsoft.com/en-us/dotnet/api/system.runtime.interopservices.runtimeinformation.runtimeidentifier?view=net-8.0#system-runtime-interopservices-runtimeinformation-runtimeidentifier

nohwnd:
Fix #3579

### DIFF
--- a/src/Platform/Microsoft.Testing.Platform/OutputDevice/TargetFrameworkParser.cs
+++ b/src/Platform/Microsoft.Testing.Platform/OutputDevice/TargetFrameworkParser.cs
@@ -7,7 +7,7 @@ namespace Microsoft.Testing.Platform.OutputDevice;
 
 internal static class TargetFrameworkParser
 {
-    internal /* for testing */ static string? GetShortTargetFramework(string frameworkDescription)
+    public static string? GetShortTargetFramework(string frameworkDescription)
     {
         // https://learn.microsoft.com/dotnet/api/system.runtime.interopservices.runtimeinformation.frameworkdescription
         string netFramework = ".NET Framework";
--- a/src/Platform/Microsoft.Testing.Platform/OutputDevice/TargetFrameworkParser.cs
+++ b/src/Platform/Microsoft.Testing.Platform/OutputDevice/TargetFrameworkParser.cs
@@ -9,7 +9,7 @@ internal static class TargetFrameworkParser
 {
     internal /* for testing */ static string? GetShortTargetFramework(string frameworkDescription)
     {
-        // https://learn.microsoft.com/en-us/dotnet/api/system.runtime.interopservices.runtimeinformation.frameworkdescription?view=net-8.0
+        // https://learn.microsoft.com/dotnet/api/system.runtime.interopservices.runtimeinformation.frameworkdescription
         string netFramework = ".NET Framework";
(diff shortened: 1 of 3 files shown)


MarcoRossignoli (approved):

src/Platform/Microsoft.Testing.Platform/OutputDevice/TargetFrameworkParser.cs
```diff
@@ -0,0 +1,69 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System.Globalization;
+
+namespace Microsoft.Testing.Platform.OutputDevice;
+
+internal static class TargetFrameworkParser
+{
+    internal /* for testing */ static string? GetShortTargetFramework(string frameworkDescription)
```
Evangelink: You could make it public as the class is `internal`. I usually go with `internal` on `internal` only if that's really not meant to be used.

src/Platform/Microsoft.Testing.Platform/OutputDevice/TargetFrameworkParser.cs
```diff
@@ -0,0 +1,69 @@
...
+    {
+        // https://learn.microsoft.com/en-us/dotnet/api/system.runtime.interopservices.runtimeinformation.frameworkdescription?view=net-8.0
```
Evangelink: Nit: let's be cultural friendly :)

```suggestion
        // https://learn.microsoft.com/dotnet/api/system.runtime.interopservices.runtimeinformation.frameworkdescription
```

src/Platform/Microsoft.Testing.Platform/OutputDevice/TargetFrameworkParser.cs
```diff
@@ -0,0 +1,69 @@
...
+                return "net472";
+            }
+            else if (major == '4' && minor == '8' && patch == '1')
+            {
+                return "net481";
+            }
+            else
+            {
+                // Just return the first 2 numbers.
+                return $"net{major}{minor}";
```
Evangelink, line 45: Let's hope we won't have some more releases :)

src/Platform/Microsoft.Testing.Platform/OutputDevice/TargetFrameworkParser.cs
nohwnd:
```suggestion
    public static string? GetShortTargetFramework(string frameworkDescription)
```

src/Platform/Microsoft.Testing.Platform/OutputDevice/TargetFrameworkParser.cs
nohwnd: If there are more it will have to evolve, there is no logic in when I would expect and not expect seeing short version. ;/

nohwnd:
failing test is flaky it seems. Cannot repro locally.

Issue #3263, opened 2024-07-15
Add analyzer for when a test member is shadowing another member


Evangelink:
## Summary

Prevent testing issues (such as NRE) when user is shadowing test members.

## Background and Motivation

.NET allows to do shadowing of members and MSTest will respect the behavior but this is nearly always causing issues. When shadowing a member (test, TestContext property...) from a base class, the base class is ignored (see https://github.com/microsoft/playwright-dotnet/issues/2958).

## Proposed Feature

Raise a diagnostic when a member is being shadowed even if the user is using the `new` keyword to force shadowing.

The analyzer should be enabled by default with a `warning` severity and its category should be `design`.

## Alternative Designs

None

AB#2200925

PR #3589, opened 2024-08-16
MSTEST0036: Add analyzer for when a test member is shadowing another member

engyebrahim:
fix: https://github.com/microsoft/testfx/issues/3263


src/Analyzers/MSTest.Analyzers/DoNotUseShadowingAnalyzer.cs
```diff
@@ -0,0 +1,113 @@
...
+            || !namedTypeSymbol.GetAttributes().Any(attr => SymbolEqualityComparer.Default.Equals(attr.AttributeClass, testClassAttributeSymbol)))
+        {
+            return;
+        }
+
+        INamedTypeSymbol? currentType = namedTypeSymbol.BaseType;
+        var baseClassesMembers = new List<ISymbol>();
+        while (currentType is not null)
+        {
+            baseClassesMembers.AddRange(currentType.GetMembers());
```
Evangelink: You could filter out the overridden members.

src/Analyzers/MSTest.Analyzers/DoNotUseShadowingAnalyzer.cs
```diff
@@ -0,0 +1,113 @@
...
+            currentType = currentType.BaseType;
+        }
+
+        foreach (ISymbol member in namedTypeSymbol.GetMembers())
+        {
+            foreach (ISymbol baseMember in baseClassesMembers)
```
Evangelink: As discussed, I think the dictionary by name would be a huge simplification over this "brute force" approach.

src/Analyzers/MSTest.Analyzers/DoNotUseShadowingAnalyzer.cs
```diff
@@ -0,0 +1,113 @@
...
+            {
+                // Check if the member is shadowing a base class member
+                if (IsMemberShadowing(member, baseMember))
+                {
+                    context.ReportDiagnostic(member.CreateDiagnostic(DoNotUseShadowingRule, member.Name));
+                    return;
```
Evangelink: You are only reporting one case at a time, so if there are multiple violations, the user has to fix, compile in chain. When possible we should show the various violations.

src/Analyzers/MSTest.Analyzers/DoNotUseShadowingAnalyzer.cs
```diff
@@ -0,0 +1,126 @@
...
+
+            currentType = currentType.BaseType;
+        }
+
+        foreach (ISymbol member in namedTypeSymbol.GetMembers())
+        {
+            foreach (ISymbol baseMember in symbolGroups.GetValueOrDefault(member.Name, new List<ISymbol>()))
+            {
+                // Check if the member is shadowing a base class member
+                if (!member.IsOverride && IsMemberShadowing(member, baseMember))
```
Evangelink: Might be worth checking that and ctor outside of the foreach. You probably can factor out the condition here and on line 69 into a method.

src/Analyzers/MSTest.Analyzers/DoNotUseShadowingAnalyzer.cs
```diff
@@ -0,0 +1,126 @@
...
+                    members = new List<ISymbol>();
+                    symbolGroups[member.Name] = members;
+                }
+
+                // Add the member to the list
+                members.Add(member);
+            }
+
+            currentType = currentType.BaseType;
+        }
```
Evangelink: Let's have a method for this part so that the main method is composing readable blocks.

test/UnitTests/MSTest.Analyzers.UnitTests/DoNotUseShadowingAnalyzerTests.cs
```diff
@@ -0,0 +1,164 @@
...
+            using Microsoft.VisualStudio.TestTools.UnitTesting;
+            public class BaseClass
+            {
+                public TestContext TestContext { get; set; } = null;
+            }
+
+            [TestClass]
+            public class DerivedClass : BaseClass
+            {
+                public TestContext [|TestContext|] { [|get|]; [|set|]; } = null;
```
Evangelink: We probably want only one case reported here (you will need to filter out accessor methods).

Evangelink (approved):

Issue #3622, opened 2024-08-19
Serilization issue using xUnit suite on .NET Framework


MarcoRossignoli:
Ref https://github.com/microsoft/testfx/issues/3478#issuecomment-2297238366

MarcoRossignoli:
Jsonite is not escaping special chars like

+ => \u002B
"  => \u0022

PR #3641, opened 2024-08-20
Fix `Content-Length`

MarcoRossignoli:
fixes https://github.com/microsoft/testfx/issues/3622


drognanar (approved):

Evangelink (approved):

nohwnd:
Maybe a test with some emojis?

MarcoRossignoli:
> Maybe a test with some emojis?

We have smoke tests at the moment. We should spend more than one test to cover everything. We can discuss about it, adding only this test looks to me not enough.

Issue #3500, opened 2024-08-08
Platform should expose a timeout option


thomhurst:
I've seen in many test suites something go wrong before where something is hanging and the test suite just becomes stuck.

Adding a timeout option, that if exceeded, will send a cancellation signal to the framework's cancellation token could help give developers faster feedback and also save them money if they're using less compute time on their build agents.

The format could be a C# stringified timespan or you could come up with your own format.

Example:

`dotnet run --timeout 00:10:00`

AB#2200943

Evangelink:
It seems like the responsibility of the test framework.

thomhurst:
I just thought since there's a cancellation token passed to us in the request context, that'd be a good place to cancel it.

But no worries, it should be easy enough to implement in my framework.

Feel free to close :)

Evangelink:
Reading again the ticket, I am thinking that maybe I didn't get what you were suggesting. Is your timeout a time per test or a global test execution timeout?

thomhurst:
Sorry if it wasn't clear. A global timeout for the whole test session

Evangelink:
Ok in that case I think it makes sense :) I was thinking about the platform handling a test method timeout and I was thinking that while technically we could do something along this line it seemed like something a lot easier to handle at framework level.

@MarcoRossignoli this seems like a useful request, WDYT?

Evangelink:
Let's go with `--timeout <val>` where `<val>` will be either an int corresponding to timout in seconds or a string in the format `<value>[h|m|s]`

PR #3642, opened 2024-08-20
Platform should expose a timeout option

engyebrahim:
fix: https://github.com/microsoft/testfx/issues/3500


src/Platform/Microsoft.Testing.Platform/Resources/PlatformResources.resx
```diff
@@ -602,4 +602,7 @@ Read more about Microsoft Testing Platform telemetry: https://aka.ms/testingplat
   <data name="TerminalOutputOptionInvalidArgument" xml:space="preserve">
     <value>--output expects a single parameter with value 'Normal' or 'Detailed'.</value>
   </data>
+  <data name="PlatformCommandLineTimeoutArgumentErrorMessage" xml:space="preserve">
+    <value>'timeout' option should have one argument as string in the format &lt;value&gt;[h|m|s] and 'value' is float</value>
```
Evangelink: Nit

```suggestion
    <value>'timeout' option should have one argument as string in the format &lt;value&gt;[h|m|s] where 'value' is float</value>
```

src/Platform/Microsoft.Testing.Platform/Hosts/TestHostBuilder.cs
```diff
@@ -236,8 +237,30 @@ public async Task<ITestHost> BuildAsync(
...
         ICommandLineOptions commandLineOptions = serviceProvider.GetCommandLineOptions();
+
+        // setting the timeout
+        if (commandLineOptions.IsOptionSet(PlatformCommandLineProvider.TimeoutOptionKey) && commandLineOptions.TryGetOptionArgumentList(PlatformCommandLineProvider.TimeoutOptionKey, out string[]? args))
+        {
+            string arg = args[0];
+            int size = arg.Length;
+            if (!float.TryParse(arg[..(size - 1)], out float value))
+            {
+                throw new InvalidOperationException(PlatformResources.PlatformCommandLineTimeoutArgumentErrorMessage);
```
Evangelink: We can use `throw ApplicationStateGuard.Unreachable()`, because this is already validated so this would be some unexpected state.

```suggestion
                throw ApplicationStateGuard.Unreachable();
```

src/Platform/Microsoft.Testing.Platform/Hosts/TestHostBuilder.cs
```diff
@@ -236,8 +237,30 @@ public async Task<ITestHost> BuildAsync(
...
+            }
+
+            TimeSpan timeout = char.ToLowerInvariant(arg[size - 1]) switch
+            {
+                'h' => TimeSpan.FromHours(value),
+                'm' => TimeSpan.FromMinutes(value),
+                's' => TimeSpan.FromSeconds(value),
+                _ => throw new InvalidOperationException(PlatformResources.PlatformCommandLineTimeoutArgumentErrorMessage),
```
Evangelink:
```suggestion
                _ => throw ApplicationStateGuard.Unreachable(),
```

src/Platform/Microsoft.Testing.Platform/Hosts/TestHostBuilder.cs
```diff
@@ -236,8 +237,30 @@ public async Task<ITestHost> BuildAsync(
             Logging.AddProvider((_, _) => loggingState.FileLoggerProvider);
         }
 
-        // Register the server mode log forwarder if needed. We follow the console --diagnostic behavior.
```
Evangelink, line 239: Why is this removed?

src/Platform/Microsoft.Testing.Platform/Hosts/TestHostBuilder.cs
```diff
@@ -236,8 +237,30 @@ public async Task<ITestHost> BuildAsync(
...
+
+            TimeSpan timeout = char.ToLowerInvariant(arg[size - 1]) switch
+            {
+                'h' => TimeSpan.FromHours(value),
+                'm' => TimeSpan.FromMinutes(value),
+                's' => TimeSpan.FromSeconds(value),
+                _ => throw new InvalidOperationException(PlatformResources.PlatformCommandLineTimeoutArgumentErrorMessage),
+            };
+
+            testApplicationCancellationTokenSource.CancelAfter(timeout);
```
Evangelink, line 259: I think that's fine as-is but I'd like to get @MarcoRossignoli's view. Marco, shall we do that directly when initializing the test app source object?

Evangelink (commented):
Some test and it's perfect!

src/Platform/Microsoft.Testing.Platform/Hosts/TestHostBuilder.cs
engyebrahim: it's still there but after the code i added

src/Platform/Microsoft.Testing.Platform/Hosts/TestHostBuilder.cs
MarcoRossignoli: I think it's fine we build the appct before the command line. And I prefer to use the `ICommandLineOptions` where timing allows it.

test/IntegrationTests/Microsoft.Testing.Platform.Acceptance.IntegrationTests/TimeoutTests.cs
```diff
@@ -0,0 +1,145 @@
...
+        testHostResult.StandardError.Contains("'timeout' option should have one argument as string in the format <value>[h|m|s] where 'value' is float");
+    }
+
+    [ArgumentsProvider(nameof(TargetFrameworks.All), typeof(TargetFrameworks))]
+    public async Task TimeoutWithValidArg_WithTestTimeOut_BuildFails(string tfm)
+    {
+        var testHost = TestInfrastructure.TestHost.LocateFrom(_testAssetFixture.NoExtensionTargetAssetPath, TestAssetFixture.AssetName, tfm);
+        TestHostResult testHostResult = await testHost.ExecuteAsync("--timeout 1s");
+
+        testHostResult.AssertExitCodeIsNot(ExitCodes.Success);
```
Evangelink, line 53: Let's assert we see the cancelling message.

test/IntegrationTests/Microsoft.Testing.Platform.Acceptance.IntegrationTests/TimeoutTests.cs
```diff
@@ -0,0 +1,145 @@
...
+    }
+
+    [ArgumentsProvider(nameof(TargetFrameworks.All), typeof(TargetFrameworks))]
+    public async Task TimeoutWithValidArg_WithHourAsSuffix_WithTestNotTimeOut_BuildPasses(string tfm)
+    {
+        var testHost = TestInfrastructure.TestHost.LocateFrom(_testAssetFixture.NoExtensionTargetAssetPath, TestAssetFixture.AssetName, tfm);
+        TestHostResult testHostResult = await testHost.ExecuteAsync("--timeout 1h");
+
+        testHostResult.AssertExitCodeIs(ExitCodes.Success);
+    }
```
Evangelink: All these tests happens after the execution is done. Maybe update the name so that's more explicit and we can also check the output doesn't contain the cancel message.

PR #3648, opened 2024-08-21
Add known report files

nohwnd:
Add extensions for report files used by xunit.

image


Evangelink (approved):

Evangelink:
Issues on native AOT tests, I'll force merge this PR and work on fixing the tests this afternoon (flakiness).

bradwilson:
FYI I changed the default xunit report extension from `.xml` to `.xunit` for parity with the others.

bradwilson:
Though I do think people, when they override filenames, will very commonly use `.xml` (and maybe `.json` for the CTRF), so leaving `.xml` in the list seems like a good idea.

nohwnd:
Ok then I should add .xunit as well.

PR #3649, opened 2024-08-21
Change --help formatting (and fix --info)

nohwnd:
Change help formatting from 

image

to 

image (4)

This avoids the help being pushed to right side by the longest option.

We **DO NOT** split the lines automatically and reflow them, you need to put newlines in the description manually, and keep each line short. This is to avoid splitting links (so they remain clickable), and to let the terminal reflow long lines, so user can widen their terminal and get the correct view without re-running.


nohwnd:
Draft because I am waiting on whether I should keep the spaces between entries. And then I need to fix help tests that list the output literally.

nohwnd:
Fails on native AOT tests now. Merging.

Issue #3646, opened 2024-08-21
Investigate why numbers are not allowed in command line options


Evangelink:
Based on @bradwilson comment https://github.com/microsoft/testfx/issues/3478#issuecomment-2299800552

Evangelink:
> I'll do an update to include the numbers in the options. I don't recall why is there, maybe is POSIX spec or maybe an error.

As per @MarcoRossignoli's comment it's possible this is a limitation, in which case it should be documented.

PR #3651, opened 2024-08-21
Allow digit inside option name

MarcoRossignoli:
`System.CommandLine` allows it too.

Fixes https://github.com/microsoft/testfx/issues/3646


MarcoRossignoli:
> Could we add some test if we have a parsing test?

The parser allows it already and we have it, it's a ctor check.

MarcoRossignoli:
> would it be appropriate for the PR to also ensure that?:
? can only be used alone: e.g. not --what??
and - cannot be used at the end of the parameter, so not --i-am-so-incomplet-

I don't know if we should block these case is it wrong?

PR #3659, opened 2024-08-22
Add `.xunit` to known files

### LINKED ISSUE #3654: add .xunit to list of known file extensions
it is the new default for report name xunit3

nohwnd:
Fix #3654

### DIFF
--- a/src/Platform/Microsoft.Testing.Platform/OutputDevice/Terminal/AnsiTerminal.cs
+++ b/src/Platform/Microsoft.Testing.Platform/OutputDevice/Terminal/AnsiTerminal.cs
@@ -37,6 +37,7 @@ internal sealed class AnsiTerminal : ITerminal
         ".nunit",
         ".trx",
         ".xml",
+        ".xunit",
     };
 
     private readonly IConsole _console;


Evangelink (approved):

Evangelink:
@nohwnd Would it make sense to have a cli option to specify extra supported files? Or/and would it make sense to have some extensibility point for some extension to be able to register its file format? For example, we could consider xunit file format to not be part of the core platform but be registered by xunit (cc @bwilsonms)

nohwnd:
Not at this point I think, I tried to describe this in the main xunit thread, but I think it was confusing. 

We have a list of known file extensions only because Linux does not require executables to have any extension, and it is common to have text files that do not have any extension. This information is used for the terminal output, only when ANSI is allowed, and only when the default output device is used (we currently don't have a way to replace that, but that might change.).

I also don't want to query the filesystem every time I am writing a file entry to the screen.  Hence the list of known extensions. 

I can imagine the reporter registration will tell us what type of file they will produce, but I am not 100% there is not a better solution to figuring out if file is executable, and so I would like to keep this list a bit longer till I have time to properly figure out what the options are.

Issue #3569, opened 2024-08-15
Generate Correlation ID between Hosts for dotnet test Feature


mariam-abdulla:
We need to correlate different executions for the same module tfm to support extensions like retry.

PR #3670, opened 2024-08-22
Generate Correlation ID between hosts

mariam-abdulla:
In order to distinguish between different test runs for the same test project (which could result from retry feature, e.g.)
and to group test results/artifacts corresponding to a single run, we introduced execution ID which could be shared between the test host, test host controller and test host orchestrator.

https://github.com/microsoft/testfx/issues/3569


src/Platform/Microsoft.Testing.Platform/DotnetTest/DotnetTestConnection.cs
```diff
@@ -0,0 +1,70 @@
...
+
+namespace Microsoft.Testing.Platform.DotnetTest;
+
+internal sealed class DotnetTestConnection
+{
+    private readonly CommandLineHandler _commandLineHandler;
+    private readonly ITestApplicationModuleInfo _testApplicationModuleInfo;
+    private readonly CTRLPlusCCancellationTokenSource _cancellationTokenSource;
+
+    public DotnetTestConnection(CommandLineHandler commandLineHandler, ITestApplicationModuleInfo testApplicationModuleInfo, CTRLPlusCCancellationTokenSource cancellationTokenSource)
```
Evangelink: I would suggest to use the interface instead of the implementation class for the token source.

src/Platform/Microsoft.Testing.Platform/DotnetTest/DotnetTestHelper.cs
```diff
@@ -0,0 +1,16 @@
...
+
+using Microsoft.Testing.Platform.CommandLine;
+
+namespace Microsoft.Testing.Platform.DotnetTest;
+
+internal static class DotnetTestHelper
+{
+    private const string DotnetTestCliProtocol = "dotnettestcli";
+
+    public static bool HasDotnetTestServerOption(CommandLineHandler commandLineHandler) =>
```
Evangelink: Extension method maybe?

src/Platform/Microsoft.Testing.Platform/DotnetTest/DotnetTestHelper.cs
```diff
@@ -0,0 +1,16 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using Microsoft.Testing.Platform.CommandLine;
+
+namespace Microsoft.Testing.Platform.DotnetTest;
+
+internal static class DotnetTestHelper
+{
+    private const string DotnetTestCliProtocol = "dotnettestcli";
```
MarcoRossignoli: I would keep this one in the Platform defaults constants. So we have everything in one place when we check what we expose.

src/Platform/Microsoft.Testing.Platform/Hosts/CommonTestHost.cs
```diff
@@ -98,7 +100,7 @@ private async Task<bool> DoHandshakeAsync(NamedPipeClient dotnetTestPipeClient,
             { HandshakeInfoPropertyNames.ProtocolVersion, ProtocolConstants.Version },
             { HandshakeInfoPropertyNames.HostType, GetType().Name },
             { HandshakeInfoPropertyNames.ModulePath, testApplicationModuleInfo?.GetCurrentTestApplicationFullPath() ?? string.Empty },
-            { HandshakeInfoPropertyNames.ExecutionId, string.Empty },
+            { HandshakeInfoPropertyNames.ExecutionId,  environment.GetEnvironmentVariable(EnvironmentVariableConstants.TESTINGPLATFORM_DOTNETTEST_EXECUTIONID) ?? string.Empty },
```
MarcoRossignoli: It cannot be empty or?  Is it a safenet if case if bug? In case add a comment above that's unexpected.

src/Platform/Microsoft.Testing.Platform/DotnetTest/DotnetTestConnection.cs
```diff
@@ -0,0 +1,70 @@
...
+    {
+        NamedPipeClient? namedPipeClient = null;
+        string executionId = string.Empty;
+
+        // If we are in server mode and the pipe name is provided
+        // then, we need to connect to the pipe server.
+        if (_commandLineHandler.HasDotnetTestServerOption() &&
+            _commandLineHandler.TryGetOptionArgumentList(PlatformCommandLineProvider.DotNetTestPipeOptionKey, out string[]? arguments))
+        {
+            executionId = Guid.NewGuid().ToString("N");
```
MarcoRossignoli: I don't see the check of current env var and the usage of this value.

MarcoRossignoli (changes requested):
Discussed different refactoring offline

src/Platform/Microsoft.Testing.Platform/CommandLine/InformativeCommandLineTestHost.cs
```diff
@@ -2,30 +2,31 @@
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
 using Microsoft.Testing.Platform.Hosts;
-using Microsoft.Testing.Platform.IPC;
+using Microsoft.Testing.Platform.Services;
 
 namespace Microsoft.Testing.Platform.CommandLine;
 
-internal sealed class InformativeCommandLineTestHost(int returnValue, NamedPipeClient? dotnetTestPipe = null) : ITestHost, IDisposable
+internal sealed class InformativeCommandLineTestHost(int returnValue, ServiceProvider serviceProvider) : ITestHost, IDisposable
```
MarcoRossignoli: `IServiceProvider`

src/Platform/Microsoft.Testing.Platform/DotnetTest/IPC/Constants.cs
```diff
@@ -3,6 +3,11 @@
 
 namespace Microsoft.Testing.Platform.IPC;
 
+internal static class Constants
+{
+    internal const string DotnetTestCliProtocol = "dotnettestcli";
```
MarcoRossignoli: I'd like to have this one inside the platform command line so we have everything in the same place when we check what's we expose.

src/Platform/Microsoft.Testing.Platform/DotnetTest/IPC/DotnetTestDataConsumer.cs
```diff
@@ -33,8 +40,17 @@ public DotnetTestDataConsumer(NamedPipeClient dotnetTestPipeClient)
 
     public string Description => "Send back information to the dotnet test";
 
+    private string? ExecutionId => Environment.GetEnvironmentVariable(EnvironmentVariableConstants.TESTINGPLATFORM_DOTNETTEST_EXECUTIONID);
+
     public async Task ConsumeAsync(IDataProducer dataProducer, IData value, CancellationToken cancellationToken)
     {
+        if (DotnetTestConnection?.IsConnected == false)
```
MarcoRossignoli: I suppose we can also avoid to register  the `DotnetTestDataConsumer` is it's not connected so we save computation. The distribution of the `IData` isn't cheap.

src/Platform/Microsoft.Testing.Platform/Hosts/CommonTestHost.cs
```diff
@@ -56,12 +53,23 @@ public async Task<int> RunAsync()
...
-            await DisposeHelper.DisposeAsync(dotnetTestPipeClient);
+            await DisposeHelper.DisposeAsync(DotnetTestConnection);
         }
 
         return exitCode;
     }
 
+    private string GetHostType()
+    {
+        string hostType = this switch
```
MarcoRossignoli, line 65: pls add a comment:

For now we don't inherit `TestHostOrchestratorHost` from the common one so we don't connect when we orchestrate.

src/Platform/Microsoft.Testing.Platform/Hosts/TestHostBuilder.cs
```diff
@@ -350,28 +349,34 @@ await LogTestHostCreatedAsync(
             return toolsTestHost;
         }
 
-        NamedPipeClient? dotnetTestPipeClient = await ConnectToDotnetTestPipeIfAvailableAsync(commandLineHandler, testApplicationCancellationTokenSource);
+        DotnetTestConnection dotnetTestConnection = new(serviceProvider, testApplicationCancellationTokenSource);
+        bool isConnectedToDotnetTest = await dotnetTestConnection.TryConnectToDotnetTestPipeIfAvailableAsync();
+
```
MarcoRossignoli: remove line pls

src/Platform/Microsoft.Testing.Platform/Hosts/TestHostBuilder.cs
```diff
@@ -350,28 +349,34 @@ await LogTestHostCreatedAsync(
...
+        if (isConnectedToDotnetTest)
+        {
+            serviceProvider.AddService(dotnetTestConnection);
```
MarcoRossignoli, line 356: I think we can always add it and retrive where is needed, we can check there `IsConnected` as condition to "do something or not" like register or not the custom `IDataConsumer`. Makes the code here easier.

MarcoRossignoli (changes requested):
Some more cleanup/reorg but LGTM

src/Platform/Microsoft.Testing.Platform/Hosts/TestHostBuilder.cs
MarcoRossignoli: Never mind me if you want to use `? GetService()` but I don't want to register the `IDataCollector` if we're not connected.

src/Platform/Microsoft.Testing.Platform/DotnetTest/DotnetTestConnection.cs
```diff
@@ -0,0 +1,167 @@
...
+
+            return _dotnetTestPipeClient.IsConnected;
+        }
+
+        return false;
+    }
+
+    public async Task SendCommandLineOptionsToDotnetTestPipeAsync()
+    {
+        if (_dotnetTestPipeClient?.IsConnected == false)
```
MarcoRossignoli: it's not expected this call or? I prefer an opposite RoslynAssert  `_dotnetTestPipeClient?.IsConnected == true`

MarcoRossignoli (approved):

PR #3673, opened 2024-08-22
Reduce output when resizing to avoid breaking the progress

nohwnd:
Before: when the window gets smaller the progress is broken, and end up scrolling down on the screen. Same thing happens when assembly name is too long.

sizing-broken

After: we hide the tfm and architecture first, then we start trimming the name from right prepending with ... to keep the output to correct width and keep the time on screen.

sizing

### DIFF
--- a/test/UnitTests/Microsoft.Testing.Extensions.UnitTests/CrashDumpTests.cs
+++ b/test/UnitTests/Microsoft.Testing.Extensions.UnitTests/CrashDumpTests.cs
@@ -19,6 +19,7 @@ public class CrashDumpTests(ITestExecutionContext testExecutionContext) : TestBa
     [Arguments("Full")]
     public async Task IsValid_If_CrashDumpType_Has_CorrectValue(string crashDumpType)
     {
+        Thread.Sleep(20_000);
         var provider = new CrashDumpCommandLineProvider();
         CommandLineOption option = provider.GetCommandLineOptions().First(x => x.Name == CrashDumpCommandLineOptions.CrashDumpTypeOptionName);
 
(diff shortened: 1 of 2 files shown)


MarcoRossignoli (approved):

test/UnitTests/Microsoft.Testing.Extensions.UnitTests/CrashDumpTests.cs
```diff
@@ -19,6 +19,7 @@ public class CrashDumpTests(ITestExecutionContext testExecutionContext) : TestBa
     [Arguments("Full")]
     public async Task IsValid_If_CrashDumpType_Has_CorrectValue(string crashDumpType)
     {
+        Thread.Sleep(20_000);
```
Evangelink: Should be reverted

Evangelink:
What about having the time just after the tfm/platform? If I have a really long line there might not be much need to have so much empty space and it seems to be causing more complexity that could be avoided.

test/UnitTests/Microsoft.Testing.Extensions.UnitTests/CrashDumpTests.cs
nohwnd:
```suggestion
```

nohwnd:
It looks worse, and it does not reduce complexity. 

It would be as complex. 

Now we write the duration at the end of the line by jumping to x chars from the end of the line and writing the string. When we see that the progress state was not updated (there is .Updated property on it that counts up), we just write over the previous duration without writing the whole line again. If we put the duration after the assembly name we need to remember on which offset it is (same amount of work as before, just different value).

And we won't save the complexity of calculating how long that duration string is, because we need to "reserve" space for it if we want to always see it on the screen.

nohwnd:
Do you see any other complexity that could be avoided?

nohwnd:
There is also max console width, if you have super wide window we don't go past 250 chars or something.

Evangelink (approved):

PR #3683, opened 2024-08-23
Banner uses runtime framework and live test output uses target framework

### LINKED ISSUE #3669: Show targetted framework rather than effective runtime
## Describe the bug

image
image

On .NET and .NET Framework we are showing the information from  RuntimeInformation.FrameworkDescription. This is the actual runtime that is being used to run the executable, and not the target framework user specifies in project.

Let's assume the dlls tested were built targetting net6 (with roll forward = latest), and net472.

To avoid confusion and losing information I would show:

- in header -> the effective runtime net8.0 / net489
- next to dll: the targeted framework as defined in TargetFrameworkAttribute on the assembly in this case net6.0 / net472

nohwnd:
Fix #3669

image

Show the runtime version in header, and targeted framework in other places.

### DIFF
--- a/src/Platform/Microsoft.Testing.Platform/OutputDevice/TargetFrameworkParser.cs
+++ b/src/Platform/Microsoft.Testing.Platform/OutputDevice/TargetFrameworkParser.cs
@@ -7,8 +7,13 @@ namespace Microsoft.Testing.Platform.OutputDevice;
 
 internal static class TargetFrameworkParser
 {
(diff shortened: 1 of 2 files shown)


Evangelink (approved):
Thanks! Not now but it'd be good to have at least one acceptance for the display of the various tfms

Evangelink:
Force merging, native AOT tests are flaky.

PR #3687, opened 2024-08-23
Remove open port server mode

MarcoRossignoli:
After a discussion with @drognanar we decided to now allow to start the module in server mode with open port:
* it's not used at the moment
* it won't work well with passive node
* when we run in constrained environment anyway you cannot open a port but only connect back to an open socket


MarcoRossignoli:
I see some tests to fix for the info.

PR #3697, opened 2024-08-26
Check if protocol version is compatible for dotnet test

mariam-abdulla:


src/Platform/Microsoft.Testing.Platform/DotnetTest/DotnetTestConnection.cs
```diff
@@ -93,24 +93,27 @@ public async Task<bool> DoHandshakeAsync(string hostType)
...
 
         HandshakeInfo response = await _dotnetTestPipeClient.RequestReplyAsync<HandshakeInfo, HandshakeInfo>(handshakeInfo, _cancellationTokenSource.CancellationToken);
 
-        return response.Properties?.TryGetValue(HandshakeInfoPropertyNames.ProtocolVersion, out string? protocolVersion) == true &&
-            protocolVersion.Equals(ProtocolConstants.Version, StringComparison.Ordinal);
+        return response.Properties?.TryGetValue(HandshakeInfoPropertyNames.SupportedProtocolVersions, out string? protocolVersion) == true &&
+            IsVersionCompatible(protocolVersion, supportedProtocolVersions);
     }
 
+    public static bool IsVersionCompatible(string protocolVersion, string supportedProtocolVersions) => supportedProtocolVersions.Split(';').Contains(protocolVersion);
```
MarcoRossignoli, line 115: I think we don't need to check here but on client or?

MarcoRossignoli (approved):

src/Platform/Microsoft.Testing.Platform/DotnetTest/DotnetTestConnection.cs
MarcoRossignoli: never mind me

Issue #3689, opened 2024-08-23
[Honor `--no-build` `--no-restore` when we run under `dotnet run` with `dotnet test`] -> Don't use `dotnet run` but look for the correct muxer manually


MarcoRossignoli:
Ref https://github.com/microsoft/testfx/issues/3478#issuecomment-2307652724

MarcoRossignoli:
First attempt is failed https://github.com/microsoft/testfx/pull/3698#issuecomment-2312076040

At this point with the current setup of the default msbuild test sdk we can only reuse the arch switch infra we did for VSTest https://github.com/microsoft/vstest/blob/main/src/Microsoft.TestPlatform.CoreUtilities/Helpers/DotnetHostHelper.cs#L121 also if it means keep it up to date with the future runtime changes.

PR #3698, opened 2024-08-26
Honor `--no-build` when we use `dotnet run`

MarcoRossignoli:
closes https://github.com/microsoft/testfx/issues/3689

image

image

image

The `--no-restore` is not available as MSBuild parameter, but looks like the perf are anyway good.


Evangelink (approved):

MarcoRossignoli:
Ok this cannot work, actually the `VSTestNoBuild` is not linked to the `--no-build` but it's used to signal the "place" from where we expect the build was done and it's not more needed.

image

PR #3712, opened 2024-08-28
Bridge message logger also forward info to output

Evangelink:
Before this fix, info messages were only added to the log whereas warnings and errors were written both to log and output.

Fixing this allows to have MSTest parallelization message showing up:

image


MarcoRossignoli (approved):

Evangelink:
Native AOT tests are still flaky, force merging

Issue #3716, opened 2024-08-28
Open the `ExecuteRequestContext` ctor for better testability


MarcoRossignoli:
Ref https://github.com/microsoft/testfx/issues/3478#issuecomment-2315852176

PR #3717, opened 2024-08-28
Expose `ExecuteRequestContext` ctor for testability

MarcoRossignoli:
Closes https://github.com/microsoft/testfx/issues/3716

### DIFF
--- a/src/Platform/Microsoft.Testing.Platform/Requests/IRequestCompleteNotifier.cs
+++ b/src/Platform/Microsoft.Testing.Platform/Requests/IRequestCompleteNotifier.cs
@@ -1,0 +1,9 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+namespace Microsoft.Testing.Platform.Requests;
+
+public interface IRequestCompleteNotifier
+{
+    void Complete();
+}
--- a/src/Platform/Microsoft.Testing.Platform/Requests/SemaphoreSlimRequestCompleteNotifier.cs
+++ b/src/Platform/Microsoft.Testing.Platform/Requests/SemaphoreSlimRequestCompleteNotifier.cs
@@ -3,7 +3,7 @@
 
 namespace Microsoft.Testing.Platform.Requests;
 
-internal sealed class SemaphoreSlimRequestCompleteNotifier : IRequestCompleteNotifier
+internal sealed class SemaphoreSlimRequestCompleteNotifier : IExecuteRequestCompletionNotifier
 {
     private readonly SemaphoreSlim _semaphore;
 
--- a/src/Platform/Microsoft.Testing.Platform/Requests/SemaphoreSlimRequestCompleteNotifier.cs
+++ b/src/Platform/Microsoft.Testing.Platform/Requests/SemaphoreSlimRequestCompleteNotifier.cs
@@ -3,7 +3,7 @@
 
 namespace Microsoft.Testing.Platform.Requests;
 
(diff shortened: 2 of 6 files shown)


src/Platform/Microsoft.Testing.Platform/Requests/SemaphoreSlimRequestCompleteNotifier.cs
```diff
@@ -0,0 +1,17 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+namespace Microsoft.Testing.Platform.Requests;
+
+internal class SemaphoreSlimRequestCompleteNotifier : IRequestCompleteNotifier
```
Evangelink:
```suggestion
internal sealed class SemaphoreSlimRequestCompleteNotifier : IRequestCompleteNotifier
```

src/Platform/Microsoft.Testing.Platform/Requests/IRequestCompleteNotifier.cs
```diff
@@ -0,0 +1,9 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+namespace Microsoft.Testing.Platform.Requests;
+
+public interface IRequestCompleteNotifier
```
Evangelink: I would go with either `IExecuteRequestCompleteHandler` or `IExecuteRequestCompletionNotifier` (we could think of simplifying `ExecuteRequest` to simply `Request` but this is currently used only for that case so I like the precision).

nohwnd: `IExecuteRequestCompletionNotifier` is better imho. 

Or `IRequestCompletionNotifier`
`IRequestCompletionSource`

 (similar to Task**Completion**Source)

`IExecuteRequestCompleteHandler` I would expect this to be used when I implement the interface myself to provide as callback.

Evangelink: The goal is for @bradwilson to mock/implement this interface and check he is calling `Complete` at the expected time in his workflow.

nohwnd: I think that is all fine. The interface should communicate what the parameter is doing. and it is used by us to notify that we completed the request ( If I get the intent ).  To me  *CompletionNotifier communicates that point better than *CompleteHandler .

Evangelink: The framework is calling `Complete` on the request context object and this `Complete` calls the `IXXX` complete method that is responsible for what to do when receiving a complete call. The current implementation release a semaphore so we can see the implementation as both notifier (as it only notify the semaphore) and handler (as it handles the action).

I think `Handler` is better because from the point of view of the context, the implementer is responsible for handling completion "request"/call.

nohwnd: I will wait for @bradwilson or someone else's opinion since we don't agree. :)

MarcoRossignoli: To me it's fine `IExecuteRequestCompletionNotifier` this object is a "proxy" that notify to the platform that the request is completed.

Adapter -> Complete() -> [Proxy for testability] -> Platform

but I won't fight too much here.

MarcoRossignoli:
FYI atm tests are failing because we use that ctor inside the HotReload extension.

PR #3718, opened 2024-08-29
Fix Console is already in batching mode when writing to terminal

### LINKED ISSUE #3714: [Testing Platform] System.InvalidOperationException: Console is already in batching mode.
Package Version: 1.4.0-preview24422.11

Looking at the code, there isn't a lock around `TestProgressStateAwareTerminal` lines 138-140.

So I'm guessing that the `StartUpdate` there is called, and then the `ThreadProc` method repeats its loop, nothing has the `_lock` lock and that tries to call `StartUpdate` too and thus an exception is thrown.

Stacktrace:

```
Error: Unhandled exception. System.InvalidOperationException: Console is already in batching mode.
   at Microsoft.Testing.Platform.OutputDevice.Terminal.NonAnsiTerminal.StartUpdate() in /_/src/Platform/Microsoft.Testing.Platform/OutputDevice/Terminal/NonAnsiTerminal.cs:line 125
   at Microsoft.Testing.Platform.OutputDevice.Terminal.TestProgressStateAwareTerminal.ThreadProc() in /_/src/Platform/Microsoft.Testing.Platform/OutputDevice/Terminal/TestProgressStateAwareTerminal.cs:line 45
   at System.Threading.Thread.StartCallback()
```

nohwnd:
Fix #3714

### DIFF
--- a/src/Platform/Microsoft.Testing.Platform/OutputDevice/Terminal/TestProgressStateAwareTerminal.cs
+++ b/src/Platform/Microsoft.Testing.Platform/OutputDevice/Terminal/TestProgressStateAwareTerminal.cs
@@ -135,9 +135,12 @@ internal sealed partial class TestProgressStateAwareTerminal : IDisposable
         }
         else
         {
-            _terminal.StartUpdate();
-            write(_terminal);
-            _terminal.StopUpdate();
+            lock (_lock)
+            {
+                _terminal.StartUpdate();
+                write(_terminal);
+                _terminal.StopUpdate();
+            }
         }
     }
 


src/Platform/Microsoft.Testing.Platform/OutputDevice/Terminal/TestProgressStateAwareTerminal.cs
```diff
@@ -135,9 +135,12 @@ internal void WriteToTerminal(Action<ITerminal> write)
         }
         else
         {
-            _terminal.StartUpdate();
-            write(_terminal);
-            _terminal.StopUpdate();
+            lock (_lock)
```
MarcoRossignoli, line 138: FYI https://learn.microsoft.com/en-us/dotnet/api/system.threading.lock?view=net-9.0 but it's there starting from 9

MarcoRossignoli (approved):

Evangelink (approved):

Issue #3668, opened 2024-08-22
Add codefix for MSTEST0029


Evangelink:
https://learn.microsoft.com/dotnet/core/testing/mstest-analyzers/mstest0029

Let's add a fixer that:

1. Adds the `[TestMethod]` attribute
2. Change accessibility to `private`.

engyebrahim:
@Evangelink please let us know if there is a way to test the two code fixes as now we provide both but we test only one case as the test takes always the first registered fix

PR #3747, opened 2024-09-02
MSTEST0029 code fix

engyebrahim:
fix: https://github.com/microsoft/testfx/issues/3668


MarcoRossignoli (changes requested):
Can you add a test for the `private` visibility?

MarcoRossignoli (approved):

Issue #3740, opened 2024-09-01
Add properties to capture a possible test output to add to the `TestNode`


MarcoRossignoli:
@bradwilson is not using the VSTestBridge where we capture the standard out/err using old properties from TestCase (PR https://github.com/microsoft/testfx/pull/3486)

In the current `TestNode` properties list we don't have the counterpart for it https://github.com/microsoft/testfx/blob/main/src/Platform/Microsoft.Testing.Platform/Messages/TestNodeProperties.cs

The proposal TBD is to add

```cs
public record TextProperty(string Text)
public record TextErrorProperty(string Text)
```
 that an adapter author can use to add text information associated with a test.

I would prefer to not have in the name StdOut/Err because these properties could be used also without the console StdOut/Err. An adapter could have a different mechanism to generate this information. Let's say the idea is to have textual information.

Otherwise we should go for specific

```cs
public record ConsoleStandardOutputProperty(string Text)
public record ConsoleStandardErrorProperty(string Text)
```

cc: @Evangelink @nohwnd

PR #3748, opened 2024-09-02
Add `StandardOutputProperty` and `StandardErrorProperty`

MarcoRossignoli:
closes https://github.com/microsoft/testfx/issues/3740

image

To cherry pick to 1.4


docs/testingplatform/Source/TestingPlatformExplorer/TestingFramework/TestingFramework.Services.cs
```diff
@@ -0,0 +1,22 @@
...
+public interface ITestOutputHelper
+{
+    void WriteLine(string message);
+    void WriteErrorLine(string message);
+}
+
+internal sealed class TestOutputHelper : ITestOutputHelper
+{
+    public StringBuilder Output { get; set; } = new StringBuilder();
+    public StringBuilder Error { get; set; } = new StringBuilder();
```
Evangelink, line 17: Does it need to be settable?

MarcoRossignoli: Nope, it's a sample code and the class is internal we remove the set I can do some cleanup as follow-up

PR #3754, opened 2024-09-03
Fix missing skip reason

MarcoRossignoli:
image

cc: @bradwilson @drognanar

### DIFF
--- a/docs/testingplatform/Source/TestingPlatformExplorer/TestingFramework/TestingFramework.Attributes.cs
+++ b/docs/testingplatform/Source/TestingPlatformExplorer/TestingFramework/TestingFramework.Attributes.cs
@@ -6,6 +6,12 @@ namespace TestingPlatformExplorer.TestingFramework;
 [AttributeUsage(AttributeTargets.Method)]
 public class SkipAttribute : Attribute
 {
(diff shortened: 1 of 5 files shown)


docs/testingplatform/Source/TestingPlatformExplorer/TestingFramework/TestingFramework.Attributes.cs
```diff
@@ -6,6 +6,12 @@ namespace TestingPlatformExplorer.TestingFramework;
 [AttributeUsage(AttributeTargets.Method)]
 public class SkipAttribute : Attribute
 {
+    public string Reason { get; private set; }
```
Evangelink, line 9: @MarcoRossignoli this should be a readonly property, could you please do a follow-up.

MarcoRossignoli: FYI it's the sample code it's not shipped.

Evangelink: Ah sorry didn't pay attention.

PR #3784, opened 2024-09-07
Aggregate Inner Exception Messages and Stacktraces

### LINKED ISSUE #3783: InnerException's are lost in the Testing Platform output
Just been wrapping some exceptions to provide more context to users (e.g. A TearDown method failed, etc.) and noticed that the output is losing some context - And makes it almost impossible for a user to see where an actual crash was.

Reproduction: Wrap some exceptions in a new exception where the original exception is passed in as an `InnerException`.

Throw your wrapped exception, and you won't see any message or stacktrace from your original exception in the testing platform output.

Repro code:

```csharp
using Microsoft.Testing.Extensions;
using Microsoft.Testing.Platform.Builder;
using Microsoft.Testing.Platform.Capabilities;
using Microsoft.Testing.Platform.Capabilities.TestFramework;
using Microsoft.Testing.Platform.TestHost;

builder.RegisterTestFramework(_ => new TestFrameworkCapabilities(), (_, _) => new DummyAdapter());
var app = await builder.BuildAsync();
return await app.RunAsync();

{

    {
        try
        {
            MyService.DoSomething();
        }
        catch (Exception e)
        {
            {
                Uid = "2",
                DisplayName = "Blah",
                Properties = new PropertyBag(new FailedTestNodeStateProperty(e))
            }));
        }
     
        context.Complete();
        
        return Task.CompletedTask;
    }

}

public class MyService
{
    public static void DoSomething()
    {
        try
        {
            InnerDoSomething();
        }
        catch (Exception e)
        {
            throw new WrappedException("Service failed!", e);
        }
    }

    private static void InnerDoSomething()
    {
        throw new Exception("Error code 488");
    }
}

public class WrappedException(string message, Exception innerException) : Exception(message, innerException);
```

Actual output:
```
failed Blah (0ms)
  Service failed!
  Stack Trace:
    at MyService.DoSomething() in C:\git\TestPlatformSingleFileCoverageExceptionRepro\TestPlatformSingleFileCoverageExceptionRepro\TestPlatformSingleFileCoverageExceptionRepro\Program.cs:64
    at DummyAdapter.ExecuteRequestAsync(ExecuteRequestContext context) in C:\git\TestPlatformSingleFileCoverageExceptionRepro\TestPlatformSingleFileCoverageExceptionRepro\TestPlatformSingleFileCoverageExceptionRepro\Program.cs:34
```

Expected:

To include the message of `Error code 488` and the stacktrace location of `MyService.InnerDoSomething` on line `70` of the original exception.

thomhurst:
Fix #3783 

After: 
image

I don't know if this actually works... I was trying to run `build.cmd -pack -test -integrationTest` but I just get:

```
C:\git\testfx\artifacts\toolset\restore.proj : error MSB4242: SDK Resolver Failure: "The SDK resolver "Microsoft.DotNet
.MSBuildSdkResolver" failed while attempting to resolve the SDK "Microsoft.DotNet.Arcade.Sdk". Exception: "System.IO.Fi
leNotFoundException: Workload manifest microsoft.net.sdk.aspire: 8.0.1/8.0.100 from workload version 9.0.100-preview.7.
24407.12 was not installed. Running "dotnet workload repair" may resolve this.
```

And I don't know how to resolve it. I've installed the aspire workload and ran workload repair but still getting it.


MarcoRossignoli:
@nohwnd can you review?

src/Platform/Microsoft.Testing.Platform/OutputDevice/Terminal/TerminalTestReporter.cs
```diff
@@ -51,7 +51,7 @@ internal sealed partial class TerminalTestReporter : IDisposable
     private bool _wasCancelled;
 
 #if NET7_0_OR_GREATER
-    [GeneratedRegex(@$"^   at ((?<code>.+) in (?<file>.+):line (?<line>\d+)|(?<code1>.+))$", RegexOptions.ExplicitCapture, 1000)]
+    [GeneratedRegex(@$"^\s*at ((?<code>.+) in (?<file>.+):line (?<line>\d+)|(?<code1>.+))$", RegexOptions.ExplicitCapture, 1000)]
```
nohwnd, line 54: There should always be exactly 3 spaces in the stack frame, is this change needed or just done to make the regex more regexy?

src/Platform/Microsoft.Testing.Platform/OutputDevice/TerminalOutputDevice.cs
```diff
@@ -404,57 +408,73 @@ public async Task ConsumeAsync(IDataProducer dataProducer, IData value, Cancella
...
                         _terminalTestReporter.TestCompleted(
                              _assemblyName,
                              _targetFramework,
                              _shortArchitecture,
                              testNodeStateChanged.TestNode.DisplayName,
                              TestOutcome.Canceled,
                              duration,
-                             errorMessage: cancelledState.Exception?.Message ?? cancelledState.Explanation,
-                             cancelledState.Exception?.StackTrace,
+                             errorMessage: string.Join(_environment.NewLine, exceptionMessages),
```
nohwnd, line 476: the receiving api can be changed to take array of strings, (we use it both here and behind a remote call in `dotnet test` so we don't want exception objects). 

That will allow us to render the erros and stack traces under each other, so you don't have to jump back and forth.

nohwnd (commented):
We can merge as is (once the regex change question is answered). 

But my preferred way of presenting multiple errors would be more along the lines of:

Failed with 3 errors: 
[1] PlaygroundWrappedException: exception message
Stack Trace: 
    at ...
[2] Exception: exception message
Stack Trace: 
    at ...
[3] Exception: exception message
Stack Trace: 
    at ...
    
 
 
 What do you think?

src/Platform/Microsoft.Testing.Platform/OutputDevice/Terminal/TerminalTestReporter.cs
thomhurst: When testing I think the first line of inner exceptions weren't rendering properly without this change. Might be that I messed up formatting then

src/Platform/Microsoft.Testing.Platform/OutputDevice/TerminalOutputDevice.cs
thomhurst: Sounds good. Do you want to take over this pr because I can't actually build it locally so it makes Dev a bit harder

nohwnd: Sure.

PR #3796, opened 2024-09-10
Add option to order tests by name

Evangelink:
Following some recent discovery that the ordering capability of MSTest is only available in Test Explorer and provided by Test Explorer (discovered through a breaking change on Test Explorer side), we are introducing a new runsettings option to allow running tests in alpabetical order.

### DIFF
--- a/src/Adapter/MSTest.TestAdapter/Execution/TestExecutionManager.cs
+++ b/src/Adapter/MSTest.TestAdapter/Execution/TestExecutionManager.cs
@@ -407,7 +407,11 @@ public class TestExecutionManager
         bool hasAnyRunnableTests = false;
         var fixtureTests = new List<TestCase>();
 
-        foreach (TestCase currentTest in tests)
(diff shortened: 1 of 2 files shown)


sbloomztf:
@Evangelink - What is the syntax for setting this option in a RunSettings file?   My tests suddenly do not run in alpha order in the test explorer, and I am trying to use this new setting to restore this functionality.

Evangelink:
Sorry I am noticing I forgot to update the doc and should have been also more explicit here on usage:

```xml
<?xml version="1.0" encoding="utf-8"?>
<RunSettings>

  <MSTest>
    <OrderTestsByNameInClass>true</OrderTestsByNameInClass>
  </MSTest>

</RunSettings>
```

sbloomztf:
@Evangelink  - Thanks!  You saved the day!  My tests now run in alpha order again!

Evangelink:
So glad to hear!

Issue #3663, opened 2024-08-22
Add code fix for MSTEST0019


Evangelink:
https://learn.microsoft.com/dotnet/core/testing/mstest-analyzers/mstest0019

PR #3812, opened 2024-09-11
Add code fix for MSTEST0019

engyebrahim:
fix: https://github.com/microsoft/testfx/issues/3663

### DIFF
--- a/src/Analyzers/MSTest.Analyzers.CodeFixes/PreferTestInitializeOverConstructorFixer.cs
+++ b/src/Analyzers/MSTest.Analyzers.CodeFixes/PreferTestInitializeOverConstructorFixer.cs
@@ -1,0 +1,116 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System.Collections.Immutable;
+using System.Composition;
+using System.Xml.Linq;
+
+using Analyzer.Utilities;
+
+using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.CodeActions;
+using Microsoft.CodeAnalysis.CodeFixes;
+using Microsoft.CodeAnalysis.CSharp;
+using Microsoft.CodeAnalysis.CSharp.Syntax;
+using Microsoft.CodeAnalysis.Editing;
+using Microsoft.CodeAnalysis.Text;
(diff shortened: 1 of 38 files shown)


src/Analyzers/MSTest.Analyzers.CodeFixes/PreferTestInitializeOverConstructorFixer.cs
```diff
@@ -0,0 +1,115 @@
...
+
+    private static async Task<Document> ReplaceConstructorWithTestInitializeAsync(Document document, ConstructorDeclarationSyntax constructorDeclaration, CancellationToken cancellationToken)
+    {
+        DocumentEditor editor = await DocumentEditor.CreateAsync(document, cancellationToken).ConfigureAwait(false);
+
+        // Find the class containing the constructor
+        if (constructorDeclaration.Parent is ClassDeclarationSyntax containingClass)
+        {
+            // Check if a TestInitialize method already exists
+            MethodDeclarationSyntax? existingTestInitialize = containingClass.Members
```
Evangelink, line 71: As I pointed out previously, checking attributes (or any type) only based on syntax is not good because it can lead to errors (e.g. you have a `MyLib.TestInitializeAttribute`). The only valid way is to check symbols.

src/Analyzers/MSTest.Analyzers.CodeFixes/PreferTestInitializeOverConstructorFixer.cs
```diff
@@ -0,0 +1,115 @@
...
+                    .SelectMany(attrList => attrList.Attributes)
+                    .Any(attr => attr.Name.ToString().Contains("TestInitialize")));
+
+            // Move the body of the constructor
+            BlockSyntax? constructorBody = constructorDeclaration.Body;
+
+            if (existingTestInitialize != null)
+            {
+                // If TestInitialize exists, append the constructor body to it
+                StatementSyntax[]? constructorStatements = constructorBody?.Statements.ToArray();
```
Evangelink, line 84: What's the type of `Statements`? Do you really need to `ToArray`?

engyebrahim: it's a `SyntaxList` so it needs the `ToArray()`

PR #3837, opened 2024-09-14
use ReadLineAsync from polyfill

SimonCropp:


SimonCropp:
whats the failed test?

Evangelink (changes requested):
You have dropped too much code, the read is fully gone.

SimonCropp:
@Evangelink i think it is still there ? https://github.com/microsoft/testfx/pull/3837/files#diff-09ebded200ee21d2dd09a73b0b6410a3bb42afd5a601763ac2c4290b0a829477R90

Evangelink:
I am just blind... Sorry

Evangelink (approved):

Evangelink (approved):

Evangelink (commented):
The cancellation token seems to not be working as expected. The tests are failing because of some hangs.

Evangelink:
I can reproduce locally the hang, I'll try to have a look at what's hanging tomorrow.

SimonCropp:
perhaps a bug in my polyfill code?

Evangelink:
@SimonCropp I don't know if that's a bug per say but for sure your implementation only look at the token as the entry of the method:
image

while your implementation do check it during the execution.

Evangelink (changes requested):
The refactoring is causing a functional difference that can lead to hangs.

PR #3838, opened 2024-09-14
remove Validate and use Guard

SimonCropp:
Not sure on this one. IMO Validate does not provide enough value over using the guard approach (that is used everywhere else)

thoughts?


Evangelink (approved):

SimonCropp:
can u tell me what tests have failed?

Evangelink:
Git checkout was failing, I have restarted it.

Evangelink:
Attaching the test log: it seems to be linked to the `Guard.DirectoryExists` that throws exception during tests.

[MSTestAdapter.PlatformServices.UnitTests_net462_x64 (2).log](https://github.com/user-attachments/files/17056078/MSTestAdapter.PlatformServices.UnitTests_net462_x64.2.log)

Evangelink (approved):

PR #3842, opened 2024-09-14
remove some redundant lambda parens

SimonCropp:


Evangelink (approved):
Oh thanks! Such a shame we don't have a rule for this one...

SimonCropp:
for people who use rider/r#, i can craft you a dotsettings file that will make it an error

PR #3867, opened 2024-09-26
Fix localization of test adapter messages

nohwnd:
Copy the resource dlls into the correct localization folders for localized runs.

image

Fix https://devdiv.visualstudio.com/DevDiv/_workitems/edit/2221503?src=WorkItemMention&src-action=artifact_link


Evangelink (commented):
`netfx-netcore-netstandard` folder is using `CultureString` instead of `UICulture` https://github.com/microsoft/testfx/blob/main/src/Adapter/MSTest.TestAdapter/build/netfx-netcore-netstandard/MSTest.TestAdapter.targets#L71. I am not sure if these should be unified.

Also uwp is having similar bug as `net`. Could you port the fix there too?

nohwnd:
ah, I thought this is a common target that is used by all the tfms. will fix.

Evangelink (approved):

Evangelink:
@nohwnd once ready, could you please backport to rel/3.6 branch too?

PR #3869, opened 2024-09-26
Fix localization info to avoid translating workers and scope

nohwnd:
Other options say to not localize workers and scope, but not the banner.


Evangelink (approved):
Looking at the translations for languages I speak, I do agree that not translating these keywords is better. Thank you @nohwnd

Issue #3572, opened 2024-08-15
Add support for testconfig.json entries in MSTest


Evangelink:
## Summary

Add support for testconfig.json entries.

## Background and Motivation

`Microsoft.Testing.Platform` provides support for `[appname].testconfig.json` as a modern replacement for `runsettings`.

## Proposed Feature

Detect if the config exists in the file and use it.

If the user is specifying the config both in the runsettings and `testconfig.json` let's fail and ask the user to specify in only one file.

Given we have a new file format, I think this is a good opportunity to refactor the settings into logical groups.

Here is the list of currently supported properties

```text
AssemblyCleanupTimeout
AssemblyInitializeTimeout
AssemblyResolution
	<AssemblyResolution> <Directory path="D:\myfolder\bin\" includeSubDirectories="false"/> </AssemblyResolution>
CaptureTraceOutput
ClassCleanupLifecycle
ClassCleanupTimeout
ClassInitializeTimeout
DeleteDeploymentDirectoryAfterTestRunIsComplete
DeploymentEnabled
DeployTestSourceDependencies
EnableBaseClassTestMethodsFromOtherAssemblies
ForcedLegacyMode
MapInconclusiveToFailed
MapNotRunnableToFailed
Parallelize
	<Parallelize><Workers>32</Workers><Scope>MethodLevel</Scope></Parallelize>
SettingsFile
TestCleanupTimeout
TestInitializeTimeout
TestTimeout
TreatClassAndAssemblyCleanupWarningsAsErrors
TreatDiscoveryWarningsAsErrors
```

Here is the suggested refactored settings

```json
{
    "platformOptions" : {
	},
	"mstest" : {
		"timeout" : {
			"assemblyInitialize" : strictly positive int,
			"assemblyCleanup" : strictly positive int,
			"classInitialize" : strictly positive int,
			"classCleanup" : strictly positive int,
			"testInitialize" : strictly positive int,
			"testCleanup" : strictly positive int,
			"test" : strictly positive int
		},
		"parallelism" : {
			"enabled": true/false,
			"workers": positive int,
			"scope": method/class,
		},
		"deployment" : {
			"enabled": true/false,
			"deployTestSourceDependencies": true/false,
			"deleteDeploymentDirectoryAfterTestRunIsComplete": true/false
		}
		... remaining settings
	}
}
```

I cannot find any good naming for all the props where we chose if something should be warning or error.

[...]

PR #3872, opened 2024-09-26
Add support for testconfig.json entries in MSTest

engyebrahim:
fix: https://github.com/microsoft/testfx/issues/3572

```

// Expected format of the json is: -
//
// "mstest" : {
//  "timeout" : {
//      "assemblyInitialize" : strictly positive int,
//      "assemblyCleanup" : strictly positive int,
//      "classInitialize" : strictly positive int,
//      "classCleanup" : strictly positive int,
//      "testInitialize" : strictly positive int,
//      "testCleanup" : strictly positive int,
//      "test" : strictly positive int
 //      "useCooperativeCancellation" : true/false
//  },
//  "parallelism" : {
//      "enabled": true/false,
//      "workers": positive int,
//      "scope": method/class,
//  },
//  "execution": {
//     "collectSourceInformation": true,
//     "executionApartmentState": "STA",
//     "disableAppDomain": true,
//     "mapInconclusiveToFailed" : true,
//     "mapNotRunnableToFailed" : true,
//     "treatDiscoveryWarningsAsErrors" : true,
//     "considerEmptyDataSourceAsInconclusive" : true,
//     "treatClassAndAssemblyCleanupWarningsAsErrors" : true,
//     "considerFixturesAsSpecialTests" : true,
//  },
//  "deployment" : {
//       "enabled": true / false,
//       "deployTestSourceDependencies": true / false,
//       "deleteDeploymentDirectoryAfterTestRunIsComplete": true / false
//  },
//  "assemblyResolution" : [
//        { "path" : "..." , includeSubDirectories: "true" } ,
//        { "path" : "..." , includeSubDirectories: "true" } ,
//        { "path" : "..." , includeSubDirectories: "true" } ,
//        ...
//     ],
//  ... remaining settings
// }
```


src/Adapter/MSTest.TestAdapter/VSTestAdapter/MSTestDiscoverer.cs
```diff
@@ -17,6 +18,8 @@ namespace Microsoft.VisualStudio.TestPlatform.MSTest.TestAdapter;
 [FileExtension(".exe")]
 public class MSTestDiscoverer : ITestDiscoverer
 {
+    internal static IConfiguration? Configuration { get; private set; }
```
MarcoRossignoli: is the config read outside or inside a possible appdomain?

src/Adapter/MSTest.TestAdapter/VSTestAdapter/MSTestDiscoverer.cs
```diff
@@ -26,12 +29,15 @@ public class MSTestDiscoverer : ITestDiscoverer
...
-    public void DiscoverTests(IEnumerable<string> sources, IDiscoveryContext discoveryContext, IMessageLogger logger, ITestCaseDiscoverySink discoverySink)
+    public void DiscoverTests(IEnumerable<string> sources, IDiscoveryContext discoveryContext, IMessageLogger logger, ITestCaseDiscoverySink discoverySink) => MSTestDiscoverer.DiscoverTests(sources, discoveryContext, logger, discoverySink, null);
+
+    internal static void DiscoverTests(IEnumerable<string> sources, IDiscoveryContext discoveryContext, IMessageLogger logger, ITestCaseDiscoverySink discoverySink, IConfiguration? configuration)
     {
         ValidateArg.NotNull(sources, "sources");
         ValidateArg.NotNull(logger, "logger");
         ValidateArg.NotNull(discoverySink, "discoverySink");
 
+        Configuration = configuration;
```
MarcoRossignoli: is it static needed? can't we flow the `IConfiguration` down?

engyebrahim: I just see it could be long from three diff places.. could it have bad effect?

src/Adapter/MSTest.TestAdapter/MSTestSettings.cs
```diff
@@ -265,6 +266,11 @@ public static void PopulateSettings(MSTestSettings settings)
     internal static void PopulateSettings(IDiscoveryContext? context, IMessageLogger? logger)
```
Evangelink: I would expect to see the `IConfiguration?` as part of the method signature.

engyebrahim:
we have that by default generated 
image

so as runsettings file could have some other settings more than mstest should i change the check to search on the key 'key' and throw on the case of it has it and we has it in configration but other than tat we okay two have two files!

engyebrahim:
i didn't use MSTESTv2 key
and from the json example that Amaury added in the task i didn't use that
`"deployment" : {
       "enabled": true/false,
        "deployTestSourceDependencies": true/false
       ,"deleteDeploymentDirectoryAfterTestRunIsComplete": true/false		
       }`
as i didn't find in mstesSettings any deployment settings

engyebrahim:
Directory working i  need more changes from vstest

engyebrahim:
check also visual studio default setting

src/Adapter/MSTest.TestAdapter/MSTestSettings.cs
```diff
@@ -262,32 +264,49 @@ public static void PopulateSettings(MSTestSettings settings)
...
-            CurrentSettings = new MSTestSettings();
-            return;
+            throw new InvalidOperationException(Resource.DuplicateConfigurationError);
         }
 
-        MSTestSettings? aliasSettings = GetSettings(context.RunSettings.SettingsXml, SettingsNameAlias, logger);
-
-        // If a user specifies MSTestV2 in the runsettings, then prefer that over the v1 settings.
-        if (aliasSettings != null)
+        if (context?.RunSettings != null && !StringEx.IsNullOrEmpty(context.RunSettings.SettingsXml))
```
MarcoRossignoli: Is it the `!StringEx.IsNullOrEmpty(context.RunSettings.SettingsXml)` needed? 
If `context?.RunSettings != null` isn't enough?

src/Adapter/MSTest.TestAdapter/MSTestSettings.cs
```diff
@@ -262,32 +264,49 @@ public static void PopulateSettings(MSTestSettings settings)
     /// <param name="logger"> The logger for messages. </param>
     /// The discovery context that contains the runsettings.
     /// </param>
-    internal static void PopulateSettings(IDiscoveryContext? context, IMessageLogger? logger)
+    internal static void PopulateSettings(IDiscoveryContext? context, IMessageLogger? logger, IConfiguration? configuration)
     {
-        RunConfigurationSettings = RunConfigurationSettings.PopulateSettings(context);
-
-        if (context?.RunSettings == null || StringEx.IsNullOrEmpty(context.RunSettings.SettingsXml))
+        if (configuration != null && context?.RunSettings != null)
```
MarcoRossignoli: We already discussed this one related to the possible `RunConfig` usage by MSTest.

src/Adapter/MSTest.TestAdapter/MSTestSettings.cs
```diff
@@ -262,32 +264,49 @@ public static void PopulateSettings(MSTestSettings settings)
...
 
             CurrentSettings = settings ?? new MSTestSettings();
+
+            return;
         }
 
-        SetGlobalSettings(context.RunSettings.SettingsXml, CurrentSettings, logger);
+        // This will contain default adapter settings
+        CurrentSettings = new MSTestSettings();
+        return;
```
MarcoRossignoli: not needed right?

src/Adapter/MSTest.TestAdapter/MSTestSettings.cs
```diff
@@ -831,4 +850,162 @@ private static void SetGlobalSettings(
...
+        //  "parallelism" : {
+        //      "enabled": true/false,
+        //      "workers": positive int,
+        //      "scope": method/class,
+        //  },
+        //  ... remaining settings
+        // }
+        MSTestSettings settings = new();
+
+        ParseBooleanSetting(configuration, "captureTraceOutput", logger, result => settings.CaptureDebugTraces = result);
```
MarcoRossignoli: I would rename `result` into `value` looks to me clearer what you think?

src/Adapter/MSTest.TestAdapter/MSTestSettings.cs
```diff
@@ -831,4 +850,162 @@ private static void SetGlobalSettings(
...
+            else
+            {
+                throw new AdapterSettingsException(string.Format(
+                    CultureInfo.CurrentCulture,
+                    Resource.InvalidClassCleanupLifecycleValue,
+                    classCleanupLifecycle,
+#if NET
+                    string.Join(", ", Enum.GetNames<ClassCleanupBehavior>())));
+#else
+                    string.Join(", ", EnumPolyfill.GetNames<ClassCleanupBehavior>())));
```
MarcoRossignoli, line 984: Can't we use polyfill for both? @Evangelink how's the story of polyfill here?

src/Adapter/MSTest.TestAdapter/MSTestSettings.cs
```diff
@@ -831,4 +850,162 @@ private static void SetGlobalSettings(
...
+#endif
+            }
+        }
+
+        if (configuration["mstest:parallelism:workers"] is string workers)
+        {
+            settings.ParallelizationWorkers = int.TryParse(workers, out int parallelWorkers)
```
MarcoRossignoli, line 991: wonder if we should warning a message in case the value is specified but actually is not a good int to parse.
Fine go on with the default proc count, but looks an important settings that users want to be correct and it's impacting the runtime.

src/Adapter/MSTest.TestAdapter/RunConfigurationSettings.cs
```diff
@@ -155,4 +159,39 @@ private static RunConfigurationSettings ToSettings(XmlReader reader)
...
+        // }
+
+        // Initialize settings object
+        RunConfigurationSettings settings = new();
+
+        if (bool.TryParse(configuration["mstest:runConfiguration:collectSourceInformation"], out bool collectSourceInformation))
+        {
+            settings.CollectSourceInformation = collectSourceInformation;
+            PlatformServiceProvider.Instance.AdapterTraceLogger.LogInfo(
+                "CollectSourceInformation value Found : {0}", collectSourceInformation);
```
MarcoRossignoli:
```suggestion
                "CollectSourceInformation value found : {0}", collectSourceInformation);
```

src/Adapter/MSTest.TestAdapter/RunConfigurationSettings.cs
```diff
@@ -155,4 +159,39 @@ private static RunConfigurationSettings ToSettings(XmlReader reader)
...
+        }
+
+        string? apartmentStateValue = configuration["mstest:runConfiguration:executionApartmentState"];
+        if (!string.IsNullOrEmpty(apartmentStateValue) &&
```
MarcoRossignoli: I think that the `Enum.TryParse` is enough, I suppose that if you pass a null to it will return `false`.

src/Adapter/MSTestAdapter.PlatformServices/Services/SettingsProvider.cs
```diff
@@ -40,6 +41,15 @@ public static MSTestAdapterSettings Settings
     public static void Reset() => s_settings = null;
 #endif
 
+    internal static void Load(IConfiguration configuration)
+    {
+#if !WINDOWS_UWP
+#pragma warning disable IDE0022 // Use expression body for method
```
MarcoRossignoli, line 47: I think that this pragma can be removed or?

MarcoRossignoli (commented):
Some question genereally looks fine

src/Adapter/MSTest.TestAdapter/MSTestSettings.cs
engyebrahim: it was like that in old code, so i think it was for a reason

src/Adapter/MSTest.TestAdapter/MSTestSettings.cs
engyebrahim: it's needed if nor runsettings or config files are there o set the default settings

src/Adapter/MSTestAdapter.PlatformServices/Services/SettingsProvider.cs
engyebrahim: when i remove it it request an arrow function and when i do it I get another error because of `#if`

src/Adapter/MSTest.TestAdapter/MSTestSettings.cs
engyebrahim: yup we are already throwing in case of invalid value

engyebrahim:
needs to add/fet spacific settings from confic that related to some frameworks like appdomain

src/Adapter/MSTest.TestAdapter/RunConfigurationSettings.cs
engyebrahim: when i remove it some tests fails

src/Adapter/MSTest.TestAdapter/MSTestSettings.cs
```diff
@@ -831,4 +906,162 @@ private static void SetGlobalSettings(
...
+        //      "workers": positive int,
+        //      "scope": method/class,
+        //  },
+        //  ... remaining settings
+        // }
+        MSTestSettings settings = new();
+
+        ParseBooleanSetting(configuration, "captureTraceOutput", logger, value => settings.CaptureDebugTraces = value);
+        ParseBooleanSetting(configuration, "enableBaseClassTestMethodsFromOtherAssemblies", logger, value => settings.EnableBaseClassTestMethodsFromOtherAssemblies = value);
+        ParseBooleanSetting(configuration, "forcedLegacyMode", logger, value => settings.ForcedLegacyMode = value);
```
Evangelink: Let's not support this setting.

src/Adapter/MSTest.TestAdapter/MSTestSettings.cs
```diff
@@ -831,4 +906,162 @@ private static void SetGlobalSettings(
...
+        ParseBooleanSetting(configuration, "mapInconclusiveToFailed", logger, value => settings.MapInconclusiveToFailed = value);
+        ParseBooleanSetting(configuration, "mapNotRunnableToFailed", logger, value => settings.MapNotRunnableToFailed = value);
+        ParseBooleanSetting(configuration, "treatDiscoveryWarningsAsErrors", logger, value => settings.TreatDiscoveryWarningsAsErrors = value);
+        ParseBooleanSetting(configuration, "considerEmptyDataSourceAsInconclusive", logger, value => settings.ConsiderEmptyDataSourceAsInconclusive = value);
+        ParseBooleanSetting(configuration, "treatClassAndAssemblyCleanupWarningsAsErrors", logger, value => settings.TreatClassAndAssemblyCleanupWarningsAsErrors = value);
+        ParseBooleanSetting(configuration, "considerFixturesAsSpecialTests", logger, value => settings.ConsiderFixturesAsSpecialTests = value);
+        ParseBooleanSetting(configuration, "cooperativeCancellationTimeout", logger, value => settings.CooperativeCancellationTimeout = value);
```
Evangelink: This should be part of the timeout section and let's use `useCooperativeCancellation`

src/Adapter/MSTest.TestAdapter/MSTestSettings.cs
```diff
@@ -831,4 +906,162 @@ private static void SetGlobalSettings(
...
+
+        ParseIntegerSetting(configuration, "timeout:test", logger, value => settings.TestTimeout = value);
+        ParseIntegerSetting(configuration, "timeout:assemblyCleanup", logger, value => settings.AssemblyCleanupTimeout = value);
+        ParseIntegerSetting(configuration, "timeout:assemblyInitialize", logger, value => settings.AssemblyInitializeTimeout = value);
+        ParseIntegerSetting(configuration, "timeout:classInitialize", logger, value => settings.ClassInitializeTimeout = value);
+        ParseIntegerSetting(configuration, "timeout:classCleanup", logger, value => settings.ClassCleanupTimeout = value);
+        ParseIntegerSetting(configuration, "timeout:testInitialize", logger, value => settings.TestInitializeTimeout = value);
+        ParseIntegerSetting(configuration, "timeout:testCleanup", logger, value => settings.TestCleanupTimeout = value);
+
+        ParseStringSetting(configuration, "settingsFile", logger, fileName => settings.TestSettingsFile = fileName);
```
Evangelink: I think we also don't want to support this one. We are in the new world so there is no point in allowing to load an old setting file

src/Adapter/MSTest.TestAdapter/MSTestSettings.cs
```diff
@@ -253,7 +257,61 @@ public static void PopulateSettings(MSTestSettings settings)
     /// The discovery context that contains the runsettings.
     /// </param>
     [Obsolete("this function will be removed in v4.0.0")]
-    public static void PopulateSettings(IDiscoveryContext? context) => PopulateSettings(context, null);
+    public static void PopulateSettings(IDiscoveryContext? context) => PopulateSettings(context, null, null);
+
+    private static bool IsRunSettingsFileHesMSTestSettings(string? runSettingsXml) => IsRunSettingsFileHasSettingName(runSettingsXml, SettingsName, out _) || IsRunSettingsFileHasSettingName(runSettingsXml, SettingsNameAlias, out _)
+                    || (IsRunSettingsFileHasSettingName(runSettingsXml, ConfigurationName, out XmlReader? runConfigrationReader) && HasRelevantKeys(runConfigrationReader));
+
+    private static bool HasRelevantKeys(XmlReader? reader)
```
Evangelink: This seems to only have sense in the cotnext above so I would make it local function of the previous or rename it to something more menaingfull

src/Adapter/MSTest.TestAdapter/MSTestSettings.cs
```diff
@@ -831,4 +906,162 @@ private static void SetGlobalSettings(
...
+        //  "parallelism" : {
+        //      "enabled": true/false,
+        //      "workers": positive int,
+        //      "scope": method/class,
+        //  },
+        //  ... remaining settings
+        // }
+        MSTestSettings settings = new();
+
+        ParseBooleanSetting(configuration, "captureTraceOutput", logger, value => settings.CaptureDebugTraces = value);
```
Evangelink: I would have it under some `output` section.

src/Adapter/MSTest.TestAdapter/MSTestSettings.cs
```diff
@@ -831,4 +906,162 @@ private static void SetGlobalSettings(
...
+        ParseBooleanSetting(configuration, "enableBaseClassTestMethodsFromOtherAssemblies", logger, value => settings.EnableBaseClassTestMethodsFromOtherAssemblies = value);
+        ParseBooleanSetting(configuration, "forcedLegacyMode", logger, value => settings.ForcedLegacyMode = value);
+        ParseBooleanSetting(configuration, "mapInconclusiveToFailed", logger, value => settings.MapInconclusiveToFailed = value);
+        ParseBooleanSetting(configuration, "mapNotRunnableToFailed", logger, value => settings.MapNotRunnableToFailed = value);
+        ParseBooleanSetting(configuration, "treatDiscoveryWarningsAsErrors", logger, value => settings.TreatDiscoveryWarningsAsErrors = value);
+        ParseBooleanSetting(configuration, "considerEmptyDataSourceAsInconclusive", logger, value => settings.ConsiderEmptyDataSourceAsInconclusive = value);
+        ParseBooleanSetting(configuration, "treatClassAndAssemblyCleanupWarningsAsErrors", logger, value => settings.TreatClassAndAssemblyCleanupWarningsAsErrors = value);
+        ParseBooleanSetting(configuration, "considerFixturesAsSpecialTests", logger, value => settings.ConsiderFixturesAsSpecialTests = value);
```
Evangelink: Let's group them under some `execution` category as suggested by @nohwnd

src/Adapter/MSTest.TestAdapter/MSTestSettings.cs
Evangelink: I would need to double check but yeah most likely it should work.

src/Adapter/MSTest.TestAdapter/MSTestSettings.cs
```diff
@@ -262,32 +290,46 @@ public static void PopulateSettings(MSTestSettings settings)
     /// <param name="logger"> The logger for messages. </param>
     /// The discovery context that contains the runsettings.
     /// </param>
-    internal static void PopulateSettings(IDiscoveryContext? context, IMessageLogger? logger)
+    internal static void PopulateSettings(IDiscoveryContext? context, IMessageLogger? logger, IConfiguration? configuration)
     {
-        RunConfigurationSettings = RunConfigurationSettings.PopulateSettings(context);
-
-        if (context?.RunSettings == null || StringEx.IsNullOrEmpty(context.RunSettings.SettingsXml))
+        if (configuration != null && context?.RunSettings != null && IsRunSettingsFileHasMSTestSettings(context?.RunSettings.SettingsXml))
```
Evangelink:
```suggestion
        if (configuration != null && context?.RunSettings != null && IsRunSettingsFileHasMSTestSettings(context.RunSettings.SettingsXml))
```

src/Adapter/MSTest.TestAdapter/MSTestSettings.cs
Evangelink: Simplification looks fine to me:

```suggestion
        if (!StringEx.IsNullOrEmpty(context?.RunSettings.SettingsXml))
```

src/Adapter/MSTest.TestAdapter/MSTestSettings.cs
```diff
@@ -831,4 +873,155 @@ private static void SetGlobalSettings(
...
+        {
+            if (bool.TryParse(value, out bool result))
+            {
+                setSetting(result);
+            }
+            else
+            {
+                logger?.SendMessage(TestMessageLevel.Warning, string.Format(CultureInfo.CurrentCulture, Resource.InvalidValue, value, key));
+            }
+        }
```
Evangelink: Nit: I find it easier mentally when there is less indenting

```suggestion
        if (configuration[$"mstest:{key}"] is not string value)
        {
            return;
        }
        
        if (bool.TryParse(value, out bool result))
        {
            setSetting(result);
        }
        else
        {
            logger?.SendMessage(TestMessageLevel.Warning, string.Format(CultureInfo.CurrentCulture, Resource.InvalidValue, value, key));
        }
```

src/Adapter/MSTest.TestAdapter/Resources/Resource.resx
```diff
@@ -414,4 +414,7 @@ but received {4} argument(s), with types '{5}'.</value>
   <data name="DynamicDataShouldBeValidMessageFormat_MemberType" xml:space="preserve">
     <value>'[DynamicData]' referenced member '{0}.{1}' should return 'IEnumerable&lt;object[]&gt;', 'IEnumerable&lt;Tuple&gt;` or 'IEnumerable&lt;ValueTuple&gt;'</value>
   </data>
+  <data name="DuplicateConfigurationError" xml:space="preserve">
+    <value>You are using both 'runsettings' and 'testconfig' files. Please use only one type of configuration file.</value>
```
Evangelink: Both '.runsettings' and '.testconfig.json' files have been detected. Please select only one of these test configuration files.

src/Adapter/MSTest.TestAdapter/RunConfigurationSettings.cs
```diff
@@ -155,4 +157,34 @@ private static RunConfigurationSettings ToSettings(XmlReader reader)
...
+        }
+
+        string? apartmentStateValue = configuration["mstest:execution:executionApartmentState"];
+        if (Enum.TryParse(apartmentStateValue, out PlatformApartmentState platformApartmentState))
+        {
+            settings.ExecutionApartmentState = platformApartmentState switch
+            {
+                PlatformApartmentState.STA => ApartmentState.STA,
+                PlatformApartmentState.MTA => ApartmentState.MTA,
+                _ => throw new NotSupportedException($"Platform apartment state '{platformApartmentState}' is not supported."),
```
Evangelink: Exception should be localized, please reuse the exception we have when reading the runsettings.

src/Adapter/MSTest.TestAdapter/TestingPlatformAdapter/MSTestBridgedTestFramework.cs
```diff
@@ -8,17 +8,24 @@
...
 {
+    private readonly IConfiguration? _configration;
+
     public MSTestBridgedTestFramework(MSTestExtension mstestExtension, Func<IEnumerable<Assembly>> getTestAssemblies,
         IServiceProvider serviceProvider, ITestFrameworkCapabilities capabilities)
         : base(mstestExtension, getTestAssemblies, serviceProvider, capabilities)
     {
+        // check if it have mstest settings
+        IConfiguration config = serviceProvider.GetConfiguration();
+        _configration = config["mstest"] is null ? null : config;
```
Evangelink: Isn't it better to always pass down the file and let the settings class detect if `mstest` section exists?

src/Adapter/MSTestAdapter.PlatformServices/MSTestAdapter.PlatformServices.csproj
```diff
@@ -31,6 +31,7 @@
 
   <ItemGroup>
     <ProjectReference Include="$(RepoRoot)src\TestFramework\TestFramework.Extensions\TestFramework.Extensions.csproj" />
+    <ProjectReference Include="$(RepoRoot)src\Platform\Microsoft.Testing.Extensions.VSTestBridge\Microsoft.Testing.Extensions.VSTestBridge.csproj" />
```
Evangelink, line 34: Why is this needed?

src/Adapter/MSTestAdapter.PlatformServices/Services/MSTestAdapterSettings.cs
```diff
@@ -304,5 +360,32 @@ private void ReadAssemblyResolutionPath(XmlReader reader)
...
+        // "mstest" : {
+        //    "assemblyResolution" : [
+        //        { "path" : "..." , includeSubDirectories: "true" } ,
+        //        { "path" : "..." , includeSubDirectories: "true" } ,
+        //        { "path" : "..." , includeSubDirectories: "true" } ,
+        //        ...
+        //     ]
+        //  ... remaining settings
+        // }
+        int indx = 0;
```
Evangelink: Let's use the full name `index`.

test/IntegrationTests/MSTest.Acceptance.IntegrationTests/ConfigurationSettingsTests.cs
```diff
@@ -0,0 +1,202 @@
...
+    [ArgumentsProvider(nameof(TargetFrameworks.All), typeof(TargetFrameworks))]
+    public async Task TestConfigJson_AndRunSettingsWithoutMstest_OverrideRunConfigration(string tfm)
+    {
+        var testHost = TestHost.LocateFrom(_testAssetFixture.ProjectPath, TestAssetFixture.ProjectName, tfm);
+        TestHostResult testHostResult = await testHost.ExecuteAsync("--settings my.runsettings");
+
+        // Assert
+        testHostResult.AssertExitCodeIs(ExitCodes.Success);
+    }
+
```
Evangelink, line 49: Maybe add a last test with no runsettings provided.

Evangelink (approved):

src/Adapter/MSTestAdapter.PlatformServices/MSTestAdapter.PlatformServices.csproj
engyebrahim: to use the Iconfiguration from the bridge

Evangelink (approved):

Issue #3824, opened 2024-09-13
MSTEST0036 is shown for cases where no shadowing happens


cremor:
## Describe the bug
The following code shows MSTEST0036 (Do not use shadowing inside test class) even though no shadowing happens.

## Steps To Reproduce
```cs
namespace TestProject1;

[TestClass]
public class BaseTest
{
    protected TObject CreateObject<TObject>()
    {
        throw new NotImplementedException();
    }
}

[TestClass]
public class ExtendedTest : BaseTest
{
    // MSTEST0036 is shown here
    private SomeType CreateObject()
    {
        throw new NotImplementedException();
    }
}

public class SomeType;
```

## Expected behavior
No MSTEST0036 warning.

## Actual behavior
MSTEST0036 is shown.

## Additional context
This is a new problem in v3.6.0

I actually see MSTEST0036 twice in the error list - with different messages. It seems like one is from IntelliSense and the other one from actually building the project.

And the message "is already exist" should be fixed to be correct English.

![grafik](https://github.com/user-attachments/assets/8743f258-2d59-4b3e-990a-97ac1216bd77)

Evangelink:
Hi @cremor!

Thanks for the bug report, indeed there are a few problems. We will work on them.

PR #3881, opened 2024-09-30
Fix MSTEST0036 is shown for cases where no shadowing happens

engyebrahim:
fix: https://github.com/microsoft/testfx/issues/3824

### DIFF
--- a/src/Analyzers/MSTest.Analyzers/DoNotUseShadowingAnalyzer.cs
+++ b/src/Analyzers/MSTest.Analyzers/DoNotUseShadowingAnalyzer.cs
@@ -112,7 +112,7 @@ public sealed class DoNotUseShadowingAnalyzer : DiagnosticAnalyzer
         }
 
         // Compare methods
-        if (member is IMethodSymbol methodSymbol && baseMember is IMethodSymbol baseMethodSymbol)
+        if (member is IMethodSymbol methodSymbol && baseMember is IMethodSymbol baseMethodSymbol && methodSymbol.IsGenericMethod == baseMethodSymbol.IsGenericMethod)
         {
             return methodSymbol.Name == baseMethodSymbol.Name &&
                    methodSymbol.Parameters.Length == baseMethodSymbol.Parameters.Length &&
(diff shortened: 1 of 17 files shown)


Evangelink (commented):
We have another FP when the methods are private (add the same private method in current and base and you will see a false positive).

Evangelink (approved):

src/Analyzers/MSTest.Analyzers/DoNotUseShadowingAnalyzer.cs
```diff
@@ -112,7 +112,8 @@ private static bool IsMemberShadowing(ISymbol member, ISymbol baseMember)
         }
 
         // Compare methods
-        if (member is IMethodSymbol methodSymbol && baseMember is IMethodSymbol baseMethodSymbol)
+        if (member is IMethodSymbol methodSymbol && baseMember is IMethodSymbol baseMethodSymbol && methodSymbol.IsGenericMethod == baseMethodSymbol.IsGenericMethod
+            && !(methodSymbol.DeclaredAccessibility == Accessibility.Private && baseMethodSymbol.DeclaredAccessibility == Accessibility.Private))
```
cremor, line 116: It also can't be shadowing if only one of the methods is private, or can it?

Evangelink: You are right, I find the syntax confusing but @engyebrahim added the not before the parenthesis so the and clause becomes a or :)

cremor: - private and private: `!(true && true) = false` (ok)
- public and public: `!(false && false) = true` (ok)
- private and public: `!(true && false) = true` (not ok, is not shadowing)

And I think the visibility check also needs to happen in the "Compare properties" section below.

Evangelink: Oh yeah good point, I overlooked last case. @engyebrahim can you please add more tests and probably fix the remaining broken part?

engyebrahim: public&public raises
public&private raises
image

private&private doesn't raise 
private&public doesn't raise -> `fixing that case`
image

Issue #3822, opened 2024-09-12
CollectionAssert.AreEqual fails for list of strings using IEqualityComparer following #3275


Smenus:
## Describe the bug
Following #3275, comparing two lists of strings (`List<string>`) fails when an `IEqualityComparer` is passed, with the intention of ignoring case for example.

## Steps To Reproduce
Setup: create a MSTest project. Put following code in a sample test:
```csharp
List<string> expected = ["one", "two"];
List<string> actual   = ["ONE", "tWo"];

CollectionAssert.AreEqual(expected, actual, StringComparer.OrdinalIgnoreCase);
```

## Expected behavior
Test passes.

## Actual behavior
Test fails.

## Additional context
From looking at the code, I expect it's because strings are `IEnumerable`.

PR #3886, opened 2024-09-30
Fix CollectionAssert.AreEqual fails for list of strings using IEqualityComparer following

engyebrahim:
fix: https://github.com/microsoft/testfx/issues/3822

### DIFF
--- a/test/UnitTests/TestFramework.UnitTests/Assertions/CollectionAssertTests.cs
+++ b/test/UnitTests/TestFramework.UnitTests/Assertions/CollectionAssertTests.cs
@@ -198,6 +198,13 @@ public class CollectionAssertTests : TestContainer
         comparer.ToString(); // no warning
     }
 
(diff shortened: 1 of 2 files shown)


test/UnitTests/TestFramework.UnitTests/Assertions/CollectionAssertTests.cs
```diff
@@ -198,6 +198,13 @@ public void CollectionAssertAreEqualComparerNullabilityPostConditions()
         comparer.ToString(); // no warning
     }
 
+    public void CollectionAssertAreEqual_WithIgnoreCaseComparer_DoesNotThrow()
```
Evangelink, line 201: Do we have a case similar to this one but where we would use a `StringComparer` that doesn't ignroe casing?

Evangelink (dismissed):

Evangelink (approved):

Issue #3751, opened 2024-09-03
Add year, minute to the `--diagnostic` filename


AliveDevil:
## Describe the bug

The `--diagnostic`-option creates files with an opaque, non-modified date sorting friendly file name.

## Steps To Reproduce

1. Run multiple successive tests with ` --diagnostic`, when process Ids are recycled, the name sorting option is useless.

## Expected behavior
Some name sorting friendly file prefix (i.e. `log_yyyymmddThhmmss_ProcessId.diag`).

## Actual behavior
![Screenshot_20240903_125137](https://github.com/user-attachments/assets/bc82a6be-0256-497c-802c-d64c94a794aa)
```
log_09030958760.diag: 11:47
log_09031011248.diag: 12:30
log_09031028245.diag: 12:34
log_09031043246.diag: 12:17
log_09031043969.diag: 12:10
log_09031045242.diag: 12:34
log_09031055355.diag: 12:09
```

nohwnd:
Adding to next milestone, looks like low hanging fruit, and I don't think change of name in diag log should be considered a breaking change.

MarcoRossignoli:
> ? Some name sorting friendly file prefix (i.e. log_yyyymmddThhmmss_ProcessId.diag)

Fine to me a format like that, keeping in mind that you should not take the "process_id" as guarantee for nothing, that id can be reused for different processes, maybe we can simply have time ordering, anyway today we have a retry loop if the file is already taken by someone else in the same folder so the order will be preserved and the pid said nothing about the nature of the file, maybe we can add the tfm the make easy to read it.

MarcoRossignoli:
@AliveDevil I did a check of the code and we're creating the file with the pattern 

`$"{_options.LogPrefixName}_{_clock.UtcNow.ToString("MMddHHssfff", CultureInfo.InvariantCulture)}.diag";`

You can change the prefix using the option `--diagnostic-output-fileprefix` as explained here https://learn.microsoft.com/en-us/dotnet/core/testing/unit-testing-platform-intro?tabs=dotnetcli#options

Now the logs are already "ordered", it's not clear to me what's the difference if we change the file name pattern.

Can you elaborate why the current stable ordering is not enough? You'll have always the log files ordered and adding the PID doesn't change much, the time granularity is small enough to make it useless for ordering scope.

AliveDevil:
When the format really is:
- Two Digit Month
- Two Digit Day
- Two Digit Hour
- Two Digit Seconds
[...]

PR #3894, opened 2024-09-30
Add yy/mm to the log filename for better ordering

MarcoRossignoli:
closes https://github.com/microsoft/testfx/issues/3751

### DIFF
--- a/src/Platform/Microsoft.Testing.Platform/Logging/FileLogger.cs
+++ b/src/Platform/Microsoft.Testing.Platform/Logging/FileLogger.cs
@@ -118,7 +118,7 @@ internal sealed class FileLogger : IDisposable
 
             try
             {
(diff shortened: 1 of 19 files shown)


MarcoRossignoli:
> Would be nice to have the year as yyyy, to make it easier to see that the first entry is a year.

I don't see 2 more char useful if not making the filename longer, the pattern is explained in the `--help`. I think that as first step is enough, after I agree that we could add a way to have composable name including placeholder for date, tfm, pid and other interesting information.
Also I would not advice to enabled always the logs, we don't have scenario where "real issue" are not crashing or failing the execution(no try/catch/go on) and so without log is hard to understand what's happened.

We can followup on it in future if we will have other user request. This one looks ok, the order by name is wrong.

PR #3920, opened 2024-10-08
Display inner exceptions

### LINKED ISSUE #3783: InnerException's are lost in the Testing Platform output

Evangelink:
Fixes #3783

Handles inner exceptions, aggregate exceptions and mix of both:

image
image
image


thomhurst:
Thanks for starting to take a look at this! Is this on pause right now? Finding it very difficult to debug wrapped exceptions currently

Evangelink:
Hi @thomhurst,

This is paused because we were discussing about the "presentation" of the data as to whether going closer to default runtime (knowing the display is different by runtime so we might be "always different") or to go with a custom look.

I also have to handle some more urgent tasks and discussions but will make sure this gets completed as soon as possible. FYI I'll include this in the 1.4.2 release so the wait period is as small as possible.

thomhurst:
> Hi @thomhurst,
> 
> This is paused because we were discussing about the "presentation" of the data as to whether going closer to default runtime (knowing the display is different by runtime so we might be "always different") or to go with a custom look.
> 
> I also have to handle some more urgent tasks and discussions but will make sure this gets completed as soon as possible. FYI I'll include this in the 1.4.2 release so the wait period is as small as possible.

No worries, thanks! Appreciate it 😄

src/Platform/Microsoft.Testing.Platform/OutputDevice/Terminal/TerminalTestReporter.cs
```diff
@@ -476,52 +474,91 @@ internal void TestCompleted(
...
+
+        terminal.SetColor(TerminalColor.Red);
+
+        if (exception is null)
+        {
+            AppendIndentedLine(terminal, errorMessage, SingleIndentation);
+        }
+        else if (outcome == TestOutcome.Fail)
+        {
+            // For failed tests, we don't prefix the message with the exception type as it's most likely some assertion failure.
```
nohwnd:
```suggestion
            // For failed tests, we don't prefix the message with the exception type because it is most likely an assertion specific exception like AssertionFailedException, and we prefer to show that without the exception type to avoid additional noise.
```

samples/Playground/Program.cs
```diff
@@ -27,8 +29,13 @@ public static async Task<int> Main(string[] args)
             // To attach to the children
             Microsoft.Testing.TestInfrastructure.DebuggerUtility.AttachCurrentProcessToParentVSProcess();
             ITestApplicationBuilder testApplicationBuilder = await TestApplication.CreateBuilderAsync(args);
+
+            // Test MSTest
             testApplicationBuilder.AddMSTest(() => [Assembly.GetEntryAssembly()!]);
 
+            // Test a custom local test framework
+            // testApplicationBuilder.RegisterTestFramework(_ => new TestFrameworkCapabilities(), (_, _) => new DummyAdapter());
```
nohwnd, line 37: Should this be commented out, or uncommented?

src/Platform/Microsoft.Testing.Platform/OutputDevice/Terminal/TerminalTestReporter.cs
thomhurst: Just my opinion here, but I actually don't like that the exception type is missing, and find it useful to see it.

From my experience, test failures can be due to flakiness (e.g. network, timeouts, etc) and the exception type makes it much quicker and easier to realise what the error is. And it's only ever something like 20 characters long so I wouldn't consider it noise.

samples/Playground/Program.cs
Evangelink: I decided to keep the playground in a state where MSTest is default while keeping the code for a custom adapter so that we don't have to always recreate the various lines. Ideally, I would need to spend a little time to use a variable/const or some pragma to switch between MSTest and custom adapter but that's not urgent as far of this PR is concerned.

src/Platform/Microsoft.Testing.Platform/OutputDevice/Terminal/TerminalTestReporter.cs
Evangelink: If your test adapter reports `TestOutcome.Error` you will go in the else branch where we do display the exception full name as prefix. As explained in the comment, most of the time this is linked to assertion failure where just displaying the message is clearer.

I would prefer to start with that version and collect some evidence of what our users like/don't like rather than making a decision based on a few of us.

thomhurst: I see. So should I be reporting assertion failures with a Failed state property and other exceptions with an Error state property?

PR #3929, opened 2024-10-10
Fix publishing as docker image via /t:PublishContainer

nohwnd:


Evangelink (commented):
Use spaces and not tabs and also please fix the `Condition` indentation

nohwnd:
Yeah sorry, I thought VS Code would fix that automatically when I ran autoformat.

Evangelink (approved):

PR #3932, opened 2024-10-11
Add extra properties for test results in dotnet test

mariam-abdulla:
This pull request includes significant updates to the test result handling and serialization logic in the `Microsoft.Testing.Platform` project. The changes enhance the detail and accuracy of test result messages, adding new fields and improving the serialization process.

### Enhancements to Test Result Handling:

* [`src/Platform/Microsoft.Testing.Extensions.VSTestBridge/ObjectModel/ObjectModelConverters.cs`](diffhunk://#diff-19d2d5674801591c37f6fe2f6788641b35869a9523c2eb07c47ea87f042b256eR170-R176): Added `StandardErrorProperty` and `StandardOutputProperty` to `TestNode` properties for better error and output tracking.

* [`src/Platform/Microsoft.Testing.Platform/ServerMode/DotnetTest/IPC/DotnetTestDataConsumer.cs`](diffhunk://#diff-986f7f7a85121d9d364713ed50cefda98cedb1bb32c37ada27a9e2c35913773fL50-R52): Refactored `GetTestNodeDetails` to return a `TestNodeDetails` record, encapsulating state, duration, reason, error message, stack trace, standard output, and standard error. Updated `ConsumeAsync` method to utilize this new structure. [[1]](diffhunk://#diff-986f7f7a85121d9d364713ed50cefda98cedb1bb32c37ada27a9e2c35913773fL50-R52) [[2]](diffhunk://#diff-986f7f7a85121d9d364713ed50cefda98cedb1bb32c37ada27a9e2c35913773fL76-R80) [[3]](diffhunk://#diff-986f7f7a85121d9d364713ed50cefda98cedb1bb32c37ada27a9e2c35913773fL94-R103) [[4]](diffhunk://#diff-986f7f7a85121d9d364713ed50cefda98cedb1bb32c37ada27a9e2c35913773fL160-R176) [[5]](diffhunk://#diff-986f7f7a85121d9d364713ed50cefda98cedb1bb32c37ada27a9e2c35913773fR186) [[6]](diffhunk://#diff-986f7f7a85121d9d364713ed50cefda98cedb1bb32c37ada27a9e2c35913773fR197-R232)

### Serialization Improvements:

* [`src/Platform/Microsoft.Testing.Platform/IPC/Serializers/BaseSerializer.cs`](diffhunk://#diff-3a5a5cb24c1c1c7e15f669cd55eb60442a821f1647df225c7edc5419236e7707R312-R323): Added a new `WriteField` method to handle `long?` values and updated the `WriteAtPosition` method to support `long` type. [[1]](diffhunk://#diff-3a5a5cb24c1c1c7e15f669cd55eb60442a821f1647df225c7edc5419236e7707R312-R323) [[2]](diffhunk://#diff-3a5a5cb24c1c1c7e15f669cd55eb60442a821f1647df225c7edc5419236e7707R385)

* [`src/Platform/Microsoft.Testing.Platform/ServerMode/DotnetTest/IPC/Models/TestResultMessages.cs`](diffhunk://#diff-2c29717383508e646d7aae91cd65b80176af324e1e3478b514fa193621aa8648L6-R8): Modified `SuccessfulTestResultMessage` and `FailedTestResultMessage` records to include `Duration`, `StandardOutput`, and `ErrorOutput` fields.

* [`src/Platform/Microsoft.Testing.Platform/ServerMode/DotnetTest/IPC/ObjectFieldIds.cs`](diffhunk://#diff-cbb7bc61903bb440f8a4dc3a414c242ed3297e1717aed55827d7f7131ca670baL69-R87): Updated field IDs to accommodate new fields in test result messages.

* [`src/Platform/Microsoft.Testing.Platform/ServerMode/DotnetTest/IPC/Serializers/TestResultMessagesSerializer.cs`](diffhunk://#diff-86fbec91c5c9d9dfea10a1fab74655330d522ac79cc075ca84c980f3bf67b40aL31-R49): Enhanced the serialization format to include new fields (`Duration`, `StandardOutput`, `StandardError`) for both successful and failed test messages. Updated the payload reading logic to handle these new fields.


src/Platform/Microsoft.Testing.Extensions.VSTestBridge/ObjectModel/ObjectModelConverters.cs
```diff
@@ -167,11 +167,13 @@ public static TestNode ToTestNode(this TestResult testResult, bool isTrxEnabled,
             if (testResultMessage.Category == TestResultMessage.StandardErrorCategory)
             {
                 testNode.Properties.Add(new SerializableKeyValuePairStringProperty("vstest.TestCase.StandardError", testResultMessage.Text ?? string.Empty));
+                testNode.Properties.Add(new StandardErrorProperty(testResultMessage.Text ?? string.Empty));
```
nohwnd, line 170: These are potentially much larger than the rest of the message, is it good idea to have 2 copies of the text in the object? Is this the only way to do this? can `"vstest.TestCase.StandardError"` be dropped in favor of the new Platform provided property?

src/Platform/Microsoft.Testing.Platform/ServerMode/DotnetTest/IPC/ObjectFieldIds.cs
```diff
@@ -66,19 +66,25 @@ internal static class SuccessfulTestResultMessageFieldsId
     public const ushort Uid = 1;
     public const ushort DisplayName = 2;
     public const ushort State = 3;
-    public const ushort Reason = 4;
-    public const ushort SessionUid = 5;
+    public const ushort Duration = 4;
```
nohwnd, line 69: isn't this a bad idea for backwards compatibility?Normally you want to add new data fields to the to the end, so you can read what you know, and skip the rest. BUT I don't know how your protocol solves this, so please feel free to ignore me / educate me, if I am wrong :)

src/Platform/Microsoft.Testing.Extensions.VSTestBridge/ObjectModel/ObjectModelConverters.cs
MarcoRossignoli: We cannot we need 2 way because the `SerializableKeyValuePairStringProperty` is an internal prop to accomodate VS. VS TE is working in 2 different mode where the values are taken using different key, value. 
Under VS protocol we declare vstest capablity and that value is the expected one.

Adding `StandardErrorProperty` like xunit did is to have the new model and avoid the need to go with an untyped "key name" that's also used in the protocol atm as vstest capability. Otherwise the core platform should embed code related to the protocol for the vstest capability.

MarcoRossignoli: My idea was to drop the vstest capability, but looks like at time wasn't good for TE. cc: @drognanar

src/Platform/Microsoft.Testing.Platform/ServerMode/DotnetTest/IPC/ObjectFieldIds.cs
mariam-abdulla: We still haven't shipped so this won't affect anything for now.

src/Platform/Microsoft.Testing.Extensions.VSTestBridge/ObjectModel/ObjectModelConverters.cs
nohwnd: So the solution here is fix the VS client, and then remove the vstest.* property, yeah?

src/Platform/Microsoft.Testing.Platform/ServerMode/DotnetTest/IPC/ObjectFieldIds.cs
nohwnd: I am correct here, but the api is not shipped so we can play around with the fields as we like.

mariam-abdulla: Yeah, exactly.

src/Platform/Microsoft.Testing.Extensions.VSTestBridge/ObjectModel/ObjectModelConverters.cs
nohwnd: Do you know if we end up serializing both when calling TE or some other client? If this is not the case then the overhead of having two "copies" of the same string on the node is minimal, but I doubt that is the case, and we actually send it twice. 

If you don't know off hand just say it , I will make an issue for investigating it.

nohwnd: created: #3933

nohwnd (approved):

MarcoRossignoli (approved):

PR #3972, opened 2024-10-26
Fix infinite loop in exception formatting when exceptions have cyclic references

### LINKED ISSUE #3971: Throwing results in test never completing
## Describe the bug
When I throw a cyclical exception from a test (like is thrown when a resilience pipeline fails) the test from which it is thrown never exits.

## Steps To Reproduce

1. Run the test `TestMethod1` in the attached project.
2. See that the test runs forever (well I left it 5 mins)

## Expected behavior
The test should finish as a failure.

## Actual behavior
The test never exits.

## Additional context
**MSTest.XYZ Version**: 3.6.1
**Microsoft.NET.Test.Sdk Version**: 17.11.0
**Operating System**: Windows 10 Pro
**Visual Studio Version**: 2022 (64-bit) - Preview (17.12.0 Preview 4.0)
**.NET Version**: net8.0

Toy test code for a repro if you don't want to download the zip:
```
    [TestMethod]
    public void TestMethod1()
    {
        // Arrange
        var exceptionA = new Exception("Exception A");
        var exceptionB = new Exception("Exception B", exceptionA);

        // Use reflection to set exceptionA's InnerException to exceptionB, creating a cycle
        FieldInfo innerExceptionField = typeof(Exception).GetField("_innerException", BindingFlags.Instance | BindingFlags.NonPublic);
        innerExceptionField.SetValue(exceptionA, exceptionB);

        // Act: Throw exceptionA, which now has a cyclic InnerException reference
        throw exceptionA;
    }
```

[MSTest.Bug.Repro.zip](https://github.com/user-attachments/files/17532204/MSTest.Bug.Repro.zip)

Vijay-Braidable:
The GetFormattedExceptionMessage method in ExceptionHelper.cs recursively traverses the inner exceptions of an exception to build a formatted error message string. However, it didn't handle the case where the inner exceptions have a cyclic reference, causing an infinite loop.

This change fixes that issue by keeping track of the exceptions seen during traversal in a HashSet. If an exception is encountered again, it breaks out of the loop and appends a "[Cyclic Exception Reference]" message to the output, avoiding the infinite recursion.

**The key changes are:**
- Introduce a HashSet seenExceptions to keep track of exceptions encountered so far.
- Check if the current exception has already been seen. If so, append a "[Cyclic Exception Reference]" message and break out of the loop.
- If the current exception hasn't been seen, add it to the seenExceptions set and continue processing as before.
- A new test class ExceptionHelperTests has also been added with two test methods:
- GetFormattedExceptionMessage_WithCyclicExceptions_ShouldNotLoopInfinitely - Verifies that cyclic exceptions are handled correctly and don't cause infinite loops.
- GetFormattedExceptionMessage_WithNoCyclicExceptions_ShouldFormatNormally - Ensures that the normal formatting still works as expected for non-cyclic exceptions.

**This fixes #3971**

### DIFF
--- a/src/Adapter/MSTest.TestAdapter/Execution/ExceptionHelper.cs
+++ b/src/Adapter/MSTest.TestAdapter/Execution/ExceptionHelper.cs
@@ -145,10 +145,19 @@ internal static class ExceptionHelper
 
         StringBuilder result = new();
         bool first = true;
+        HashSet<Exception> seenExceptions = new();
+
         for (Exception? curException = ex;
              curException != null;
              curException = curException.InnerException)
         {
+            if (seenExceptions.Contains(curException))
+            {
+                result.Append(" ---> [Cyclic Exception Reference]");
+                break;
+            }
+            seenExceptions.Add(curException);
+
             // Get the exception message. Need to check for errors because the Message property
             // may have been overridden by the exception type in user code.
             string msg;
--- a/test/UnitTests/MSTestAdapter.UnitTests/Execution/ExceptionHelperTests.cs
+++ b/test/UnitTests/MSTestAdapter.UnitTests/Execution/ExceptionHelperTests.cs
@@ -1,0 +1,46 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System;
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+
+namespace Microsoft.VisualStudio.TestPlatform.MSTestAdapter.UnitTests.Execution
+{
+    [TestClass]
+    public class ExceptionHelperTests
+    {
+        [TestMethod]
+        public void GetFormattedExceptionMessage_WithCyclicExceptions_ShouldNotLoopInfinitely()
+        {
+            // Arrange
+            var exceptionA = new Exception("Exception A");
+            var exceptionB = new Exception("Exception B", exceptionA);
+            
+            // Use reflection to set exceptionA's InnerException to exceptionB, creating a cycle
+            var innerExceptionField = typeof(Exception).GetField("_innerException", BindingFlags.Instance | BindingFlags.NonPublic);
+            innerExceptionField.SetValue(exceptionA, exceptionB);
+
+            // Act
+            string formattedMessage = ExceptionHelper.GetFormattedExceptionMessage(exceptionA);
+
+            // Assert
+            StringAssert.Contains(formattedMessage, "[Cyclic Exception Reference]");
+        }
+
+        [TestMethod]
+        public void GetFormattedExceptionMessage_WithNoCyclicExceptions_ShouldFormatNormally()
+        {
+            // Arrange
+            var exceptionA = new Exception("Exception A");
+            var exceptionB = new Exception("Exception B", exceptionA);
+
+            // Act  
+            string formattedMessage = ExceptionHelper.GetFormattedExceptionMessage(exceptionB);
+
+            // Assert
+            StringAssert.Contains(formattedMessage, "Exception A");
+            StringAssert.Contains(formattedMessage, "Exception B");
+            StringAssert.DoesNotMatch(formattedMessage, new Regex(@"\[Cyclic Exception Reference\]"));
+        }
+    }
+}


Vijay-Braidable:

src/Adapter/MSTest.TestAdapter/Execution/ExceptionHelper.cs
```diff
@@ -145,10 +145,20 @@ internal static string GetFormattedExceptionMessage(this Exception ex)
...
              curException != null;
              curException = curException.InnerException)
         {
+            if (seenExceptions.Contains(curException))
+            {
+                result.Append(" ---> [Cyclic Exception Reference]");
+                break;
+            }
+            
+            seenExceptions.Add(curException);
```
campersau: Could be reduced to:

```suggestion
            if (!seenExceptions.Add(curException))
            {
                result.Append(" ---> [Cyclic Exception Reference]");
                break;
            }
```

test/UnitTests/MSTestAdapter.UnitTests/Execution/ExceptionHelperTests.cs
```diff
@@ -0,0 +1,46 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System;
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+
+namespace Microsoft.VisualStudio.TestPlatform.MSTestAdapter.UnitTests.Execution
+{
+    [TestClass]
+    public class ExceptionHelperTests
```
Evangelink, line 10: We are not using MSTest to test MSTest for some historical reason.

```suggestion
    public sealed class ExceptionHelperTests : TestContainer
```

test/UnitTests/MSTestAdapter.UnitTests/Execution/ExceptionHelperTests.cs
```diff
@@ -0,0 +1,46 @@
...
+    {
+        [TestMethod]
+        public void GetFormattedExceptionMessage_WithCyclicExceptions_ShouldNotLoopInfinitely()
```
Evangelink, line 13:
```suggestion
        public void GetFormattedExceptionMessage_WithCyclicExceptions_ShouldNotLoopInfinitely()
```

test/UnitTests/MSTestAdapter.UnitTests/Execution/ExceptionHelperTests.cs
```diff
@@ -0,0 +1,46 @@
...
+
+            // Act
+            string formattedMessage = ExceptionHelper.GetFormattedExceptionMessage(exceptionA);
+
+            // Assert
+            StringAssert.Contains(formattedMessage, "[Cyclic Exception Reference]");
+        }
+
+        [TestMethod]
+        public void GetFormattedExceptionMessage_WithNoCyclicExceptions_ShouldFormatNormally()
```
Evangelink, line 31:
```suggestion
        public void GetFormattedExceptionMessage_WithNoCyclicExceptions_ShouldFormatNormally()
```

src/Adapter/MSTest.TestAdapter/Execution/ExceptionHelper.cs
```diff
@@ -145,10 +145,18 @@ internal static string GetFormattedExceptionMessage(this Exception ex)
...
         bool first = true;
+        HashSet<Exception> seenExceptions = new();
+
         for (Exception? curException = ex;
              curException != null;
              curException = curException.InnerException)
         {
+            if (!seenExceptions.Add(curException))
+            {
+                result.Append(" ---> [Cyclic Exception Reference]");
```
Evangelink, line 156: Please use localized message.

Evangelink (commented):
You will also need to handle the StackTrace method helper.

Evangelink (commented):
You will also need to handle the StackTrace method helper.

thomhurst:
I still don't understand this. An exception can't have a cyclical inner exception reference without using reflection to hack it (unless someone can give me source code to prove me wrong?) And that would hang on the throw anyway so this wouldn't solve that?

Evangelink:
Good point! I assumed that the throw was good and it was hanging inside MSTest but that's the throw that's causing exception so nothing to do on MSTest side.

Thanks @thomhurst

Issue #3831, opened 2024-09-13
Under hot reload the UX is confusing


MarcoRossignoli:
Under the extension https://learn.microsoft.com/en-us/dotnet/core/testing/unit-testing-platform-extensions-hosting the UX  is not behaving correctly, it's printing always the timer.

Evangelink:
Fixed by #3983

PR #3983, opened 2024-10-31
Fix live output with HotReload

nohwnd:
Fix https://github.com/microsoft/testfx/issues/3831

### DIFF
--- a/src/Platform/Microsoft.Testing.Platform/OutputDevice/IHotReloadPlatformOutputDevice.cs
+++ b/src/Platform/Microsoft.Testing.Platform/OutputDevice/IHotReloadPlatformOutputDevice.cs
@@ -1,0 +1,11 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+namespace Microsoft.Testing.Platform.OutputDevice;
+
+internal interface IHotReloadPlatformOutputDevice : IPlatformOutputDevice
+{
+    Task DisplayBeforeHotReloadSessionStartAsync();
+
+    Task DisplayAfterHotReloadSessionEndAsync();
+}
--- a/src/Platform/Microsoft.Testing.Platform/OutputDevice/Terminal/AnsiTerminalTestProgressFrame.cs
+++ b/src/Platform/Microsoft.Testing.Platform/OutputDevice/Terminal/AnsiTerminalTestProgressFrame.cs
@@ -177,7 +177,12 @@ internal sealed class AnsiTerminalTestProgressFrame
             terminal.MoveCursorUp(previousFrame.ProgressCount + 2);
         }
 
-        terminal.AppendLine();
+        // When there is nothing to render, don't write empty lines, e.g. when we start the test run, and then we kick off build
+        // in dotnet test, there is a long pause where we have no assemblies and no test results (yet).
+        if (ProgressCount > 0)
+        {
+            terminal.AppendLine();
+        }
 
         int i = 0;
         for (; i < ProgressCount; i++)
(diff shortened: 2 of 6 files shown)


Evangelink:
I would have used the same display as for the regular live output with support for the same options but at least it's fixing the hot reload experience so that's good step forward.

nohwnd:
This is using the same display as the normal logger with all the options it has. User can suppress the progress for example by calling with --no-progress. The special methods for hot reload call to the same api as normal run. We just need to NOT call the final completion because user would see Aborted message on ctrl_C and that looked ugly and unexpected.

Evangelink:
@nohwnd do we see the passed tests by default or is it just you have enabled the detailed output for the gif? I'd really want for the various modes to have the same view by default to provide consistent experience.

src/Platform/Microsoft.Testing.Platform/OutputDevice/IHotReloadPlatformOutputDevice.cs
```diff
@@ -0,0 +1,11 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+namespace Microsoft.Testing.Platform.OutputDevice;
+
+internal interface IHotReloadPlatformOutputDevice : IPlatformOutputDevice
```
Evangelink, line 6: Let's go as-is for now but I really feel like we should move forward with the concept of capability and check if the registered platform output device is "hot reload aware".

Actually looking at the hot reload usage of this interface, I am wondering if it's at all needed? Why don't we simply always call the start/end methods of the `IPlatformOutputDevice` when the hot reload mode is enabled?

Evangelink (changes requested):
Branch is not ready for the next iteration, let's not merge now

Evangelink:
I think this is also fixing  #3831

src/Platform/Microsoft.Testing.Platform/OutputDevice/IHotReloadPlatformOutputDevice.cs
nohwnd: This is because if you do that you will get Test run aborted message when you press CTRL+C, due to the nature in which hot reload works. This message is confusing because most often no tests will be running at that time. 

These two additional methods are a way to explicitly indicate what is happening, because otherwise we don't know which  ~abort~ assembly completed event is the last one which should be ignored. 

Normal run is like this: 

session start
assembly start -> tests start
tests finished -> assembly end OR  ctrl+C -> assembly end
session end

But hot reload is like this:

session start
assembly start -> tests start
tests finished -> assembly end OR ctrl+C -> assembly end

... session still continues and runs can repeat

ctrl+C -> assembly end (and this prints the extra abort we don't want to see)
session end

nohwnd:
Yes the default output is to show passed tests in hotReload mode.

nohwnd:
> I think this is also fixing #3831

Yes, true. Adding to auto-close.

src/Platform/Microsoft.Testing.Platform/OutputDevice/IHotReloadPlatformOutputDevice.cs
Evangelink: Yes exactly why I think having a capability you could check when in the init/end to display or not the message would be better. As said, this is more of a where we should go than what this PR needs to do.

Evangelink:
Can we make it to have consistent output in the various mode and don't show passed tests by default (like for a normal run)?

Evangelink (approved):
Approving but would love to have the consistent experience.

Evangelink (approved):
Approving but would love to have the consistent experience.

PR #3992, opened 2024-11-04
Fix hangdump not showing tests in progress

nohwnd:
Fixes cases where method names are same in multiple classes (e.g. TestMethod1) and are not visible among the hanging tests.

The bug happens when Class1.TestMethod1 and Class2.TestMethod1 run in parallel, and one of them is not hanging. The first one to complete will remove the entry from tests in progress. 

Using UID reduces the chance of this happening significantly, and will also correctly show  that two tests with name TestMethod1 were running.

### DIFF
--- a/src/Platform/Microsoft.Testing.Extensions.HangDump/HangDumpActivityIndicator.cs
+++ b/src/Platform/Microsoft.Testing.Extensions.HangDump/HangDumpActivityIndicator.cs
@@ -35,7 +35,7 @@ internal sealed class HangDumpActivityIndicator : IDataConsumer, ITestSessionLif
     private readonly ManualResetEventSlim _signalActivity = new(false);
     private readonly ManualResetEventSlim _mutexCreated = new(false);
     private readonly bool _traceLevelEnabled;
-    private readonly ConcurrentDictionary<string, (Type, DateTimeOffset)> _testsCurrentExecutionState = new();
+    private readonly ConcurrentDictionary<string, (string Name, Type Type, DateTimeOffset StartTime)> _testsCurrentExecutionState = new();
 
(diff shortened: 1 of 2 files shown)


src/Platform/Microsoft.Testing.Extensions.HangDump/HangDumpActivityIndicator.cs
```diff
@@ -35,7 +35,7 @@ internal sealed class HangDumpActivityIndicator : IDataConsumer, ITestSessionLif
     private readonly ManualResetEventSlim _signalActivity = new(false);
     private readonly ManualResetEventSlim _mutexCreated = new(false);
     private readonly bool _traceLevelEnabled;
-    private readonly ConcurrentDictionary<string, (Type, DateTimeOffset)> _testsCurrentExecutionState = new();
+    private readonly ConcurrentDictionary<string, (string Name, Type Type, DateTimeOffset StartTime)> _testsCurrentExecutionState = new();
```
Evangelink: Can we change the key type from string to UID for safety?

Evangelink (commented):
Could you please add some test? Acceptance is probably the easiest.

src/Platform/Microsoft.Testing.Extensions.HangDump/HangDumpActivityIndicator.cs
nohwnd: yes.

Evangelink (approved):

Evangelink:
Please forward port this.

nohwnd:
https://github.com/microsoft/testfx/pull/3999

PR #4014, opened 2024-11-06
Apply IDE0032 use auto-properties

Evangelink:


Evangelink:
Looks like the compiler version is differnet on CI so closing the PR.

PR #4015, opened 2024-11-07
Add TestData property to the TestContext

### LINKED ISSUE #385: Access to Test Method parameters from TestCleanup.
## Description
I have test method hooked up to a data source attribute that is passing to a parameter an object that needs to be disposed at the end of the test.  Consequently, I would like to have another function in the ITestDataSource interface that gives me back the object array that was just passed to the test function so that I can handle any objects in it that need to be disposed of.

## Steps to reproduce
Here is a rough example. 
This is an overridden data source attribute creating the object for the test, and using the object type in its display name:
```
public class ComplexTestAttribute : Attribute, ITestDataSource
{
    {
        for(int i = 0; i < 3; i++)
            yield return new object[1] { new ComplexObject(i) };
        }
    }

    {
        if (data != null)
        {
            return string.Format(CultureInfo.CurrentCulture, "{0} ({1})", methodInfo.Name, data[0].GetType().Name);
        }
        return string.Format(CultureInfo.CurrentCulture, "{0} ({1})", methodInfo.Name, "Unknown driver");
    }

    //Theoretical cleanup function.
    public void TestCleanup(object[] data, TestContext testContext)
    {
        ComplexObject obj =  (ComplexObject)data[0];
        if(testContext.CurrentTestOutcome == UnitTestOutcome.Failed)
        {
            obj.HandleError();
        }
        obj.Dispose();
    }
}
```
And here is the test class that uses it:
```
[TestClass]
class myClass 
{
    [ComplexTest]
    [TestMethod]
    void myTest(ComplexObject obj) 
    {
        obj.RunTest();
    }
}
```

## Expected behavior
It would be good to have a TestCleanup function in the ITestDataSource interface that takes the latest test record data AND the test context of the test.  This way any cleanup or other follow up logic could happen within the test thread.

## Actual behavior
Right now, I have the ITestDataSource class as a subclass of my test class, so that it can call a function with access to the TestContext object.  However, since I have no access to the actual TestContext, or even to the test class instance that ran the object, I have to make my underlying TestContext object static, which is clearly not thread safe, and thus not conducive to parallelization.

## Environment
Using Windows 10,
MSTest.TestFramework:  version 1.2.0

Evangelink:
Makes TestContext aware of the parameters that will be passed to the test method so user could have different behavior in the test init or cleanup,

Fixes #385

### DIFF
--- a/src/TestFramework/TestFramework.Extensions/TestContext.cs
+++ b/src/TestFramework/TestFramework.Extensions/TestContext.cs
@@ -53,6 +53,8 @@ public abstract class TestContext
     /// </summary>
     public virtual CancellationTokenSource CancellationTokenSource { get; protected set; } = new();
 
+    public object?[]? TestData { get; protected set; }
+
 #if NETFRAMEWORK
     /// <summary>
     /// Gets the current data row when test is used for data driven testing.
--- a/src/Adapter/MSTest.TestAdapter/Execution/TestMethodRunner.cs
+++ b/src/Adapter/MSTest.TestAdapter/Execution/TestMethodRunner.cs
@@ -354,8 +354,8 @@ internal class TestMethodRunner
     private TestResult[] ExecuteTestWithDataSource(UTF.ITestDataSource? testDataSource, object?[]? data)
     {
         var stopwatch = Stopwatch.StartNew();
-
         _testMethodInfo.SetArguments(data);
+        _testContext.SetTestData(data);
         TestResult[] testResults = ExecuteTest(_testMethodInfo);
         stopwatch.Stop();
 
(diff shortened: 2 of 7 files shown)


src/TestFramework/TestFramework.Extensions/TestContext.cs
```diff
@@ -53,6 +53,8 @@ public abstract class TestContext
     /// </summary>
     public virtual CancellationTokenSource CancellationTokenSource { get; protected set; } = new();
 
+    public object?[]? TestData { get; protected set; }
```
Evangelink, line 56: The pattern for the other properties is to have the property virtual to have the setter done in the implementation class but I don't see much value so I am using a protected setter instead.

fhnaseer (approved):

PR #4024, opened 2024-11-08
Fix IsNonDerived implementation for attribute comparison

Youssef1313:

### DIFF
--- a/src/Adapter/MSTest.TestAdapter/Helpers/AttributeComparer.cs
+++ b/src/Adapter/MSTest.TestAdapter/Helpers/AttributeComparer.cs
@@ -6,6 +6,8 @@ namespace Microsoft.VisualStudio.TestPlatform.MSTest.TestAdapter.Helpers;
 internal static class AttributeComparer
 {
     public static bool IsNonDerived<TAttribute>(Attribute attribute) =>
+        // We are explicitly checking the given type *exactly*. We don't want to consider inheritance.
+        // So, this should NOT be refactored to 'attribute is TAttribute'.
         attribute.GetType() == typeof(TAttribute);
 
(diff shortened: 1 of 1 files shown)


src/Adapter/MSTest.TestAdapter/Helpers/AttributeComparer.cs
```diff
@@ -6,7 +6,7 @@ namespace Microsoft.VisualStudio.TestPlatform.MSTest.TestAdapter.Helpers;
 internal static class AttributeComparer
 {
     public static bool IsNonDerived<TAttribute>(Attribute attribute) =>
-        attribute is TAttribute;
+        attribute.GetType() == typeof(TAttribute);
```
nohwnd, line 11: Add a comment please that we don't want to do `is` check here and why.

nohwnd: do you know if compiler will optimize this call to be about the same or better perf? I don't think it will create the whole type object and compare it. But I am not 100% sure :)

Youssef1313: `GetType() == typeof(...)` is generally faster than `is` type checking as far as I know. The compiler doesn't need to optimize actually

Youssef1313: I have learnt that earlier from https://github.com/dotnet/runtime/blob/1d86001220b04cd9d074c28ab62610c6acb7698e/src/libraries/System.Linq/src/System/Linq/Enumerable.cs#L42-L47 :)

Youssef1313:
No, it's not linked to it.

Evangelink (approved):

Issue #334, opened 2017-12-17
Provide additional information for assert failures


codito:
## Description
Default `Assert` methods should provide information in case of failures. They should at least answer the next question for a developer once a failure occurs. I've seen this scenarios with `CollectionAsserts`, request to review other APIs as well.

## Steps to reproduce
1. Create a hello world test project
2. Add two `List<int>` variables. `var actual = { 1, 2, 3 }` and `expected = { 1, 3, 2 }`
3. Add an assert. `CollectionAssert.AreEqual(expected, actual)`
4. Run tests

## Expected behavior
A developer would immediately ask what were the elements if the collection doesn't match. Note the `**additional**` information below.

```
Failed   C.Tests.N.D
Error Message:
 CollectionAssert.AreEqual failed. (Element at index 1 do not match.)
 **Expected = { 2, 1, 3 }, Actual = { 1, 2, 3 }**
Stack Trace:
   at C.Tests.N.D()
```

## Actual behavior
```
Failed   C.Tests.N.D
Error Message:
 CollectionAssert.AreEqual failed. (Element at index 1 do not match.)
Stack Trace:
   at C.Tests.N.D()
```

## Environment
MSTest 1.2.0, VSTest 15.3.0, dotnet-sdk 2.0 (linux)

GoFightNguyen:
This is a great issue. If this functionality gets added, then great. But I think we should also consider using [Fluent Assertions](https://github.com/fluentassertions/fluentassertions).

I used your example and Fluent Assertions for the asserts.

`actual.Should().BeEquivalentTo(expected);`
This passes, because it does not care about ordering.

If I do `actual.Should().BeEquivalentTo(expected, o => o.WithStrictOrdering());`, then the test output is
> Message: Expected item[1] to be a collection with 3 item(s)Expected item[1] to be 3, but found 2.
Expected item[2] to be a collection with 3 item(s)Expected item[2] to be 2, but found 3.

Admittedly, that message has unnecessary and redundant information
>Expected item[1] to be a collection with 3 item(s)
>Expected item[2] to be a collection with 3 item(s)

I'll have to open an issue, because the previous version of Fluent Assertions did not include that cruft.

ckamunya11:
cruft

Evangelink:
[...]

PR #4027, opened 2024-11-08
Partially improve CollectionAssert message

Youssef1313:
It still doesn't give the complete info as requested in #334


Evangelink (commented):
I would prefer to have actual and expected on new lines and not surrounded by quotes because for most complex data it could render weirdly.

@nohwnd I don't recall if your idea of expected/actual as exception data is already supported in TE or VSTest. If that's the case, it could be better than doing it explicitly (as done here).

@Youssef1313 Could we add a few tests?

Evangelink (commented):
I would prefer to have actual and expected on new lines and not surrounded by quotes because for most complex data it could render weirdly.

@nohwnd I don't recall if your idea of expected/actual as exception data is already supported in TE or VSTest. If that's the case, it could be better than doing it explicitly (as done here).

@Youssef1313 Could we add a few tests?

test/UnitTests/TestFramework.UnitTests/Assertions/CollectionAssertTests.cs
```diff
@@ -465,6 +465,17 @@ public void CollectionAssertAreNotEquivalent_FailsWithTwoNullsAndComparer_WithMe
...
 
+    public void CollectionAssertAreEqual_FailsWithGoodMessage()
+    {
+        Exception ex = VerifyThrows(() => CollectionAssert.AreEqual(new[] { 1, 2, 3 }, new[] { 1, 5, 3 }));
+        Assert.AreEqual(
+            """
+            CollectionAssert.AreEqual failed. (Element at index 1 do not match.
+            Expected: 2
+            Actual: 5)
+            """, ex.Message);
```
Evangelink: Could we get rid of the parenthesis? I think it's only reducing readability.

Evangelink (dismissed):
One last readability nit and LGTM, thanks @Youssef1313

Evangelink (approved):
Awesome :shipit:

PR #4033, opened 2024-11-11
remove redundant reflection in CreateClassInfo

SimonCropp:


Evangelink (commented):
I hope this was not a side-effect call that was used to populate ctors.

@nohwnd is this used for the source gen part?

nohwnd:
This looks like code that was added for source gen so we get the constructors from the metadata, it should not be just side-effect code. If it is not connected to the results of the code we call now, then we should connect it.

nohwnd:
I can have a more detailed look on this, but unfortunately not now.

nohwnd:
This was most likely an oversight from my side. The current code would still work even in native aot, but eventually we should be able to query the "metadata provider" for only the data that are relevant to tests. Replacing with the linked PR for remedy this.

Thanks for pointing this out Simon. 🙂

PR #4037, opened 2024-11-11
Add TestDisplayName property to the TestContext

### LINKED ISSUE #459: TestContext should provide that is a TestMethod or a DataTestMethod
I have some unit tests where I use the TestContext.TestName.
However if I use it in a DataTestMethod my tests are failing because there is a collision between the tests because of the TestName. 

Is there a way to determine if the test is a DataTestMethod? Or even better to retrieve the DisplayName from the TestContext.

Evangelink:
Fixes #459

### DIFF
--- a/src/Adapter/MSTest.TestAdapter/Execution/TestMethodRunner.cs
+++ b/src/Adapter/MSTest.TestAdapter/Execution/TestMethodRunner.cs
@@ -182,6 +182,7 @@ internal class TestMethodRunner
             }
             else
             {
+                _testContext.SetDisplayName(_test.DisplayName);
                 TestResult[] testResults = ExecuteTest(_testMethodInfo);
 
                 foreach (TestResult testResult in testResults)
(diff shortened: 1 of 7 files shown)


src/Adapter/MSTest.TestAdapter/Execution/TestMethodRunner.cs
```diff
@@ -182,6 +182,7 @@ internal UnitTestResult[] RunTestMethod()
             }
             else
             {
+                _testContext.SetDisplayName(_test.DisplayName);
```
Youssef1313, line 185: Not sure why it needs to be conditional. Any harm doing it unconditionally at the beginning of `RunTestMethod` or even in the constructor [here](https://github.com/microsoft/testfx/blob/15f2bbd7abf8ebae96070006eff2c3bf921b62b8/src/Adapter/MSTest.TestAdapter/Execution/TestMethodRunner.cs#L61-L62)?

Youssef1313 (approved):

src/Adapter/MSTest.TestAdapter/Execution/TestMethodRunner.cs
Evangelink: It's using a different logic for the parameterized tests. I was thinking about setting it always first and then resetting it based on parameterized tests submethods but thought it would be less clear.

Issue #4034, opened 2024-11-11
`AvoidExpectedExceptionAttributeAnalyzer` doesn't handle attributes that derive from `ExpectedExceptionBaseAttribute`


Youssef1313:
## Describe the bug

The analyzer specifically looks for ExpectedExceptionAttribute, while users could have their own custom `ExpectedExceptionBaseAttribute` derived class

PR #4038, opened 2024-11-11
Implement codefix for MSTEST0006 (AvoidExpectedException)

### LINKED ISSUE #3399: MSTEST0006: Add code fix
AB#2200928

Youssef1313:
Fixes #3399

### DIFF
--- a/src/Analyzers/MSTest.Analyzers.CodeFixes/AvoidExpectedExceptionAttributeFixer.cs
+++ b/src/Analyzers/MSTest.Analyzers.CodeFixes/AvoidExpectedExceptionAttributeFixer.cs
@@ -49,9 +49,7 @@ public sealed class AvoidExpectedExceptionAttributeFixer : CodeFixProvider
 
         // Find the method declaration identified by the diagnostic.
         MethodDeclarationSyntax methodDeclaration = syntaxToken.Parent.AncestorsAndSelf().OfType<MethodDeclarationSyntax>().First();
-        SemanticModel semanticModel = await context.Document.GetSemanticModelAsync(context.CancellationToken).ConfigureAwait(false)
-            ?? throw new InvalidOperationException("SemanticModel cannot be null.");
-
+        SemanticModel semanticModel = await context.Document.GetRequiredSemanticModelAsync(context.CancellationToken).ConfigureAwait(false);
         if (!semanticModel.Compilation.TryGetOrCreateTypeByMetadataName(WellKnownTypeNames.MicrosoftVisualStudioTestToolsUnitTestingExpectedExceptionAttribute, out INamedTypeSymbol? expectedExceptionAttributeSymbol))
         {
             return;
--- a/src/Analyzers/MSTest.Analyzers.CodeFixes/AvoidExpectedExceptionAttributeFixer.cs
+++ b/src/Analyzers/MSTest.Analyzers.CodeFixes/AvoidExpectedExceptionAttributeFixer.cs
@@ -1,0 +1,145 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System.Collections.Immutable;
+using System.Composition;
+
+using Analyzer.Utilities;
+using Analyzer.Utilities.Extensions;
+
+using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.CodeActions;
+using Microsoft.CodeAnalysis.CodeFixes;
+using Microsoft.CodeAnalysis.CSharp;
+using Microsoft.CodeAnalysis.CSharp.Syntax;
+using Microsoft.CodeAnalysis.Editing;
+using Microsoft.CodeAnalysis.Text;
+
+using MSTest.Analyzers.Helpers;
+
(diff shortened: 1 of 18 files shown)


Youssef1313:
@Evangelink Current implementation only wraps last statement. Let me know if we want to implement both fixes or if we want to go with this one first until we hear user feedback

Evangelink:
I think that will be the most used code fix so let's ship this one only for now and wait for feedback. Please once this is merged, update the documentation https://learn.microsoft.com/dotnet/core/testing/mstest-analyzers/mstest0006 to mention we have code fix and the version it's introduced into

src/Analyzers/MSTest.Analyzers.CodeFixes/AvoidExpectedExceptionAttributeFixer.cs
```diff
@@ -0,0 +1,145 @@
...
+        SyntaxNode newLambdaExpression = oldStatement switch
+        {
+            ExpressionStatementSyntax oldLambdaExpression => oldLambdaExpression.Expression,
+            _ => oldStatement,
+        };
+
+        SyntaxGenerator generator = editor.Generator;
+        newLambdaExpression = generator.VoidReturningLambdaExpression(newLambdaExpression);
+
+        bool containsAsyncCode = newLambdaExpression.DescendantNodesAndSelf().Any(n => n is AwaitExpressionSyntax);
```
Evangelink, line 117: Isn't it easier/faster to just look if method declaration as async?

test/UnitTests/MSTest.Analyzers.UnitTests/AvoidExpectedExceptionAttributeAnalyzerTests.cs
```diff
@@ -44,6 +44,276 @@ public class TestClass
...
+                    code,
+                },
+            },
+            FixedState =
+            {
+                Sources =
+                {
+                    fixedCode,
+                },
+                ExpectedDiagnostics =
```
Evangelink: Let's link this issue https://github.com/microsoft/testfx/issues/4034 and say the expected diagnostics should be gone when it's fixed.

Evangelink (commented):
Could you please add a test, sync and async (it's fine to do both in the same test) with multiple statements to clearly document we only wrap last statement?

src/Analyzers/MSTest.Analyzers.CodeFixes/AvoidExpectedExceptionAttributeFixer.cs
```diff
@@ -0,0 +1,145 @@
...
+
+        if (diagnostic.Properties.ContainsKey(DiagnosticDescriptorHelper.CannotFixPropertyKey))
+        {
+            return;
+        }
+
+        // Find the method declaration identified by the diagnostic.
+        MethodDeclarationSyntax methodDeclaration = syntaxToken.Parent.AncestorsAndSelf().OfType<MethodDeclarationSyntax>().First();
+        SemanticModel semanticModel = await context.Document.GetSemanticModelAsync(context.CancellationToken).ConfigureAwait(false)
+            ?? throw new InvalidOperationException("SemanticModel cannot be null.");
```
Evangelink: It should be updated to be based on #4039

src/Analyzers/MSTest.Analyzers.CodeFixes/AvoidExpectedExceptionAttributeFixer.cs
Youssef1313: @Evangelink The test method could be async, but the specific line being wrapped in ThrowsException may be synchronous. I'll add a test to cover this case and make sure it doesn't regress.

Something along the lines of

```csharp
[TestMethod]
[ExpectedException(typeof(Exception))]
public async Task MyTestMethod()
{
    await Task.Delay(100);
    MyMethod();
}

private static void MyMethod() => throw new Exception();
```

It should just be `Assert.ThrowsException<Exception>(() => MyMethod);` and not ThrowsExceptionAsync.

It goes the other way around where we need to use Assert.ThrowsExceptionAsync if the last statement in the test had the `await` deeper, e.g, `MyMethod(await MyOtherMethod())`

test/UnitTests/MSTest.Analyzers.UnitTests/AvoidExpectedExceptionAttributeAnalyzerTests.cs
Youssef1313: It's not related. #4034 is about work to be done on the analyzer. The test here uses ExpectedExceptionAttribute which is supported. However, what is not supported (intentionally) by the codefix here is when the attribute has `AllowDerivedTypes = true` because using `Assert.ThrowsException<ExceptionType>(...)` isn't going to preserve the semantics. The user needs to specify the exact exception type, and not a base one.

src/Analyzers/MSTest.Analyzers.CodeFixes/AvoidExpectedExceptionAttributeFixer.cs
Evangelink: Definitely! Sorry I overlooked that case.

Evangelink (approved):

PR #2839, opened 2024-05-10
Lower reflection overhead

nohwnd:
This PR simplifies the way we access reflection in discovery and run. By improving the attribute cache and using it whenever possible. 

It also simplifies ReflectHelper code, by removing dependencies between methods, and giving them more descriptive names. To make it much easier to reason about the current and desired behavior.

I am keeping the existing *non-consistent* behavior, to avoid breaking changes.  

- IsAttributeDefined => IsNonDerivedAttributeDefined

- GetCustomAttribute => GetFirstDerivedAttributeOrDefault + inherit: true - when used with attributes that allow only one instance

- GetCustomAttributes -> GetDerivedAttributes + inherit: true - for attributes that allow multiple. 

# Before

Initial measurements, on SDK 3.3.1.

Discovery of 10k empty tests: 
discovered: 10000 tests in 1992,1448 ms, sent them in 289,5684 ms, total: 2281,7132
discovered: 10000 tests in 899,1829 ms, sent them in 212,4562 ms, total: 1111,6391
discovered: 10000 tests in 552,0988 ms, sent them in 321,4334 ms, total: 873,5322

Run 10k empty tests:

image

discovery: 1327ms
Microsoft(R) Testing Platform Execution Command Line Tool
Version: 1.1.0+8c0a8fd8e (UTC 2024/04/03)
RuntimeInformation: win-x64 - .NET 8.0.5
Copyright(c) Microsoft Corporation.  All rights reserved.
Passed! - Failed: 0, Passed: 10000, Skipped: 0, Total: 10000, Duration: 2s 272ms - mstest133.dll (win-x64 - .NET 8.0.5)
run: 2616ms

After: 

## After
Discover 10 empty tests: 
discovered: 10000 tests in 1219,4222 ms, sent them in 227,5277 ms, total: 1446,9499 <<<
2088
discovered: 10000 tests in 324,3993 ms, sent them in 193,5436 ms, total: 517,9429 <<<
1024
discovered: 10000 tests in 259,1427 ms, sent them in 291,6613 ms, total: 550,804 <<<
discovered: discovery alloc: 50 MB send alloc: 72 MB
971

Run 10k empty tests:

image

discovery: 999ms
discovered: 10000 tests in 279,4147 ms, sent them in 333,1555 ms, total: 612,5702 <<<
discovered: discovery alloc: 49 MB send alloc: 72 MB

[...]

PR #4042, opened 2024-11-12
Simplify HasCorrectTimeout

SimonCropp:
it was never called with a null value. so move it to a calculated property on the attribute

### DIFF
--- a/src/TestFramework/TestFramework/Attributes/TestMethod/TimeoutAttribute.cs
+++ b/src/TestFramework/TestFramework/Attributes/TestMethod/TimeoutAttribute.cs
@@ -25,7 +25,11 @@ public sealed class TimeoutAttribute : Attribute
     /// <param name="timeout">
     /// The timeout.
     /// </param>
-    public TimeoutAttribute(TestTimeout timeout) => Timeout = (int)timeout;
+    public TimeoutAttribute(TestTimeout timeout)
+    {
+        Timeout = (int)timeout;
+        HasCorrectTimeout = Timeout > 0;
+    }
 
     /// <summary>
     /// Gets the timeout in milliseconds.
@@ -51,6 +55,11 @@ public sealed class TimeoutAttribute : Attribute
             _isCooperativeCancellation = value;
         }
     }
+
+    /// <summary>
+    /// Gets a value indicating whether the instance has the correct test timeout signature.
+    /// </summary>
+    internal bool HasCorrectTimeout { get; private set; }
 
     internal bool IsCooperativeCancellationSet { get; private set; }
 }
--- a/src/TestFramework/TestFramework/Attributes/TestMethod/TimeoutAttribute.cs
+++ b/src/TestFramework/TestFramework/Attributes/TestMethod/TimeoutAttribute.cs
@@ -17,7 +17,11 @@ public sealed class TimeoutAttribute : Attribute
     /// <param name="timeout">
     /// The timeout in milliseconds.
     /// </param>
-    public TimeoutAttribute(int timeout) => Timeout = timeout;
+    public TimeoutAttribute(int timeout)
+    {
+        Timeout = timeout;
+        HasCorrectTimeout = Timeout > 0;
+    }
 
     /// <summary>
     /// Initializes a new instance of the <see cref="TimeoutAttribute"/> class with a preset timeout.
@@ -26,9 +30,8 @@ public sealed class TimeoutAttribute : Attribute
     /// The timeout.
     /// </param>
     public TimeoutAttribute(TestTimeout timeout)
+        : this((int)timeout)
     {
-        Timeout = (int)timeout;
-        HasCorrectTimeout = Timeout > 0;
     }
 
     /// <summary>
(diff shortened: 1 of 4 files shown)


Evangelink (commented):
Assigning @nohwnd for review

src/TestFramework/TestFramework/Attributes/TestMethod/TimeoutAttribute.cs
```diff
@@ -52,5 +59,10 @@ public bool CooperativeCancellation
         }
     }
 
+    /// <summary>
+    /// Gets a value indicating whether the instance has the correct test timeout signature.
+    /// </summary>
+    internal bool HasCorrectTimeout { get; private set; }
```
Youssef1313: My personal preference here is to keep the constructors as they were before and switch this to:

```suggestion
    internal bool HasCorrectTimeout => Timeout > 0;
```

test/UnitTests/MSTestAdapter.UnitTests/Extensions/MethodInfoExtensionsTests.cs
```diff
@@ -244,28 +244,6 @@ public void HasCorrectTestMethodSignatureShouldReturnFalseForAsyncTestMethodsWit
...
-    {
-        MethodInfo methodInfo = typeof(DummyTestClass).GetMethod("PublicMethodWithInvalidTimeout");
-        Verify(!methodInfo.HasCorrectTimeout());
-    }
-
-    public void HasCorrectTimeoutShouldReturnTrueForMethodsWithTimeoutAttribute()
-    {
-        MethodInfo methodInfo = typeof(DummyTestClass).GetMethod("PublicMethodWithTimeout");
-        Verify(methodInfo.HasCorrectTimeout());
-    }
```
Youssef1313: These tests should probably be replaced by:

```csharp
    public void HasCorrectTimeoutShouldReturnFalseForMethodsWithInvalidTimeoutAttribute()
    {
        var timeoutAttribute = new TimeoutAttribute(-11);
        Verify(!timeoutAttribute.HasCorrectTimeout);
    }

    public void HasCorrectTimeoutShouldReturnTrueForMethodsWithTimeoutAttribute()
    {
        var timeoutAttribute = new TimeoutAttribute(11);
        Verify(timeoutAttribute.HasCorrectTimeout);
    }
```

Also, now `PublicMethodWithInvalidTimeout` and `PublicMethodWithTimeout` are unused I think

nohwnd:
The code looks correct, I've initially added it here, as a replacement for repeated calls to GetAttributes, 

https://github.com/microsoft/testfx/pull/2839/files

which reduced the numbers of calls to reflection (I think we queried the attributes 4 or 5 times to figure out the attribute data).

And then with further optimizations the code was never called with null as pointed out by this PR. 

The code where we query the attributes is using the reflection helper methods, so it all should be correct.

test/UnitTests/MSTestAdapter.UnitTests/Extensions/MethodInfoExtensionsTests.cs
nohwnd: Looks like it. Can you add the code please @SimonCropp ?

src/TestFramework/TestFramework/Attributes/TestMethod/TimeoutAttribute.cs
```diff
@@ -17,15 +17,22 @@ public sealed class TimeoutAttribute : Attribute
     /// <param name="timeout">
     /// The timeout in milliseconds.
     /// </param>
-    public TimeoutAttribute(int timeout) => Timeout = timeout;
+    public TimeoutAttribute(int timeout)
+    {
+        Timeout = timeout;
+        HasCorrectTimeout = Timeout > 0;
```
nohwnd:
```suggestion
```

src/TestFramework/TestFramework/Attributes/TestMethod/TimeoutAttribute.cs
```diff
@@ -17,15 +17,21 @@ public sealed class TimeoutAttribute : Attribute
     /// <param name="timeout">
     /// The timeout in milliseconds.
     /// </param>
-    public TimeoutAttribute(int timeout) => Timeout = timeout;
+    public TimeoutAttribute(int timeout)
+    {
+        Timeout = timeout;
+    }
```
nohwnd:
```suggestion
    public TimeoutAttribute(int timeout) => Timeout = timeout;
```

src/TestFramework/TestFramework/Attributes/TestMethod/TimeoutAttribute.cs
```diff
@@ -17,15 +17,21 @@ public sealed class TimeoutAttribute : Attribute
...
     /// Initializes a new instance of the <see cref="TimeoutAttribute"/> class with a preset timeout.
     /// </summary>
     /// <param name="timeout">
     /// The timeout.
     /// </param>
-    public TimeoutAttribute(TestTimeout timeout) => Timeout = (int)timeout;
+    public TimeoutAttribute(TestTimeout timeout)
+        : this((int)timeout)
+    {
+    }
```
nohwnd:
```suggestion
    public TimeoutAttribute(TestTimeout timeout) => Timeout = (int)timeout;
```

test/UnitTests/MSTestAdapter.UnitTests/Extensions/MethodInfoExtensionsTests.cs
nohwnd: added

PR #4044, opened 2024-11-12
remove private types

SimonCropp:
just checking if this passes CI


Evangelink (changes requested):
These types are used for .NET Framework to allow testing private types

PR #4046, opened 2024-11-12
GetCustomAttributes cant return null

SimonCropp:


Evangelink:
This is changing the public API so it will have to wait until the V4. It's not super easy to know the impact this could have, although nullability is usually minimal.

PR #4048, opened 2024-11-12
leverage out null pattern in IsValidDeploymentItem

SimonCropp:
to prevent incorrect use of the out variable


Evangelink (dismissed):

Evangelink:
@SimonCropp this test https://github.com/microsoft/testfx/blob/87aed30d19af055e443a2a86c89c554e03501c4d/test/UnitTests/MSTestAdapter.PlatformServices.UnitTests/Utilities/ns13DeploymentItemUtilityTests.cs#L392 is failing as it expects the string to be empty and not null.

Evangelink:
@SimonCropp would you mind fixing the tests?

Evangelink (approved):

Issue #3315, opened 2024-07-23
new Analyzer+CodeFix: reccomend against wrong assertion method (e.g. `Assert.IsTrue(x != null)`)


LukasGelke:
## Summary
As some assertions may be written in different ways, some of them provide a better experience: better/richer message, clearer intent, better readability/maintainability.

## Background and Motivation
- default Exception-Messages are better
```cs
string text = "Mouse";
Assert.IsTrue("House" == text);
//  Assert.IsTrue failed.
Assert.AreEqual("House", text);
// Assert.AreEqual failed. Expected:<House>. Actual:<Mouse>. 
```
This is also useful when a test is failing in a pipeline, or you get the stack trace only. This way you *may* not have to dig too deep and "waste" time debugging, just to get to that point to know what `text` actually is. Or when you get a notification from the pipeline a la `test method "VeryImportantTest" failed with message "Assert.IsTrue failed."`

- we have a company-internal analyzer that does this.
- xUnit has a similar analyzers: 
  - https://xunit.net/xunit.analyzers/rules/xUnit2003
  - https://xunit.net/xunit.analyzers/rules/xUnit2004
- (though for some tests, it is explicitly wanted due to explicit testing of operator overload; but for those case the warning is just disabled via `#pragma` or Attribute)

## Proposed Feature
essentially all permutations of IsFalse/IsTrue with `is`/`is not`/`==`/`!=`
- Assert.IsNotNull, e.g.
  -  `Assert.IsTrue(x != null)` => `Assert.IsNotNull(x)`
  -  `Assert.IsTrue(null != x)` => `Assert.IsNotNull(x)`
  -  `Assert.IsFalse(x != null)` => `Assert.IsNull(x)`
  -  `Assert.IsTrue(x is not null)` => `Assert.IsNotNull(x)`
  - ...
  -  `Assert.AreNotEqual(null, x)` => `Assert.IsNotNull(x)`
  - ...
  -  either side may be `null` literal

- Assert.Are[Not]Equal, e.g.
  - `Assert.IsTrue(x == y)` => `Assert.AreEqual(x, y)`
  - `Assert.IsTrue(x != y)` => `Assert.AreNotEqual(x, y)`
  -  `Assert.IsFalse(x == y)` => `Assert.AreNotEqual(x, y)`
  -  `Assert.IsFalse(x != y)` => `Assert.AreEqual(x, y)`
  - ...
[...]

PR #4052, opened 2024-11-12
Implement 'Use proper Assert methods' analyzer

Youssef1313:
Implements the analyzer part of #3315. Codefix isn't yet implemented.

### DIFF
--- a/src/Analyzers/MSTest.Analyzers/UseProperAssertMethodsAnalyzer.cs
+++ b/src/Analyzers/MSTest.Analyzers/UseProperAssertMethodsAnalyzer.cs
@@ -29,7 +29,7 @@ namespace MSTest.Analyzers;
 /// <code>Assert.[IsTrue|IsFalse](x [==|!=] y)</code>
 /// </item>
 /// <item>
-/// <code>Assert.[AreEqual|AreNotEqual]([true|false], x)</code>
+/// <code>Assert.AreEqual([true|false], x)</code>
 /// </item>
 /// <item>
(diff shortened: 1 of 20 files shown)


Evangelink (approved):
Awesome work @Youssef1313

src/Analyzers/MSTest.Analyzers/UseProperAssertMethodsAnalyzer.cs
```diff
@@ -0,0 +1,238 @@
...
+    private static readonly LocalizableResourceString MessageFormat = new(nameof(Resources.UseProperAssertMethodsMessageFormat), Resources.ResourceManager, typeof(Resources));
+
+    internal static readonly DiagnosticDescriptor Rule = DiagnosticDescriptorHelper.Create(
+        DiagnosticIds.UseProperAssertMethodsRuleId,
+        Title,
+        MessageFormat,
+        null,
+        Category.Usage,
+        DiagnosticSeverity.Info,
+        isEnabledByDefault: false);
```
Youssef1313, line 66: @Evangelink I just remembered that we haven't discussed this. Is the current severity and being not enabled by default okay? Or should it be enabled by default?

Evangelink: I think we can enable by default, I am only careful about the `warning` severity and outside of breaking change version I try to set it as `warning` only for rules where failure to comply would result in runtime error as we have many users with warn as error who would get broken by bumping.

PR #4053, opened 2024-11-12
Consider readonly TestContext as valid

### LINKED ISSUE #4032: FP warning UTA031
`UTA031: class TestContextCtorWithReadOnlyProperty does not have valid TestContext property. TestContext must be of type TestContext, must be non-static, public and must not be read-only. For example: public TestContext TestContext.`

This is now a valid construct so the code warning should be updated

AB#2300884

Youssef1313:
Fixes #4032

### DIFF
--- a/src/Adapter/MSTest.TestAdapter/Discovery/TypeValidator.cs
+++ b/src/Adapter/MSTest.TestAdapter/Discovery/TypeValidator.cs
@@ -121,6 +121,9 @@ internal class TypeValidator
             return true;
         }
 
+        // NOTE: It's okay to not have a setter for the property at all.
+        // The user may set the TestContext property in the constructor.
+        // In case the user didn't set it in the constructor, the TestContextShouldBeValidAnalyzer will report a diagnostic.
         foreach (PropertyInfo pinfo in propertyInfo)
         {
             MethodInfo? getInfo = pinfo.GetMethod;
--- a/src/Adapter/MSTest.TestAdapter/Discovery/TypeValidator.cs
+++ b/src/Adapter/MSTest.TestAdapter/Discovery/TypeValidator.cs
@@ -123,14 +123,14 @@ internal class TypeValidator
 
         foreach (PropertyInfo pinfo in propertyInfo)
         {
-            MethodInfo? setInfo = pinfo.SetMethod;
-            if (setInfo == null)
+            MethodInfo? getInfo = pinfo.GetMethod;
+            if (getInfo == null)
             {
                 // we have a getter, but not a setter.
                 return false;
             }
 
-            if (setInfo.IsPrivate || setInfo.IsStatic || setInfo.IsAbstract)
+            if (getInfo.IsPrivate || getInfo.IsStatic || getInfo.IsAbstract)
             {
                 return false;
             }
--- a/src/Adapter/MSTest.TestAdapter/Discovery/TypeValidator.cs
+++ b/src/Adapter/MSTest.TestAdapter/Discovery/TypeValidator.cs
@@ -126,7 +126,7 @@ internal class TypeValidator
             MethodInfo? getInfo = pinfo.GetMethod;
             if (getInfo == null)
             {
-                // we have a getter, but not a setter.
+                // we don't have a getter.
                 return false;
             }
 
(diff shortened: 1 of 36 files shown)


Evangelink (commented):
Could you please double check if we have a test case where we would have readonly `TestContext` property and no ctor accepting `TestContext` parameter? I think we should still raise in this case because the property would end up not being set.

Youssef1313:
We don't have a good way to detect that at run-time. Do you think going for an analyzer would be the best approach for now?

test/IntegrationTests/MSTest.Acceptance.IntegrationTests/TestContextTests.cs
Youssef1313, line 1: Note to self: add tests showing that no warning is issued when the field/property is assigned in constructor

src/Adapter/MSTest.TestAdapter/Discovery/TypeValidator.cs
```diff
@@ -121,16 +121,19 @@ internal static bool HasCorrectTestContextSignature(Type type)
             return true;
         }
 
+        // NOTE: It's okay to not have a setter for the property at all.
+        // The user may set the TestContext property in the constructor.
+        // In case the user didn't set it in the constructor, the TestContextShouldBeValidAnalyzer will report a diagnostic.
```
Evangelink, line 126: I have just looked at the code and we could query the list of ctor to verify if one as a `TestContext` parameter but I am also fine with the current simplification.

src/Analyzers/MSTest.Analyzers.CodeFixes/TestContextShouldBeValidFixer.cs
```diff
@@ -87,6 +83,12 @@ public static async Task<Document> FixMemberDeclarationAsync(Document document,
         {
             // ensure that the property has setter and getter
             var propertyDeclaration = (PropertyDeclarationSyntax)newMemberDeclaration;
+            if (!propertyDeclaration.Identifier.ValueText.Equals("TestContext", StringComparison.Ordinal))
```
Evangelink: Let's extract all these `TestContext` string literals into a constant please.

src/Analyzers/MSTest.Analyzers/Resources.resx
```diff
@@ -361,11 +361,11 @@ The type declaring these methods should also respect the following rules:
     <value>TestCleanup method should have valid layout</value>
   </data>
   <data name="TestContextShouldBeValidDescription" xml:space="preserve">
-    <value>TestContext property should follow the following layout to be valid:
+    <value>TestContext should be a non-static property or field that is assigned in the constructor. Otherwise, all the following is required for it to be valid:
```
Evangelink: 'TestContext' should be a field or property assigned in ctor or for a property set by MSTest, it should follow the layout:
...

Evangelink (commented):
Thanks for the updates @Youssef1313, a few nits but otherwise LGTM

src/Adapter/MSTest.TestAdapter/Discovery/TypeValidator.cs
Youssef1313: The analyzer should report that and it has the additional capability of being able to investigate the constructor body and make sure the property is actually assigned.

I didn't feel it's worth complicating that logic. Thinking more, I even started to think the analyzer could have false positives.

```csharp
[TestClass]
public abstract class BaseTestClass
{
    protected BaseTestClass(TestContext context, string extraParam)
    {
        TestContext = context;
        // Do whatever with 'extraParam'.
    }

    public TestContext TestContext { get; } // This is readonly, but the constructor isn't single parameter.
}

[TestClass]
public class ConcreteTestClass
{
    public ConcreteTestClass(TestContext context) : base(context, nameof(ConcreteTestClass))
    {
    }

    // test methods...
}
```

If we do the analysis in TypeValidator and there are any false positive warnings, that would actually prevent the test from running, while having the FPs from analyzer is safer as it won't affect whether the test is run or not, and users could just add suppressions when they encounter false positives (for sure we want to eliminate/reduce FPs as much as possible).

For the current "potential" FP, do you think it's worth investigating?

src/Analyzers/MSTest.Analyzers/Resources.resx
Youssef1313: Adjusted it per your suggestion but a little bit differently. Let me know if it looks good?

Evangelink (approved):

src/Analyzers/MSTest.Analyzers/Resources.resx
Evangelink: LGTM

src/Adapter/MSTest.TestAdapter/Discovery/TypeValidator.cs
Evangelink: The case above should work. I think the args validation should be a different rule.

Evangelink:
Happy with the current state so I'll move the PR as-is. Thanks @Youssef1313

PR #4054, opened 2024-11-12
Pass multiple errors

nohwnd:
Allow passing multiple errors to dotnet test, fix code to handle multiple errors without relying on Exception type that we won't serialize.

Fix discovery in dotnet test, and report exit codes and outputs when exe fails so we have more info in dotnet test.

### DIFF
--- a/eng/Version.Details.xml
+++ b/eng/Version.Details.xml
@@ -1,25 +1,25 @@
 <?xml version="1.0" encoding="utf-8"?>
 <Dependencies>
   <ToolsetDependencies>
-    <Dependency Name="Microsoft.DotNet.Arcade.Sdk" Version="10.0.0-beta.24560.1">
+    <Dependency Name="Microsoft.DotNet.Arcade.Sdk" Version="10.0.0-beta.24561.1">
       <Uri>https://github.com/dotnet/arcade</Uri>
-      <Sha>232061b49ae2157efbb83acde9acae79ef3d6d40</Sha>
+      <Sha>32b8acf5b6f79e78a7cc5f7b7c619f68ff0c97ad</Sha>
     </Dependency>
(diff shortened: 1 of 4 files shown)


src/Platform/Microsoft.Testing.Platform/ServerMode/DotnetTest/IPC/ObjectFieldIds.cs
```diff
@@ -80,13 +80,19 @@ internal static class FailedTestResultMessageFieldsId
     public const ushort State = 3;
     public const ushort Duration = 4;
     public const ushort Reason = 5;
-    public const ushort ErrorMessage = 6;
-    public const ushort ErrorStackTrace = 7;
+    public const ushort ExceptionMessageList = 6;
     public const ushort StandardOutput = 8;
```
mariam-abdulla, line 84: We are missing the 7 value here.

src/Platform/Microsoft.Testing.Platform/ServerMode/DotnetTest/IPC/Serializers/TestResultMessagesSerializer.cs
```diff
@@ -248,13 +248,43 @@ private static List<FailedTestResultMessage> ReadFailedTestMessagesPayload(Strea
                         reason = ReadStringValue(stream, fieldSize);
                         break;
 
-                    case FailedTestResultMessageFieldsId.ErrorMessage:
-                        errorMessage = ReadStringValue(stream, fieldSize);
-                        break;
+                    case FailedTestResultMessageFieldsId.ExceptionMessageList:
```
mariam-abdulla, line 251: Please extract this as a separate method.

nohwnd:
Merging this as agreed with Mariam so she has easier time using my changes, will address the refactorings in separate PR/

PR #4055, opened 2024-11-12
[PROTOTYPE] Implement TestCultureAttribute

### LINKED ISSUE #579: Create a TestCulture attribute to set the culture of the test
Pretty much the same as NUnit's SetCultureAttribute and SetUICultureAttribute (although I personally don't like the naming convention) and maybe one attribute could be used for both?

Youssef1313:
Fixes #579

### DIFF
--- a/src/TestFramework/TestFramework/Attributes/TestMethod/TestMethodAttribute.cs
+++ b/src/TestFramework/TestFramework/Attributes/TestMethod/TestMethodAttribute.cs
@@ -1,6 +1,8 @@
 ﻿// Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
+using System.Globalization;
+
 namespace Microsoft.VisualStudio.TestTools.UnitTesting;
 
 /// <summary>
@@ -9,6 +11,29 @@ namespace Microsoft.VisualStudio.TestTools.UnitTesting;
 [AttributeUsage(AttributeTargets.Method)]
 public class TestMethodAttribute : Attribute
 {
+    private protected readonly struct ScopedCultureDisposable : IDisposable
+    {
+        private readonly CultureInfo _previousCulture;
+        private readonly CultureInfo _previousUICulture;
+
+        public ScopedCultureDisposable(string cultureName)
+        {
+            _previousCulture = CultureInfo.CurrentCulture;
+            _previousUICulture = CultureInfo.CurrentUICulture;
+
+            var newCulture = new CultureInfo(cultureName);
+            // TODO: Should we set both? Should we have different attribute? Same attribute with two arguments?
+            CultureInfo.CurrentCulture = newCulture;
+            CultureInfo.CurrentUICulture = newCulture;
+        }
+
+        public void Dispose()
+        {
+            CultureInfo.CurrentCulture = _previousCulture;
+            CultureInfo.CurrentUICulture = _previousUICulture;
+        }
+    }
+
     /// <summary>
     /// Initializes a new instance of the <see cref="TestMethodAttribute"/> class.
     /// </summary>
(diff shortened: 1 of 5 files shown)


src/TestFramework/TestFramework/Attributes/TestMethod/TestMethodAttribute.cs
```diff
@@ -9,6 +11,29 @@ namespace Microsoft.VisualStudio.TestTools.UnitTesting;
...
+        private readonly CultureInfo _previousCulture;
+        private readonly CultureInfo _previousUICulture;
+
+        public ScopedCultureDisposable(string cultureName)
+        {
+            _previousCulture = CultureInfo.CurrentCulture;
+            _previousUICulture = CultureInfo.CurrentUICulture;
+
+            var newCulture = new CultureInfo(cultureName);
+            // TODO: Should we set both? Should we have different attribute? Same attribute with two arguments?
```
Youssef1313, line 25: This needs to be discussed first.

src/TestFramework/TestFramework/Attributes/TestMethod/TestMethodAttribute.cs
```diff
@@ -36,5 +61,19 @@ public TestMethodAttribute()
...
+    {
+        using (SetCultureForTest(testMethod))
+        {
+            return [testMethod.Invoke(null)];
+        }
+    }
+
+    // TestMethodInfo isn't accessible here :/
+    // Can we add TestCultureName to the *public* interface?
+    // Or should we introduce an internal interface ITestMethod2 : ITestMethod :/
```
Youssef1313, line 74: cc @Evangelink

src/TestFramework/TestFramework/Attributes/TestMethod/TestMethodAttribute.cs
```diff
@@ -9,6 +11,29 @@ namespace Microsoft.VisualStudio.TestTools.UnitTesting;
...
+        private readonly CultureInfo _previousUICulture;
+
+        public ScopedCultureDisposable(string cultureName)
+        {
+            _previousCulture = CultureInfo.CurrentCulture;
+            _previousUICulture = CultureInfo.CurrentUICulture;
+
+            var newCulture = new CultureInfo(cultureName);
+            // TODO: Should we set both? Should we have different attribute? Same attribute with two arguments?
+            CultureInfo.CurrentCulture = newCulture;
```
Youssef1313, line 26: Important consideration: Validate if that approach is going to work well for tests running in parallel

nohwnd: https://github.com/dotnet/runtime/blob/main/src/libraries/System.Private.CoreLib/src/System/Globalization/CultureInfo.cs#L376-L380

Looks to by async level safe, which is ideal for us, because all tests run in their own async context.

This small experiment where I try to "spray" the new culture all over the thread pool seems to confirm that async is properly isolated both on .net and .net framework.

```cs
using System.Collections.Concurrent;
using System.Globalization;


var tsk = Task.Run(async () =>
{
    while (true)
    {
        await Task.Delay(1000);
        Console.WriteLine($"Thread: {Thread.CurrentThread.ManagedThreadId}, Culture: {CultureInfo.CurrentCulture.Name}");
        await Task.Yield();
    }
});

await Task.Delay(1000);

var tasks = new List<Task>();
var threads = new ConcurrentBag<int>();
foreach (var i in Enumerable.Range(0, 1000))
{
    tasks.Add(Task.Run(() =>
    {
        threads.Add(Thread.CurrentThread.ManagedThreadId);
        CultureInfo.CurrentCulture = CultureInfo.GetCultureInfo("cs-CZ");
        Console.WriteLine($"Thread: {Thread.CurrentThread.ManagedThreadId}, Culture: {CultureInfo.CurrentCulture.Name}");
        return Task.CompletedTask;
    }));
}

Task.WaitAll(tasks.ToArray());

Console.WriteLine(string.Join(", ", threads.Distinct()));
Task.WaitAll(tsk);
```

Youssef1313:
Postponing this for now.

PR #4057, opened 2024-11-13
Refactor unnecessary try/catch

### LINKED ISSUE #3006: ConsiderEmptyDataSourceAsInconclusive throws to catch the exception on the next line
When 0 items are produced and the code does not throw, we throw just to catch the exception.  This could be made more efficient by extracting the catch and using it in `if` and `catch` paths.

https://grep.app/search?q=%26%26%20.%2AConsiderEmptyDataSourceAsInconclusive&regexp=true

Code is introduced here: https://github.com/microsoft/testfx/pull/2771/files

PR has no comments that would mention this.

Youssef1313:
Fixes #3006

### DIFF
--- a/src/Adapter/MSTest.TestAdapter/Discovery/AssemblyEnumerator.cs
+++ b/src/Adapter/MSTest.TestAdapter/Discovery/AssemblyEnumerator.cs
@@ -413,13 +413,26 @@ internal class AssemblyEnumerator : MarshalByRefObject
 
             if (!data.Any())
             {
-                UnitTestElement discoveredTest = test.Clone();
-                // Make the test not data driven, because it had no data.
-                discoveredTest.TestMethod.DataType = DynamicDataType.None;
-                discoveredTest.DisplayName = dataSource.GetDisplayName(methodInfo, null) ?? discoveredTest.DisplayName;
+                if (MSTestSettings.CurrentSettings.ConsiderEmptyDataSourceAsInconclusive)
+                {
+                    UnitTestElement discoveredTest = test.Clone();
+                    // Make the test not data driven, because it had no data.
+                    discoveredTest.TestMethod.DataType = DynamicDataType.None;
+                    discoveredTest.DisplayName = dataSource.GetDisplayName(methodInfo, null) ?? discoveredTest.DisplayName;
 
-                tests.Add(discoveredTest);
-                continue;
+                    tests.Add(discoveredTest);
+
+                    continue;
+                }
+                else
+                {
+                    throw new ArgumentException(
+                        string.Format(
+                            CultureInfo.InvariantCulture,
+                            FrameworkMessages.DynamicDataIEnumerableEmpty,
+                            "GetData",
+                            dataSource.GetType().Name));
+                }
             }
 
             var testDisplayNameFirstSeen = new Dictionary<string, int>();
--- a/src/Adapter/MSTest.TestAdapter/Discovery/AssemblyEnumerator.cs
+++ b/src/Adapter/MSTest.TestAdapter/Discovery/AssemblyEnumerator.cs
@@ -426,12 +426,7 @@ internal class AssemblyEnumerator : MarshalByRefObject
                 }
                 else
                 {
-                    throw new ArgumentException(
-                        string.Format(
-                            CultureInfo.InvariantCulture,
-                            FrameworkMessages.DynamicDataIEnumerableEmpty,
-                            "GetData",
-                            dataSource.GetType().Name));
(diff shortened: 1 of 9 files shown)


src/Adapter/MSTest.TestAdapter/Discovery/AssemblyEnumerator.cs
```diff
@@ -409,24 +409,22 @@ private static bool ProcessTestDataSourceTests(UnitTestElement test, ReflectionT
 
             // This code is to discover tests. To run the tests code is in TestMethodRunner.ExecuteDataSourceBasedTests.
             // Any change made here should be reflected in TestMethodRunner.ExecuteDataSourceBasedTests as well.
-            try
-            {
-                data = dataSource.GetData(methodInfo);
+            data = dataSource.GetData(methodInfo);
 
-                if (!data.Any())
+            if (!data.Any())
```
Youssef1313, line 419: This actually was NEVER true before.

GetData used to throw the exception when it's about to return empty.

Evangelink (dismissed):

Youssef1313:
@Evangelink I'm disabling auto-merge as I'm making changes since your last review. Once I'm able to get it green, I'll let you know to re-review.

src/TestFramework/TestFramework/Interfaces/EmptyDataSourceExceptionInfoExtensions.cs
```diff
@@ -0,0 +1,26 @@
...
+
+internal static class EmptyDataSourceExceptionInfoExtensions
+{
+    internal static ArgumentException GetExceptionForEmptyDataSource(this ITestDataSource dataSource, MethodInfo testMethodInfo)
+        => dataSource is ITestDataSourceEmptyDataSourceExceptionInfo info ? info.GetExceptionForEmptyDataSource(testMethodInfo) : new ArgumentException(
+            string.Format(
+                CultureInfo.InvariantCulture,
+                FrameworkMessages.DynamicDataIEnumerableEmpty,
+                "GetData",
+                dataSource.GetType().Name));
```
Evangelink: Not mandatory but I tend to find it easier when scanning code to have the ternary split like

```suggestion
        => dataSource is ITestDataSourceEmptyDataSourceExceptionInfo info
            ? info.GetExceptionForEmptyDataSource(testMethodInfo)
            : new ArgumentException(
                string.Format(
                    CultureInfo.InvariantCulture,
                    FrameworkMessages.DynamicDataIEnumerableEmpty,
                    "GetData",
                    dataSource.GetType().Name));
```

src/Adapter/MSTest.TestAdapter/Discovery/AssemblyEnumerator.cs
```diff
@@ -409,25 +409,26 @@ private static bool ProcessTestDataSourceTests(UnitTestElement test, ReflectionT
 
             // This code is to discover tests. To run the tests code is in TestMethodRunner.ExecuteDataSourceBasedTests.
             // Any change made here should be reflected in TestMethodRunner.ExecuteDataSourceBasedTests as well.
-            try
+            data = dataSource.GetData(methodInfo);
+
+            if (!data.Any())
             {
-                data = dataSource.GetData(methodInfo);
+                if (MSTestSettings.CurrentSettings.ConsiderEmptyDataSourceAsInconclusive)
```
Evangelink: Optional nit: I usually to prefer exit conditions first to reduce mental overhead so I would suggest to revert the if to throw and not to have some else.

src/Adapter/MSTest.TestAdapter/Execution/TestMethodRunner.cs
```diff
@@ -311,25 +311,26 @@ private bool ExecuteDataSourceBasedTests(List<TestResult> results)
...
-                        dataSource = testDataSource.GetData(_testMethodInfo.MethodInfo);
 
-                        if (!dataSource.Any())
+                    // This code is to execute tests. To discover the tests code is in AssemblyEnumerator.ProcessTestDataSourceTests.
+                    // Any change made here should be reflected in AssemblyEnumerator.ProcessTestDataSourceTests as well.
+                    dataSource = testDataSource.GetData(_testMethodInfo.MethodInfo);
+
+                    if (!dataSource.Any())
+                    {
+                        if (MSTestSettings.CurrentSettings.ConsiderEmptyDataSourceAsInconclusive)
```
Evangelink: Optional nit: I usually to prefer exit conditions first to reduce mental overhead so I would suggest to revert the if to throw and not to have some else.

Evangelink (approved):
Small nits but LGTM, thanks @Youssef1313

PR #4058, opened 2024-11-13
Fix DataSourceAttribute not being discovered with TestDataSourceDiscoveryOption.DuringDiscovery

### LINKED ISSUE #4029: Datasource is not working with MSTest.TestFramework Adapter and MSTest.TestAdapter from version 3.5.0.
We were using MSTest.TestFramework Adapter and MSTest.TestAdapter version **2.2.3** and tried to update it to the latest version **3.6.2**. But after updating,  [DataSource("TC123456")] does not seem to work. We do not see any error either in the Testlog. It just can not start execution of the test.

We are using **app.config** file which referes to this Datasource.

> <connectionStrings>
> <add name="ABC.TestAutomation.Application.TestData.ABC" connectionString="|DataDirectory|\Data\TestData.xml"
>   providerName="Microsoft.VisualStudio.TestTools.DataSource.XML" />
> </connectionStrings>
> 
> <dataSources>
>  <add name="TC123456" connectionString="ABC.TestAutomation.Application.TestData.ABC" dataTableName="TC123456" dataAccessMethod="Sequential" />
> </dataSources>

We are using Microsoft.NET.Test.Sdk - 17.11.1 and .NET Framework 4.7.2.

Also we noticed that Datasource problems exists with some of the versions.

2.2.5    works
2.2.10 - works

3.0.0 - does not work
3.0.1 - works 
3.1.1 - works
3.2.0 - does not work
3.3.0 - works

3.4.0 - does not work
3.4.2 - works
3.4.3 - works

3.5.0 - does not work
3.5.1 - does not work
3.5.2 - does not work
3.6.0 - does not work
3.6.1 - does not work
3.6.2 - does not work

Why are some of the versions giving problem with Datasource ? Could you pleaase help us using the latest version 3.6.2 and still be able to work with Datasource ?

Thanks and Regards.

AB#2300883

Youssef1313:
Fixes #4029

### DIFF
--- a/eng/install-access-database-engine.ps1
+++ b/eng/install-access-database-engine.ps1
@@ -1,0 +1,7 @@
+# This is required for DataSourceTests
+# Otherwise, the tests will fail with:
+# The unit test adapter failed to connect to the data source or to read the data. For more information on troubleshooting this error, see "Troubleshooting Data-Driven Unit Tests" (http://go.microsoft.com/fwlink/?LinkId=62412) in the MSDN Library. Error details: The 'Microsoft.Ace.OLEDB.12.0' provider is not registered on the local machine.
+#     at Microsoft.VisualStudio.TestPlatform.MSTestAdapter.PlatformServices.TestDataSource.GetData(ITestMethod testMethodInfo, ITestContext testContext) in /_/src/Adapter/MSTestAdapter.PlatformServices/Services/TestDataSource.cs:84
+# The direct download link originates from https://www.microsoft.com/en-us/download/details.aspx?id=54920&msockid=01fa77be234c617f31936293223560aa
+Invoke-RestMethod https://download.microsoft.com/download/3/5/C/35C84C36-661A-44E6-9324-8786B8DBE231/accessdatabaseengine_X64.exe -OutFile ./accessdatabaseengine_X64.exe
+Start-Process ./accessdatabaseengine_X64.exe -Wait -ArgumentList "/quiet /passive /norestart"
--- a/src/Adapter/MSTest.TestAdapter/Discovery/AssemblyEnumerator.cs
+++ b/src/Adapter/MSTest.TestAdapter/Discovery/AssemblyEnumerator.cs
@@ -285,6 +285,8 @@ internal class AssemblyEnumerator : MarshalByRefObject
             return false;
         }
 
+        DynamicDataType originalDataType = test.TestMethod.DataType;
+
         // PERF: For perf we started setting DataType in TypeEnumerator, so when it is None we will not reach this line.
         // But if we do run this code, we still reset it to None, because the code that determines if this is data drive test expects the value to be None
         // and only sets it when needed.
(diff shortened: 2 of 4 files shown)


Youssef1313:
@Evangelink the test here fails in CI with:

```
    The unit test adapter failed to connect to the data source or to read the data. For more information on troubleshooting this error, see "Troubleshooting Data-Driven Unit Tests" (http://go.microsoft.com/fwlink/?LinkId=62412) in the MSDN Library. Error details: The 'Microsoft.Ace.OLEDB.12.0' provider is not registered on the local machine.
```

That is a missing installation of something in CI. Not sure yet what we need to do here.

src/Adapter/MSTest.TestAdapter/Discovery/AssemblyEnumerator.cs
```diff
@@ -403,8 +403,14 @@ private static bool TryProcessTestDataSourceTests(UnitTestElement test, TestMeth
     private static bool ProcessTestDataSourceTests(UnitTestElement test, ReflectionTestMethodInfo methodInfo, IEnumerable<ITestDataSource> testDataSources,
```
Youssef1313: A different possible fix is to change:

https://github.com/microsoft/testfx/blob/04f901ceb6e7069a65c2734ab49081ba1fd5c07c/src/Adapter/MSTest.TestAdapter/Discovery/AssemblyEnumerator.cs#L288

to:

```csharp
if (test.TestMethod.DataType != DynamicDataType.ITestDataSource)
```

nohwnd: I think this might be a better fix to make, together with a comment that DataSourceAttribute tests connect to external systems like databases, and we generally want to postpone that until run.

And then renaming the ProcessTestDataSourceTests to ProcessITestDataSourceTests, and adding a comment that this does no apply to DataSourceAttribute tests. 

Is this code path taken only when TestDataSourceDiscoveryOption.DuringDiscovery is set? It might be also worth to mention that in the comment.

eng/install-access-database-engine.ps1
```diff
@@ -0,0 +1,7 @@
+# This is required for DataSourceTests
+# Otherwise, the tests will fail with:
+# The unit test adapter failed to connect to the data source or to read the data. For more information on troubleshooting this error, see "Troubleshooting Data-Driven Unit Tests" (http://go.microsoft.com/fwlink/?LinkId=62412) in the MSDN Library. Error details: The 'Microsoft.Ace.OLEDB.12.0' provider is not registered on the local machine.
+#     at Microsoft.VisualStudio.TestPlatform.MSTestAdapter.PlatformServices.TestDataSource.GetData(ITestMethod testMethodInfo, ITestContext testContext) in /_/src/Adapter/MSTestAdapter.PlatformServices/Services/TestDataSource.cs:84
+# The direct download link originates from https://www.microsoft.com/en-us/download/details.aspx?id=54920&msockid=01fa77be234c617f31936293223560aa
+Invoke-RestMethod https://download.microsoft.com/download/3/5/C/35C84C36-661A-44E6-9324-8786B8DBE231/accessdatabaseengine_X64.exe -OutFile ./accessdatabaseengine_X64.exe
+Start-Process ./accessdatabaseengine_X64.exe -Wait -ArgumentList "/quiet /passive /norestart"
```
nohwnd, line 7: Would be nice to have a lighter way to test this, but I guess you investigated that. E.g inmemory "db" data source.

Youssef1313: Not really sure if it's possible with the current way TestDataSource is implemented

nohwnd (approved):

PR #4059, opened 2024-11-13
Support ValueTask-returning test methods for all target frameworks

### LINKED ISSUE #4056: Allow tests that return a ValueTask to be discovered
## Summary
When using DataTestMethod and DataRow, sometimes only some of the tests will be asyncronous. It'd be nice if test methods that return ValueTask instead of Task or Void can be discovered.

This would be useful if the tests sometime run an async method to get data to pass to it or if the awaited methods in a test are all ValueTask

Here's some example code of when this might be useful :
<details>
<summary>Example Code</summary>
<div>

```csharp

using System;
using System.IO;
using System.Threading.Tasks;

namespace TestProject1
{

    [TestClass]
    public class UnitTest1
    {
        public static class Uploader
        {
            // Uploads data or returns false if data is missing
            public static bool Upload(MyFileObject mfo)
            {
                return !(mfo.Data == null || mfo.Data.Length == 0) && mfo.FileName == "Report.pdf";
            }
        }

        public class MyFileObject
        {
            public string FileName { get; set; }
            public byte[] Data { get; set; }
        }


        public enum FileUploadValues
        {
            FileIsUploaded,
            FileBytesIsNull,
            FileBytesIsEmpty,
        }

        [ExcludeFromCodeCoverage]
        [DataTestMethod]
        [DataRow(FileUploadValues.FileIsUploaded)]
        [DataRow(FileUploadValues.FileBytesIsNull)]
        [DataRow(FileUploadValues.FileBytesIsEmpty)]
        public async ValueTask Early_Return_Tests(FileUploadValues fileUploadValue)
        {
            byte[] fileBytes = null;
            if (fileUploadValue == FileUploadValues.FileIsUploaded)
            {
                using (var ms = new MemoryStream())
                {
                    using (var fs = File.Open("./myfile.txt", FileMode.Open, FileAccess.Read, FileShare.Read))
                    {
                        await fs.CopyToAsync(ms);
                        fileBytes = ms.ToArray();
                    }
                }
            }
            else if (fileUploadValue == FileUploadValues.FileBytesIsEmpty)
            {
                fileBytes = Array.Empty<byte>();
            }

            var fileObj = new MyFileObject
            {
                Data = fileBytes,
                FileName = "INVALID_FILE_NAME"
            };

            bool isSuccessful = Uploader.Upload(fileObj);

            Assert.IsFalse(isSuccessful);
        }
    }
}
```

</div>

AB#2300885

Youssef1313:
Fixes #4056

### DIFF
--- a/Directory.Packages.props
+++ b/Directory.Packages.props
@@ -45,6 +45,7 @@
     <PackageVersion Include="Microsoft.TestPlatform.TranslationLayer" Version="$(MicrosoftNETTestSdkVersion)" />
     <PackageVersion Include="Microsoft.WindowsAppSDK" Version="1.0.0" />
     <PackageVersion Include="Polyfill" Version="7.4.0" />
+    <PackageVersion Include="System.Threading.Tasks.Extensions" Version="4.5.4" />
     <!-- CVE-2019-0820 -->
     <PackageVersion Include="System.Diagnostics.TextWriterTraceListener" Version="4.3.0" />
     <!-- CVE-2019-0981, CVE-2019-0980, CVE-2019-0657 -->
--- a/src/Adapter/MSTest.TestAdapter/Extensions/MethodInfoExtensions.cs
+++ b/src/Adapter/MSTest.TestAdapter/Extensions/MethodInfoExtensions.cs
@@ -103,9 +103,7 @@ internal static class MethodInfoExtensions
     /// <returns>True if the method has a void/task return type..</returns>
     internal static bool IsValidReturnType(this MethodInfo method)
         => ReflectHelper.MatchReturnType(method, typeof(Task))
-#if NETCOREAPP
         || ReflectHelper.MatchReturnType(method, typeof(ValueTask))
-#endif
         || (ReflectHelper.MatchReturnType(method, typeof(void)) && method.GetAsyncTypeName() == null);
 
     /// <summary>
@@ -201,11 +199,9 @@ internal static class MethodInfoExtensions
         {
             task.GetAwaiter().GetResult();
         }
-#if NET6_0_OR_GREATER
         else if (invokeResult is ValueTask valueTask)
         {
             valueTask.GetAwaiter().GetResult();
         }
-#endif
     }
 }
--- a/src/Adapter/MSTest.TestAdapter/MSTest.TestAdapter.csproj
+++ b/src/Adapter/MSTest.TestAdapter/MSTest.TestAdapter.csproj
@@ -53,6 +53,7 @@
 
   <ItemGroup>
     <PackageReference Include="Microsoft.TestPlatform.AdapterUtilities" />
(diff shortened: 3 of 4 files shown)


eng/Versions.props
```diff
@@ -13,4 +13,7 @@
     <MicrosoftTestingInternalFrameworkVersion>1.5.0-preview.24560.1</MicrosoftTestingInternalFrameworkVersion>
     <MSTestEngineVersion>1.0.0-alpha.24473.2</MSTestEngineVersion>
   </PropertyGroup>
+  <PropertyGroup Label="MSTest prod dependencies">
+    <SystemThreadingTasksExtensionsVersion>4.5.4</SystemThreadingTasksExtensionsVersion>
```
Evangelink: Let's have the version in the `Directory.Packages.props` file. I am trying to have only darc/maestro updated versions here.

src/Adapter/MSTest.TestAdapter/MSTest.TestAdapter.nuspec
```diff
@@ -6,15 +6,18 @@
...
       <group targetFramework="net462">
         <dependency id="Microsoft.Testing.Extensions.VSTestBridge" version="$TestingPlatformVersion$" />
         <dependency id="Microsoft.Testing.Platform.MSBuild" version="$TestingPlatformVersion$" />
+        <dependency id="System.Threading.Tasks.Extensions" version="$SystemThreadingTasksExtensionsVersion$" />
       </group>
       <group targetFramework="uap10.0" />
       <group targetFramework="netcoreapp3.1">
         <dependency id="Microsoft.Testing.Extensions.VSTestBridge" version="$TestingPlatformVersion$" />
         <dependency id="Microsoft.Testing.Platform.MSBuild" version="$TestingPlatformVersion$" />
+        <dependency id="System.Threading.Tasks.Extensions" version="$SystemThreadingTasksExtensionsVersion$" />
```
Evangelink: From the docs it seems that `ValueTask` is available for netcore3.1 so no need to add the deps.

Evangelink (commented):
A few small nits but otherwise LGTM. Let's see if there are some unanticipated problems with  our net framework users.

src/Adapter/MSTest.TestAdapter/MSTest.TestAdapter.nuspec
Youssef1313: Yes. The intent was for UWP. Thanks for catching that.

Youssef1313: @Evangelink uap10.0 seems to currently have no dependencies at all. I'm wondering why it's different from all others which have VSTestBridge and MTP.MSBuild?

Evangelink: The new platform hasn't been validated as working.

Evangelink (approved):

PR #4077, opened 2024-11-18
Use new output for VSTest acceptance test

Evangelink:
Hope to fix flakiness on acceptance test by removing the non-terminal logger display.


Evangelink:
Flakiness, force merging

PR #4078, opened 2024-11-18
Fix DataRow issues related to serialization/deserialization

Youssef1313:

### DIFF
--- a/src/Adapter/MSTest.TestAdapter/ObjectModel/TestMethod.cs
+++ b/src/Adapter/MSTest.TestAdapter/ObjectModel/TestMethod.cs
@@ -1,6 +1,7 @@
 ﻿// Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
+using System.Collections.Concurrent;
 using System.Collections.ObjectModel;
 using System.Diagnostics.CodeAnalysis;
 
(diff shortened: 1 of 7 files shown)


src/Adapter/MSTest.TestAdapter/ObjectModel/TestMethod.cs
```diff
@@ -28,6 +29,8 @@ public sealed class TestMethod : ITestMethod
     private string? _declaringClassFullName;
     private string? _declaringAssemblyName;
 
+    private static readonly ConcurrentDictionary<string, object?[]> DataDictionary = new();
```
Youssef1313: Notes:

- This is bad. The dictionary in theory can grow up indefinitely.
- Currently, UniqueName isn't really unique and we need a better mechanism here. One possibility is to attach a test property resulting from `Interlocked.Increment` on a static counter and use that as the dictionary key. Then, instead of deserializing we look up that key.
- This PR is mostly to open a discussion about the possibilities to improve the situation for some parameterized tests broken scenarios.

@Evangelink thoughts? I feel like some of the parameterized test issues could be fixable without the need for breaking changes?

Youssef1313: Resolved as I switched to the Interlocked.Increment. Will add another comment with the current concerns.

src/Adapter/MSTest.TestAdapter/ObjectModel/TestMethod.cs
```diff
@@ -28,20 +29,23 @@ public sealed class TestMethod : ITestMethod
     private string? _declaringClassFullName;
     private string? _declaringAssemblyName;
 
+    private static readonly ConcurrentDictionary<int, object?[]> DataDictionary = new();
```
Youssef1313, line 32: Notes:

- This is bad. The dictionary in theory can grow up indefinitely.
- This PR is mostly to open a discussion about the possibilities to improve the situation for some parameterized tests broken scenarios.

@Evangelink thoughts? I feel like some of the parameterized test issues could be fixable by trying to move forward in this direction without the need for breaking changes?

Youssef1313:
Closing for now per our discussion.

PR #4079, opened 2024-11-19
Use reflection provider for class constructors

nohwnd:
Fixes oversight in code where we grab the constructors from Reflection rather than getting them from the metadata provider.

Replace https://github.com/microsoft/testfx/pull/4033

### DIFF
--- a/src/Adapter/MSTest.TestAdapter/Execution/TypeCache.cs
+++ b/src/Adapter/MSTest.TestAdapter/Execution/TypeCache.cs
@@ -260,23 +260,41 @@ internal class TypeCache : MarshalByRefObject
     private TestClassInfo CreateClassInfo(Type classType, TestMethod testMethod)
     {
         IEnumerable<ConstructorInfo> constructors = PlatformServiceProvider.Instance.ReflectionOperations.GetDeclaredConstructors(classType);
-        ConstructorInfo? constructor = constructors.FirstOrDefault(ctor => ctor.GetParameters().Length == 0 && ctor.IsPublic);
-        bool isParameterLessConstructor;
+        (ConstructorInfo? CtorInfo, bool IsParameterless)? selectedConstructor = null;
 
-        if (classType.GetConstructor([typeof(TestContext)]) is { } testContextCtor)
+        foreach (ConstructorInfo ctor in constructors)
         {
-            constructor = testContextCtor;
-            isParameterLessConstructor = false;
+            if (!ctor.IsPublic)
+            {
+                continue;
+            }
+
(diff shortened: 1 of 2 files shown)


Evangelink (dismissed):

Evangelink (commented):
Some tests are now breaking on `MSTestAdapter.UnitTest`

src/Adapter/MSTest.TestAdapter/Execution/TypeCache.cs
```diff
@@ -260,24 +260,42 @@ private static bool TryGetUnescapedManagedTypeName(TestMethod testMethod, [NotNu
...
+            }
+
+            ParameterInfo[] parameters = ctor.GetParameters();
+
+            // There are just 2 ctor shapes that we know, so the code is quite simple,
+            // but if we add more, add a priority to the search, and shortcircuit this search so we only iterate
+            // through the collection once, to avoid re-allocating GetParameters multiple times.
+            if (parameters.Length == 1 && parameters[0].ParameterType == typeof(TestContext))
+            {
+                selectedConstructor ??= (ctor, IsParameterless: false);
```
Youssef1313: If we first found a parameterless constructor, then we found this, we want to always assign right?

```suggestion
                selectedConstructor = (ctor, IsParameterless: false);
```

src/Adapter/MSTest.TestAdapter/Execution/TypeCache.cs
```diff
@@ -260,24 +260,42 @@ private static bool TryGetUnescapedManagedTypeName(TestMethod testMethod, [NotNu
...
+
+                // This is the preferred constructor, no point in searching for more.
+                break;
+            }
+
+            if (parameters.Length == 0)
+            {
+                selectedConstructor = (ctor, IsParameterless: true);
```
Youssef1313: Maybe add `Debug.Assert(selectorConstructor is null);` before this line?

src/Adapter/MSTest.TestAdapter/Execution/TypeCache.cs
nohwnd: I think I meant to put this on the other case, so we select the first parameterless one.

src/Adapter/MSTest.TestAdapter/Execution/TypeCache.cs
nohwnd: moving the `??=` to the correct place solves the issue that was motivation for this Assert. Which was to error when we forget to break in the most preferred case, then falling through here would choose the non-preferred constructor.

PR #4112, opened 2024-11-21
DataSource draft

### LINKED ISSUE #4092: RFC011 - Custom data sources
# RFC 011 - Custom data sources

- [ ] Approved in principle
- [x] Under discussion
- [ ] Implementation
- [ ] Shipped

## Summary

A replacement for DataSourceAttribute, which is currently .NET Framework only, to make it more modern and bring the support for .NET Core.

## Motivation

User needs for DataSource to be supported on .NET Core. https://github.com/microsoft/testfx/issues/233

## Detailed design

Instead of having a single giant attribute like DataSourceAttribute, we will divide it into multiple attributes.

### `XmlDataSourceAttribute`

The attribute constructor will have three parameters, `(string xmlFilePath, string tableName)`

Previously with `DataSrouceAttribute`, the code used to look like this:

```csharp
[TestMethod]
[DeploymentItem("MyTests.Configuration.xml")]
[DataSource("Microsoft.VisualStudio.TestTools.DataSource.XML", "|DataDirectory|\\MyTests.Configuration.xml", "MyTable1", DataAccessMethod.Sequential)]
public void TestMethod()
{
    var foo = TestContext.DataRow["MyData"]);
}
```

XML file:

```xml
<Rows>
  <MyTable1>
    <MyData>true</MyData>
  </MyTable1>
  <MyTable1>
    <MyData>false</MyData>
  </MyTable1>

  <MyTable2>
    <MyData2>false</MyData2>
  </MyTable2>
</Rows>
```

Now, it will be:

```csharp
[TestMethod]
[XmlDataSource("MyTests.Configuration.xml", "MyTable1")]
public void TestMethod(bool myData /*parameters can't be validated at compile-time*/)
{
    // The parameters are the alternative of TestContext.DataRow["MyData"]);
}
```

The attribute will implement `ITestDataSource` and may be in a separate package.

### `CSVDataSourceAttribute`

The attribute constructor will have one parameters, `(string csvFilePath)`

Previously with `DataSrouceAttribute`, the code used to look like this:

```csharp
[TestMethod]
[DataSource("Microsoft.VisualStudio.TestTools.DataSource.CSV", "TestData.csv", "TestData#csv", DataAccessMethod.Sequential)]
public void TestMethod()
{
    var header1 = TestContext.DataRow["CsvHeader1"]);
}
```
Now, it will be:

```csharp
[TestMethod]
[CSVDataSource("TestData.csv")]
public void TestMethod(string header1, int header2, ... /*parameters can't be validated at compile-time*/)
{
    // The parameters are the alternative of TestContext.DataRow["CsvHeader1"]);
}
```

The attribute will implement `ITestDataSource` and may be in a separate package.

### `SqlDataSourceAttribute`

TBD

Concerns:

- Security: What's the best way to store connection strings?

## Drawbacks

Why should we *not* do this?

## Alternatives

What other designs have been considered? What is the impact of not doing this?

## Compatibility

This change isn't a breaking change. But it brings functionality to .NET ("Core") that is done differently on .NET Framework

### Unresolved questions

What parts of the design are still TBD?

Youssef1313:
Fixes #4092

### DIFF
--- a/src/TestFramework/TestFramework.Extensions.Csv/CsvDataSourceAttribute.cs
+++ b/src/TestFramework/TestFramework.Extensions.Csv/CsvDataSourceAttribute.cs
@@ -37,13 +37,7 @@ public sealed class CsvDataSourceAttribute : Attribute, ITestDataSource
             throw new FileNotFoundException($"Csv file '{fullPath}' cannot be found.", fullPath);
         }
 
-        string tableName = Path.GetFileName(fullPath).Replace('.', '#');
-
-        // We can map simplified CSVs to an OLEDB/Text connection, then proceed as normal
         using OleDbConnection connection = new();
-        using OleDbDataAdapter dataAdapter = new();
-        using OleDbCommandBuilder commandBuilder = new();
-        using OleDbCommand command = new();
 
         // We have to use the name of the folder which contains the CSV file in the connection string
         // If target platform is x64, then use CsvConnectionTemplate64 connection string.
@@ -56,13 +50,20 @@ public sealed class CsvDataSourceAttribute : Attribute, ITestDataSource
         // The connection will get closed when we dispose of it
         connection.Open();
 
+        using OleDbCommandBuilder commandBuilder = new();
+        string tableName = Path.GetFileName(fullPath).Replace('.', '#');
         string quotedTableName = commandBuilder.QuoteIdentifier(tableName, connection);
 
-        command.Connection = connection;
+        using OleDbCommand command = new()
+        {
+            Connection = connection,
+            CommandText = $"SELECT * FROM {quotedTableName}",
+        };
 
-        command.CommandText = $"SELECT * FROM {quotedTableName}";
-
-        dataAdapter.SelectCommand = command;
+        using OleDbDataAdapter dataAdapter = new()
+        {
+            SelectCommand = command,
+        };
 
         DataTable table = new()
         {
--- a/src/TestFramework/TestFramework.Extensions.Csv/CsvDataSourceAttribute.cs
+++ b/src/TestFramework/TestFramework.Extensions.Csv/CsvDataSourceAttribute.cs
@@ -1,0 +1,83 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System.Data;
+using System.Data.OleDb;
+using System.Globalization;
+using System.Reflection;
+
+namespace Microsoft.VisualStudio.TestTools.UnitTesting;
+
+/// <summary>
+/// Attribute to define dynamic data from a CSV file for a test method.
+/// </summary>
+[AttributeUsage(AttributeTargets.Method, AllowMultiple = true)]
+public sealed class CsvDataSourceAttribute : Attribute, ITestDataSource
+{
+    // Template used to map from a filename to a DB connection string
+    private const string CsvConnectionTemplate = "Provider=Microsoft.Jet.OLEDB.4.0;Data Source={0};Persist Security Info=False;Extended Properties=\"text;HDR=YES;FMT=Delimited\"";
+    private const string CsvConnectionTemplate64 = "Provider=Microsoft.Ace.OLEDB.12.0;Data Source={0};Persist Security Info=False;Extended Properties=\"text;HDR=YES;FMT=Delimited\"";
+
+    public CsvDataSourceAttribute(string fileName)
+        => FileName = fileName;
+
+    internal string FileName { get; }
+
+    IEnumerable<object?[]> ITestDataSource.GetData(MethodInfo methodInfo)
+    {
+        // We specifically use OleDb to read a CSV file...
+        // We better work with a full path, if nothing else, errors become easier to report
+        string fullPath = Path.GetFullPath(FileName);
+        if (!File.Exists(fullPath))
+        {
+            // TODO: Localize.
+            throw new FileNotFoundException($"Csv file '{fullPath}' cannot be found.", fullPath);
+        }
+
+        string tableName = Path.GetFileName(fullPath).Replace('.', '#');
+
+        // We can map simplified CSVs to an OLEDB/Text connection, then proceed as normal
+        using OleDbConnection connection = new();
+        using OleDbDataAdapter dataAdapter = new();
+        using OleDbCommandBuilder commandBuilder = new();
+        using OleDbCommand command = new();
+
+        // We have to use the name of the folder which contains the CSV file in the connection string
+        // If target platform is x64, then use CsvConnectionTemplate64 connection string.
+        connection.ConnectionString = IntPtr.Size == 8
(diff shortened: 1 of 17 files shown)


src/TestFramework/TestFramework.Extensions.Csv/CsvDataSourceAttribute.cs
```diff
@@ -0,0 +1,86 @@
...
+            throw new FileNotFoundException($"Csv file '{fullPath}' cannot be found.", fullPath);
+        }
+
+        string tableName = Path.GetFileName(fullPath).Replace('.', '#');
+
+        // We can map simplified CSVs to an OLEDB/Text connection, then proceed as normal
+        using OleDbConnection connection = new();
+        using OleDbDataAdapter dataAdapter = new();
+        using OleDbCommandBuilder commandBuilder = new();
+        using OleDbCommand command = new();
```
Evangelink: Nit: I usually prefer to initialize variable close to where they are used rather than having init blocks and usage blocks.

src/TestFramework/TestFramework.Extensions.Csv/CsvDataSourceAttribute.cs
```diff
@@ -0,0 +1,86 @@
...
+        object?[][] dataRows = new object?[table.Rows.Count][];
+        for (int i = 0; i < dataRows.Length; i++)
+        {
+            dataRows[i] = [table.Rows[i]];
+        }
+
+        return dataRows;
+    }
+
+    string? ITestDataSource.GetDisplayName(MethodInfo methodInfo, object?[]? data)
```
Evangelink, line 86: For the most part, you should be able to reuse the helper from `DynamicDataAttribute`.

src/TestFramework/TestFramework.Extensions.Csv/TestFramework.Extensions.Csv.csproj
```diff
@@ -0,0 +1,43 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFramework>netstandard2.0</TargetFramework>
+    <!-- TODO: (before merge) Is this hack relevant for a new project? -->
```
Evangelink: Good question, it's probably fine as this is (was?) mainly for some VS binding redirects.

src/TestFramework/TestFramework.Extensions.Csv/CsvDataSourceAttribute.cs
```diff
@@ -0,0 +1,86 @@
...
+using System.Data.OleDb;
+using System.Globalization;
+using System.Reflection;
+
+namespace Microsoft.VisualStudio.TestTools.UnitTesting;
+
+/// <summary>
+/// Attribute to define dynamic data from a CSV file for a test method.
+/// </summary>
+[AttributeUsage(AttributeTargets.Method, AllowMultiple = false/*TODO: Do we want to allow multiple*/)]
```
Evangelink: I think we allow multiple `DynamicData` so we should probably do the same.

Evangelink (commented):
For the name of the package, I was thinking about something like `MSTest.Data.Csv` or `MSTest.DataSource.Csv` or because we go for a functionally oriented name maybe the NuGet package could be tuned around the technology used so `MSTest.Data.Oledb` or `MSTest.DataSource.Oledb`. WDYT?

As for the name of the dll, my plan for v4 is to simplify naming so I think (if there is no technical constraint) we should do the same for these new packages.

src/TestFramework/TestFramework.Extensions.Csv/TestFramework.Extensions.Csv.csproj
Youssef1313: @Evangelink by fine, do you mean it can be removed for the new projects? or do you mean just keep as-is? I don't have much context on what issues is it trying to solve

src/TestFramework/TestFramework.Extensions.Csv/CsvDataSourceAttribute.cs
Youssef1313: Agreed. This was almost a copy/paste from the existing code. Fixed now

src/TestFramework/TestFramework.Extensions.Csv/CsvDataSourceAttribute.cs
Youssef1313: Agreed. Done

src/TestFramework/TestFramework.Extensions.Csv/CsvDataSourceAttribute.cs
```diff
@@ -0,0 +1,88 @@
...
+        if (!File.Exists(fullPath))
+        {
+            // TODO: Localize.
+            throw new FileNotFoundException($"Csv file '{fullPath}' cannot be found.", fullPath);
+        }
+
+        using OleDbConnection connection = new();
+
+        // We have to use the name of the folder which contains the CSV file in the connection string
+        // If target platform is x64, then use CsvConnectionTemplate64 connection string.
```
nohwnd, line 45: will this work also on arm64?

src/TestFramework/TestFramework.Extensions.Csv/CsvDataSourceAttribute.cs
```diff
@@ -0,0 +1,88 @@
...
+    private const string CsvConnectionTemplate64 = "Provider=Microsoft.Ace.OLEDB.12.0;Data Source={0};Persist Security Info=False;Extended Properties=\"text;HDR=YES;FMT=Delimited\"";
+
+    public CsvDataSourceAttribute(string fileName)
+        => FileName = fileName;
+
+    internal string FileName { get; }
+
+    IEnumerable<object?[]> ITestDataSource.GetData(MethodInfo methodInfo)
+    {
+        // TODO: Avoid using OleDb and instead parse Csv directly. Maybe use https://www.nuget.org/packages/CsvHelper? Write our own?
```
nohwnd, line 30: I would definitely avoid writing our own. 😁

test/IntegrationTests/MSTest.Acceptance.IntegrationTests/DataSourceTests.cs
```diff
@@ -52,6 +57,7 @@ public class DataSourceTests : AcceptanceTestBase
 
 
 #file MyTestClass.cs
+using System.Data;
```
nohwnd, line 60: Is this using needed?

Youssef1313:
> `MSTest.DataSource.Csv` sounds good. Maybe MSTest.DataSource.OleDb.Csv would be better? Or something that makes it less generic so we can replace it with other better solution in the future?

My view here is that if we wanted to replace OleDb with something else, it should be done in the same package, not as a new package. To me, the fact that we use OleDb is very implementation detail. Are you thinking of any scenarios where the user will still want to use OleDb with Csv if we provided an alternative?

Evangelink:
Linked to AB#2304879

Evangelink:
As described on the linked tickets, we have decided not to move forward with the feature request.

PR #4124, opened 2024-11-22
Implement `--config` for specifying testconfig.json file

### LINKED ISSUE #3562: Add support for --config <path-to-testconfig.json> file
## Summary

Add support for specifying a different configuration file.

## Background and Motivation

Currently the platform will auto-detect a `[appname].testconfig.json` file and load it. As part of a multi-test project, I probably want to share the same config file for all/most/multiple of my test projects. Having the ability to provide the common configuration file would be really nice.

## Proposed Feature

Add support for `--config <path>`.

## Alternative Designs

1. We could have each config file _testconfig.json_ specifying a path to a common file to load but this is cumbersome as I need to create the config file for each of my test projects.

2. We could use an auto-lookup mechanism, similar to request #3559. This seems interesting but raise the question of merging of configs.

AB#2307192

Youssef1313:
Fixes #3562

### DIFF
--- a/src/Platform/Microsoft.Testing.Platform/CommandLine/PlatformCommandLineProvider.cs
+++ b/src/Platform/Microsoft.Testing.Platform/CommandLine/PlatformCommandLineProvider.cs
@@ -31,6 +31,7 @@ internal sealed class PlatformCommandLineProvider : ICommandLineOptionsProvider
     public const string MinimumExpectedTestsOptionKey = "minimum-expected-tests";
     public const string TestHostControllerPIDOptionKey = "internal-testhostcontroller-pid";
     public const string ExitOnProcessExitOptionKey = "exit-on-process-exit";
+    public const string ConfigOptionKey = "config";
 
     public const string ServerOptionKey = "server";
     public const string ClientPortOptionKey = "client-port";
@@ -59,6 +60,7 @@ internal sealed class PlatformCommandLineProvider : ICommandLineOptionsProvider
         new(DiscoverTestsOptionKey, PlatformResources.PlatformCommandLineDiscoverTestsOptionDescription, ArgumentArity.Zero, false, isBuiltIn: true),
         new(IgnoreExitCodeOptionKey, PlatformResources.PlatformCommandLineIgnoreExitCodeOptionDescription, ArgumentArity.ExactlyOne, false, isBuiltIn: true),
         new(ExitOnProcessExitOptionKey, PlatformResources.PlatformCommandLineExitOnProcessExitOptionDescription, ArgumentArity.ExactlyOne, false, isBuiltIn: true),
+        new(ConfigOptionKey, PlatformResources.PlatformCommandLineConfigOptionDescription, ArgumentArity.ExactlyOne, false, isBuiltIn: true),
 
         // Hidden options
         new(HelpOptionQuestionMark, PlatformResources.PlatformCommandLineHelpOptionDescription, ArgumentArity.Zero, true, isBuiltIn: true),
--- a/src/Platform/Microsoft.Testing.Platform/Configurations/JsonConfigurationProvider.cs
+++ b/src/Platform/Microsoft.Testing.Platform/Configurations/JsonConfigurationProvider.cs
@@ -1,6 +1,7 @@
 ﻿// Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
+using Microsoft.Testing.Platform.CommandLine;
 using Microsoft.Testing.Platform.Helpers;
 using Microsoft.Testing.Platform.Logging;
 using Microsoft.Testing.Platform.Services;
@@ -9,10 +10,15 @@ namespace Microsoft.Testing.Platform.Configurations;
 
 internal sealed partial class JsonConfigurationSource
 {
-    internal sealed class JsonConfigurationProvider(ITestApplicationModuleInfo testApplicationModuleInfo, IFileSystem fileSystem, ILogger? logger) : IConfigurationProvider
+    internal sealed class JsonConfigurationProvider(
+        ITestApplicationModuleInfo testApplicationModuleInfo,
+        IFileSystem fileSystem,
+        CommandLineParseResult commandLineParseResult,
+        ILogger? logger) : IConfigurationProvider
     {
         private readonly ITestApplicationModuleInfo _testApplicationModuleInfo = testApplicationModuleInfo;
         private readonly IFileSystem _fileSystem = fileSystem;
+        private readonly CommandLineParseResult _commandLineParseResult = commandLineParseResult;
         private readonly ILogger? _logger = logger;
         private Dictionary<string, string?>? _propertyToAllChildren;
         private Dictionary<string, string?>? _singleValueData;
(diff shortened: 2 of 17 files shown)


Evangelink (commented):
This is looking good ❤️ :shipit: 

@MarcoRossignoli please also review

src/Platform/Microsoft.Testing.Platform/Configurations/JsonConfigurationProvider.cs
```diff
@@ -29,13 +35,26 @@ private async Task LogInformationAsync(string message)
...
+            string configFileName;
+            if (_commandLineParseResult.TryGetOptionArgumentList(PlatformCommandLineProvider.ConfigOptionKey, out string[]? configOptions))
             {
-                await LogInformationAsync($"Config file '{configFileName}' not found.");
-                return;
+                configFileName = Path.GetFullPath(configOptions[0]);
+                if (!_fileSystem.Exists(configFileName))
+                {
+                    // TODO: Localize
+                    throw new FileNotFoundException($"The configuration file '{configFileName}' specified with '--config' could not be found.", configFileName);
```
Evangelink: Fix this todo please

test/IntegrationTests/MSTest.Acceptance.IntegrationTests/ConfigurationSettingsTests.cs
```diff
@@ -57,6 +57,41 @@ public async Task TestConfigJson_WithoutRunSettings_BuildSuccess(string tfm)
...
+
+    public async Task TestWithConfigFromCommandLineWithNonExistingFile()
+    {
+        var testHost = TestHost.LocateFrom(_testAssetFixture.ProjectPath, TestAssetFixture.ProjectName, TargetFrameworks.NetCurrent.Arguments);
+        TestHostResult testHostResult = await testHost.ExecuteAsync("--config dummyconfigfile_not_existing_file.json", environmentVariables: new()
+        {
+            ["TestWithConfigFromCommandLine"] = "true",
+        });
+
+        testHostResult.AssertStandardErrorContains("FileNotFoundException");
```
Evangelink, line 92: If the name of the file is part of the output, can we capture it in the assert please.

test/IntegrationTests/Microsoft.Testing.Platform.Acceptance.IntegrationTests/HelpInfoTests.cs
```diff
@@ -27,6 +27,8 @@ .NET Testing Platform v*
 Usage {TestAssetFixture.NoExtensionAssetName}* [option providers] [extension option providers]
 Execute a .NET Test Application.
 Options:
+    --config
```
Evangelink: I mean wondering if we should have `--config-file` instead. @MarcoRossignoli @nohwnd any opinion?

MarcoRossignoli: --config-file is good so

src/Platform/Microsoft.Testing.Platform/Configurations/JsonConfigurationProvider.cs
```diff
@@ -29,13 +35,26 @@ private async Task LogInformationAsync(string message)
...
-            string configFileName = $"{Path.Combine(
-                Path.GetDirectoryName(_testApplicationModuleInfo.GetCurrentTestApplicationFullPath())!,
-                Path.GetFileNameWithoutExtension(_testApplicationModuleInfo.GetCurrentTestApplicationFullPath()))}{PlatformConfigurationConstants.PlatformConfigSuffixFileName}";
-            if (!_fileSystem.Exists(configFileName))
+            string configFileName;
+            if (_commandLineParseResult.TryGetOptionArgumentList(PlatformCommandLineProvider.ConfigOptionKey, out string[]? configOptions))
             {
-                await LogInformationAsync($"Config file '{configFileName}' not found.");
-                return;
+                configFileName = Path.GetFullPath(configOptions[0]);
```
MarcoRossignoli: wonder if we should avoid this api that could throw https://learn.microsoft.com/en-us/dotnet/api/system.io.path.getfullpath?view=net-8.0 and simply test for existance

MarcoRossignoli: Exists rule looks good to me

`The path parameter is permitted to specify relative or absolute path information. Relative path information is interpreted as relative to the current working directory. To obtain the current working directory, see GetCurrentDirectory.`

src/Platform/Microsoft.Testing.Platform/CommandLine/PlatformCommandLineProvider.cs
```diff
@@ -59,6 +60,7 @@ internal sealed class PlatformCommandLineProvider : ICommandLineOptionsProvider
         new(DiscoverTestsOptionKey, PlatformResources.PlatformCommandLineDiscoverTestsOptionDescription, ArgumentArity.Zero, false, isBuiltIn: true),
         new(IgnoreExitCodeOptionKey, PlatformResources.PlatformCommandLineIgnoreExitCodeOptionDescription, ArgumentArity.ExactlyOne, false, isBuiltIn: true),
         new(ExitOnProcessExitOptionKey, PlatformResources.PlatformCommandLineExitOnProcessExitOptionDescription, ArgumentArity.ExactlyOne, false, isBuiltIn: true),
+        new(ConfigOptionKey, PlatformResources.PlatformCommandLineConfigOptionDescription, ArgumentArity.ExactlyOne, false, isBuiltIn: true),
```
MarcoRossignoli: I would validate the existance/validity of the file inside the `ValidateOptionArgumentsAsync` so in case of we will have a meaninful message instead of an exception thrown below.

MarcoRossignoli (commented):
Generally good, I would add the validation and remove the usage of  the GetFullPath api.

src/Platform/Microsoft.Testing.Platform/Configurations/JsonConfigurationProvider.cs
Youssef1313: My intent here was to provide an exception with the full path when it's not found.
So that the user can clearly see a relative path in `--config-file` is relative to "what".

Are you okay to evaluate GetFullPath only when we are about to throw an exception, and possible catch any exception during GetFullPath and do nothing?

Somthing like this?

```csharp
if (_commandLineParseResult.TryGetOptionArgumentList(PlatformCommandLineProvider.ConfigOptionKey, out string[]? configOptions))
{
    configFileName = configOptions[0];
    if (!_fileSystem.Exists(configFileName))
    {
        try
        {
            // Get the full path for better error messages.
            // As this is only for the purpose of throwing an exception, ignore any exceptions during the GetFullPath call.
            configFileName = Path.GetFullPath(configFileName);
        }
        catch
        {
        }

        // TODO: Localize
        throw new FileNotFoundException($"The configuration file '{configFileName}' specified with '--config-file' could not be found.", configFileName);
    }
}
```

Evangelink: I think it sounds reasonable, let's get confirmation from @MarcoRossignoli.

MarcoRossignoli: Ok fine to me, anyway I don't expect we will reach this point if not for some bug, I'd like to have this check on the verify option value, where we define the option.

Youssef1313: I added it there as well.

Issue #4146, opened 2024-11-25
System.InvalidOperationException: Console is already in batching mode


thomhurst:
I'm getting this on my pipelines on GitHub actions sometimes:

```
  Error: Unhandled exception. System.InvalidOperationException: Console is already
  in batching mode.
     at 
  Microsoft.Testing.Platform.OutputDevice.Terminal.NonAnsiTerminal.StartUpdate() 
  in 
  /_/src/Platform/Microsoft.Testing.Platform/OutputDevice/Terminal/NonAnsiTerminal
  .cs:line 125
     at 
  Microsoft.Testing.Platform.OutputDevice.Terminal.TestProgressStateAwareTerminal.
  ThreadProc() in 
  /_/src/Platform/Microsoft.Testing.Platform/OutputDevice/Terminal/TestProgressSta
  teAwareTerminal.cs:line 45
     at System.Threading.Thread.StartCallback()
```

Looking at the code, every call to `StartUpdate` is within a `lock` statement so that all looks correct, so the only thing I can think of is we're hitting some exception (although I can't see anything in the output - Maybe I need to tweak my verbosity?) and therefore not calling `StopUpdate`?

It should probably be in a `finally` block to ensure it's always called.

Youssef1313:
@thomhurst Is that fixed by #4147?

thomhurst:
When's the next package release?

Youssef1313:
On NuGet.org, that would be a question for @Evangelink.

For a nightly/dogfood build, you can use this feed: `https://dnceng.pkgs.visualstudio.com/public/_packaging/test-tools/nuget/v3/index.json`

The version `3.7.0-preview.24575.7` includes your change.

nohwnd:
@Youssef1313  idk if you are aware of it, but here is a full guide to the preview releases: <https://aka.ms/mstest/preview>

PR #4147, opened 2024-11-25
StopUpdate in Finally Block

thomhurst:
#4146 

I'm getting this on my pipelines on GitHub actions sometimes:

```
  Error: Unhandled exception. System.InvalidOperationException: Console is already
  in batching mode.
     at 
  Microsoft.Testing.Platform.OutputDevice.Terminal.NonAnsiTerminal.StartUpdate() 
  in 
  /_/src/Platform/Microsoft.Testing.Platform/OutputDevice/Terminal/NonAnsiTerminal
  .cs:line 125
     at 
  Microsoft.Testing.Platform.OutputDevice.Terminal.TestProgressStateAwareTerminal.
  ThreadProc() in 
  /_/src/Platform/Microsoft.Testing.Platform/OutputDevice/Terminal/TestProgressSta
  teAwareTerminal.cs:line 45
     at System.Threading.Thread.StartCallback()
```

Looking at the code, every call to `StartUpdate` is within a `lock` statement so that all looks correct, so the only thing I can think of is we're hitting some exception (although I can't see anything in the output - Maybe I need to tweak my verbosity?) and therefore not calling `StopUpdate`?

It should probably be in a `finally` block to ensure it's always called.

### DIFF
--- a/src/Platform/Microsoft.Testing.Platform/OutputDevice/Terminal/NonAnsiTerminal.cs
+++ b/src/Platform/Microsoft.Testing.Platform/OutputDevice/Terminal/NonAnsiTerminal.cs
@@ -5,6 +5,7 @@ using System.Globalization;
 using System.Text;
 
 using Microsoft.Testing.Platform.Helpers;
+using Microsoft.Testing.Platform.Resources;
 
 namespace Microsoft.Testing.Platform.OutputDevice.Terminal;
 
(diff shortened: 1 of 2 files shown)


MarcoRossignoli:
@nohwnd can you take a look?

nohwnd (approved):

MarcoRossignoli (approved):

PR #4151, opened 2024-11-25
Set IsTestingPlatformApplication to true in ClassicEngine.targets

mariam-abdulla:
When we are using mstest sdk-style project, we have IsTestingPlatformApplication set to false, thus it's causing issues.
Because we are setting this property based on EnableMSTestRunner which is set to false because of the order of import (mstest testadapter props then ClassicEngine.targets), we need to set IsTestingPlatformApplication to true in ClassicEngine.targets to fix this issue.

The IsTestingPlatformApplication property is needed for the dotnet test driver to distinguish if we need to run the test project using the new testing platform.

### DIFF
--- a/src/Package/MSTest.Sdk/Sdk/Runner/ClassicEngine.targets
+++ b/src/Package/MSTest.Sdk/Sdk/Runner/ClassicEngine.targets
@@ -9,7 +9,10 @@
     <!-- When we are using mstest sdk-style project, we have IsTestingPlatformApplication set to false, thus it's causing issues.
     Because we are setting this property based on EnableMSTestRunner which is set to false,
     because of the order of import (mstest testadapter props then ClassicEngine.targets),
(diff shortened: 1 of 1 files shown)


src/Package/MSTest.Sdk/Sdk/Runner/ClassicEngine.targets
```diff
@@ -5,6 +5,12 @@
 
   <PropertyGroup Condition=" '$(IsTestApplication)' == 'true' ">
     <EnableMSTestRunner>true</EnableMSTestRunner>
+
+    <!-- When we are using mstest sdk-style project, we have IsTestingPlatformApplication set to false, thus it's causing issues.
```
MarcoRossignoli, line 9: add that we need the `IsTestingPlatformApplication` for the dotnet test driver

MarcoRossignoli (approved):
cc: @Evangelink

Evangelink (approved):

PR #4156, opened 2024-11-25
Further `System.Threading.Lock` usages

### LINKED ISSUE #4152: Polyfill Lock
NET9 brings more performant locking. Thom used it here, but it looks like this should use a polyfil instead. 

https://github.com/microsoft/testfx/blob/7d0cd29ae1fa85772fa08a58f50eb68ee472dcb0/src/Platform/Microsoft.Testing.Platform/OutputDevice/Terminal/TestProgressStateAwareTerminal.cs#L19-L23

The polyfill library we use from Simon already has Lock support it seems (there are new Lock related entries in Readme).

thomhurst:
Fix #4152


Youssef1313:
@dotnet-policy-service rerun

Youssef1313 (approved):
LGTM

CLA is stuck, let's see if marking as draft then ready for review again can fix re-trigger it. Otherwise I will close & re-open for a complete fresh run.

Evangelink (approved):

PR #4162, opened 2024-11-26
Implement codefix for MSTEST0037: Use proper 'Assert' methods

### LINKED ISSUE #3315: new Analyzer+CodeFix: reccomend against wrong assertion method (e.g. `Assert.IsTrue(x != null)`)

Youssef1313:
Fixes #3315

### DIFF
--- a/src/Analyzers/MSTest.Analyzers.CodeFixes/UseProperAssertMethodsFixer.cs
+++ b/src/Analyzers/MSTest.Analyzers.CodeFixes/UseProperAssertMethodsFixer.cs
@@ -72,7 +72,7 @@ public sealed class UseProperAssertMethodsFixer : CodeFixProvider
                 createChangedDocument = ct => FixAssertMethodForAddArgumentModeAsync(context.Document, diagnostic.AdditionalLocations[0], diagnostic.AdditionalLocations[1], diagnostic.AdditionalLocations[2], root, identifierNameSyntax, properAssertMethodName, ct);
                 break;
             case UseProperAssertMethodsAnalyzer.CodeFixModeRemoveArgument:
-                createChangedDocument = ct => FixAssertMethodForRemoveArgumentModeAsync(context.Document, diagnostic.AdditionalLocations[0], root, identifierNameSyntax, properAssertMethodName, ct);
+                createChangedDocument = ct => FixAssertMethodForRemoveArgumentModeAsync(context.Document, diagnostic.AdditionalLocations, root, identifierNameSyntax, properAssertMethodName, diagnostic.Properties.ContainsKey(UseProperAssertMethodsAnalyzer.NeedsNullableBooleanCastKey), ct);
                 break;
             default:
                 break;
@@ -151,13 +151,20 @@ public sealed class UseProperAssertMethodsFixer : CodeFixProvider
         return editor.GetChangedDocument();
     }
 
-    private static async Task<Document> FixAssertMethodForRemoveArgumentModeAsync(Document document, Location location1, SyntaxNode root, IdentifierNameSyntax identifierNameSyntax, string properAssertMethodName, CancellationToken cancellationToken)
+    private static async Task<Document> FixAssertMethodForRemoveArgumentModeAsync(
+        Document document,
+        IReadOnlyList<Location> additionalLocations,
+        SyntaxNode root,
+        IdentifierNameSyntax identifierNameSyntax,
+        string properAssertMethodName,
+        bool needsNullableBoolCast,
+        CancellationToken cancellationToken)
     {
         // This doesn't properly handle cases like Assert.AreEqual(message: "My message", expected: true, actual: x)
         // The proper handling of this may be Assert.IsTrue(message: "My message", condition: x)
         // Or: Assert.IsTrue(x, "My message")
         // For now this is not handled.
-        if (root.FindNode(location1.SourceSpan) is not ArgumentSyntax node1)
+        if (root.FindNode(additionalLocations[0].SourceSpan) is not ArgumentSyntax node1)
         {
             return document;
         }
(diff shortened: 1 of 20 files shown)


test/UnitTests/MSTest.Analyzers.UnitTests/UseProperAssertMethodsAnalyzerTests.cs
```diff
@@ -354,10 +582,28 @@ public void MyTestMethod()
             }
             """;
 
-        await VerifyCS.VerifyAnalyzerAsync(
+        // error CS1503: Argument 1: cannot convert from 'object' to 'bool'
```
Evangelink: Shall we instead not raise? We could detect if that can be converted to the other type. I guess

test/UnitTests/MSTest.Analyzers.UnitTests/UseProperAssertMethodsAnalyzerTests.cs
```diff
@@ -397,10 +646,28 @@ public void MyTestMethod()
...
```
Evangelink: Same remark about avoiding the analyzer diagnostic instead.

Evangelink (commented):
For the cases leading to compilation error, could we tweak the rule not to report instead?

test/UnitTests/MSTest.Analyzers.UnitTests/UseProperAssertMethodsAnalyzerTests.cs
Youssef1313: The diagnostic is still correct IMO. It's that the codefix should be producing a cast.

Youssef1313: I adjusted the codefix to produce a cast.

src/Analyzers/MSTest.Analyzers.CodeFixes/UseProperAssertMethodsFixer.cs
```diff
@@ -0,0 +1,209 @@
...
+            context.RegisterCodeFix(
+            CodeAction.Create(
+                title: string.Format(CultureInfo.InvariantCulture, CodeFixResources.UseProperAssertMethodsFix, properAssertMethodName),
+                createChangedDocument,
+                equivalenceKey: nameof(UseProperAssertMethodsFixer)),
+            diagnostic);
+        }
+    }
+
+    private static async Task<Document> FixAssertMethodForSimpleModeAsync(Document document, Location location1, Location location2, SyntaxNode root, SimpleNameSyntax simpleNameSyntax, string properAssertMethodName, CancellationToken cancellationToken)
```
Evangelink: Could you please replace `location1` and `location2` by more meaningful names?

src/Analyzers/MSTest.Analyzers.CodeFixes/UseProperAssertMethodsFixer.cs
```diff
@@ -0,0 +1,209 @@
...
+        }
+
+        DocumentEditor editor = await DocumentEditor.CreateAsync(document, cancellationToken).ConfigureAwait(false);
+        FixInvocationMethodName(editor, simpleNameSyntax, properAssertMethodName);
+        editor.ReplaceNode(node1, SyntaxFactory.Argument(node2).WithAdditionalAnnotations(Formatter.Annotation));
+
+        return editor.GetChangedDocument();
+    }
+
+    private static async Task<Document> FixAssertMethodForAddArgumentModeAsync(Document document, Location location1, Location location2, Location location3, SyntaxNode root, SimpleNameSyntax simpleNameSyntax, string properAssertMethodName, CancellationToken cancellationToken)
```
Evangelink: Same remark about location variables.

src/Analyzers/MSTest.Analyzers.CodeFixes/UseProperAssertMethodsFixer.cs
```diff
@@ -0,0 +1,209 @@
...
+        SimpleNameSyntax simpleNameSyntax,
+        string properAssertMethodName,
+        bool needsNullableBoolCast,
+        CancellationToken cancellationToken)
+    {
+        // This doesn't properly handle cases like Assert.AreEqual(message: "My message", expected: true, actual: x)
+        // The proper handling of this may be Assert.IsTrue(message: "My message", condition: x)
+        // Or: Assert.IsTrue(x, "My message")
+        // For now this is not handled.
+        if (root.FindNode(additionalLocations[0].SourceSpan) is not ArgumentSyntax node1)
```
Evangelink: Same remark for `node1`

Evangelink (commented):
LGTM - please rename the few variables with numbering suffix.

Evangelink (approved):

Issue #4159, opened 2024-11-26
System.Text.RegularExpressions.RegexMatchTimeoutException


thomhurst:
Seems to be affecting Macs again: https://github.com/thomhurst/TUnit/actions/runs/12029859325/job/33536030398

Happening since upgrading to 1.5.0-preview.24575.7
```
Error: Unhandled exception. System.Text.RegularExpressions.RegexMatchTimeoutException: The Regex engine has timed out while trying to match a pattern to an input string. This can occur for many reasons, including very large inputs or excessive backtracking caused by nested quantifiers, back-references and other factors.
   at System.Text.RegularExpressions.RegexRunner.<CheckTimeout>g__ThrowRegexTimeout|25_0()
   at System.Text.RegularExpressions.Generated.<RegexGenerator_g>F06D33C3F8C8C3FD257C1A1967E3A3BAC4BE9C8EC41CC9366C764C2205C68F0CE__GetFrameRegex_1.RunnerFactory.Runner.TryMatchAtCurrentPosition(ReadOnlySpan`1) in /_/artifacts/obj/Microsoft.Testing.Platform/Release/net8.0/System.Text.RegularExpressions.Generator/System.Text.RegularExpressions.Generator.RegexGenerator/RegexGenerator.g.cs:line 639
   at System.Text.RegularExpressions.Generated.<RegexGenerator_g>F06D33C3F8C8C3FD257C1A1967E3A3BAC4BE9C8EC41CC9366C764C2205C68F0CE__GetFrameRegex_1.RunnerFactory.Runner.Scan(ReadOnlySpan`1) in /_/artifacts/obj/Microsoft.Testing.Platform/Release/net8.0/System.Text.RegularExpressions.Generator/System.Text.RegularExpressions.Generator.RegexGenerator/RegexGenerator.g.cs:line 537
   at System.Text.RegularExpressions.Regex.ScanInternal(RegexRunnerMode, Boolean, String, Int32, RegexRunner, ReadOnlySpan`1, Boolean)
   at System.Text.RegularExpressions.Regex.RunSingleMatch(RegexRunnerMode, Int32, String, Int32, Int32, Int32)
   at System.Text.RegularExpressions.Regex.Match(String)
[...]

PR #4175, opened 2024-11-27
Show the text that timed out in regex match

nohwnd:
Help diagnose #4159

### DIFF
--- a/src/Platform/Microsoft.Testing.Platform/OutputDevice/Terminal/TerminalTestReporter.cs
+++ b/src/Platform/Microsoft.Testing.Platform/OutputDevice/Terminal/TerminalTestReporter.cs
@@ -647,7 +647,16 @@ internal sealed partial class TerminalTestReporter : IDisposable
     internal /* for testing */ static void AppendStackFrame(ITerminal terminal, string stackTraceLine)
     {
         terminal.Append(DoubleIndentation);
-        Match match = GetFrameRegex().Match(stackTraceLine);
+        Match match;
+        try
+        {
+            match = GetFrameRegex().Match(stackTraceLine);
+        }
+        catch (RegexMatchTimeoutException ex)
+        {
+            throw new RegexMatchTimeoutException($"Timeout matching '{stackTraceLine}'.", ex);
+        }
+
         if (match.Success)
         {
             bool weHaveFilePathAndCodeLine = !RoslynString.IsNullOrWhiteSpace(match.Groups["code"].Value);
--- a/src/Platform/Microsoft.Testing.Platform/OutputDevice/Terminal/TerminalTestReporter.cs
+++ b/src/Platform/Microsoft.Testing.Platform/OutputDevice/Terminal/TerminalTestReporter.cs
@@ -4,9 +4,13 @@
 #if !NET7_0_OR_GREATER
 using System.Diagnostics.CodeAnalysis;
 using System.Reflection;
+
 #endif
 
 using System.Globalization;
+#if !NETSTANDARD
+using System.Runtime.ExceptionServices;
+#endif
 using System.Text.RegularExpressions;
 
 using Microsoft.Testing.Platform.Helpers;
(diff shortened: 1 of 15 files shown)


Youssef1313 (dismissed):

Evangelink (commented):
Is it a temp change that will be removed before we ship? If not then the exception message should be localized.

nohwnd:
Thanks for pointing that out. localized so we can keep it for the future as well.

src/Platform/Microsoft.Testing.Platform/Resources/PlatformResources.resx
```diff
@@ -671,4 +671,8 @@ Takes one argument as string in the format &lt;value&gt;[h|m|s] where 'value' is
   <data name="ConfigurationFileNotFound" xml:space="preserve">
     <value>The configuration file '{0}' specified with '--config-file' could not be found.</value>
   </data>
-</root>
+  <data name="TimeoutInRegexStackTraceLineParsing" xml:space="preserve">
+    <value>Parsing stack trace line with regex timeout: '{0}'</value>
```
Youssef1313: Feels grammatically more correct this way?

```suggestion
    <value>Parsing stack trace line '{0}' with regex timed out.</value>
```

src/Platform/Microsoft.Testing.Platform/OutputDevice/Terminal/TerminalTestReporter.cs
```diff
@@ -647,7 +647,16 @@ private static void AppendAssemblyLinkTargetFrameworkAndArchitecture(ITerminal t
...
         terminal.Append(DoubleIndentation);
-        Match match = GetFrameRegex().Match(stackTraceLine);
+        Match match;
+        try
+        {
+            match = GetFrameRegex().Match(stackTraceLine);
+        }
+        catch (RegexMatchTimeoutException ex)
+        {
+            throw new RegexMatchTimeoutException(string.Format(CultureInfo.CurrentCulture, PlatformResources.TimeoutInRegexStackTraceLineParsing, stackTraceLine), ex);
```
Youssef1313: Hmm, we use InvariantCulture everywhere else I think?

nohwnd: we don't, this is correct imho

Youssef1313: Do we want to preserve the original stack trace?

Maybe with `ExceptionDispatchInfo.Capture(ex).Throw(new RegexMatchTimeoutException(...));`?

src/Platform/Microsoft.Testing.Platform/Resources/PlatformResources.resx
nohwnd: yes, but then the long text you are matching will be in the middle of the sentence, and you will never learn what the problem is until you read till the end. 

In the original way you see: <this is problem>: <this is additional info for that problem.>

Youssef1313: Maybe:

```suggestion
    <value>Parsing stack trace line with regex timed out. Input: '{0}'</value>
```

src/Platform/Microsoft.Testing.Platform/OutputDevice/Terminal/TerminalTestReporter.cs
nohwnd: It does not seem to be working like this, I can capture the exception and rethrow it, but not capture, change message and rethrow with the same stack trace. But yes it would be nice to keep more information.

Youssef1313: I'm not very sure tbh.

Searching across the codebase it seems we do both and are inconsistent there.

Looking into runtime https://grep.app/search?q=SR.Format%28CultureInfo, they mostly use InvariantCulture. But I see the point of "this is user facing so should be current culture".

That can be discussed separately though as we are already inconsistent.

Youssef1313: My bad, wrong API. Here is an example with the correct API to preserve the stack trace but update the message

```csharp
using System;
using System.Runtime.ExceptionServices;

try
{
    ThisWillBePreserved();
}
catch (Exception ex)
{
    var exNew = new Exception("Updated message");
    throw ex.StackTrace is null
        ? exNew
        : ExceptionDispatchInfo.SetRemoteStackTrace(exNew, ex.StackTrace);
}

void ThisWillBePreserved() => throw new System.Exception("Original message");
```

[SharpLab](https://sharplab.io/#v2:C4LgTgrgdgNAJiA1AHwAICYCMBYAUKgBgAJVMA6AJWmAEsBbAUzIFEAPAYwYAdaB7KAMoMwANxqcAzgG48eYGACeeAN54i6ogBUAFjQkB1GgBsjAIQYAFMAwnCRDOAAoAlDNwBfPOwCGwdtqJHNk4eGn4iBlZnFTUNEW8wCNYAOQYAdyIAXiIodKJg7j4oRwAiAFUuOF8HIkYJCW8AcwYS11j1YG0wXgzIsgFgb3YAa00wIYYiPRyIE3aNdQB+JNS0+YWQfI5CsKgAET0uX38ASSgAM15+hmAKBjpeYAYBodHxzkdI1Zgk/sGRsYTNoeWT4AAsWl0BmMZks1lsogcLiyAD4iJ1uhlchlSCxtqF+KUAPJgGiNGhQbxGWo2BrNVpSIA===)

image

nohwnd: ok added

PR #4178, opened 2024-11-27
Only push output device messages to Test Explorer, don't push logs

### LINKED ISSUE #4213: Output messages should be forward to clients instead of logs (server mode)
As of today, the logs are forwarded to the client and are displayed there. This is not right because the logs are meant to be troubleshooting oriented hence untranslated and possibly with many messages intended to help platform/framework/extension authors and not directly the end user.

On the other hand, the output devices messages are designed as user facing messaging and are currently not forwarded to be printed into the client (e.g. Test Explorer).

This is leading to multiple localization issues and missing some important/useful notifications for the end users.

Youssef1313:
Logs are intended to be used for diagnosing. They shouldn't be localized and they shouldn't be pushed to Test Explorer.

Fixes #4213

Linked to AB#2275339


test/IntegrationTests/MSTest.Acceptance.IntegrationTests/ServerModeTests.cs
```diff
@@ -40,7 +40,7 @@ public async Task DiscoverAndRun(string tfm)
         await Task.WhenAll(discoveryListener.WaitCompletion(), runListener.WaitCompletion());
         Assert.AreEqual(1, discoveryCollector.TestNodeUpdates.Count(x => x.Node.NodeType == "action"), $"Wrong number of discovery");
         Assert.AreEqual(2, runCollector.TestNodeUpdates.Count, $"Wrong number of updates");
-        Assert.IsFalse(logs.IsEmpty, $"Logs are empty");
+        Assert.IsTrue(logs.IsEmpty, $"Logs are not empty");
```
Youssef1313: Need to carefully double check if this is expected. And if yes, it's ideal to have another test that shows logs being not empty.

src/Platform/Microsoft.Testing.Platform/ServerMode/JsonRpc/ServerModePerCallOutputDevice.cs
```diff
@@ -19,11 +34,72 @@ internal class ServerModePerCallOutputDevice : IPlatformOutputDevice
 
     public Task DisplayAfterSessionEndRunAsync() => Task.CompletedTask;
 
-    public Task DisplayAsync(IOutputDeviceDataProducer producer, IOutputDeviceData data) => Task.CompletedTask;
+    public async Task DisplayAsync(IOutputDeviceDataProducer producer, IOutputDeviceData data)
+    {
+        using (await _asyncMonitor.LockAsync(TimeoutHelper.DefaultHangTimeSpanTimeout))
```
Evangelink: I don't think we need this lock, we could just forward down and the thing below is responsible for that.

@MarcoRossignoli shall we lock to ensure ordering below?

src/Platform/Microsoft.Testing.Platform/ServerMode/JsonRpc/ServerModePerCallOutputDevice.cs
```diff
@@ -19,11 +34,72 @@ internal class ServerModePerCallOutputDevice : IPlatformOutputDevice
...
+                    break;
+
+                case ExceptionOutputDeviceData exceptionOutputDeviceData:
+                    await LogAsync(LogLevel.Error, exceptionOutputDeviceData.Exception.ToString(), padding: null);
+                    break;
+            }
+        }
+    }
 
     public Task DisplayBannerAsync(string? bannerMessage) => Task.CompletedTask;
```
Evangelink: Let's handle this case and print as debug.

@MarcoRossignoli @drognanar it feels like it would be too verbose to print as Info for TE but maybe we still prefer doing it. WDYT?

src/Platform/Microsoft.Testing.Platform/ServerMode/JsonRpc/ServerModePerCallOutputDevice.cs
```diff
@@ -19,11 +34,72 @@ internal class ServerModePerCallOutputDevice : IPlatformOutputDevice
...
 
     public Task DisplayBeforeSessionStartAsync() => Task.CompletedTask;
```
Evangelink: Let's emit a debug message here and when in trace we should emit a localized message with the path to the log file to help users looking up the log.

Evangelink (commented):
Let's add a test where we have a fake adapter that emits the various messages and check we catch them with the right level.

src/Platform/Microsoft.Testing.Platform/Hosts/ServerTestHost.cs
```diff
@@ -450,13 +444,20 @@ private async Task<ResponseArgsBase> ExecuteRequestAsync(RequestArgsBase args, s
 
         DateTimeOffset adapterLoadStart = _clock.UtcNow;
 
+        IPlatformOutputDevice outputDevice = ServiceProvider.GetRequiredService<IPlatformOutputDevice>();
+        // TODO: What if some one wants to have their own output device that's also compatible with server mode in IDE?
```
MarcoRossignoli: I think here we need to do something better. 

Despite `IPlatformOutputDevice` is a bit special and the implementation needs to take into  account more scenario(like app restart and correct handling for the "output") I'd like to abstract it from the "modes". 

`IPlatformOutputDevice` implements `IOutputDevice` and today we pass to the `IOutputDevice` requestors the `IPlatformOutputDevice` implementation as-is. 

Clearly this is not more good if we need to route display info due to modes.

So we should when  `IOutputDevice` is requested from an extension to proxy it to be able to forward to the `IPlatformOutputDevice` but also to other internal part of the platform that needs it, for instance the server mode.

High level idea
```
 GetService<IOutputDevice>() -> OutputDeviceRouter -> ArbitraryPlatformOutputDeviceImplementation & ServerModeForwarder
```

In this way we don't need special or custom output device implementation for old and possible new modes and we can keep the "current" output device choice.

We cannot forward the `ArbitraryPlatformOutputDeviceImplementation` display..but I would expect that the info printed by the "device" are not important...the device output goal is to "show information" in "some way" it doesn't affect and should not affect in any way the execution, so information produced by this extension shouldn't be so useful.

cc: @Evangelink

src/Platform/Microsoft.Testing.Platform/ServerMode/JsonRpc/ProxyPlatformOutputDevice.cs
```diff
@@ -0,0 +1,136 @@
...
+using Microsoft.Testing.Platform.Helpers;
+using Microsoft.Testing.Platform.Hosts;
+using Microsoft.Testing.Platform.OutputDevice;
+using Microsoft.Testing.Platform.TestHost;
+
+namespace Microsoft.Testing.Platform.ServerMode;
+
+// Any interfaces that can have special treatment for any output device should be implemented by
+// this proxy class and be forwarded properly.
+// This is not so good. How can we make sure we are not missing any interfaces that may be implemented by external output devices?
```
Youssef1313: @Evangelink @MarcoRossignoli Can you think of a good way to solve this?

src/Platform/Microsoft.Testing.Platform/Hosts/ServerTestHost.cs
```diff
@@ -93,13 +96,38 @@ public ServerTestHost(
...
+
+#pragma warning disable VSTHRD100 // Avoid async void methods
+    private async void OnCurrentDomainUnhandledException(object sender, UnhandledExceptionEventArgs e)
+#pragma warning restore VSTHRD100 // Avoid async void methods
+    {
+        _logger.LogWarning($"[ServerTestHost.OnCurrentDomainUnhandledException] {e.ExceptionObject}{_environment.NewLine}IsTerminating: {e.IsTerminating}");
+
+        // Looks like nothing in this message to really be localized?
+        // All are class names, method names, property names, and placeholders. So none is localizable?
+        await ServiceProvider.GetOutputDevice().DisplayAsync(
```
MarcoRossignoli: I think that use the `GetAwaiter().GetResult()` is better here. I don't expect a log of call here and awaiting a void we could lose the message.

src/Platform/Microsoft.Testing.Platform/OutputDevice/OutputDeviceManager.cs
```diff
@@ -15,15 +16,15 @@ public void SetPlatformOutputDevice(Func<IServiceProvider, IPlatformOutputDevice
         _platformOutputDeviceFactory = platformOutputDeviceFactory;
     }
 
-    public IPlatformOutputDevice Build(ServiceProvider serviceProvider)
+    internal ProxyPlatformOutputDevice Build(ServiceProvider serviceProvider, bool useServerModeOutputDevice)
```
MarcoRossignoli: why internal ctor?

src/Platform/Microsoft.Testing.Platform/ServerMode/JsonRpc/ProxyPlatformOutputDevice.cs
MarcoRossignoli: Why do we need it?

The proxy should only implement `IOutputDevice` and returned when users asks to the service provider. Here after the proxy should have today at max 2 "receivers", the real `IPlatformOutputDevice` and the servicehost logger(you can directly add a method to that instead to implement a custom one).

src/Platform/Microsoft.Testing.Platform/Hosts/ServerTestHost.cs
```diff
@@ -93,13 +96,36 @@ public ServerTestHost(
...
+    {
+        _logger.LogWarning($"[ServerTestHost.OnCurrentDomainUnhandledException] {e.ExceptionObject}{_environment.NewLine}IsTerminating: {e.IsTerminating}");
+
+        // Looks like nothing in this message to really be localized?
+        // All are class names, method names, property names, and placeholders. So none is localizable?
+        ServiceProvider.GetOutputDevice().DisplayAsync(
+            this,
+            new WarningMessageOutputDeviceData(
+                $"[ServerTestHost.OnCurrentDomainUnhandledException] {e.ExceptionObject}{_environment.NewLine}IsTerminating: {e.IsTerminating}"))
+            .GetAwaiter().GetResult();
```
Evangelink, line 118: @MarcoRossignoli is this better to do this or use await on async void? I think I recall the await async void was the "recommended" way for eventing methods in WPF but I might say something wrong.

src/Platform/Microsoft.Testing.Platform/Hosts/ServerTestHost.cs
```diff
@@ -269,6 +288,12 @@ private async Task HandleNotificationAsync(NotificationMessage message, Cancella
                         if (cancellationException is null)
                         {
                             await _logger.LogWarningAsync($"Exception during the cancellation of request id '{args.CancelRequestId}'");
+
+                            // TODO: Localize
```
Evangelink: TODO to fix

src/Platform/Microsoft.Testing.Platform/Hosts/ServerTestHost.cs
Youssef1313: @Evangelink I initially had it `async void` as I'm coming from the background that in general async void is okay for events in UI frameworks (e.g, WPF). Marco suggested using `GetAwaiter().GetResult()` instead.

src/Platform/Microsoft.Testing.Platform/OutputDevice/ProxyOutputDevice.cs
```diff
@@ -0,0 +1,68 @@
...
+
+using Microsoft.Testing.Platform.Extensions.OutputDevice;
+using Microsoft.Testing.Platform.Hosts;
+using Microsoft.Testing.Platform.ServerMode;
+
+namespace Microsoft.Testing.Platform.OutputDevice;
+
+// Any interfaces that can have special treatment for any output device should be implemented by
+// this proxy class and be forwarded properly.
+// This is not so good. How can we make sure we are not missing any interfaces that may be implemented by external output devices?
```
Evangelink: Comment seems to be outdated and could be dropped.

Youssef1313: It is. Thanks!

src/Platform/Microsoft.Testing.Platform/OutputDevice/TerminalOutputDevice.cs
```diff
@@ -25,6 +25,7 @@ namespace Microsoft.Testing.Platform.OutputDevice;
 /// <summary>
 /// Implementation of output device that writes to terminal with progress and optionally with ANSI.
 /// </summary>
+// NOTE: !!! Any interfaces here will likely need to be implemented by ProxyPlatformOutputDevice as well. !!!
```
Evangelink: Outdated?

src/Platform/Microsoft.Testing.Platform/ServerMode/JsonRpc/ServerModePerCallOutputDevice.cs
```diff
@@ -1,14 +1,44 @@
...
+    private readonly ConcurrentBag<ServerLogMessage> _messages = new();
+
+    private IServerTestHost? _serverTestHost;
+
+    private static readonly string[] NewLineStrings = { "\r\n", "\n" };
+
+    public ServerModePerCallOutputDevice(IServiceProvider serviceProvider)
+        => _serviceProvider = serviceProvider;
+
+    internal async Task InitializeAsync(IServerTestHost serverTestHost)
```
Evangelink, line 30: Could you add a small comment for future readers as to why we do that

test/IntegrationTests/Microsoft.Testing.Platform.Acceptance.IntegrationTests/Microsoft.Testing.Platform.Acceptance.IntegrationTests.csproj
```diff
@@ -1,17 +1,21 @@
-<Project Sdk="Microsoft.NET.Sdk">
+﻿<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
     <TargetFrameworks>$(NetCurrent)</TargetFrameworks>
     <TestRunnerAdditionalArguments>$(TestRunnerAdditionalArguments) --retry-failed-tests 3</TestRunnerAdditionalArguments>
     <UseVSTestRunner>false</UseVSTestRunner>
     <DefineConstants Condition=" '$(FastAcceptanceTest)' == 'true'">$(DefineConstants);SKIP_INTERMEDIATE_TARGET_FRAMEWORKS</DefineConstants>
+    <DefineConstants>$(DefineConstants);IS_MTP</DefineConstants>
```
Evangelink: Unused?

test/IntegrationTests/Microsoft.Testing.Platform.Acceptance.IntegrationTests/Microsoft.Testing.Platform.Acceptance.IntegrationTests.csproj
```diff
@@ -1,17 +1,21 @@
...
     <OutputType>Exe</OutputType>
     <UseAppHost>true</UseAppHost>
   </PropertyGroup>
 
   <ItemGroup>
     <PackageReference Include="MSBuild.StructuredLogger" />
+    <PackageReference Include="StreamJsonRpc" />
     <PackageReference Include="Microsoft.Testing.Extensions.Retry" />
+
+    <Compile Include="$(RepoRoot)test\IntegrationTests\MSTest.Acceptance.IntegrationTests\ServerMode\**\*.cs" Link="%(RecursiveDir)%(FileName)%(Extension)" />
```
Evangelink: Nit: I would reverse the order (e.g. move file here and reference from MSTest) as platform is more "core" than MSTest but I am fine if we keep as-is.

src/Platform/Microsoft.Testing.Platform/OutputDevice/TerminalOutputDevice.cs
Youssef1313: Yup. Will drop it

Evangelink (approved):
Some nits and questions but LGTM.

Thanks for the massive work here @Youssef1313! And thanks @MarcoRossignoli for technical consulting <3

test/IntegrationTests/Microsoft.Testing.Platform.Acceptance.IntegrationTests/Microsoft.Testing.Platform.Acceptance.IntegrationTests.csproj
Youssef1313: Let's do it as follow-up to not pollute the diff with too many file moves.

src/Platform/Microsoft.Testing.Platform/Hosts/ServerTestHost.cs
Evangelink: Ok so let's go with this suggestion. Thanks!

test/IntegrationTests/Microsoft.Testing.Platform.Acceptance.IntegrationTests/Microsoft.Testing.Platform.Acceptance.IntegrationTests.csproj
Youssef1313: Tracking in https://github.com/microsoft/testfx/issues/4231

test/IntegrationTests/Microsoft.Testing.Platform.Acceptance.IntegrationTests/Microsoft.Testing.Platform.Acceptance.IntegrationTests.csproj
Youssef1313: Removed

src/Platform/Microsoft.Testing.Platform/ServerMode/JsonRpc/ServerModePerCallOutputDevice.cs
Youssef1313: Added a comment. Does it look good?

src/Platform/Microsoft.Testing.Platform/OutputDevice/TerminalOutputDevice.cs
Youssef1313: Fixed

src/Platform/Microsoft.Testing.Platform/OutputDevice/ProxyOutputDevice.cs
Youssef1313: Fixed

src/Platform/Microsoft.Testing.Platform/Hosts/ServerTestHost.cs
Youssef1313: Done

Evangelink (approved):

PR #4183, opened 2024-11-27
Mark IgnoreAttribute as not inherited

Youssef1313:
It makes better sense to reflect the real semantics of the attribute.

@Evangelink Any backcompat concerns with this change?


Evangelink (approved):
I think it's safe.

PR #4185, opened 2024-11-27
Fix local warning (and fix typo)

Youssef1313:
This seems to be producing a warning in IDE but not in command-line builds?

image

Fixed a small typo along the way. Sorry for having both fixes in one PR. @Evangelink Let me know if you prefer to split them.


Evangelink (approved):
It's fine, thanks for calling it out. LGTM!

Issue #4199, opened 2024-11-29
How to build locally?


thomhurst:
I'd like to build locally but just get compile errors:

```
C:\Users\Tom\.nuget\packages\msbuild.sdk.extras\3.0.44\Build\LanguageTargets\CheckMissing.targets(44,5): error : The sp
ecified language targets for uap10.0.16299 is missing. Ensure correct tooling is installed for 'uap'. Missing: 'C:\Prog
ram Files\Microsoft Visual Studio\2022\Community\MSBuild\Microsoft\WindowsXaml\v17.0\Microsoft.Windows.UI.Xaml.CSharp.t
argets' [C:\git\testfx\src\TestFramework\TestFramework.Extensions\TestFramework.Extensions.csproj::TargetFramework=uap1
0.0.16299]
```

```
C:\Users\Tom\.nuget\packages\msbuild.sdk.extras\3.0.44\Build\LanguageTargets\CheckMissing.targets(44,5): error : The sp
ecified language targets for uap10.0.16299 is missing. Ensure correct tooling is installed for 'uap'. Missing: 'C:\Prog
ram Files\Microsoft Visual Studio\2022\Community\MSBuild\Microsoft\WindowsXaml\v17.0\Microsoft.Windows.UI.Xaml.CSharp.t
argets' [C:\git\testfx\src\Adapter\MSTest.TestAdapter\MSTest.TestAdapter.csproj::TargetFramework=uap10.0.16299]
```

```
C:\Users\Tom\.nuget\packages\msbuild.sdk.extras\3.0.44\Build\LanguageTargets\CheckMissing.targets(44,5): error : The sp
ecified language targets for uap10.0.16299 is missing. Ensure correct tooling is installed for 'uap'. Missing: 'C:\Prog
ram Files\Microsoft Visual Studio\2022\Community\MSBuild\Microsoft\WindowsXaml\v17.0\Microsoft.Windows.UI.Xaml.CSharp.t
argets' [C:\git\testfx\src\Adapter\MSTestAdapter.PlatformServices\MSTestAdapter.PlatformServices.csproj::TargetFramewor
k=uap10.0.16299]
```

Evangelink:
Hi @thomhurst,

It's probably easier for you not to install UWP on your machine (required to build MSTest) so I'd recommend you to use ` .\Build.cmd /p:ProductsToBuild=testing-platform`

thomhurst:
Wonder if it's worth splitting mstest and testing platform into different solutions?

[...]

PR #4200, opened 2024-11-29
Extensible Filters + AggregateFilter

### LINKED ISSUE #3530: [Testing Platform] Aggregate Filter
I was actually thinking about mixes of filters earlier and whether that was worth supporting. You'd need a new type like `AggregateFilter` (like `AggregateException`) that wraps around multiple. Might be worth splitting that into its own issue if that sounds like a good idea to you?

_Originally posted by @thomhurst in https://github.com/microsoft/testfx/issues/3528#issuecomment-2285603508_
            

As mentioned here, it would be good to be able to mix and match filters. This would need a new filter type that can wrap around a collection of filters.

### LINKED ISSUE #3590: [Testing Platform] Open up filters for custom implementations
as mentioned in other threads, you are planning to allow test framework creators to be able to write their own filter implementations. This is currently locked down behind internal code.

This is just to track this.

thomhurst:
WIP.

I'm struggling to build this locally, have raised #4199 

Fixes #3530 
Fixes #3590


thomhurst:
@Evangelink Hope you don't mind me having a go at tackling a couple of issues that have been sitting around for months?

`Microsoft.Testing.Extensions.Retry` isn't open-source/in this repo? Got failing tests because I've changed code, but that's in a package it seems so I can't make changes to it?

thomhurst:
Same happening with integration tests

> Unhandled exception. System.TypeLoadException: Could not load type 'Microsoft.Testing.Platform.Requests.ITestExecutionFilterFactory' from assembly 'Microsoft.Testing.Platform, Version=42.42.42.42,

Why does it reference packages instead of the local project's code?

thomhurst:
@Evangelink @nohwnd @MarcoRossignoli Providing you were okay with implementing this, how would I go about being unblocked due to private/internal code that I can't edit?

Evangelink:
@thomhurst I wasn't able to play a little with the PR but although I like the concept of the aggregate filter, I don't think that the extensibility as you are doing it is the right way forward as it would allow users to register any kind of filter including filters that are not handled by the framework. As of now, the filters have to be specifically supported by the platform or the framework.

thomhurst:
Hmm that's a good point. The registration code is going to have to be surfaced publicly though in order for frameworks to register it.

Might be that filters work in conjunction with capabilities? So filters are only considered active/available if the framework has a capability for it? Kinda like the trx extension. That way we know a framework can handle that filter.

thomhurst:
Another alternative is the filter knows how to perform the logic itself, so that users could actually add filters on the fly without specific support from the framework.

E.g.

```
public class MyFilter
{
    ...
    bool CanRun(TestNode testNode) => testNode.Id == "1"; 
}
````

**Comment by thomhurst** [comment]
@Evangelink did you have any opinions? I think the filter defining its own logic means users could make their own filters. Way more extensible then

**Comment by thomhurst** [comment]
@Evangelink @nohwnd @MarcoRossignoli @Youssef1313 

Here's my latest idea on filtering. Each implementation implements its own `MatchesFilter(TestNode)` method which defines its own logic on how it should filter.

This means all frameworks need to do is call: `ITestExecutionFilter.MatchesFilter(testNode)` - They don't need to do any casting, or type checking, or implementing any logic, because it's defined by the filter itself.

Of course, they still could override the behaviour by doing what they do currently - Casting to the concrete type and implementing the filtering themselves, but I don't envision a lot of scenarios where you'd want to change the logic.

This also makes filtering more consistent between frameworks as they can all use centralised logic instead of recreating it themselves, with the risk of introducing bugs.

This also means users could create their own filters and register them, and they'd automatically flow through as the ITestExecutionFilter, or inside the AggregateFilter, and the filtering will work with their newly introduced logic without the framework even having to know about the specific type/logic.

Keen to know your thoughts. I think this works nicely with the Open/Closed principle - Closed for modification but highly open to extensibility 😄

**Comment by thomhurst** [comment]
Bump 😇

**Comment by Evangelink** [comment]
Hey @thomhurst !

Sorry for the delay, the colleagues are in holidays and I was trying to make sure the new version gets released before I also leave but we had many infra issues.

I'll try to have a look asap and write my feedback

**Comment by thomhurst** [comment]
Did you ever manage to take a look @Evangelink ?

**Comment by Evangelink** [comment]
Not yet sadly. I am finishing up fixes for 3.7.1 and will be on it. 

Let me auto-assign the PR.

**Comment by Evangelink** [comment]
@thomhurst I am really sorry but I am not able to get some focus on this for the time being.

**Comment by thomhurst** [comment]
Anyone going to have any capacity at any point? I'd really like to write my own batching filter.

**Comment by Youssef1313** [comment]
Thanks for the ping @thomhurst 

It looks like there are merge conflicts here now, if you can fix them please.

**Comment by thomhurst** [comment]
> Thanks for the ping @thomhurst
> 
> It looks like there are merge conflicts here now, if you can fix them please.

Pushed a newer version. It's not tidied up completely (public API analyzers etc.) but I thought I'd get feedback before progressing any further.

**Comment by thomhurst** [comment]
Had another conflict again - Just resolved

File: src/Platform/Microsoft.Testing.Platform/Extensions/IReceivesAllTestNodesExtension.cs [hunk 1]
```diff
@@ -0,0 +1,20 @@
...
+/// An example is a Batch filter that needs to compare all of the test nodes to know whether to filter or not.
+/// Or an extension that might add some useful Test Node properties to all tests.
+/// </summary>
+public interface IReceivesAllTestNodesExtension
+{
+    /// <summary>
+    /// Receives a list of test nodes.
+    /// </summary>
+    /// <param name="testNodes">A read-only list of test nodes.</param>
+    void ReceiveTestNodes(IReadOnlyList<TestNode> testNodes);
```
Youssef1313: Looks like this is not used currently.

thomhurst: Not by any of the out of the box filters (apart from aggregate filter so it can pass them to inner filters) - but filter authors can place implement this interface on their custom filters etc. to ask the framework that it needs to see all the test nodes for use in its logic

src/Platform/Microsoft.Testing.Platform/Requests/ConsoleTestExecutionFilterFactory.cs
```diff
@@ -1,28 +0,0 @@
...
-
-internal sealed class ConsoleTestExecutionFilterFactory(ICommandLineOptions commandLineService) : ITestExecutionFilterFactory
-{
-    private readonly ICommandLineOptions _commandLineService = commandLineService;
-
-    public string Uid => nameof(ConsoleTestExecutionFilterFactory);
-
-    public string Version => AppVersion.DefaultSemVer;
-
-    public string DisplayName => PlatformResources.ConsoleTestExecutionFilterFactoryDisplayName;
```
Youssef1313, line 18: This resource is no longer used. Could you delete it as well?

src/Platform/Microsoft.Testing.Platform/Requests/ConsoleTestExecutionFilterFactory.cs
```diff
@@ -1,28 +0,0 @@
...
-
-    public string Description => PlatformResources.ConsoleTestExecutionFilterFactoryDescription;
```
Youssef1313, line 20: This resource is no longer used. Could you delete it as well?

src/Platform/Microsoft.Testing.Platform/Requests/ConsoleTestExecutionRequestFactory.cs
```diff
@@ -3,30 +3,20 @@
...
-    private readonly ICommandLineOptions _commandLineService = commandLineService;
-    private readonly ITestExecutionFilterFactory _testExecutionFilterFactory = testExecutionFilterFactory;
-
-    public async Task<TestExecutionRequest> CreateRequestAsync(TestSessionContext session)
+    public Task<TestExecutionRequest> CreateRequestAsync(TestSessionContext session)
     {
-        (bool created, ITestExecutionFilter? testExecutionFilter) = await _testExecutionFilterFactory.TryCreateAsync();
-        if (!created)
-        {
-            throw new InvalidOperationException(PlatformResources.CannotCreateTestExecutionFilterErrorMessage);
```
Youssef1313, line 21: This resource is no longer used. Could you delete it as well?

src/Platform/Microsoft.Testing.Platform/Requests/ITestExecutionFilter.cs
```diff
@@ -1,9 +1,24 @@
...
 /// <summary>
 /// Represents a filter for test execution.
 /// </summary>
-public interface ITestExecutionFilter;
+public interface ITestExecutionFilter
+{
+    /// <summary>
+    /// Gets a value indicating whether the filter should be used for filtering.
+    /// </summary>
+    bool IsEnabled { get; }
```
Youssef1313: I'm not sure whether or not we want to expose `IsEnabled`.

One feeling I have is that implementations of the interface could decide to return `true` from `MatchesFilter` whenever they want to "disable" the filter. But I don't have a strong opinion.

One thing we need to think about is if we want `IsEnabled` to be a property or a Task-returning method, in case implementations could need to be async. (I'll discuss with the team)

src/Platform/Microsoft.Testing.Platform/Requests/ITestExecutionFilter.cs
```diff
@@ -1,9 +1,24 @@
...
+
+    /// <summary>
+    /// Determines whether the specified test node matches the filter criteria.
+    /// </summary>
+    /// <param name="testNode">The test node to check against the filter.</param>
+    /// <returns><c>true</c> if the test node matches the filter; otherwise, <c>false</c>.</returns>
+    bool MatchesFilter(TestNode testNode);
```
Youssef1313: Same here, I'm not sure if this is better to be Task-returning or not. (I'll discuss with the team)

src/Platform/Microsoft.Testing.Platform/TestHost/TestHostManager.cs
```diff
@@ -57,35 +58,37 @@ internal async Task<ActionResult<ITestFrameworkInvoker>> TryBuildTestAdapterInvo
...
+    public async Task<ITestExecutionFilter> BuildFilterAsync(
+        IServiceProvider serviceProvider,
+        ICollection<TestNode>? testNodes)
     {
-        Guard.NotNull(testExecutionFilterFactory);
-        if (_testExecutionFilterFactory is not null)
+        if (testNodes?.Count > 0)
         {
-            throw new InvalidOperationException(PlatformResources.TEstExecutionFilterFactoryFactoryAlreadySetErrorMessage);
+            return new TestNodeUidListFilter(testNodes.Select(x => x.Uid).ToArray());
```
Youssef1313: Is the intent here to run whatever was selected in IDE, even if there is some other filter that would filter it out?

thomhurst: Yeah exactly. `TestNodeUidListFilter` seems to be special and work a bit differently to other filters.

Alternatively I could add it to the built filters list and combine it with others in an aggregate filter (if any)?

But not sure that's even supported at the moment if its only IDEs/Test Explorer driving this.

But if it is going to be extended to the CLI in the future then it might make sense.

src/Platform/Microsoft.Testing.Platform/Requests/ITestExecutionFilter.cs
thomhurst: Maybe ValueTask to avoid allocations in sync scenarios but gives the flexibility of async?

Youssef1313: For sync scenarios (common scenario), I think `Task.FromResult(...)` will be used which uses cached tasks internally.

src/Platform/Microsoft.Testing.Platform/TestHost/TestHostManager.cs
Youssef1313: I tend to think `TestHostManager` could have two `BuildFilterAsync` overloads. One that only takes `ICollection<TestNode>? testNodes` which is only used by `ServerTestHost`, and the signature kinda implies that it only considers the given test nodes, and another overload that only takes `IServiceProvider serviceProvider` which is used by ConsoleTestHost.

So in future, if we decided to add `TestNodeUidListFilter` for CLI scenarios, we go through the proper code path that combines all filters properly.

src/Platform/Microsoft.Testing.Platform/Extensions/IReceivesAllTestNodesExtension.cs
Youssef1313: @thomhurst filter authors can implement it, but who is responsible of calling it?

thomhurst: My thought was the test framework - Because they're responsible for building the test nodes.

Alternatively, if we wanted it to happen automatically, maybe there could be a consumer of discovered test node messages on the bus, that could aggregate all test nodes and pass them on?

thomhurst: Can a data consumer access extensions and filters though?

src/Platform/Microsoft.Testing.Platform/Requests/ConsoleTestExecutionRequestFactory.cs
thomhurst: Deleted

src/Platform/Microsoft.Testing.Platform/Requests/ConsoleTestExecutionFilterFactory.cs
thomhurst: Deleted

src/Platform/Microsoft.Testing.Platform/Requests/ConsoleTestExecutionFilterFactory.cs
thomhurst: Deleted

src/Platform/Microsoft.Testing.Platform/TestHost/TestHostManager.cs
thomhurst: Added a separate method for the test node path

thomhurst:
Aside from tweaking the signatures of things, do you think this PR makes sense and is a good idea?

Youssef1313:
Yes @thomhurst, I think this is a very good improvement.

I'd like to see some tests that use and demonstrate `IReceivesAllTestNodesExtension` though, and there will be few things to discuss with the team to agree on the final API shape we want.

And very sorry for the delays in reviewing this.

thomhurst:
Added a test for `AggregateFilter` calling inner `IReceivesAllTestNodesExtension` when it's called itself.

But apart from that there's no logic to test - It's just an interface that serves the purpose of passing data via the parameter. The logic is up to whoever is implementing it.

test/UnitTests/Microsoft.Testing.Platform.UnitTests/Filters/FiltersTests.cs
```diff
@@ -0,0 +1,254 @@
...
+
+        Assert.IsTrue(testExecutionFilter is AggregateFilter);
+
+        ((AggregateFilter)testExecutionFilter).ReceiveTestNodes([]);
+
+        Assert.IsTrue(((AggregateFilter)testExecutionFilter).InnerFilters.Cast<ReceivableTestNodesFilter>().All(x => x.ReceivedTestNodes));
+    }
+
+    [TestMethod]
+    public async Task Batch_ReceivesAllTestNodes_Filter_Example()
```
thomhurst: @Youssef1313 Here's an example of how I'd like to to use `IReceivesAllTestNodesExtension` within a filter.

Youssef1313: I assume in real-world, you will do something like `if ([run|discover]TestExecutionRequest.Filter is IReceivesAllTestNodesExtension receivesTestNodeFilter) receivesTestNodeFilter.ReceiveTestNodes()`?

Isn't that achievable if you define the interface on your side, and have the special handling for `AggregateFilter`? Something like:

```csharp
if (request.Filter is AggregateFilter aggregateFilter)
{
    foreach (var inner in aggregateFilter.InnerFilters.OfType<IReceivesAllTestNodes>())
    {
        inner.ReceiveTestNodes();
    }
}

(request.Filter as IReceivesAllTestNodes)?.ReceiveTestNodes();
```

thomhurst: Yeah you're right actually. I can just define this my side. I'll remove it from this PR!

src/Platform/Microsoft.Testing.Extensions.VSTestBridge/Requests/VSTestTestExecutionFilter.cs
```diff
@@ -24,4 +25,10 @@ private VSTestTestExecutionFilter()
     public ImmutableArray<TestCase>? TestCases => null;
 
     internal static VSTestTestExecutionFilter Instance { get; } = new();
+
+    /// <inheritdoc />
+    public bool IsEnabled => false;
```
drognanar: why is this filter disabled?

src/Platform/Microsoft.Testing.Platform/Requests/AggregateFilter.cs
```diff
@@ -0,0 +1,24 @@
...
+/// <param name="innerFilters">The list of inner filters to combine.</param>
+public sealed class AggregateFilter(params IReadOnlyList<ITestExecutionFilter> innerFilters) : ITestExecutionFilter
+{
+    /// <summary>
+    /// Gets the list of inner filters.
+    /// </summary>
+    public IReadOnlyList<ITestExecutionFilter> InnerFilters { get; } = innerFilters;
+
+    /// <inheritdoc />
+    public bool IsEnabled => true;
```
drognanar: we should check whether any of the subfilters are enabled

src/Platform/Microsoft.Testing.Platform/TestHost/TestHostManager.cs
```diff
@@ -57,35 +62,33 @@ internal async Task<ActionResult<ITestFrameworkInvoker>> TryBuildTestAdapterInvo
...
-            await testExecutionFilterFactory.TryInitializeAsync();
+            await testExecutionFilter.TryInitializeAsync();
 
-            return ActionResult.Ok(testExecutionFilterFactory);
+            list.Add(testExecutionFilter);
         }
 
-        return ActionResult.Fail<ITestExecutionFilterFactory>();
+        ITestExecutionFilter[] requestedFilters = list
+            .Where(x => x.IsEnabled)
```
drognanar: I think we should pass all filters rather than filter them out. this allows us to keep the async nature of the factories and build them at the time that the request is invoked.

especially in case of the server mode, where the request is invoked once JSON-RPC method is invoked and that might not happen immediately. and might impact the filter generated.

for instance, this affects how the RetryFilter operates. with a delayed, async factory, the RetryFilterFactory can check with the retry orchestrator if there's tests to be retried or not and disable the filter on construction. once built, the filter should be immediate, i.e. not contain any async code.

with the non-async approach, either each filter match needs to become async, or all of the filter creation needs to be done at process startup, which the RetryFilter does not have the luxury of doing.

src/Platform/Microsoft.Testing.Platform/PublicAPI/PublicAPI.Unshipped.txt
```diff
@@ -9,11 +9,22 @@ Microsoft.Testing.Platform.Extensions.Messages.FileArtifactProperty.FileArtifact
...
 Microsoft.Testing.Platform.Extensions.Messages.FileArtifactProperty.FileInfo.get -> System.IO.FileInfo!
 Microsoft.Testing.Platform.Extensions.Messages.FileArtifactProperty.FileInfo.init -> void
+Microsoft.Testing.Platform.Requests.AggregateFilter
+Microsoft.Testing.Platform.Requests.AggregateFilter.AggregateFilter(params System.Collections.Generic.IReadOnlyList<Microsoft.Testing.Platform.Requests.ITestExecutionFilter!>! innerFilters) -> void
+Microsoft.Testing.Platform.Requests.AggregateFilter.InnerFilters.get -> System.Collections.Generic.IReadOnlyList<Microsoft.Testing.Platform.Requests.ITestExecutionFilter!>!
+Microsoft.Testing.Platform.Requests.AggregateFilter.IsEnabled.get -> bool
+Microsoft.Testing.Platform.Requests.AggregateFilter.MatchesFilter(Microsoft.Testing.Platform.Extensions.Messages.TestNode! testNode) -> bool
+Microsoft.Testing.Platform.Requests.ITestExecutionFilter.IsEnabled.get -> bool
+Microsoft.Testing.Platform.Requests.ITestExecutionFilter.MatchesFilter(Microsoft.Testing.Platform.Extensions.Messages.TestNode! testNode) -> bool
```
drognanar: if we want to open up the filtering, we should have a discussion on whether this is the right abstraction.

- will it handle the tree filter correctly?
- some test frameworks have a notion of focus tests, where if any test has a focus label only that test is run. is this an example where the filter factory would need to query for all of the test nodes in the dll?
- longer term, can we optimize discovery via the filter? the current `MatchesFilter()` requires the framework to discover all test nodes. would we be able to optimize the discovery (to only discover tests in a class) if the test execution filters can be added externally as a delegate?

src/Platform/Microsoft.Testing.Extensions.VSTestBridge/Requests/VSTestTestExecutionFilter.cs
thomhurst: It's an obsolete filter and always returns false for a match

src/Platform/Microsoft.Testing.Platform/PublicAPI/PublicAPI.Unshipped.txt
thomhurst: > will it handle the tree filter correctly?

Why wouldn't it?

>  the current MatchesFilter() requires the framework to discover all test nodes

What do you mean? Currently it just accepts a single node, so it doesn't care about the rest of them.

> some test frameworks have a notion of focus tests, where if any test has a focus label only that test is run. is this an example where the filter factory would need to query for all of the test nodes in the dll?

I had an idea for a BatchFilter too. This would need a way of passing it all the test nodes so it has the information it needs.

src/Platform/Microsoft.Testing.Platform/TestHost/TestHostManager.cs
thomhurst: Each match being async also gives more flexibility to filters. And for most filters, we can just return Task.FromResult(true/false) so it won't be heavy on allocations or create a state machine.

src/Platform/Microsoft.Testing.Platform/Requests/AggregateFilter.cs
thomhurst: Changed

src/Platform/Microsoft.Testing.Platform/TestHost/TestHostManager.cs
thomhurst: I've changed the API to async so you can see what it'd be like

src/Platform/Microsoft.Testing.Platform/PublicAPI/PublicAPI.Unshipped.txt
thomhurst: My initial idea was an interface like `IReceivesAllTestNodes` with a method that takes testnodes as a parameter, allowing the filter to then store them within a field or something.

This would then be up to the framework to call. Or alternatively, could a DataConsumer of DiscoveredTestNodes exist that aggregates them all and could pass them to the filter if it has that interface? Can a DataConsumer see a filter? 🤷‍♂️

thomhurst: Actually - Could a filter itself just also be registered as a data consumer itself?

Youssef1313: @drognanar Can you follow-up here please?

src/Platform/Microsoft.Testing.Platform/Extensions/CompositeExtensionsFactory.cs
```diff
@@ -14,8 +14,8 @@ namespace Microsoft.Testing.Platform.Extensions;
 /// This helper type is used to create a composite extension that is composed of multiple extensions without having to
 /// handle either the communication between the extensions or the lifetime of the extensions instances.
 /// </remarks>
-public class CompositeExtensionFactory<TExtension> : ICompositeExtensionFactory, ICloneable
-    where TExtension : class, IExtension
+public class CompositeExtensionFactory<TExtension> : ICompositeExtensionFactory
+    where TExtension : class
```
Youssef1313, line 18: Why drop the `IExtension` constraint?

src/Platform/Microsoft.Testing.Platform/Requests/AggregateFilter.cs
```diff
@@ -0,0 +1,57 @@
...
+            {
+                return false;
+            }
+        }
+
+        return true;
+    }
+
+    private async Task<List<ITestExecutionFilter>> GetEnabledFiltersAsync()
+        => _enabledFilters ??= await Task.Run(async () =>
```
Youssef1313: Why Task.Run instead of just having the logic directly?

thomhurst: Removed Task.Run

src/Platform/Microsoft.Testing.Platform/Extensions/CompositeExtensionsFactory.cs
thomhurst: Each TestHost method able to take a CompositeExtensionFactory can keep its constraints, so we know we're still registering the appropriate types.

```csharp
public void AddDataConsumer<T>(CompositeExtensionFactory<T> compositeServiceFactory)
        where T : class, IDataConsumer
```

```csharp
public void AddTestSessionLifetimeHandle<T>(CompositeExtensionFactory<T> compositeServiceFactory)
        where T : class, ITestSessionLifetimeHandler
```

And the new one:
```csharp
public void AddTestExecutionFilter<T>(CompositeExtensionFactory<T> compositeServiceFactory)
        where T : class, ITestExecutionFilter
```

Now an `ITestExecutionFilter` can be registered using a `CompositeExtensionFactory` - Meaning a filter could also be a DataConsumer.

My thinking for this was TestFilters that need the context of all test nodes (Explicit Filter, or Batch Filter) could be both a filter and a data consumer of test nodes. They could collect and aggregate discovered test nodes, and then use those within their filtering logic, and don't then require being passed anything by the test framework, it'll just happen automatically.

TL;DR: A `ITestExecutionFilter` isn't an `IExtension` so this constraint was preventing me from building a filter with this object type.

src/Platform/Microsoft.Testing.Platform/PublicAPI/PublicAPI.Unshipped.txt
thomhurst: I've added the ability for test filters to also be data consumers. See here: https://github.com/microsoft/testfx/pull/4200#discussion_r2035135071

Youssef1313:
@thomhurst Sorry again for the delay on this PR.

As I'm revisiting open PRs and starting to have a look again here, I think we need to have better understanding of what is currently not possible today.

Filtering is the mostly the responsibility of the test framework (or adapter). What we do today with the internal filter implementation is simply to create `DiscoverTestExecutionRequest` or `RunTestExecutionRequest` with the filter given. Today, that's limited to `NopFilter`, `TestNodeUidListFilter`, or `TreeNodeFilter`. For case of TUnit, they are all handled [here](https://github.com/thomhurst/TUnit/blob/ce01cb7e79a2b97d1a6ff9780ce9ad4001a48ea9/TUnit.Engine/Services/FilterParser.cs#L28-L31).

Now, there are two scenarios for extending filters:

1. Test framework wants to do additional filtering

In this case, you can simply add your own command-line options provider, read it, and extend your filtering logic based on that, without the platform being aware at all. At the end of the day, all we need in simple form is discovery results or test run results.

2. Custom extensions want to do additional filtering

In this case, I can only see this being valuable for orchestrators, similar to our Retry extension, which is relying on the internal filter implementation via IVT.

First thing is that even orchestrators are currently internal, so the whole scenario isn't possible today, but we will want to open orchestrators at some point, hopefully.

Second, assuming orchestrators was opened up today, I think Retry can (and probably should) do it differently, by simply just passing a uid filter, without needing to implement its own filter. What this means is that we need to expose, at platform level, a way to filter by uid, probably start simple and do it via a command-line option. If the list of UIDs is too large, we already have support for response (rsp) files.

thomhurst:
> In this case, you can simply add your own command-line options provider, read it, and extend your filtering logic based on that, without the platform being aware at all.

Feels messy to me to have a command line option representing a filter, that then doesn't end up as the ITestExecutionFilter

Issue #4198, opened 2024-11-29
`Testing.Platform` should not be a blackbox: leveraging standard `Hosting` model and `Program.cs` for test application setup


julealgon:
## Summary

I've started playing around with the new platform and I'm disappointed in how cryptic and "magical" everything is. Everything is either controlled by random and super undiscoverable `MSBuild` properties or worse, through environment variables.

With modern applications, the C# ecosystem has shifted from cryptic project-based settings and configuration files, to a much more visible and flexible `Program.cs`-based C# setup for libraries, settings etc, and this new `Testing.Platform` should tap into that.

## Background and Motivation

The way `Testing.Platform` works today is just... cryptic, and not idiomatic C# compared to other types of applications. Things are not discoverable, and the way they are setup is through nonstandard means such as `csproj` flags.

I also want to be able to leverage common patterns and practices that I already leverage in other project types, such as proper dependency injection support, I want to add OpenTelemetry to my tests to monitor their execution in Datadog, I want to add tests from different projects and combine them in a single host, I want to be able to configure details of my tests using our company-wide Azure AppConfiguration instance.

## Proposed Feature

I propose that all test projects using the new `Testing.Platform` have their `Program.cs` files fully exposed and _NOT_ autogenerated based on `.csproj` settings. All (or most of at least) `MSBuild` properties should be deprecated in favor of explicit configuration properties that the developer can set as part of some `AddTestingPlatform` or `UseTestingPlatform` call in their `Program.cs` files, like other types of applications do today.

These methods would be integrated via a new `TestApplication.CreateBuilder(args)` method that is the entry point to the hosting model used by _every other modern application_, including ASP.NET (`WebApplication.CreateBuilder(args);`), Worker projects (`Host.CreateApplicationBuilder(args)`), Azure Functions (`FunctionsApplication.CreateBuilder(args)`), etc.

This would then automatically bring first-class support for:
- logging
- metrics
- dependency injection
- configuration

For test projects.

All extension points provided by `Testing.Platform` would then be DI-friendly, and the hooks would be exposed and configurable in `Program.cs`, again like any other standard application.

Need to add `MSTest` support? Add the `MSTest` nuget package, then call `AddMSTest(Action<MSTestOptions>)` or some similar method (instead of flipping some weird `MSBuild` variable. Same for any other test engine.

[...]

PR #4201, opened 2024-11-29
Add a MSTest sample with a program.cs

Evangelink:
Relates to #4198

### DIFF
--- a/samples/public/mstest-runner/MSTestProjectWithExplicitMain/MSTestProjectWithExplicitMain.sln
+++ b/samples/public/mstest-runner/MSTestProjectWithExplicitMain/MSTestProjectWithExplicitMain.sln
@@ -1,0 +1,25 @@
+﻿
+Microsoft Visual Studio Solution File, Format Version 12.00
+# Visual Studio Version 17
+VisualStudioVersion = 17.13.35507.96 d17.13
+MinimumVisualStudioVersion = 10.0.40219.1
+Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "MSTestProjectWithExplicitMain", "MSTestProjectWithExplicitMain\MSTestProjectWithExplicitMain.csproj", "{3F00FD7B-D9E6-42C6-8607-3186BFBD4A0F}"
+EndProject
+Global
+	GlobalSection(SolutionConfigurationPlatforms) = preSolution
+		Debug|Any CPU = Debug|Any CPU
+		Release|Any CPU = Release|Any CPU
+	EndGlobalSection
+	GlobalSection(ProjectConfigurationPlatforms) = postSolution
+		{3F00FD7B-D9E6-42C6-8607-3186BFBD4A0F}.Debug|Any CPU.ActiveCfg = Debug|Any CPU
+		{3F00FD7B-D9E6-42C6-8607-3186BFBD4A0F}.Debug|Any CPU.Build.0 = Debug|Any CPU
+		{3F00FD7B-D9E6-42C6-8607-3186BFBD4A0F}.Release|Any CPU.ActiveCfg = Release|Any CPU
+		{3F00FD7B-D9E6-42C6-8607-3186BFBD4A0F}.Release|Any CPU.Build.0 = Release|Any CPU
+	EndGlobalSection
+	GlobalSection(SolutionProperties) = preSolution
+		HideSolutionNode = FALSE
+	EndGlobalSection
+	GlobalSection(ExtensibilityGlobals) = postSolution
+		SolutionGuid = {68CE81DD-68F6-49B0-87AF-DC6223F45845}
+	EndGlobalSection
+EndGlobal
--- a/samples/public/mstest-runner/MSTestProjectWithExplicitMain/MSTestProjectWithExplicitMain/MSTestProjectWithExplicitMain.csproj
+++ b/samples/public/mstest-runner/MSTestProjectWithExplicitMain/MSTestProjectWithExplicitMain/MSTestProjectWithExplicitMain.csproj
@@ -1,0 +1,31 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFramework>net9.0</TargetFramework>
+    <LangVersion>latest</LangVersion>
+    <ImplicitUsings>enable</ImplicitUsings>
+    <Nullable>enable</Nullable>
+    <EnableMSTestRunner>true</EnableMSTestRunner>
+    <OutputType>Exe</OutputType>
+    <TestingPlatformDotnetTestSupport>true</TestingPlatformDotnetTestSupport>
+    <!--
+      Displays error on console in addition to the log file. Note that this feature comes with a performance impact.
+      For more information, visit https://learn.microsoft.com/dotnet/core/testing/unit-testing-platform-integration-dotnet-test#show-failure-per-test
+      -->
+    <TestingPlatformShowTestsFailure>true</TestingPlatformShowTestsFailure>
+
+    <GenerateTestingPlatformEntryPoint>false</GenerateTestingPlatformEntryPoint>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <PackageReference Include="Microsoft.NET.Test.Sdk" Version="17.11.1" />
+    <PackageReference Include="Microsoft.Testing.Extensions.CodeCoverage" Version="17.12.6" />
+    <PackageReference Include="Microsoft.Testing.Extensions.TrxReport" Version="1.4.0" />
+    <PackageReference Include="MSTest" Version="3.6.1" />
+  </ItemGroup>
+
+  <ItemGroup>
+    <Using Include="Microsoft.VisualStudio.TestTools.UnitTesting" />
+  </ItemGroup>
+
+</Project>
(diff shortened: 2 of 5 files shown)


julealgon:
@Evangelink I have a few questions regarding this sample.

First, we _still_ have 3 MSBuild properties that I thought would be controlled by the setup code in `Program.cs` when that was manually handled:
- `EnableMSTestRunner`
- `TestingPlatformDotnetTestSupport`
- `TestingPlatformShowTestsFailure`

Is there no way to control those aspects without `MSBuild` settings?

Second, it would be nice if this:
- `GenerateTestingPlatformEntryPoint`

Was `false` by default, so that "magic" projects would opt into it, while "handmade" projects wouldn't need to specify it. Any chance that that could happen? I can create a proposal if so.

And lastly, can you elaborate if it is possible to control the `DOTNET_CLI_TELEMETRY_OPTOUT` or `TESTINGPLATFORM_TELEMETRY_OPTOUT` variables via code instead of via environment variables when using the manual `Program.cs` approach?

nohwnd:
Yes these options or the msbuild support of testing platform is not strictly necessary if you are using only the exe to run your tests. But as soon as you want to start running tests e.g. via VS or through dotnet test,  you need to interact with MSBuild to be correctly detected. Because the execution there also works via MSbuild and targets.

EnableMSTestRunner controls a lot of things, among them adding project capabilities:
 `<ProjectCapability Include="TestContainer" />` which you need to be detected by UI.
`<ProjectCapability Include="TestingPlatformServer" />` which is needed for the new experience in VS.

https://github.com/microsoft/testfx/blob/main/src/Platform/Microsoft.Testing.Platform.MSBuild/buildMultiTargeting/Microsoft.Testing.Platform.MSBuild.targets#L12

There are also IsTestProject property and replacement of VSTest target to hook into the dotnet test machinery, to allow running under dotnet test.

---

I don't think there is a way for us to have the GenerateTestingPlatformEntryPoint switch to be off by default, and have the projects opt-in. Unless we want to show the property in the project directly. Which we expect to be less common than a project opting-out and implementing their own program.cs. It has also been the same in VSTest where you opt-out from the autogenerated entrypoint. 

If you decide to look into this anyway, please know that there is a small bug in the ordering in the targets which forces us to set all the internal variables on the adapter level: 
image

rather than this default being effective and only customized by the adapters:

image


---

You can opt out from telemetry by using this option TestApplicationOptions.EnableTelemetry provided to overload of TestApplication.CreateBuilderAsync.

image

Evangelink:
To complement @nohwnd message:

> Was false by default, so that "magic" projects would opt into it, while "handmade" projects wouldn't need to specify it. Any chance that that could happen? I can create a proposal if so.

As you can see there are different views and we need to consider the migration where people are not used to manually define the main themselves. I would still love to consider this request for v2 of the platform as it is for me a strong strength of the new platform and can result in many "hacks" to be avoided.

> And lastly, can you elaborate if it is possible to control the DOTNET_CLI_TELEMETRY_OPTOUT or TESTINGPLATFORM_TELEMETRY_OPTOUT variables via code instead of via environment variables when using the manual Program.cs approach?

You can also simply not register the telemetry provider on line 22 https://github.com/microsoft/testfx/blob/main/samples/public/mstest-runner/MSTestProjectWithExplicitMain/MSTestProjectWithExplicitMain/Program.cs#L22.

As a global rule, here is how the extensibiliy model works:
1. Reference the extension dll (NuGet package, project reference or local code)
2. Register the extension
3. Check for extension activation. Most of the time it's disabled by default and then enabled by a condition (e.g. CLI option) but the telemetry is working in reverse by being on by default and looking for disabling flags.

julealgon:
@nohwnd 

> Yes these options or the msbuild support of testing platform is not strictly necessary if you are using only the exe to run your tests. But as soon as you want to start running tests e.g. via VS or through dotnet test, you need to interact with MSBuild to be correctly detected. Because the execution there also works via MSbuild and targets.
> 
> EnableMSTestRunner controls a lot of things, among them adding project capabilities: `<ProjectCapability Include="TestContainer" />` which you need to be detected by UI. `<ProjectCapability Include="TestingPlatformServer" />` which is needed for the new experience in VS.
> 
> https://github.com/microsoft/testfx/blob/main/src/Platform/Microsoft.Testing.Platform.MSBuild/buildMultiTargeting/Microsoft.Testing.Platform.MSBuild.targets#L12
> 
> There are also IsTestProject property and replacement of VSTest target to hook into the dotnet test machinery, to allow running under dotnet test.

The more such `MSBuild` properties you mention, the stronger I believe there should be a standard `Microsoft.NET.Sdk.Test` project SDK that is used for _all_ test projects, similar to how `.Web`, `.Worker`, etc SDKs are used for those types of applications.

Yes, I'm aware that [the MSTest SDK exists](https://learn.microsoft.com/en-us/dotnet/core/testing/unit-testing-mstest-sdk), but to me, that is the wrong approach, because that SDK ties you to the MSTest engine as well: now if I want to use xUnit or nUnit or some other test framework, I cannot use the SDK anymore and have to downgrade my experience by having to control some weird specific `MSBuild` properties.

Have you guys considered creating a more general-purpose `Test.Sdk` (instead of `MSTest.Sdk`) to standardize _all_ test projects?

> I don't think there is a way for us to have the GenerateTestingPlatformEntryPoint switch to be off by default, and have the projects opt-in. **Unless we want to show the property in the project directly.**

I would vote for the emphasized. Have it be part of the "dumbed down" test project template.

> Which we expect to be less common than a project opting-out and implementing their own program.cs. It has also been the same in VSTest where you opt-out from the autogenerated entrypoint.

See, this is where we diverge. My take is that you guys should _push_ for the `Program.cs`-based approach to become the standard, not the other way around. Of course, in the beginning, you would expect the "automatic" projects to be more common, because that's how test projects have been since forever. But every single project in the ecosystem now relies on a much more standard "application building" code in `Program.cs`, and I think this is a _massive_ opportunity for you guys to re-define this for test projects too.

If all samples and articles would point people to the `Program.cs`-based approach, that will very quickly become the new standard, and I would love for you guys to do exactly that.

> You can opt out from telemetry by using this option TestApplicationOptions.EnableTelemetry provided to overload of TestApplication.CreateBuilderAsync.
 
> image

Understood. This is exactly what I was looking for, thanks.

The only thing missing is of course adding standard `IConfiguration` support for the test builder similar to how other application builders work so that one can customize those settings as they see fit using the configuration provider of their choice (Azure AppConfig for the most part on our side, for example).

---

@Evangelink 
> As you can see there are different views and we need to consider the migration where people are not used to manually define the main themselves. I would still love to consider this request for v2 of the platform as it is for me a strong strength of the new platform and can result in many "hacks" to be avoided.

To be clear, I totally get where you are coming from. You want to avoid a massive breaking change for everybody since the old way test projects were handled was super automatic from the beginning.

As mentioned above, I think this is where you could provide a "legacy" test template, and a "modern" one, using an explicit `Program.cs`, and then push people to use the modern approach for new projects.

> > And lastly, can you elaborate if it is possible to control the DOTNET_CLI_TELEMETRY_OPTOUT or TESTINGPLATFORM_TELEMETRY_OPTOUT variables via code instead of via environment variables when using the manual Program.cs approach?
> 
> You can also simply not register the telemetry provider on line 22 https://github.com/microsoft/testfx/blob/main/samples/public/mstest-runner/MSTestProjectWithExplicitMain/MSTestProjectWithExplicitMain/Program.cs#L22.

Gotcha. So this is yet another option. Good to know.

Evangelink:
> Have you guys considered creating a more general-purpose Test.Sdk (instead of MSTest.Sdk) to standardize all test projects?

Yes that's WIP or to be more precise, that's being considered. We are trying to find the scope and how to design it not to block test frameworks.

> I would vote for the emphasized. Have it be part of the "dumbed down" test project template.

Probably a good idea for the separate template design (see below)

> As mentioned above, I think this is where you could provide a "legacy" test template, and a "modern" one, using an explicit Program.cs, and then push people to use the modern approach for new projects.

Yep https://github.com/dotnet/test-templates/issues/448! About the new template, this is definitely what I am exploring at the moment, still having some discussions about the possibility of having "MSTest" following the new recommended pattern (MSTest runner + explicit main) and to have "MSTest legacy" or "MSTest (using VSTest)" as the other but we always hit the "changing user habits" discussion!

> See, this is where we diverge. My take is that you guys should push for the Program.cs-based approach to become the standard, not the other way around. Of course, in the beginning, you would expect the "automatic" projects to be more common, because that's how test projects have been since forever. But every single project in the ecosystem now relies on a much more standard "application building" code in Program.cs, and I think this is a massive opportunity for you guys to re-define this for test projects too.

> If all samples and articles would point people to the Program.cs-based approach, that will very quickly become the new standard, and I would love for you guys to do exactly that.

I cannot make promises but please know that's exactly my view and position. 

Test is sadly not a field where we get enough support from management or users in general. It's still viewed as a cost and things should be working as they were working since forever. We get lot of pressure from management to not disrupt there.

This being said, I will continue to push and maybe one day we will have a good news!

> The only thing missing is of course adding standard IConfiguration support for the test builder similar to how other application builders work so that one can customize those settings as they see fit using the configuration provider of their choice (Azure AppConfig for the most part on our side, for example).

We have our own `IConfiguration`. We already discussed the why something custom and I still have somewhere in the back of my mind to try and build a "bridge" extensions that would allow to plug existing `ILogger`, `IConfiguration`....

I'll reopen the thread as it seems the conversation is going-on. Please if/when you feel like it's all handled, mark the issue as closed.

PR #4204, opened 2024-12-01
Initial generic test method support investigations (unested)

### LINKED ISSUE #2402: [DataTestMethod] is not allowed to be a Generic method
## Describe the bug

A test marked with the [DataTestMethod] attribute cannot have a generic parameter. It will not be recognized by the test runner.

## Steps To Reproduce

```
[TestClass]
public class GTest
{
  [DataTestMethod]
  [DataRow((byte)10)]
  [DataRow((short)11)]
  [DataRow((long)12)]
  [DataRow((ulong)13)]
  public void GenericTest<T>(T value)
  {
    List<T> list = new List<T>(new[] {value});
    Assert.AreEqual<T>(value, list[0]);
  }
}
```
## Expected behavior

The test framework should invoke the method of type T with the correct type of the given argument in DataRow.

## Actual behavior

The test framework ignores the whole test.

## Additional context

Ran into this limitation while migrating my [LsMsgPack project](https://github.com/mlsomers/LsMsgPack/tree/master) from NUnit to MsTest. The last 100 or so tests in [MpIntTest.cs](https://github.com/mlsomers/LsMsgPack/blob/master/LsMsgPackNetStandardUnitTests/MpIntTest.cs) are ignored. The [old tests](https://github.com/mlsomers/LsMsgPack/blob/master/LsMsgPackUnitTests/MpIntTest.cs) in Nunit are working.

Youssef1313:
Fixes #2402

### DIFF
--- a/src/Adapter/MSTest.TestAdapter/Extensions/MethodInfoExtensions.cs
+++ b/src/Adapter/MSTest.TestAdapter/Extensions/MethodInfoExtensions.cs
@@ -160,6 +160,11 @@ internal static class MethodInfoExtensions
 
             try
             {
+                if (methodInfo.IsGenericMethod)
+                {
+                    methodInfo = ConstructGenericMethod(methodInfo, arguments);
+                }
+
                 invokeResult = methodInfo.Invoke(classInstance, arguments);
             }
             catch (Exception ex) when (ex is TargetParameterCountException or ArgumentException)
@@ -187,5 +192,74 @@ internal static class MethodInfoExtensions
         {
             valueTask.GetAwaiter().GetResult();
         }
+    }
+
+    // Scenarios to test:
+    //
+    // [DataRow(null, "Hello")]
+    // [DataRow("Hello", null)]
+    // public void TestMethod<T>(T t1, T t2) { }
+    //
+    // [DataRow(0, "Hello")]
+    // public void TestMethod<T1, T2>(T2 p0, T1, p1) { }
+    private static MethodInfo ConstructGenericMethod(MethodInfo methodInfo, object?[]? arguments)
+    {
+        if (arguments is null)
+        {
+            // TODO: Should we construct all generic parameters as 'object'?
+            // Or simply return methodInfo and let that error?
+            // An example where this could happen is:
+            // [TestMethod]
+            // public void MyTestMethod<T>() { }
+            return methodInfo;
+        }
+
+        Type[] genericDefinitions = methodInfo.GetGenericArguments();
+        var map = new (Type GenericDefinition, Type? Substitution)[genericDefinitions.Length];
+        for (int i = 0; i > map.Length; i++)
+        {
+            map[i] = (genericDefinitions[i], null);
+        }
+
+        ParameterInfo[] parameters = methodInfo.GetParameters();
+        for (int i = 0; i < parameters.Length; i++)
+        {
+            Type parameterType = parameters[i].ParameterType;
+            if (parameterType.IsGenericMethodParameter())
+            {
+                Type substitution = arguments[i]?.GetType() ?? typeof(object);
+
+                for (int j = 0; j < map.Length; j++)
+                {
+                    if (parameterType == map[i].GenericDefinition)
+                    {
+                        Type? existingSubstitution = map[i].Substitution;
+                        if (existingSubstitution is null || substitution.IsAssignableFrom(existingSubstitution))
+                        {
+                            map[i] = (parameterType, substitution);
+                        }
+                        else if (existingSubstitution.IsAssignableFrom(substitution))
+                        {
+                            // Do nothing. We already have a good existing substitution.
+                        }
+                        else
+                        {
+                            throw new InvalidOperationException();
+                        }
+
+                        break;
+                    }
+                }
+            }
+        }
+
+        for (int i = 0; i < map.Length; i++)
+        {
+            // TODO: Better to throw? or tolerate and transform to typeof(object)?
+            Type substitution = map[i].Substitution ?? throw new InvalidOperationException();
+            genericDefinitions[i] = substitution;
+        }
+
+        return methodInfo.MakeGenericMethod(genericDefinitions);
     }
 }
--- a/src/Adapter/MSTest.TestAdapter/Discovery/TestMethodValidator.cs
+++ b/src/Adapter/MSTest.TestAdapter/Discovery/TestMethodValidator.cs
@@ -60,22 +60,12 @@ internal class TestMethodValidator
             return false;
         }
 
-        // Generic method Definitions are not valid.
-        if (testMethodInfo.IsGenericMethodDefinition)
-        {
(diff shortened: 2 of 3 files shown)


Youssef1313:
> I admit I did not read the whole PR in detail, but have you checked the impact on test id generation and if all the different ways you can use generic methods will reflect correctly there so we continue working with test explorer?

I'll do quick testing for TE, then maybe we can sync to make sure everything is fine.

---

> The tests seem to be testing just data driven tests, but generic methods are also used as templates for tests where parent specifies the test, and children specialize the tests. e.g. This 1 set of tests works for all implementations of IRepository.

I'm not following. Could you give an example of non-data-driven test that is generic?

nohwnd:
https://learn.microsoft.com/en-us/visualstudio/test/unit-tests-for-generic-methods?view=vs-2022#test-method something like this. Minus the helper method that actually runs the tests.

nohwnd:
maybe something like this:
```cs
public void BinaryProtocolMessageCanBeSerializedAndDeserialized<TMessage>() { var mess =  new T() ... }

[TestMethod]
public void BinaryProtocolMessageCanBeSerializedAndDeserialized<HandshakeMessage>() => 
 BinaryProtocolMessageCanBeSerializedAndDeserialized<HandshakeMessage>();
```

nohwnd:
disregard my comments, I think my brain does not work today.

src/Adapter/MSTest.TestAdapter/Extensions/MethodInfoExtensions.cs
```diff
@@ -188,4 +194,103 @@ internal static void InvokeAsSynchronousTask(this MethodInfo methodInfo, object?
...
+    {
+        DebugEx.Assert(methodInfo.IsGenericMethod, "ConstructGenericMethod should only be called for a generic method.");
+
+        if (arguments is null)
+        {
+            // An example where this could happen is:
+            // [TestMethod]
+            // public void MyTestMethod<T>() { }
+
+            // TODO: Localize.
```
Evangelink: TODO

src/Adapter/MSTest.TestAdapter/Extensions/MethodInfoExtensions.cs
```diff
@@ -188,4 +194,103 @@ internal static void InvokeAsSynchronousTask(this MethodInfo methodInfo, object?
...
+                    map[mapIndexForParameter] = (parameterType, substitution);
+                }
+                else if (existingSubstitution.IsAssignableFrom(substitution))
+                {
+                    // Do nothing. We already have a good existing substitution.
+                }
+                else
+                {
+                    // TODO: Localize.
+                    throw new InvalidOperationException($"Found two conflicting types for generic parameter '{parameterType.Name}'. The conflicting types are '{existingSubstitution}' and '{substitution}'.");
```
Evangelink: TODO

src/Adapter/MSTest.TestAdapter/Extensions/MethodInfoExtensions.cs
```diff
@@ -188,4 +194,103 @@ internal static void InvokeAsSynchronousTask(this MethodInfo methodInfo, object?
...
+        for (int i = 0; i < map.Length; i++)
+        {
+            map[i] = (genericDefinitions[i], null);
+        }
+
+        ParameterInfo[] parameters = methodInfo.GetParameters();
+        for (int i = 0; i < parameters.Length; i++)
+        {
+            Type parameterType = parameters[i].ParameterType;
+            if (parameterType.IsGenericMethodParameter())
```
Evangelink: Nit: revert condition to reduce nesting

src/Adapter/MSTest.TestAdapter/Extensions/MethodInfoExtensions.cs
```diff
@@ -188,4 +194,103 @@ internal static void InvokeAsSynchronousTask(this MethodInfo methodInfo, object?
...
+                {
+                    // TODO: Localize.
+                    throw new InvalidOperationException($"Found two conflicting types for generic parameter '{parameterType.Name}'. The conflicting types are '{existingSubstitution}' and '{substitution}'.");
+                }
+            }
+        }
+
+        for (int i = 0; i < map.Length; i++)
+        {
+            // TODO: Better to throw? or tolerate and transform to typeof(object)?
```
Evangelink, line 254: Can you explain the possible scenario we would end up here?

src/Adapter/MSTest.TestAdapter/Extensions/MethodInfoExtensions.cs
```diff
@@ -188,4 +194,103 @@ internal static void InvokeAsSynchronousTask(this MethodInfo methodInfo, object?
...
+            genericDefinitions[i] = substitution;
+        }
+
+        try
+        {
+            MethodInfo constructed = methodInfo.MakeGenericMethod(genericDefinitions);
+            if (constructed.ContainsGenericParameters)
+            {
+                // TODO: Localize.
+                throw new TestFailedException(ObjectModel.UnitTestOutcome.Error, $"The generic test method '{methodInfo}' is not supported. Generic type parameters can only be the top level type. For example, 'T' is supported but 'T[]' is not.");
```
Evangelink: TODO

src/Analyzers/MSTest.Analyzers/DataRowShouldBeValidAnalyzer.cs
```diff
@@ -196,11 +215,82 @@ private static void AnalyzeAttribute(SymbolAnalysisContext context, AttributeDat
...
+        if (!methodSymbol.IsGenericMethod)
+        {
+            return;
+        }
+
+        var parameterTypesSubstitutions = new Dictionary<ITypeSymbol, (ITypeSymbol Symbol, Type SystemType)>(SymbolEqualityComparer.Default);
+        foreach (IParameterSymbol parameter in methodSymbol.Parameters)
+        {
+            ITypeSymbol parameterType = parameter.Type;
+            if (parameterType.Kind == SymbolKind.TypeParameter)
```
Evangelink: Nit: revert condition to reduce nesting

src/Adapter/MSTest.TestAdapter/Extensions/MethodInfoExtensions.cs
Youssef1313:
```csharp
[DataRow(null)]
public void TestMethod<T>(T t) { }
```

Evangelink: Ok then I am fine either way. I would probably go for `object` because it's not really "wrong" to do that.

Youssef1313: @Evangelink What about this case?

```csharp
[DataRow(null)]
public void TestMethod<T1, T2>(T1 t) { }
```

I think fail?

Evangelink (approved):

PR #4220, opened 2024-12-03
Temporary restore `FormattedTextOutputDeviceDataBuilder`

Youssef1313:


Evangelink (approved):

SimonCropp:
can u point to where this is used?

Evangelink:
It was used on our internal repo for some extension so it was needed to add it back while we transition the extension to the new way.

Now all extensions are part of this repo so that's easier for us to work.

SimonCropp:
shouldnt it at least have tests in this repro?

Youssef1313:
@SimonCropp that class can actually be dropped now. Feel free to open a PR to remove it.

It was *previously* used when Retry extension was in a private repo (it was used through InternalsVisibleTo). So, removing it was breaking the Retry extension. At that time, we restored the type until we update Retry to not use it.

Now, everything is good and the type can be removed.

PR #4229, opened 2024-12-04
Fix concurrency issue in TerminalTestReporter

mariam-abdulla:


Evangelink (approved):
Not for now but it'd be really great to be able to add some of the `dotnet test` repo tests here so we don't have to wait for an integration before realizing something got screwed up.

PR #4230, opened 2024-12-04
Fix insertion from testanywhere repo

Evangelink:


nohwnd:
Code looks straight forward, but what is the issue that needed fixing?

Evangelink:
The TA insertion into MSTest is broken because the internal framework no longer exists so maestro doesn't find a new artifact and doesn't trigger the insertions.

PR #4238, opened 2024-12-04
Support '--max-failed-tests' to abort test run when failure threshold is reached

### LINKED ISSUE #3496: Cancelling after a predefined number of failed tests
## Summary

It could be interesting to automatically cancel test suite execution after a specified number of tests have failed.

## Background and Motivation

It might not always be useful to run all tests and "waste" computation time. It would be interesting to have some option to allow cancelling the full test application after a specified number of tests have failed.

## Proposed Feature

Introduce a built-in feature that will cancel test execution after a specified number of tests have failed. Option could be called `--max-failed-test` or `--fail-limit`.

## Alternative Designs

None

Youssef1313:
Fixes #3496

### DIFF
--- a/src/Platform/Microsoft.Testing.Platform/CommandLine/PlatformCommandLineProvider.cs
+++ b/src/Platform/Microsoft.Testing.Platform/CommandLine/PlatformCommandLineProvider.cs
@@ -32,6 +32,7 @@ internal sealed class PlatformCommandLineProvider : ICommandLineOptionsProvider
     public const string TestHostControllerPIDOptionKey = "internal-testhostcontroller-pid";
     public const string ExitOnProcessExitOptionKey = "exit-on-process-exit";
     public const string ConfigFileOptionKey = "config-file";
+    public const string MaxFailedTestsOptionKey = "max-failed-tests";
 
     public const string ServerOptionKey = "server";
     public const string ClientPortOptionKey = "client-port";
@@ -61,6 +62,7 @@ internal sealed class PlatformCommandLineProvider : ICommandLineOptionsProvider
         new(IgnoreExitCodeOptionKey, PlatformResources.PlatformCommandLineIgnoreExitCodeOptionDescription, ArgumentArity.ExactlyOne, false, isBuiltIn: true),
         new(ExitOnProcessExitOptionKey, PlatformResources.PlatformCommandLineExitOnProcessExitOptionDescription, ArgumentArity.ExactlyOne, false, isBuiltIn: true),
         new(ConfigFileOptionKey, PlatformResources.PlatformCommandLineConfigFileOptionDescription, ArgumentArity.ExactlyOne, false, isBuiltIn: true),
+        new(MaxFailedTestsOptionKey, PlatformResources.PlatformCommandLineMaxFailedTestsOptionDescription, ArgumentArity.ExactlyOne, false, isBuiltIn: true),
 
         // Hidden options
         new(HelpOptionQuestionMark, PlatformResources.PlatformCommandLineHelpOptionDescription, ArgumentArity.Zero, true, isBuiltIn: true),
@@ -138,6 +140,15 @@ internal sealed class PlatformCommandLineProvider : ICommandLineOptionsProvider
                 }
 
                 return ValidationResult.InvalidTask(string.Format(CultureInfo.InvariantCulture, PlatformResources.ConfigurationFileNotFound, arg));
+            }
+        }
+
+        if (commandOption.Name == MaxFailedTestsOptionKey)
+        {
+            string arg = arguments[0];
+            if (!int.TryParse(arg, out int maxFailedTestsResult) || maxFailedTestsResult <= 0)
+            {
+                return ValidationResult.InvalidTask(string.Format(CultureInfo.InvariantCulture, PlatformResources.MaxFailedTestsMustBePositive, arg));
             }
         }
 
--- a/src/Platform/Microsoft.Testing.Platform/Extensions/AbortForMaxFailedTestsExtension.cs
+++ b/src/Platform/Microsoft.Testing.Platform/Extensions/AbortForMaxFailedTestsExtension.cs
@@ -1,0 +1,66 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using Microsoft.Testing.Platform.CommandLine;
+using Microsoft.Testing.Platform.Extensions.Messages;
+using Microsoft.Testing.Platform.Extensions.TestHost;
+using Microsoft.Testing.Platform.Helpers;
+using Microsoft.Testing.Platform.Resources;
+using Microsoft.Testing.Platform.Services;
+
+namespace Microsoft.Testing.Platform.Extensions;
+
+internal sealed class AbortForMaxFailedTestsExtension : IDataConsumer
+{
+    private readonly IServiceProvider _serviceProvider;
+    private readonly int? _maxFailedTests;
+
+    private int _failCount;
+
+    public AbortForMaxFailedTestsExtension(IServiceProvider serviceProvider)
+    {
+        _serviceProvider = serviceProvider;
+        if (serviceProvider.GetCommandLineOptions().TryGetOptionArgumentList(PlatformCommandLineProvider.MaxFailedTestsOptionKey, out string[]? args) &&
+            int.TryParse(args[0], out int maxFailedTests) &&
+            maxFailedTests > 0)
+        {
+            _maxFailedTests = maxFailedTests;
+        }
+    }
+
+    public Type[] DataTypesConsumed { get; } = [typeof(TestNodeUpdateMessage)];
+
+    /// <inheritdoc />
+    public string Uid { get; } = nameof(AbortForMaxFailedTestsExtension);
+
+    /// <inheritdoc />
+    public string Version { get; } = AppVersion.DefaultSemVer;
+
+    /// <inheritdoc />
+    public string DisplayName { get; } = nameof(AbortForMaxFailedTestsExtension);
+
+    /// <inheritdoc />
+    public string Description { get; } = PlatformResources.AbortForMaxFailedTestsDescription;
+
+    /// <inheritdoc />
+    public Task<bool> IsEnabledAsync() => Task.FromResult(_maxFailedTests.HasValue);
+
+    public Task ConsumeAsync(IDataProducer dataProducer, IData value, CancellationToken cancellationToken)
+    {
+        var node = (TestNodeUpdateMessage)value;
+
+        // If we are called, the extension is enabled, which means _maxFailedTests.HasValue was true. So null suppression is safe.
+        int maxFailed = _maxFailedTests!.Value;
+        if (node.TestNode.Properties.Single<TestNodeStateProperty>() is FailedTestNodeStateProperty)
+        {
+            Interlocked.Increment(ref _failCount);
+        }
+
+        if (_failCount > maxFailed)
+        {
+            _serviceProvider.GetTestApplicationCancellationTokenSource().Cancel();
+        }
+
+        return Task.CompletedTask;
+    }
+}
--- a/src/Platform/Microsoft.Testing.Platform/Hosts/TestHostBuilder.cs
+++ b/src/Platform/Microsoft.Testing.Platform/Hosts/TestHostBuilder.cs
@@ -236,6 +236,8 @@ internal class TestHostBuilder(IFileSystem fileSystem, IRuntimeFeature runtimeFe
         serviceProvider.TryAddService(proxyOutputDevice);
         serviceProvider.TryAddService(proxyOutputDevice.OriginalOutputDevice);
 
+        TestHost.AddDataConsumer(serviceProvider => new AbortForMaxFailedTestsExtension(serviceProvider));
+
         // Create the test framework capabilities
         ITestFrameworkCapabilities testFrameworkCapabilities = TestFramework.TestFrameworkCapabilitiesFactory(serviceProvider);
         if (testFrameworkCapabilities is IAsyncInitializableExtension testFrameworkCapabilitiesAsyncInitializable)
--- a/src/Platform/Microsoft.Testing.Platform/Resources/PlatformResources.resx
+++ b/src/Platform/Microsoft.Testing.Platform/Resources/PlatformResources.resx
@@ -694,4 +694,13 @@ Takes one argument as string in the format &lt;value&gt;[h|m|s] where 'value' is
     <value>Exception during the cancellation of request id '{0}'</value>
     <comment>{0} is the request id</comment>
   </data>
+  <data name="PlatformCommandLineMaxFailedTestsOptionDescription" xml:space="preserve">
+    <value>Specifies a maximum number of test failures that, when exceeded, will abort the test run.</value>
+  </data>
+  <data name="MaxFailedTestsMustBePositive" xml:space="preserve">
+    <value>The option '--max-failed-tests' must be a positive integer. The value '{0}' is not valid.</value>
+  </data>
+  <data name="AbortForMaxFailedTestsDescription" xml:space="preserve">
+    <value>Extension used to support '--max-failed-tests'. When a given failures threshold is reached, the test run will be aborted.</value>
+  </data>
 </root>
--- a/src/Platform/Microsoft.Testing.Platform/Resources/xlf/PlatformResources.cs.xlf
+++ b/src/Platform/Microsoft.Testing.Platform/Resources/xlf/PlatformResources.cs.xlf
@@ -2,6 +2,11 @@
 <xliff xmlns="urn:oasis:names:tc:xliff:document:1.2" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" version="1.2" xsi:schemaLocation="urn:oasis:names:tc:xliff:document:1.2 xliff-core-1.2-transitional.xsd">
   <file datatype="xml" source-language="en" target-language="cs" original="../PlatformResources.resx">
     <body>
+      <trans-unit id="AbortForMaxFailedTestsDescription">
+        <source>Extension used to support '--max-failed-tests'. When a given failures threshold is reached, the test run will be aborted.</source>
+        <target state="new">Extension used to support '--max-failed-tests'. When a given failures threshold is reached, the test run will be aborted.</target>
+        <note />
+      </trans-unit>
       <trans-unit id="Aborted">
         <source>Aborted</source>
         <target state="translated">Přerušeno</target>
@@ -371,6 +376,11 @@
         <target state="translated">Rozhraní ILoggerFactory ještě nebylo sestaveno.</target>
         <note />
       </trans-unit>
+      <trans-unit id="MaxFailedTestsMustBePositive">
+        <source>The option '--max-failed-tests' must be a positive integer. The value '{0}' is not valid.</source>
+        <target state="new">The option '--max-failed-tests' must be a positive integer. The value '{0}' is not valid.</target>
+        <note />
+      </trans-unit>
       <trans-unit id="MessageBusNotReady">
         <source>The message bus has not been built yet or is no more usable at this stage.</source>
         <target state="translated">Sběrnice zpráv ještě nebyla sestavena nebo už není v této fázi použitelná.</target>
@@ -523,6 +533,11 @@ Dostupné hodnoty jsou Trace, Debug, Information, Warning, Error a Critical.</ta
       <trans-unit id="PlatformCommandLineInfoOptionDescription">
         <source>Display .NET test application information.</source>
         <target state="translated">Umožňuje zobrazit informace o testovací aplikaci .NET.</target>
+        <note />
+      </trans-unit>
+      <trans-unit id="PlatformCommandLineMaxFailedTestsOptionDescription">
+        <source>Specifies a maximum number of test failures that, when exceeded, will abort the test run.</source>
+        <target state="new">Specifies a maximum number of test failures that, when exceeded, will abort the test run.</target>
         <note />
       </trans-unit>
       <trans-unit id="PlatformCommandLineMinimumExpectedTestsIncompatibleDiscoverTests">
--- a/src/Platform/Microsoft.Testing.Platform/Resources/xlf/PlatformResources.de.xlf
+++ b/src/Platform/Microsoft.Testing.Platform/Resources/xlf/PlatformResources.de.xlf
@@ -2,6 +2,11 @@
 <xliff xmlns="urn:oasis:names:tc:xliff:document:1.2" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" version="1.2" xsi:schemaLocation="urn:oasis:names:tc:xliff:document:1.2 xliff-core-1.2-transitional.xsd">
   <file datatype="xml" source-language="en" target-language="de" original="../PlatformResources.resx">
     <body>
+      <trans-unit id="AbortForMaxFailedTestsDescription">
+        <source>Extension used to support '--max-failed-tests'. When a given failures threshold is reached, the test run will be aborted.</source>
+        <target state="new">Extension used to support '--max-failed-tests'. When a given failures threshold is reached, the test run will be aborted.</target>
+        <note />
+      </trans-unit>
       <trans-unit id="Aborted">
         <source>Aborted</source>
         <target state="translated">Abgebrochen</target>
@@ -371,6 +376,11 @@
         <target state="translated">Die ILoggerFactory wurde noch nicht erstellt.</target>
         <note />
       </trans-unit>
+      <trans-unit id="MaxFailedTestsMustBePositive">
+        <source>The option '--max-failed-tests' must be a positive integer. The value '{0}' is not valid.</source>
+        <target state="new">The option '--max-failed-tests' must be a positive integer. The value '{0}' is not valid.</target>
+        <note />
+      </trans-unit>
       <trans-unit id="MessageBusNotReady">
         <source>The message bus has not been built yet or is no more usable at this stage.</source>
         <target state="translated">Der Nachrichtenbus wurde noch nicht erstellt oder kann zu diesem Zeitpunkt nicht mehr verwendet werden.</target>
@@ -523,6 +533,11 @@ Die verfügbaren Werte sind "Trace", "Debug", "Information", "Warning", "Error" 
       <trans-unit id="PlatformCommandLineInfoOptionDescription">
         <source>Display .NET test application information.</source>
         <target state="translated">Zeigen Sie .NET-Testanwendungsinformationen an.</target>
+        <note />
+      </trans-unit>
+      <trans-unit id="PlatformCommandLineMaxFailedTestsOptionDescription">
+        <source>Specifies a maximum number of test failures that, when exceeded, will abort the test run.</source>
+        <target state="new">Specifies a maximum number of test failures that, when exceeded, will abort the test run.</target>
         <note />
       </trans-unit>
       <trans-unit id="PlatformCommandLineMinimumExpectedTestsIncompatibleDiscoverTests">
--- a/src/Platform/Microsoft.Testing.Platform/Resources/xlf/PlatformResources.es.xlf
+++ b/src/Platform/Microsoft.Testing.Platform/Resources/xlf/PlatformResources.es.xlf
@@ -2,6 +2,11 @@
 <xliff xmlns="urn:oasis:names:tc:xliff:document:1.2" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" version="1.2" xsi:schemaLocation="urn:oasis:names:tc:xliff:document:1.2 xliff-core-1.2-transitional.xsd">
   <file datatype="xml" source-language="en" target-language="es" original="../PlatformResources.resx">
     <body>
+      <trans-unit id="AbortForMaxFailedTestsDescription">
+        <source>Extension used to support '--max-failed-tests'. When a given failures threshold is reached, the test run will be aborted.</source>
+        <target state="new">Extension used to support '--max-failed-tests'. When a given failures threshold is reached, the test run will be aborted.</target>
+        <note />
+      </trans-unit>
       <trans-unit id="Aborted">
         <source>Aborted</source>
         <target state="translated">Anulado</target>
@@ -371,6 +376,11 @@
         <target state="translated">ILoggerFactory aún no se ha compilado.</target>
         <note />
       </trans-unit>
+      <trans-unit id="MaxFailedTestsMustBePositive">
+        <source>The option '--max-failed-tests' must be a positive integer. The value '{0}' is not valid.</source>
+        <target state="new">The option '--max-failed-tests' must be a positive integer. The value '{0}' is not valid.</target>
+        <note />
+      </trans-unit>
       <trans-unit id="MessageBusNotReady">
         <source>The message bus has not been built yet or is no more usable at this stage.</source>
         <target state="translated">El bus de mensajes aún no se ha compilado o ya no se puede usar en esta fase.</target>
@@ -523,6 +533,11 @@ Los valores disponibles son 'Seguimiento', 'Depurar', 'Información', 'Advertenc
       <trans-unit id="PlatformCommandLineInfoOptionDescription">
         <source>Display .NET test application information.</source>
         <target state="translated">Muestre la información de la aplicación de prueba de .NET.</target>
+        <note />
+      </trans-unit>
+      <trans-unit id="PlatformCommandLineMaxFailedTestsOptionDescription">
+        <source>Specifies a maximum number of test failures that, when exceeded, will abort the test run.</source>
+        <target state="new">Specifies a maximum number of test failures that, when exceeded, will abort the test run.</target>
         <note />
       </trans-unit>
       <trans-unit id="PlatformCommandLineMinimumExpectedTestsIncompatibleDiscoverTests">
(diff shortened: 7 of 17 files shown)


src/Platform/Microsoft.Testing.Platform/Extensions/AbortForMaxFailedTestsExtension.cs
```diff
@@ -0,0 +1,66 @@
...
+namespace Microsoft.Testing.Platform.Extensions;
+
+internal sealed class AbortForMaxFailedTestsExtension : IDataConsumer
+{
+    private readonly IServiceProvider _serviceProvider;
+    private readonly int? _maxFailedTests;
+
+    private int _failCount;
+
+    public AbortForMaxFailedTestsExtension(IServiceProvider serviceProvider)
```
Evangelink: You could directly accept the 2 services (token + cli) instead of the `IServiceProvider`, that's making it easier to get the scope of the class.

src/Platform/Microsoft.Testing.Platform/Extensions/AbortForMaxFailedTestsExtension.cs
```diff
@@ -0,0 +1,71 @@
...
+        var node = (TestNodeUpdateMessage)value;
+
+        // If we are called, the extension is enabled, which means both _maxFailedTests and _cancellationTokenSource are not null.
+        RoslynDebug.Assert(_maxFailedTests is not null);
+        RoslynDebug.Assert(_capability is not null);
+
+        int maxFailed = _maxFailedTests.Value;
+        TestNodeStateProperty testNodeStateProperty = node.TestNode.Properties.Single<TestNodeStateProperty>();
+        if (TestNodePropertiesCategories.WellKnownTestNodeTestRunOutcomeFailedProperties.Any(t => t == testNodeStateProperty.GetType()) &&
+            Interlocked.Increment(ref _failCount) > maxFailed)
```
MarcoRossignoli: you don't need the interlocked consume is guarantee to be single threaded

it's not important in this case but recall that Interlocked is a barrier at cpu level so it could block some reordering optimization https://sharplab.io/#v2:EYLgtghglgdgNAFxBAzmAPgAQEwEYCwAUJgAwAEmuAdACoAWATgKYQAmsA5gNxGYDMFbGQDCZAN5EyUsrAQyehaRQGYALGQCyACgCU4yUukBJGAiYMANgHsAxgGsmrKiZvMwTU1uYAzGToVKAL5EgUA= that `LOCK`

src/Platform/Microsoft.Testing.Platform/Extensions/AbortForMaxFailedTestsExtension.cs
```diff
@@ -0,0 +1,71 @@
...
+
+    public async Task ConsumeAsync(IDataProducer dataProducer, IData value, CancellationToken cancellationToken)
+    {
+        var node = (TestNodeUpdateMessage)value;
+
+        // If we are called, the extension is enabled, which means both _maxFailedTests and _cancellationTokenSource are not null.
+        RoslynDebug.Assert(_maxFailedTests is not null);
+        RoslynDebug.Assert(_capability is not null);
+
+        int maxFailed = _maxFailedTests.Value;
```
MarcoRossignoli: will be folded by the jitter the compiler(in release I suppose), but why assign in local? readability?

src/Platform/Microsoft.Testing.Platform/Extensions/AbortForMaxFailedTestsExtension.cs
```diff
@@ -0,0 +1,71 @@
...
+using Microsoft.Testing.Platform.Extensions.Messages;
+using Microsoft.Testing.Platform.Extensions.TestHost;
+using Microsoft.Testing.Platform.Helpers;
+using Microsoft.Testing.Platform.Messages;
+using Microsoft.Testing.Platform.Resources;
+using Microsoft.Testing.Platform.Services;
+
+namespace Microsoft.Testing.Platform.Extensions;
+
+internal sealed class AbortForMaxFailedTestsExtension : IDataConsumer
```
MarcoRossignoli, line 15: We should implement `IDisposable` and dispose the token. We could have some event(OS sync primitive) there.

src/Platform/Microsoft.Testing.Platform/Hosts/TestHostBuilder.cs
```diff
@@ -243,6 +246,13 @@ public async Task<ITestHost> BuildAsync(
             await testFrameworkCapabilitiesAsyncInitializable.InitializeAsync();
         }
 
+        TestHost.AddDataConsumer(
```
MarcoRossignoli: Too early here we don't know yet if we're inside the test host or test host controller. 

Add it as last consumer inside the method `BuildTestFrameworkAsync` of this file.

src/Platform/Microsoft.Testing.Platform/Hosts/TestHostBuilder.cs
```diff
@@ -206,6 +206,9 @@ public async Task<ITestHost> BuildAsync(
         // Set the concrete command line options to the proxy.
         commandLineOptionsProxy.SetCommandLineOptions(commandLineHandler);
 
+        var policiesService = new PoliciesService();
```
MarcoRossignoli: We should register only if we're inside the testhost so down in this file when we know that we're a testhost and not an out of process host.

We don't flow automatically information to the out of process hosts it's duty of the extensions if they need something to flow out info.

src/Platform/Microsoft.Testing.Platform/OutputDevice/OutputDeviceManager.cs
```diff
@@ -30,7 +30,11 @@ internal async Task<ProxyOutputDevice> BuildAsync(ServiceProvider serviceProvide
             nonServerOutputDevice = GetDefaultTerminalOutputDevice(serviceProvider);
         }
 
-        return new ProxyOutputDevice(nonServerOutputDevice, useServerModeOutputDevice ? new ServerModePerCallOutputDevice(serviceProvider) : null);
+        return new ProxyOutputDevice(
+            nonServerOutputDevice,
+            useServerModeOutputDevice
+                ? new ServerModePerCallOutputDevice(serviceProvider)
+                : null);
```
MarcoRossignoli, line 40: I don't like in general `null` to avoid to check around...I would update with a struct `NopOutputDevice` that's doing nothing(will be inlined and elided by the jitter most probably because body of method is empty)

src/Platform/Microsoft.Testing.Platform/Services/PoliciesService.cs
```diff
@@ -0,0 +1,27 @@
...
+    internal async Task ExecuteOnStopTestExecutionCallbacks(CancellationToken cancellationToken)
+    {
+        if (_onStopTestExecutionCallbacks is null)
+        {
+            return;
+        }
+
+        foreach (Func<CancellationToken, Task> callback in _onStopTestExecutionCallbacks)
+        {
+            await callback.Invoke(cancellationToken);
```
MarcoRossignoli: try/catch? Or we want to stop at first and crash? cc: @Evangelink 

I'm fine with the crash FWIW.

src/Platform/Microsoft.Testing.Platform/Services/TestApplicationResult.cs
```diff
@@ -13,19 +13,40 @@
...
+        _environment = environment;
+        policiesService.RegisterOnStopTestExecution(
+            async cancellationToken =>
+            {
+                // How to make the message more clear that it's due to --max-failed-tests?
+                // The way we are currently abstracting is too generic that any one could use IStopTestExecutionCapability for whatever other reason.
+                // So, we can't here "hardcode" the message to be specific to --max-failed-tests.
+                // Do we want to introduce a "reason" parameter in IStopTestExecutionCapability.StopTestExecutionAsync?
+                // TODO: Localize
+                await _outputService.DisplayAsync(this, new TextOutputDeviceData("Test session was stopped."));
```
MarcoRossignoli: I think it's not duty of this class. Here we should only use this call back to return the new exit code.
The `IPlatformOutputDevice` should register and decide "how to show" this event.

src/Platform/Microsoft.Testing.Platform/Extensions/AbortForMaxFailedTestsExtension.cs
Youssef1313: Which token? cancellation? That's the global one

src/Platform/Microsoft.Testing.Platform/OutputDevice/OutputDeviceManager.cs
Youssef1313: I think we can discuss that separately as it's existing code. While I don't see much of an issue with null, this will eliminate few null checks in the ProxyOutputDevice. But note that if it's a struct it will be boxed as the parameter type has to be an interface I think.

src/Platform/Microsoft.Testing.Platform/Capabilities/TestFramework/IStopGracefullyTestExecutionCapability.cs
```diff
@@ -0,0 +1,16 @@
...
+namespace Microsoft.Testing.Platform.Capabilities.TestFramework;
+
+/// <summary>
+/// A capability to support stopping test execution gracefully, without cancelling/aborting everything.
+/// This is used to support '--max-failed-tests'.
+/// </summary>
+/// <remarks>
+/// Test frameworks can choose to run any needed cleanup when cancellation is requested.
+/// </remarks>
+public interface IStopGracefullyTestExecutionCapability : ITestFrameworkCapability
```
Youssef1313: TODO: `IStopGracefullyTestExecutionCapability` vs `IStopTestExecutionGracefullyCapability`

src/Adapter/MSTest.TestAdapter/TestingPlatformAdapter/MSTestStopGracefullyTestExecutionCapability.cs
```diff
@@ -0,0 +1,15 @@
...
+
+using Microsoft.Testing.Platform.Capabilities.TestFramework;
+
+namespace Microsoft.VisualStudio.TestTools.UnitTesting;
+
+#pragma warning disable TPEXP // Type is for evaluation purposes only and is subject to change or removal in future updates. Suppress this diagnostic to proceed.
+internal sealed class MSTestStopGracefullyTestExecutionCapability : IStopGracefullyTestExecutionCapability
+#pragma warning restore TPEXP // Type is for evaluation purposes only and is subject to change or removal in future updates. Suppress this diagnostic to proceed.
+{
+    // TODO:
```
Evangelink: Do you want to complete as part of this PR or do you want to split platform PR and then MSTest implementation PR? If the latter, let's drop the class.

src/Platform/Microsoft.Testing.Platform/Capabilities/TestFramework/IStopGracefullyTestExecutionCapability.cs
Evangelink: I would suggest `IGracefulStopTestExecutionCapability` or `IGracefulStopTestExecutionRequestCapability`.

src/Platform/Microsoft.Testing.Platform/CommandLine/PlatformCommandLineProvider.cs
```diff
@@ -32,6 +32,7 @@ internal sealed class PlatformCommandLineProvider : ICommandLineOptionsProvider
     public const string TestHostControllerPIDOptionKey = "internal-testhostcontroller-pid";
     public const string ExitOnProcessExitOptionKey = "exit-on-process-exit";
     public const string ConfigFileOptionKey = "config-file";
+    public const string MaxFailedTestsOptionKey = "max-failed-tests";
```
Evangelink: Reading that, I recall we forgot to agree on either always having the feature on platform side and having potential failure if the option is enabled by user or if we say this is something the framework has to register (similar to services in VSTest bridge). 

If there are no technical issues with that, I think that's probably better. WDYT @MarcoRossignoli?

src/Platform/Microsoft.Testing.Platform/Extensions/AbortForMaxFailedTestsExtension.cs
```diff
@@ -0,0 +1,70 @@
...
+{
+    private readonly int? _maxFailedTests;
+    private readonly IStopGracefullyTestExecutionCapability? _capability;
+    private readonly PoliciesService _policiesService;
+    private readonly CancellationToken _cancellationToken;
+    private int _failCount;
+
+    public AbortForMaxFailedTestsExtension(ICommandLineOptions commandLineOptions, IStopGracefullyTestExecutionCapability? capability, PoliciesService policiesService, CancellationToken cancellationToken)
```
Evangelink: I know we only use the token and not the cancel but I'd be tempted to declare the `ITestApp...` instead of the token here. So that's clearler we plug to the global token.

src/Platform/Microsoft.Testing.Platform/OutputDevice/OutputDeviceManager.cs
Evangelink: See https://github.com/microsoft/testfx/issues/4256

src/Platform/Microsoft.Testing.Platform/Resources/PlatformResources.resx
```diff
@@ -694,4 +694,16 @@ Takes one argument as string in the format &lt;value&gt;[h|m|s] where 'value' is
...
+    <value>Specifies a maximum number of test failures that, when exceeded, will abort the test run.</value>
+  </data>
+  <data name="MaxFailedTestsMustBePositive" xml:space="preserve">
+    <value>The option '--max-failed-tests' must be a positive integer. The value '{0}' is not valid.</value>
+  </data>
+  <data name="AbortForMaxFailedTestsDescription" xml:space="preserve">
+    <value>Extension used to support '--max-failed-tests'. When a given failures threshold is reached, the test run will be aborted.</value>
+  </data>
+  <data name="ReachedMaxFailedTestsMessage" xml:space="preserve">
+    <value>Test session is aborting due to reaching failures specified by the '--max-failed-tests' option.</value>
```
Evangelink: Shall we specify max number tests here? Something like `Test session aborted: the maximum number of failed tests ({0}) has been reached.`

src/Platform/Microsoft.Testing.Platform/Services/CTRLPlusCCancellationTokenSource.cs
```diff
@@ -11,14 +11,18 @@ internal sealed class CTRLPlusCCancellationTokenSource : ITestApplicationCancell
...
     {
         if (console is not null)
         {
             console.CancelKeyPress += OnConsoleCancelKeyPressed;
         }
 
         _logger = logger;
+        if (policiesService is not null)
+        {
+            _cancellationTokenSource.Token.Register(() => policiesService.AbortPolicy.ExecuteCallbacksAsync(cancellationToken: default).GetAwaiter().GetResult());
```
Evangelink:
```suggestion
            _cancellationTokenSource.Token.Register(async () => await policiesService.AbortPolicy.ExecuteCallbacksAsync(cancellationToken: default));
```

src/Platform/Microsoft.Testing.Platform/Services/IPoliciesService.cs
```diff
@@ -0,0 +1,11 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+namespace Microsoft.Testing.Platform.Services;
+
+internal interface IPoliciesService
```
Evangelink:
```suggestion
internal interface IStopPoliciesService
```

src/Platform/Microsoft.Testing.Platform/Services/PoliciesService.cs
```diff
@@ -0,0 +1,42 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System.Collections.Concurrent;
+
+namespace Microsoft.Testing.Platform.Services;
+
+internal sealed class PoliciesService : IPoliciesService
+{
+    internal sealed class Policy
```
Evangelink: I would probably move this type up and have it part of the interface.

src/Platform/Microsoft.Testing.Platform/Services/PoliciesService.cs
```diff
@@ -0,0 +1,42 @@
...
+    {
+        private BlockingCollection<Func<CancellationToken, Task>>? _callbacks;
+
+        public bool IsPolicyTriggered { get; private set; }
```
Evangelink: I think `IsTriggered` is enough

test/UnitTests/Microsoft.Testing.Platform.UnitTests/Services/TestApplicationResultTests.cs
```diff
@@ -68,15 +68,20 @@ public async Task GetProcessExitCodeAsync_If_Failed_Tests_Returns_AtLeastOneTest
     public async Task GetProcessExitCodeAsync_If_Canceled_Returns_TestSessionAborted()
     {
         Mock<ITestApplicationCancellationTokenSource> testApplicationCancellationTokenSource = new();
+        var policiesService = new PoliciesService();
         testApplicationCancellationTokenSource.SetupGet(x => x.CancellationToken).Returns(() =>
         {
             CancellationTokenSource cancellationTokenSource = new();
+
+            // In reality, this is done by CTRLPlusCCancellationTokenSource (the real production impl of ITestApplicationCancellationTokenSource).
+            cancellationTokenSource.Token.Register(() => policiesService.AbortPolicy.ExecuteCallbacksAsync(cancellationToken: default).GetAwaiter().GetResult());
```
Evangelink:
```suggestion
            cancellationTokenSource.Token.Register(async () => await policiesService.AbortPolicy.ExecuteCallbacksAsync(cancellationToken: default));
```

src/Platform/Microsoft.Testing.Platform/Capabilities/TestFramework/IStopGracefullyTestExecutionCapability.cs
Youssef1313: I like `IGracefulStopTestExecutionCapability`

src/Adapter/MSTest.TestAdapter/TestingPlatformAdapter/MSTestStopGracefullyTestExecutionCapability.cs
Youssef1313: I think this PR, to at least try the design and see if it's good enough

src/Platform/Microsoft.Testing.Platform/Services/CTRLPlusCCancellationTokenSource.cs
Youssef1313: @Evangelink That will be fire-and-forget. CancellationToken.Register doesn't have a `Func<Task>` overload 😕

test/UnitTests/Microsoft.Testing.Platform.UnitTests/Services/TestApplicationResultTests.cs
Youssef1313: @Evangelink That will be fire-and-forget. CancellationToken.Register doesn't have a `Func<Task>` overload 😕

src/Platform/Microsoft.Testing.Platform/Services/CTRLPlusCCancellationTokenSource.cs
Youssef1313: effectively same as `async void`

src/Platform/Microsoft.Testing.Platform/Services/PoliciesService.cs
Youssef1313: Even if not used in the interface at all?

Youssef1313: Class is removed now completely :(

src/Platform/Microsoft.Testing.Platform/TestFramework/MaxFailedTestsCommandLineOptionsProvider.cs
```diff
@@ -0,0 +1,55 @@
...
+using Microsoft.Testing.Platform.Extensions.CommandLine;
+using Microsoft.Testing.Platform.Helpers;
+using Microsoft.Testing.Platform.Resources;
+
+namespace Microsoft.Testing.Platform.TestFramework;
+
+[Experimental("TPEXP", UrlFormat = "https://aka.ms/testingplatform/diagnostics#{0}")]
+public sealed class MaxFailedTestsCommandLineOptionsProvider : ICommandLineOptionsProvider
+{
+    // TODO: We have 'minimum-expected-tests', so should we use "maximum" instead of "max" here as well for consistency?
```
Evangelink: Good catch, yes!

src/Platform/Microsoft.Testing.Platform/TestFramework/MaxFailedTestsCommandLineOptionsProvider.cs
```diff
@@ -0,0 +1,55 @@
...
+using Microsoft.Testing.Platform.CommandLine;
+using Microsoft.Testing.Platform.Extensions;
+using Microsoft.Testing.Platform.Extensions.CommandLine;
+using Microsoft.Testing.Platform.Helpers;
+using Microsoft.Testing.Platform.Resources;
+
+namespace Microsoft.Testing.Platform.TestFramework;
+
+[Experimental("TPEXP", UrlFormat = "https://aka.ms/testingplatform/diagnostics#{0}")]
+public sealed class MaxFailedTestsCommandLineOptionsProvider : ICommandLineOptionsProvider
```
Evangelink: The class could probably remain internal as we are exposing the extension method that is public

src/Platform/Microsoft.Testing.Platform/Services/StopPoliciesService.cs
```diff
@@ -0,0 +1,90 @@
...
+    {
+        IsMaxFailedTestsTriggered = true;
+        if (_maxFailedTestsCallbacks is null)
+        {
+            return;
+        }
+
+        foreach (Func<int, CancellationToken, Task> callback in _maxFailedTestsCallbacks)
+        {
+            await callback.Invoke(maxFailedTests, cancellationToken);
```
Evangelink, line 50: As suggested by @MarcoRossignoli, we probably want to try/catch to execute the other callbacks in case one callback is failing.

src/Platform/Microsoft.Testing.Platform/Services/StopPoliciesService.cs
```diff
@@ -0,0 +1,90 @@
...
+        }
+
+        if (_abortCallbacks is null)
+        {
+            return;
+        }
+
+        foreach (Func<Task> callback in _abortCallbacks)
+        {
+            await callback.Invoke();
```
Evangelink, line 67: MaxFailedTestsCommandLineOptionsProvider

src/Platform/Microsoft.Testing.Platform/Extensions/AbortForMaxFailedTestsExtension.cs
MarcoRossignoli: Oh I see you're right never mind me, FYI everything inside the service bag will be disposed at the end.

src/Platform/Microsoft.Testing.Platform/Services/StopPoliciesService.cs
MarcoRossignoli: Or crash...we have to decide(I'm for crash for now)...in case of catch we need to show in some way to users that something went wrong.

src/Platform/Microsoft.Testing.Platform/OutputDevice/OutputDeviceManager.cs
MarcoRossignoli: I would cache a static instance. But not so important.

src/Platform/Microsoft.Testing.Platform/Services/StopPoliciesService.cs
```diff
@@ -0,0 +1,90 @@
...
+    }
+
+    private BlockingCollection<Func<int, CancellationToken, Task>>? _maxFailedTestsCallbacks;
+    private BlockingCollection<Func<Task>>? _abortCallbacks;
+
+    public bool IsMaxFailedTestsTriggered { get; private set; }
+
+    public bool IsAbortTriggered { get; private set; }
+
+    private static void RegisterCallback<T>(ref BlockingCollection<T>? callbacks, T callback)
```
Youssef1313, line 34: TODO: Discuss what we want to do in case this is called **after** the policy is already triggered.

src/Platform/Microsoft.Testing.Platform/TestFramework/MaxFailedTestsCommandLineOptionsProvider.cs
Youssef1313: Renamed.

src/Platform/Microsoft.Testing.Platform/TestFramework/MaxFailedTestsCommandLineOptionsProvider.cs
Youssef1313: Done

src/Platform/Microsoft.Testing.Platform/Services/StopPoliciesService.cs
Youssef1313: Not sure what you mean here.

src/Adapter/MSTest.TestAdapter/Execution/ClassCleanupManager.cs
```diff
@@ -62,4 +62,19 @@ public void MarkTestComplete(TestMethodInfo testMethodInfo, TestMethod testMetho
             ShouldRunEndOfAssemblyCleanup = _remainingTestsByClass.IsEmpty;
         }
     }
+
+    internal static void ForceCleanup(TypeCache typeCache)
+    {
+        IEnumerable<TestClassInfo> classInfoCache = typeCache.ClassInfoListWithExecutableCleanupMethods;
```
Evangelink, line 68: Putting a note for investigation: I am wondering if that would not trigger the cleanup of all test classes and not only the ones for which the init was called.

src/Adapter/MSTest.TestAdapter/Execution/TestExecutionManager.cs
```diff
@@ -399,6 +399,11 @@ private void ExecuteTestsInSource(IEnumerable<TestCase> tests, IRunContext? runC
             ExecuteTestsWithTestRunner(testsToRun, frameworkHandle, source, sourceLevelParameters, testRunner);
         }
 
+        if (MSTestGracefulStopTestExecutionCapability.Instance.IsStopRequested)
```
Evangelink, line 402: Do we still have any cleanup to process at this point? It feels like we are done with everything there.

src/Platform/Microsoft.Testing.Platform/CommandLine/MaxFailedTestsCommandLineOptionsProvider.cs
```diff
@@ -0,0 +1,53 @@
...
+        => ValidationResult.ValidTask;
+
+    public Task<ValidationResult> ValidateOptionArgumentsAsync(CommandLineOption commandOption, string[] arguments)
+    {
+        if (commandOption.Name == MaxFailedTestsOptionKey)
+        {
+            string arg = arguments[0];
+            // We consider --maximum-failed-tests 0 as valid.
+            // The idea is that we stop the execution when we *exceed* the max failed tests, not when *reach*.
+            // So zero means, stop execution on the first failure.
```
Evangelink: I would prefer the value to match so that `--maximum-failed-tests 0` is invalid and `--maximum-failed-tests 1` fails when receiving the first failed test.

Looking up at some other tools (for different languages) this seems to be matching their behavior.

src/Platform/Microsoft.Testing.Platform/Extensions/AbortForMaxFailedTestsExtension.cs
```diff
@@ -0,0 +1,76 @@
...
+    {
+        var node = (TestNodeUpdateMessage)value;
+
+        // If we are called, the extension is enabled, which means both _maxFailedTests and capability are not null.
+        RoslynDebug.Assert(_maxFailedTests is not null);
+        RoslynDebug.Assert(_capability is not null);
+
+        TestNodeStateProperty testNodeStateProperty = node.TestNode.Properties.Single<TestNodeStateProperty>();
+        if (TestNodePropertiesCategories.WellKnownTestNodeTestRunOutcomeFailedProperties.Any(t => t == testNodeStateProperty.GetType()) &&
+            ++_failCount > _maxFailedTests.Value &&
```
Evangelink: `>=` to match the previous comment on having the number matching to received and not exceeded.

src/Platform/Microsoft.Testing.Platform/Hosts/TestHostBuilder.cs
```diff
@@ -724,6 +734,17 @@ private async Task<ITestFramework> BuildTestFrameworkAsync(TestFrameworkBuilderD
...
+        var abortForMaxFailedTestsExtension = new AbortForMaxFailedTestsExtension(
+            serviceProvider.GetCommandLineOptions(),
+            serviceProvider.GetTestFrameworkCapabilities().GetCapability<IGracefulStopTestExecutionCapability>(),
+            serviceProvider.GetRequiredService<IStopPoliciesService>(),
+            serviceProvider.GetTestApplicationCancellationTokenSource());
+
+        if (await abortForMaxFailedTestsExtension.IsEnabledAsync())
+        {
+            dataConsumersBuilder.Add(abortForMaxFailedTestsExtension);
+        }
```
Evangelink, line 746: I think it's better to use `RegisterAsServiceOrConsumerOrBothAsync` here.

src/Platform/Microsoft.Testing.Platform/Services/StopPoliciesService.cs
Evangelink: Let's do like CancellationToken and trigger immediatly.

src/Platform/Microsoft.Testing.Platform/Services/StopPoliciesService.cs
Evangelink: It was meant to be copy-paste from the previous comment but something went wrong, sorry!

Evangelink (approved):
A few minor updates but looking great overall! Nice job here @Youssef1313

src/Adapter/MSTest.TestAdapter/Execution/TestExecutionManager.cs
Youssef1313: We are not done with class cleanups or assembly cleanups at this point. ForceCleanup does take care of that.

src/Adapter/MSTest.TestAdapter/Execution/ClassCleanupManager.cs
Youssef1313: If init wasn't called, then we will not call cleanup. This is handled in ExecuteClassCleanup by this piece of code:

```
if (IsClassCleanupExecuted
    // If there is a ClassInitialize method and it has not been executed, then we should not execute ClassCleanup
    || (!IsClassInitializeExecuted && ClassInitializeMethod is not null))
{
    return;
}
```

However, thinking now more, it's curious what will happen if there is only class cleanup but there is no class initialize.
Generally, if the class didn't yet start executing, it's not in cache, but it's definitely not a good idea to rely on that behavior.

The logic I mentioned above (which is already existing) feels somewhat suspicious to me as well in case the class doesn't define ClassCleanup, but base class does.
It doesn't make sense we check whether ClassInitializeMethod is null or not without also a corresponding check for base classes.

src/Platform/Microsoft.Testing.Platform/Hosts/TestHostBuilder.cs
Youssef1313: @Evangelink Do we want to register it in ServiceProvider though?

src/Adapter/MSTest.TestAdapter/Execution/ClassCleanupManager.cs
Youssef1313: Adjusted the logic there a bit. It should be better now.

src/Platform/Microsoft.Testing.Platform/Hosts/TestHostBuilder.cs
Evangelink: I see, it's probably not changing much as it's internal but it might help some other extension points we would build. I am fine if we don't.

@MarcoRossignoli any opinion here?

src/Platform/Microsoft.Testing.Platform/Services/StopPoliciesService.cs
Evangelink: Done on the public register methods.

Evangelink (approved):
Awesome work here @Youssef1313. Let's merge and do a potential follow-up for https://github.com/microsoft/testfx/pull/4238/files#r1878337032

src/Platform/Microsoft.Testing.Platform/Hosts/TestHostBuilder.cs
MarcoRossignoli: I'm fine with the current code, I prefer to be specific and not have it as service as "internal" feature.

PR #4249, opened 2024-12-06
Support 'TestPropertyAttribute' on test classes

### LINKED ISSUE #4248: Allow `[TestProperty]` to be set on a class
## Summary

Update `[TestProperty]` to allow usage on test classes.

## Background and Motivation

In xUnit, the `[Trait]` can be set on classes and with NUnit `[Property]` can also be set on classes. By doing this change, we would make it easier for people converting from NUnit and xUnit while allowing grouping of traits.

## Proposed Feature

1. Update the attribute to allow usage on classes.
2. Update related code analyzer to allow usage.

Youssef1313:
Fixes #4248

### DIFF
--- a/src/Adapter/MSTest.TestAdapter/Execution/TypeCache.cs
+++ b/src/Adapter/MSTest.TestAdapter/Execution/TypeCache.cs
@@ -911,17 +911,17 @@ internal class TypeCache : MarshalByRefObject
     /// </summary>
     /// <param name="testMethodInfo"> The test Method Info. </param>
     /// <param name="testContext"> The test Context. </param>
-    private static void SetCustomProperties(TestMethodInfo testMethodInfo, ITestContext testContext)
+    private void SetCustomProperties(TestMethodInfo testMethodInfo, ITestContext testContext)
     {
         DebugEx.Assert(testMethodInfo != null, "testMethodInfo is Null");
         DebugEx.Assert(testMethodInfo.TestMethod != null, "testMethodInfo.TestMethod is Null");
 
-        IEnumerable<TestPropertyAttribute> attributes = testMethodInfo.TestMethod.GetCustomAttributes<TestPropertyAttribute>(inherit: false);
+        IEnumerable<TestPropertyAttribute> attributes = _reflectionHelper.GetDerivedAttributes<TestPropertyAttribute>(testMethodInfo.TestMethod, inherit: true);
         DebugEx.Assert(attributes != null, "attributes is null");
 
         if (testMethodInfo.TestMethod.DeclaringType is { } testClass)
         {
-            attributes = attributes.Concat(testClass.GetCustomAttributes<TestPropertyAttribute>(inherit: false));
+            attributes = attributes.Concat(_reflectionHelper.GetDerivedAttributes<TestPropertyAttribute>(testClass, inherit: true));
         }
 
         foreach (TestPropertyAttribute attribute in attributes)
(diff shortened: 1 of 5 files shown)


src/Adapter/MSTest.TestAdapter/Execution/TypeCache.cs
```diff
@@ -916,10 +916,15 @@ private static void SetCustomProperties(TestMethodInfo testMethodInfo, ITestCont
         DebugEx.Assert(testMethodInfo != null, "testMethodInfo is Null");
         DebugEx.Assert(testMethodInfo.TestMethod != null, "testMethodInfo.TestMethod is Null");
 
-        object[] attributes = testMethodInfo.TestMethod.GetCustomAttributes(typeof(TestPropertyAttribute), false);
+        IEnumerable<TestPropertyAttribute> attributes = testMethodInfo.TestMethod.GetCustomAttributes<TestPropertyAttribute>(inherit: false);
         DebugEx.Assert(attributes != null, "attributes is null");
 
-        foreach (TestPropertyAttribute attribute in attributes.Cast<TestPropertyAttribute>())
+        if (testMethodInfo.TestMethod.DeclaringType is { } testClass)
```
Evangelink, line 922: We may want to look at parent classes (or use inherit - if that works)

nohwnd: maybe the helpers from reflection operations should be use here, they encode getting inherited / non-inherited, and deriving / non-deriving attributes in a standard way.

Youssef1313: The current behavior for methods is that we do **not** inherit. Do we want to change that as well?

Inheritance for methods basically means that if the test method is an **override**, do we get the attributes from the **overridden** method(s) or not.

To me, I think yes we should inherit in both cases (on class and on method)

Evangelink: I am not sure what to do for test methods. I would maybe avoid to change. 

I am wondering if we should also add an analyzer that guards around overrides of a TestMethod. I can't think of a scenario where this would make sense (@nohwnd can you think of any?)

Youssef1313: https://github.com/dotnet/runtime/blob/ad304792cfd5af4155120f054b48850bf91acf62/src/libraries/System.Text.Json/tests/System.Text.Json.SourceGeneration.Tests/RealWorldContextTests.cs :)

Youssef1313: and https://github.com/dotnet/roslyn/blob/aea1c90d8a651112b646c3c8a389d1293d246ddd/src/Compilers/Core/CodeAnalysisTest/Collections/ImmutableSegmentedDictionaryTest.nonnetstandard.cs#L19-L24

nohwnd: I must be blind I don't see overrides of test methods in the linked files. I think there might be potentially some usage, but  I don't have a good example where it would be necessary to do it. It makes the most sense to me when you write unit tests for methods that themselves are overriden. So you can keep the same names of tests and just change the parameters and data.

Youssef1313: The example from dotnet/runtime declares test methods as virtual, and some classes can override it, e.g: https://github.com/dotnet/runtime/blob/731a96b0018cda0c67bb3fab7a4d06538fe2ead5/src/libraries/System.Text.Json/tests/System.Text.Json.SourceGeneration.Tests/SerializationContextTests.cs#L240

The example from roslyn points out to an override test method right away.

(and yes they are not MSTest but shows that the concept isn't uncommon)

Evangelink: @nohwnd I do see them. I have read the code in diagonale and I am not sure to really understand why it's needed but that's a different story.

Evangelink (approved):

PR #3511, opened 2024-08-09
avoid GetTypeInfo

SimonCropp:

PR #4263, opened 2024-12-07
More efficient reflection

Youssef1313:
These changes seems to be more efficient, even if not super significant, but also will help when https://github.com/dotnet/runtime/issues/53217 is implemented (this analyzer is already approved). See also https://github.com/dotnet/runtime/pull/89317 which now hides GetTypeInfo for .NET 10

### DIFF
--- a/src/Adapter/MSTest.TestAdapter/Discovery/TypeValidator.cs
+++ b/src/Adapter/MSTest.TestAdapter/Discovery/TypeValidator.cs
@@ -106,7 +106,7 @@ internal class TypeValidator
     {
         DebugEx.Assert(type != null, "HasCorrectTestContextSignature type is null");
 
-        IEnumerable<PropertyInfo> propertyInfoEnumerable = PlatformServiceProvider.Instance.ReflectionOperations.GetDeclaredProperties(type);
+        PropertyInfo[] propertyInfoEnumerable = PlatformServiceProvider.Instance.ReflectionOperations.GetDeclaredProperties(type);
         var propertyInfo = new List<PropertyInfo>();
 
         foreach (PropertyInfo pinfo in propertyInfoEnumerable)
--- a/src/Adapter/MSTest.TestAdapter/Execution/TypeCache.cs
+++ b/src/Adapter/MSTest.TestAdapter/Execution/TypeCache.cs
@@ -399,7 +399,7 @@ internal sealed class TypeCache : MarshalByRefObject
 
         assemblyInfo = new TestAssemblyInfo(assembly);
 
-        IReadOnlyList<Type> types = AssemblyEnumerator.GetTypes(assembly, assembly.FullName!, null);
+        Type[] types = AssemblyEnumerator.GetTypes(assembly, assembly.FullName!, null);
 
         foreach (Type t in types)
         {
--- a/src/Adapter/MSTest.TestAdapter/Execution/TypeCache.cs
+++ b/src/Adapter/MSTest.TestAdapter/Execution/TypeCache.cs
@@ -259,7 +259,7 @@ internal sealed class TypeCache : MarshalByRefObject
     /// <returns> The <see cref="TestClassInfo"/>. </returns>
     private TestClassInfo CreateClassInfo(Type classType, TestMethod testMethod)
     {
-        IEnumerable<ConstructorInfo> constructors = PlatformServiceProvider.Instance.ReflectionOperations.GetDeclaredConstructors(classType);
+        ConstructorInfo[] constructors = PlatformServiceProvider.Instance.ReflectionOperations.GetDeclaredConstructors(classType);
         (ConstructorInfo CtorInfo, bool IsParameterless)? selectedConstructor = null;
 
         foreach (ConstructorInfo ctor in constructors)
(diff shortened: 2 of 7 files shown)


src/Adapter/MSTest.TestAdapter/Discovery/TypeValidator.cs
```diff
@@ -106,7 +106,7 @@ internal static bool HasCorrectTestContextSignature(Type type)
     {
         DebugEx.Assert(type != null, "HasCorrectTestContextSignature type is null");
 
-        IEnumerable<PropertyInfo> propertyInfoEnumerable = PlatformServiceProvider.Instance.ReflectionOperations.GetDeclaredProperties(type);
+        PropertyInfo[] propertyInfoEnumerable = PlatformServiceProvider.Instance.ReflectionOperations.GetDeclaredProperties(type);
```
Youssef1313, line 109: Now that the type is strongly-typed as array, this should avoid an allocation for an enumerator.

src/Adapter/MSTest.TestAdapter/Execution/TypeCache.cs
```diff
@@ -259,7 +259,7 @@ private static bool TryGetUnescapedManagedTypeName(TestMethod testMethod, [NotNu
     /// <returns> The <see cref="TestClassInfo"/>. </returns>
     private TestClassInfo CreateClassInfo(Type classType, TestMethod testMethod)
     {
-        IEnumerable<ConstructorInfo> constructors = PlatformServiceProvider.Instance.ReflectionOperations.GetDeclaredConstructors(classType);
+        ConstructorInfo[] constructors = PlatformServiceProvider.Instance.ReflectionOperations.GetDeclaredConstructors(classType);
```
Youssef1313, line 262: Same here about enumerator allocation.

src/Adapter/MSTest.TestAdapter/SourceGeneration/SourceGeneratedReflectionOperations.cs
```diff
@@ -68,7 +68,7 @@ public IEnumerable<PropertyInfo> GetDeclaredProperties(Type type)
     public Type[] GetDefinedTypes(Assembly assembly)
         => ReflectionDataProvider.Types;
 
-    public IEnumerable<MethodInfo> GetRuntimeMethods(Type type)
+    public MethodInfo[] GetRuntimeMethods(Type type)
         => ReflectionDataProvider.TypeMethods[type];
```
Youssef1313, line 72: The return type here was already an array, but making at strongly-typed allows callsites to not allocate an enumerator and be more efficient.

src/Adapter/MSTestAdapter.PlatformServices/Services/ReflectionOperations2.cs
```diff
@@ -23,26 +26,26 @@ public ReflectionOperations2()
 #pragma warning disable IL2026 // Members attributed with RequiresUnreferencedCode may break when trimming
 #pragma warning disable IL2067 // 'target parameter' argument does not satisfy 'DynamicallyAccessedMembersAttribute' in call to 'target method'.
 #pragma warning disable IL2057 // Unrecognized value passed to the typeName parameter of 'System.Type.GetType(String)'
-    public IEnumerable<ConstructorInfo> GetDeclaredConstructors(Type classType)
-        => classType.GetTypeInfo().DeclaredConstructors;
+    public ConstructorInfo[] GetDeclaredConstructors(Type classType)
+        => classType.GetConstructors(DeclaredOnlyLookup);
```
Youssef1313, line 30: `GetTypeInfo` has an extra unnecessary allocation, and forces us to use IEnumerable which will cause an enumerator allocation at callsites as well. The same applies everywhere below.

src/Adapter/MSTestAdapter.PlatformServices/Services/ReflectionOperations2.cs
```diff
@@ -23,26 +26,26 @@ public ReflectionOperations2()
...
+    public PropertyInfo[] GetDeclaredProperties(Type type)
+        => type.GetProperties(DeclaredOnlyLookup);
 
     public PropertyInfo? GetDeclaredProperty(Type type, string propertyName)
-        => type.GetTypeInfo().GetDeclaredProperty(propertyName);
+        => type.GetProperty(propertyName, DeclaredOnlyLookup);
 
     public Type[] GetDefinedTypes(Assembly assembly)
-        => assembly.DefinedTypes.ToArray();
+        => assembly.GetTypes();
```
Youssef1313, line 45: `DefinedTypes` alone would internally allocate two arrays, and then we do `ToArray` again which will allocate a third array.

`GetTypes()` seems to be much more efficient. I assume it doesn't introduce any behavior changes.

Evangelink (approved):
I was going to say that I think there is a change of behavior and that we had some issue with a similar change from @SimonCropp but I don't see any failed test so I guess I remember wrong or something changed.

Let's move on and see :)

Youssef1313:
@Evangelink Looks like you are referring to https://github.com/microsoft/testfx/pull/3511. The problem there wasn't related to which reflection methods are used and was fixed.

PR #4269, opened 2024-12-07
Allow to disable test expansion on implementations of ITestDataSource

### LINKED ISSUE #4255: Add ability to disable expansion on `[DynamicData]`
## Summary

Add ability to disable expansion on `[DynamicData]`.

## Background and Motivation

MSTest is globally configured to expand parameterized tests during discovery to enhance user experience but sometimes the parameterized types are not serializable leading to warnings or errors. As of now, users can disable the expansion altogether using the assembly attribute `[assembly: TestDataSourceDiscoveryAttribute(TestDataSourceDiscoveryOption.DuringExecution)]` but this is forcing all parameterized tests not to be expanded.

## Proposed Feature

While waiting for us to get rid of the serialization in v4, it would be good to introduce an option on each `[DynamicData]` (settable property) defaulting to `DuringDiscovery` but that can be set to `DuringExecution` that would allow to have this unique data source not to be expanded.

Evangelink:
Fixes #4255

Note that I decided to not allow `[DataRow]` to disable test expansion because AFAIK we don't have issue there but that's arbitrary call and I don't mind changing it.


Youssef1313:
> Note that I decided to not allow `[DataRow]` to disable test expansion because AFAIK we don't have issue there but that's arbitrary call and I don't mind changing it.

There are issues with DataRow currently. If you do this:

```csharp
[DataRow(new object[] { (byte)0 })]
[DataRow(new object[] { (short)0 })]
```

the types byte and short are lost IIRC and you end up with int. This happens even though we have:

https://github.com/microsoft/testfx/blob/e101a9d48773cc935c7b536d25d378d9a3211fee/src/Adapter/MSTest.TestAdapter/Helpers/DataSerializationHelper.cs#L16

src/Adapter/MSTest.TestAdapter/Discovery/AssemblyEnumerator.cs
```diff
@@ -411,6 +411,16 @@ private static bool ProcessITestDataSourceTests(UnitTestElement test, Reflection
     {
         foreach (ITestDataSource dataSource in testDataSources)
         {
+            // When reaching this point the global setup is to expand the parameterized tests but each data source
+            // could decide not to expand so we need to respect it.
+            if (dataSource is IExpandableDataSource { ExpandDataSource: false })
```
Youssef1313: Does it go the other way around? On assembly level it's set to not expand, but specific data sources can expand? Reading the code, I don't think the current implementation handles it the opposite way. But the API shape gives the impression that it would work.

src/TestFramework/TestFramework/Attributes/DataSource/DynamicDataAttribute.cs
```diff
@@ -84,6 +84,12 @@ public DynamicDataAttribute(string dynamicDataSourceName, Type dynamicDataDeclar
     /// </summary>
     public Type? DynamicDataDisplayNameDeclaringType { get; set; }
 
+    /// <summary>
+    /// Gets or sets a value indicating whether the data source should be expanded, allowing to treat each data row as a separate test.
+    /// The default value is <c>true</c>.
+    /// </summary>
+    public bool ExpandDataSource { get; set; } = true;
```
Youssef1313: My feeling is for this to be `bool?`, and it means:

- null: Use what is specified by the assembly attribute and default to DuringDiscovery when the attribute is not present.
- true: force DuringDiscovery, even if the assembly attribute is set to DuringExecution
- false: force DuringExecution, even if the assembly attribute is set to DuringDiscovery

src/Adapter/MSTest.TestAdapter/Discovery/AssemblyEnumerator.cs
Youssef1313: Actually for that to work both ways, it needs to be nullable bool. Added another comment on that.

test/IntegrationTests/MSTest.IntegrationTests/Parameterized tests/DataExtensibilityTests.cs
```diff
@@ -94,4 +97,24 @@ public void ExecuteCustomTestExtensibilityWithTestDataTests()
...
+
+        // Assert
+        Verify(testCases.Length == 1);
+
+        VerifyE2E.TestsPassed(
+            testResults,
+            "CustomDisableExpansionTestDataSourceTestMethod1 (1,2,3)",
+            "CustomDisableExpansionTestDataSourceTestMethod1 (4,5,6)");
+
+        VerifyE2E.TestsFailed(testResults);
```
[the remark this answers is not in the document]
Evangelink: Incorrect understanding, this assert there is no failed tests

src/Adapter/MSTest.TestAdapter/Discovery/AssemblyEnumerator.cs
Evangelink: You are right, I haven't considered the other way around. I'll think whether to rename to specify this is only to disable or expand behavior (there is probably some performance impact).

src/TestFramework/TestFramework/Attributes/DataSource/DynamicDataAttribute.cs
Evangelink: Does it bring any advantage to consider the null state over simple true/false?

Youssef1313: @Evangelink If we are going to consider it in both ways (disable and enable), then there is no way for simple true/false to work, I think?

Reading the value as true will mean "use DuringDiscover", and reading the value as false will mean "use DuringExecution". But then nothing will mean "use whatever specified by the assembly attribute".

Evangelink: Ah right, good point! I'll try to do the rework tonight

src/Adapter/MSTest.TestAdapter/Discovery/AssemblyEnumerator.cs
```diff
@@ -387,16 +355,36 @@ static UnitTestElement GetFixtureTest(string classFullName, string assemblyLocat
...
+        // Since we depend on those properties, if they don't exist, we bail out early.
+        if (!test.TestMethod.HasManagedMethodAndTypeProperties)
+        {
+            return false;
+        }
+
         // We don't have a special method to filter attributes that are not derived from Attribute, so we take all
         // attributes and filter them. We don't have to care if there is one, because this method is only entered when
         // there is at least one (we determine this in TypeEnumerator.GetTestFromMethod.
         IEnumerable<ITestDataSource> testDataSources = ReflectHelper.Instance.GetDerivedAttributes<Attribute>(testMethodInfo.MethodInfo, inherit: false).OfType<ITestDataSource>();
```
Youssef1313, line 367: Feel free to cleanup in a follow-up as it's not directly related to your changes

```suggestion
        IEnumerable<ITestDataSource> testDataSources = ReflectHelper.Instance.GetDerivedAttributes<ITestDataSource>(testMethodInfo.MethodInfo, inherit: false);
```

src/Adapter/MSTest.TestAdapter/Discovery/AssemblyEnumerator.cs
```diff
@@ -227,25 +237,23 @@ private List<UnitTestElement> DiscoverTestsInType(
...
                         // Add fixture tests like AssemblyInitialize, AssemblyCleanup, ClassInitialize, ClassCleanup.
-                        if (MSTestSettings.CurrentSettings.ConsiderFixturesAsSpecialTests && testMethodInfo.Value is not null)
+                        if (MSTestSettings.CurrentSettings.ConsiderFixturesAsSpecialTests)
                         {
-                            AddFixtureTests(testMethodInfo.Value, tests, fixturesTests);
+                            AddFixtureTests(testMethodInfo, tests, fixturesTests);
                         }
 
-                        if (DynamicDataAttached(test, testMethodInfo, tests))
+                        if (TryUnfoldITestDataSources(test, testMethodInfo, dataSourcesUnfoldingStrategy, tests))
```
Youssef1313, line 253: Here, if `TryUnfoldITestDataSources` returned false, we will add the "folded"/"unexpanded" test to the `tests` list. But in `TryUnfoldITestDataSource`, one data source could have been unfolded already.

src/Adapter/MSTest.TestAdapter/Helpers/ReflectHelper.cs
```diff
@@ -278,11 +278,22 @@ internal static TestIdGenerationStrategy GetTestIdGenerationStrategy(Assembly as
     /// Gets TestDataSourceDiscovery assembly level attribute.
     /// </summary>
     /// <param name="assembly"> The test assembly. </param>
+    [Obsolete]
```
Youssef1313, line 281: It should still be good to use the method (and we do use it) to still support the old attribute. Not sure why obsoleting is necessary/good.

src/TestFramework/TestFramework/Interfaces/ITestDataSourceUnfoldingCapability.cs
```diff
@@ -0,0 +1,36 @@
...
+
+/// <summary>
+/// Specifies how parameterized tests should be executed, either as individual test cases for each data row or as a
+/// single test case. This affects the test results and the UI representation of the tests.
+/// </summary>
+public enum TestDataSourceUnfoldingStrategy : byte
+{
+    /// <summary>
+    /// The application decides whether to unfold the parameterized test.
+    /// </summary>
```
Youssef1313: The doc can be expanded a bit for clarity I think. Something that mentions the assembly-level attribute(s), and the default behavior when there is no assembly-level attribute.

test/IntegrationTests/TestAssets/DynamicDataTestProject/DisableExpansionTests.cs
```diff
@@ -0,0 +1,60 @@
...
+    public void TestMethodSourceOnDifferentType(int a, string s)
+    {
+    }
+
+    [TestMethod]
+    [DynamicData(nameof(PropertySource), UnfoldingStrategy = TestDataSourceUnfoldingStrategy.Fold)]
+    [DynamicData(nameof(PropertySource), typeof(DataSourceHelper))]
+    public void TestPropertyWithTwoSourcesButOnlyOneDisablingExpansion(int a, string s)
+    {
+    }
```
Youssef1313: We should have the same test but with the order of attributes swapped (i.e, the first attribute is unfolding and the second is folding)

src/Adapter/MSTest.TestAdapter/Helpers/ReflectHelper.cs
Evangelink: Type is obsolete so it forces me to suppress the obsolete. It feels easiers and clearer to have the method also obsolete. It will also help with cleanup IMO when we are finally getting to v4.

src/TestFramework/TestFramework/Interfaces/ITestDataSourceUnfoldingCapability.cs
```diff
@@ -0,0 +1,36 @@
...
+}
+
+/// <summary>
+/// Specifies how parameterized tests should be executed, either as individual test cases for each data row or as a
+/// single test case. This affects the test results and the UI representation of the tests.
+/// </summary>
+public enum TestDataSourceUnfoldingStrategy : byte
+{
+    /// <summary>
+    /// The application decides whether to unfold the parameterized test.
```
Evangelink:
```suggestion
    /// MSTest will decide whether to unfold the parameterized test based on value from the assembly level attribute <see cref="TestDataSourceOptionsAttribute" />. If no assembly level attribute is specified, then the default configuration is to unfold.
```

WDYT @Youssef1313?

src/Adapter/MSTest.TestAdapter/Discovery/AssemblyEnumerator.cs
Evangelink: Fixed

src/Adapter/MSTest.TestAdapter/Discovery/AssemblyEnumerator.cs
```diff
@@ -384,100 +352,142 @@ static UnitTestElement GetFixtureTest(string classFullName, string assemblyLocat
...
         try
         {
-            return ProcessITestDataSourceTests(test, new(testMethodInfo.MethodInfo, test.DisplayName), testDataSources, tests);
+            bool isDataDriven = false;
+            foreach (ITestDataSource dataSource in testDataSources)
+            {
+                isDataDriven = true;
+                if (!TryUnfoldITestDataSource(dataSource, dataSourcesUnfoldingStrategy, test, new(testMethodInfo.MethodInfo, test.DisplayName), tempListOfTests))
+                {
+                    // TODO: Improve multi-source design!
```
Youssef1313, line 380: Is it too much work to have each data source treated individually? Some can fold while others not?

Evangelink: Yes as it would force me to have some kind of identifier of the data source so that I know the ones to exclude and there is no good way to do that for now.

Youssef1313 (dismissed):

src/Adapter/MSTest.TestAdapter/Discovery/AssemblyEnumerator.cs
Evangelink: see #4322

src/Adapter/MSTest.TestAdapter/Discovery/AssemblyEnumerator.cs
Evangelink: Not working there is a constraint of type to be attribute.

PR #4270, opened 2024-12-08
[Performance] Don't parse the same runsettings for each class in the assembly

Youssef1313:
It's easiest to review each commit on its own. The second commit is a pure refactoring to remove `RunSettingsXml` as it can be passed from GetTestsInIsolation to EnumerateAssembly and it makes the code easier and more simpler to follow.

This change was motivated by:

image

### DIFF
--- a/src/Adapter/MSTest.TestAdapter/Discovery/AssemblyEnumerator.cs
+++ b/src/Adapter/MSTest.TestAdapter/Discovery/AssemblyEnumerator.cs
@@ -51,11 +51,6 @@ internal class AssemblyEnumerator : MarshalByRefObject
         MSTestSettings.PopulateSettings(settings);
 
     /// <summary>
-    /// Gets or sets the run settings to use for current discovery session.
-    /// </summary>
-    public string? RunSettingsXml { get; set; }
-
-    /// <summary>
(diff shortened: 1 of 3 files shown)


src/Adapter/MSTest.TestAdapter/Discovery/AssemblyEnumerator.cs
```diff
@@ -215,7 +217,7 @@ private List<UnitTestElement> DiscoverTestsInType(
         HashSet<string> fixturesTests)
     {
         IDictionary<string, object> tempSourceLevelParameters = PlatformServiceProvider.Instance.SettingsProvider.GetProperties(assemblyFileName);
-        tempSourceLevelParameters = RunSettingsUtilities.GetTestRunParameters(runSettingsXml)?.ConcatWithOverwrites(tempSourceLevelParameters)
+        tempSourceLevelParameters = testRunParametersFromRunSettings?.ConcatWithOverwrites(tempSourceLevelParameters)
```
Youssef1313, line 219: Here, `RunSettingsUtilities.GetTestRunParameters(runSettingsXml)` was part of `DiscoverTestsInType` which is called for every type in the assembly.

Evangelink (approved):

PR #4271, opened 2024-12-08
[Performance] Ensure ReflectionOperations return `Attribute[]` when possible

Youssef1313:
Motivation:

image

### DIFF
--- a/src/Adapter/MSTestAdapter.PlatformServices/Services/ReflectionOperations.cs
+++ b/src/Adapter/MSTestAdapter.PlatformServices/Services/ReflectionOperations.cs
@@ -1,6 +1,7 @@
 ﻿// Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
+using System.Diagnostics;
 using System.Diagnostics.CodeAnalysis;
 using System.Reflection;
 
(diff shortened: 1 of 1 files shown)


src/Adapter/MSTestAdapter.PlatformServices/Services/ReflectionOperations.cs
```diff
@@ -60,6 +72,6 @@ public object[] GetCustomAttributes(Assembly assembly, Type type) =>
 #if NETFRAMEWORK
         ReflectionUtility.GetCustomAttributes(assembly, type).ToArray();
 #else
-        assembly.GetCustomAttributes(type).ToArray<object>();
```
Evangelink, line 63: Isn't this line changing behavior by setting inherit to true?

Evangelink (commented):
Would be good to have a before/after on the allocation screenshot

src/Adapter/MSTestAdapter.PlatformServices/Services/ReflectionOperations.cs
Youssef1313: That is the default already

https://github.com/dotnet/runtime/blob/2d81cbb28a8feb1af5b790664b511b264d032877/src/coreclr/System.Private.CoreLib/src/System/Attribute.CoreCLR.cs#L697-L701

Youssef1313: And actually, for assemblies specifically, this parameter has no meaning and is unused.

https://github.com/dotnet/runtime/blob/2d81cbb28a8feb1af5b790664b511b264d032877/src/coreclr/System.Private.CoreLib/src/System/Reflection/RuntimeAssembly.cs#L332-L340

Evangelink (approved):

PR #4272, opened 2024-12-08
[Performance] Avoid System.Action allocations for test context cancellation

Youssef1313:
image


Evangelink (approved):
It'd be good to have some before/after screenshot

PR #4279, opened 2024-12-08
Refactor GetTestFromMethod

Youssef1313:

### DIFF
--- a/src/Adapter/MSTest.TestAdapter/Discovery/TypeEnumerator.cs
+++ b/src/Adapter/MSTest.TestAdapter/Discovery/TypeEnumerator.cs
@@ -201,29 +201,39 @@ internal class TypeEnumerator
 
         testElement.Traits = traits.ToArray();
 
-        if (_reflectHelper.GetFirstDerivedAttributeOrDefault<CssIterationAttribute>(method, inherit: true) is CssIterationAttribute cssIteration)
+        Attribute[] attributes = _reflectHelper.GetCustomAttributesCached(method, inherit: true);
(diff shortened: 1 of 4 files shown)


src/Adapter/MSTest.TestAdapter/Discovery/TypeEnumerator.cs
```diff
@@ -201,29 +201,39 @@ internal UnitTestElement GetTestFromMethod(MethodInfo method, bool isDeclaredInT
 
         testElement.Traits = traits.ToArray();
 
-        if (_reflectHelper.GetFirstDerivedAttributeOrDefault<CssIterationAttribute>(method, inherit: true) is CssIterationAttribute cssIteration)
-        {
-            testElement.CssIteration = cssIteration.CssIteration;
-        }
+        Attribute[] attributes = _reflectHelper.GetCustomAttributesCached(method, inherit: true);
```
Youssef1313, line 204: @Evangelink This one is hypothetical and not based on any profiling.

My idea there is that we were doing `GetFirstDerivedAttributeOrDefault` multiple times. Each will internally call GetCustomAttributesCached and loop over of the result. While it's already cached, and should be fast enough, in theory there is still a bit of overhead for the dictionary lookup and for the multiple type checks.

This refactoring reduces the number of dictionary lookups as well as the number of type checks. However, it may not have much of an impact 😕 

Feel free to close the PR if the change isn't convincing enough.

Evangelink (approved):

Evangelink (dismissed):

PR #4280, opened 2024-12-08
Simplify AttributeComparer.IsDerived

Youssef1313:


Evangelink (approved):
That seems like an equivalent change but I hope we have enough tests there.

PR #4284, opened 2024-12-08
fix nullability of test hierarchy

SimonCropp:


Evangelink (commented):
I am not sure what's best to do here. At the moment, the hierarchy is a Test Explorer concept and we expect to have the following 4 levels:
- 0: assembly
- 1: namespace
- 2: class
- 3: method

we are allowed to nullify the assembly to let Test Explorer set it (to have the tfm associated for example) but the other entries are meant to be non-nullable. 

Given that, I think it might be easier to keep the array non-nullable and to "force null" on the first item than to give a chance to have them all nullable. But I am open to suggestions. Note that ideally, this should be wrapped in a type with 4 properties as explained above which would make things easier (allowing to have only assembly nullable).

SimonCropp:
is it possible to make this a string type structure? instead of an array?

Evangelink:
> is it possible to make this a strong typed structure? instead of an array?

On MSTest side yes. On Test Explorer/VSTest, probably harder to no go.

SimonCropp:
ok. then based on your comments, i think we abandon this?

PR #4285, opened 2024-12-08
cleanup SourceGeneratedFileOperations

SimonCropp:
* remove redundant casts
 * make ReflectionDataProvider `null!` and add comment as to why. better than the `warning disable CS8618` that gives no context
 * fix nullability of return values
 * remove null suppression for `ReflectionDataProvider` usage

### DIFF
--- a/src/Adapter/MSTest.TestAdapter/SourceGeneration/SourceGeneratedFileOperations.cs
+++ b/src/Adapter/MSTest.TestAdapter/SourceGeneration/SourceGeneratedFileOperations.cs
@@ -7,8 +7,6 @@ using System.Reflection;
 using Microsoft.VisualStudio.TestPlatform.MSTestAdapter.PlatformServices;
 using Microsoft.VisualStudio.TestPlatform.MSTestAdapter.PlatformServices.Interface;
 
-#pragma warning disable CS0618 // ignore FileOperations deprecation
-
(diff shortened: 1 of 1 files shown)


src/Adapter/MSTest.TestAdapter/SourceGeneration/SourceGeneratedFileOperations.cs
```diff
@@ -7,18 +7,19 @@
 using Microsoft.VisualStudio.TestPlatform.MSTestAdapter.PlatformServices;
 using Microsoft.VisualStudio.TestPlatform.MSTestAdapter.PlatformServices.Interface;
 
+#pragma warning disable CS0618 // ignore FileOperations deprecation
```
Youssef1313: This shouldn't be needed. The warning shows in IDE only due to a Roslyn bug that's fixed.

Youssef1313 (dismissed):

src/Adapter/MSTest.TestAdapter/SourceGeneration/SourceGeneratedFileOperations.cs
SimonCropp: fixed

Youssef1313 (approved):
Thanks for the cleanup!

PR #4286, opened 2024-12-08
remove param values where same as default

SimonCropp:

### DIFF
--- a/src/Platform/Microsoft.Testing.Platform/Hosts/TestHostBuilder.cs
+++ b/src/Platform/Microsoft.Testing.Platform/Hosts/TestHostBuilder.cs
@@ -383,7 +383,7 @@ internal sealed class TestHostBuilder(IFileSystem fileSystem, IRuntimeFeature ru
         // Check if we're in the test host or we should check test controllers extensions
         // Environment variable check should not be needed but in case we will rollback to use only env var we will need it.
         if ((!testHostControllerInfo.HasTestHostController ||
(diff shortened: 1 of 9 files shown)


src/Platform/Microsoft.Testing.Platform/Hosts/TestHostBuilder.cs
```diff
@@ -383,7 +383,7 @@ await LogTestHostCreatedAsync(
         // Check if we're in the test host or we should check test controllers extensions
         // Environment variable check should not be needed but in case we will rollback to use only env var we will need it.
         if ((!testHostControllerInfo.HasTestHostController ||
-            systemEnvironment.GetEnvironmentVariable($"{EnvironmentVariableConstants.TESTINGPLATFORM_TESTHOSTCONTROLLER_SKIPEXTENSION}_{testHostControllerInfo.GetTestHostControllerPID(true)}") != "1")
```
Youssef1313, line 386: @Evangelink @MarcoRossignoli FYI, we never pass false to `GetTestHostControllerPID`. Not sure if we should remove that parameter altogether?

Evangelink: It's internal so I don't mind removing as we don't need it for now.

Evangelink (approved):

src/Platform/Microsoft.Testing.Platform/Hosts/TestHostBuilder.cs
Evangelink: Merging the PR anyways, let's do a follow-up if @MarcoRossignoli agrees.

PR #4295, opened 2024-12-09
Enroll mtp to source build

nohwnd:


Evangelink (approved):

eng/SourceBuildPrebuiltBaseline.xml
```diff
@@ -0,0 +1,8 @@
+<!-- When altering this file or making other Source Build related changes, include @dotnet/source-build as a reviewer. -->
+<!-- See aka.ms/dotnet/prebuilts for guidance on what pre-builts are and how to eliminate them. -->
+
+<UsageData>
+  <IgnorePatterns>
+    <UsagePattern IdentityGlob="Polyfill/*" />
```
MichaelSimons: The reference to Polyfill will either need to be removed/conditionally excluded from source-build else it will also need to be added to source-build-externals as well.

nohwnd: I tried to exclude but we rely on it in code, it does not just polyfill for older tfms. What does it take to get it into source-build-externals?

Can you also please advise on how to get rid of these prebuilts? 

image

MichaelSimons: How is polyfill used in the product?  I'm am just trying to make sure it is really needed for source-build.  The instructions for adding new source-build-externals is at https://github.com/dotnet/source-build-externals/blob/main/README.md#adding-a-new-external-component.  You will want to explore the dependencies Polyfill has to ensure it is source build compatible (at least the components .NET would have a dependency on).

MichaelSimons: Your prebuilts appear to be arcade related.  Why is the version.details arcade dependency pointing to 10.0 while the global.json is downgrading to a 9.0 version.  I would expect 10 to be used in both.  What version of the SDK will these changes be flowing into?

nohwnd: These are the versions we are using in vstest and there we don't have these problems. Not sure what is missing here.

nohwnd: Polyfill does not have any deps, it is a source only package. It brings code as nuget package so the same modern C# syntax can be used in older tfms.

eng/SourceBuildPrebuiltBaseline.xml
```diff
@@ -0,0 +1,11 @@
+<!-- When altering this file or making other Source Build related changes, include @dotnet/source-build as a reviewer. -->
+<!-- See aka.ms/dotnet/prebuilts for guidance on what pre-builts are and how to eliminate them. -->
+
+<UsageData>
+  <IgnorePatterns>
+     <!-- These are coming in via runtime but the source-build infra isn't able to automatically pick up the right intermediate. -->
+    <UsagePattern IdentityGlob="Microsoft.NET.ILLink.Tasks/*9.0.*" />
+    <!-- This is a source only nuget package, there are no dlls coming to the build from this. -->
+    <UsagePattern IdentityGlob="Polyfill/*" />
```
nohwnd: @MichaelSimons this package only contains source code. There are no pre-builds in it.

Directory.Build.targets
```diff
@@ -10,5 +10,11 @@
     <AssemblyVersion>14.0.0.0</AssemblyVersion>
     <AssemblyFileVersion>14.0.0.0</AssemblyFileVersion>
   </PropertyGroup>
-
+  
+    <!--
+    CI source build leg: this needs to build only the current TFM.
+  -->
```
Evangelink:
```suggestion
  <!--
    CI source build leg: this needs to build only the current TFM.
  -->
```

eng/SourceBuildPrebuiltBaseline.xml
```diff
@@ -0,0 +1,11 @@
+<!-- When altering this file or making other Source Build related changes, include @dotnet/source-build as a reviewer. -->
+<!-- See aka.ms/dotnet/prebuilts for guidance on what pre-builts are and how to eliminate them. -->
+
+<UsageData>
+  <IgnorePatterns>
+     <!-- These are coming in via runtime but the source-build infra isn't able to automatically pick up the right intermediate. -->
+    <UsagePattern IdentityGlob="Microsoft.NET.ILLink.Tasks/*9.0.*" />
```
nohwnd, line 7: @MichaelSimons this is copied from asp.net core repo, without it I get warning about pre-built, please let me know if this is acceptable.

eng/Analyzers.props
```diff
@@ -7,7 +7,8 @@
     <EnforceCodeStyleInBuild>true</EnforceCodeStyleInBuild>
   </PropertyGroup>
 
-  <ItemGroup>
+  <!-- Source build requires that everything needed for build is source-buildable, so we don't use analyzers for source build. -->
+  <ItemGroup Condition="'$(DotNetBuildSourceOnly)' != 'true'">
```
Evangelink: Below you are using `'$(DotNetBuildFromSource)' == 'true' or '$(DotNetBuild)' == 'true'` shall we do the same here?

Evangelink: Nit:

```suggestion
  <ItemGroup Condition=" '$(DotNetBuildSourceOnly)' != 'true' ">
```

eng/Build.props
```diff
@@ -4,38 +4,47 @@
     <ProductsToBuild Condition=" '$(ProductsToBuild)' == '' ">all</ProductsToBuild>
   </PropertyGroup>
 
-  <!-- When building all products on windows OS -->
-  <ItemGroup Condition=" '$(ProductsToBuild)' == 'all' AND '$(OS)' == 'Windows_NT' ">
-    <ProjectToBuild  Include="$(RepoRoot)TestFx.sln" />
-  </ItemGroup>
+  <Choose>
+    <When Condition="'$(DotNetBuildFromSource)' == 'true' or '$(DotNetBuild)' == 'true'">
```
Evangelink: Nit:

```suggestion
    <When Condition=" '$(DotNetBuildFromSource)' == 'true' or '$(DotNetBuild)' == 'true' ">
```

eng/DotNetBuild.props
```diff
@@ -0,0 +1,10 @@
+<!-- When altering this file, please include @dotnet/product-construction as a reviewer. -->
```
Evangelink, line 1: Can you please edit codeowner for that file?

eng/SourceBuildPrebuiltBaseline.xml
```diff
@@ -0,0 +1,11 @@
+<!-- When altering this file or making other Source Build related changes, include @dotnet/source-build as a reviewer. -->
```
Evangelink, line 1: Can you please edit codeowner for that file?

eng/Version.Details.xml
```diff
@@ -25,5 +25,29 @@
...
+      <Uri>https://github.com/dotnet/source-build-reference-packages</Uri>
+      <Sha>1ebd9ce245112164207d961c0d2faea741c7c489</Sha>
+      <SourceBuild RepoName="source-build-reference-packages" ManagedOnly="true" />
+    </Dependency>
+    <!-- Intermediate is necessary for source build. -->    
+    <Dependency Name="Microsoft.SourceBuild.Intermediate.arcade" Version="10.0.0-beta.24604.4">    
+      <Uri>https://github.com/dotnet/arcade</Uri>    
+      <Sha>45d845e04c05fbe5da9838c454bbc3af1df6be81</Sha>    
+      <SourceBuild RepoName="arcade" ManagedOnly="true" />    
+    </Dependency>
```
Evangelink: As discussed offline, I would like to preserve this repo using NET 9, I need to change our subscription and unblock the arcade flow. Shall we downgrade these 2 to 9.X?

Directory.Build.targets
```diff
@@ -10,5 +10,11 @@
...
+  <PropertyGroup Condition=" '$(DotNetBuildSourceOnly)' == 'true'">
```
nohwnd:
```suggestion
  <PropertyGroup Condition=" '$(DotNetBuildSourceOnly)' == 'true' ">
```

eng/Build.props
nohwnd:
```suggestion
    <When Condition=" '$(DotNetBuildSourceOnly)' == 'true' ">
```

eng/Analyzers.props
nohwnd: " '$(DotNetBuildSourceOnly)' == 'true' " should be the more modern way of doing this based on this: https://github.com/dotnet/arcade/blob/main/Documentation/UnifiedBuild/Unified-Build-Controls.md

eng/Version.Details.xml
nohwnd: all good to me, I am just using the version that is used above by the repository to not see any prebuilts.

eng/SourceBuildPrebuiltBaseline.xml
nohwnd: done, but I don't think they will get notified cross organizations

eng/DotNetBuild.props
nohwnd: yup

eng/Version.Details.xml
nohwnd: how do you want to do this, I don't think you discussed this with me :)

eng/SourceBuildPrebuiltBaseline.xml
nohwnd: image
 does not work. the team is not visible we are in different org

Evangelink: @dotnet/source-build

Evangelink: @MichaelSimons would you consider having the team duplicated on MS org?

Thanks for testing this out @nohwnd

eng/Version.Details.xml
Evangelink: Let's keep as-is for now and I'll handle when I try to downgrade

Evangelink (approved):
LGTM but please let's wait approval from @MichaelSimons

eng/SourceBuildPrebuiltBaseline.xml
MichaelSimons: This is actually great.  Since this is a source code only package then we should add it to SBRP as a text-only package.  This is a lot easier than adding it as a source build external. See https://github.com/dotnet/source-build-reference-packages/blob/main/README.md#text-only

eng/Version.Details.xml
```diff
@@ -25,5 +25,29 @@
...
       <Sha>aa2fcc8616d988b234bc1d218465b20c56d0b82f</Sha>
     </Dependency>
+    <!-- Intermediate is necessary for source build. -->
+    <Dependency Name="Microsoft.SourceBuild.Intermediate.diagnostics" Version="9.0.0-preview.24566.1">
+      <Uri>https://github.com/dotnet/diagnostics</Uri>
+      <Sha>8c505ca6921b5f7e9b8acc234cc8f15035537ee4</Sha>
+      <SourceBuild RepoName="diagnostics" ManagedOnly="true" />
+    </Dependency>
+    <!-- Intermediate is necessary for source build. -->
+    <Dependency Name="Microsoft.SourceBuild.Intermediate.source-build-externals" Version="9.0.0-alpha.1.24175.1">
```
MichaelSimons, line 35: What version(s) of the SDK is this going to flow into?  Given arcade and SBRP is a 10.0 version, I would expect this to be the same version.

eng/SourceBuildPrebuiltBaseline.xml
MichaelSimons: I have no objection to duplicating the team - the name would have to be different microsoft/dotnet-source-build.

eng/SourceBuildPrebuiltBaseline.xml
MichaelSimons: I think this one is acceptable.

eng/Version.Details.xml
nohwnd: This will flow into .net sdk 10. But above Amaury mentions that he wants to downgrade the build to 9.

https://github.com/microsoft/testfx/pull/4295#discussion_r1878054720

eng/Version.Details.xml
MichaelSimons: What is the reason to downgrade?  I ask to see if this is going to flow into a 9.0 SDK or if it is for other reasons?

eng/SourceBuildPrebuiltBaseline.xml
nohwnd: https://github.com/dotnet/source-build-reference-packages/pull/1095

nohwnd:
```suggestion
```

eng/SourceBuildPrebuiltBaseline.xml
Evangelink: Cool, let me know when they are available so we can add the write access and update the codeowner file.

eng/Version.Details.xml
Evangelink: @MichaelSimons in past when we were on net8 and starting to move to net9 we got some "complaints" for community that it was painful and limiting their ability to contribute to be forced to be on alpha/beta/preview SDK. I know `build.cmd` installs local etc.

PR #4306, opened 2024-12-10
Refactor GetResultOrRunClassInitialize to have clearer intent

Evangelink:

### DIFF
--- a/src/Adapter/MSTest.TestAdapter/Execution/TestClassInfo.cs
+++ b/src/Adapter/MSTest.TestAdapter/Execution/TestClassInfo.cs
@@ -29,6 +29,8 @@ public class TestClassInfo
 {
     private readonly Lock _testClassExecuteSyncObject = new();
 
+    private UnitTestResult? _cachedClassInitializeResult;
+
     /// <summary>
     /// Initializes a new instance of the <see cref="TestClassInfo"/> class.
     /// </summary>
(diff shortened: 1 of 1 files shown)


src/Adapter/MSTest.TestAdapter/Execution/TestClassInfo.cs
```diff
@@ -387,7 +395,9 @@ internal UnitTestResult GetResultOrRunClassInitialize(ITestContext testContext,
                 PlatformServiceProvider.Instance.AdapterTraceLogger.LogWarning(Resource.STAIsOnlySupportedOnWindowsWarning);
             }
 
-            return DoRun();
+            UnitTestResult result = DoRun();
+            _cachedClassInitializeResult = result;
```
Youssef1313, line 399: The value set from this code path (non-STA) is not read. We only read in STA path. Do we want to do a check at the beginning that's common for STA and non-STA like this?

```csharp
if (_cachedClassInitializeResult is not null)
{
    return _cachedClassInitializeResult;
}

Evangelink: It's returned on 366, we could do `if (IsClassInitializeExecuted || _...)` or even simplify as the if you described and invert the assert,

Youssef1313: @Evangelink Line 366 is for STA specifically, while line 399 is for non-STA.

Evangelink: Nope. 366 is generic, 381 STA and here 399 for non-STA

PR #472, opened 2018-08-06
Enabling tcm properties flow to test context

abhishkk:
**Description**:
VS IDE has support of association in test explorer. Currently this support is disabled for mstest v2. 
Support for mstest v2 will be enabled with this PR: https://devdiv.visualstudio.com/DevDiv/Automated%20Testing/_git/VS/pullrequest/135489?_a=overview

When association is enabled for mstest v2 tests, user can associate mstest v2 tests with a test case and then run it via test plan flow.

In test plan flow, user might need to use tcm properties (like test case id, test run id, test plan id etc) in test method code. This PR contains changes to allow tcm properties to flow to test context.

**Validation Scenarios**:
Validated that for full framework and net core, tcm properties are properly passed to test context.

**Note**: In talk with @PBoraMSFT if RFC is required in mstest v2 or not as the flow is controled from IDE. There is a plan to add the documentation on association window in test explorer.

PR #4314, opened 2024-12-11
Remove dead code

Youssef1313:


src/Adapter/MSTest.TestAdapter/Constants.cs
```diff
@@ -88,36 +88,6 @@ internal static class Constants
 
     internal static readonly TestProperty InnerResultsCountProperty = TestProperty.Register("InnerResultsCount", InnerResultsCountLabel, typeof(int), TestPropertyAttributes.Hidden, typeof(TestResult));
 
-    internal static readonly TestProperty TestRunIdProperty = TestProperty.Register(TestRunId, TestRunId, typeof(int), TestPropertyAttributes.Hidden, typeof(TestCase));
```
Youssef1313, line 91: We never set a value for any of these properties. @Evangelink Could they be magically set by something else (not us)? Or is it really dead code?

src/Adapter/MSTest.TestAdapter/Execution/TestExecutionManager.cs
```diff
@@ -438,8 +438,7 @@ private void ExecuteTestsWithTestRunner(
             PlatformServiceProvider.Instance.AdapterTraceLogger.LogInfo("Executing test {0}", unitTestElement.TestMethod.Name);
 
             // Run single test passing test context properties to it.
-            IDictionary<TestProperty, object?> tcmProperties = TcmTestPropertiesProvider.GetTcmProperties(currentTest);
```
Youssef1313, line 441: This dictionary was always empty because we don't set `TestCaseIdProperty` anywhere, so GetTcmProperties returns an empty dictionary right away.

Evangelink:
I sadly have no idea... Looks dead but I don't know.

@nohwnd is it ringing any bell?

nohwnd:
all of them look azdo related (when it was still called tfs) https://github.com/microsoft/testfx-docs/issues/36  https://github.com/microsoft/testfx/pull/472

I would check with azdo team if they rely on these names.

Youssef1313:
A seemingly relevant doc: https://learn.microsoft.com/en-us/azure/devops/test/associate-automated-test-with-test-case

PR #4315, opened 2024-12-11
[Refactor] Use `List<T>` rather than `Collection<T>`

Youssef1313:
There shouldn't any noticable difference with this change. It's a minor cleanup.

`Collection<T>` ([Source](https://github.com/dotnet/runtime/blob/f0e1d7ba91f601064b90475541d5e7e91507f63e/src/libraries/System.Private.CoreLib/src/System/Collections/ObjectModel/Collection.cs)) is a mostly a thin wrapper around `List<T>` (except for cases where the constructor with `IList<T>` is used - but we don't use that constructor). I think it's a better idea to just use `List<T>` right away.

### DIFF
--- a/src/Adapter/MSTest.TestAdapter/Discovery/AssemblyEnumerator.cs
+++ b/src/Adapter/MSTest.TestAdapter/Discovery/AssemblyEnumerator.cs
@@ -72,7 +72,7 @@ internal class AssemblyEnumerator : MarshalByRefObject
     internal ICollection<UnitTestElement> EnumerateAssembly(
         string assemblyFileName,
         [StringSyntax(StringSyntaxAttribute.Xml, nameof(runSettingsXml))] string? runSettingsXml,
-        out ICollection<string> warnings)
+        out List<string> warnings)
     {
         DebugEx.Assert(!StringEx.IsNullOrWhiteSpace(assemblyFileName), "Invalid assembly file name.");
         var warningMessages = new List<string>();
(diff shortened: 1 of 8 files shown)


src/Adapter/MSTest.TestAdapter/Discovery/AssemblyEnumerator.cs
```diff
@@ -228,8 +228,7 @@ private List<UnitTestElement> DiscoverTestsInType(
         {
             typeFullName = type.FullName;
             TypeEnumerator testTypeEnumerator = GetTypeEnumerator(type, assemblyFileName, discoverInternals, discoveryOption, testIdGenerationStrategy);
-            ICollection<UnitTestElement>? unitTestCases = testTypeEnumerator.Enumerate(out ICollection<string> warningsFromTypeEnumerator);
-            warningMessages.AddRange(warningsFromTypeEnumerator);
+            List<UnitTestElement>? unitTestCases = testTypeEnumerator.Enumerate(warningMessages);
```
Youssef1313, line 229: I'm passing the `warningMessages` list directly. There is no point for `TypeEnumerator.Enumerate` to allocate a new list and then add that list to `warningMessages`.

src/Adapter/MSTest.TestAdapter/Discovery/TypeEnumerator.cs
```diff
@@ -119,12 +117,11 @@ internal Collection<UnitTestElement> GetTests(ICollection<string> warnings)
             currentType = currentType.BaseType;
         }
 
-        return new Collection<UnitTestElement>(
-            tests.GroupBy(
-                t => t.TestMethod.Name,
-                (_, elements) =>
-                    elements.OrderBy(t => inheritanceDepths[t.TestMethod.DeclaringClassFullName ?? t.TestMethod.FullClassName]).First())
-                .ToList());
+        return tests.GroupBy(
```
Youssef1313, line 119: This was created a `List<UnitTestElement>`, then wrapped it in a `Collection<UnitTestElement>`. Now that we return `List<UnitTestElement>`, there is no need to create an extra object.

Evangelink (dismissed):
Ideally not a big fan of mutable params passed around but it was already the case and it does improve logic so LGTM.

Evangelink (dismissed):

Evangelink (approved):

PR #4321, opened 2024-12-11
Display an error when '--maximum-failed-tests' is used with a framework not implementing the required capability

Youssef1313:


src/Platform/Microsoft.Testing.Platform/Extensions/AbortForMaxFailedTestsExtension.cs
```diff
@@ -53,7 +58,21 @@ public AbortForMaxFailedTestsExtension(
...
+        if (!_maxFailedTests.HasValue)
+        {
+            return false;
+        }
+
+        if (_capability is null)
+        {
+            await _outputDevice.DisplayAsync(this, new ErrorMessageOutputDeviceData(PlatformResources.AbortForMaxFailedTestsCapabilityNotAvailable));
+            return false;
+        }
```
Evangelink: Maybe we should hard fail instead (wdyt @MarcoRossignoli?). Pros to hardfail, forces you to fix. Cons for it, if you did it wrong and shipped, people cannot use your version (but maybe that's not "too bad").

MarcoRossignoli: Crash it with "wrong implementation".

src/Platform/Microsoft.Testing.Platform/Resources/PlatformResources.resx
```diff
@@ -713,4 +713,7 @@ Takes one argument as string in the format &lt;value&gt;[h|m|s] where 'value' is
     <value>Test session is aborting due to reaching failures ('{0}') specified by the '--maximum-failed-tests' option.</value>
     <comment>{0} is the number of max failed tests.</comment>
   </data>
-</root>
+  <data name="AbortForMaxFailedTestsCapabilityNotAvailable" xml:space="preserve">
+    <value>The current test framework does not support '--maximum-failed-tests'.</value>
```
Evangelink: I am wondering if we should add even more details like they need to open a bug on the test framework repo.

Or change message to be `The current test framework does not implement correctly the maximum failed test capability`

Evangelink (approved):

src/Platform/Microsoft.Testing.Platform/Resources/PlatformResources.resx
Youssef1313: @Evangelink I chose a slightly different wording to mention the interface name `IGracefulStopTestExecutionCapability`. Please let me know if you think what I chose is too wordy.

Evangelink (approved):

PR #4327, opened 2024-12-12
Upload test results for all OSes

Evangelink:

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -104,6 +104,15 @@ stages:
             name: Test
             displayName: Test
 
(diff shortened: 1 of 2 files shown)


azure-pipelines.yml
```diff
@@ -104,23 +104,23 @@ stages:
             name: Test
             displayName: Test
 
-          # Upload code coverage to codecov.io
```
Evangelink, line 107: Swapped so we always upload results first and then we can do codecov.io

nohwnd (approved):

PR #4329, opened 2024-12-12
Add source build to official build

nohwnd:


nohwnd:
force merged after checking with michael from source build. The change has no impact on PR build, so no point in waiting for it.

PR #4340, opened 2024-12-13
Auto detect DynamicDataSourceType

### LINKED ISSUE #4254: Simplify `[DynamicData]` to not require user to specify the member type
## Background and Motivation

Reflection call to detect if the member is property or method isn't expensive so there is no good reason to force user to have to specify member type. Besides, nowadays most people will use `nameof(MemberName)` that help to guarantee the member exists.

## Proposed Feature

I think the easiest and most transparent solution is probably to add a new state to the enum `DynamicDataSourceType` like `AutoDetect` that would be the default value. Although this is technically a breaking change, it should not have any impact for users who are using the default value or the explicit value.

Note we need to update the related code analyzer.

## Alternative Designs

Deprecate enum and related ctors but that would flow to the user for no real impact.

Youssef1313:
Fixes #4254


src/Adapter/MSTest.TestAdapter/DynamicDataOperations.cs
```diff
@@ -25,36 +25,33 @@ public IEnumerable<object[]> GetData(Type? _dynamicDataDeclaringType, DynamicDat
...
+                {
+                    obj = GetDataFromProperty(dynamicDataPropertyInfo);
+                }
+                else if (PlatformServiceProvider.Instance.ReflectionOperations.GetDeclaredMethod(_dynamicDataDeclaringType, _dynamicDataSourceName) is { } dynamicDataMethodInfo)
+                {
+                    obj = GetDataFromMethod(dynamicDataMethodInfo);
+                }
+                else
+                {
+                    // throw
```
Evangelink: Yes let's throw `NotSupportedException`. I know some other frameworks do support fields but I have never really found any good use case for it so not sure if that's worth supporting.

@nohwnd any opinion?

Evangelink (dismissed):
Some issue with public API it seems but otherwise LGTM

src/Adapter/MSTest.TestAdapter/DynamicDataOperations.cs
nohwnd: properties and methods seem fine if that is the current thing we support in mstest.

Evangelink: @Youssef1313 please add the throw here

Youssef1313: In progress

Evangelink (approved):

PR #4348, opened 2024-12-13
Move TATF based tests to MSTest

Evangelink:


Directory.Packages.props
```diff
@@ -65,12 +66,13 @@
...
     <PackageVersion Include="Microsoft.Extensions.FileSystemGlobbing" Version="9.0.0" />
-    <PackageVersion Include="Microsoft.Testing.Internal.Framework" Version="$(MicrosoftTestingInternalFrameworkVersion)" />
-    <PackageVersion Include="Microsoft.Testing.Internal.Framework.SourceGeneration" Version="$(MicrosoftTestingInternalFrameworkVersion)" />
     <PackageVersion Include="Microsoft.TestPlatform" Version="$(MicrosoftNETTestSdkVersion)" />
     <!-- Pinned to 4.18.4 for security -->
     <PackageVersion Include="Moq" Version="4.18.4" />
     <PackageVersion Include="MSBuild.StructuredLogger" Version="2.2.386" />
+    <PackageVersion Include="MSTest.Analyzers" Version="$(MSTestVersion)" />
+    <PackageVersion Include="MSTest.TestAdapter" Version="$(MSTestVersion)" />
+    <PackageVersion Include="MSTest.TestFramework" Version="$(MSTestVersion)" />
```
Youssef1313, line 75: Could we switch to the meta package?

samples/Playground/Tests.cs
```diff
@@ -18,7 +18,7 @@ public class TestClass
     public void Test() => TestContext.AddResultFile(@"c:\hello2");
 
     [TestMethod]
-    public void Test2() => Assert.AreEqual(1, 0, "few");
+    public void Test2() => Assert.AreEqual(0, 1, "few");
```
Youssef1313: Unnecessary change but it doesn't harm.

src/Platform/Microsoft.Testing.Extensions.VSTestBridge/ObjectModel/FastFilter.cs
```diff
@@ -195,14 +195,15 @@ private void AddProperty(string name, string value)
         {
             if (!_filterDictionaryBuilder.TryGetValue(name, out ImmutableHashSet<string>.Builder? values))
             {
-                values = ImmutableHashSet.CreateBuilder<string>(StringComparer.OrdinalIgnoreCase);
+                values = ImmutableHashSet.CreateBuilder(StringComparer.OrdinalIgnoreCase)!;
```
Youssef1313, line 198: Why is the suppression needed?

test/IntegrationTests/Microsoft.Testing.Platform.Acceptance.IntegrationTests/Properties/launchSettings.json
```diff
@@ -2,7 +2,7 @@
   "profiles": {
     "Microsoft.Testing.Platform.Acceptance.IntegrationTests": {
       "commandName": "Project",
-      "commandLineArgs": "--treenode-filter /*/*/*/**",
+      // "commandLineArgs": "--treenode-filter /*/*/*/**",
```
Youssef1313: Is this change intended?

Directory.Packages.props
Evangelink: I tried and something wasn't working but I don't recall what. Let me double check.

samples/Playground/Tests.cs
Evangelink: Part of the bulk fix on other assertions, I'll leave it as-is.

test/IntegrationTests/Microsoft.Testing.Platform.Acceptance.IntegrationTests/Properties/launchSettings.json
Evangelink: Yes! The filter isn't supported by MSTest so when you try to run the project it fails. I need to double check this is done everywhere!

test/UnitTests/Microsoft.Testing.Platform.MSBuild.UnitTests/Microsoft.Testing.Platform.MSBuild.UnitTests.csproj
```diff
@@ -1,9 +1,9 @@
 <Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFrameworks>$(MicrosoftTestingTargetFrameworks);net462</TargetFrameworks>
+    <TargetFrameworks>net8.0;net9.0</TargetFrameworks>
```
Youssef1313, line 4: Maybe use `$(NetMinimum);$(NetCurrent)`? (I assume NetMinimum flows properly from Arcade.

test/Utilities/Microsoft.Testing.TestInfrastructure/Microsoft.Testing.TestInfrastructure.csproj
```diff
@@ -7,18 +7,17 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="Microsoft.Testing.Internal.Framework" />
+    <PackageReference Include="MSTest.TestFramework" />
```
Youssef1313, line 11: Should we explicitly reference MSTest.Analyzers here as well? At least until we are using 3.8 which will have https://github.com/microsoft/testfx/pull/4351

Directory.Packages.props
Evangelink: Looks like it was just to avoid `Microsoft.NET.Test.Sdk` (the VSTest deps) that is part of the bundle.

src/Platform/Microsoft.Testing.Extensions.VSTestBridge/ObjectModel/FastFilter.cs
Evangelink: It's complaining about inconsistent nullability.

test/Utilities/Microsoft.Testing.TestInfrastructure/Microsoft.Testing.TestInfrastructure.csproj
Evangelink: Yes!

test/UnitTests/Microsoft.Testing.Platform.MSBuild.UnitTests/Microsoft.Testing.Platform.MSBuild.UnitTests.csproj
Evangelink: It would not work well when we move to arcade 10, we would only test `net8.0` and `net10.0`

src/Platform/Microsoft.Testing.Extensions.VSTestBridge/ObjectModel/FastFilter.cs
Youssef1313: Oh, was the code previously creating `ImmutableHashSet<string>` and it's not `ImmutableHashSet<string?>` ? In that case, I think it's better to keep the type parameter explicitly specified as `string`

MarcoRossignoli (approved):

PR #4349, opened 2024-12-13
Avoid more overhead of assembly initialize

nohwnd:
Looks like we are spending about 1% of the test run on 1k tests trying to setup assembly initialize contexts inside of RunAssemblyInitializeIfNeeded, only to enter the child method that would run AssemblyInitialize where we just return if we already run the code.

### DIFF
--- a/src/Adapter/MSTest.TestAdapter/Execution/UnitTestRunner.cs
+++ b/src/Adapter/MSTest.TestAdapter/Execution/UnitTestRunner.cs
@@ -160,15 +160,22 @@ internal sealed class UnitTestRunner : MarshalByRefObject
                 _fixtureTests.TryAdd(testMethod.AssemblyName, testMethodInfo);
                 _fixtureTests.TryAdd(testMethod.AssemblyName + testMethod.FullClassName, testMethodInfo);
 
-                UnitTestResult assemblyInitializeResult = RunAssemblyInitializeIfNeeded(testMethodInfo, testContext);
+                UnitTestResult? assemblyInitializeResult = null;
 
-                if (assemblyInitializeResult.Outcome != UnitTestOutcome.Passed)
+                if (!testMethodInfo.Parent.Parent.IsAssemblyInitializeExecuted)
+                {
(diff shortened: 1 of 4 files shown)


src/Adapter/MSTest.TestAdapter/Execution/UnitTestRunner.cs
```diff
@@ -160,15 +160,25 @@ internal UnitTestResult[] RunSingleTest(TestMethod testMethod, IDictionary<strin
...
-                UnitTestResult assemblyInitializeResult = RunAssemblyInitializeIfNeeded(testMethodInfo, testContext);
+                UnitTestResult? assemblyInitializeResult = null;
 
-                if (assemblyInitializeResult.Outcome != UnitTestOutcome.Passed)
+                if (!testMethodInfo.Parent.Parent.IsAssemblyInitializeExecuted)
+                {
+                    assemblyInitializeResult = RunAssemblyInitializeIfNeeded(testMethodInfo, testContext);
+                }
+
+                if (assemblyInitializeResult != null && assemblyInitializeResult.Outcome != UnitTestOutcome.Passed)
```
Youssef1313, line 170: This changes the semantics of the code I think.

Previously, if assembly initialize failed, we will attach the failure for every single test run. While now, it's only attached for the first test.

nohwnd: from the code it looks like we prepare  to capture outputs of the initialize method in every single test, but we actually run zero times or one time. so hopefully this changes nothing. We just prevent setting up capturing of output for method that will immediately return.

nohwnd:
Not sure why that test complains, but running against local build produces error for every test. So probably the acceptance test is wrong: 
image

nohwnd:
same for class init: 
image

Youssef1313:
@nohwnd There were large refactorings here. Probably the overhead was resolved already along the way. Can you check if there is anything to do here or if we should close this PR?

nohwnd:
re measured and no longer shows up.

Issue #4257, opened 2024-12-06
Add relaxed versions of Assert.ThrowsException and Assert.ThrowsExceptionAsync


Evangelink:
## Summary

Allows to assert some derived exceptions are thrown.

## Background and Motivation

`Assert.ThrowsException<T>` and `Assert.ThrowsExceptionAsync<T>` are checking that the exception throws is exactly matching the one specified as generic type argument. It can often happen that you would instead prefer to say type or derived types.

## Proposed Feature

Add `Assert.ThrowsAnyException<T>`/`Assert.ThrowsAnyExceptionAsync<T>` or add overloads of the current APIs with a boolean allowing to control whether to check exact type or derived.

## Alternative Designs

Relax current APIs and introduce `Assert.ThrowsExactException<T>` and `Assert.ThrowsExactExceptionAsync`. Although I prefer the naming, this would introduce a potentially strong functional breaking change for users so I would be reluctant to do it.

nohwnd:
My preferred syntax would also be changing the current to "any" and ThrowsExactException. But I see why that is not possible.

So with that not being an option I think the proposed design is the best. I do like the name being reasonably simple.

Evangelink:
The simplest idea would be to add `ThrowsAnyException(Async)` APIs but the name could be seen as confusing. 

Given that in v4 we would like to ensure simplified naming, and for example remove the `Exception` part of the API, it seems better to:

1. Add 2 new APIs (`Throws(Async)` and `ThrowsExactly(Async)`)
2. Add an analyzer + code fix to change `ThrowsException(Async)` into `ThrowsExactly(Async)`
3. In v4 either obsolete the older API or obsolete in a later release and drop in v4

PR #4350, opened 2024-12-13
Introduce Assert.Throws(Async) and Assert.ThrowsExactly(Async) APIs

Evangelink:
Contribute to #4257

### DIFF
--- a/src/TestFramework/TestFramework/Assertions/Assert.ThrowsException.cs
+++ b/src/TestFramework/TestFramework/Assertions/Assert.ThrowsException.cs
@@ -3,6 +3,7 @@
 
 using System.Diagnostics.CodeAnalysis;
 using System.Globalization;
+using System.Runtime.CompilerServices;
 
 namespace Microsoft.VisualStudio.TestTools.UnitTesting;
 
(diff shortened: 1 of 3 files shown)


src/TestFramework/TestFramework/Assertions/Assert.ThrowsException.cs
```diff
@@ -163,6 +218,10 @@ public static T ThrowsException<T>(Func<object?> action, string message, params
     [SuppressMessage("Microsoft.Design", "CA1031:DoNotCatchGeneralExceptionTypes", Justification = "Requirement is to handle all kinds of user exceptions and format appropriately.")]
```
Youssef1313: No longer relevant?

Youssef1313 (commented):
Should we add EditorBrowsable(Never) to the existing ThrowsException(Async) APIs to avoid confusion when typing Assert.Throws and invoke completion in IDE?

Evangelink:
> Should we add EditorBrowsable(Never) to the existing ThrowsException(Async) APIs to avoid confusion when typing Assert.Throws and invoke completion in IDE?

Yes! I never remember this trick! Done!

PR #4367, opened 2024-12-17
Reduce alloc in GetStackTraceInformation

SimonCropp:
* cache s_utaEndOfInnerExceptionTrace  statically to avoid the string concat on every iteration of the loop
 * remove CurrentCulture usage. they are all strings
 * avoid the concat with the popped value. yes this is redundant on newer runtimes, but not for older

### DIFF
--- a/src/Adapter/MSTest.TestAdapter/Execution/ExceptionHelper.cs
+++ b/src/Adapter/MSTest.TestAdapter/Execution/ExceptionHelper.cs
@@ -79,7 +79,7 @@ internal static class ExceptionHelper
         {
             if (!first)
             {
-                result.Append(s_utaEndOfInnerExceptionTrace);
+                result.Append(utaEndOfInnerExceptionTrace);
             }
 
             result.Append(' ');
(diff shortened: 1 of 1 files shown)


src/Adapter/MSTest.TestAdapter/Execution/ExceptionHelper.cs
```diff
@@ -77,18 +77,22 @@ internal static class ExceptionHelper
...
+            result.Append(' ');
+            result.AppendLine(stackTraces.Pop());
+
             first = false;
         }
 
         return CreateStackTraceInformation(ex, true, result.ToString());
     }
 
+    private static readonly string utaEndOfInnerExceptionTrace = Resource.UTA_EndOfInnerExceptionTrace + Environment.NewLine;
```
Youssef1313: Note: Caching will change behavior.

Previously, CurrentUICulture was respected every time `Resource.UTA_EndOfInnerExceptionTrace` is accessed.

src/Adapter/MSTest.TestAdapter/Execution/ExceptionHelper.cs
```diff
@@ -77,18 +77,22 @@ internal static class ExceptionHelper
...
         {
-            result.AppendFormat(
-                    CultureInfo.CurrentCulture,
-                    "{0} {1}{2}",
-                    first ? string.Empty : (Resource.UTA_EndOfInnerExceptionTrace + Environment.NewLine),
-                    stackTraces.Pop(),
-                    Environment.NewLine);
+            if (!first)
+            {
+                result.Append(utaEndOfInnerExceptionTrace);
```
Youssef1313: Maybe `result.AppendLine(Resource.UTA_EndOfInnerExceptionTrace)` so that we still respect CurrentUICulture for resource restrieval?

Youssef1313 (approved):
LGTM

Evangelink (approved):
It feels strange to always have the space prefix but you preserved intent so LGTM

Issue #4266, opened 2024-12-07
Revise Regex usage in TrimStackTrace


Youssef1313:
https://github.com/microsoft/testfx/blob/beb890d4fe699cf718748884c1e642e28239c1fc/src/Adapter/MSTest.TestAdapter/Execution/ExceptionHelper.cs#L109

It doesn't make much sense to me to use Regex here.

Not so much allocations but would be good to reduce:

![Image](https://github.com/user-attachments/assets/2597b08c-7116-48c5-979d-77dbb1bb6e0e)

Youssef1313:
This can even be more efficient than `string.Split`.

We can do the loop ourselves and use spans to not allocate anything here.

nohwnd:
Agree. Probably was used because it has overload that takes string, rather than having to wrap the string into array to call string.split.

PR #4368, opened 2024-12-17
Span trim stack trace

SimonCropp:


Youssef1313:
@SimonCropp Thanks for improving this!

I'm curious, did you have a scenario where you have seen this in a performance trace? Or did you work on this because of https://github.com/microsoft/testfx/issues/4266 ?

SimonCropp:
@Youssef1313 just 4266

SimonCropp:
@Youssef1313 
thoughts on adding System.ValueTuple and System.Memory TestFramework.csproj for some older tfms? i need that to make the polyfill work. if u dont like it can can go a different direction

Youssef1313:
Sorry @SimonCropp, this slipped off our radar and I and @Evangelink haven't discussed this specifically yet. But generally we try to be very careful when introducing dependencies.

PR #4373, opened 2024-12-18
Fix main

Youssef1313:

### DIFF
--- a/test/Directory.Build.targets
+++ b/test/Directory.Build.targets
@@ -28,7 +28,11 @@
     <ProjectReference Include="$(RepoRoot)src\Platform\Microsoft.Testing.Extensions.TrxReport\Microsoft.Testing.Extensions.TrxReport.csproj" />
     <PackageReference Include="Microsoft.Testing.Extensions.CodeCoverage" GeneratePathProperty="True" />
     <PackageReference Include="MSTest.Analyzers" />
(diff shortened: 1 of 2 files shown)


test/Directory.Build.targets
```diff
@@ -28,7 +28,9 @@
     <ProjectReference Include="$(RepoRoot)src\Platform\Microsoft.Testing.Extensions.TrxReport\Microsoft.Testing.Extensions.TrxReport.csproj" />
     <PackageReference Include="Microsoft.Testing.Extensions.CodeCoverage" GeneratePathProperty="True" />
     <PackageReference Include="MSTest.Analyzers" />
+
     <PackageReference Include="MSTest.TestFramework" />
+    <PackageReference Update="MSTest.TestFramework" Condition=" '$(UseMSTestFromSource)' == 'true' " ExcludeAssets="runtime;compile" />
```
Evangelink, line 33: I guess that works too, I was envisioning something like


```suggestion
     <PackageReference Include="MSTest.TestFramework" Condition=" '$(UseMSTestFromSource)' != 'true' " />
```

Evangelink (approved):

test/Directory.Build.targets
Evangelink: I'll merge as-is to unblock ci. Feel free to do a follow-up if you find my syntax better.

PR #4389, opened 2024-12-18
Allow any IEnumerable as return type to DynamicData source member

### LINKED ISSUE #4363: Allow single type as return of `[DynamicData]` sources
## Summary

Simplify even further allowed type for the data souce methods.

## Background and Motivation

Data sources provided for `[DynamicData]` have been simplified to allow `IEnumerable<ITuple>` in addition to `IEnumerable<object[]>` and we should go even further by allowing `T` to be any type (e.g. `IEnumerable<int>`, `IEnumerable<MyType>`...). As you may still want to use a source despite having a single parameter in your test method as you benefit from unfolding/expanding.

## Proposed Feature

Allow any `IEnumerable<T>` type.

Evangelink:
Fixes #4363


Youssef1313 (dismissed):

Evangelink:
Flakiness on timeout acceptance tests, I'll force merge

PR #4354, opened 2024-12-14
Open source Retry and Hot Reload extensions

Evangelink:
Open-source the retry and hot reload extensions but retain the dual-license on these 2 extensions.

We should wait for #4348 to be merged before considering this PR.

TODO:
- Add readme files in the 2 dual-license extensions
- Update root readme
- Update repo license to not show MIT by default to ensure we are not misleading users.

PR #4393, opened 2024-12-19
Cleanup retry and hot reload versions and refs

Evangelink:
Follow-up from #4354 to cleanup references (use project ref instead of package ref) and fix version included into MSTest.Sdk


Youssef1313 (approved):

Evangelink:
Flakiness on timeout acceptance tests, I'll force merge

PR #4394, opened 2024-12-19
Fix flakiness on test Log_WhenAsyncFlush_StreamWriterIsCalledOnlyWhenLogLevelAllowsIt

Evangelink:
For netfx, we would queue the message and we have a second loop reading from the queue. It's possible that when we assert, we still have not read. Now we ensure the loop is flushed before reading.

### DIFF
--- a/test/UnitTests/Microsoft.Testing.Platform.UnitTests/Logging/FileLoggerTests.cs
+++ b/test/UnitTests/Microsoft.Testing.Platform.UnitTests/Logging/FileLoggerTests.cs
@@ -197,7 +197,7 @@ public sealed class FileLoggerTests : IDisposable
 
     [TestMethod]
     [DynamicData(nameof(LogTestHelpers.GetLogLevelCombinations), typeof(LogTestHelpers), DynamicDataSourceType.Method)]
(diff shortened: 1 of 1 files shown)


Youssef1313 (approved):

MarcoRossignoli (approved):
💘

test/UnitTests/Microsoft.Testing.Platform.UnitTests/Logging/FileLoggerTests.cs
```diff
@@ -214,13 +214,11 @@ public async Task Log_WhenAsyncFlush_StreamWriterIsCalledOnlyWhenLogLevelAllowsI
             _mockFileStreamFactory.Object);
         fileLogger.Log(currentLogLevel, Message, null, Formatter, Category);
 
+        // Ensures that the async flush is completed before the file is read
+        fileLogger.Dispose();
```
Youssef1313: The variable is declared with `using`, so Dispose will be called twice. It's usually not a problem as the second dispose will simple `return;`, but it's redundant.

I would remove `using` from line 207, and wrap line 215 with `try/finally` and have the dispose call in finally.

The current code is still okay though and will not cause issues.

Youssef1313 (approved):

Issue #4371, opened 2024-12-17
Cannot run UWP tests with 3.6.2


jnyrup:
## Describe the bug
After upgrading from 3.6.1 to 3.6.4 I could no longer run tests from [UWP.Specs](https://github.com/fluentassertions/fluentassertions/tree/develop/Tests/UWP.Specs) in Visual Studio.

v3.6.2 was the first version that no longer worked for me.
I suspect it being related to #3949 as if I remove these three lines
https://github.com/microsoft/testfx/pull/3949/files#diff-9dfba7d4ed3c8c0e52817be015e3b388f7e1cd95cd6301c04684b604d720cb99R37-R39
or explicitly set `<EnableMSTestV2CopyResources>false</EnableMSTestV2CopyResources>` in my csproj the test is again runnable.

## Steps To Reproduce
1. Clone https://github.com/fluentassertions/fluentassertions
2. Checkout `c050f6bcaec68195f1d0dd0e44d233c6bbecf033`
3. Open `FluentAssertions.sln`
4. Change to release mode.
5. Right click on `UWP.Specs` and select build (the project is excluded from the configuration manager)
6. See that `Determining_caller_identity_should_not_throw_for_native_programs` is runnable and succeeds.
7. Edit `Tests\UWP.Specs\UWP.Specs.csproj` to use 3.6.2
8. Build `UWP.Specs` again
9. See that `Determining_caller_identity_should_not_throw_for_native_programs` is no longer runnable.

## Expected behavior
`Determining_caller_identity_should_not_throw_for_native_programs` should succeed

## Actual behavior
`Determining_caller_identity_should_not_throw_for_native_programs` is not runnable from VS

Evangelink:
Perfect timing @jnyrup! I was about to start release of 3.7, including into that milestone so we fix and ship tomorrow.

Evangelink:
Unrelated to your issue but as I am building I noticed:

![Image](https://github.com/user-attachments/assets/5a037514-7690-473b-9826-d3a7155b4796)

MSTest is also supporting the new platform and xUnit and NUnit are on their way too (beta/previews for them).

Evangelink:
@jnyrup I am not managing to build the uwp package on the commit you gave me.

Evangelink:
Ok so looks like we have a few problems, it's not yet fully clear as the issue is different for 3.7

PR #4402, opened 2024-12-20
Don't reference MTP in UWP

Evangelink:
Contributes to #4371

Thanks to the report, I have realized we were about to ship a version of MSTest broken for UWP. We are sadly having issues with having acceptance test for UWP because of arcade infra.

### DIFF
--- a/src/Adapter/MSTest.TestAdapter/RunConfigurationSettings.cs
+++ b/src/Adapter/MSTest.TestAdapter/RunConfigurationSettings.cs
@@ -2,10 +2,14 @@
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
 using System.Diagnostics.CodeAnalysis;
-using System.Globalization;
 using System.Xml;
 
+#if !WINDOWS_UWP
+using System.Globalization;
+
 using Microsoft.Testing.Platform.Configurations;
+#endif
+
 using Microsoft.VisualStudio.TestPlatform.ObjectModel.Adapter;
 using Microsoft.VisualStudio.TestPlatform.ObjectModel.Utilities;
 using Microsoft.VisualStudio.TestPlatform.PlatformAbstractions;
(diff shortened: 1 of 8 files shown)


src/Adapter/MSTest.TestAdapter/RunConfigurationSettings.cs
```diff
@@ -2,10 +2,14 @@
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
 using System.Diagnostics.CodeAnalysis;
-using System.Globalization;
 using System.Xml;
 
+#if !WINDOWS_UWP
+using System.Globalization;
+
```
fhnaseer, line 9: nit: empty line,

fhnaseer (dismissed):

src/Adapter/MSTest.TestAdapter/RunConfigurationSettings.cs
Evangelink: We have policy to have empty line between groups of usings hence the space. Thanks @fhnaseer

src/Adapter/MSTestAdapter.PlatformServices/Interfaces/IConfiguration.cs
```diff
@@ -0,0 +1,10 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+#if WINDOWS_UWP
+namespace Microsoft.Testing.Platform.Configurations;
+
+public interface IConfiguration
```
Youssef1313: Does this have to be public?

Evangelink: I thought it was on some public path but looks like not, thanks for pointing it out. Fixing that.

PR #4413, opened 2024-12-20
use EnumPolyfill.GetNames correctly

SimonCropp:
the *Polyfill types are meant to be used as a replacement to remove the need for the ifdef

on newer frameworks they redirect to correct implementation


Evangelink (approved):

Evangelink:
Flakiness again :c

PR #4420, opened 2024-12-21
remove redundant null check in ComputeStringHash

SimonCropp:
also fixes a nullability warning


Evangelink (approved):

src/Platform/Microsoft.Testing.Extensions.HangDump/Helpers/FNV1HashHelper.cs
```diff
@@ -14,16 +14,12 @@ internal static class FNV_1aHashHelper
     /// </summary>
     public static uint ComputeStringHash(string s)
     {
-        uint num = default;
-        if (s != null)
+        uint num = 2166136261u;
+        int num2 = 0;
+        while (num2 < s.Length)
```
Youssef1313: Maybe more readable this way:

```csharp
for (int i = 0; i < s.Length; i++)
{
    num = (s[i] ^ num) * 16777619;
}
```

Evangelink (approved):

PR #4422, opened 2024-12-21
make RegisterSerializer generic

SimonCropp:
this is purely cosmetic. so feel free to burn it if u dont like it


Evangelink (commented):
Fine by me. @MarcoRossignoli WDYT ?

SimonCropp:
yeah it just makes the usage a little more succinct

src/Platform/Microsoft.Testing.Platform/IPC/NamedPipeBase.cs
```diff
@@ -12,9 +12,11 @@ internal abstract class NamedPipeBase
     private readonly Dictionary<Type, object> _typeSerializer = [];
     private readonly Dictionary<int, object> _idSerializer = [];
 
-    public void RegisterSerializer(INamedPipeSerializer namedPipeSerializer, Type type)
+    public void RegisterSerializer<TSerializer, TInput>()
+        where TSerializer : INamedPipeSerializer, new()
     {
-        _typeSerializer.Add(type, namedPipeSerializer);
+        INamedPipeSerializer namedPipeSerializer = new TSerializer();
```
jnyrup, line 18: A minor difference here is that `Activator.CreateInstance` is now used to construct `namedPipeSerializer`.

Evangelink (changes requested):
@SimonCropp I have discussed it with the team and we will stick with current approach. `Activator.CreateInstance` should work with NAOT and trimming but we don't know if there are cases where it would be different. We don't yet have automatized enough tests around NAOT and trimming and we know current approach does work on all manual tests we did.

PR #4431, opened 2024-12-23
UnreachableException is more correct

SimonCropp:


Evangelink (commented):
AFAIK this is a NET9 exception so it won't work for analyzers (targeting `netstandard2.0`)

Evangelink (approved):

PR #4433, opened 2024-12-23
remove redundant requestedName != null check

SimonCropp:
not possible for it to be null

### DIFF
--- a/src/Adapter/MSTestAdapter.PlatformServices/AssemblyResolver.cs
+++ b/src/Adapter/MSTestAdapter.PlatformServices/AssemblyResolver.cs
@@ -358,8 +358,6 @@ class AssemblyResolver :
             return null;
         }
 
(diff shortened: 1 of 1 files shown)


src/Adapter/MSTestAdapter.PlatformServices/AssemblyResolver.cs
```diff
@@ -358,8 +358,6 @@ protected virtual
             return null;
         }
 
-        DebugEx.Assert(requestedName != null && !StringEx.IsNullOrEmpty(requestedName.Name), "MSTest.AssemblyResolver.OnResolve: requested is null or name is empty!");
```
Youssef1313, line 361: This line was asserting two two conditions. Did you intend to only remove the `requestedName != null`?

Youssef1313 (approved):
LGTM with a small comment

src/Adapter/MSTestAdapter.PlatformServices/AssemblyResolver.cs
SimonCropp: shit. yes, will ifx

Youssef1313 (approved):

PR #4434, opened 2024-12-23
Fix analyzer messages not localized

Youssef1313:
Fixes [AB#2325919](https://devdiv.visualstudio.com/DevDiv/_workitems/edit/2325919)

### DIFF
--- a/test/IntegrationTests/MSTest.Acceptance.IntegrationTests/AnalyzersTests.cs
+++ b/test/IntegrationTests/MSTest.Acceptance.IntegrationTests/AnalyzersTests.cs
@@ -1,0 +1,63 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using Microsoft.Testing.Platform.Acceptance.IntegrationTests;
+using Microsoft.Testing.Platform.Acceptance.IntegrationTests.Helpers;
+
+namespace MSTest.Acceptance.IntegrationTests;
+
+[TestClass]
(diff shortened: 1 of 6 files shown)


test/IntegrationTests/MSTest.Acceptance.IntegrationTests/AnalyzersTests.cs
```diff
@@ -0,0 +1,49 @@
...
+    {
+    }
+}
+""".PatchTargetFrameworks(TargetFrameworks.NetCurrent)
+    .PatchCodeWithReplace("$MSTestVersion$", MSTestVersion);
+
+        using TestAsset testAsset = await TestAsset.GenerateAssetAsync("Analyzers", code);
+        DotnetMuxerResult result = await DotnetCli.RunAsync($"build {testAsset.TargetAssetPath}", AcceptanceFixture.NuGetGlobalPackagesFolder.Path, environmentVariables: new()
+        {
+            ["DOTNET_CLI_UI_LANGUAGE"] = "it-IT",
```
Evangelink, line 43: Why are you setting the 3 vars? Only one should be working, right?

Evangelink (approved):

test/IntegrationTests/MSTest.Acceptance.IntegrationTests/AnalyzersTests.cs
Youssef1313: @Evangelink We [propagate the *existing* env variables to DotnetCli](https://github.com/microsoft/testfx/blob/2564c52733fde212f367e5869abdc12046c6c8e5/test/Utilities/Microsoft.Testing.TestInfrastructure/DotnetCli.cs#L64-L65). I didn't spend the time to check which one takes precedence as I thought it's okay to just set everything.

PR #4444, opened 2024-12-24
Better handling of CTS and timeout tokens

Evangelink:


src/Adapter/MSTest.TestAdapter/Execution/TestMethodInfo.cs
```diff
@@ -782,18 +792,27 @@ private bool SetTestContext(object classInstance, TestResult result)
...
-            string exceptionMessage = actualException.GetFormattedExceptionMessage();
-            StackTraceInformation? stackTraceInfo = actualException.GetStackTraceInformation();
+            Exception realException = ex.GetRealException();
 
-            string errorMessage = string.Format(
-                CultureInfo.CurrentCulture,
-                Resource.UTA_InstanceCreationError,
-                TestClassName,
-                exceptionMessage);
+            if (realException.IsOperationCanceledExceptionFromToken(TestMethodOptions.TestContext!.Context.CancellationTokenSource.Token))
```
Evangelink, line 818: Better message when timeout happens during ctor

src/Adapter/MSTest.TestAdapter/Helpers/FixtureMethodRunner.cs
```diff
@@ -16,6 +16,13 @@ internal static class FixtureMethodRunner
         Action action, CancellationTokenSource cancellationTokenSource, TimeoutInfo? timeoutInfo, MethodInfo methodInfo,
         IExecutionContextScope executionContextScope, string methodCanceledMessageFormat, string methodTimedOutMessageFormat)
     {
+        if (cancellationTokenSource.Token.IsCancellationRequested)
```
Evangelink, line 23: Don't bother doing any processing if the token is canceled.

test/IntegrationTests/MSTest.Acceptance.IntegrationTests/InitializeAndCleanupTimeoutTests.cs
```diff
@@ -1,762 +0,0 @@
-﻿// Copyright (c) Microsoft Corporation. All rights reserved.
```
Evangelink, line 1: All tests are moved to the `TimeoutTests.cs`  class

Youssef1313:
Looks like something is still flaky? `CooperativeCancellation_WhenTestCleanupTimeoutExpiresAndUserChecksToken_StepThrows` is currently failing on MacOS release.

src/Adapter/MSTest.TestAdapter/Execution/TestMethodInfo.cs
```diff
@@ -902,4 +920,7 @@ void ExecuteAsyncAction()
         RunTestCleanupMethod(timeoutResult);
         return timeoutResult;
     }
+
+    void IDisposable.Dispose()
+        => _timeoutTokenSource?.Dispose();
```
Youssef1313: I'm curious why is it needed? Don't we dispose in the finally block already in the same method that creates the CTS?

src/Adapter/MSTest.TestAdapter/Extensions/ExceptionExtensions.cs
```diff
@@ -32,6 +32,10 @@ internal static Exception GetRealException(this Exception exception)
         return exception;
     }
 
+    internal static bool IsOperationCanceledExceptionFromToken(this Exception ex, CancellationToken cancellationToken)
+        => (ex is OperationCanceledException oce && oce.CancellationToken == cancellationToken)
+        || (ex is AggregateException aggregateEx && aggregateEx.InnerExceptions.OfType<OperationCanceledException>().Any(oce => oce.CancellationToken == cancellationToken));
```
Youssef1313: Could we have an AggregateException with another inner AggregateException?

In that case, would it be better to change this to the following?

```csharp
    internal static bool IsOperationCanceledExceptionFromToken(this Exception ex, CancellationToken cancellationToken)
        => (ex is OperationCanceledException oce && oce.CancellationToken == cancellationToken)
        || (ex is AggregateException aggregateEx && aggregateEx.InnerExceptions.Any(innerEx => innerEx.IsOperationCanceledExceptionFromToken(cancellationToken));
```

src/Adapter/MSTest.TestAdapter/Helpers/FixtureMethodRunner.cs
Youssef1313: Do we precisely know when this can happen?

src/Adapter/MSTest.TestAdapter/Extensions/ExceptionExtensions.cs
```diff
@@ -32,6 +32,10 @@ internal static Exception GetRealException(this Exception exception)
         return exception;
     }
 
+    internal static bool IsOperationCanceledExceptionFromToken(this Exception ex, CancellationToken cancellationToken)
```
Youssef1313: Is it intentional this doesn't handle TaskCanceledException?

Youssef1313: If not intentional, it may be a good idea to rename the method when fixing the impl. Something like `IsOperationCanceledOrTaskCanceledExceptionFromToken` or `IsOperationOrTaskCanceledExceptionFromToken` or anything that clearly expresses the intent.

test/IntegrationTests/MSTest.Acceptance.IntegrationTests/Properties/launchSettings.json
```diff
@@ -2,7 +2,7 @@
   "profiles": {
     "MSTest.Acceptance.IntegrationTests": {
       "commandName": "Project",
-      "commandLineArgs": ""
+      //"commandLineArgs": "--filter ClassName~TimeoutTests"
```
Youssef1313: nit: Consider reverting this change before merge.

src/Adapter/MSTest.TestAdapter/Execution/TestMethodInfo.cs
Evangelink: We do but I prefer to potentially double dispose than suppress the CA warning in case the implementation changes.

src/Adapter/MSTest.TestAdapter/Extensions/ExceptionExtensions.cs
Evangelink: TCE inherits from OCE

src/Adapter/MSTest.TestAdapter/Extensions/ExceptionExtensions.cs
Evangelink: I'm not aware this could happen. Maybe using Flatten would be best

src/Adapter/MSTest.TestAdapter/Execution/TestMethodInfo.cs
Youssef1313: I think the CA warning is because we are storing the CTS in a field. But do we actually need to store it in a field? We should be using it in a very limited scope and I guess it should be possible to just pass it around?

src/Adapter/MSTest.TestAdapter/Extensions/ExceptionExtensions.cs
Youssef1313: Ah okay. That's good then!

src/Adapter/MSTest.TestAdapter/Extensions/ExceptionExtensions.cs
Evangelink: I have tried a few things and cannot find a single scenario where this happens so I'll keep current logic.

src/Adapter/MSTest.TestAdapter/Execution/TestMethodInfo.cs
Evangelink: You are right, I am refactoring.

src/Adapter/MSTest.TestAdapter/Helpers/FixtureMethodRunner.cs
Evangelink: With all the refactorings I don't think it's now hit (outside of special race on cancellation).

Evangelink:
@Youssef1313 I have done multiple rerun and it seems this is no longer flaky (or that the level of flakiness dropped enough)

src/Adapter/MSTest.TestAdapter/Helpers/FixtureMethodRunner.cs
Evangelink: Keeping it for now, I'll try to run more tests to see if we can get rid of it.

PR #4453, opened 2024-12-26
perf in LogMessage

SimonCropp:
I assume logging is in the hot path. so a few more lines for better oerf is justified

move over to using EnumerateInvocationList


src/TestFramework/TestFramework/Logger.cs
```diff
@@ -36,13 +36,34 @@ public static void LogMessage(string format, params object?[] args)
         Guard.NotNull(format);
         Guard.NotNull(args);
 
+        Delegate[] delegates = OnLogMessage.GetInvocationList();
+
+        if (delegates.Length == 0)
```
Youssef1313: Can this ever happen? I'd assume if there is nothing to invoke, we will find OnLogMessage null?

SimonCropp: good point. i will check

SimonCropp: u r correct. i will fix

Evangelink:
I have synced the Polyfill.

Youssef1313:
@Evangelink We need to add the updated version to https://github.com/dotnet/source-build-reference-packages/tree/main/src/textOnlyPackages/src/polyfill/

Youssef1313:
I opened dotnet/source-build-reference-packages#1105
@SimonCropp It may take a while to get that PR merged due to holidays.

src/TestFramework/TestFramework/Logger.cs
```diff
@@ -40,13 +40,12 @@ public static void LogMessage(string format, params object?[] args)
             ? format
             : string.Format(CultureInfo.InvariantCulture, format, args);
 
-        object?[] parameters = [message];
         // Making sure all event handlers are called in sync on same thread.
-        foreach (Delegate invoker in OnLogMessage.GetInvocationList())
+        foreach (LogMessageHandler invoker in DelegatePolyfill.EnumerateInvocationList(OnLogMessage))
```
Youssef1313, line 44: Note: we don't call Logger.LogMessage **at all**. I don't know how much consumers will be calling this, but it doesn't seem to be a hot path.

@Evangelink I'm not sure if there are "valid" use cases for this Logger class or not. If not, we should consider breaking in v4.

SimonCropp: @Youssef1313 want me to burn this PR, and submit another that marks it as obsolete?

Youssef1313: I'll let @Evangelink answer that on whether there are valid use cases for `Logger.LogMessage` or if people are using that API

PR #4456, opened 2024-12-27
MSTEST0037 (proper assert analyzer): Don't report for user defined equality operators

### LINKED ISSUE #4414: MSTEST0037 testing operators == and !=
If you have a class that overloads operators `==` and `!=`, writing the proper unit tests results in `MSTEST0037: Use proper 'Assert' methods`:

```
public sealed class SomeClass
{
    public static Boolean operator ==(SomeClass left, SomeClass right)
    {
        ...
    }

    public static Boolean operator !=(SomeClass left, SomeClass right)
    {
        ...
    }
}
```

[TestMethod]
public void SomeClass_TestOperatorEquals()
{
    var left = new SomeClass(...);
    var right = new SomeClass(...);
    Assert.IsTrue(laft == right); // MSTEST0037
}

Youssef1313:
Fixes #4414


Evangelink (commented):
Does it work if the user operators are in the base class?

Youssef1313:
I think it should. It should be the compiler responsibility to give us the correct OperatorMethod with the correct MethodKind.

Evangelink (approved):

PR #4457, opened 2024-12-27
Add ability to ignore specific test cases

### LINKED ISSUE #1043: How to ignore particular test case (DataRow)
## Description

I am trying to ignore _one_ test case by adding `Ignore` keyword for `DataRow` attribute:

```C#
[TestClass]
public class MathTests
{
    [TestMethod]
    [DataRow(1, 1, 2)]
    [DataRow(2, 2, 3), Ignore]
    public void Sum_Test(int a, int b, int expectedSum)
    {
        var sut = new Math();
        
        var sum = sut.Sum(a, b);
        
        Assert.IsTrue(sum == expectedSum);
    }
}

public class Math
{
    public int Sum(int a, int b)
    {
        return a + b;
    }
}
```

but it ignores the _whole_ test:

image

I also asked this on StackOverflow: [How to ignore particular test case in MSTest](https://stackoverflow.com/questions/71077359/how-to-ignore-particular-test-case-in-mstest).

## Expected behavior
Only `Sum_Test(2,2,3)` test case should be ignored.

## Actual behavior
It ignores both test cases.

## Environment 
- Operating system: Windows 10 Pro
- Target: .NET 6
- MSTest.TestFramework: v2.2.8

### LINKED ISSUE #1411: RFC: Ignore/Skip test or test case
# RFC: Ignore/Skip test or test case

~- [ ] Approved in principle~
- [x] Under discussion
- [ ] Implementation
- [ ] Shipped

## Summary

Improve test/test case skipping mechanism.

## Motivation

Today, it is possible to use [IgnoreAttribute](https://learn.microsoft.com/dotnet/api/microsoft.visualstudio.testtools.unittesting.ignoreattribute), that can be set on a method or on a class to allow ignoring respectively a test method or all test methods of a given class.

There are multiple design issues with this solution:

1. It is possible to set it on any class or method even if they are not a test class or test method. When set in some invalid location, the attribute will have no effect so the impact on user is limited but this is providing a bad user experience. This can also be confusion for users in contexts such as initialization or cleanup methods (`[AssemblyInitialize]`, `[ClassInitialize]`, `[TestInitialize]`, `[AssemblyCleanup]`, `[ClassCleanup]`, `[TestCleanup]`) where it might be thought this would ignore the marked initialize/cleanup method.

1. It is not possible to ignore only specific entries of parameterized tests. For example, it's currently impossible to ignore a `DataRow` entry (see #1043).

## Detailed design

We would like to obsolete the existing attribute and replace it with some property (e.g. `Ignore = "some reason"`) on the various attributes where we want to support this feature. 

By doing so, we ease discoverability as the property is available on the attribute we have already defined and we would fix the design issues raised above.

We would keep the multi-level support:

1. If a test class is marked as ignored (e.g. `[TestClass(Ignore = "...")]`) then all test methods are ignored.

1. If a data test method is marked as ignored (e.g. `[DataTestMethod(Ignore = "...")]`) then all test cases are ignored.

To complete furthermore this feature, we may also want to introduce some new concepts for other data sources:

1. Another property like `IgnoreRows` that would allow to specific a list of rows to skip with either only 1 message or a list of messages (e.g. `[DynamicData(IgnoreRows = ([0, 2, 3], "message"))]` or `[DynamicData(IgnoreRowIndexes = [(0, "message 1"), (2, "message 2"), (3, "message 3")])]`).

1. Replace (or add) the `object[]` part of `IEnumerable<object[]>` by some specific type (we could reuse `DataRow` or introduce some new type) that would allow to skip the entry for all tests using this source.

## Drawbacks

Users would need to update their tests to use the new syntax if we obsolete the old one.

## Alternatives

We could extend current `[Ignore]` attribute to provide properties for rows to ignore.

There aren't lot of requests/complaints about the discoverability or the fact this is not possible to suppress a specific row/entry of a parameterized test.

## Compatibility

Not per se but the goal is to drop the older solution.

### Unresolved questions

None.

Youssef1313:
Fixes #1043 and fixes #1411


src/Adapter/MSTest.TestAdapter/ObjectModel/UnitTestElement.cs
```diff
@@ -33,11 +33,6 @@ public UnitTestElement(TestMethod testMethod)
     /// </summary>
     public TestMethod TestMethod { get; private set; }
 
-    /// <summary>
-    /// Gets or sets a value indicating whether the unit test should be ignored at run-time.
-    /// </summary>
-    public bool Ignored { get; set; }
```
Youssef1313, line 39: This was never read (except in a unit test) and is adding noise to the code. The class is internal so I removed it altogether

src/TestFramework/TestFramework/PublicAPI/PublicAPI.Unshipped.txt
```diff
@@ -1,11 +1,22 @@
...
+Microsoft.VisualStudio.TestTools.UnitTesting.DynamicDataAttribute.Ignore.get -> string?
+Microsoft.VisualStudio.TestTools.UnitTesting.DynamicDataAttribute.Ignore.set -> void
 Microsoft.VisualStudio.TestTools.UnitTesting.DynamicDataSourceType.AutoDetect = 2 -> Microsoft.VisualStudio.TestTools.UnitTesting.DynamicDataSourceType
 *REMOVED*Microsoft.VisualStudio.TestTools.UnitTesting.DynamicDataAttribute.DynamicDataAttribute(string! dynamicDataSourceName, Microsoft.VisualStudio.TestTools.UnitTesting.DynamicDataSourceType dynamicDataSourceType = Microsoft.VisualStudio.TestTools.UnitTesting.DynamicDataSourceType.Property) -> void
 *REMOVED*Microsoft.VisualStudio.TestTools.UnitTesting.DynamicDataAttribute.DynamicDataAttribute(string! dynamicDataSourceName, System.Type! dynamicDataDeclaringType, Microsoft.VisualStudio.TestTools.UnitTesting.DynamicDataSourceType dynamicDataSourceType = Microsoft.VisualStudio.TestTools.UnitTesting.DynamicDataSourceType.Property) -> void
+Microsoft.VisualStudio.TestTools.UnitTesting.ITestDataSourceIgnoreCapability
```
Youssef1313, line 202: TODO: Discuss the naming. Possible options:

- `Ignore`
- `IgnoreReason`
- `IgnoreMessage`

Note that the property on the existing IgnoreAttribute is named `IgnoreMessage`. So we may want to keep the consistency.

Evangelink: I like the `IgnoreMessage` in case we start adding more "options" linked to ignoring a test (e.g. I like NUnit concept of `Until` on the `[Ignore]` attribute). I guess we could also think about creating a `IgnoreOptions` type (I never recall the limitations of types in attributes).

src/TestFramework/TestFramework/Interfaces/ITestDataSourceIgnoreCapability.cs
```diff
@@ -0,0 +1,15 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+namespace Microsoft.VisualStudio.TestTools.UnitTesting;
+
+/// <summary>
+/// Specifies the capability of a test data source to be ignored and define the ignore reason.
+/// </summary>
+public interface ITestDataSourceIgnoreCapability
```
Evangelink, line 9: Could it be worth making this one more generic (not linked to data sources) so we can apply it to test methods and test classes too?

src/TestFramework/TestFramework/Attributes/DataSource/DynamicDataAttribute.cs
```diff
@@ -114,6 +114,11 @@ public DynamicDataAttribute(string dynamicDataSourceName, Type dynamicDataDeclar
     /// <inheritdoc />
     public TestDataSourceUnfoldingStrategy UnfoldingStrategy { get; set; } = TestDataSourceUnfoldingStrategy.Auto;
 
+    /// <summary>
+    /// Gets or sets a reason to ignore this dynamic data source. Setting the property to non-null value will ignore the dynamic data source.
+    /// </summary>
```
Evangelink, line 119:
```suggestion
    /// <inheritdoc />
```

src/TestFramework/TestFramework/Attributes/DataSource/DataRowAttribute.cs
```diff
@@ -53,6 +53,11 @@ public DataRowAttribute(string?[]? stringArrayData)
     /// </summary>
     public string? DisplayName { get; set; }
 
+    /// <summary>
+    /// Gets or sets a reason to ignore the specific test case. Setting the property to non-null value will ignore the test case.
+    /// </summary>
```
Evangelink, line 58:
```suggestion
    /// <inheritdoc />
```

src/Adapter/MSTest.TestAdapter/Execution/TestMethodRunner.cs
```diff
@@ -306,6 +311,20 @@ private bool ExecuteDataSourceBasedTests(List<TestResult> results)
...
                     isDataDriven = true;
+
+                    if (testDataSource is ITestDataSourceIgnoreCapability { Ignore: { } ignoreReason })
+                    {
+                        results.Add(new()
+                        {
+                            // This is closest to ignore. This enum doesn't have a value specific to Ignore.
+                            // It may be a better idea to add a value there, but the enum is public and we need to think more carefully before adding the API.
+                            // For now, TestResultExtensions.ToUnitTestResults method will convert this to Ignored value of ObjectModel enum when IgnoreReason is non-null.
+                            Outcome = UTF.UnitTestOutcome.NotRunnable,
```
Evangelink: It's probably good to have more states but I will have a closer look later.

src/Adapter/MSTest.TestAdapter/Execution/UnitTestRunner.cs
```diff
@@ -347,9 +347,17 @@ private bool IsTestMethodRunnable(
             }
         }
 
+        // TODO: Executor should never be null. Is it incorrectly annotated?
+        string? ignoreMessage = testMethodInfo.TestMethodOptions.Executor?.Ignore ?? testMethodInfo.Parent.ClassAttribute.Ignore;
```
Youssef1313: TODO: I think we should check the class first, then the method.

src/TestFramework/TestFramework/Interfaces/ITestDataSourceIgnoreCapability.cs
Youssef1313: @Evangelink The impl is quite different, at least currently. For example, for specific test cases, we need to have a VSTest property for that to restore the associated ignore message. But we don't need that for when it's the whole test method or test class being ignored.

src/TestFramework/TestFramework/Attributes/DataSource/DynamicDataAttribute.cs
Youssef1313: I was intentionally not inheriting as I'm adjusting the wording a little bit for the specific concrete implementation. It's not super important though as the wording from the interface should still be clear enough for the concrete implementations.

src/Adapter/MSTest.TestAdapter/Execution/TestMethodRunner.cs
Youssef1313: Sure. Let's discuss later

Youssef1313:
@Evangelink This doesn't yet implement a new type that can be specialized under `IEnumerable<object[]>`. I assume this can come in future? If not, we can discuss the design and I'll implement in this PR.

src/Adapter/MSTest.TestAdapter/Execution/TestMethodRunner.cs
Youssef1313: I added UnitTestOutcome.Ignored

Evangelink (approved):

PR #4458, opened 2024-12-27
fix .net install links

SimonCropp:
https://devblogs.microsoft.com/dotnet/critical-dotnet-install-links-are-changing/


Youssef1313 (commented):
I guess such a change should flow from Arcade through an automatic PR from maestro bot.

Files under `eng/common` are not intended to be manually changed.

SimonCropp:
ok. then we can burn this

PR #4460, opened 2024-12-27
Fix MSTEST0017 false positive when both actual/expected are constants

### LINKED ISSUE #4299: MSTEST0017 (Assertion arguments should be passed in the correct order) shouldn't warn if both arguments are literals
## Describe the bug

If both arguments are literals, e.g:

```csharp
Assert.AreEqual(0, 1);
```

we currently issue a warning, but following the codefix will still keep the warning.

We shouldn't issue a diagnostic when "both" arguments look like they should be "expected".

Currently, the above snippet will already produce MSTEST0025 (Use 'Assert.Fail' instead of an always-failing 'Assert.AreEqual' assert).

Youssef1313:
Fixes #4299

### DIFF
--- a/src/Analyzers/MSTest.Analyzers/AssertionArgsShouldBePassedInCorrectOrderAnalyzer.cs
+++ b/src/Analyzers/MSTest.Analyzers/AssertionArgsShouldBePassedInCorrectOrderAnalyzer.cs
@@ -57,6 +57,9 @@ public sealed class AssertionArgsShouldBePassedInCorrectOrderAnalyzer : Diagnost
         });
     }
 
+    private static bool IsConstant(IArgumentOperation argumentOperation)
+        => argumentOperation.Value.ConstantValue.HasValue;
+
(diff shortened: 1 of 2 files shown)


src/Analyzers/MSTest.Analyzers/AssertionArgsShouldBePassedInCorrectOrderAnalyzer.cs
```diff
@@ -57,6 +57,9 @@ public override void Initialize(AnalysisContext context)
         });
     }
 
+    private static bool IsConstant(IArgumentOperation argumentOperation)
+        => argumentOperation.Value.ConstantValue.HasValue;
```
Youssef1313, line 61: NOTE: I removed the check for OperationKind.Literal.
If we have a literal, then ConstantValue.HasValue should always be true. All literals are constants (but not the other way around). The tests are already covering the case where we have a constant which is not literal.

Evangelink (approved):

PR #4467, opened 2024-12-29
Add non-serialization strategy for data unfolding

### LINKED ISSUE #1462: Use a better serialization/deserialization mechanism for data discovery
## Describe the bug

Today, MSTest uses the `DataContractJsonSerializer` serializer in order to perform the serialization and deserialization of the data during discovery and execution.

The problem with current solution is that it requires custom objects to be marked with `DataContractAttribute`/`DataMemberAttribute` which isn't what user expects to do or to be doing for the sake of test.

At the moment, the `BinaryFormatter` seems to be the best option but it's not ideal as this formatter has a lot of security concerns associated (see https://learn.microsoft.com/dotnet/api/system.runtime.serialization.formatters.binary.binaryformatter#remarks)

## Additional context

Related issues:
- MS Test 2.2.4 - 2.2.10: Tests fail because DynamicData gets only structs with default values (#1094)
- wrong value received by test method with ITestDataSource and values of struct with property of nullable primitive type (#1022)
- DynamicData test is not passing exception objects correctly (#1037)
- Using an interface in dynamic data causes a de-/serialization exception (#908)
- DateTime equality test fails after upgrading from 2.2.3 (#1019)
- Using dynamic data source method fails to work (#1588)
- [DataRow]: Original nested types are lost when nested deeper than one level (#2390)
- #3259
- #1054
- #3962

Evangelink:
This pull request includes significant changes to the `MSTest.TestAdapter` and `TestFramework` to introduce a new strategy for identifying test data in data-driven tests. The changes involve updating the test discovery and execution logic to support the new `TestDataIdentifierStrategy`.

Fixes #1462

**IMPORTANT: when using the data index strategy any impact on the order of the data will impact the index which in turn will impact the test ID. This can have important side effect if you are using a system tracking the test ID over time.**

Key changes include:

### Enhancements to Test Discovery:
* Modified the `EnumerateAssembly` method in `AssemblyEnumerator.cs` to include the new `TestDataIdentifierStrategy` when discovering tests. [[1]](diffhunk://#diff-5fd277a76b57620b4fe5c5085dd13fe9a5204139340c42b049edf928f0c36000L89-R90) [[2]](diffhunk://#diff-5fd277a76b57620b4fe5c5085dd13fe9a5204139340c42b049edf928f0c36000R111-R118) [[3]](diffhunk://#diff-5fd277a76b57620b4fe5c5085dd13fe9a5204139340c42b049edf928f0c36000L118-R127)
* Updated the `DiscoverTestsInType` method to pass the `TestDataIdentifierStrategy` to the test discovery process. [[1]](diffhunk://#diff-5fd277a76b57620b4fe5c5085dd13fe9a5204139340c42b049edf928f0c36000R227) [[2]](diffhunk://#diff-5fd277a76b57620b4fe5c5085dd13fe9a5204139340c42b049edf928f0c36000L249-R258)

### Enhancements to Test Execution:
* Added a new method `TryExecuteITestDataSource` in `TestMethodRunner.cs` to handle the execution of tests using the new identifier strategy. [[1]](diffhunk://#diff-264b553419c7a41f96c540062c88e7547bdf2fe95fc6271d2efd17dc19fb76aaL184-R189) [[2]](diffhunk://#diff-264b553419c7a41f96c540062c88e7547bdf2fe95fc6271d2efd17dc19fb76aaR252-R291)

### Interface and Attribute Updates:
* Introduced the `ITestDataIdentifierStrategy` interface and the `TestDataIdentifierStrategy` enum to define the strategy for uniquely identifying test data.
* Updated the `DataRowAttribute` and `DynamicDataAttribute` to implement the new `ITestDataIdentifierStrategy` interface. [[1]](diffhunk://#diff-d2c9ddf08df83d0e90b05ee6c2e7290b5f23e310de0e1dd8e94737d3883efb8cL12-R12) [[2]](diffhunk://#diff-d2c9ddf08df83d0e90b05ee6c2e7290b5f23e310de0e1dd8e94737d3883efb8cR59-R61) [[3]](diffhunk://#diff-ef817dcecca2469992a54f4e6c654e4c381fbcb72654f50d5054260b62f09be6L35-R35) [[4]](diffhunk://#diff-ef817dcecca2469992a54f4e6c654e4c381fbcb72654f50d5054260b62f09be6R117-R119)
* Enhanced the `TestDataSourceOptionsAttribute` to include the new identifier strategy.

### TODOs:
- [ ] Add tests
- [ ] Add documentation
- [x] Finish todo exceptions (or failed state?)
- [x] Better explain consequences on test ID
- [x] Better explain impact when ordering changes

### DIFF
--- a/src/Adapter/MSTest.TestAdapter/Discovery/AssemblyEnumerator.cs
+++ b/src/Adapter/MSTest.TestAdapter/Discovery/AssemblyEnumerator.cs
@@ -86,10 +86,11 @@ internal class AssemblyEnumerator : MarshalByRefObject
         DataRowAttribute.TestIdGenerationStrategy = testIdGenerationStrategy;
         DynamicDataAttribute.TestIdGenerationStrategy = testIdGenerationStrategy;
 
-        TestDataSourceUnfoldingStrategy dataSourcesUnfoldingStrategy = ReflectHelper.GetTestDataSourceOptions(assembly)?.UnfoldingStrategy switch
+        TestDataSourceOptionsAttribute? testDataSourceOptionsAttribute = ReflectHelper.GetTestDataSourceOptions(assembly);
+        TestDataSourceUnfoldingStrategy dataSourcesUnfoldingStrategy = testDataSourceOptionsAttribute?.UnfoldingStrategy switch
         {
             // When strategy is auto we want to unfold
-            TestDataSourceUnfoldingStrategy.Auto => TestDataSourceUnfoldingStrategy.Unfold,
+            TestDataSourceUnfoldingStrategy.Auto => TestDataSourceUnfoldingStrategy.UnfoldUsingDataContractJsonSerializer,
             // When strategy is set, let's use it
             { } value => value,
             // When the attribute is not set, let's look at the legacy attribute
@@ -103,7 +104,7 @@ internal class AssemblyEnumerator : MarshalByRefObject
                 // as the ID generator will ignore it.
                 (null, TestIdGenerationStrategy.Legacy) => TestDataSourceUnfoldingStrategy.Fold,
 #pragma warning restore CS0618 // Type or member is obsolete
-                _ => TestDataSourceUnfoldingStrategy.Unfold,
+                _ => TestDataSourceUnfoldingStrategy.UnfoldUsingDataContractJsonSerializer,
             },
         };
 
@@ -115,8 +116,7 @@ internal class AssemblyEnumerator : MarshalByRefObject
                 continue;
             }
 
-            List<UnitTestElement> testsInType = DiscoverTestsInType(assemblyFileName, testRunParametersFromRunSettings, type, warnings, discoverInternals,
-                dataSourcesUnfoldingStrategy, testIdGenerationStrategy, fixturesTests);
+            List<UnitTestElement> testsInType = DiscoverTestsInType(assemblyFileName, testRunParametersFromRunSettings, type, warnings, discoverInternals, dataSourcesUnfoldingStrategy, testIdGenerationStrategy, fixturesTests);
             tests.AddRange(testsInType);
         }
 
@@ -367,10 +367,12 @@ internal class AssemblyEnumerator : MarshalByRefObject
         try
         {
             bool isDataDriven = false;
+            int dataSourceIndex = -1;
             foreach (ITestDataSource dataSource in testDataSources)
             {
+                dataSourceIndex++;
                 isDataDriven = true;
-                if (!TryUnfoldITestDataSource(dataSource, dataSourcesUnfoldingStrategy, test, new(testMethodInfo.MethodInfo, test.DisplayName), tempListOfTests))
+                if (!TryUnfoldITestDataSource(dataSource, dataSourceIndex, dataSourcesUnfoldingStrategy, test, new(testMethodInfo.MethodInfo, test.DisplayName), tempListOfTests))
                 {
                     // TODO: Improve multi-source design!
                     // Ideally we would want to consider each data source separately but when one source cannot be expanded,
@@ -400,7 +402,7 @@ internal class AssemblyEnumerator : MarshalByRefObject
         }
     }
 
-    private static bool TryUnfoldITestDataSource(ITestDataSource dataSource, TestDataSourceUnfoldingStrategy dataSourcesUnfoldingStrategy, UnitTestElement test, ReflectionTestMethodInfo methodInfo, List<UnitTestElement> tests)
+    private static bool TryUnfoldITestDataSource(ITestDataSource dataSource, int dataSourceIndex, TestDataSourceUnfoldingStrategy dataSourcesUnfoldingStrategy, UnitTestElement test, ReflectionTestMethodInfo methodInfo, List<UnitTestElement> tests)
     {
         var unfoldingCapability = dataSource as ITestDataSourceUnfoldingCapability;
 
@@ -455,6 +457,7 @@ internal class AssemblyEnumerator : MarshalByRefObject
             // If strategy is DisplayName and we have a duplicate test name don't expand the test, bail out.
 #pragma warning disable CS0618 // Type or member is obsolete
             if (test.TestMethod.TestIdGenerationStrategy == TestIdGenerationStrategy.DisplayName
+#pragma warning restore CS0618 // Type or member is obsolete
                 && testDisplayNameFirstSeen.TryGetValue(discoveredTest.DisplayName!, out int firstIndexSeen))
             {
                 string warning = string.Format(CultureInfo.CurrentCulture, Resource.CannotExpandIDataSourceAttribute_DuplicateDisplayName, firstIndexSeen, index, discoveredTest.DisplayName);
@@ -464,24 +467,38 @@ internal class AssemblyEnumerator : MarshalByRefObject
                 // Duplicated display name so bail out. Caller will handle adding the original test.
                 return false;
             }
-#pragma warning restore CS0618 // Type or member is obsolete
 
-            try
+            if (unfoldingCapability?.UnfoldingStrategy == TestDataSourceUnfoldingStrategy.UnfoldUsingDataIndex
+                || (unfoldingCapability?.UnfoldingStrategy is null or TestDataSourceUnfoldingStrategy.Auto
+                    && dataSourcesUnfoldingStrategy == TestDataSourceUnfoldingStrategy.UnfoldUsingDataIndex))
             {
-                discoveredTest.TestMethod.SerializedData = DataSerializationHelper.Serialize(d);
-                discoveredTest.TestMethod.TestDataSourceIgnoreMessage = testDataSourceIgnoreMessage;
+                discoveredTest.TestMethod.SerializedData = new string[3]
+                {
+                    TestDataSourceUnfoldingStrategy.UnfoldUsingDataIndex.ToString(),
+                    dataSourceIndex.ToString(CultureInfo.InvariantCulture),
+                    index.ToString(CultureInfo.InvariantCulture),
+                };
                 discoveredTest.TestMethod.DataType = DynamicDataType.ITestDataSource;
             }
-            catch (SerializationException ex)
+            else
             {
-                string warning = string.Format(CultureInfo.CurrentCulture, Resource.CannotExpandIDataSourceAttribute_CannotSerialize, index, discoveredTest.DisplayName);
-                warning += Environment.NewLine;
-                warning += ex.ToString();
-                warning = string.Format(CultureInfo.CurrentCulture, Resource.CannotExpandIDataSourceAttribute, test.TestMethod.ManagedTypeName, test.TestMethod.ManagedMethodName, warning);
-                PlatformServiceProvider.Instance.AdapterTraceLogger.LogWarning($"DynamicDataEnumerator: {warning}");
+                try
+                {
+                    discoveredTest.TestMethod.SerializedData = DataSerializationHelper.Serialize(d);
+                    discoveredTest.TestMethod.TestDataSourceIgnoreMessage = testDataSourceIgnoreMessage;
+                    discoveredTest.TestMethod.DataType = DynamicDataType.ITestDataSource;
+                }
+                catch (SerializationException ex)
+                {
+                    string warning = string.Format(CultureInfo.CurrentCulture, Resource.CannotExpandIDataSourceAttribute_CannotSerialize, index, discoveredTest.DisplayName);
+                    warning += Environment.NewLine;
+                    warning += ex.ToString();
+                    warning = string.Format(CultureInfo.CurrentCulture, Resource.CannotExpandIDataSourceAttribute, test.TestMethod.ManagedTypeName, test.TestMethod.ManagedMethodName, warning);
+                    PlatformServiceProvider.Instance.AdapterTraceLogger.LogWarning($"DynamicDataEnumerator: {warning}");
 
-                // Serialization failed for the type, bail out. Caller will handle adding the original test.
-                return false;
+                    // Serialization failed for the type, bail out. Caller will handle adding the original test.
+                    return false;
+                }
             }
 
             discoveredTests.Add(discoveredTest);
--- a/src/Adapter/MSTest.TestAdapter/Execution/TestMethodRunner.cs
+++ b/src/Adapter/MSTest.TestAdapter/Execution/TestMethodRunner.cs
@@ -169,22 +169,11 @@ internal sealed class TestMethodRunner
 
         bool isDataDriven = false;
         var parentStopwatch = Stopwatch.StartNew();
-        if (_test.DataType == DynamicDataType.ITestDataSource)
+        if (TryExecuteITestDataSource(results))
         {
-            if (_test.TestDataSourceIgnoreMessage is not null)
-            {
-                return [new() { Outcome = UTF.UnitTestOutcome.Ignored, IgnoreReason = _test.TestDataSourceIgnoreMessage }];
-            }
-
-            object?[]? data = DataSerializationHelper.Deserialize(_test.SerializedData);
-            TestResult[] testResults = ExecuteTestWithDataSource(null, data);
-            results.AddRange(testResults);
         }
-        else if (TryExecuteDataSourceBasedTests(results))
-        {
-            isDataDriven = true;
-        }
-        else if (TryExecuteFoldedDataDrivenTests(results))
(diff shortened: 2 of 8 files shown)


src/TestFramework/TestFramework/Interfaces/ITestDataIdentifierStrategy.cs
```diff
@@ -0,0 +1,40 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+namespace Microsoft.VisualStudio.TestTools.UnitTesting;
+
+/// <summary>
+/// Defines the strategy for uniquely identifying test data.
+/// This is only used when <see cref="ITestDataSourceUnfoldingCapability.UnfoldingStrategy" /> is set to <see cref="TestDataSourceUnfoldingStrategy.Unfold"/>.
+/// </summary>
+public interface ITestDataIdentifierStrategy
```
Evangelink: Given this enum and interface only work when using the unfolding strategy, I was wondering if we should merge the 2 together so we have something like `Auto`, `Fold`, `UnfoldUsingDataContractSerialization`, `UnfoldUsingDataIndex`.

WDYT?

src/Adapter/MSTest.TestAdapter/Execution/TestMethodRunner.cs
```diff
@@ -254,6 +249,46 @@ internal UnitTestResult[] RunTestMethod()
...
+        UTF.ITestDataSource? dataSource;
+        object?[]? data;
+        if (_test.SerializedData?.Length == 3
+            && Enum.TryParse(_test.SerializedData[0], out TestDataIdentifierStrategy _))
+        {
+            if (!int.TryParse(_test.SerializedData[1], out int dataSourceIndex)
+                || !int.TryParse(_test.SerializedData[2], out int dataIndex))
+            {
+                // TODO
+                throw new InvalidOperationException();
```
Evangelink: Shall I handle gracefully or throw unreachable?

src/Adapter/MSTest.TestAdapter/Execution/TestMethodRunner.cs
```diff
@@ -254,6 +249,46 @@ internal UnitTestResult[] RunTestMethod()
...
+            }
+
+            dataSource = _testMethodInfo.GetAttributes<Attribute>(false)?.OfType<UTF.ITestDataSource>().Skip(dataSourceIndex).FirstOrDefault();
+            if (dataSource is null)
+            {
+                // TODO
+                throw new InvalidOperationException();
```
Evangelink: Shall I handle gracefully or throw unreachable?

src/Adapter/MSTest.TestAdapter/Discovery/AssemblyEnumerator.cs
```diff
@@ -463,23 +475,37 @@ private static bool TryUnfoldITestDataSource(ITestDataSource dataSource, TestDat
...
             }
-#pragma warning restore CS0618 // Type or member is obsolete
 
-            try
+            TestDataIdentifierStrategy currentDataSourceIdentifierStrategy = (dataSource as ITestDataIdentifierStrategy)?.IdentifierStrategy ?? TestDataIdentifierStrategy.Auto;
+            if (currentDataSourceIdentifierStrategy == TestDataIdentifierStrategy.DataIndex
+                || (currentDataSourceIdentifierStrategy == TestDataIdentifierStrategy.Auto && dataSourcesIdentifierStrategy == TestDataIdentifierStrategy.DataIndex))
             {
-                discoveredTest.TestMethod.SerializedData = DataSerializationHelper.Serialize(d);
+                discoveredTest.TestMethod.SerializedData = new string[3]
```
Evangelink, line 475: I am reusing the serialized data property as it's not used and allows me to avoid adding more properties to the TestCase object as it has some perf impact. This isn't perfect but as far as I can tell, there is no way for the serialized data to be serialized this way as we prefix with the data type.

src/TestFramework/TestFramework/Interfaces/ITestDataIdentifierStrategy.cs
Youssef1313: @Evangelink You mean not needing the interface at all, and only relying on the existing enum? This is fine for me.

We will have to keep `Unfold` and make it the same as `UnfoldUsingDataContractSerialization` for backcompat I assume. But that's fine.

src/Adapter/MSTest.TestAdapter/Execution/TestMethodRunner.cs
Youssef1313: I think this should be unreachable. Otherwise something very bad happened.

If SerializedData is coming through data contract serialization, the length must be even. So, we are sure if the length is 3, it should be coming through the new alternative. In this case, we should always be able to correctly parse everything back.

I'd even move the `Enum.TryParse` check from the outer if to the inner if (negated, of course)

src/Adapter/MSTest.TestAdapter/Execution/TestMethodRunner.cs
```diff
@@ -254,6 +249,46 @@ internal UnitTestResult[] RunTestMethod()
...
+            && Enum.TryParse(_test.SerializedData[0], out TestDataIdentifierStrategy _))
+        {
+            if (!int.TryParse(_test.SerializedData[1], out int dataSourceIndex)
+                || !int.TryParse(_test.SerializedData[2], out int dataIndex))
+            {
+                // TODO
+                throw new InvalidOperationException();
+            }
+
+            dataSource = _testMethodInfo.GetAttributes<Attribute>(false)?.OfType<UTF.ITestDataSource>().Skip(dataSourceIndex).FirstOrDefault();
```
Youssef1313: It may be a good idea to check with runtime team that it's safe to assume reflection will return attributes in a deterministic order. I *think* it should be a safe assumption, but just a sanity check.

src/Adapter/MSTest.TestAdapter/Execution/TestMethodRunner.cs
Youssef1313: I think it shouldn't happen?

src/TestFramework/TestFramework/Interfaces/ITestDataIdentifierStrategy.cs
```diff
@@ -0,0 +1,40 @@
...
+
+    /// <summary>
+    /// Identifies test data using data contract serialization.
+    /// </summary>
+    DataContractSerialization,
+
+    /// <summary>
+    /// Identifies test data by its index in the data source.
+    /// </summary>
+    DataIndex,
```
Youssef1313: In the context of HotReload, if an attribute is removed, re-ordered, etc, is everything going to work fine?

src/Adapter/MSTest.TestAdapter/Discovery/AssemblyEnumerator.cs
Youssef1313: Fine for me to reuse. At least until we have a strong reason not to reuse.

src/TestFramework/TestFramework/Interfaces/ITestDataIdentifierStrategy.cs
Evangelink: Will double check that scenario!

src/Adapter/MSTest.TestAdapter/Execution/TestMethodRunner.cs
Evangelink: From my tests it is deterministic but yes I'll double check with them.

src/TestFramework/TestFramework/Interfaces/ITestDataIdentifierStrategy.cs
Evangelink: Yes that's what I meant. My last commit is doing the change, please review and let me know if you think that's better or worse.

src/TestFramework/TestFramework/PublicAPI/PublicAPI.Unshipped.txt
```diff
@@ -1,11 +1,26 @@
...
+Microsoft.VisualStudio.TestTools.UnitTesting.TestDataIdentifierStrategy.Auto = 0 -> Microsoft.VisualStudio.TestTools.UnitTesting.TestDataIdentifierStrategy
+Microsoft.VisualStudio.TestTools.UnitTesting.TestDataIdentifierStrategy.DataContractSerialization = 1 -> Microsoft.VisualStudio.TestTools.UnitTesting.TestDataIdentifierStrategy
+Microsoft.VisualStudio.TestTools.UnitTesting.TestDataIdentifierStrategy.DataIndex = 2 -> Microsoft.VisualStudio.TestTools.UnitTesting.TestDataIdentifierStrategy
+Microsoft.VisualStudio.TestTools.UnitTesting.TestDataSourceOptionsAttribute.IdentifierStrategy.get -> Microsoft.VisualStudio.TestTools.UnitTesting.TestDataIdentifierStrategy
+Microsoft.VisualStudio.TestTools.UnitTesting.TestDataSourceOptionsAttribute.TestDataSourceOptionsAttribute(Microsoft.VisualStudio.TestTools.UnitTesting.TestDataIdentifierStrategy identifierStrategy) -> void
+Microsoft.VisualStudio.TestTools.UnitTesting.TestDataSourceOptionsAttribute.TestDataSourceOptionsAttribute(Microsoft.VisualStudio.TestTools.UnitTesting.TestDataSourceUnfoldingStrategy unfoldingStrategy, Microsoft.VisualStudio.TestTools.UnitTesting.TestDataIdentifierStrategy identifierStrategy) -> void
```
Evangelink:
```suggestion
```

src/TestFramework/TestFramework/PublicAPI/PublicAPI.Unshipped.txt
```diff
@@ -1,11 +1,26 @@
 #nullable enable
+Microsoft.VisualStudio.TestTools.UnitTesting.DataRowAttribute.IdentifierStrategy.get -> Microsoft.VisualStudio.TestTools.UnitTesting.TestDataIdentifierStrategy
+Microsoft.VisualStudio.TestTools.UnitTesting.DataRowAttribute.IdentifierStrategy.set -> void
```
Evangelink:
```suggestion
```

src/TestFramework/TestFramework/PublicAPI/PublicAPI.Unshipped.txt
```diff
@@ -1,11 +1,26 @@
...
 Microsoft.VisualStudio.TestTools.UnitTesting.DynamicDataAttribute.DynamicDataAttribute(string! dynamicDataSourceName) -> void
 Microsoft.VisualStudio.TestTools.UnitTesting.DynamicDataAttribute.DynamicDataAttribute(string! dynamicDataSourceName, Microsoft.VisualStudio.TestTools.UnitTesting.DynamicDataSourceType dynamicDataSourceType) -> void
 Microsoft.VisualStudio.TestTools.UnitTesting.DynamicDataAttribute.DynamicDataAttribute(string! dynamicDataSourceName, System.Type! dynamicDataDeclaringType) -> void
 Microsoft.VisualStudio.TestTools.UnitTesting.DynamicDataAttribute.DynamicDataAttribute(string! dynamicDataSourceName, System.Type! dynamicDataDeclaringType, Microsoft.VisualStudio.TestTools.UnitTesting.DynamicDataSourceType dynamicDataSourceType) -> void
+Microsoft.VisualStudio.TestTools.UnitTesting.DynamicDataAttribute.IdentifierStrategy.get -> Microsoft.VisualStudio.TestTools.UnitTesting.TestDataIdentifierStrategy
+Microsoft.VisualStudio.TestTools.UnitTesting.DynamicDataAttribute.IdentifierStrategy.set -> void
```
Evangelink:
```suggestion
```

src/Adapter/MSTest.TestAdapter/Execution/TestMethodRunner.cs
```diff
@@ -254,6 +249,44 @@ internal UnitTestResult[] RunTestMethod()
...
+            return false;
+        }
+
+        UTF.ITestDataSource? dataSource;
+        object?[]? data;
+        if (_test.SerializedData?.Length == 3)
+        {
+            if (!Enum.TryParse(_test.SerializedData[0], out TestDataSourceUnfoldingStrategy _)
+                || !int.TryParse(_test.SerializedData[1], out int dataSourceIndex)
+                || !int.TryParse(_test.SerializedData[2], out int dataIndex))
```
Youssef1313: Consider using the `IFormatProvider` overload and pass `CultureInfo.InvariantCulture`.

src/TestFramework/TestFramework/Interfaces/ITestDataSourceUnfoldingCapability.cs
```diff
@@ -22,17 +22,34 @@ public enum TestDataSourceUnfoldingStrategy : byte
...
     /// <summary>
     /// The parameterized test is not unfolded; all data rows are treated as a single test case.
     /// </summary>
     Fold,
+
+    /// <summary>
+    /// Each data row is treated as a separate test case, and the data is unfolded using
+    /// <see cref="System.Runtime.Serialization.Json.DataContractJsonSerializer"/>.
+    /// </summary>
+    UnfoldUsingDataContractJsonSerializer,
```
Youssef1313, line 45: Should we make this the same numeric value as `Unfold`?

src/TestFramework/TestFramework/Interfaces/ITestDataSourceUnfoldingCapability.cs
```diff
@@ -22,17 +22,34 @@ public enum TestDataSourceUnfoldingStrategy : byte
...
     /// <see cref="TestDataSourceOptionsAttribute" />. If no assembly level attribute is specified, then the default
-    /// configuration is to unfold.
+    /// configuration is to unfold using <see cref="System.Runtime.Serialization.Json.DataContractJsonSerializer"/>.
     /// </summary>
     Auto,
 
     /// <summary>
     /// Each data row is treated as a separate test case.
     /// </summary>
+    [Obsolete("Use 'UnfoldUsingDataContractJsonSerializer' instead")]
```
Youssef1313, line 33: Consider EditorBrowsable(Never), and maybe also inheritdoc from `UnfoldUsingDataContractJsonSerializer`

src/TestFramework/TestFramework/Interfaces/ITestDataIdentifierStrategy.cs
Youssef1313: Looks better to me

src/Adapter/MSTest.TestAdapter/Execution/TestMethodRunner.cs
Evangelink: I'm surprised there was no CA warning here. I need to verify the existing rules.

nohwnd:
This is interesting. Have you considered implementing in a way that is more abstracted from the execution, like we discussed (and you realized it) in tatf? Where there was an underlying concept of data entry + string key that represented single test case, and the index based key way just representation of that. 

I haven't dug into this as much as you did so please let me know if or why it is not possible or desirable (e.g you want to do it without braking changes).

Evangelink:
Yes the key based is ideal but requires way more changes in our infra so I wanted to start with index. I would expect in the long run to keep only key based and serialization (with a change to `System.Text.Json` most likely).

src/TestFramework/TestFramework/Interfaces/ITestDataSourceUnfoldingCapability.cs
Evangelink: It's probably not important, is it?

src/Adapter/MSTest.TestAdapter/Execution/TestMethodRunner.cs
```diff
@@ -526,4 +515,42 @@ private static List<TestResult> UpdateResultsWithParentInfo(
...
+            {
+                throw ApplicationStateGuard.Unreachable();
+            }
+
+            dataSource = _testMethodInfo.GetAttributes<Attribute>(false)?.OfType<UTF.ITestDataSource>().Skip(dataSourceIndex).FirstOrDefault();
```
Evangelink, line 537: Runtime team just told me not to rely on guaranteed reflection order. I'll try to see what other option we have. Given that this is currently always working fine, I'd vote for adding a comment and moving on with the PR.

PR #4476, opened 2024-12-30
Introduce interpolated string handler overloads for assertions

### LINKED ISSUE #4273: Add interpolated string handler overloads for Assert methods
## Summary

It's [common](https://grep.app/search?q=%5CsAssert%5C..%2B%3F%5C%28.%2B%3F%2C.%2B%3F%5C%24%22&regexp=true) to have `Assert.SomeMethod(arguments, message: $"... {something} ... etc");`.
The message is only needed when the assert actually fails, which hopefully isn't the common case when running tests :)
Currently, we are paying the cost to construct the string while it ends up not being used. That is exactly what interpolated string handlers tries to solve.

## Background and Motivation

Introducing new interpolated string handlers overloads will solve the performance costs for this scenario.

See https://learn.microsoft.com/dotnet/csharp/advanced-topics/performance/interpolated-string-handler for info.

Youssef1313:
Fixes #4273


src/TestFramework/TestFramework/Assertions/Assert.AreEqual.cs
```diff
@@ -10,6 +12,256 @@ namespace Microsoft.VisualStudio.TestTools.UnitTesting;
 /// </summary>
 public sealed partial class Assert
 {
+    [InterpolatedStringHandler]
+    [EditorBrowsable(EditorBrowsableState.Never)]
+    public readonly struct AssertAreEqualInterpolatedStringHandler<TArgument>
```
Evangelink, line 17: On the official documentation, they use `ref struct` is it working when not `ref`?

src/TestFramework/TestFramework/Assertions/Assert.AreEqual.cs
```diff
@@ -10,6 +12,256 @@ namespace Microsoft.VisualStudio.TestTools.UnitTesting;
...
+        {
+            if (_builder is not null)
+            {
+                FailAreEqual(_expected, _actual, _builder.ToString());
+            }
+        }
+
+        public readonly void AppendLiteral(string value) => _builder!.Append(value);
+
+        public readonly void AppendFormatted<T>(T value) => _builder!.Append(value);
```
Evangelink: Shall we have `AppendFormatted<T>(T t, string format) where T : IFormattable`?

src/TestFramework/TestFramework/Assertions/Assert.AreEqual.cs
```diff
@@ -10,6 +12,256 @@ namespace Microsoft.VisualStudio.TestTools.UnitTesting;
...
+
+        internal void FailIfNeeded()
+        {
+            if (_builder is not null)
+            {
+                FailAreEqual(_expected, _actual, _builder.ToString());
+            }
+        }
+
+        public readonly void AppendLiteral(string value) => _builder!.Append(value);
```
Evangelink: I think you can drop these `readonly` as the struct is alreadyd `readonly`.

src/TestFramework/TestFramework/Assertions/Assert.AreEqual.cs
```diff
@@ -10,6 +12,256 @@ namespace Microsoft.VisualStudio.TestTools.UnitTesting;
...
+
+        public readonly void AppendFormatted<T>(T value) => _builder!.Append(value);
+
+#if NETCOREAPP3_1_OR_GREATER
+        public readonly void AppendFormatted(ReadOnlySpan<char> value) => _builder!.Append(value.ToString());
```
Evangelink: Why calling `ToString` when the `Append` has an overload with `ReadOnlySpan<char>`?

Evangelink: Shall we also add `ReadOnlyMemory<char>` like `Append` has?

src/TestFramework/TestFramework/Assertions/Assert.AreEqual.cs
```diff
@@ -10,6 +12,256 @@ namespace Microsoft.VisualStudio.TestTools.UnitTesting;
...
+            shouldAppend = AreEqualFailing(expected, actual);
+            if (shouldAppend)
+            {
+                _builder = new StringBuilder(literalLength + formattedCount);
+                _expected = expected;
+                _actual = actual;
+            }
+        }
+
+        internal void FailIfNeeded()
```
Evangelink: What about either a generic name like `ComputeAssertion` or `Assert`? Or a more specialized name like `AssertAreEqual`?

src/TestFramework/TestFramework/Assertions/Assert.AreEqual.cs
```diff
@@ -272,17 +525,44 @@ public static void AreEqual<T>(IEquatable<T>? expected, IEquatable<T>? actual, s
...
+            Math.Abs(expected - actual) > delta;
+
+    private static bool AreEqualFailing(decimal expected, decimal actual, decimal delta)
+        => Math.Abs(expected - actual) > delta;
+
+    private static bool AreEqualFailing(long expected, long actual, long delta)
+        => Math.Abs(expected - actual) > delta;
+
+    [DoesNotReturn]
+    private static void FailAreEqual(object? expected, object? actual, string userMessage)
```
Evangelink: Let's use a naming like `ThrowAssertAreEqualFailed` or `ThrowAreEqualFailed`.

Evangelink (commented):
Could you please apply the various suggestions on `AreEqual` globally?

src/TestFramework/TestFramework/Assertions/Assert.AreEqual.cs
Youssef1313: Some handlers of the BCL are using ref struct, some others are not. At least for Debug.Assert, it's not a ref struct.

But it's worth understanding and making sure we are doing the right thing for our scenario.

src/TestFramework/TestFramework/Assertions/Assert.AreEqual.cs
Youssef1313: Indeed.

image

Youssef1313: Code snippet from BCL:

I think it's safest to include all these overloads of AppendFormatted

```csharp
            /// <summary>Writes the specified string to the handler.</summary>
            /// <param name="value">The string to write.</param>
            public void AppendLiteral(string value) => _stringBuilderHandler.AppendLiteral(value);

            /// <summary>Writes the specified value to the handler.</summary>
            /// <param name="value">The value to write.</param>
            /// <typeparam name="T">The type of the value to write.</typeparam>
            public void AppendFormatted<T>(T value) => _stringBuilderHandler.AppendFormatted(value);

            /// <summary>Writes the specified value to the handler.</summary>
            /// <param name="value">The value to write.</param>
            /// <param name="format">The format string.</param>
            /// <typeparam name="T">The type of the value to write.</typeparam>
            public void AppendFormatted<T>(T value, string? format) => _stringBuilderHandler.AppendFormatted(value, format);

            /// <summary>Writes the specified value to the handler.</summary>
            /// <param name="value">The value to write.</param>
            /// <param name="alignment">Minimum number of characters that should be written for this value.  If the value is negative, it indicates left-aligned and the required minimum is the absolute value.</param>
            /// <typeparam name="T">The type of the value to write.</typeparam>
            public void AppendFormatted<T>(T value, int alignment) => _stringBuilderHandler.AppendFormatted(value, alignment);

            /// <summary>Writes the specified value to the handler.</summary>
            /// <param name="value">The value to write.</param>
            /// <param name="format">The format string.</param>
            /// <param name="alignment">Minimum number of characters that should be written for this value.  If the value is negative, it indicates left-aligned and the required minimum is the absolute value.</param>
            /// <typeparam name="T">The type of the value to write.</typeparam>
            public void AppendFormatted<T>(T value, int alignment, string? format) => _stringBuilderHandler.AppendFormatted(value, alignment, format);

            /// <summary>Writes the specified character span to the handler.</summary>
            /// <param name="value">The span to write.</param>
            public void AppendFormatted(ReadOnlySpan<char> value) => _stringBuilderHandler.AppendFormatted(value);

            /// <summary>Writes the specified string of chars to the handler.</summary>
            /// <param name="value">The span to write.</param>
            /// <param name="alignment">Minimum number of characters that should be written for this value.  If the value is negative, it indicates left-aligned and the required minimum is the absolute value.</param>
            /// <param name="format">The format string.</param>
            public void AppendFormatted(ReadOnlySpan<char> value, int alignment = 0, string? format = null) => _stringBuilderHandler.AppendFormatted(value, alignment, format);

            /// <summary>Writes the specified value to the handler.</summary>
            /// <param name="value">The value to write.</param>
            public void AppendFormatted(string? value) => _stringBuilderHandler.AppendFormatted(value);

            /// <summary>Writes the specified value to the handler.</summary>
            /// <param name="value">The value to write.</param>
            /// <param name="alignment">Minimum number of characters that should be written for this value.  If the value is negative, it indicates left-aligned and the required minimum is the absolute value.</param>
            /// <param name="format">The format string.</param>
            public void AppendFormatted(string? value, int alignment = 0, string? format = null) => _stringBuilderHandler.AppendFormatted(value, alignment, format);

            /// <summary>Writes the specified value to the handler.</summary>
            /// <param name="value">The value to write.</param>
            /// <param name="alignment">Minimum number of characters that should be written for this value.  If the value is negative, it indicates left-aligned and the required minimum is the absolute value.</param>
            /// <param name="format">The format string.</param>
            public void AppendFormatted(object? value, int alignment = 0, string? format = null) => _stringBuilderHandler.AppendFormatted(value, alignment, format);
```

src/TestFramework/TestFramework/Assertions/Assert.AreSame.cs
```diff
@@ -10,6 +12,74 @@ namespace Microsoft.VisualStudio.TestTools.UnitTesting;
...
+                FailAreSame(_expected, _actual, _builder.ToString());
+            }
+        }
+
+        public void AppendLiteral(string value) => _builder!.Append(value);
+
+        public void AppendFormatted<T>(T value) => _builder!.Append(value);
+
+#if NETCOREAPP3_1_OR_GREATER
+        public void AppendFormatted(ReadOnlySpan<char> value) => _builder!.Append(value);
```
Youssef1313, line 47: We may have problems if people are using ROS in .NET Framework via NuGet package. Not sure if it's acceptable to tell them "call `.ToString()`", or if we should take an extra dependency for netstandard/netframework

Youssef1313: No concern here. ROS (ref structs in general) already won't work on .NET Framework in an interpolated string hole. The compiler will complain that it cannot box it and will not automagically call `.ToString()`

src/TestFramework/TestFramework/Assertions/Assert.AreEqual.cs
Youssef1313: I don't think we should. This should fall onto the generic overload.

ReadOnlySpan can't use that overload because it's a ref struct, so it cannot be used as a generic type parameter.

src/TestFramework/TestFramework/Assertions/Assert.AreEqual.cs
Youssef1313: Added that overload, though it shouldn't have `where T : IFormattable` to allow enums and ISpanFormattable.

src/TestFramework/TestFramework/Assertions/Assert.AreEqual.cs
Youssef1313: We shouldn't use a ref struct. I added tests that cover us if this is a ref struct. The tests I added will fail to compile altogether.

Youssef1313:
Thanks for the review @Evangelink. Those comments were really very helpful and revealed some important issues.

src/TestFramework/TestFramework/Assertions/Assert.AreSame.cs
```diff
@@ -10,6 +12,120 @@ namespace Microsoft.VisualStudio.TestTools.UnitTesting;
...
+        {
+            if (_builder is not null)
+            {
+                ThrowAssertAreSameFailed(_expected, _actual, _builder.ToString());
+            }
+        }
+
+        public void AppendLiteral(string value) => _builder!.Append(value);
+
+        public void AppendFormatted<T>(T value) => AppendFormatted(value, format: null);
```
Youssef1313, line 44: I should be adding `AppendFormatted(string?)` overload. This should be resolved before merging.

src/TestFramework/TestFramework/Assertions/Assert.AreEqual.cs
```diff
@@ -80,6 +432,12 @@ public static void AreEqual<T>(T? expected, T? actual, IEqualityComparer<T>? com
     public static void AreEqual<T>(T? expected, T? actual, string? message)
         => AreEqual(expected, actual, null, message, null);
 
+    /// <inheritdoc cref="AreEqual{T}(IEquatable{T}?, IEquatable{T}?, string?)" />
+#pragma warning disable IDE0060 // Remove unused parameter - https://github.com/dotnet/roslyn/issues/76578
+    public static void AreEqual<T>(T? expected, T? actual, [InterpolatedStringHandlerArgument(nameof(expected), nameof(actual))] ref AssertAreEqualInterpolatedStringHandler<T> message)
```
Evangelink, line 437: I don't know if runtime calls this parameter `message`, I would prefer `handler` or `builder` like in the example.

Evangelink (approved):
Amazing work. I didn't review every single test added but I assume you have used the same pattern everywhere so LGTM

src/TestFramework/TestFramework/Assertions/Assert.AreEqual.cs
Youssef1313: Hmmm. My idea was for existing code written on the form `Assert.AreEqual(expected, actual, message: $"... {...} ...")` to pick up the new overload. But looking at `StringBuilder`, it wasn't done this way. However, for `Debug.Assert`, the BCL reused the same parameter name `message`.

I have no idea if this was a very explicit design design, and why they decided to use different approach for StringBuilder vs Debug.Assert

Evangelink: Ok that works for me

PR #4482, opened 2024-12-31
Allow DynamicData source to be on base types

### LINKED ISSUE #4347: Do not require member type in [DynamicData] when member is defined on base class
## Summary

Make usage as simple as possible when member info can be inferred.

## Background and Motivation

The use case below should be simplified:

```cs
public class Base
{
    public static IEnumerable<object[]> GetData() => yield return new object[] { 1 };
}

[TestClass]
public class MyTest : Base
{
    [TestMethod]
    [DynamicData(nameof(GetData), DynamicDataSourceType.Method)] // MSTEST0018 + error at runtime
    public void M(int i) {}
}
```

## Proposed Feature

Do not require the `typeof` when we know the member is part of the class (through inheritance).

Youssef1313:
Fixes #4347


src/Adapter/MSTest.TestAdapter/Execution/TypeCache.cs
```diff
@@ -810,7 +810,8 @@ private MethodInfo GetMethodInfoForTestMethod(TestMethod testMethod, TestClassIn
         else if (methodBase != null)
         {
             Type[] parameters = methodBase.GetParameters().Select(i => i.ParameterType).ToArray();
-            testMethodInfo = PlatformServiceProvider.Instance.ReflectionOperations.GetRuntimeMethod(methodBase.DeclaringType!, methodBase.Name, parameters);
+            // TODO: Should we pass true for includeNonPublic?
+            testMethodInfo = PlatformServiceProvider.Instance.ReflectionOperations.GetRuntimeMethod(methodBase.DeclaringType!, methodBase.Name, parameters, includeNonPublic: false);
```
Evangelink, line 814: I'd say no except if after the refactoring we no longer support internal test methods (I hope we have a test for that).

Please remove the comment or fix the code.

Evangelink (approved):
One TODO to drop or fix but other LGTM

src/Adapter/MSTest.TestAdapter/Execution/TypeCache.cs
Youssef1313: `includeNonPublic: false` preserves the existing behavior. But I'm suspecting that to be a bug when `DiscoverInternals` is used, which is why I left it as "TODO". Let's discuss/investigate more on Monday.

Evangelink: Moving on by merging, I am creating a follow-up task #4508

Evangelink (approved):

Issue #4478, opened 2024-12-30
Warning for Assert.AreNotSame calls with value types


Youssef1313:
## Summary

Similar to #4352, we should also warn when Assert.AreNotSame is called with at least one argument being a value type.
Such assert will never fail and is likely not the intent.

## Background and Motivation

Searching quickly with grep.app, I found a violation example here:

https://github.com/jmpoep/vmprotect-3.5.1/blob/d8fcb7c0ffd4fb45a8cfbd770c8b117d7dbe52b5/runtime/VMProtect.Runtime/Tests/UnitTestProject/UnitTest1.cs#L34

The linked assert will **not** fail even if `hwid.ToString().Length` is zero, which is clearly not the intent of the assertion.

## Proposed Feature

An analyzer to warn when Assert.AreNotSame is called with value types.

## Alternative Designs

AB#2331954

Youssef1313:
FYI @Evangelink

PR #4493, opened 2025-01-02
Add analyzer for value type usage with Assert.AreSame

### LINKED ISSUE #4352: Improve handling of incorrect calls to Assert.AreSame
It is currently possible to call Assert.AreSame with value ~reference~ types. The assertion will fail when the test is run, but it would be better if
1) this was not possible at all (by constraining the generic type parameter), or
2) an analyzer could warn immediately about calls that will fail

Not a big problem since the test will fail, but would be nice since it would shorten the time before the mistake was discovered.

Youssef1313:
Fixes #4352

### DIFF
--- a/src/Analyzers/MSTest.Analyzers/AvoidAssertAreSameWithValueTypesAnalyzer.cs
+++ b/src/Analyzers/MSTest.Analyzers/AvoidAssertAreSameWithValueTypesAnalyzer.cs
@@ -22,12 +22,13 @@ public sealed class AvoidAssertAreSameWithValueTypesAnalyzer : DiagnosticAnalyze
 {
     private static readonly LocalizableResourceString Title = new(nameof(Resources.AvoidAssertAreSameWithValueTypesTitle), Resources.ResourceManager, typeof(Resources));
     private static readonly LocalizableResourceString MessageFormat = new(nameof(Resources.AvoidAssertAreSameWithValueTypesMessageFormat), Resources.ResourceManager, typeof(Resources));
+    private static readonly LocalizableResourceString Description = new(nameof(Resources.AvoidAssertAreSameWithValueTypesDescription), Resources.ResourceManager, typeof(Resources));
 
     internal static readonly DiagnosticDescriptor Rule = DiagnosticDescriptorHelper.Create(
         DiagnosticIds.AvoidAssertAreSameWithValueTypesRuleId,
         Title,
         MessageFormat,
-        null,
+        Description,
         Category.Usage,
         DiagnosticSeverity.Warning,
         isEnabledByDefault: true);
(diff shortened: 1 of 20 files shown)


src/Analyzers/MSTest.Analyzers/AvoidAssertAreSameWithValueTypesAnalyzer.cs
```diff
@@ -0,0 +1,77 @@
...
+    private static readonly LocalizableResourceString Title = new(nameof(Resources.AvoidAssertAreSameWithValueTypesTitle), Resources.ResourceManager, typeof(Resources));
+    private static readonly LocalizableResourceString MessageFormat = new(nameof(Resources.AvoidAssertAreSameWithValueTypesMessageFormat), Resources.ResourceManager, typeof(Resources));
+
+    internal static readonly DiagnosticDescriptor Rule = DiagnosticDescriptorHelper.Create(
+        DiagnosticIds.AvoidAssertAreSameWithValueTypesRuleId,
+        Title,
+        MessageFormat,
+        null,
+        Category.Usage,
+        DiagnosticSeverity.Info,
```
Evangelink: I would bump this to warning as it's impossible for the actual call to be valid.

src/Analyzers/MSTest.Analyzers/AvoidAssertAreSameWithValueTypesAnalyzer.cs
```diff
@@ -0,0 +1,77 @@
...
+                context.RegisterOperationAction(context => AnalyzeOperation(context, assertSymbol), OperationKind.Invocation);
+            }
+        });
+    }
+
+    private static void AnalyzeOperation(OperationAnalysisContext context, INamedTypeSymbol assertSymbol)
+    {
+        var operation = (IInvocationOperation)context.Operation;
+        IMethodSymbol targetMethod = operation.TargetMethod;
+        if (targetMethod.Name != "AreSame" ||
```
Evangelink, line 59: Shall we consider `AreNotSame`? It feels like it would always be true so the assert is pointless (maybe this should be handled in the other rule we have about always true/false assertions).

Evangelink (approved):
LGTM, I would still add a rule description explaining why this is never working.

src/Analyzers/MSTest.Analyzers/AvoidAssertAreSameWithValueTypesAnalyzer.cs
Youssef1313: I have a separate issue for it: https://github.com/microsoft/testfx/issues/4478.

The always failing asserts (e.g, `Assert.IsTrue(false, ...)`) isn't warning by default as it's a more common pattern. So I'd assume we will want to make AreSame/AreNotSame separate as they are "more" wrong. I guess AreSame/AreNotSame can just go both in the exact same analyzer in this PR? What do you think?

Evangelink: Lgtm

Evangelink (approved):

PR #4495, opened 2025-01-02
Fix match all filter not working when property expression is present

### LINKED ISSUE #3936: Treenode Filtering not working with All Nodes Wildcard `/**`
I'm not sure if I'm doing something wrong, but according to this, it should work: https://github.com/microsoft/testfx/blob/main/docs/mstest-runner-graphqueryfiltering/graph-query-filtering.md

Here's my CLI input and output:

With all nodes wildcard:
```
PS C:\git\TUnit\TUnit.TestProject> dotnet run --treenode-filter /**[MyKey=MyDynamicallyAddedValue] --no-restore -f net8.0

████████╗██╗   ██╗███╗   ██╗██╗████████╗
╚══██╔══╝██║   ██║████╗  ██║██║╚══██╔══╝
   ██║   ██║   ██║██╔██╗ ██║██║   ██║
   ██║   ██║   ██║██║╚██╗██║██║   ██║
   ██║   ╚██████╔╝██║ ╚████║██║   ██║
   ╚═╝    ╚═════╝ ╚═╝  ╚═══╝╚═╝   ╚═╝

   TUnit v1.0.0.0 | 64-bit | Microsoft Windows 10.0.22631 | win-x64 | .NET 8.0.8 | Microsoft Testing Platform v1.4.1


Test run summary: Zero tests ran - bin\Debug\net8.0\TUnit.TestProject.dll (net8.0|x64)
  total: 0
  failed: 0
  succeeded: 0
  skipped: 0
  duration: 223ms

PS C:\git\TUnit\TUnit.TestProject>
```

Specifying class name:
```
PS C:\git\TUnit\TUnit.TestProject> dotnet run --treenode-filter /*/*/FilterByDynamicAddedPropertyTests/*[MyKey=MyDynamicallyAddedValue] --no-restore -f net8.0

████████╗██╗   ██╗███╗   ██╗██╗████████╗
╚══██╔══╝██║   ██║████╗  ██║██║╚══██╔══╝
   ██║   ██║   ██║██╔██╗ ██║██║   ██║
   ██║   ██║   ██║██║╚██╗██║██║   ██║
   ██║   ╚██████╔╝██║ ╚████║██║   ██║
   ╚═╝    ╚═════╝ ╚═╝  ╚═══╝╚═╝   ╚═╝

   TUnit v1.0.0.0 | 64-bit | Microsoft Windows 10.0.22631 | win-x64 | .NET 8.0.8 | Microsoft Testing Platform v1.4.1


Test run summary: Passed! - bin\Debug\net8.0\TUnit.TestProject.dll (net8.0|x64)
  total: 31
  failed: 0
  succeeded: 31
  skipped: 0
  duration: 285ms

PS C:\git\TUnit\TUnit.TestProject>
```

As you can see, the filter doesn't match when using the double asterix wildcard?

Filter match returns false:

Image

The path I build/pass to the filter service is:
`/TUnit.TestProject/TUnit.TestProject/FilterByDynamicAddedPropertyTests/Test1`

And the property bag contains:
`KeyValuePairStringProperty { Key = MyKey, Value = MyDynamicallyAddedValue }`

Youssef1313:
Fixes #3936

### DIFF
--- a/src/Platform/Microsoft.Testing.Platform/Requests/TreeNodeFilter/TreeNodeFilter.cs
+++ b/src/Platform/Microsoft.Testing.Platform/Requests/TreeNodeFilter/TreeNodeFilter.cs
@@ -441,7 +441,13 @@ public sealed class TreeNodeFilter : ITestExecutionFilter
             if (currentFragmentIndex >= _filters.Count)
             {
                 // Note: The regex for ** is .*.*, so we match against such a value expression.
-                return currentFragmentIndex > 0 && _filters.Last() is ValueExpression { Value: ".*.*" };
+                FilterExpression lastFilter = _filters.Last();
+                if (lastFilter is ValueAndPropertyExpression valueAndPropertyExpression)
+                {
+                    lastFilter = valueAndPropertyExpression.Value;
+                }
+
+                return currentFragmentIndex > 0 && lastFilter is ValueExpression { Value: ".*.*" };
             }
 
             if (!MatchFilterPattern(
--- a/test/UnitTests/Microsoft.Testing.Platform.UnitTests/Requests/TreeNodeFilterTests.cs
+++ b/test/UnitTests/Microsoft.Testing.Platform.UnitTests/Requests/TreeNodeFilterTests.cs
@@ -167,4 +167,23 @@ public sealed class TreeNodeFilterTests
             Assert.IsFalse(filterInstance.MatchesFilter(nodePath, nodeProperties));
         }
     }
+
+    [TestMethod]
+    public void MatchAllFilterWithPropertyExpression()
+    {
+        TreeNodeFilter filter = new("/**[A=B]");
+        Assert.IsTrue(filter.MatchesFilter("/A/B/C/D", new PropertyBag(new KeyValuePairStringProperty("A", "B"))));
+        Assert.IsFalse(filter.MatchesFilter("/A/B/C/D", new PropertyBag(new KeyValuePairStringProperty("A", "C"))));
+    }
+
+    [TestMethod]
+    public void MatchAllFilterSubpathWithPropertyExpression()
+    {
+        TreeNodeFilter filter = new("/A/**[A=B]");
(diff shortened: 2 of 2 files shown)


src/Platform/Microsoft.Testing.Platform/Requests/TreeNodeFilter/TreeNodeFilter.cs
```diff
@@ -441,7 +441,13 @@ public bool MatchesFilter(string testNodeFullPath, PropertyBag filterablePropert
...
             {
                 // Note: The regex for ** is .*.*, so we match against such a value expression.
-                return currentFragmentIndex > 0 && _filters.Last() is ValueExpression { Value: ".*.*" };
+                FilterExpression lastFilter = _filters.Last();
+                if (lastFilter is ValueAndPropertyExpression valueAndPropertyExpression)
+                {
+                    lastFilter = valueAndPropertyExpression.Value;
+                }
+
+                return currentFragmentIndex > 0 && lastFilter is ValueExpression { Value: ".*.*" };
```
Evangelink, line 450: I would need to make some tests and read the rest of the code but this is usually not recommended regex as it's a double eat all. It was there before so I won't block but it'd be good to investigate this part. Would you mind creating a ticket and assigning it to me?

Evangelink (approved):

src/Platform/Microsoft.Testing.Platform/Requests/TreeNodeFilter/TreeNodeFilter.cs
Youssef1313: @Evangelink For context, I think the way the code is done like this is to differentiate `*` and `**`. `TokenizeFilter` will just substitute `*` with `.*`, then we are able to special case `.*.*` and are sure it was `**`.

I'll create an issue.

Youssef1313: #4496

PR #4497, opened 2025-01-02
Simplify unnecessary extension methods in analyzers

Youssef1313:


src/Analyzers/MSTest.Analyzers/ClassCleanupShouldBeValidAnalyzer.cs
```diff
@@ -63,7 +63,7 @@ private static void AnalyzeSymbol(
     {
         var methodSymbol = (IMethodSymbol)context.Symbol;
         bool isInheritanceModeSet = methodSymbol.IsInheritanceModeSet(inheritanceBehaviorSymbol, classCleanupAttributeSymbol);
-        if (methodSymbol.IsClassInitializeMethod(classCleanupAttributeSymbol)
```
Youssef1313, line 66: Note: This was calling `IsClassInitializeMethod` but passing `classCleanupAttributeSymbol`.

The end result is that it's checking for `[ClassCleanup]` which is the intended behavior. But this example kinda shows why those extension methods are unnecessary. They all basically have the same implementation, just different parameter name.

Evangelink (approved):

PR #4503, opened 2025-01-03
Small ClassInitialize refactoring

Youssef1313:


src/Adapter/MSTest.TestAdapter/Execution/TestClassInfo.cs
```diff
@@ -343,18 +345,29 @@ public void RunClassInitialize(TestContext testContext)
...
         {
-            // For optimization purposes, we duplicate some of the logic of RunClassInitialize here so we don't need to start
-            // a thread for nothing.
-            if ((ClassInitializeMethod is null && BaseClassInitMethods.Count == 0)
-                || IsClassInitializeExecuted)
+            // For optimization purposes, avoid starting a thread when we know it will do nothing.
+            // This is the case when:
+            // 1. We have already executed class initialize (this is handled at the beginning for STA and non-STA)
+            // 2. There is no class initialize method to execute. In this case we mark IsClassInitializeExecuted and return a passed result (this condition here).
+            if (ClassInitializeMethod is null && BaseClassInitMethods.Count == 0)
```
Evangelink: Looking from the phone so I have a less good view but can't we also move this condition early in the method? It seems independent to using sta thread or not.

Youssef1313: Indeed. That makes sense.

Evangelink (approved):

PR #4523, opened 2025-01-06
Report diagnostic when AreNotSame is passed value type

### LINKED ISSUE #4478: Warning for Assert.AreNotSame calls with value types

Youssef1313:
Fixes #4478


Evangelink (approved):
Neat, thanks @Youssef1313

PR #4537, opened 2025-01-07
Fix "Could not load file or assembly 'Microsoft.Testing.Platform'" exception

### LINKED ISSUE #4426: MSTest.Sdk 3.7.0 issues when targeting .NET Framework 4.8
When using MSTest.Sdk 3.7.0 in the test project targeting .NET Framework 4.8 I found the following issues:

- No DLL is produced, only EXE.
- Executing tests from Visual Studio fails with:
 `An exception occurred while test discoverer 'MSTestDiscoverer' was loading tests. Exception: Could not load file or assembly 'Microsoft.Testing.Platform, Version=1.5.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a' or one of its dependencies. The system cannot find the file specified.`

When projects target .NET 8 the aforementioned issues are not present. See the [sample project](https://github.com/user-attachments/files/18223588/MSTestSdk.zip).

Evangelink:
During the investigation, I realized that VSTest custom assembly resolver, enabled only when AppDomains are enabled and only for netfx, includes only VS folders and MSTest.TestAdapter. On 3.7, we started to depend upon some types from MTP leading to the dll being required but not loaded in this mode (the deps is part of a different NuGet). 

As part of the analysis, I have also noticed that we are required to keep the framework under the adapter folder for the exact same reason.

I have restored the framework, only for netfx, that will be resolved only when under the customer resolver for netfx in VSTest. For MTP, I have decided not to copy the dll but instead to have wrapper types so we don't "pollute" the engine with types specific to a given platform.

Fixes #4426

### DIFF
--- a/src/Adapter/MSTest.TestAdapter/TestingPlatformAdapter/BridgedConfiguration.cs
+++ b/src/Adapter/MSTest.TestAdapter/TestingPlatformAdapter/BridgedConfiguration.cs
@@ -1,0 +1,23 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+#if !WINDOWS_UWP
+using System.Diagnostics.CodeAnalysis;
+
+using Microsoft.Testing.Platform.Configurations;
+
+using PlatformServicesConfiguration = Microsoft.VisualStudio.TestPlatform.MSTestAdapter.PlatformServices.Interface.IConfiguration;
+
+namespace Microsoft.VisualStudio.TestTools.UnitTesting;
+
+[SuppressMessage("ApiDesign", "RS0030:Do not use banned APIs", Justification = "We can use MTP from this folder")]
+internal sealed class BridgedConfiguration : PlatformServicesConfiguration
+{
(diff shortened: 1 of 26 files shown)


src/Adapter/MSTest.TestAdapter/TestingPlatformAdapter/BridgedConfiguration.cs
```diff
@@ -0,0 +1,23 @@
...
+#if !WINDOWS_UWP
+using System.Diagnostics.CodeAnalysis;
+
+using Microsoft.Testing.Platform.Configurations;
+
+using PlatformServicesConfiguration = Microsoft.VisualStudio.TestPlatform.MSTestAdapter.PlatformServices.Interface.IConfiguration;
+
+namespace Microsoft.VisualStudio.TestTools.UnitTesting;
+
+[SuppressMessage("ApiDesign", "RS0030:Do not use banned APIs", Justification = "We can use MTP from this folder")]
```
Evangelink, line 13: This is not ideal as it suppresses too much. I would prefer to build a custom analyzer but wanted to ship a quick fix and this is "good enough" for now.

Youssef1313 (approved):

PR #4541, opened 2025-01-07
Add messageBuilder overloads for Assert.Throws[Exactly][Async]

### LINKED ISSUE #1655: Pass actual exception to Assert.ThrowsException(Async) assertion error message builder
It would be helpful to get an actual exception when creating an assertion error message in the `Assert.ThrowsException(Async)<T>` methods.
```
var ex = await Assert.ThrowsExceptionAsync<ExpectedException>(
    () => DoAsync(),
    actualException =>
    {
        return actualException switch
        {
            RpcException rpc => $"{rpc.Message}, Trailers: {string.Join(", ", rpc.Trailers)}",
            _ => actualException.Message
        };
    });
```

AB#2332772

Youssef1313:
Fixes #1655


Evangelink:
@Youssef1313 another flakiness on `WhenSpecificDataSourceIsIgnoredViaIgnoreMessageProperty`. I'll force merge but let's fix this test as discussed offline.

Evangelink (approved):

PR #4546, opened 2025-01-07
missing temp dir cleanup in TestAssetFixtureBase

SimonCropp:


test/Utilities/Microsoft.Testing.TestInfrastructure/TestAssetFixtureBase.cs
```diff
@@ -63,6 +63,8 @@ protected virtual void Dispose(bool disposing)
 
             _disposedValue = true;
         }
+
+        _nugetGlobalPackagesDirectory.Dispose();
```
Youssef1313, line 67: We create a singleton of `NuGetGlobalPackagesFolder` and we dispose in assembly cleanup. See `AcceptanceFixture.cs`

PR #4549, opened 2025-01-08
remove messageFormatOnException in CreateNavigationSession

SimonCropp:
Draft until we confirm it should not be used?


Evangelink (approved):
Seems good to me.

PR #4552, opened 2025-01-08
remove redundant finalArguments.Clear

SimonCropp:
it is out of scope, so clear should be redundant?

### DIFF
--- a/src/Platform/Microsoft.Testing.Extensions.Retry/RetryOrchestrator.cs
+++ b/src/Platform/Microsoft.Testing.Extensions.Retry/RetryOrchestrator.cs
@@ -275,7 +275,6 @@ internal sealed class RetryOrchestrator : ITestHostOrchestrator, IOutputDeviceDa
                     }
                 }
 
-                finalArguments.Clear();
                 lastListOfFailedId = retryFailedTestsPipeServer.FailedUID?.ToArray();
                 continue;
             }


src/Platform/Microsoft.Testing.Extensions.Retry/RetryOrchestrator.cs
```diff
@@ -275,7 +275,6 @@ public async Task<int> OrchestrateTestHostExecutionAsync()
                     }
                 }
 
-                finalArguments.Clear();
```
Youssef1313, line 278: Isn't the continue below going back to the next iteration of `while (attemptCount < userMaxRetryCount + 1)` which will start using finalArguments again?

Evangelink: As far as I can see we don't use this variable after the foreach on lines 176-183.

Evangelink (approved):
LGTM, thanks @SimonCropp

src/Platform/Microsoft.Testing.Extensions.Retry/RetryOrchestrator.cs
Evangelink: Sorry I missed the outer loop, you are right this is not correct. I need to refactor this class, there is too much indentation and things to keep in mind.

Evangelink:
As pointed out by @Youssef1313 this is changing the behavior. The class does need some refactoring to reduce cognitive complexity.

src/Platform/Microsoft.Testing.Extensions.Retry/RetryOrchestrator.cs
SimonCropp: yeah i made the same mistake. and agreed this class is difficult to reason about

PR #4555, opened 2025-01-08
Fix potential race condition in GetResultOrRunClassInitialize

Youssef1313:


Evangelink:
Looks like there is another hang.

Youssef1313:
@Evangelink This is green now

Evangelink (approved):
Does it need to be backported?

PR #4564, opened 2025-01-08
make some methods static

SimonCropp:


Youssef1313 (approved):
LGTM, but this is curious. Why the one in MTP product code wasn't flagged by CA1822? The rest are test-only and we disable the rule on tests so that's expected

https://github.com/microsoft/testfx/blob/9682c1d6ef850fac69b250a48bb9216b556a8741/src/.editorconfig#L13

@Evangelink Do you see any obvious reason? If not, probably we should track it in an issue for future investigation.

Evangelink (approved):
LGTM

Evangelink:
> @Evangelink Do you see any obvious reason? If not, probably we should track it in an issue for future investigation.

I have seen a couple of analyzers that are not well raised in our repo but never was able to spend much time on it.

> The rest are test-only and we disable the rule on tests so that's expected

We may want to re-enable for the test infra projects.

PR #4567, opened 2025-01-09
Set MTP editorconfig as non-root explicitly

Youssef1313:
Maybe this is already the default, I don't know. But it's better to be explicit IMO.


Evangelink (approved):
Fine by me

PR #4575, opened 2025-01-09
simplify some dictionary usage

SimonCropp:


src/Adapter/MSTest.TestAdapter/Execution/TestClassInfo.cs
```diff
@@ -483,11 +483,7 @@ UnitTestResult DoRun()
             return null;
         }
 
-        TimeoutInfo? timeout = null;
-        if (ClassInitializeMethodTimeoutMilliseconds.TryGetValue(methodInfo, out TimeoutInfo localTimeout))
-        {
-            timeout = localTimeout;
-        }
+        TimeoutInfo? timeout = ClassInitializeMethodTimeoutMilliseconds.GetValueOrDefault(methodInfo);
```
Youssef1313: ⚠️ The dictionary is `Dictionary<MethodInfo, TimeoutInfo>`. As `TimeoutInfo` is a struct, `GetValueOrDefault` is going to return `default(TimeoutInfo)` when not found, whereas previously we would have used null. So this feels like a change in behavior

src/Adapter/MSTest.TestAdapter/SourceGeneration/SourceGeneratedReflectionOperations.cs
```diff
@@ -84,7 +84,7 @@ public MethodInfo[] GetRuntimeMethods(Type type)
         Dictionary<string, PropertyInfo> type = ReflectionDataProvider.TypePropertiesByName[classType];
 
         // We as asking for TestContext here, it may not be there.
-        PropertyInfo? property = type.TryGetValue(propertyName, out PropertyInfo? propertyInfo) ? propertyInfo : null;
+        PropertyInfo? property = type.GetValueOrDefault(propertyName);
```
Youssef1313, line 87:
```suggestion
        _ = type.TryGetValue(propertyName, PropertyInfo? property);
```

SimonCropp: why not GetValueOrDefault ?

Youssef1313: They are the same, nothing really wrong about `GetValueOrDefault`.

`GetValueOrDefault` will end up calling TryGetValue at the end of the day, so I preferred to call TryGetValue directly unless `GetValueOrDefault` is saving annoying conditional syntax (like the case in `GetMetadata` below in the PR)

It's mostly personal opinion/style though.

SimonCropp: give my preference for this style does not match the projects's, i will close

PR #4578, opened 2025-01-09
inline string literals

SimonCropp:


Evangelink (approved):
Oh wow I wasn't aware we did that!

SimonCropp:
@Evangelink IMO it should be a compile error. can u think of a scenario where is is desired?

PR #4586, opened 2025-01-09
Introduce `RetryAttribute` for test methods

### LINKED ISSUE #3161: Develop Retry Attribute for TestMethod
## Summary
I'm curious to know if there's any built-in mechanism to retry tests. Case in point, I have a unit test code something like this:

```
[TestMethod]
public async Task Get_Banner_ReturnOK()
{
    //tests
}
```

Im just looking for something I Can retry this TestMethod Twice or more, something like this:

```
[TestMethod]
[Retry(2)]
public async Task Get_Banner_ReturnOK()
{
    //tests
}
```

AB#2332770

Youssef1313:
Fixes #3161


src/TestFramework/TestFramework/Attributes/TestMethod/RetryAttribute.cs
```diff
@@ -0,0 +1,22 @@
...
+    /// <summary>
+    /// Initializes a new instance of the <see cref="RetryAttribute"/> class with the given number of max retries.
+    /// </summary>
+    public RetryAttribute(int maxRetries)
+        => MaxRetries = maxRetries;
+
+    /// <summary>
+    /// Gets the number of retries that the test should make in case of failures.
+    /// </summary>
+    public int MaxRetries { get; }
```
Evangelink: What about `MaxRetryAttempts`?

src/TestFramework/TestFramework/Attributes/TestMethod/RetryAttribute.cs
```diff
@@ -0,0 +1,22 @@
...
+/// <summary>
+/// This attribute is used to set a retry count on a test method in case of failure.
+/// </summary>
+[AttributeUsage(AttributeTargets.Method, Inherited = false, AllowMultiple = false)]
+public sealed class RetryAttribute : Attribute
+{
+    /// <summary>
+    /// Initializes a new instance of the <see cref="RetryAttribute"/> class with the given number of max retries.
+    /// </summary>
+    public RetryAttribute(int maxRetries)
```
Evangelink: I don't want to build something too complex but I think it would be good to have:

- Delay: wait time between 2 retries
- optional: some backoff strategy, I think constant and exponential are the most used.

src/TestFramework/TestFramework/Attributes/TestMethod/RetryAttribute.cs
```diff
@@ -0,0 +1,22 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+namespace Microsoft.VisualStudio.TestTools.UnitTesting;
+
+/// <summary>
+/// This attribute is used to set a retry count on a test method in case of failure.
+/// </summary>
+[AttributeUsage(AttributeTargets.Method, Inherited = false, AllowMultiple = false)]
+public sealed class RetryAttribute : Attribute
```
Evangelink: It'd be good to have this attribute non-sealed so we allow extensibility, which means the logic should also be partially here so it can be overrided. I mainly want that so people (or we?) could ship a retry based off Polly.

src/TestFramework/TestFramework/Attributes/TestMethod/RetryAttribute.cs
```diff
@@ -0,0 +1,22 @@
...
+{
+    /// <summary>
+    /// Initializes a new instance of the <see cref="RetryAttribute"/> class with the given number of max retries.
+    /// </summary>
+    public RetryAttribute(int maxRetries)
+        => MaxRetries = maxRetries;
```
Evangelink: Shall we validate value here?

src/TestFramework/TestFramework/Attributes/TestMethod/RetryAttribute.cs
Evangelink: Let's be explicit here this is in addition to the original call so a value of 1 means call + 1 retry, so 2 attempts in total.

src/Adapter/MSTest.TestAdapter/Execution/TestMethodInfo.cs
```diff
@@ -260,6 +263,20 @@ public virtual TestResult Invoke(object?[]? arguments)
         return expectedExceptions.FirstOrDefault();
     }
 
+    /// <summary>
+    /// Gets the number of retries this test method should make in case of failure.
+    /// </summary>
+    /// <returns>
+    /// The number of retries, which is always greater than or equal to 1.
+    /// If RetryAttribute is not present, returns 1.
```
thomhurst, line 266: I guess it's how you interpret language, but in TUnit, lack of a retry attribute means a retry count of 0;

Because if I do `[Retry(1)]` I'd expect it to try again after a failure, once.

If I'm reading this PR correctly, `[Retry(1)]` currently means, run the test normally as you would and never retry? So that's retrying 0 times.

Does that make sense? The number doesn't work with the wording. This usage would work better with `[Attempt(1)]` or `[Try(1)]` or something.

Youssef1313: @thomhurst Yup indeed. I'll be working on a fix so that `[Retry(n)]` means it can run `1 + n` times at max (normal first attempt + n retries).

src/TestFramework/TestFramework/Attributes/TestMethod/RetryAttribute.cs
nohwnd: I think we have a precedens for that with data sources that don't use non-sealed class, but rather than that an interface (ITestDataSource) on the attribute, so any implementer does not have to provide the same ctor parameters we provided.

src/Adapter/MSTest.TestAdapter/Execution/TestMethodInfo.cs
nohwnd: Perfect, NUnit imho does it the way when Retry(1) means normal run, and I only see that being source of confusion.

src/TestFramework/TestFramework/Attributes/TestMethod/RetryAttribute.cs
nohwnd: Not strong opinion, but the comment already says "in case of failure", to me it is clear.

nohwnd:
FYI From previous talks with Playwright where they have good retries, that imho were attempted to be duplicated in mstest.playwright, Max also mentioned that they wanted: 

- max retries to be configurable via config
- see what max retries value is in assembly initialize (might be specific to them because they are in position of test framework in vstest, so that story might not be relevant here, where they have way more power via extensions easily)
- have good reporting of the retries, to be able to show retries that were not passing

Youssef1313:
> have good reporting of the retries, to be able to show retries that were not passing

We are going to have a design that allows that, but that is unlikely to be implemented in this PR though, unfortunately. We need to know better how to communicate that information to tools that will make use of it.

src/Adapter/MSTest.TestAdapter/Execution/UnitTestRunner.cs
```diff
@@ -122,13 +123,19 @@ internal FixtureTestResult GetFixtureTestResult(TestMethod testMethod, string fi
         static UnitTestOutcome GetOutcome(Exception? exception) => exception == null ? UnitTestOutcome.Passed : UnitTestOutcome.Failed;
     }
 
+    // Task cannot cross app domains.
+    // For now, TestExecutionManager will call this sync method which is hacky.
+    // If we removed AppDomains in v4, we should use the async method and remove this one.
+    internal UnitTestResult[] RunSingleTest(TestMethod testMethod, IDictionary<string, object?> testContextProperties, IMessageLogger messageLogger)
+        => RunSingleTestAsync(testMethod, testContextProperties, messageLogger).GetAwaiter().GetResult();
```
Youssef1313, line 130: Could we maybe instead have our custom task-like type that's serializable and can cross app domains? Is it worth doing?

src/TestFramework/TestFramework/Attributes/TestMethod/RetryAttribute.cs
```diff
@@ -0,0 +1,115 @@
...
+    /// </summary>
+    public int MaxRetryAttempts { get; }
+
+    /// <summary>
+    /// Gets or sets the delay, in milliseconds, between retries.
+    /// This delay is also applied after the first run and before the first retry attempt.
+    /// </summary>
+    public int MillisecondsDelayBetweenRetries { get; set; }
+
+    public DelayBackoffType BackoffType
```
Youssef1313, line 39: @Evangelink A user may want to create a custom `RetryAttribute`, but not respect these properties and have their own stuff. In that case, the fact that the properties are there but not having effect feels bad.

I'm thinking of separating this class into `RetryBaseAttribute` and `RetryAttribute`.

- `RetryBaseAttribute` will be abstract with only an abstract method `ExecuteAsync`
- `RetryAttribute` will remain as it is currently.

The `UnitTestRunner` only cares about `ExecuteAsync` anyways.

What do you think? Does that make sense to you?

src/Adapter/MSTest.TestAdapter/Execution/TestMethodInfo.cs
```diff
@@ -260,6 +263,16 @@ public virtual TestResult Invoke(object?[]? arguments)
...
+    /// </returns>
+    private RetryAttribute? GetRetryAttribute()
+        => ReflectHelper.Instance.GetFirstDerivedAttributeOrDefault<RetryAttribute>(TestMethod, inherit: true);
```
Youssef1313: TODO: We should throw if we found multiple.

src/Adapter/MSTest.TestAdapter/Execution/UnitTestRunner.cs
Youssef1313: Will keep it simplified for now, and we will adjust in future if needed.

src/Adapter/MSTest.TestAdapter/Execution/TestMethodInfo.cs
```diff
@@ -249,17 +252,44 @@ public virtual TestResult Invoke(object?[]? arguments)
...
+    /// Gets the number of retries this test method should make in case of failure.
+    /// </summary>
+    /// <returns>
+    /// The number of retries, which is always greater than or equal to 1.
+    /// If RetryAttribute is not present, returns 1.
+    /// </returns>
+    private RetryBaseAttribute? GetRetryAttribute()
+    {
+        IEnumerable<RetryBaseAttribute> attributes = ReflectHelper.Instance.GetDerivedAttributes<RetryBaseAttribute>(TestMethod, inherit: true);
+        if (attributes.Count() > 1)
```
Evangelink: Not a problem as we don't expect to have many attributes but I would usually recommend to avoid `Count` when you just want to check more than 1.

test/IntegrationTests/MSTest.Acceptance.IntegrationTests/RetryTests.cs
```diff
@@ -0,0 +1,146 @@
...
+{
+    [TestMethod]
+    public async Task BasicRetryScenarioTest()
+    {
+        var testHost = TestHost.LocateFrom(AssetFixture.ProjectPath, TestAssetFixture.ProjectName, TargetFrameworks.NetCurrent);
+        TestHostResult testHostResult = await testHost.ExecuteAsync("--settings my.runsettings");
+
+        testHostResult.AssertExitCodeIs(ExitCodes.AtLeastOneTestFailed);
+        testHostResult.AssertOutputContains("""
+            TestMethod1 executed '1' times.
```
Evangelink: Nit: time and not times.

test/IntegrationTests/MSTest.Acceptance.IntegrationTests/RetryTests.cs
```diff
@@ -0,0 +1,146 @@
...
+            TestMethod2 executed '2' times.
+            TestMethod3 executed '3' times.
+            TestMethod4 executed '4' times.
+            TestMethod5 executed '4' times.
+            """);
+
+        testHostResult.AssertOutputContains("failed TestMethod5");
+        testHostResult.AssertOutputContains("Assert.Fail failed. Failing TestMethod4. Attempt: 4");
```
Evangelink: Nit: Attempts

test/IntegrationTests/MSTest.Acceptance.IntegrationTests/RetryTests.cs
```diff
@@ -0,0 +1,146 @@
...
+    }
+
+    [ClassCleanup]
+    public static void ClassCleanup()
+    {
+        Console.WriteLine($"TestMethod1 executed '{_count1}' times.");
+        Console.WriteLine($"TestMethod2 executed '{_count2}' times.");
+        Console.WriteLine($"TestMethod3 executed '{_count3}' times.");
+        Console.WriteLine($"TestMethod4 executed '{_count4}' times.");
+        Console.WriteLine($"TestMethod5 executed '{_count5}' times.");
```
Evangelink: I would remove the single quote, it's not helping much with readability.

Evangelink (dismissed):

src/Adapter/MSTest.TestAdapter/Resources/Resource.resx
```diff
@@ -267,8 +267,8 @@ Error: {1}</value>
     <value>The ExpectedException attribute defined on test method {0}.{1} threw an exception during construction.
 {2}</value>
   </data>
-  <data name="UTA_MultipleExpectedExceptionsOnTestMethod" xml:space="preserve">
-    <value>The test method {0}.{1} has multiple attributes derived from ExpectedExceptionBaseAttribute defined on it. Only one such attribute is allowed.</value>
+  <data name="UTA_MultipleAttributesOnTestMethod" xml:space="preserve">
+    <value>The test method {0}.{1} has multiple attributes derived from '{2}' defined on it. Only one such attribute is allowed.</value>
```
Evangelink: Nit: 
```suggestion
    <value>The test method '{0}.{1}' has multiple attributes derived from '{2}' defined on it. Only one such attribute is allowed.</value>
```

src/TestFramework/TestFramework/Attributes/TestMethod/RetryBaseAttribute.cs
```diff
@@ -0,0 +1,53 @@
...
+        {
+            UnitTestOutcome outcome = result.Outcome;
+            if (outcome is UnitTestOutcome.Failed or UnitTestOutcome.Timeout)
+            {
+                return false;
+            }
+        }
+
+        return true;
+    }
```
Evangelink: We need only one method accepting `IEnumerable<>`

src/TestFramework/TestFramework/Attributes/TestMethod/RetryContext.cs
```diff
@@ -0,0 +1,12 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+namespace Microsoft.VisualStudio.TestTools.UnitTesting;
+
+public sealed class RetryContext
```
Evangelink: Shall we use `struct` like roslyn? We probably have very few callback so it might not be needed.

src/TestFramework/TestFramework/Attributes/TestMethod/RetryAttribute.cs
```diff
@@ -0,0 +1,87 @@
...
+    /// Note that a first run of the method was already executed and failed before this method is called.
+    /// </summary>
+    /// <param name="retryContext">An object to encapsulate the state needed for retry execution.</param>
+    /// <returns>
+    /// Returns a <see cref="RetryResult"/> object that contains the results of all attempts. Only
+    /// the last added element is used to determine the test outcome.
+    /// The other results are currently not used, but may be used in the future for tooling to show the
+    /// state of the failed attempts.
+    /// </returns>
+    protected internal override async Task<RetryResult> ExecuteAsync(RetryContext retryContext)
```
Evangelink, line 64: Shall we use an interface instead to give us more freedom in future without breaking change? I don't know if that's support useful or not given roslyn mainly goes with `struct`.

src/TestFramework/TestFramework/Attributes/TestMethod/RetryBaseAttribute.cs
Youssef1313: Using `IEnumerable<TestResult>` will box the enumerator, but not a big deal. I'll switch.

src/TestFramework/TestFramework/Attributes/TestMethod/RetryContext.cs
Youssef1313: Fine for me to use struct.

Evangelink (approved):

src/Adapter/MSTest.TestAdapter/Execution/TestExecutionManager.cs
```diff
@@ -121,7 +124,9 @@ public void RunTests(IEnumerable<TestCase> tests, IRunContext? runContext, IFram
         CacheSessionParameters(runContext, frameworkHandle);
 
         // Execute the tests
-        ExecuteTests(tests, runContext, frameworkHandle, isDeploymentDone);
+        // This is a public API, so we can't change it to be async.
+        // Consider not using this API internally, and introduce an async version, and mark this as obsolete.
+        ExecuteTestsAsync(tests, runContext, frameworkHandle, isDeploymentDone).GetAwaiter().GetResult();
```
Youssef1313, line 129: I think I need to address that in a follow-up

PR #4591, opened 2025-01-09
Refactor TestMethodRunner.Execute so that it returns TestResult (from TestFramework) rather than UnitTestResult (from TestAdapter)

Youssef1313:

### DIFF
--- a/src/Adapter/MSTest.TestAdapter/Execution/UnitTestRunner.cs
+++ b/src/Adapter/MSTest.TestAdapter/Execution/UnitTestRunner.cs
@@ -127,8 +127,8 @@ internal sealed class UnitTestRunner : MarshalByRefObject
     /// </summary>
     /// <param name="testMethod"> The test Method. </param>
     /// <param name="testContextProperties"> The test context properties. </param>
-    /// <returns> The <see cref="UnitTestResult"/>. </returns>
-    internal UnitTestResult[] RunSingleTest(TestMethod testMethod, IDictionary<string, object?> testContextProperties, IMessageLogger messageLogger)
+    /// <returns> The <see cref="TestResult"/>. </returns>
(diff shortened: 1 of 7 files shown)


src/Adapter/MSTest.TestAdapter/Execution/UnitTestRunner.cs
```diff
@@ -180,7 +180,7 @@ internal UnitTestResult[] RunSingleTest(TestMethod testMethod, IDictionary<strin
                         // Run the test method
                         testContextForTestExecution.SetOutcome(testContextForClassInit.Context.CurrentTestOutcome);
                         var testMethodRunner = new TestMethodRunner(testMethodInfo, testMethod, testContextForTestExecution);
-                        result = testMethodRunner.Execute(classInitializeResult.StandardOut!, classInitializeResult.StandardError!, classInitializeResult.DebugTrace!, classInitializeResult.TestContextMessages!);
+                        result = testMethodRunner.Execute(classInitializeResult.StandardOut!, classInitializeResult.StandardError!, classInitializeResult.DebugTrace!, classInitializeResult.TestContextMessages!).ToUnitTestResults();
```
Youssef1313, line 183: @Evangelink My idea here is now that Execute returns TestResult, we can make the RetryAttribute work with TestResult as well, which is in TestFramework.

Basically, internally in Execute we were already dealing with TestResult but we converted it to UnitTestResult. I'm making it such that the conversion happens outside the Execute.

Does that make sense?

Evangelink (approved):
LGMT but do we need the change to `List`?

Youssef1313:
> LGMT but do we need the change to `List`?

Avoiding `List<TestResult>` -> `TestResult[]` -> `UnitTestResult[]` and instead going `List<TestResult>` -> `UnitTestResult[]` directly. We shouldn't be needing `UnitTestResult` at all and I'll investigate that separately.

PR #4620, opened 2025-01-12
Fix DynamicData.GetData regression preventing it to work without adapter

### LINKED ISSUE #4619: [Regression] DynamicDataAttribute.GetData no longer works in 3.7 when using TestFramework without TestAdapter
This regression is likely caused by https://github.com/microsoft/testfx/pull/3801

In this PR, pay attention to the change in `DynamicDataAttribute.cs`.

Before the changes, DynamicDataAttribute.GetData would always work and use reflection.

After the change, DynamicDataAttribute will use `DynamicDataProvider.Instance.GetData`, but `DynamicDataProvider.Instance` is set by the adapter. So, when the adapter isn't used, this fails because DynamicDataProvider.Instance is null:

https://github.com/microsoft/testfx/blob/f2de56070d8a6ce61649eea84a955aa6f24b9fc9/src/TestFramework/TestFramework/Attributes/DataSource/DynamicDataProvider.cs#L12-L14

Worse, the exception says:

> If you are seeing this error, you are using Test Framework without Test Adapter, and your adapter should set DynamicDataProvider.Instance

However, DynamicDataProvider is even internal.

Youssef1313:
Fixes #4619


src/TestFramework/TestFramework/Attributes/DataSource/DynamicDataOperations.cs
```diff
@@ -235,6 +234,7 @@ private static bool IsTupleOrValueTuple(Type type, out int tupleSize)
             return true;
         }
 
+#if !NET462 // TODO: Add dependency on System.ValueTuple for net462?
```
Youssef1313: @Evangelink Thoughts?

Evangelink: We have it on the adapter today, right? If so that's probably fine, I would maybe just be worried to introduce a deps in a bug fix release (3.7.1).

Evangelink (dismissed):
Nice work

PR #4625, opened 2025-01-13
Use FailFast instead of throwing in DebugEx.Assert under .NET Framework

Youssef1313:


Evangelink (approved):

Evangelink:
Restarted failed job, flakiness again :c

PR #4639, opened 2025-01-14
Fix Tuple/ValueTuple handling of TRest

### LINKED ISSUE #4624: Handle Tuples/ValueTuples that have size >= 8
https://github.com/microsoft/testfx/blob/c7496765495fe8027474ba0154105a295b813ab5/src/Adapter/MSTest.TestAdapter/DynamicDataOperations.cs#L245

The last generic parameter here is a TRest and is not handled correctly.

The same applies to `Tuple`s as well.

AB#2342947

Youssef1313:
Fixes #4624


Evangelink (commented):
Can we add some tests for these?

Evangelink (approved):

PR #4641, opened 2025-01-14
Reword unhandled TestMethodAttribute.Execute exception message

### LINKED ISSUE #4616: Are we not supposed to create subclasses of TestMethodAttribute anymore? ("If using extension of TestMethodAttribute then please contact vendor")
I created a subclass of the TestMethodAttribute, and while investigating an exception thrown by a test, the error message was prefixed with the following text:
> Exception thrown while executing test. If using extension of TestMethodAttribute then please contact vendor.

My need for this subclass, is to log each execution of a test to Application Insights, so that we can correlate the test result with the tested HTTP APIs (which also logs to Application Insights).

Basically:
```
[AttributeUsage(AttributeTargets.Method)]
public class LoggingTestMethodAttribute : Microsoft.VisualStudio.TestTools.UnitTesting.TestMethodAttribute
{
    public override TestResult[] Execute(ITestMethod testMethod)
    {
        using var requestLogger = new RequestLogger(GlobalTestSetup.Instance.TelemetryClient, testMethod);
        requestLogger.Start();

        var result = base.Execute(testMethod);

        requestLogger.Stop(result);
        return result;
    }
}
```

I couldn't find any other way to do it, except with a subclass of TestMethodAttribute.
If we shouldn't be subclassing TestMethodAttribute, what are we supposed to do instead? Is there already a better way of doing this, or should we just ignore the ominous text/warning?

I guess this has something to do with the breaking changes you are planning for V4 ?

Best regards,
Marcus Nilsson

Evangelink:
Fixes #4616

@marnilss is this message better?

### DIFF
--- a/test/UnitTests/MSTestAdapter.UnitTests/Execution/TestMethodRunnerTests.cs
+++ b/test/UnitTests/MSTestAdapter.UnitTests/Execution/TestMethodRunnerTests.cs
@@ -85,11 +85,12 @@ public class TestMethodRunnerTests : TestContainer
 
         UnitTestResult[] results = testMethodRunner.Execute(string.Empty, string.Empty, string.Empty, string.Empty).ToUnitTestResults();
         Verify(results[0].Outcome == AdapterTestOutcome.Failed);
-        Verify(results[0].ErrorMessage.Contains(
+        Verify(results[0].ErrorMessage.StartsWith(
             """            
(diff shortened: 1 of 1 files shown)


Youssef1313 (approved):

marnilss:
Looks good 👍

Youssef1313 (approved):

PR #4642, opened 2025-01-14
Avoid keeping trx stream open in unit tests

Youssef1313:
TrxTests works by creating a MemoryStream, and mocking NewFileStream to return that instance. Then later, the tests tries to use that stream. For that reason, tests are passing a flag to prevent production code form disposing the stream.

This PR cleans up product code, but makes tests more verbose. Now, product code will **always** dispose, without any flag that's used by tests. However, the "fake" stream created by tests will create the `XDocument` that the test needs when Dispose is called. That way, we don't need the stream anymore and we use the XDocument.

@Evangelink How do you feel about this change? I generally don't like when there is a code path in product code that's test only and prefer more ugly tests but cleaner product code. That's a personal taste though.


Evangelink (approved):
LGTM

PR #4662, opened 2025-01-15
Drop unused source gen related code

### LINKED ISSUE #4338: `SourceGeneratedReflectionOperations.GetDeclaredMethod` should throw if multiple methods are found.
The existing behavior of reflection's GetDeclaredMethod is that if multiple methods are found, an exception is thrown.

One impact this will have is on `[DynamicData(nameof(MyMethod), DynamicDataSourceType.Method]`

If MyMethod has multiple overloads, we would throw an exception when using reflection, but the behavior is kinda undefined when using source generator. It depends on which overload is returned (current impl uses FirstOrDefault).

Evangelink:
Fixes #4338


Youssef1313 (approved):
LGTM. The new source gen doesn't seem to be moving forward for now. We can bring that back if it ever happened.

PR #4678, opened 2025-01-16
Fix tuple regression before it's shipped :)

Youssef1313:


Youssef1313:
Let's see if the backport is smart enough to force push the existing PR :)

PR #4702, opened 2025-01-17
Simplify reflection

Youssef1313:


Youssef1313:
> if you are okay with the slight change in behavior where we will now consider more derived attributes.

I think it's low risk, probably no change in behavior at all. I think all places where we were looking for the specific attribute type, the attribute was already sealed. I can't see any place where we look for non-sealed attributes using NonDerived.

PR #4712, opened 2025-01-18
Implement MSTestAnalysisMode

### LINKED ISSUE #3325: Add an easy way to increase analyzer severity for rules that are good default, but a breaking change
## Summary

MSTest v3 ships with a set of analyzers which are balanced to avoid braking changes. 

Add a way to increase the severity of such rules for new projects so they can start with stricter rule set. 

Get Warning for the case above, and use the recommended solution of making the class abstract when it is just a base class.

## Background and Motivation

As an example: MSTEST0030, which reports `info` on non-abstract test class that has `[TestMethod]` but is not decorated with `[TestClass]`. This leads to the TestMethod1 being skipped by the test framework.

```cs
public class UnitTest1
{
    [TestMethod]
    public void TestMethod1()
    {
    }
}
```

The rule cannot be set to `warning` by default, because this is also a common pattern to create a base class with tests that are inherited, but are not intended to run. A recommended solution for this case is to mark the class abstract.

In my test base I would like this rule, and other such rules to be warnings for any new project, without breaking my old projects that also use mstest v3.

## Proposed Feature

Add build property or other knob to enable more strict analyzer rules for new projects, and make this the default in dotnet new templates, so new projects start with stricter rules, because they are more likely to embrace our recommended fixes and workarounds.

Possibly take advantage of analysis level https://learn.microsoft.com/dotnet/fundamentals/code-analysis/overview?tabs=net-8#enable-additional-rules

Youssef1313:
Fixes #3325

### DIFF
--- a/src/Analyzers/MSTest.GlobalConfigsGenerator/AnalyzerSeverityDecider.cs
+++ b/src/Analyzers/MSTest.GlobalConfigsGenerator/AnalyzerSeverityDecider.cs
@@ -48,7 +48,7 @@ internal static class AnalyzerSeverityDecider
 
         if (rule.DefaultSeverity >= DiagnosticSeverity.Warning)
         {
-            throw new InvalidOperationException("Rules with severity >= Warning are expected to be enabled by default.");
+            throw new InvalidOperationException($"Rule '{rule.Id}' with severity >= Warning is expected to be enabled by default.");
         }
 
         // If a rule is enabled by default, Recommended keeps it as Info.
--- a/src/Analyzers/MSTest.GlobalConfigsGenerator/AnalyzerSeverityDecider.cs
+++ b/src/Analyzers/MSTest.GlobalConfigsGenerator/AnalyzerSeverityDecider.cs
@@ -3,15 +3,23 @@
 
 using Microsoft.CodeAnalysis;
 
+using MSTest.Analyzers.Helpers;
+
 // NOTE for our current matrix.
 // We decide whether to include a rule in a specific mode or not based on two factors.
 // 1. IsEnabledByDefault (true/false)
-// 2. DefaultSeverity (Info, Warn) - note: we currently don't use severity Hidden or Error.
-// So, we have 4 possible combinations.
-// 1. IsEnabledByDefault = true, DefaultSeverity = Warn (we include that in All, Recommended, and Default)
-// 2. IsEnabledByDefault = true, DefaultSeverity = Info (we include that in All and Recommended)
-// 3. IsEnabledByDefault = false, DefaultSeverity = Warn (we don't combine Warn with not enabled by default)
-// 4. IsEnabledByDefault = false, DefaultSeverity = Info (we include that in All only)
+// 2. DefaultSeverity (Info, Warn, Error) - note: we currently don't use severity Hidden or Error (but we intend to use Error in future)
+// So, we have 6 possible combinations.
+// 1. IsEnabledByDefault = true, DefaultSeverity = Error (we include that in All, Recommended, and Default)
+// 2. IsEnabledByDefault = true, DefaultSeverity = Warn (we include that in All, Recommended, and Default)
+// 3. IsEnabledByDefault = true, DefaultSeverity = Info (we include that in All and Recommended)
+// 4. IsEnabledByDefault = false, DefaultSeverity = Error (we don't combine Error with not enabled by default)
+// 5. IsEnabledByDefault = false, DefaultSeverity = Warn (we don't combine Warn with not enabled by default)
+// 6. IsEnabledByDefault = false, DefaultSeverity = Info (we include that in All only)
+//
+// In addition to that, there are two custom tags that influence the logic.
+// 1. EscalateToErrorInRecommended (makes a specific rule as "error" when using Recommended or All mode)
+// 2. DisableInAllMode (Always disable the rule - it's completely opt-in)
 internal static class AnalyzerSeverityDecider
 {
     public static DiagnosticSeverity? GetSeverity(DiagnosticDescriptor rule, MSTestAnalysisMode mode)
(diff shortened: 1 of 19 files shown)


src/Analyzers/MSTest.GlobalConfigsGenerator/AnalyzerSeverityDecider.cs
```diff
@@ -0,0 +1,79 @@
...
+            MSTestAnalysisMode.Default => DecideForModeDefault(rule),
+            MSTestAnalysisMode.None => DecideForModeNone(rule),
+            _ => throw new ArgumentException($"Unexpected MSTestAnalysisMode '{mode}'.", nameof(mode)),
+        };
+
+        return severity;
+    }
+
+    private static DiagnosticSeverity? DecideForModeAll(DiagnosticDescriptor rule) =>
+        // TODO: We should consider at least not enabling "conflicting" rules.
```
Youssef1313: We should address this, I think.

Evangelink: Yes I agree. Shall we have some extra tags for such rules?

src/Analyzers/MSTest.GlobalConfigsGenerator/AnalyzerSeverityDecider.cs
```diff
@@ -0,0 +1,79 @@
...
+
+// NOTE for our current matrix.
+// We decide whether to include a rule in a specific mode or not based on two factors.
+// 1. IsEnabledByDefault (true/false)
+// 2. DefaultSeverity (Info, Warn) - note: we currently don't use severity Hidden or Error.
+// So, we have 4 possible combinations.
+// 1. IsEnabledByDefault = true, DefaultSeverity = Warn (we include that in All, Recommended, and Default)
+// 2. IsEnabledByDefault = true, DefaultSeverity = Info (we include that in All and Recommended)
+// 3. IsEnabledByDefault = false, DefaultSeverity = Warn (we don't combine Warn with not enabled by default)
+// 4. IsEnabledByDefault = false, DefaultSeverity = Info (we include that in All only)
```
Evangelink: Not a bad strategy, I am wondering if we should have some extra tags on each rules to help with that. WDYT?

src/Analyzers/MSTest.GlobalConfigsGenerator/AnalyzerSeverityDecider.cs
```diff
@@ -0,0 +1,79 @@
...
+        };
+
+        return severity;
+    }
+
+    private static DiagnosticSeverity? DecideForModeAll(DiagnosticDescriptor rule) =>
+        // TODO: We should consider at least not enabling "conflicting" rules.
+        // Or alternatively, have a configuration with reasonable default for such rules.
+        // Example of conflicting rules is an analyzer that suggests Dispose instead of
+        // cleanup, and another analyzer that suggests the opposite.
```
Evangelink: For this example, I am wondering if we should have either some user defined property/editorconfig or have more profiles. We have a few rules that are more around "favor .NET lifecycle" (in which case we prefer ctor and dispose) or favor consistency in which case we go for the opposite.

src/Analyzers/MSTest.GlobalConfigsGenerator/MSTest.GlobalConfigsGenerator.csproj
```diff
@@ -0,0 +1,12 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+  <PropertyGroup>
```
Evangelink: Nit:

```suggestion
﻿<Project Sdk="Microsoft.NET.Sdk">
﻿
  <PropertyGroup>
```

src/Analyzers/MSTest.GlobalConfigsGenerator/MSTest.GlobalConfigsGenerator.csproj
```diff
@@ -0,0 +1,12 @@
...
+    <TargetFramework>$(NetCurrent)</TargetFramework>
+    <OutputType>Exe</OutputType>
+    <IsPackable>false</IsPackable>
+    <IsShipping>false</IsShipping>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <PackageReference Include="Microsoft.CodeAnalysis.CSharp.Workspaces" />
+  </ItemGroup>
+</Project>
```
Evangelink: Nit:

```suggestion
  </ItemGroup>

</Project>
```

src/Analyzers/MSTest.GlobalConfigsGenerator/MSTestAnalysisMode.cs
```diff
@@ -0,0 +1,27 @@
...
+internal enum MSTestAnalysisMode
+{
+    /// <summary>
+    /// Disables all analyzers. User will need to opt-in to all analyzers.
+    /// </summary>
+    None,
+
+    /// <summary>
+    /// Follow the current behavior.
+    /// Rules that are enabled by default as warnings remain enabled as warnings. Anything else isn't enabled by default.
```
Evangelink, line 13: Maybe we can simplify by saying we respect definition of the analyzer (enabled by default + severity).

Evangelink (commented):
So glad to see this PR, you really rock @Youssef1313 🚀

src/Analyzers/MSTest.GlobalConfigsGenerator/AnalyzerSeverityDecider.cs
Youssef1313: @Evangelink We can either implement it "in place" inside the AnalyzerSeverityDecider or have an extra tag. I'm not sure why roslyn-analyzers went with extra tags but implementing in-place is a bit easier for sure. Let's discuss later (by later I don't mean after merging the PR but on Tuesday maybe).

src/Analyzers/MSTest.GlobalConfigsGenerator/AnalyzerSeverityDecider.cs
Youssef1313: @Evangelink You mean have define the minimum mode in each rule? Tags like `AllOnly`, `RecommendedAndAll`, `DefaultAndRecommendedAndAll` (putting naming aside)? Then each rule has to explicitly declare the right tag?

Or did you mean something like "decide the applicable modes automatically based on `IsEnabledByDefault` and `DefaultSeverity`, but have tags to "overwrite" that? (e.g, for rules that we want to disable even in `All` mode)

I'm not really sure if we can start with the simpler approach of hard-coding that inside AnalyzerSeverityDecider, and implement tags later if we found a problem with hard-coding. That kinda overlaps with your previous comment as well. Let's discuss on Tuesday.

src/Analyzers/MSTest.GlobalConfigsGenerator/AnalyzerSeverityDecider.cs
Youssef1313: We could go with an editorconfig option to determine the preferred style for "conflicting" rules and have what we think is a "reasonable" default. Then the analyzer can get the property value and decide if it should analyzer or not. i.e, it will be that only one of the two can be analyzing. It's kind of a separate feature but is related so I can have at least a quick investigation on that.

Evangelink (approved):

PR #4723, opened 2025-01-19
[Refactor] Simplify FixtureTestResult

Youssef1313:

### DIFF
--- a/src/Adapter/MSTest.TestAdapter/Execution/TestExecutionManager.cs
+++ b/src/Adapter/MSTest.TestAdapter/Execution/TestExecutionManager.cs
@@ -590,7 +590,7 @@ public class TestExecutionManager
             {
                 var result = new TestTools.UnitTesting.TestResult()
                 {
-                    Outcome = fixtureTestResult.Outcome.ToAdapterOutcome(),
+                    Outcome = fixtureTestResult.Outcome,
                 };
                 SendTestResults(currentTest, [result], DateTimeOffset.Now, DateTimeOffset.Now, testExecutionRecorder);
             }
--- a/src/Adapter/MSTest.TestAdapter/ObjectModel/FixtureTestResult.cs
+++ b/src/Adapter/MSTest.TestAdapter/ObjectModel/FixtureTestResult.cs
@@ -1,16 +1,17 @@
 ﻿// Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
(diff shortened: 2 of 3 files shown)


src/Adapter/MSTest.TestAdapter/Execution/TestExecutionManager.cs
```diff
@@ -590,7 +590,7 @@ private async Task ExecuteTestsWithTestRunnerAsync(
             {
                 var result = new TestTools.UnitTesting.TestResult()
                 {
-                    Outcome = fixtureTestResult.Outcome.ToAdapterOutcome(),
+                    Outcome = fixtureTestResult.Outcome,
```
Youssef1313, line 593: `ToAdapterOutcome` is actually taking adapter outcome and converts it to TestFramework outcome, but anyways. Changed the type of `Outcome` property inside `FixtureTestResult` so that now no conversion is needed here.

src/Adapter/MSTest.TestAdapter/ObjectModel/FixtureTestResult.cs
```diff
@@ -21,10 +22,5 @@ internal FixtureTestResult(bool isExecuted, UnitTestOutcome outcome, string? exc
     /// <summary>
     /// Gets the outcome of the test.
     /// </summary>
-    public UnitTestOutcome Outcome { get; }
-
-    /// <summary>
-    /// Gets the exception message if any.
-    /// </summary>
-    public string? ExceptionMessage { get; }
```
Youssef1313, line 29: This was never used. Thankfully, `FixtureTestResult` isn't public so we can remove.

src/Adapter/MSTest.TestAdapter/ObjectModel/FixtureTestResult.cs
```diff
@@ -1,16 +1,17 @@
 ﻿// Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
+using UTF = Microsoft.VisualStudio.TestTools.UnitTesting;
+
 namespace Microsoft.VisualStudio.TestPlatform.MSTest.TestAdapter.ObjectModel;
 
 [Serializable]
-internal sealed class FixtureTestResult
+internal readonly struct FixtureTestResult
```
Youssef1313, line 9: FixtureTestResult is very lightweight that's returned from a method just to read the properties immediately and do nothing more with it. So changed it to struct.

Evangelink (dismissed):

PR #4725, opened 2025-01-19
Remove AsyncTestProperty

Youssef1313:
I can't see any meaningful usage of it.

@Evangelink Do you know if it's used by a component outside of testfx repo?


Evangelink (commented):
I don't. Have you checked on grep.app?

Youssef1313:
> I don't. Have you checked on grep.app?

@Evangelink I was more suspecting that if there is a usage, it could be buried somewhere in an internal repo (similar to #4314 which is *probably* used by AzDO)

Evangelink:
I'm fine to drop and bring back in case that was used somewhere we didn't know about

Evangelink (approved):

PR #4734, opened 2025-01-20
Support conditional tests

### LINKED ISSUE #4700: Add OS related attribute
## Summary

Add a new attribute that will skip running the test if the OS doesn't match.

## Background and Motivation

There are many implementation of OS skip attributes in other test framework but none in MSTest. We should close the gap here.

Looking at people implementation, there are 2 main trends, an inclusive attribute like `SupportedOS` and an exclusive attribute like `IgnoreOnOS`. Both of these ways have cons when you want to have all but one (few) in which case the other attribute would be better.

## Proposed Feature

I would like to have our implementation allowing to choose between "include" and "exclude" mode (with most likely "include" as default behavior).
The attribute name could be:

- ConditionalOSAttribute
- OSConditionAttribute
- OSFilterAttribute

I think a naming like `Condition`, `Conditional` or `Filter` would give us more flexibility to introduce more features alike (e.g. conditional run based on architecture, based on locale, based on env var...)

## Alternative Designs

We could also ship 2 specific attributes
- SupportedOSAttribute
- UnsupportedOSAttribute

AB#2349734

Youssef1313:
Fixes #4700


src/TestFramework/TestFramework/Attributes/TestMethod/ConditionalTestBaseAttribute.cs
```diff
@@ -0,0 +1,25 @@
...
+namespace Microsoft.VisualStudio.TestTools.UnitTesting;
+
+/// <summary>
+/// This attribute is used to ignore a test class or a test method, based on a condition and using an optional message.
+/// </summary>
+/// <remarks>
+/// This attribute isn't inherited. Applying it to a base class will not cause derived classes to be ignored.
+/// </remarks>
+[AttributeUsage(AttributeTargets.Class | AttributeTargets.Method, Inherited = false)]
+public abstract class ConditionalTestBaseAttribute : Attribute
```
Youssef1313: @Evangelink Is `IgnoreIfBaseAttribute` or `IgnoreIfAttribute` a better name?

The derived classes can then be named `IgnoreIfOperatingSystem` and `IgnoreIfNotOperatingSystem`. That makes the whole attribute hierarchy consistent in terms of naming and makes the intent clear.

src/TestFramework/TestFramework/Attributes/TestMethod/ConditionBaseAttribute.cs
```diff
@@ -0,0 +1,52 @@
...
+/// <remarks>
+/// This attribute isn't inherited. Applying it to a base class will not affect derived classes.
+/// </remarks>
+[AttributeUsage(AttributeTargets.Class | AttributeTargets.Method, Inherited = false)]
+public abstract class ConditionBaseAttribute : Attribute
+{
+    protected ConditionBaseAttribute(Mode mode)
+        => Mode = mode;
+
+    protected ConditionBaseAttribute()
```
Evangelink: Do we want to allow this one or force subclasses to select the mode?

src/TestFramework/TestFramework/Attributes/TestMethod/ConditionBaseAttribute.cs
```diff
@@ -0,0 +1,52 @@
...
+    {
+    }
+
+    internal Mode Mode { get; }
+
+    /// <summary>
+    /// Gets the ignore message (in case <see cref="ShouldRun"/> returns <see langword="false"/>) indicating
+    /// the reason for ignoring the test method or test class.
+    /// </summary>
+    public abstract string? ConditionalIgnoreMessage { get; }
```
Evangelink: Maybe just `IgnoreMessage`?

src/TestFramework/TestFramework/Attributes/TestMethod/OSConditionAttribute.cs
```diff
@@ -0,0 +1,62 @@
...
+[AttributeUsage(AttributeTargets.Class | AttributeTargets.Method, Inherited = false)]
+public sealed class OSConditionAttribute : ConditionBaseAttribute
+{
+    private readonly OperatingSystems _operatingSystems;
+
+    /// <summary>
+    /// Initializes a new instance of the <see cref="OSConditionAttribute"/> class.
+    /// </summary>
+    /// <param name="operatingSystems">The operating systems that this test supports.</param>
+    public OSConditionAttribute(OperatingSystems operatingSystems, Mode mode)
```
Evangelink: I was thinking to have mode first for consistency with other attributes we will/would ship and will be string based (params).

src/TestFramework/TestFramework/Attributes/TestMethod/IgnoreAttribute.cs
```diff
@@ -26,10 +26,16 @@ public IgnoreAttribute()
...
     /// Gets the ignore message indicating the reason for ignoring the test method or test class.
     /// </summary>
-    public string? IgnoreMessage { get; }
+    public string? IgnoreMessage => ConditionalIgnoreMessage;
+
+    public override bool ShouldRun => false;
+
+    public override string? ConditionalIgnoreMessage { get; }
+
+    public override string GroupName => nameof(IgnoreAttribute);
```
Evangelink: Shall we skip the `attribute` suffix?

src/TestFramework/TestFramework/Attributes/TestMethod/OSConditionAttribute.cs
```diff
@@ -0,0 +1,62 @@
...
+
+namespace Microsoft.VisualStudio.TestTools.UnitTesting;
+
+/// <summary>
+/// This attribute is used to ignore a test class or a test method, with an optional message.
+/// </summary>
+/// <remarks>
+/// This attribute isn't inherited. Applying it to a base class will not cause derived classes to be ignored.
+/// </remarks>
+[AttributeUsage(AttributeTargets.Class | AttributeTargets.Method, Inherited = false)]
```
Evangelink: Let's be explicit on the allow multiple please.

src/Adapter/MSTest.TestAdapter/Helpers/AttributeHelpers.cs
```diff
@@ -0,0 +1,40 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+
+namespace Microsoft.VisualStudio.TestPlatform.MSTest.TestAdapter.Helpers;
+
+internal static class AttributeHelpers
+{
+    public static bool IsIgnored(ICustomAttributeProvider type, out string? ignoreMessage)
```
Evangelink: Let's make this an extension method

src/Adapter/MSTest.TestAdapter/Helpers/AttributeHelpers.cs
```diff
@@ -0,0 +1,40 @@
...
+        IEnumerable<ConditionBaseAttribute> attributes = ReflectHelper.Instance.GetDerivedAttributes<ConditionBaseAttribute>(type, inherit: false);
+        IEnumerable<IGrouping<string, ConditionBaseAttribute>> groups = attributes.GroupBy(attr => attr.GroupName);
+        foreach (IGrouping<string, ConditionBaseAttribute>? group in groups)
+        {
+            bool atLeastOneInGroupIsSatisfied = false;
+            string? firstNonSatisfiedMatch = null;
+            foreach (ConditionBaseAttribute attribute in group)
+            {
+                bool shouldRun = attribute.Mode == Mode.Include ? attribute.ShouldRun : !attribute.ShouldRun;
+                if (attribute.ShouldRun)
```
Evangelink: Shall you use the variable above instead?

src/TestFramework/TestFramework/Attributes/TestMethod/ConditionBaseAttribute.cs
```diff
@@ -0,0 +1,52 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+namespace Microsoft.VisualStudio.TestTools.UnitTesting;
+
+public enum Mode
```
Evangelink: Let's go with `ConditionMode`

src/TestFramework/TestFramework/Attributes/TestMethod/ConditionBaseAttribute.cs
```diff
@@ -0,0 +1,52 @@
...
+{
+    Include,
```
Evangelink: Xml doc: 
Include: runs only when the condition is met.
Exclude: skips when the condition is met.

src/TestFramework/TestFramework/Attributes/TestMethod/ConditionBaseAttribute.cs
Youssef1313: We can force for now. Easy to introduce the ctor later **if** we wanted.

src/TestFramework/TestFramework/Attributes/TestMethod/IgnoreAttribute.cs
Youssef1313: Yes. I was only preferring `nameof` rather than literal strings. But doesn't harm as `"Ignore"`

src/Adapter/MSTest.TestAdapter/Helpers/AttributeHelpers.cs
Youssef1313: I'm surprised we don't have a warning for unused local variable. Definitely a bug to not use it.

Evangelink (approved):

PR #4754, opened 2025-01-22
Add UsingMSTestSdk property

aelij:
The `Microsoft.NET.Sdk` sets a `UsingMicrosoftNETSdk` property which allows other imported files to rely on it, for example in Directory.Build.{props, targets}. This PR adds a similar `UsingMSTestSdk` for the MSTest SDK.


Evangelink (commented):
Thanks! It seems to make sense. @baronfel @rainersigwald could you please confirm?

Evangelink:
Thanks @aelij

PR #4758, opened 2025-01-23
Fix wrong Trx warning

Youssef1313:


Evangelink:
Let's add some acceptance test. It should be relatively easy to reproduce by calling run twice and asserting no message on first and message on second call .

Evangelink (dismissed):

PR #4769, opened 2025-01-23
MSTEST0030: add codefix

Evangelink:

### DIFF
--- a/src/Analyzers/MSTest.Analyzers.CodeFixes/AddTestClassFixer.cs
+++ b/src/Analyzers/MSTest.Analyzers.CodeFixes/AddTestClassFixer.cs
@@ -18,12 +18,14 @@ using MSTest.Analyzers.Helpers;
 
 namespace MSTest.Analyzers;
 
-[ExportCodeFixProvider(LanguageNames.CSharp, Name = nameof(PublicTypeShouldBeTestClassFixer))]
+[ExportCodeFixProvider(LanguageNames.CSharp, Name = nameof(AddTestClassFixer))]
 [Shared]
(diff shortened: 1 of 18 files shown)


src/Analyzers/MSTest.Analyzers.CodeFixes/AddTestClassFixer.cs
```diff
@@ -49,7 +51,7 @@ public override async Task RegisterCodeFixesAsync(CodeFixContext context)
             CodeAction.Create(
                 title: CodeFixResources.PublicTypeShouldBeTestClassFix,
                 createChangedDocument: c => AddTestClassAttributeAsync(context.Document, declaration, c),
-                equivalenceKey: nameof(PublicTypeShouldBeTestClassFixer)),
+                equivalenceKey: nameof(AddTestClassFixer)),
```
Youssef1313: I don't know what behavior would we prefer, but doing it this way will cause a FixAll invocation for one diagnostic to fix things for both diagnostics. So not sure if that's desired from user point of view.

src/Analyzers/MSTest.Analyzers.CodeFixes/AddTestClassFixer.cs
```diff
@@ -49,7 +51,7 @@ public override async Task RegisterCodeFixesAsync(CodeFixContext context)
             CodeAction.Create(
                 title: CodeFixResources.PublicTypeShouldBeTestClassFix,
```
Youssef1313: Not sure if it's worth renaming the resource similar to the fixer rename.

Evangelink:
@Youssef1313 ready for review

Youssef1313 (approved):

PR #4784, opened 2025-01-25
Disable MSTEST0015 in various profiles

### LINKED ISSUE #4780: MSTEST0015: Test method should not be ignored shouldn't be enabled in Recommended (maybe even All)

Evangelink:
Fixes #4780

### DIFF
--- a/src/Analyzers/MSTest.Analyzers/TestMethodShouldNotBeIgnoredAnalyzer.cs
+++ b/src/Analyzers/MSTest.Analyzers/TestMethodShouldNotBeIgnoredAnalyzer.cs
@@ -29,7 +29,8 @@ public sealed class TestMethodShouldNotBeIgnoredAnalyzer : DiagnosticAnalyzer
         Description,
         Category.Design,
         DiagnosticSeverity.Info,
-        isEnabledByDefault: true);
+        isEnabledByDefault: true,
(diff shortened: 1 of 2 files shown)


src/Analyzers/MSTest.Analyzers/TestMethodShouldNotBeIgnoredAnalyzer.cs
```diff
@@ -29,7 +29,8 @@ public sealed class TestMethodShouldNotBeIgnoredAnalyzer : DiagnosticAnalyzer
         Description,
         Category.Design,
         DiagnosticSeverity.Info,
-        isEnabledByDefault: true);
+        isEnabledByDefault: true,
+        disableInAllMode: true);
```
Youssef1313: The implementation for globalconfig generation will prevent having something enabled by default but disabled in all mode.
Because it's not super obvious to me what the behavior should be.

Maybe in this case, it should remain "info" in all mode, and not completely disabled?

Evangelink: I'll make it info but still disabled as it can be seen as noise

Youssef1313 (approved):

PR #4786, opened 2025-01-26
Refactor some logic out of TypeCache

Youssef1313:


src/Adapter/MSTest.TestAdapter/Execution/TypeCache.cs
```diff
@@ -699,26 +697,7 @@ private TestMethodInfo ResolveTestMethodInfoForDiscovery(TestMethod testMethod,
         MethodInfo methodInfo = GetMethodInfoForTestMethod(testMethod, testClassInfo);
 
         // Let's build a fake options type as it won't be used.
```
Youssef1313: @Evangelink Here, things like timeout, test method attribute were not calculated. With the refactoring of the logic into TestMethodInfo, they will be calculated.

Is there a perf concern here for discovery? Maybe we can use `Lazy<T>`?

Youssef1313: Since #4958, there shouldn't be a perf concern. I'm more confident now to proceed with this PR :)
I'll need to resolve the conflict and get it green.

Evangelink:
Some types are not found @Youssef1313

Evangelink (dismissed):

Evangelink (approved):

PR #4795, opened 2025-01-27
Open experimentally Configuration and Logging

Evangelink:

### DIFF
--- a/src/Platform/Microsoft.Testing.Platform/Builder/TestApplicationBuilder.cs
+++ b/src/Platform/Microsoft.Testing.Platform/Builder/TestApplicationBuilder.cs
@@ -58,9 +58,11 @@ internal sealed class TestApplicationBuilder : ITestApplicationBuilder
 
     internal ITestHostOrchestratorManager TestHostControllersManager => _testHostBuilder.TestHostOrchestratorManager;
 
-    internal IConfigurationManager Configuration => _testHostBuilder.Configuration;
+    [Experimental("TPEXP", UrlFormat = "https://aka.ms/testingplatform/diagnostics#{0}")]
+    public IConfigurationManager Configuration => _testHostBuilder.Configuration;
 
-    internal ILoggingManager Logging => _testHostBuilder.Logging;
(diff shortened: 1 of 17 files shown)


src/Platform/Microsoft.Testing.Platform/Builder/TestApplicationBuilder.cs
```diff
@@ -58,9 +58,11 @@ internal TestApplicationBuilder(
 
     internal ITestHostOrchestratorManager TestHostControllersManager => _testHostBuilder.TestHostOrchestratorManager;
```
Youssef1313, line 59: Unrelated to your changes, but all other properties of type "SomethingManager" are named "Something" without "Manager". This is currently internal so not a big deal, but may be worth renaming in a separate PR so that we don't unintentionally end up with naming inconsistency if we decided to make it public in future.

src/Platform/Microsoft.Testing.Platform/CommandLine/OptionRecord.cs
```diff
@@ -3,7 +3,8 @@
 
 namespace Microsoft.Testing.Platform.CommandLine;
 
-internal sealed class OptionRecord(string option, string[] arguments)
+[Experimental("TPEXP", UrlFormat = "https://aka.ms/testingplatform/diagnostics#{0}")]
+public sealed class OptionRecord(string option, string[] arguments)
```
Youssef1313: Double-checking if the naming is intended. Do we want to rename `option` to `name`? Or are we fine with having "option" duplicated in type name and the property/parameter names?

Youssef1313: Not sure also if `CommandLineOption` is a better name than `OptionRecord`

Youssef1313 (dismissed):
LGTM with few comments about naming

Youssef1313 (dismissed):

Youssef1313:
@Evangelink I think this was accidentally closed by the other PR.

Evangelink:
Thanks 👍

PR #4803, opened 2025-01-27
MSTEST0029 do not report on implementation of interface

### LINKED ISSUE #4660: MSTEST0029 (Public method should be test method) false positive on public `Dispose`
Having a TestClass that implements `IDisposable` and implement `Dispose` method implicitly will trigger MSTEST0029

```csharp
	[TestClass]
	public class TestClass1 : IDisposable
	{
		[TestMethod]
		public void TestMethod1()
		{
		}

		public void Dispose() { }
	}
```

Evangelink:
Fixes #4660.

I am not sure what's the best way to really cover/document that we exclude all interfaces.

### DIFF
--- a/src/Analyzers/MSTest.Analyzers/PublicMethodShouldBeTestMethodAnalyzer.cs
+++ b/src/Analyzers/MSTest.Analyzers/PublicMethodShouldBeTestMethodAnalyzer.cs
@@ -64,7 +64,9 @@ public sealed class PublicMethodShouldBeTestMethodAnalyzer : DiagnosticAnalyzer
             return;
         }
 
-        if (!methodSymbol.HasValidTestMethodSignature(taskSymbol, valueTaskSymbol, canDiscoverInternals))
+        if (!methodSymbol.HasValidTestMethodSignature(taskSymbol, valueTaskSymbol, canDiscoverInternals)
+            || methodSymbol.IsVirtual
+            || methodSymbol.IsOverride)
         {
             return;
         }
--- a/src/Analyzers/MSTest.Analyzers/PublicMethodShouldBeTestMethodAnalyzer.cs
+++ b/src/Analyzers/MSTest.Analyzers/PublicMethodShouldBeTestMethodAnalyzer.cs
@@ -85,6 +85,12 @@ public sealed class PublicMethodShouldBeTestMethodAnalyzer : DiagnosticAnalyzer
             return;
         }
 
+        // We consider that if the method implements an interface member, it is not a test method.
+        if (methodSymbol.IsImplementationOfAnyInterfaceMember())
+        {
+            return;
+        }
+
         ImmutableArray<AttributeData> methodAttributes = methodSymbol.GetAttributes();
         // check if the method has testMethod, testInitialize or testCleanup attribute
         bool hasValidAttribute = false;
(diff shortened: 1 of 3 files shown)


test/UnitTests/MSTest.Analyzers.UnitTests/PublicMethodShouldBeTestMethodAnalyzerTests.cs
```diff
@@ -252,4 +252,22 @@ public void TestCleanup()
...
+
+    [TestMethod]
+    public async Task WhenMethodIsPublicAndImplementsInterface_NoDiagnostic()
+    {
+        string code = """
+            using Microsoft.VisualStudio.TestTools.UnitTesting;
+            using System;
+
+            [TestClass]
+            public class MyTestClass : IDisposable
```
Youssef1313, line 264: It might be worth adding a test for user-defined interface

src/Analyzers/MSTest.Analyzers/RoslynAnalyzerHelpers/IMethodSymbolExtensions.cs
```diff
@@ -10,6 +10,96 @@ namespace Analyzer.Utilities.Extensions;
...
+
+        if (symbol is IEventSymbol eventSymbol && !eventSymbol.ExplicitInterfaceImplementations.IsEmpty)
+        {
+            return true;
+        }
+
+        return false;
+    }
+
+    public static bool IsImplementationOfAnyImplicitInterfaceMember(this ISymbol symbol)
```
Youssef1313: Not sure if this has changed since I filed https://github.com/dotnet/roslyn-analyzers/issues/5878. It may be worth actually double checking if my report is correct. But it's best if we can avoid porting bugs from roslyn-analyzers.

test/UnitTests/MSTest.Analyzers.UnitTests/PublicMethodShouldBeTestMethodAnalyzerTests.cs
Youssef1313, line 1: Very niche case, but consider adding a test for an abstract test class that implements IDisposable as a virtual method, and having the concrete test class *override* Dispose from base.

src/Analyzers/MSTest.Analyzers/RoslynAnalyzerHelpers/IMethodSymbolExtensions.cs
Evangelink: I confirm this is handling both implicit and explicit.

src/Analyzers/MSTest.Analyzers/PublicMethodShouldBeTestMethodAnalyzer.cs
```diff
@@ -64,7 +64,9 @@ private static void AnalyzeSymbol(SymbolAnalysisContext context, INamedTypeSymbo
             return;
         }
 
-        if (!methodSymbol.HasValidTestMethodSignature(taskSymbol, valueTaskSymbol, canDiscoverInternals))
+        if (!methodSymbol.HasValidTestMethodSignature(taskSymbol, valueTaskSymbol, canDiscoverInternals)
+            || methodSymbol.IsVirtual
+            || methodSymbol.IsOverride)
```
Youssef1313, line 69: This is skipping virtual/override completely, is it intended?
Or do we want to walk the OverriddenMethod chain and skip only if something in the chain is an implementation of interface member?

Youssef1313 (approved):

src/Analyzers/MSTest.Analyzers/PublicMethodShouldBeTestMethodAnalyzer.cs
Evangelink: Let's ignore that for now.

PR #4812, opened 2025-01-28
Retry: mark as experimental extension points

Evangelink:
Following internal discussions, there are some unclear parts as to how we want to really allow extension of the retry capabilities so marking it as experimental so we can allow some extensibility while reserving right to modify the public API.

### DIFF
--- a/src/TestFramework/TestFramework/Attributes/TestMethod/RetryBaseAttribute.cs
+++ b/src/TestFramework/TestFramework/Attributes/TestMethod/RetryBaseAttribute.cs
@@ -21,6 +21,7 @@ public abstract class RetryBaseAttribute : Attribute
     /// The other results are currently not used, but may be used in the future for tooling to show the
     /// state of the failed attempts.
     /// </returns>
+    [Experimental("MSTESTEXP", UrlFormat = "https://aka.ms/mstest/diagnostics#{0}")]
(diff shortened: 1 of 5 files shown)


src/TestFramework/TestFramework/Attributes/TestMethod/RetryBaseAttribute.cs
```diff
@@ -21,6 +21,7 @@ public abstract class RetryBaseAttribute : Attribute
     /// The other results are currently not used, but may be used in the future for tooling to show the
     /// state of the failed attempts.
     /// </returns>
+    [Experimental("MSTESTEXP", UrlFormat = "https://aka.ms/mstest/diagnostics#{0}")]
```
Youssef1313, line 24: This redirects to https://learn.microsoft.com/en-us/dotnet/core/testing/mstest-analyzers/overview, are we going to documentation MSTESTEXP there (even though it's not an analyzer)?

Youssef1313 (approved):
LGTM with small comment

src/TestFramework/TestFramework/Attributes/TestMethod/RetryBaseAttribute.cs
Evangelink: I was wondering whether to have a separate page and decided it was probably "a kind of analyzer". I am doing the additional section as we speak.

PR #4814, opened 2025-01-28
Make SelfRegisteredExtensions and EntryPoint types part of namespace

### LINKED ISSUE #3813: Build error when one test project references another
## Describe the bug

Starting with MSTest 3.6.0, when one test project references another test project, warning CS0436 is emitted by the compiler because the same type is being compiled into two assemblies.

## Steps To Reproduce

1. Pull [`build-error-on-reference` branch](https://github.com/bdukes/mstest-demo/tree/build-error-on-reference) from [bdukes/mstest-demo](https://github.com/bdukes/mstest-demo)
2. `dotnet build`

## Expected behavior

The build command is successful.

## Actual behavior

The build command fails with the following error:

```
I:\code\MSTestDemo\ProjectTwo.Tests\obj\Debug\net8.0\TestPlatformEntryPoint.cs(13,9): warning CS0436: The type 'SelfRegi
steredExtensions' in 'I:\code\MSTestDemo\ProjectTwo.Tests\obj\Debug\net8.0\AutoRegisteredExtensions.cs' conflicts with t
he imported type 'SelfRegisteredExtensions' in 'ProjectOne.Tests, Version=1.0.0.0, Culture=neutral, PublicKeyToken=null'
. Using the type defined in 'I:\code\MSTestDemo\ProjectTwo.Tests\obj\Debug\net8.0\AutoRegisteredExtensions.cs'. [I:\code
\MSTestDemo\ProjectTwo.Tests\ProjectTwo.Tests.csproj]
```

### LINKED ISSUE #3935: Build error when one test project references another, redux
## Describe the bug
I have the same exact issue as in #3813, it is not solved for me by the PR, because my test assemblies use InternalsVisibleTo, so they can see autogenerated Internal classes, so they still get the conflict:

`D:\dbs\sh\l23t\1010_175302_0\cmd\7\Test\UnitTests\BIAndReporting\Reports.Shared\obj\Debug\net48\TestPlatformEntryPoint.cs(13,9): warning CS0436: The type 'SelfRegisteredExtensions' in 'D:\dbs\sh\l23t\1010_175302_0\cmd\7\Test\UnitTests\BIAndReporting\Reports.Shared\obj\Debug\net48\AutoRegisteredExtensions.cs' conflicts with the imported type 'SelfRegisteredExtensions' in 'MS.Dynamics.Test.BIAndReporting.Common.UnitTests, Version=7.0.0.0, Culture=neutral, PublicKeyToken=31bf3856ad364e35'. Using the type defined in 'D:\dbs\sh\l23t\1010_175302_0\cmd\7\Test\UnitTests\BIAndReporting\Reports.Shared\obj\Debug\net48\AutoRegisteredExtensions.cs'. [D:\dbs\sh\l23t\1010_175302_0\cmd\7\Test\UnitTests\BIAndReporting\Reports.Shared\Shared.UnitTests.csproj::TargetFramework=net48]`

## Steps To Reproduce

[assembly: InternalsVisibleTo("MS.Dynamics.Test.BIAndReporting.RunTime.UnitTests, PublicKey=blah")]

Evangelink:
Fixes #3935 and relates/fixes #3813


src/Platform/Microsoft.Testing.Platform.MSBuild/Tasks/TestingPlatformAutoRegisteredExtensions.cs
```diff
@@ -41,6 +41,9 @@ internal TestingPlatformAutoRegisteredExtensions(IFileSystem fileSystem)
     [Required]
     public ITaskItem[] AutoRegisteredExtensionsBuilderHook { get; set; }
 
+    [Required]
+    public string? RootNamespace { get; set; }
```
Evangelink: As far as I can see it cannot be null, not sure if that's worth "handling" or not. I am also not sanitizing or anything as something else would fail anyways.

src/Platform/Microsoft.Testing.Platform.MSBuild/buildMultiTargeting/Microsoft.Testing.Platform.MSBuild.targets
```diff
@@ -195,6 +196,7 @@
           Condition=" '$(GenerateSelfRegisteredExtensions)' == 'True' " >
     <TestingPlatformAutoRegisteredExtensions AutoRegisteredExtensionsSourcePath="$(_AutoRegisteredExtensionsSourcePath)"
                                              AutoRegisteredExtensionsBuilderHook="@(TestingPlatformBuilderHook)"
+                                             RootNamespace="$(RootNamespace)"
```
Youssef1313, line 199: Is it valid for users to do `<RootNamespace></RootNamespace>`? In this case, our generated code will not compile.

src/Platform/Microsoft.Testing.Platform.MSBuild/Tasks/TestingPlatformEntryPointTask.cs
```diff
@@ -54,82 +57,89 @@ public override bool Execute()
...
-Module TestingPlatformEntryPoint
+Namespace {{rootNamespace}}
+    <System.Diagnostics.CodeAnalysis.ExcludeFromCodeCoverage>
+    Module TestingPlatformEntryPoint
 
-    Function Main(args As Global.System.String()) As Global.System.Int32
-        Return MainAsync(args).Result
-    End Function
+        Function Main(args As Global.System.String()) As Global.System.Int32
+            Return MainAsync(args).Result
```
Youssef1313, line 164: Unrelated to your changes, but this potentially doesn't match the behavior of C#. It may be worth updating this to `GetAwaiter().GetResult()` instead to match C#. And maybe have some test (at least manual) to verify that.

Do you want us to track it in an issue? Or was this an explicit decision for some reason?

src/Platform/Microsoft.Testing.Platform.MSBuild/buildMultiTargeting/Microsoft.Testing.Platform.MSBuild.targets
Evangelink: From looking up online, this is not supposed to be valid syntax but it can work when done in a target. I don't know if that's worth falling back to some default namespace or considering this as not supported. Any opinion?

src/Platform/Microsoft.Testing.Platform.MSBuild/Tasks/TestingPlatformEntryPointTask.cs
Evangelink: No idea. Let's see if fixing it causes any issue.

src/Platform/Microsoft.Testing.Platform.MSBuild/buildMultiTargeting/Microsoft.Testing.Platform.MSBuild.targets
Youssef1313: I'd say if it's possible for `RootNamespace` to be empty, then we should generate in global namespace (the old behavior). I see for example Roslyn is doing this:

https://github.com/dotnet/roslyn/blob/1b6cd2ab7a0862ea0c4d61b60d826429b2de2236/src/Workspaces/VisualBasicTest/Microsoft.CodeAnalysis.VisualBasic.Workspaces.UnitTests.vbproj#L9

Not entirely sure why this can be useful though.

Evangelink: Handled now

Youssef1313 (approved):

PR #4818, opened 2025-01-29
Move away from nuspec for MSTest metapackage

Youssef1313:


eng/verify-nupkgs.ps1
```diff
@@ -22,7 +22,7 @@ function Confirm-NugetPackages {
         "MSTest.Sdk"                                  = 15
         "MSTest.TestFramework"                        = 148
         "MSTest.TestAdapter"                          = 75
-        "MSTest"                                      = 6
```
Youssef1313, line 25: Extra `_._` in lib for:

- net6.0
- net7.0
- net8.0
- net9.0
- net462
- netcoreapp3.1
- netstandard2.0
- uap10.0

Total 8 TFMs. 6 (old files) + 8 (new files) = 14

Evangelink (commented):
Is `Microsoft.NET.Test.Sdk` still working as expected following this change? It's not using `buildTransitive`: 
image

eng/verify-nupkgs.ps1
Evangelink: Do we need all net TFMs? Or shall we match the adapter ones?

Youssef1313: The ones I added are the adapter ones per https://www.nuget.org/packages/MSTest.TestAdapter/3.7.3#dependencies-body-tab

Evangelink: Oh right! I thought we were only targeting net6.0! Sorry

Youssef1313:
@Evangelink That was indeed a real untested issue. Fixed and ensured it's tested.

Evangelink (approved):

PR #4823, opened 2025-01-29
Cleanup props/targets copy to output

Evangelink:

### DIFF
--- a/src/Adapter/MSTest.TestAdapter/MSTest.TestAdapter.csproj
+++ b/src/Adapter/MSTest.TestAdapter/MSTest.TestAdapter.csproj
@@ -89,19 +89,10 @@
   </ItemGroup>
 
   <ItemGroup>
-    <None Update="build\common\MSTest.TestAdapter.props" Condition=" '$(TargetFramework)' != '$(UwpMinimum)' ">
+    <None Update="build\common\*" Condition=" '$(TargetFramework)' != '$(UwpMinimum)' ">
       <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
     </None>
-    <None Update="build\netfx-netcore-netstandard\MSTest.TestAdapter.targets">
-      <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
-    </None>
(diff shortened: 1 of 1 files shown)


src/Adapter/MSTest.TestAdapter/MSTest.TestAdapter.csproj
```diff
@@ -89,19 +89,10 @@
   </ItemGroup>
 
   <ItemGroup>
-    <None Update="build\common\MSTest.TestAdapter.props" Condition=" '$(TargetFramework)' != '$(UwpMinimum)' ">
+    <None Update="build\common\*" Condition=" '$(TargetFramework)' != '$(UwpMinimum)' ">
```
Youssef1313, line 92: Is this a change in behavior? Previously the props was done with a condition while the targets didn't have a condition.

src/Adapter/MSTest.TestAdapter/MSTest.TestAdapter.csproj
```diff
@@ -89,19 +89,10 @@
...
       <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
     </None>
-    <None Update="build\netfx-netcore-netstandard\MSTest.TestAdapter.targets">
```
Youssef1313, line 95: This is gone completely now? Was it not needed? If so we should delete the file?

src/Adapter/MSTest.TestAdapter/MSTest.TestAdapter.csproj
Evangelink: Some "bugs" + dead code

src/Adapter/MSTest.TestAdapter/MSTest.TestAdapter.csproj
Evangelink: I don't see the file. Do you?

Youssef1313: Never mind. It's not there.

Issue #4070, opened 2024-11-15
MSTestRunner does not integrate with "VS MSBuild"


RuiAlias:
## Describe the bug

An MSTest.Sdk project cannot be tested with "VS MSBuild" (or "Full MSBuild").

## Steps To Reproduce

On a Windows machine with "VS MSBuild" and dotnet SDK installed:
```
dotnet new mstest
MSBuild.exe -restore -t:Build;Test mstest.csproj
```

You can also see the repro: https://github.com/RuiAlias/mtp-msbuild-repro. On this repro I created a GitHub Actions Workflow that runs the msbuild command and fails: https://github.com/RuiAlias/mtp-msbuild-repro/actions/runs/11858511987/job/33049304509.

## Expected behavior

`MSBuild.exe` runs the tests successfully.

## Actual behavior

`MSBuild.exe` fails with error:
```
D:\a\mtp-msbuild-repro\mtp-msbuild-repro\bin\Debug\net8.0\mtp-msbuild-repro.dll : error run failed: Tests failed: Possible reasons for this include: [D:\a\mtp-msbuild-repro\mtp-msbuild-repro\mtp-msbuild-repro.csproj]
D:\a\mtp-msbuild-repro\mtp-msbuild-repro\bin\Debug\net8.0\mtp-msbuild-repro.dll : error run failed: Could not execute because the specified command or file was not found. [D:\a\mtp-msbuild-repro\mtp-msbuild-repro\mtp-msbuild-repro.csproj]
D:\a\mtp-msbuild-repro\mtp-msbuild-repro\bin\Debug\net8.0\mtp-msbuild-repro.dll : error run failed:   * You misspelled a built-in dotnet command. [D:\a\mtp-msbuild-repro\mtp-msbuild-repro\mtp-msbuild-repro.csproj]
D:\a\mtp-msbuild-repro\mtp-msbuild-repro\bin\Debug\net8.0\mtp-msbuild-repro.dll : error run failed:   * You intended to execute a .NET program, but dotnet---internal-msbuild-node does not exist. [D:\a\mtp-msbuild-repro\mtp-msbuild-repro\mtp-msbuild-repro.csproj]
[...]

PR #4840, opened 2025-01-30
Fix InvokeTestingPlatformTask to handle running with msbuild.exe and 'Test' target

Youssef1313:
Fixes https://github.com/microsoft/testfx/issues/4070

### DIFF
--- a/src/Platform/Microsoft.Testing.Platform.MSBuild/Microsoft.Testing.Platform.MSBuild.csproj
+++ b/src/Platform/Microsoft.Testing.Platform.MSBuild/Microsoft.Testing.Platform.MSBuild.csproj
@@ -1,4 +1,4 @@
-<Project Sdk="Microsoft.NET.Sdk">
+﻿<Project Sdk="Microsoft.NET.Sdk">
   <PropertyGroup>
     <TargetFrameworks>$(MicrosoftTestingTargetFrameworks);netstandard2.0</TargetFrameworks>
     <DefineConstants>PLATFORM_MSBUILD</DefineConstants>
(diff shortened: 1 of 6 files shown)


Evangelink (dismissed):

Evangelink:
@Youssef1313 don't forget to link the ticket

rainersigwald:
Any hope of getting a binlog? I wonder if something is setting `DOTNET_HOST_PATH` or something?

Youssef1313:
@rainersigwald It's very possible that `DOTNET_HOST_PATH` is set. It's also being used by the logic in the task.

https://github.com/microsoft/testfx/blob/6294747c732ff6f343411d2142eccbaee62db995/src/Platform/Microsoft.Testing.Platform.MSBuild/buildMultiTargeting/Microsoft.Testing.Platform.MSBuild.targets#L322-L331

https://github.com/microsoft/testfx/blob/6294747c732ff6f343411d2142eccbaee62db995/src/Platform/Microsoft.Testing.Platform.MSBuild/Tasks/InvokeTestingPlatformTask.cs#L145-L149

It should be possible to get a binlog to be 100% sure. But I'm not sure why using `DOTNET_HOST_PATH` would be problematic?

Youssef1313:
[msbuild.zip](https://github.com/user-attachments/files/18653214/msbuild.zip)

Nope, `DOTNET_HOST_PATH` doesn't seem to be set

Youssef1313:
@rainersigwald Note: Microsoft.Testing.Platform.MSBuild project has a dependency on `Microsoft.Win32.Registry` NuGet package because we try to read registry key when resolving for `dotnet.exe` on Windows. I assume our packaging here for the MSBuild task is incorrect and we should ship dlls of dependencies (such as Microsoft.Win32.Registry) in `_MSBuildTasks` folder of the package (this is the folder that we use the task from in `UsingTask`)

src/Platform/Microsoft.Testing.Platform.MSBuild/Microsoft.Testing.Platform.MSBuild.csproj
```diff
@@ -47,6 +47,7 @@ This package provides MSBuild integration of the platform, its extensions and co
     <ItemGroup>
       <TfmSpecificPackageFile Include="@(ReferenceCopyLocalPaths->WithMetadataValue('ReferenceSourceTarget', 'ProjectReference')->WithMetadataValue('PrivateAssets', 'All'))" PackagePath="lib\$(TargetFramework)\%(ReferenceCopyLocalPaths.DestinationSubDirectory)" />
       <BuildOutputInPackage Include="@(ReferenceCopyLocalPaths->WithMetadataValue('ReferenceSourceTarget', 'ProjectReference')->WithMetadataValue('CopyToBuildOutput', 'true'))" TargetPath="%(ReferenceCopyLocalPaths.DestinationSubDirectory)" />
+      <BuildOutputInPackage Include="@(RuntimeTargetsCopyLocalItems->HasMetadata('NuGetPackageId'))" TargetPath="%(RuntimeCopyLocalPaths.DestinationSubPath)" />
```
Youssef1313, line 50: Hopefully this gets us green.

It adds three extra dlls to the package:

image

@rainersigwald Is this the right way to package?

Evangelink (dismissed):

Evangelink (approved):

PR #4845, opened 2025-01-31
fix Warning IDE0052 member can be removed as the value assigned to it is never read

SimonCropp:


Evangelink (approved):

src/Platform/Microsoft.Testing.Platform/OutputDevice/TerminalOutputDevice.cs
```diff
@@ -578,7 +573,6 @@ public Task ConsumeAsync(IDataProducer dataProducer, IData value, CancellationTo
 
                 break;
             case TestRequestExecutionTimeInfo testRequestExecutionTimeInfo:
-                _testRequestExecutionTimeInfo = testRequestExecutionTimeInfo;
                 break;
```
Youssef1313: The whole case is redundant now, I think?

Then, TerminalOutputDevice is no longer doing anything with `TestRequestExecutionTimeInfo` and I guess it can be dropped from the `DataTypesConsumed` array?

Youssef1313 (approved):

PR #4846, opened 2025-01-31
fix Warning CS9113 : Parameter 'testFramework' is unread.

SimonCropp:


Youssef1313 (approved):
I don't see the warning locally, not sure why. But the change looks good to me.

Evangelink (approved):

SimonCropp:
@Youssef1313 

> I don't see the warning locally, not sure why. But the change looks good to me.

i dont either in the ide, but they do show up in the text output of the build. i didnt bother digging in further into the disparity

PR #4848, opened 2025-01-31
remove redundant IPlatformOutputDeviceManager

SimonCropp:


src/Platform/Microsoft.Testing.Platform/OutputDevice/OutputDeviceManager.cs
```diff
@@ -7,7 +7,7 @@
 
 namespace Microsoft.Testing.Platform.OutputDevice;
 
-internal sealed class PlatformOutputDeviceManager : IPlatformOutputDeviceManager
+internal sealed class PlatformOutputDeviceManager
```
Youssef1313, line 10: It was there because of potential future plans to make `IPlatformOutputDeviceManager` public and allow framework authors to plug their own output device through `SetPlatformOutputDevice`.

@Evangelink @MarcoRossignoli Are you fine dropping the interface for now, and re-introduce later if we want to expose it?

SimonCropp: @Evangelink @MarcoRossignoli bump

Evangelink: Given shift in priorities, it's probably safe to remove for now.

Evangelink (dismissed):

Youssef1313 (approved):

PR #4850, opened 2025-01-31
remove redundant GetCommandLineArguments

SimonCropp:


Evangelink (approved):

src/Platform/Microsoft.Testing.Platform/Services/ITestApplicationModuleInfo.cs
```diff
@@ -11,8 +11,6 @@ internal interface ITestApplicationModuleInfo
 
     bool IsAppHostOrSingleFileOrNativeAot { get; }
 
-    string GetCommandLineArguments();
```
Youssef1313, line 14: Is the implementation in `CurrentTestApplicationModuleInfo` unused as well? In this case could we clean it up there as well? Thanks!

Youssef1313 (approved):

PR #4852, opened 2025-01-31
remove redundant PropertyPerEdgeChars

SimonCropp:


Evangelink (commented):
Looks like a bug

src/Platform/Microsoft.Testing.Platform/Requests/TreeNodeFilter/TreeNodeFilter.cs
```diff
@@ -12,8 +12,6 @@ namespace Microsoft.Testing.Platform.Requests;
 public sealed class TreeNodeFilter : ITestExecutionFilter
 {
     public const char PathSeparator = '/';
-    internal const char PropertyPerEdgeStartChar = '[';
-    internal const char PropertyPerEdgeEndChar = ']';
```
Youssef1313, line 16: Maybe these should have been used here:

https://github.com/microsoft/testfx/blob/8ad37096aedd7dffdfc4df56b8c3e2294224778b/src/Platform/Microsoft.Testing.Platform/Requests/TreeNodeFilter/TreeNodeFilter.cs#L389-L395

@Evangelink the switch is already using string literals for everything else. So I think dropping the constants for now is fine?

SimonCropp: i agree with @Youssef1313. i dont think there is any value in extracting them out to constants

PR #4853, opened 2025-01-31
remove redundant exit code

SimonCropp:


src/Platform/Microsoft.Testing.Platform/Helpers/ExitCodes.cs
```diff
@@ -15,7 +15,6 @@ internal static class ExitCodes
     public const int TestSessionAborted = 3;
     public const int InvalidPlatformSetup = 4;
     public const int InvalidCommandLine = 5;
-    public const int FeatureNotImplemented = 6;
```
Youssef1313, line 18: It may be worth leaving a comment to not re-use the code 6 as it's already documented in https://learn.microsoft.com/en-us/dotnet/core/testing/unit-testing-platform-exit-codes.

@Evangelink @MarcoRossignoli Was this intended to be used in some cases?

Youssef1313 (approved):

src/Platform/Microsoft.Testing.Platform/Helpers/ExitCodes.cs
SimonCropp: @Evangelink @MarcoRossignoli bump

Evangelink: Thanks for the bump! I would keep it commented as suggested by @Youssef1313 as it's already public (not sure if someone relies on it).

src/Platform/Microsoft.Testing.Platform/Helpers/ExitCodes.cs
```diff
@@ -15,7 +15,6 @@ internal static class ExitCodes
     public const int TestSessionAborted = 3;
     public const int InvalidPlatformSetup = 4;
     public const int InvalidCommandLine = 5;
```
Evangelink, line 17:
```suggestion
    public const int InvalidCommandLine = 5;
    // public const int FeatureNotImplemented = 6;
```

Evangelink (approved):

PR #4855, opened 2025-01-31
remove ITool.Hidden

SimonCropp:


Youssef1313 (approved):
LGTM. @Evangelink @MarcoRossignoli Was this introduced for "potential" future usage that never happened?

Evangelink:
Yes all the tooling concept is for now unused. Given shift in priorities, I have no idea if/when this will be used so fine to cleanup.

PR #4859, opened 2025-01-31
remove redundant ServiceProviderExtensions

SimonCropp:


Evangelink (commented):
@MarcoRossignoli do you want to keep them for future or are you fine adding back when we will have the need?

SimonCropp:
IMO half finished work should be quarantined to a branch until is is decided it is really required

Evangelink (approved):

PR #4862, opened 2025-01-31
remove redundant TestDiscovered

SimonCropp:


Evangelink (commented):
@nohwnd is it dead code or is something missing?

nohwnd:
The nicer intergration for --list-tests we have in dotnet test is not used here when running single assembly. Looking if it would be useful here as well (at least it shows test count).

image

Evangelink:
Closing this one as it was a missing implementation feature.

PR #4865, opened 2025-01-31
remove SendRequestAsync

SimonCropp:


Youssef1313:
@Evangelink @MarcoRossignoli Looks like this is doc'ed in https://github.com/microsoft/testfx/blob/a1cb98d646681d58601ce69b09ea11c3a2691053/docs/mstest-runner-protocol/001-protocol-intro.md but implementation is incomplete?

SimonCropp:
@Evangelink @MarcoRossignoli  bump

Evangelink (commented):
@MarcoRossignoli @drognanar is it missing something or shall we drop it and remove this from the docs?

drognanar:
@evangelink The VS side is handling these RPCs, i.e. if you'd be debugging a test project and starting child processes, these should use the RPC calls back to VS.

What should happen on the MTP side is that:

1. When debugging a test, if the outer test runner starts an inner runner (for instance Fakes needs a child process with environment variables), this should happen via this RPC call. This was tricky to do given the existing MTP architecture and for most scenarios is not distinguishible from what we're currently doing.
2. If MTP has an equivalent of https://github.com/microsoft/vstest/blob/main/src/Microsoft.TestPlatform.ObjectModel/Adapter/Interfaces/IFrameworkHandle.cs#L30 it should be implemented via this RPC call. If we do not provide this functionality, we won't start the child processes under a debugger.

For the second one, if you'd point me to where the IFrameworkHandle is exposed in the bridge/adapter layers, we could handle it in a similar way to how we setup the output device here https://github.com/microsoft/testfx/blob/main/src/Platform/Microsoft.Testing.Platform/Hosts/ServerTestHost.cs#L476C1-L476C50.

Evangelink (changes requested):
Requesting change as we need to fix implementation and not drop it.

SimonCropp:
@Evangelink so burn this?

PR #4867, opened 2025-01-31
remove redundant ITask members

SimonCropp:


Evangelink (dismissed):

src/Platform/Microsoft.Testing.Platform/Helpers/System/ITask.cs
```diff
@@ -10,21 +10,11 @@ internal interface ITask
 
     Task Run(Action action);
 
-    Task<T> Run<T>(Func<Task<T>?> function, CancellationToken cancellationToken);
```
Youssef1313: This overload seems to be used by MSTest.Engine (closed source) through InternalsVisibleTo, which is why CI is failing.

Youssef1313 (approved):

PR #4876, opened 2025-01-31
remove WellKnownCustomTags.CompilationEnd

SimonCropp:


src/Analyzers/MSTest.Analyzers/Helpers/WellKnownCustomTags.cs
```diff
@@ -5,7 +5,6 @@ namespace MSTest.Analyzers.Helpers;
 
 internal static class WellKnownCustomTags
 {
-    public const string CompilationEnd = nameof(CompilationEnd);
```
Youssef1313, line 8: @Evangelink Thoughts? We currently don't have any compilation-end analyzers. Do you prefer keeping the infra around it in place? Or remove and remember to restore it later when needed?

Evangelink: I think it's fine to steal back from roslyn analyzers when (if) we need it.

Evangelink (approved):

Evangelink:
Looks like this is used by the diagnostic helper. You probably want to do some extra cleanup in the calling methods

PR #4881, opened 2025-01-31
remove LifecycleWithParallelAttributesTaskProjectNamePath

SimonCropp:


Evangelink (changes requested):
I need to investigate if we need to cleanup more or to add the missing tests

PR #4882, opened 2025-01-31
remove GetDisplayName

SimonCropp:


Youssef1313 (approved):
Very nice!

Thanks @SimonCropp for the cleanup. This was likely introduced during some refactorings moving things over from DynamicDataAttribute to be on the adapter then moving it back to the framework.

PR #4883, opened 2025-01-31
remove redundant null suppressions

SimonCropp:


src/Adapter/MSTest.TestAdapter/Execution/TypeCache.cs
```diff
@@ -390,7 +390,7 @@ private TestAssemblyInfo GetAssemblyInfo(Type type)
 
         assemblyInfo = new TestAssemblyInfo(assembly);
 
-        Type[] types = AssemblyEnumerator.GetTypes(assembly, assembly.FullName!, null);
```
Youssef1313, line 393: Removing this should produce a warning. `Assembly.FullName` is annotated as nullable on .NET Core.

SimonCropp:
damn. i am moving between too many codebases

PR #4884, opened 2025-01-31
Cleanup VB entrypoint

### LINKED ISSUE #4815: Improve VB generated entry point
> Unrelated to your changes, but this potentially doesn't match the behavior of C#. It may be worth updating this to `GetAwaiter().GetResult()` instead to match C#. And maybe have some test (at least manual) to verify that.

_Originally posted by @Youssef1313 in https://github.com/microsoft/testfx/pull/4814#discussion_r1932754554_

Youssef1313:
Fixes #4815

### DIFF
--- a/src/Platform/Microsoft.Testing.Platform.MSBuild/Tasks/TestingPlatformEntryPointTask.cs
+++ b/src/Platform/Microsoft.Testing.Platform.MSBuild/Tasks/TestingPlatformEntryPointTask.cs
@@ -1,4 +1,4 @@
-﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
 #pragma warning disable CS8618 // Properties below are set by MSBuild.
@@ -160,7 +160,7 @@ Namespace {{rootNamespace}}
     <System.Diagnostics.CodeAnalysis.ExcludeFromCodeCoverage>
     Module TestingPlatformEntryPoint
 
-        Function Main(args As String) As Integer
+        Function Main(args As String()) As Integer
             Return MainAsync(args).GetAwaiter().GetResult()
         End Function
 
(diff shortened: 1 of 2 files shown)


src/Platform/Microsoft.Testing.Platform.MSBuild/Tasks/TestingPlatformEntryPointTask.cs
```diff
@@ -135,11 +135,11 @@ internal sealed class TestingPlatformEntryPoint
 <System.Diagnostics.CodeAnalysis.ExcludeFromCodeCoverage>
 Module TestingPlatformEntryPoint
 
-    Function Main(args As Global.System.String()) As Global.System.Int32
-        Return MainAsync(args).Result
+    Function Main(args As String()) As Integer
```
Youssef1313, line 138: Below (line 142) we use `Integer` (language keyword), while here we used `Global.System.Int32`. I'm switching to use language keywords every where.

Evangelink (dismissed):

src/Platform/Microsoft.Testing.Platform.MSBuild/Tasks/TestingPlatformEntryPointTask.cs
```diff
@@ -135,11 +135,11 @@ internal sealed class TestingPlatformEntryPoint
...
+        Return MainAsync(args).GetAwaiter().GetResult()
```
Youssef1313, line 139: Matches C# behavior

src/Platform/Microsoft.Testing.Platform.MSBuild/Tasks/TestingPlatformEntryPointTask.cs
```diff
@@ -135,11 +135,11 @@ internal sealed class TestingPlatformEntryPoint
...
     End Function
 
-    Public Async Function MainAsync(ByVal args() As Global.System.String) As Global.System.Threading.Tasks.Task(Of Integer)
+    Public Async Function MainAsync(args As String()) As Global.System.Threading.Tasks.Task(Of Integer)
```
Youssef1313, line 142: - `ByVal` is redundant. Parameters are `ByVal` by default.
- `()` can be put either after parameter name or after type and both are exactly the same and equivalent. Few lines above we already used the style of putting it after the type, so I switched this one to match.

PR #4888, opened 2025-02-01
Disable CentralPackageTransitivePinningEnabled

Youssef1313:

### DIFF
--- a/Directory.Packages.props
+++ b/Directory.Packages.props
@@ -1,6 +1,15 @@
 <Project>
   <PropertyGroup>
     <ManagePackageVersionsCentrally>true</ManagePackageVersionsCentrally>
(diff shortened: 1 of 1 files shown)


Evangelink:
Why do you want to disable it? I found it pretty nice when we have to fix a security issue on a transitive deps. Before that we always had to add the package to the impacted projects.

Evangelink (commented):
As discussed offline, it's better to disable transitive pinning. I think it's better to keep the option and set it to false so we know this was considered and it's an explicit choice.

Ideally, adding a small comment about it.

Youssef1313:
@Evangelink Good to merge now?

Evangelink (approved):

PR #4897, opened 2025-02-02
JoinSingleLineAndShorten first cant be null

SimonCropp:


src/Platform/Microsoft.Testing.Platform.MSBuild/Tasks/FailedTestHelper.cs
```diff
@@ -77,10 +77,10 @@ internal static void FromFailedTest(this FailedTestInfoRequest failedTestInfoReq
         }
     }
 
-    private static string? JoinSingleLineAndShorten(string? first, string? second)
-        => first != null && second != null
-            ? SingleLineAndShorten(first) + " " + SingleLineAndShorten(second)
-            : SingleLineAndShorten(first) ?? SingleLineAndShorten(second);
+    private static string? JoinSingleLineAndShorten(string first, string? second)
+        => second == null
+            ? SingleLineAndShorten(first) ?? SingleLineAndShorten(second)
```
Youssef1313:
```suggestion
        => second == null
            ? SingleLineAndShorten(first)
```

src/Platform/Microsoft.Testing.Platform.MSBuild/Tasks/FailedTestHelper.cs
```diff
@@ -77,10 +77,10 @@ internal static void FromFailedTest(this FailedTestInfoRequest failedTestInfoReq
         }
     }
 
-    private static string? JoinSingleLineAndShorten(string? first, string? second)
-        => first != null && second != null
-            ? SingleLineAndShorten(first) + " " + SingleLineAndShorten(second)
-            : SingleLineAndShorten(first) ?? SingleLineAndShorten(second);
+    private static string? JoinSingleLineAndShorten(string first, string? second)
```
Youssef1313, line 80: Maybe worth an assert that first is not null?

SimonCropp: while i am happy to.... My preference is

 * for internal APIs, that are not invoked with reflection, is to rely on the compilers nullability checking.
 * for public APIs guard with arg exceptions. 
 * for internal APIs invoked with reflection, then perhaps consider debug asserts.

the reason i prefer not to debug asser in internal apis is that if u apply that approach consitently for all methods, the negative inpact on readability outwhere they rare chance that it finds an issue. Also debug assers are usually only a local dev box feature. ie in CI and in prodcution a release build is used. So if u are dev'ing on your local machine, a null ref exception with a stack trace is as good as a debug assert message.

IMO the majority of Debug asserts in this code base should be burnt.

@Evangelink @nohwnd thoughts?

SimonCropp: bump

Evangelink: If we never pass null there is no need to complexify the logic IMO.

Evangelink (approved):

PR #4900, opened 2025-02-02
GetExceptionMessage exception cant be null

SimonCropp:
and move it to a string builder

### DIFF
--- a/src/Adapter/MSTestAdapter.PlatformServices/Extensions/ExceptionExtensions.cs
+++ b/src/Adapter/MSTestAdapter.PlatformServices/Extensions/ExceptionExtensions.cs
@@ -13,22 +13,18 @@ internal static class ExceptionExtensions
     /// </summary>
     /// <param name="exception"> The exception. </param>
     /// <returns> Custom exception message that includes inner exceptions. </returns>
-    internal static string GetExceptionMessage(this Exception? exception)
+    internal static string GetExceptionMessage(this Exception exception)
     {
-        if (exception == null)
-        {
-            return string.Empty;
-        }
-
-        string exceptionString = exception.Message;
+        var builder = new StringBuilder(exception.Message);
         Exception? inner = exception.InnerException;
         while (inner != null)
         {
-            exceptionString += Environment.NewLine + inner.Message;
+            builder.AppendLine();
+            builder.AppendLine(inner.Message);
             inner = inner.InnerException;
         }
 
-        return exceptionString;
+        return builder.ToString();
     }
 
     internal static bool IsOperationCanceledExceptionFromToken(this Exception ex, CancellationToken cancellationToken)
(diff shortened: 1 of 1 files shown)


src/Adapter/MSTestAdapter.PlatformServices/Extensions/ExceptionExtensions.cs
```diff
@@ -13,22 +13,18 @@ internal static class ExceptionExtensions
...
-        }
-
-        string exceptionString = exception.Message;
+        var builder = new StringBuilder(exception.Message);
         Exception? inner = exception.InnerException;
         while (inner != null)
         {
-            exceptionString += Environment.NewLine + inner.Message;
+            builder.AppendLine();
+            builder.AppendLine(inner.Message);
```
Youssef1313: This appends an extra new line compared to the old logic.

src/Adapter/MSTestAdapter.PlatformServices/Extensions/ExceptionExtensions.cs
```diff
@@ -13,22 +13,18 @@ internal static class ExceptionExtensions
...
-    internal static string GetExceptionMessage(this Exception? exception)
+    internal static string GetExceptionMessage(this Exception exception)
     {
-        if (exception == null)
-        {
-            return string.Empty;
-        }
-
-        string exceptionString = exception.Message;
+        var builder = new StringBuilder(exception.Message);
```
Youssef1313, line 18: I'm not sure if it's worth switching to `StringBuilder`.

- This code path is (hopefully) not hot and won't be performance sensitive. So, "hopefully" it won't matter that much whether or not we are using a `StringBuilder`.
- I'm thinking that in many cases, `InnerException` may actually be null. In this case we created a `StringBuilder` for nothing.

@Evangelink What do you think?

src/Adapter/MSTestAdapter.PlatformServices/Extensions/ExceptionExtensions.cs
SimonCropp: oops. will fix

src/Adapter/MSTestAdapter.PlatformServices/Extensions/ExceptionExtensions.cs
SimonCropp: i was also on the fence on this.

Evangelink: I don't think there is much more bad in one or the other. I do think it's true that most of the time you don't have inner exceptions so the stringbuilder isn't adding a lot. But again it should be neglictible.

Youssef1313 (approved):

PR #4902, opened 2025-02-02
use range in TestMethodInfo.cs

SimonCropp:


Youssef1313:
> src\Adapter\MSTest.TestAdapter\Execution\TestMethodInfo.cs(522,50): error CS0518: (NETCORE_ENGINEERING_TELEMETRY=Build) Predefined type 'System.Range' is not defined or imported

Youssef1313:
It looks like it's failing:

> src\Adapter\MSTest.TestAdapter\Execution\TestMethodInfo.cs(522,50): error CS0518: Predefined type 'System.Range' is not defined or imported
>
> src\Adapter\MSTest.TestAdapter\Execution\TestMethodInfo.cs(522,52): error CS0518: Predefined type 'System.Index' is not defined or imported

SimonCropp:
unfortunately this would require a ref to ValueTuple package in MSTest.TestAdapter. so will close

PR #4905, opened 2025-02-03
Work towards supporting tree node filtering in MSTest

Youssef1313:


src/Platform/Microsoft.Testing.Extensions.VSTestBridge/ObjectModel/ContextAdapterBase.cs
```diff
@@ -21,16 +23,34 @@ public ContextAdapterBase(ICommandLineOptions commandLineOptions)
...
 
+    protected TreeNodeFilter? TreeNodeFilter { get; set; }
+
     // NOTE: Implementation is borrowed from VSTest
     // MSTest relies on this method existing and access it through reflection: https://github.com/microsoft/testfx/blob/main/src/Adapter/MSTest.TestAdapter/TestMethodFilter.cs#L115
     public ITestCaseFilterExpression? GetTestCaseFilter(
         IEnumerable<string>? supportedProperties,
         Func<string, TestProperty?> propertyProvider)
     {
+        if (TreeNodeFilter is not null)
```
Youssef1313, line 49: TODO: We can either say that VSTest's filter and tree node filter are not supported together at the same time, or we can create a new `CompoundTestCaseFilterExpression` which takes the two `ITestCaseFilterExpression` and does `&&` between them.

Putting this comment as a reminder, but let's first finalize something that works first then get back to this.

Youssef1313:
This is likely going to be much easier when we are able to ship MTP adapter without VSTestBridge and without VSTest object model altogether.

PR #4907, opened 2025-02-03
Use terminal logger for discovery

nohwnd:
image

Dismiss #4862


src/Platform/Microsoft.Testing.Platform/Hosts/TestHostBuilder.cs
```diff
@@ -742,36 +742,14 @@ private static async Task<ITestFramework> BuildTestFrameworkAsync(TestFrameworkB
...
-            AsynchronousMessageBus concreteMessageBusService = new(
-                dataConsumerServices,
-                serviceProvider.GetTestApplicationCancellationTokenSource(),
-                serviceProvider.GetTask(),
-                serviceProvider.GetLoggerFactory(),
-                serviceProvider.GetEnvironment());
-            await concreteMessageBusService.InitAsync();
-            testFrameworkBuilderData.MessageBusProxy.SetBuiltMessageBus(concreteMessageBusService);
-        }
+        AsynchronousMessageBus concreteMessageBusService = new(
```
nohwnd, line 747: this removes the first leg of the if. it is possible to implement all this with the custom message bus, but it means duplicating the logic that output device already has. And this makes it similar to how hotreload works.

Evangelink:
Before was
image

src/Platform/Microsoft.Testing.Platform/Hosts/TestHostBuilder.cs
Evangelink: I am not able to understand what's the impact of the change so I'll trust you and say that if that's green it's good for me.

Evangelink:
@nohwnd many acceptance tests need to be updated followng this change.

nohwnd:
> @nohwnd many acceptance tests need to be updated followng this change.

good

src/Platform/Microsoft.Testing.Platform/Hosts/TestHostBuilder.cs
nohwnd: I am also not able to tell, because I did not add the custom messagebus. But it seems to make the execution less custom, because we are using the standard message bus, in the same mode as if we were running tests.

@MarcoRossignoli  any objections removing this code?

test/IntegrationTests/MSTest.Acceptance.IntegrationTests/TestFilterTests.cs
```diff
@@ -34,8 +34,9 @@ public async Task DiscoverTestsWithFilter_UsingTestProperty_FilteredTests(string
 
         testHostResult.AssertExitCodeIs(ExitCodes.Success);
         testHostResult.AssertOutputContains("""
-The following Tests are available:
+Discovered 1 tests in assembly - .*\.dll \(net.+\|.+\)
 Test2
+Discovered 1 tests.$
 """);
```
Youssef1313: Should this be using `AssertOutputMatchesRegex`?

test/IntegrationTests/MSTest.Acceptance.IntegrationTests/TestFilterTests.cs
```diff
@@ -34,8 +34,9 @@ public async Task DiscoverTestsWithFilter_UsingTestProperty_FilteredTests(string
 
         testHostResult.AssertExitCodeIs(ExitCodes.Success);
         testHostResult.AssertOutputContains("""
-The following Tests are available:
+Discovered 1 tests in assembly - .*\.dll \(net.+\|.+\)
 Test2
+Discovered 1 tests.$
```
Youssef1313: Intended to be `\.`?

test/IntegrationTests/Microsoft.Testing.Platform.Acceptance.IntegrationTests/ExecutionTests.cs
```diff
@@ -49,8 +50,9 @@ public async Task Exec_WhenListTestsAndFilterAreSpecified_OnlyFilteredTestsAreFo
         testHostResult.AssertExitCodeIs(ExitCodes.Success);
 
         const string OutputPattern = """
-The following Tests are available:
-Test1$
+Discovered 1 tests in assembly - .*\.dll \(net.+\|.+\)
+Test1
+Discovered 1 tests.$
```
Youssef1313: Intended to be `\.`?

test/IntegrationTests/Microsoft.Testing.Platform.Acceptance.IntegrationTests/ExecutionTests.cs
```diff
@@ -19,9 +19,10 @@ public async Task Exec_WhenListTestsIsSpecified_AllTestsAreFound(string tfm)
         testHostResult.AssertExitCodeIs(ExitCodes.Success);
 
         const string OutputPattern = """
-The following Tests are available:
+Discovered 1 tests in assembly - .*\.dll \(net.+\|.+\)
 Test1
-Test2$
+Test2
+Discovered 2 tests.$
```
Youssef1313: Same here

src/Platform/Microsoft.Testing.Platform/Hosts/TestHostBuilder.cs
Youssef1313: The comment was saying:

> adding a custom message bus that will simply forward to the output display for better performance

Not sure if it's still relevant?

test/IntegrationTests/MSTest.Acceptance.IntegrationTests/TestFilterTests.cs
```diff
@@ -33,9 +33,10 @@ public async Task DiscoverTestsWithFilter_UsingTestProperty_FilteredTests(string
         TestHostResult testHostResult = await testHost.ExecuteAsync("--filter tree=one --list-tests");
 
         testHostResult.AssertExitCodeIs(ExitCodes.Success);
-        testHostResult.AssertOutputContains("""
-The following Tests are available:
+        testHostResult.AssertOutputMatchesRegex("""
+Discovered 1 tests in assembly - .*\.dll \(net.+\|.+\)
```
Evangelink: Could we update the message to have `X test(s)`?

test/IntegrationTests/MSTest.Acceptance.IntegrationTests/TestFilterTests.cs
```diff
@@ -33,9 +33,10 @@ public async Task DiscoverTestsWithFilter_UsingTestProperty_FilteredTests(string
...
 Test2
+Discovered 1 tests\.$
```
Evangelink: Is this line equivalent to the summary for the run?

test/IntegrationTests/MSTest.Acceptance.IntegrationTests/TestFilterTests.cs
nohwnd: yes, but I find it pretty noisy, when only `1 test` is the special case here and every other number is tests (at least in english). If only we had a better way of humanizing the output.

Evangelink: We could use different resource entries based on tests count :)

test/IntegrationTests/MSTest.Acceptance.IntegrationTests/TestFilterTests.cs
nohwnd: yes, why do you ask?

Evangelink: It feels a little redundant and the shape is pretty inconsistent with the run. For run, the summary does contain asm name + tfm and arch

Evangelink: And duration

Evangelink: Maybe something like `Test discovery summary: Found X tests - Asm (tfm|arch)`

nohwnd: This assumes there won't be more than 1 assemblies in the run, can I count on that? there were some multi-module run ideas, and addMStest allows multiple dll paths.

test/IntegrationTests/MSTest.Acceptance.IntegrationTests/TestFilterTests.cs
nohwnd: done as test(s)

test/IntegrationTests/MSTest.Acceptance.IntegrationTests/TestFilterTests.cs
Evangelink: Isn't it the same assumption (good or bad) for the run?

test/IntegrationTests/Microsoft.Testing.Platform.Acceptance.IntegrationTests/ExecutionTests.cs
```diff
@@ -199,22 +201,34 @@ public Task<CloseTestSessionResult> CloseTestSessionAsync(CloseTestSessionContex
 
     public async Task ExecuteRequestAsync(ExecuteRequestContext context)
     {
-        await context.MessageBus.PublishAsync(this, new TestNodeUpdateMessage(context.Request.Session.SessionUid,
-            new TestNode() { Uid = "0", DisplayName = "Test1", Properties = new(DiscoveredTestNodeStateProperty.CachedInstance) }));
+        bool isDiscovery = _sp.GetCommandLineOptions().TryGetOptionArgumentList("--list-tests", out _);
```
Youssef1313: As this changing test "input", I assume that now there is some behavior difference. It would be good to explicitly describe in PR description, and how is it going to affect known frameworks today (if affects them at all)

nohwnd: I am not sure to understand your comment. The change is removing a custom message bus that captures the --list-tests before, and instead it uses the standard bus to handle --list-tests. I don't know if that has any negative impact, but all tests are still passing, so the only change that should be happening is having a different output for the discovery.

If you think there is a different side effect then we don't have tests for it :)

nohwnd: From private chat, we now handle both events in the same handler, so if test framework previously produced both events (especially on run) the new code will break the framework because it expects discovery to be reported via discovery, and execution events to be reported  during execution but not both.

Evangelink: It seems correct to me that discover events are produced during the run. I am tempted to say discovery should produce only discovery events but maybe it would be fine to allow skipped (if you know upfront this test cannot be run) or errors if there is a misconfiguration.

Maybe what we are missing is to have the output informed about the kind of request we are into (maybe just detecting if `--list-tests` is set is enough) to know what kind of info to display.

Youssef1313: > It seems correct to me that discover events are produced during the run.

@nohwnd Does the behavior change for this case? Let's ignore the case where non-discovery events are produced during discovery. It would be good to still test the case for producing discovery + non-discovery during run.

nohwnd: handled in code, and tests fail if we revert it the if, so also tested.

nohwnd:
now failed on AbortWithCTRLPlusC_TestHost_Succeeded which is also flaky test, all my luck is gone for today.

PR #4908, opened 2025-02-03
Add tests for analyzers using metapackage and TestFramework package

Youssef1313:


Youssef1313:
Adding these tests in https://github.com/microsoft/testfx/pull/4818

PR #4921, opened 2025-02-05
VSTestBridge: support filter at discovery + runsettings TestCaseFilter

### LINKED ISSUE #4704: TestCaseFilter not working anymore with Microsoft Test Platform
Hello,

I switched recently to Xunit v3 and MS Test Platform (enabled in VS Preview Features) + Enabled via <UseMicrosoftTestingPlatformRunner>true</UseMicrosoftTestingPlatformRunner>

I'm using a .RunSettings file:
<RunSettings>
  <RunConfiguration>
    <TestCaseFilter>( Category !~ Integration )</TestCaseFilter>
  </RunConfiguration>
</RunSettings>

Before the migration to xunit v3 + MS Test Platform, this filter was taken into account into VS Test Explorer
After the migration, this filter is not taken into account anymore

My workarround is to run filter via the trait column

Image

Is there a solution to filter via a runsetting (or equivalent)?

I opened a ticket on xunit repo, but @bradwilson redirected me here https://github.com/xunit/xunit/issues/3134

Thank you

Thomas

Youssef1313:
Fixes #4704

### DIFF
--- a/src/Platform/Microsoft.Testing.Extensions.VSTestBridge/ObjectModel/ContextAdapterBase.cs
+++ b/src/Platform/Microsoft.Testing.Extensions.VSTestBridge/ObjectModel/ContextAdapterBase.cs
@@ -4,6 +4,8 @@
 using Microsoft.Testing.Extensions.VSTestBridge.CommandLine;
 using Microsoft.Testing.Platform;
 using Microsoft.Testing.Platform.CommandLine;
+using Microsoft.Testing.Platform.Extensions.Messages;
+using Microsoft.Testing.Platform.Requests;
 using Microsoft.VisualStudio.TestPlatform.ObjectModel;
 using Microsoft.VisualStudio.TestPlatform.ObjectModel.Adapter;
 
@@ -11,17 +13,27 @@ namespace Microsoft.Testing.Extensions.VSTestBridge.ObjectModel;
 
 internal abstract class ContextAdapterBase
 {
-    protected ContextAdapterBase(ICommandLineOptions commandLineOptions)
+    protected ContextAdapterBase(ICommandLineOptions commandLineOptions, IRunSettings runSettings, ITestExecutionFilter filter)
     {
+        RunSettings = runSettings;
+
+        RoslynDebug.Assert(runSettings.SettingsXml is not null);
+
+        string? filterFromRunsettings = XElement.Parse(runSettings.SettingsXml).Descendants("TestCaseFilter").SingleOrDefault()?.Value;
+        string? filterFromCommandLineOption = null;
         if (commandLineOptions.TryGetOptionArgumentList(
             TestCaseFilterCommandLineOptionsProvider.TestCaseFilterOptionName,
             out string[]? filterExpressions)
             && filterExpressions is not null
             && filterExpressions.Length == 1)
         {
-            FilterExpressionWrapper = new(filterExpressions[0]);
+            filterFromCommandLineOption = filterExpressions[0];
         }
+
+        HandleFilter(filter, filterFromRunsettings, filterFromCommandLineOption);
     }
+
+    public IRunSettings? RunSettings { get; }
 
     protected FilterExpressionWrapper? FilterExpressionWrapper { get; set; }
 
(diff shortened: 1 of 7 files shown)


src/Platform/Microsoft.Testing.Extensions.VSTestBridge/ObjectModel/ContextAdapterBase.cs
```diff
@@ -4,25 +4,37 @@
...
 internal abstract class ContextAdapterBase
 {
-    protected ContextAdapterBase(ICommandLineOptions commandLineOptions)
+    protected ContextAdapterBase(ICommandLineOptions commandLineOptions, IRunSettings runSettings, ITestExecutionFilter filter)
     {
+        RunSettings = runSettings;
+
+        RoslynDebug.Assert(runSettings.SettingsXml is not null);
+
+        string? filterFromRunsettings = XElement.Parse(runSettings.SettingsXml).Descendants("TestCaseFilter").SingleOrDefault()?.Value;
```
Evangelink: I am usually not a big fan of `Single` API as it throws an unhelpful exception. Ideally I'd prefer to throw ourselves an exception saying the `TestCaseFilter` appears more than once.

src/Platform/Microsoft.Testing.Extensions.VSTestBridge/ObjectModel/ContextAdapterBase.cs
```diff
@@ -61,4 +73,117 @@ protected ContextAdapterBase(ICommandLineOptions commandLineOptions)
 
         return adapterSpecificTestCaseFilter;
     }
+
+    private void HandleFilter(ITestExecutionFilter? filter, string? filterFromRunsettings, string? filterFromCommandLineOption)
+    {
+        // No filters at all, we can return immediately as there is nothing to do.
+        if (filter is null or NopFilter && filterFromRunsettings is null && filterFromCommandLineOption is null)
```
Evangelink: Nit: for readability (feel free to ignore)

```suggestion
        if (filter is null or NopFilter 
            && filterFromRunsettings is null 
            && filterFromCommandLineOption is null)
```

src/Platform/Microsoft.Testing.Extensions.VSTestBridge/ObjectModel/ContextAdapterBase.cs
```diff
@@ -61,4 +73,117 @@ protected ContextAdapterBase(ICommandLineOptions commandLineOptions)
...
+        }
+
+        var filterBuilder = new StringBuilder();
+
+        AppendFilter(filterFromRunsettings, filterBuilder);
+        AppendFilter(filterFromCommandLineOption, filterBuilder);
+
+        // TODO: Handle TreeNodeFilter. Note that this is unlikely to be handled here.
+        // This code path tries to construct a VSTest compatible filter.
+        // This is probably not possible for TreeNodeFilter.
```
Evangelink: I'd drop this comment. It won't help when implementing the `TreeNodeFilter` and if we don't find this line, it'll be there forever.

src/Platform/Microsoft.Testing.Extensions.VSTestBridge/ObjectModel/ContextAdapterBase.cs
```diff
@@ -61,4 +73,117 @@ protected ContextAdapterBase(ICommandLineOptions commandLineOptions)
...
+        switch (filter)
```
Evangelink: You can simplify as a simple `if` for now.

src/Platform/Microsoft.Testing.Extensions.VSTestBridge/ObjectModel/ContextAdapterBase.cs
```diff
@@ -61,4 +73,117 @@ protected ContextAdapterBase(ICommandLineOptions commandLineOptions)
...
+            default:
+                break;
+        }
+
+        if (filterBuilder.Length > 0)
+        {
+            FilterExpressionWrapper = new(filterBuilder.ToString());
+        }
+
+        static void AppendFilter(string? filter, StringBuilder builder)
```
Evangelink, line 105:
```suggestion
        // Local functions
        static void AppendFilter(string? filter, StringBuilder builder)
```

src/Platform/Microsoft.Testing.Extensions.VSTestBridge/ObjectModel/ContextAdapterBase.cs
```diff
@@ -61,4 +73,117 @@ protected ContextAdapterBase(ICommandLineOptions commandLineOptions)
...
+                            }
+
+                            filter.Append(currentChar);
+                            break;
+
+                        default:
+                            filter.Append(currentChar);
+                            break;
+                    }
+                }
```
Evangelink:
```suggestion
                filter.Append(guid.ToString());
                continue;
            }
            
            TestNodeUid currentTestNodeUid = testNodesUid[i];
            filter.Append("FullyQualifiedName=");
            for (int k = 0; k < currentTestNodeUid.Value.Length; k++)
            {
                char currentChar = currentTestNodeUid.Value[k];
                switch (currentChar)
                {
                    case '\\':
                    case '(':
                    case ')':
                    case '&':
                    case '|':
                    case '=':
                    case '!':
                    case '~':
                        // If the symbol is not escaped, add an escape character.
                        if (i - 1 < 0 || currentTestNodeUid.Value[k - 1] != '\\')
                        {
                            filter.Append('\\');
                        }

                        filter.Append(currentChar);
                        break;

                    default:
                        filter.Append(currentChar);
                        break;
                }
```

Evangelink (commented):
Link the ticket and add some tests but LGTM!

src/Platform/Microsoft.Testing.Extensions.VSTestBridge/ObjectModel/ContextAdapterBase.cs
Youssef1313: I was replicating https://github.com/microsoft/testfx/blob/ab9e1494a285006703cc3703be7f4ed4268cff23/src/Platform/Microsoft.Testing.Extensions.VSTestBridge/ObjectModel/RunContextAdapter.cs#L25

The logic is inaccurate anyways. We shouldn't be using `Descendants` altogether. Instead, we should look at direct child called `RunConfiguration`, then from there find `TestCaseFilter`.

src/Platform/Microsoft.Testing.Extensions.VSTestBridge/ObjectModel/ContextAdapterBase.cs
Youssef1313: @Evangelink I think this suggestion doesn't handle the `|` part that is done after the if/else. I slightly changed the logic so that the `|` is appended in the beginning instead of at the end so that we can do `continue;` without affecting it.

src/Platform/Microsoft.Testing.Extensions.VSTestBridge/ObjectModel/ContextAdapterBase.cs
Youssef1313: Fixed.

test/IntegrationTests/MSTest.Acceptance.IntegrationTests/TestFilterTests.cs
```diff
@@ -83,6 +83,37 @@ public async Task RunWithFilter_UsingTestPropertyForPriorityAndTestCategory_NotF
...
+    [DynamicData(nameof(TargetFrameworks.AllForDynamicData), typeof(TargetFrameworks))]
+    public async Task RunWithFilterFromRunsettings(string currentTfm)
+    {
+        var testHost = TestHost.LocateFrom(AssetFixture.TargetAssetPath, AssetName, currentTfm);
+
+        TestHostResult testHostResult = await testHost.ExecuteAsync("--settings CategoryA.runsettings");
+        testHostResult.AssertOutputContains("Running test: CategoryAOnly");
+        testHostResult.AssertOutputDoesNotContain("Running test: CategoryBOnly");
+        testHostResult.AssertOutputContains("Running test: CategoryAAndB");
+        testHostResult.AssertExitCodeIs(ExitCodes.AtLeastOneTestFailed);
```
Evangelink: I don't see a failed test, am I missing something?

Youssef1313: `PriorityTest`, `OwnerTest`, and `TestCategoryTest` are reported as failing. See the existing test `UsingTestPropertyForOwnerAndPriorityAndTestCategory_TestsFailed`

Evangelink: Oh ok! I wasn't seeing any failed assert hence the question :)

Evangelink (dismissed):

Evangelink:
Flakiness, force merging.

PR #4923, opened 2025-02-05
Obsolete PrivateType and PrivateObject

### LINKED ISSUE #4919: `PrivateType` and `PrivateObject` classes should be marked obsolete
https://github.com/microsoft/testfx/blob/main/src/TestFramework/TestFramework.Extensions/PrivateType.cs

https://github.com/microsoft/testfx/blob/main/src/TestFramework/TestFramework.Extensions/PrivateObject.cs

Youssef1313:
Fixes #4919

### DIFF
--- a/src/TestFramework/TestFramework.Extensions/PrivateObject.cs
+++ b/src/TestFramework/TestFramework.Extensions/PrivateObject.cs
@@ -8,6 +8,7 @@ namespace Microsoft.VisualStudio.TestTools.UnitTesting;
 /// <summary>
 /// This class represents the live NON public INTERNAL object in the system.
 /// </summary>
+[Obsolete(Constants.PublicTypeObsoleteMessage)]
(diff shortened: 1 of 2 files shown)


Evangelink (commented):
This means that all netfx users will now see a warning and potentially have their build failing (warn as error). It seems a bit harsh, isn't it?

Youssef1313:
@Evangelink I'm expecting these to types to be used by almost no one. Anything I'm missing? Is there a known scenario where these are used?

Evangelink:
This is a weird technic I don't particularly like that allows to test private members. This is netfx specific but this is used by quite some users (especially on large companies) so causing break there could be too much (warn as error).

Youssef1313:
Yeah. I thought this is one of those classes that are around but no one is using them 😕 
Closing then. Not sure if you want to consider a break in v4 though

PR #4924, opened 2025-02-05
MSTEST0018: Fix FP with types implementing IEnumerable

### LINKED ISSUE #4922: MSTEST0018 analyzer warning should not be displayed if the referenced "data source" member is derived from IEnumerable
## Describe the bug

https://developercommunity.visualstudio.com/t/MSTEST0018-analyzer-warning-should-not-b/10841731

### Current Situation

When using the [DynamicData](https://learn.microsoft.com/dotnet/api/microsoft.visualstudio.testtools.unittesting.dynamicdataattribute) attribute to create a data-driven unit test for MSTest, the referenced “data source” member type must be of [IEnumerable<T>](https://learn.microsoft.com/dotnet/api/system.collections.generic.ienumerable-1).

Using a data type derived from `IEnumerable<T>`, e.g., [List<T>](https://learn.microsoft.com/dotnet/api/system.collections.generic.list-1), the above mentioned analyzer warning [MSTEST0018: DynamicData should be valid](https://learn.microsoft.com/dotnet/core/testing/mstest-analyzers/mstest0018) is issued.

### Desired Situation

Don’t issue warning MSTEST0018 for data types derived from `IEnumerable<T>` when used for the referenced “data source” member.

Evangelink:
Fix #4922


test/UnitTests/MSTest.Analyzers.UnitTests/DynamicDataShouldBeValidAnalyzerTests.cs
```diff
@@ -1356,4 +1356,30 @@ await VerifyCS.VerifyAnalyzerAsync(
...
+                }
+
+                private static List<object[]> GetData()
+                {
+                    return new List<object[]>
+                    {
+                        new object[] { 1, 2 },
+                        new object[] { 3, 4 },
+                    };
+                }
```
Youssef1313, line 1578: Hypothetical scenario, what if the return type is `object` while the actual implementation returns `List`.

I think this is allowed but will be a false positive (and nothing much we can do about it).

Probably not for this PR, but maybe in future we can use two diagnostic IDs? One when we are confident that this will break something, and another when we are less confident which is info by default.

Or maybe in future we can add some dataflow analysis 👀

src/Analyzers/MSTest.Analyzers/DynamicDataShouldBeValidAnalyzer.cs
```diff
@@ -276,8 +276,7 @@ private static void AnalyzeDataSource(SymbolAnalysisContext context, AttributeDa
         // Validate member return type.
         ITypeSymbol? memberTypeSymbol = member.GetMemberType();
         if (memberTypeSymbol is INamedTypeSymbol memberNamedType
-            && (!SymbolEqualityComparer.Default.Equals(memberNamedType.ConstructedFrom, ienumerableTypeSymbol)
-                || memberNamedType.TypeArguments.Length != 1))
+            && !memberNamedType.OriginalDefinition.Inherits(ienumerableTypeSymbol))
```
Youssef1313: This is referring to `IEnumerable<T>`, but I think it should refer to `IEnumerable`?

nit: `SpecialType.System_Collections_IEnumerable` can be used.

test/UnitTests/MSTest.Analyzers.UnitTests/DynamicDataShouldBeValidAnalyzerTests.cs
```diff
@@ -1356,4 +1356,30 @@ await VerifyCS.VerifyAnalyzerAsync(
...
+            [TestClass]
+            public class MyTestClass
+            {
+                [TestMethod]
+                [DynamicData(nameof(GetData), DynamicDataSourceType.Method)]
+                public void TestMethod2(int i, int j)
+                {
+                }
+
+                private static List<object[]> GetData()
```
Youssef1313, line 1571: Consider a test for `string`. I think it needs to be special cased in the analyzer the same way it was special cased in the implementation

test/UnitTests/MSTest.Analyzers.UnitTests/DynamicDataShouldBeValidAnalyzerTests.cs
Evangelink: Added a test to show we report. We can improve later if we want.

I do like the idea of multiple IDs based on confidence but I'll consider this out of the scope of this PR.

src/Analyzers/MSTest.Analyzers/DynamicDataShouldBeValidAnalyzer.cs
```diff
@@ -275,13 +275,17 @@ private static void AnalyzeDataSource(SymbolAnalysisContext context, AttributeDa
 
         // Validate member return type.
         ITypeSymbol? memberTypeSymbol = member.GetMemberType();
-        if (memberTypeSymbol is INamedTypeSymbol memberNamedType
-            && (!SymbolEqualityComparer.Default.Equals(memberNamedType.ConstructedFrom, ienumerableTypeSymbol)
```
Youssef1313, line 279: I think the `ienumerableTypeSymbol` parameter is no longer used and can be cleaned up

src/Analyzers/MSTest.Analyzers/DynamicDataShouldBeValidAnalyzer.cs
```diff
@@ -275,13 +275,17 @@ private static void AnalyzeDataSource(SymbolAnalysisContext context, AttributeDa
...
-                || memberNamedType.TypeArguments.Length != 1))
+        if (memberTypeSymbol is not INamedTypeSymbol memberNamedType)
         {
-            context.ReportDiagnostic(attributeSyntax.CreateDiagnostic(MemberTypeRule, declaringType.Name, memberName));
             return;
```
Youssef1313, line 277: Looks like a false negative to me. Let's say the method returns a pointer type. We will return right away because that's not an `INamedTypeSymbol`.

If we fix this, we need to make sure whether `IArrayTypeSymbol` works or not. i.e, if the return type is `object[][]` instead of `IEnumerable<object[]>`

src/Analyzers/MSTest.Analyzers/DynamicDataShouldBeValidAnalyzer.cs
Evangelink: Good catch!

test/UnitTests/MSTest.Analyzers.UnitTests/DynamicDataShouldBeValidAnalyzerTests.cs
```diff
@@ -1618,6 +1618,27 @@ public void TestMethod2(char c)
...
+            [TestClass]
+            public class MyTestClass
+            {
+                [TestMethod]
+                [{|#0:DynamicData(nameof(GetData), DynamicDataSourceType.Method)|}]
+                public void TestMethod2(char c)
+                {
+                }
+
+                private static IntPtr GetData() => new IntPtr(1);
```
Youssef1313: I think this is considered as `INamedTypeSymbol` and is not testing the proper check?

In that case, the following should test the case for getting `IPointerTypeSymbol`:

```csharp
private static unsafe int* GetData() => null;
```

Youssef1313 (dismissed):

test/UnitTests/MSTest.Analyzers.UnitTests/DynamicDataShouldBeValidAnalyzerTests.cs
```diff
@@ -1356,4 +1552,93 @@ await VerifyCS.VerifyAnalyzerAsync(
...
```
Evangelink:
```suggestion
                private static unsafe int* GetData() => null;
```

Youssef1313 (approved):

PR #4942, opened 2025-02-08
Preserve TestIdGenerationStrategy in the execution AppDomain

Youssef1313:
Related to #3053, Fixes #4941

PR #4930, opened 2025-02-06
Add TestDataRow class

Youssef1313:
Fixes https://github.com/microsoft/testfx/issues/2227

PR #4950, opened 2025-02-10
Preserve TestIdGenerationStrategy in the execution AppDomain - round 2

Youssef1313:
Follow-up to #4942

Initially, we set the correct value here:

https://github.com/microsoft/testfx/blob/0f6963807ad19fdcc11a860158a1c34261879ddb/src/Adapter/MSTest.TestAdapter/Discovery/AssemblyEnumerator.cs#L82-L85

This is the discovery app domain (we create it [here](https://github.com/microsoft/testfx/blob/b9cef47d99842580a111001dc84ba4ebb14ead31/src/Adapter/MSTest.TestAdapter/Discovery/AssemblyEnumeratorWrapper.cs#L98))

Then, the code in this PR will set the `TestIdGenerationStrategy` in the main AppDomain.

After that, when we create the UnitTestRunner, the value from the main AppDomain will flow to the newly created AppDomain via the code added in #4942

It may be hard to follow, but... AppDomains.

The test added in https://github.com/microsoft/testfx/pull/4930 will make sure this doesn't regress, and I'll explicitly document the test in the other PR.

### DIFF
--- a/src/Adapter/MSTest.TestAdapter/Execution/TestAssemblySettingsProvider.cs
+++ b/src/Adapter/MSTest.TestAdapter/Execution/TestAssemblySettingsProvider.cs
@@ -30,6 +30,13 @@ internal sealed class TestAssemblySettingsProvider : MarshalByRefObject
         // Load the source.
         Assembly testAssembly = PlatformServiceProvider.Instance.FileOperations.LoadAssembly(source, isReflectionOnly: false);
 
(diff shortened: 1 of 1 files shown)


Evangelink (dismissed):

Evangelink (approved):

Youssef1313:
The description here will no longer be correct with #4952 because the code being modified here will no longer be run on the main AppDomain. I'll create a separate PR that also cleans up the code added in #4942 as it will not be needed.

PR #4951, opened 2025-02-10
Cleanup seemingly unnecessary complexity around `TestAssemblySettingsProvider`

Youssef1313:


src/Adapter/MSTest.TestAdapter/Execution/TestExecutionManager.cs
```diff
@@ -372,20 +372,7 @@ private async Task ExecuteTestsInSourceAsync(IEnumerable<TestCase> tests, IRunCo
 
         _testRunCancellationToken?.ThrowIfCancellationRequested();
 
-        TestAssemblySettingsProvider? sourceSettingsProvider = null;
-
-        try
-        {
-            sourceSettingsProvider = isolationHost.CreateInstanceForType(
```
Youssef1313, line 379: We previously were creating an instance of `TestAssemblySettingsProvider` in a separate AppDomain, but we did "nothing" with it other than a null check. Then we simply call `TestAssemblySettingsProvider.GetSettings`

Youssef1313:
Replacing with #4952

Issue #4934, opened 2025-02-06
Is TestNodeFileArtifact the way we report attachments in MTP?


bradwilson:
I'm trying to do the Microsoft.Testing.Platform equivalent to VSTest's `TestResult.Attachments` to add attachments from xUnit.net v3.

This test:

```csharp
using System.IO;
using Xunit;

namespace Empty;

public class UnitTest
{
    [Fact]
    public void Passing()
    {
        var bytes = File.ReadAllBytes("logo-128-transparent.png");

        TestContext.Current.AddAttachment("Note", "This is important information");
        TestContext.Current.AddAttachment("Logo", bytes, "image/png");
    }
}
```

Yields this output when run, which seems reasonable:

```
xUnit.net v3 Microsoft.Testing.Platform Runner v1.1.0-pre.11-dev+4fdcdbe543 (64-bit .NET 8.0.12)

  In process file artifacts produced:
    - For test 'Empty.UnitTest.Passing': C:\...\Note.txt
    - For test 'Empty.UnitTest.Passing': C:\...\Logo.png

Test run summary: Passed! - bin\Debug\net8.0\cs.dll (net8.0|x64)
  total: 1
  failed: 0
  succeeded: 1
  skipped: 0
  duration: 91ms
```

(Paths shortened for clarity)

_**Q1: Is this the correct replacement for attachments?**_

-----

Running the test in Test Explorer in MTP mode, does not yield any attachments:

![Image](https://github.com/user-attachments/assets/a31a273c-3c64-41ac-9269-00794c681fef)

Compared to running the test in Test Explorer in VSTest mode:

![Image](https://github.com/user-attachments/assets/2780a73b-d134-4ec1-85db-3d46a0f40002)

_**Q2: Am I doing something wrong? Or does Test Explorer not yet support attachments like this?**_

nohwnd:
I see attachments to be described in the protocol. but the only 2 mentions of the update messages are in the docs: 

https://grep.app/search?q=AttachmentUpdatesParams

is this implemented for vs @drognanar ?

bradwilson:
I don't see AttachmentUpdatesParams in the libraries other than the protocol definition: https://github.com/search?q=repo%3Amicrosoft%2Ftestfx%20AttachmentUpdatesParams&type=code

Evangelink:
I confirm this is also not working for MSTest.

bradwilson:
[...]

PR #4953, opened 2025-02-10
Fix ordering in server mode data consume

Evangelink:
While investigating #4934, I have noticed an ordering issue.

### DIFF
--- a/src/Platform/Microsoft.Testing.Platform/ServerMode/JsonRpc/PerRequestServerDataConsumerService.cs
+++ b/src/Platform/Microsoft.Testing.Platform/ServerMode/JsonRpc/PerRequestServerDataConsumerService.cs
@@ -216,22 +216,24 @@ internal sealed class PerRequestServerDataConsumer(IServiceProvider serviceProvi
             return;
         }
 
-        if (value is TestNodeUpdateMessage update)
+        switch (value)
         {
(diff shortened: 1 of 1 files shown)


src/Platform/Microsoft.Testing.Platform/ServerMode/JsonRpc/PerRequestServerDataConsumerService.cs
```diff
@@ -216,22 +216,24 @@ public async Task ConsumeAsync(IDataProducer dataProducer, IData value, Cancella
...
-        }
-        else if (value is TestNodeFileArtifact testNodeFileArtifact)
-        {
-            Artifacts.Add(new Artifact(testNodeFileArtifact.FileInfo.FullName, dataProducer.Uid, FileType, testNodeFileArtifact.DisplayName, testNodeFileArtifact.Description));
+            case TestNodeUpdateMessage update:
+                await ProcessTestNodeUpdateAsync(update, cancellationToken);
+                PopulateTestNodeStatistics(update);
+                break;
+
+            case TestNodeFileArtifact testNodeFileArtifact:
```
Evangelink, line 226: This inherits from `SessionFileArtifact` so we need to handle subclass before base class.

Youssef1313 (approved):

PR #4958, opened 2025-02-11
Refactor discovery to build specific info for discovery

Youssef1313:
Building "fake" options type doesn't seem like a good idea. It's simply caller responsibility to be aware that these info are not correct and cannot be relied on. When reading the code, you have to go to references few levels up to understand and be sure that the "fake" options are indeed not used.

Instead, a new internal `DiscoveryTestMethodInfo` is introduced which holds only information relevant for discovery. It also doesn't calculate things that will end up not being used for discovery.

As the new type is very lightweight, I chose to make it a struct.

### DIFF
--- a/src/Adapter/MSTest.TestAdapter/Execution/TypeCache.cs
+++ b/src/Adapter/MSTest.TestAdapter/Execution/TypeCache.cs
@@ -103,7 +103,7 @@ internal sealed class TypeCache : MarshalByRefObject
     /// Get the test method info corresponding to the parameter test Element.
     /// </summary>
     /// <returns> The <see cref="TestMethodInfo"/>. </returns>
-    public TestMethodInfo? GetTestMethodInfoForDiscovery(TestMethod testMethod)
+    public DiscoveryTestMethodInfo? GetTestMethodInfoForDiscovery(TestMethod testMethod)
     {
         Guard.NotNull(testMethod);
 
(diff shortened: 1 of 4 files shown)


src/Adapter/MSTest.TestAdapter/Execution/TypeCache.cs
```diff
@@ -694,12 +694,10 @@ private TestMethodInfo ResolveTestMethodInfo(TestMethod testMethod, TestClassInf
         return testMethodInfo;
     }
 
-    private TestMethodInfo ResolveTestMethodInfoForDiscovery(TestMethod testMethod, TestClassInfo testClassInfo)
+    private DiscoveryTestMethodInfo ResolveTestMethodInfoForDiscovery(TestMethod testMethod, TestClassInfo testClassInfo)
     {
         MethodInfo methodInfo = GetMethodInfoForTestMethod(testMethod, testClassInfo);
-
-        // Let's build a fake options type as it won't be used.
-        return new TestMethodInfo(methodInfo, testClassInfo, new(TimeoutInfo.FromTimeout(-1), null, false, null!));
```
Youssef1313, line 702: Building "fake" options type doesn't seem like a good idea. It's simply caller responsibility to be aware that these info are not correct and cannot be relied on. When reading the code, you have to go to references few levels up to understand and be sure that the "fake" options are indeed not used.

Instead, a new internal `DiscoveryTestMethodInfo` is introduced which holds only information relevant for discovery. It also doesn't calculate things that will end up not being used for discovery.

src/Adapter/MSTest.TestAdapter/MSTest.TestAdapter.csproj
```diff
@@ -63,8 +63,7 @@
 
   <ItemGroup>
     <PackageReference Include="Microsoft.TestPlatform.AdapterUtilities" />
-    <PackageReference Include="System.Threading.Tasks.Extensions"
-                      Condition=" '$(TargetFramework)' == 'netstandard2.0' OR '$(TargetFramework)' == '$(NetFrameworkMinimum)' OR '$(TargetFramework)' == '$(UwpMinimum)' " />
+    <PackageReference Include="System.Threading.Tasks.Extensions" Condition=" '$(TargetFramework)' == 'netstandard2.0' OR '$(TargetFramework)' == '$(NetFrameworkMinimum)' OR '$(TargetFramework)' == '$(UwpMinimum)' " />
```
Youssef1313, line 66: VS made this change when I added the new file. So it seems it doesn't like the existing formatting 😕

Evangelink: No worries

Evangelink (approved):

PR #4966, opened 2025-02-11
Add SourceGen + Engine

Evangelink:

PR #4971, opened 2025-02-11
Follow-up to the move of SourceGen + Engine

Youssef1313:
Follow-up to #4966

### DIFF
--- a/src/Adapter/MSTest.Engine/MSTest.Engine.csproj
+++ b/src/Adapter/MSTest.Engine/MSTest.Engine.csproj
@@ -7,13 +7,8 @@
     <PackageLicenseExpressionInternal></PackageLicenseExpressionInternal>
     <PackageLicenseExpression></PackageLicenseExpression>
     <PackageLicenseFile>License.txt</PackageLicenseFile>
-    <!--
-      This is a early preview package, keep 1.0.0-alpha or similar suffix even in official builds.
-    -->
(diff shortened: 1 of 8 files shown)


src/Adapter/MSTest.Engine/MSTest.Engine.csproj
```diff
@@ -5,15 +5,19 @@
...
     <!-- Cancel default license (closed-source) and set license as packaged license file -->
     <PackageLicenseExpressionInternal></PackageLicenseExpressionInternal>
+    <PackageLicenseExpression></PackageLicenseExpression>
     <PackageLicenseFile>License.txt</PackageLicenseFile>
     <!--
       This is a early preview package, keep 1.0.0-alpha or similar suffix even in official builds.
     -->
+    <!-- Whenever VersionPrefix or PreReleaseVersionLabel change, MSTest.Sdk need to be in sync. -->
+    <!-- See GenerateTemplates target in MSTest.Sdk.targets -->
     <VersionPrefix>1.0.0</VersionPrefix>
```
Evangelink: Can we create a property on the main `Directory.Build.props`, like `MSTestEngineVersionPrefix` reuse it here and in the SDK csproj. This would reduce the risk of having them out of sync.

Evangelink (dismissed):

src/Adapter/MSTest.Engine/MSTest.Engine.csproj
Youssef1313: Done, and improved the logic

Issue #5007, opened 2025-02-13
MSTEST0038: Use 'AreNotSame' instead of 'AreNotEqual'


Evangelink:
## Describe the bug

Message is reversed, we should recommend to use `AreNotEqual`.

PR #5008, opened 2025-02-13
Fix MSTEST0038 message

Youssef1313:
Fixes https://github.com/microsoft/testfx/issues/5007

### DIFF
--- a/test/UnitTests/MSTest.Analyzers.UnitTests/AvoidAssertAreSameWithValueTypesAnalyzerTests.cs
+++ b/test/UnitTests/MSTest.Analyzers.UnitTests/AvoidAssertAreSameWithValueTypesAnalyzerTests.cs
@@ -189,4 +189,51 @@ public sealed class AvoidAssertAreSameWithValueTypesAnalyzerTests
 
         await VerifyCS.VerifyCodeFixAsync(code, fixedCode);
     }
+
+    [TestMethod]
+    public async Task AnalyzerMessageShouldBeCorrect()
+    {
+        string code = """
+            using Microsoft.VisualStudio.TestTools.UnitTesting;
+            using System.Collections.Generic;
+
+            [TestClass]
+            public class MyTestClass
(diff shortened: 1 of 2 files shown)


Evangelink (dismissed):

test/UnitTests/MSTest.Analyzers.UnitTests/AvoidAssertAreSameWithValueTypesAnalyzerTests.cs
```diff
@@ -189,4 +189,51 @@ public void TestMethod()
...
+            """;
+
+        await new VerifyCS.Test()
+        {
+            TestCode = code,
+            FixedCode = fixedCode,
+            ExpectedDiagnostics =
+            {
+                VerifyCS.Diagnostic().WithLocation(0).WithMessage("Use 'AreEqual' instead of 'AreSame' when comparing value types"),
+                VerifyCS.Diagnostic().WithLocation(1).WithMessage("Use 'AreNotEqual' instead of 'AreNotSame' when comparing value types"),
```
Evangelink: Let's use `WithArguments` instead of `WithMessage` so that the test still works when using different culture.

Youssef1313:
```suggestion
                VerifyCS.Diagnostic().WithLocation(0).WithArguments("AreEqual", "AreSame"),
                VerifyCS.Diagnostic().WithLocation(1).WithArguments("AreEqual", "AreSame"),
```

Youssef1313: It makes it less clear to see from the test that the message is correct, but I agree that it's better to make the test run under different culture.

Evangelink (approved):

PR #5009, opened 2025-02-13
Use ReferenceTrimmer

stan-sz:


Evangelink:
TODO: Review number of tests + content of nupkg

FYI @Youssef1313

test/UnitTests/Microsoft.Testing.Platform.UnitTests/Program.cs
```diff
@@ -19,8 +19,10 @@
 #if ENABLE_CODECOVERAGE
 builder.AddCodeCoverageProvider();
 #endif
-builder.AddCrashDumpProvider();
+builder.AddAppInsightsTelemetryProvider();
+builder.AddCrashDumpProvider(ignoreIfNotSupported: true);
```
Youssef1313, line 23: I'm curious why ignoreIfNotSupported was not needed previously

test/UnitTests/Microsoft.Testing.Platform.UnitTests/Program.cs
```diff
@@ -19,8 +19,10 @@
...
 builder.AddHangDumpProvider();
+builder.AddRetryProvider();
```
Youssef1313, line 25: @Evangelink Do we want to use re-try for unit tests? I'm afraid this could hide race conditions and flaky tests that indicate real issues

test/UnitTests/Microsoft.Testing.Platform.UnitTests/Program.cs
```diff
@@ -19,8 +19,10 @@
 #if ENABLE_CODECOVERAGE
 builder.AddCodeCoverageProvider();
 #endif
-builder.AddCrashDumpProvider();
+builder.AddAppInsightsTelemetryProvider();
```
Youssef1313, line 22: @Evangelink Any reason telemetry wasn't enabled previously? Is it intentional for our tests to opt-out for telemetry? It's already enabled for two other test projects

test/UnitTests/Microsoft.Testing.Extensions.VSTestBridge.UnitTests/Program.cs
```diff
@@ -20,14 +20,15 @@
 #if ENABLE_CODECOVERAGE
 builder.AddCodeCoverageProvider();
 #endif
+Console.WriteLine("NATIVE_AOT disabled");
```
Youssef1313, line 23: Intentional?

src/Platform/Microsoft.Testing.Extensions.VSTestBridge/Microsoft.Testing.Extensions.VSTestBridge.csproj
```diff
@@ -17,7 +17,7 @@
   <ItemGroup>
     <ProjectReference Include="$(RepoRoot)src\Platform\Microsoft.Testing.Platform\Microsoft.Testing.Platform.csproj" />
     <ProjectReference Include="$(RepoRoot)src\Platform\Microsoft.Testing.Extensions.TrxReport.Abstractions\Microsoft.Testing.Extensions.TrxReport.Abstractions.csproj" />
-    <ProjectReference Include="$(RepoRoot)src\Platform\Microsoft.Testing.Extensions.Telemetry\Microsoft.Testing.Extensions.Telemetry.csproj" />
+    <ProjectReference Include="$(RepoRoot)src\Platform\Microsoft.Testing.Extensions.Telemetry\Microsoft.Testing.Extensions.Telemetry.csproj" ReferenceOutputAssembly="false" />
```
Youssef1313, line 21: Does this affect the publicly visible dependencies of the VSTestBridge NuGet package?

src/Platform/Microsoft.Testing.Extensions.CrashDump/Microsoft.Testing.Extensions.CrashDump.csproj
```diff
@@ -37,7 +37,6 @@ This package extends Microsoft Testing Platform to provide a crash dump function
 
   <ItemGroup>
     <ProjectReference Include="$(RepoRoot)src\Platform\Microsoft.Testing.Platform\Microsoft.Testing.Platform.csproj" />
-    <ProjectReference Include="$(RepoRoot)src\Platform\Microsoft.Testing.Extensions.TrxReport.Abstractions\Microsoft.Testing.Extensions.TrxReport.Abstractions.csproj" />
```
Youssef1313, line 40: Same here. Does this affect the publicly visible dependencies of CrashDump?

test/UnitTests/Microsoft.Testing.Platform.UnitTests/Program.cs
nohwnd: Probably because crash dump is passive (unlike hang dump), so what happens is that on activation you set env variables and hope that some crash dumps will appear in the dump folder (those will be created by the built-in .net crash reporter). This setup won't fail in native aot, and we probably don't test that it actually works there because we don't expect it to work?

If the process was more active, like in VSTest where we use procdump for .net framework processes, then even the setup would probably fail in native aot, or other incompatible scenario, and we would see it fail early, even without explicitly testing for it.

nohwnd:
Replacing with https://github.com/microsoft/testfx/pull/5549, to see how far it is from being done. Picked the commits from the start of the PR to avoid stealing code.

PR #5014, opened 2025-02-13
Convert internal framework to MTP

Evangelink:
Move our internal framework to be MTP based so we can start dogfooding the new `dotnet test` experience.


Directory.Build.props
```diff
@@ -13,6 +13,7 @@
     <LangVersion>preview</LangVersion>
     <ImplicitUsings>enable</ImplicitUsings>
     <GenerateDocumentationFile>true</GenerateDocumentationFile>
+    <Nullable>enable</Nullable>
```
Evangelink, line 16: Enabled globally, cleaning up the various files.

src/Platform/Microsoft.Testing.Platform/Microsoft.Testing.Platform.csproj
```diff
@@ -56,6 +56,8 @@ This package provides the core platform and the .NET implementation of the proto
     <InternalsVisibleTo Include="MSTest.Acceptance.IntegrationTests" Key="$(VsPublicKey)" />
     <InternalsVisibleTo Include="MSTest.Engine" Key="$(VsPublicKey)" />
     <InternalsVisibleTo Include="MSTest.Engine.UnitTests" Key="$(VsPublicKey)" />
+    <InternalsVisibleTo Include="TestFramework.ForTestingMSTest" Key="$(VsPublicKey)" />
+    <InternalsVisibleTo Include="MSTest.IntegrationTests" Key="$(VsPublicKey)" />
```
Evangelink, line 61: Needed for polyfill

test/Utilities/TestFramework.ForTestingMSTest/TestFramework.ForTestingMSTest.csproj
```diff
@@ -1,20 +1,16 @@
-<Project Sdk="Microsoft.NET.Sdk">
+﻿<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFramework>netstandard2.0</TargetFramework>
+    <TargetFrameworks>netstandard2.0;$(MicrosoftTestingTargetFrameworks)</TargetFrameworks>
```
Youssef1313, line 4: Why is `netstandard2.0` not enough?

Evangelink: It was for polyfill but I think I got rid of the deps there so I would need to check if that's enough with netstandard

Evangelink: Keeping it so that we can benefit from platform optimizations

PR #5020, opened 2025-02-14
Fix parameterized test treated as ignored when using VSTest in Test E…

### LINKED ISSUE #5018: DynamicDataAttribute and DataRowAttribute annotated test methods are being skipped in v3.8
## Describe the bug

After updating MSTest to version 3.8.0 data-driven tests are being skipped. Update from 3.7.3 to 3.8.0 shouldn't bring any breaking changes.

## Steps To Reproduce

1. Create a MSTest project. Update MSTest to v3.7.3.
2.  Create a data-driven test using DynamicDataAttribute or DataRowAttribute.
3. Run tests. None should be skipped.
4. Update MSTest to 3.8.0
5. Run tests. Tests using DynamicDataAttribute or DataRowAttribute are being skipped.

Working example can be found in repo [polyline-algorithm-csharp@repro/mtstest-breaking-change-v3.8](https://github.com/petesramek/polyline-algorithm-csharp/tree/repro/mtstest-breaking-change-v3.8) and tests folder. There is side by side the same test project, one using 3.7.* and the other 3.8.*.

## Expected behavior

Test methods using DynamicDataAttribute or DataRowAttribute are not being skipped after an update to 3.8.0.

## Actual behavior

Test methods using DynamicDataAttribute or DataRowAttribute are being skipped after an update to 3.8.0.

## Additional context

Image

Youssef1313:
Issue is only on TE. No good way to write a test. The fix is manually verified using locally-built packages.

Fixes #5018

### DIFF
--- a/src/Adapter/MSTest.TestAdapter/Execution/TestMethodRunner.cs
+++ b/src/Adapter/MSTest.TestAdapter/Execution/TestMethodRunner.cs
@@ -170,7 +170,7 @@ internal sealed class TestMethodRunner
         var parentStopwatch = Stopwatch.StartNew();
         if (_test.DataType == DynamicDataType.ITestDataSource)
         {
-            if (_test.TestDataSourceIgnoreMessage is not null)
+            if (!string.IsNullOrEmpty(_test.TestDataSourceIgnoreMessage))
             {
                 _testContext.SetOutcome(UTF.UnitTestOutcome.Ignored);
                 return [TestResult.CreateIgnoredResult(_test.TestDataSourceIgnoreMessage)];
@@ -269,7 +269,7 @@ internal sealed class TestMethodRunner
 
         foreach (UTF.ITestDataSource testDataSource in testDataSources)
         {
-            if (testDataSource is ITestDataSourceIgnoreCapability { IgnoreMessage: { } ignoreMessage })
+            if (testDataSource is ITestDataSourceIgnoreCapability { IgnoreMessage: { Length: > 0 } ignoreMessage })
             {
                 results.Add(TestResult.CreateIgnoredResult(ignoreMessage));
                 continue;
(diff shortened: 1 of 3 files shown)


src/Adapter/MSTest.TestAdapter/Execution/TestMethodRunner.cs
```diff
@@ -170,7 +170,7 @@ internal TestResult[] RunTestMethod()
         var parentStopwatch = Stopwatch.StartNew();
         if (_test.DataType == DynamicDataType.ITestDataSource)
         {
-            if (_test.TestDataSourceIgnoreMessage is not null)
+            if (!string.IsNullOrEmpty(_test.TestDataSourceIgnoreMessage))
```
Evangelink, line 173: Shall we also exclude full whitespace?

src/Adapter/MSTest.TestAdapter/Execution/TestMethodRunner.cs
```diff
@@ -269,7 +269,7 @@ private bool TryExecuteFoldedDataDrivenTests(List<TestResult> results)
 
         foreach (UTF.ITestDataSource testDataSource in testDataSources)
         {
-            if (testDataSource is ITestDataSourceIgnoreCapability { IgnoreMessage: { } ignoreMessage })
+            if (testDataSource is ITestDataSourceIgnoreCapability { IgnoreMessage: { Length: > 0 } ignoreMessage })
```
Evangelink, line 272: Shall we consider whitespace like previous comment?

src/Adapter/MSTest.TestAdapter/Execution/TestMethodRunner.cs
Youssef1313: @Evangelink I thought about it, but I think if the user explicitly sets ignore message to `"  "`, then he wants to ignore. It's unlikely someone will do it though.

I actually think that the check for empty is a workaround for VSTest bug and I think it should be allowed to set to empty string and have the test ignored.

Evangelink: @nohwnd are you aware of VSTest serialization not handling nulls for string?

Evangelink:
Let's merge as-is but be prepared to do a different fix based on @nohwnd input for VSTest behavior.

Evangelink (approved):

PR #5021, opened 2025-02-14
Try out props/targets refactoring

### LINKED ISSUE #5016: [Regression] Setting `IsTestingPlatformApplication` to false no longer prevents the entry-point generation
Scenario:

Two projects are involved:

1. A typical test project using is a testing platform application, and has generated entry-point.
2. Another executable, but is not a test project, and csproj sets `IsTestingPlatformApplication` to false. This executable is referencing the test project

Because we changed MSTest.TestAdapter to be transitive, its targets file is now imported in the second project. So:

https://github.com/microsoft/testfx/blob/1461958ee68d012061f024d3c2c51f6d62065965/src/Adapter/MSTest.TestAdapter/buildTransitive/common/MSTest.TestAdapter.targets#L5

Youssef1313:
Fixes #5016 
The general idea here is:

1. MSBuild extension props file is imported first. We do nothing
2. Then TestAdapter props. We do nothing.
3. csproj is imported, and the user can set `IsTestingPlatformApplication` and/or `EnableMSTestRunner`.
4. MSBuild extension targets are imported. At this point, we default `IsTestingPlatformApplication` to true if it wasn't set from csproj or from MSTest.Sdk. But we don't yet decide whether or not we can generate anything.
5. TestAdapter targets is imported, now we default `EnableMSTestRunner` to false if not already set. And we overwrite `IsTestingPlatformApplication` to use the same value as EnableMSTestRunner
6. In MSBuild extension `Target`s (referring to really "Target", not `.targets` file), this is only when we can reliable use the value of `IsTestingPlatformApplication`.

The core thing here is that, `IsTestingPlatformApplication` value should stabilize during evaluation, and it stabilizes late during evaluation (in TestAdapter targets, after MSBuild extension targets).
So, the value can only be reliably read inside of a `Target`, and we allow it to be set very late during evaluation. After evaluation, it shouldn't be set by any one.

### DIFF
--- a/src/Platform/Microsoft.Testing.Platform.MSBuild/buildMultiTargeting/Microsoft.Testing.Platform.MSBuild.props
+++ b/src/Platform/Microsoft.Testing.Platform.MSBuild/buildMultiTargeting/Microsoft.Testing.Platform.MSBuild.props
@@ -1,10 +1,5 @@
 ﻿<?xml version="1.0" encoding="utf-8" ?>
 <Project>
-  <PropertyGroup>
-    <!-- This is the knob to disable completely for this project the machinery -->
-    <IsTestingPlatformApplication Condition=" '$(IsTestingPlatformApplication)' == '' ">True</IsTestingPlatformApplication>
-  </PropertyGroup>
-
   <!-- Register the PlatformOutputDevice extension -->
   <ItemGroup>
     <TestingPlatformBuilderHook Include="8214564B-6107-422D-8A82-462F97602287" >
--- a/src/Platform/Microsoft.Testing.Platform.MSBuild/buildMultiTargeting/Microsoft.Testing.Platform.MSBuild.targets
+++ b/src/Platform/Microsoft.Testing.Platform.MSBuild/buildMultiTargeting/Microsoft.Testing.Platform.MSBuild.targets
@@ -153,7 +153,7 @@
   </Target>
   <Target Name="_IncludeGenerateTestingPlatformEntryPointIntoCompilation" BeforeTargets="BeforeCompile;XamlPreCompile"
           DependsOnTargets="_CalculateGenerateTestingPlatformEntryPoint;_GenerateTestingPlatformEntryPointFileInputCache;_GenerateTestingPlatformEntryPoint"
-          Condition=" '$(GenerateTestingPlatformEntryPoint)' == 'True' " />
+          Condition=" '$(GenerateTestingPlatformEntryPoint)' == 'True' ">
 
     <ItemGroup>
       <Compile Include="$(_TestingPlatformEntryPointSourcePath)" />
--- a/src/Platform/Microsoft.Testing.Platform.MSBuild/buildMultiTargeting/Microsoft.Testing.Platform.MSBuild.targets
+++ b/src/Platform/Microsoft.Testing.Platform.MSBuild/buildMultiTargeting/Microsoft.Testing.Platform.MSBuild.targets
@@ -7,6 +7,9 @@
   </PropertyGroup>
 
   <!-- Declare our capability to Visual Studio/Visual Studio Code -->
+  <!-- Note: DisableTestingPlatformServerCapability may be set by TestAdapter targets -->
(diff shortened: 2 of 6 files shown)


src/Platform/Microsoft.Testing.Platform.MSBuild/buildMultiTargeting/Microsoft.Testing.Platform.MSBuild.targets
```diff
@@ -48,6 +46,16 @@
...
   </PropertyGroup>
+
+  <Target Name="_CalculateGenerateTestingPlatformConfigurationFile" BeforeTargets="CopyFilesToOutputDirectory" Condition="'$(GenerateTestingPlatformConfigurationFile)' == 'true'">
+    <PropertyGroup>
+      <!-- We add this to a target to allow setting IsTestingPlatformApplication as late as possible -->
+      <!-- For example, MSTest.TestAdapter.targets is imported *after* Microsoft.Testing.Platform.MSBuild.targets -->
+      <!-- We want to allow MSTest.TestAdapter.targets to set IsTestingPlatformApplication to false if EnableMSTestRunner is not set to true. -->
+      <GenerateTestingPlatformConfigurationFile Condition=" '$(GenerateTestingPlatformConfigurationFile)' == '' " >$(IsTestingPlatformApplication)</GenerateTestingPlatformConfigurationFile>
+    </PropertyGroup>
+  </Target>
```
Youssef1313: Not sure if there is a better way here. Applicable to the new 2 targets as well.

Youssef1313:
What I want to do here is probably too big for 3.8.1. I'll go with a simple approach for 3.8.1, and have the work here for 3.9.0

src/Platform/Microsoft.Testing.Platform.MSBuild/buildMultiTargeting/Microsoft.Testing.Platform.MSBuild.targets
Youssef1313: Improved a bit what I had previously

src/Platform/Microsoft.Testing.Platform.MSBuild/buildMultiTargeting/Microsoft.Testing.Platform.MSBuild.props
```diff
@@ -1,10 +1,5 @@
 ﻿<?xml version="1.0" encoding="utf-8" ?>
 <Project>
-  <PropertyGroup>
-    <!-- This is the knob to disable completely for this project the machinery -->
-    <IsTestingPlatformApplication Condition=" '$(IsTestingPlatformApplication)' == '' ">True</IsTestingPlatformApplication>
```
Evangelink, line 5: The idea of setting it that early was to say that if you use `MTP MSBuild` you are a test app and you could rely on this property to make your actions (similar to `IsTestProject`).

Youssef1313: This doesn't look correct to me.
Is `IsTestingPlatformApplication` intended to be read by us only, and set by anyone? In that case, we know we read it late enough, and we can set the default late enough.

If it's intended to be read by others, then there is a design flaw here. It should only be read inside of a `Target` to allow it to be set as late as possible

Evangelink: To me the idea is similar to `IsTestProject` except this is MTP

Youssef1313: It's more complicated for the case of MTP. Let's discuss offline

Evangelink: Discussed offline and we can no longer do that because now the package is supporting both VSTest and MTP.

Evangelink (approved):

PR #5032, opened 2025-02-15
Use `global::` prefix for generated code

### LINKED ISSUE #5031: The type or namespace name 'VisualStudio' does not exist in the namespace 'Something.Microsoft' (are you missing an assembly reference?)
## Steps To Reproduce

Update at https://github.com/Lumoin/Verifiable/blob/main/global.json#L7 `"MSTest.Sdk": "3.7.3" ` -> `"MSTest.Sdk": "3.8.0"`

## Expected behavior

Project compiles and tests can be run.

## Actual behavior

An error message like in the header from one project (Verifiable.Microsoft).

Youssef1313:
Fixes #5031


Evangelink (commented):
Could you please add some acceptance test for this scenario so we avoid regressing in the future?

Evangelink (approved):

PR #5037, opened 2025-02-17
Refactor logic around GetCurrentTestApplicationFullPath

Youssef1313:

### DIFF
--- a/src/Platform/Microsoft.Testing.Platform/Builder/TestApplication.cs
+++ b/src/Platform/Microsoft.Testing.Platform/Builder/TestApplication.cs
@@ -179,7 +179,7 @@ public sealed class TestApplication : ITestApplication
 #endif
         await logger.LogInformationAsync($"IsDynamicCodeSupported: {isDynamicCodeSupported}");
 
-        string moduleName = testApplicationModuleInfo.GetCurrentTestApplicationFullPath();
+        string? moduleName = testApplicationModuleInfo.TryGetCurrentTestApplicationFullPath();
         await logger.LogInformationAsync($"Test module: {moduleName}");
         await logger.LogInformationAsync($"Command line arguments: '{(args.Length == 0 ? string.Empty : args.Aggregate((a, b) => $"{a} {b}"))}'");
 
@@ -311,7 +311,7 @@ public sealed class TestApplication : ITestApplication
         }
 
         // Set the directory to the default test result directory
-        string directory = Path.Combine(Path.GetDirectoryName(testApplicationModuleInfo.GetCurrentTestApplicationFullPath())!, AggregatedConfiguration.DefaultTestResultFolderName);
+        string directory = Path.Combine(testApplicationModuleInfo.GetCurrentTestApplicationDirectory(), AggregatedConfiguration.DefaultTestResultFolderName);
         bool customDirectory = false;
 
         if (result.TryGetOptionArgumentList(PlatformCommandLineProvider.ResultDirectoryOptionKey, out string[]? resultDirectoryArg))
--- a/src/Platform/Microsoft.Testing.Platform/Builder/TestApplication.cs
+++ b/src/Platform/Microsoft.Testing.Platform/Builder/TestApplication.cs
@@ -179,7 +179,7 @@ public sealed class TestApplication : ITestApplication
 #endif
         await logger.LogInformationAsync($"IsDynamicCodeSupported: {isDynamicCodeSupported}");
 
-        string? moduleName = testApplicationModuleInfo.TryGetCurrentTestApplicationFullPath();
+        string moduleName = testApplicationModuleInfo.TryGetCurrentTestApplicationFullPath() ?? testApplicationModuleInfo.TryGetAssemblyName() ?? "unknown";
         await logger.LogInformationAsync($"Test module: {moduleName}");
         await logger.LogInformationAsync($"Command line arguments: '{(args.Length == 0 ? string.Empty : args.Aggregate((a, b) => $"{a} {b}"))}'");
 
--- a/src/Platform/Microsoft.Testing.Platform/Builder/TestApplication.cs
+++ b/src/Platform/Microsoft.Testing.Platform/Builder/TestApplication.cs
@@ -179,7 +179,7 @@ public sealed class TestApplication : ITestApplication
 #endif
         await logger.LogInformationAsync($"IsDynamicCodeSupported: {isDynamicCodeSupported}");
 
-        string moduleName = testApplicationModuleInfo.TryGetCurrentTestApplicationFullPath() ?? testApplicationModuleInfo.TryGetAssemblyName() ?? "unknown";
+        string moduleName = testApplicationModuleInfo.GetDisplayName();
         await logger.LogInformationAsync($"Test module: {moduleName}");
         await logger.LogInformationAsync($"Command line arguments: '{(args.Length == 0 ? string.Empty : args.Aggregate((a, b) => $"{a} {b}"))}'");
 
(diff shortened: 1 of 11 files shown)


src/Platform/Microsoft.Testing.Platform/Services/CurrentTestApplicationModuleInfo.cs
```diff
@@ -46,24 +46,38 @@ public bool IsAppHostOrSingleFileOrNativeAot
...
         string? moduleName = Assembly.GetEntryAssembly()?.Location;
-#pragma warning restore IL3000
 
-        moduleName = RoslynString.IsNullOrEmpty(moduleName)
+        return RoslynString.IsNullOrEmpty(moduleName)
             ? GetProcessPath(_environment, _process)
             : moduleName;
+    }
 
-        ApplicationStateGuard.Ensure(moduleName is not null);
```
Youssef1313, line 63: This guard isn't correct. For example, in case of Wasm (which is obviously not supported currently), this guard fails.

Some callers don't really require non-null value. So, we now have `TryGetCurrentTestApplicationFullPath` (return is nullable) and `GetCurrentTestApplicationFullPath` (return is not nullable).

src/Platform/Microsoft.Testing.Platform/Builder/TestApplication.cs
```diff
@@ -179,7 +179,7 @@ private static async Task LogInformationAsync(
 #endif
         await logger.LogInformationAsync($"IsDynamicCodeSupported: {isDynamicCodeSupported}");
 
-        string moduleName = testApplicationModuleInfo.GetCurrentTestApplicationFullPath();
+        string moduleName = testApplicationModuleInfo.TryGetCurrentTestApplicationFullPath() ?? testApplicationModuleInfo.TryGetAssemblyName() ?? "unknown";
```
Youssef1313: This is used only for logging. So, we can tolerate null (e.g, in future when Wasm is supported). Then, we can fallback to just assembly name if we can find it, or otherwise just log it as "unknown"

src/Platform/Microsoft.Testing.Platform/Builder/TestApplication.cs
```diff
@@ -311,7 +311,7 @@ private static ApplicationLoggingState CreateFileLoggerIfDiagnosticIsEnabled(
         }
 
         // Set the directory to the default test result directory
-        string directory = Path.Combine(Path.GetDirectoryName(testApplicationModuleInfo.GetCurrentTestApplicationFullPath())!, AggregatedConfiguration.DefaultTestResultFolderName);
+        string directory = Path.Combine(testApplicationModuleInfo.GetCurrentTestApplicationDirectory(), AggregatedConfiguration.DefaultTestResultFolderName);
```
Youssef1313, line 314: Refactoring the pattern of `Path.GetDirectoryName(GetCurrentTestApplicationFullPath())` to be using a new method `GetCurrentTestApplicationDirectory`. This is equivalent to the original logic, except that we don't throw on null path and fallback to `AppContext.BaseDirectory` instead.

src/Platform/Microsoft.Testing.Platform/CommandLine/CommandLineHandler.cs
```diff
@@ -102,7 +102,7 @@ async Task DisplayPlatformInfoAsync()
             await outputDevice.DisplayAsync(this, new TextOutputDeviceData($"  Runtime location: {runtimeLocation}"));
 #endif
 
-            string moduleName = _testApplicationModuleInfo.GetCurrentTestApplicationFullPath();
+            string moduleName = _testApplicationModuleInfo.TryGetCurrentTestApplicationFullPath() ?? _testApplicationModuleInfo.TryGetAssemblyName() ?? "unknown";
```
Youssef1313: Same as the comment in `TestApplication.cs`. This is only for display, so we can tolerate null path and fallback to assembly name, or finally fallback to "unknown"

src/Platform/Microsoft.Testing.Platform/Configurations/AggregatedConfiguration.cs
```diff
@@ -61,7 +61,7 @@ public void SetTestHostWorkingDirectory(string workingDirectory) =>
     public async Task CheckTestResultsDirectoryOverrideAndCreateItAsync(ICommandLineOptions commandLineOptions, IFileLoggerProvider? fileLoggerProvider)
     {
         // Load Configuration
-        _currentWorkingDirectory = Path.GetDirectoryName(_testApplicationModuleInfo.GetCurrentTestApplicationFullPath())!;
+        _currentWorkingDirectory = _testApplicationModuleInfo.GetCurrentTestApplicationDirectory();
```
Youssef1313, line 64: Similar change as pointed out before. The implementation of `GetCurrentTestApplicationDirectory` is equivalent to the original code, except that we tolerate null and use `AppContext.BaseDirectory`.

src/Platform/Microsoft.Testing.Platform/OutputDevice/TerminalOutputDevice.cs
```diff
@@ -93,7 +93,7 @@ public TerminalOutputDevice(
             _targetFramework = TargetFrameworkParser.GetShortTargetFramework(Assembly.GetEntryAssembly()?.GetCustomAttribute<TargetFrameworkAttribute>()?.FrameworkDisplayName) ?? _runtimeFramework;
         }
 
-        _assemblyName = testApplicationModuleInfo.GetCurrentTestApplicationFullPath();
+        _assemblyName = testApplicationModuleInfo.TryGetCurrentTestApplicationFullPath() ?? testApplicationModuleInfo.TryGetAssemblyName() ?? "unknown assembly name";
```
Youssef1313: This is used for display only. So, if we don't have a full path, we can try assembly name, or fallback to unknown assembly

src/Platform/Microsoft.Testing.Platform/Builder/TestApplication.cs
Evangelink: Maybe `<unknown-assembly>`

src/Platform/Microsoft.Testing.Platform/CommandLine/CommandLineHandler.cs
Evangelink: Same comment about `unknown` + maybe let's introduce a constant so we have all places matching.

Evangelink: Let's make some test (manual is fine) with and without `UseAppHost` to confirm we are not impacted by https://github.com/dotnet/runtime/issues/3704

Youssef1313: `AppContext.BaseDirectory` is used by `GetCurrentTestApplicationDirectory` not `TryGetCurrentTestApplicationFullPath`. Is this comment on the wrong line?

I'm not sure how we can reach the code to use AppContext.BaseDirectory, because in normal scenarios we are currently supporting, `TryGetCurrentTestApplicationFullPath` already works fine.
The code for AppContext.BaseDirectory will be reachable if we support Wasm in future. So not sure how to test.

src/Platform/Microsoft.Testing.Platform/Builder/TestApplication.cs
Youssef1313: Done

src/Platform/Microsoft.Testing.Platform/CommandLine/CommandLineHandler.cs
Evangelink: Ok let's ignore then.

Evangelink (approved):

PR #5040, opened 2025-02-17
Fix breaking change with telemetry

### LINKED ISSUE #5015: MTP extensions - problem between 1.5.3 and 1.6.0
## Describe the bug

Bump TRX report on the given sample to `1.6.0` and there will be crash.

[test_plat_sent-fixed.zip](https://github.com/user-attachments/files/18789866/test_plat_sent-fixed.zip)

Evangelink:
Fixes #5015, manually validated.


src/Platform/Microsoft.Testing.Platform/Builder/TestApplicationBuilder.cs
```diff
@@ -68,6 +68,9 @@ internal TestApplicationBuilder(
 
     internal ITelemetryManager Telemetry => _testHostBuilder.Telemetry;
 
+    [Obsolete("Remove in v2. Avoid breaking change with the rename of the property. See https://github.com/microsoft/testfx/issues/5015")]
+    internal ITelemetryManager TelemetryManager => _testHostBuilder.Telemetry;
```
Youssef1313, line 75: There was another rename, something used by `Retry` extension IIRC. Will we fix it as well?

src/Platform/Microsoft.Testing.Platform/Builder/TestApplicationBuilder.cs
```diff
@@ -68,6 +68,9 @@ internal TestApplicationBuilder(
 
     internal ITelemetryManager Telemetry => _testHostBuilder.Telemetry;
 
+    [Obsolete("Remove in v2. Avoid breaking change with the rename of the property. See https://github.com/microsoft/testfx/issues/5015")]
```
Youssef1313: Do we want to maybe go strong and make it error?

```suggestion
    [Obsolete("Remove in v2. Avoid breaking change with the rename of the property. See https://github.com/microsoft/testfx/issues/5015", error: true)]
```

src/Platform/Microsoft.Testing.Platform/Builder/TestApplicationBuilder.cs
Evangelink: Good catch, thanks!

Youssef1313 (approved):

PR #5042, opened 2025-02-17
Follow-up to ignore fix

Youssef1313:
Follow-up to #5020

Per discussion with @nohwnd, the serialization/deserialization logic in vstest doesn't seem to round trip null string correctly, and we get it back as empty string. Changing the vstest behavior is risky as there may be users/components out there relying on that behavior. Not setting the property at all when it's null should be fine, and potentially having better performance.

This PR gets back the behavior of empty string being allowed to ignore the test, but still not regress the null case as we now don't set the property at all.

The change here is manually validated.

### DIFF
--- a/src/Adapter/MSTest.TestAdapter/ObjectModel/UnitTestElement.cs
+++ b/src/Adapter/MSTest.TestAdapter/ObjectModel/UnitTestElement.cs
@@ -195,7 +195,10 @@ internal sealed class UnitTestElement
 
             testCase.SetPropertyValue(Constants.TestDynamicDataTypeProperty, (int)TestMethod.DataType);
             testCase.SetPropertyValue(Constants.TestDynamicDataProperty, data);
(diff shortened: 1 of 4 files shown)


src/Adapter/MSTest.TestAdapter/ObjectModel/UnitTestElement.cs
```diff
@@ -195,7 +195,10 @@ internal TestCase ToTestCase()
 
             testCase.SetPropertyValue(Constants.TestDynamicDataTypeProperty, (int)TestMethod.DataType);
             testCase.SetPropertyValue(Constants.TestDynamicDataProperty, data);
-            testCase.SetPropertyValue(Constants.TestDataSourceIgnoreMessageProperty, TestMethod.TestDataSourceIgnoreMessage);
+            if (TestMethod.TestDataSourceIgnoreMessage is not null)
```
Evangelink, line 199:
```suggestion
            // VSTest serialization doesn't handle null so instead don't set the property so that it's deserialized as null
            if (TestMethod.TestDataSourceIgnoreMessage is not null)
```

Evangelink (approved):

Issue #5039, opened 2025-02-17
MTP should be built with `SupportedPlatform` items for Android, iOS, Browser


Youssef1313:
We should have the following when building on MTP so that we get warnings from unsupported APIs and we explicitly decide how to handle for the unsupported platform

```
  <ItemGroup>
    <SupportedPlatform Include="android" />
    <SupportedPlatform Include="ios" />
    <SupportedPlatform Include="browser" />
  </ItemGroup>
```

PR #5046, opened 2025-02-17
Enable platform compatibility warnings for android, ios, tvos, and browser

Youssef1313:
Contributes to #5039

### DIFF
--- a/src/Platform/Microsoft.Testing.Platform/Helpers/System/SystemConsole.cs
+++ b/src/Platform/Microsoft.Testing.Platform/Helpers/System/SystemConsole.cs
@@ -51,61 +51,14 @@ internal sealed class SystemConsole : IConsole
         };
     }
 
-    // This is based on the UnsupportedOSPlatformAttribute on the API:
-    // https://github.com/dotnet/runtime/blob/8a79637e1454c751c24a984b7fb8af4c4d43394b/src/libraries/System.Console/src/System/Console.cs#L560-L563
-    [SupportedOSPlatformGuard("android")]
-    [SupportedOSPlatformGuard("ios")]
-    [SupportedOSPlatformGuard("tvos")]
-    [SupportedOSPlatformGuard("browser")]
-    private static bool IsCancelKeyPressNotSupported()
-        => RuntimeInformation.IsOSPlatform(OSPlatform.Create("ANDROID")) ||
-            RuntimeInformation.IsOSPlatform(OSPlatform.Create("IOS")) ||
-            RuntimeInformation.IsOSPlatform(OSPlatform.Create("TVOS")) ||
-            RuntimeInformation.IsOSPlatform(OSPlatform.Create("BROWSER"));
-
-    // This is based on the UnsupportedOSPlatformAttribute on the API:
-    // https://github.com/dotnet/runtime/blob/d6cfa2039309c1d5cef630e95eb5274a556dc726/src/libraries/System.Console/src/System/Console.cs#L334-L337
-    [SupportedOSPlatformGuard("android")]
-    [SupportedOSPlatformGuard("ios")]
-    [SupportedOSPlatformGuard("tvos")]
-    [SupportedOSPlatformGuard("browser")]
-    private static bool IsForegroundColorNotSupported()
-        => RuntimeInformation.IsOSPlatform(OSPlatform.Create("ANDROID")) ||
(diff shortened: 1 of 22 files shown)


src/Platform/Microsoft.Testing.Platform/Microsoft.Testing.Platform.csproj
```diff
@@ -74,6 +74,13 @@ This package provides the core platform and the .NET implementation of the proto
     <EmbeddedResource Update="Resources\PlatformResources.resx" GenerateSource="true" />
   </ItemGroup>
 
+  <ItemGroup>
+    <SupportedPlatform Include="android" />
```
Evangelink, line 80: Shall we add `windows`, `linux` and `macOS` or is it default?

src/Platform/Microsoft.Testing.Platform/Helpers/System/SystemConsole.cs
```diff
@@ -13,11 +13,19 @@ internal sealed class SystemConsole : IConsole
     /// <summary>
     /// Gets the height of the buffer area.
     /// </summary>
+    [UnsupportedOSPlatform("android")]
```
Evangelink, line 16: We should probably have all these also on the `IConsole`.

src/Platform/Microsoft.Testing.Platform/Helpers/System/SystemConsole.cs
```diff
@@ -45,12 +53,38 @@ static SystemConsole()
...
             RuntimeInformation.IsOSPlatform(OSPlatform.Create("TVOS")) ||
             RuntimeInformation.IsOSPlatform(OSPlatform.Create("BROWSER"));
 
+    // This is based on the UnsupportedOSPlatformAttribute on the API:
+    // https://github.com/dotnet/runtime/blob/d6cfa2039309c1d5cef630e95eb5274a556dc726/src/libraries/System.Console/src/System/Console.cs#L334-L337
+    [SupportedOSPlatformGuard("android")]
+    [SupportedOSPlatformGuard("ios")]
+    [SupportedOSPlatformGuard("tvos")]
+    [SupportedOSPlatformGuard("browser")]
+    private static bool IsForegroundColorNotSupported()
```
Evangelink: Can we drop the 3 `Is...` and instead just annotate the API?

src/Platform/Microsoft.Testing.Platform/Microsoft.Testing.Platform.csproj
Youssef1313: They are default. https://github.com/dotnet/sdk/blob/cb459ebc2d9374b2bc5ce944cc633a6e79ed8275/src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.SupportedPlatforms.props

src/Platform/Microsoft.Testing.Platform/Helpers/System/SystemConsole.cs
Youssef1313: Yes, this is actually a false negative I reported in https://github.com/dotnet/roslyn-analyzers/issues/7579

src/Platform/Microsoft.Testing.Platform/Helpers/System/SystemConsole.cs
Youssef1313: The other APIs (CancelKeyPress, ForegroundColor, Clear), we want to make them callable on any platform. So they are *supported* (supported here means they don't throw, not that they really work), but we make them no-op on those platforms.

Evangelink: Don't we instead want to push the decision of calling or not to the caller? Like it's done on Console.

Youssef1313: @Evangelink The *existing* code was already to try to be no-op on Android and iOS.

Generally, I think whether or not to push the decision to the caller depends on the API itself. If it's really an important functionality, push to caller

Youssef1313: Addressed per our discussion offline.

Evangelink (approved):

PR #5054, opened 2025-02-18
Fix serialization of exceptions by BinaryFormatter in .NET Framework

### LINKED ISSUE #5048: AssertException is not compatible with BinaryFormatter
## Describe the bug

AssertException is not compatible with BinaryFormatter which is still available and used in .NET Framework.

## Steps To Reproduce

Run this code in net472 project (here: [TestProject134.zip](https://github.com/user-attachments/files/18828284/TestProject134.zip) )

```cs
using System.Runtime.Serialization.Formatters.Binary;

namespace TestProject134
{
    [TestClass]
    public sealed class Test1
    {
        [TestMethod]
        public void AssertExceptionDeserializationFails()
        {
            try
            {
                Assert.AreEqual(1, 2);
            }
            catch (Exception ex)
            {
                Write(ex);
            }
        }

        [TestMethod]
        public void InvalidOperationExceptionDeserializationPasses()
        {
            try
            {
                throw new InvalidOperationException("aaaaa");
            }
            catch (Exception ex)
            {
                Write(ex);
            }
        }

        private static void Write(Exception ex)
        {
            var mem = new MemoryStream();
            new BinaryFormatter().Serialize(mem, ex);
            mem.Position = 0;
            var deser = new BinaryFormatter().Deserialize(mem);

            Console.WriteLine(deser);
        }
    }
}
```

## Expected behavior

Both tests pass an show me the actual error.

## Actual behavior

AssertException fails to deserialize with:

```
 Test method TestProject134.Test1.AssertExceptionDeserializationFails threw exception: 
System.Runtime.Serialization.SerializationException: The constructor to deserialize an object of type 'Microsoft.VisualStudio.TestTools.UnitTesting.AssertFailedException' was not found. ---> System.Runtime.Serialization.SerializationException: The constructor to deserialize an object of type 'Microsoft.VisualStudio.TestTools.UnitTesting.AssertFailedException' was not found.

  Stack Trace: 
ObjectManager.GetConstructor(RuntimeType t)
ObjectManager.CompleteISerializableObject(Object obj, SerializationInfo info, StreamingContext context)
--- End of inner exception stack trace ---
ObjectManager.CompleteISerializableObject(Object obj, SerializationInfo info, StreamingContext context)
ObjectManager.FixupSpecialObject(ObjectHolder holder)
ObjectManager.DoFixups()
ObjectReader.Deserialize(HeaderHandler handler, __BinaryParser serParser, Boolean fCheck, Boolean isCrossAppDomain, IMethodCallMessage methodCallMessage)
BinaryFormatter.Deserialize(Stream serializationStream, HeaderHandler handler, Boolean fCheck, Boolean isCrossAppDomain, IMethodCallMessage methodCallMessage)
BinaryFormatter.Deserialize(Stream serializationStream)
Test1.Write(Exception ex) line 39
Test1.AssertExceptionDeserializationFails() line 17
```

## Additional context

I am unable to see details of failures when writing UI tests for VS. I am not sure if this worked previously, but I don't see it working in mstest 3.3.1 either.

nohwnd:
The change adds protected constructors that binary formatter needs to deserialize exceptions. All added api targets .NET Framework, and .NET Standard 2.0, to avoid bringing the new apis to .NET.

All those apis are obsoleted the same way they are obsoleted in .NET.

Fix #5048


Evangelink (approved):

src/TestFramework/TestFramework/Exceptions/AssertFailedException.cs
```diff
@@ -35,4 +40,18 @@ public AssertFailedException()
         : base()
     {
     }
+
+#if NETFRAMEWORK || NETSTANDARD2_0
```
Youssef1313: Looks dangerous to me to include an API in netstandard2.0 but not when building for net6.0.

I guess we should either go for all, or only .NET Framework. @Evangelink thoughts?

Evangelink (changes requested):
Reviewing the link https://github.com/dotnet/docs/issues/34893, I think we should be really following the advices there and add the protected ctor always and make it obsolete for NET8.0+:

```csharp
#if NET8_0_OR_GREATER
    [Obsolete(DiagnosticId = "SYSLIB0051")]
#endif
```

src/TestFramework/TestFramework/Exceptions/AssertFailedException.cs
nohwnd: this should be solved by adding a warning to the package targets when it restores for a version that is compatible with netstandard2.0, but is older than your lowest supported net version. I thought we had an issue for this, but cannot find it.

nohwnd: Idk where that ended up, but the idea is like what system.io.hashing is doing:

image



```
<Project InitialTargets="NETStandardCompatError_System_IO_Hashing_net8_0">
  <Target Name="NETStandardCompatError_System_IO_Hashing_net8_0"
          Condition="'$(SuppressTfmSupportBuildWarnings)' == ''">
    <Warning Text="System.IO.Hashing 9.0.2 doesn't support $(TargetFramework) and has not been tested with it. Consider upgrading your TargetFramework to net8.0 or later. You may also set &lt;SuppressTfmSupportBuildWarnings&gt;true&lt;/SuppressTfmSupportBuildWarnings&gt; in the project file to ignore this warning and attempt to run in this unsupported configuration at your own risk." />
  </Target>
</Project>


```

nohwnd: My logic here was that binary formatter is removed / heavily deprecated in all net8 and newer, so why add new api to support it, even if we still build for net6.

nohwnd: okay so no issue was filed, but we discussed it here: https://teams.microsoft.com/l/message/19:28aafb7e5e764075b048b64fd4d9d4a5@thread.v2/1737016870258?context=%7B%22contextType%22%3A%22chat%22%7D

nohwnd:
As I read it the recommendation does not address the case where we did not have the serialization apis before we decided to Obsolete them. So the recommendation does not apply here. 

They recommend to not rely on binary formatter first and foremost. This we cannot influence, but we can avoid people using the new constructor by accident, hance the obsoletion and browsable Never.

They recommend to remove the serialization constructor if we can, which we effectively do in .NET, by having compiler condition.

If we cannot  remove in .net (which is not our case) we should obsolete it. For that they recommend to conditionally obsolete it in .NET 8.

So afaik my code is correct.

nohwnd:
I am referring to:

If you're creating a custom System.Exception-derived type

Consider whether you truly need your custom exception type to be serializable. Chances are you do not need it to be serializable, ... _<- yup we don't need that in .net._

Consider simply removing the [Serializable] attribute, the serialization constructor, and the GetObjectData method override, as shown below. _<- done by not defining them for .NET_

There may be some cases where you cannot remove these APIs from your custom exception types. This might occur if you produce a library constrained by API compatibility requirements. _<- we did not have them till now so we don't need to keep compatibility._

Youssef1313:
I'd prefer to have consistent behavior. If we want to support .NET Framework, let's support everywhere with obsolete.

My *second* preference is to only support .NET Framework, and not include the serialization constructor on netstandard2.0.

nohwnd:
amary will continue here on the suggestion above, with having the api everywhere and only conditional obsolete. Thanks

@Evangelink  I took the obsolete message from here:         

internal const string LegacyFormatterImplMessage = "This API supports obsolete formatter-based serialization. It should not be called or extended by application code.";

internal const string LegacyFormatterImplDiagId = "SYSLIB0051"; 
        
internal const string SharedUrlFormat = "https://aka.ms/dotnet-warnings/{0}";
        
https://github.com/dotnet/runtime/blob/main/src/libraries/Common/src/System/Obsoletions.cs#L8

Evangelink (approved):

PR #5057, opened 2025-02-18
Avoid using unsupported APIs

Youssef1313:


src/Platform/Microsoft.Testing.Platform/OutputDevice/BrowserOutputDevice.cs
```diff
@@ -0,0 +1,401 @@
...
+    public string Description { get; } = "Test Platform default browser console service";
+
+    /// <inheritdoc />
+    public Task<bool> IsEnabledAsync() => Task.FromResult(true);
+
+    public async Task DisplayBannerAsync(string? bannerMessage)
+    {
+        using (await _asyncMonitor.LockAsync(TimeoutHelper.DefaultHangTimeSpanTimeout))
+        {
+            if (!_bannerDisplayed)
```
Evangelink: Nit: revert to reduce nesting

src/Platform/Microsoft.Testing.Platform/OutputDevice/BrowserOutputDevice.cs
```diff
@@ -0,0 +1,401 @@
...
+            {
+                // skip the banner for the children processes
+                _environment.SetEnvironmentVariable(TESTINGPLATFORM_CONSOLEOUTPUTDEVICE_SKIP_BANNER, "1");
+
+                _bannerDisplayed = true;
+
+                if (bannerMessage is not null)
+                {
+                    ConsoleLog(bannerMessage);
```
Evangelink: Nit: add `return;` and remove `else` branch to reduce nesting.

src/Platform/Microsoft.Testing.Platform/OutputDevice/BrowserOutputDevice.cs
```diff
@@ -0,0 +1,401 @@
...
+    {
+        AppendAssemblyLinkTargetFrameworkAndArchitecture(_console, _assemblyName, _targetFramework, _longArchitecture);
+        return Task.CompletedTask;
+    }
+
+    private static void AppendAssemblyLinkTargetFrameworkAndArchitecture(IConsole console, string assembly, string? targetFramework, string? architecture)
+    {
+        var builder = new StringBuilder();
+        builder.Append(assembly);
+        if (targetFramework != null || architecture != null)
```
Evangelink: Nit: I'd probably revert here too

src/Platform/Microsoft.Testing.Platform/OutputDevice/BrowserOutputDevice.cs
```diff
@@ -0,0 +1,401 @@
...
+        }
+
+        console.WriteLine(builder.ToString());
+    }
+
+    public async Task DisplayAfterSessionEndRunAsync()
+    {
+        using (await _asyncMonitor.LockAsync(TimeoutHelper.DefaultHangTimeSpanTimeout))
+        {
+            if (!_firstCallTo_OnSessionStartingAsync)
```
Evangelink: Nit: revert condition to reduce nesting

src/Platform/Microsoft.Testing.Platform/OutputDevice/BrowserOutputDevice.cs
```diff
@@ -0,0 +1,401 @@
...
+    private static partial void ConsoleError(string? message);
+
+    [JSImport("globalThis.console.log")]
+    private static partial void ConsoleLog(string? message);
+#else
+    private void ConsoleWarn(string? message) => _console.WriteLine(message);
+
+    private void ConsoleError(string? message) => _console.WriteLine(message);
+
+    private void ConsoleLog(string? message) => _console.WriteLine(message);
```
Evangelink, line 301: Suggestion that doesn't need to be done here. I know nothing about web so maybe that's totally stupid. Is it common to have some kind of prefix for warning/errors in the console? If so we may want to prefix the message

Youssef1313: @Evangelink Usually we would want to use the right API (e.g, `console.error` and `console.warning`), which is already done when `JSImport` is available (net7+).

Ideally, we will want to still always call `console.error` and `console.warning` for all TFMs.

@captainsafia Sorry for the ping here, but I can't seem to find the "old" way of doing JS interop as probably all docs are updated to the source-generated attributes. Do you know how to call `console.error` and `console.warning` when the source generator isn't available? (Note: we are a netstandard2.0 and net6+ library)

captainsafia: I don't know off the top of my head but @MackinnonBuck might have ideas here.

Evangelink:
As discussed offline with @Youssef1313, let's merge as-is and follow up

PR #5070, opened 2025-02-19
Fix ClassCleanup not called when the first test in class is ignored

### LINKED ISSUE #5062: ClassCleanup is not executed when Ignore (Attribute) is applied to the first test method in a series of test methods in a test class
## Describe the bug

The ClassCleanup is not executed when Ignore (Attribute) is applied to the first test method in a series of test methods in a test class.

## Steps To Reproduce

The following code reproduces the issue. You can set a breakpoint in `ClassCleanup` method and observe that it is not executed.

Note that the `ClassCleanup` method is executed if we move the `[Ignore]` attribute from `TestMethod1` to `TestMethod2` instead.

I am able to reproduce the bug with MSTest 3.8.0.

```
namespace UnitTestBugRepro
{
    using System;

    [TestClass]
    public class UnitTest1
    {
        [ClassInitialize()]
        public static void ClassInit(TestContext context)
        {
            Console.WriteLine("Class Init");
        }

        [ClassCleanup(ClassCleanupBehavior.EndOfClass)]
        {
            Console.WriteLine("Class Cleanup");
        }

        [Ignore]
        [TestMethod]
        public void TestMethod1()
        {
            Console.WriteLine("TestMethod1");
        }

        [TestMethod]
        public void TestMethod2()
        {
            Console.WriteLine("TestMethod2");
        }
    }
}
```

## Expected behavior

`TestMethod1()` is ignored and `ClassCleanup()` method is executed.

## Actual behavior

`TestMethod1()` is ignored and `ClassCleanup()` method is **not** executed.

Youssef1313:
When the first test of a class is ignored (and the first test ever in the assembly), we will not call AssemblyInitialize, so `IsAssemblyInitializeExecuted` will be false.

However, in that case, we still want to call `RunClassCleanup` so that the test is marked as completed (see `MarkTestComplete`). Not marking the ignored test as completed is problematic, because later on when all tests in the class are executed, we will still be viewing remaining tests in the class, hence we will not call class cleanup.

Note that even with this change, we respect the case where all tests in class are ignored, and we will not call ClassCleanup. The logic in `ExecuteClassCleanup` handles the case when class init wasn't called.

Fixes #5062

### DIFF
--- a/test/IntegrationTests/MSTest.Acceptance.IntegrationTests/IgnoreTests.cs
+++ b/test/IntegrationTests/MSTest.Acceptance.IntegrationTests/IgnoreTests.cs
@@ -193,6 +193,7 @@ public class SubClass : IntermediateClass
     public static void SubClassCleanup()
         => Console.WriteLine("SubClass.ClassCleanup");
 
+    // Ignore the first test on purpose, see https://github.com/microsoft/testfx/issues/5062
     [TestMethod]
     [Ignore]
     public void IgnoredMethod()
--- a/test/IntegrationTests/MSTest.Acceptance.IntegrationTests/IgnoreTests.cs
+++ b/test/IntegrationTests/MSTest.Acceptance.IntegrationTests/IgnoreTests.cs
@@ -18,9 +18,11 @@ public sealed class IgnoreTests : AcceptanceTestBase<IgnoreTests.TestAssetFixtur
 
         // Assert
         testHostResult.AssertExitCodeIs(ExitCodes.Success);
-        testHostResult.AssertOutputContainsSummary(failed: 0, passed: 11, skipped: 7);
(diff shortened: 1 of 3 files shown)


test/IntegrationTests/MSTest.Acceptance.IntegrationTests/IgnoreTests.cs
```diff
@@ -191,6 +193,11 @@ public class SubClass : IntermediateClass
     public static void SubClassCleanup()
         => Console.WriteLine("SubClass.ClassCleanup");
 
+    [TestMethod]
```
Evangelink, line 197:
```suggestion
    // Ignore the first test on purpose, see https://github.com/microsoft/testfx/issues/5062
    [TestMethod]
```

Evangelink (approved):
Suggesting a comment on the test so we don't change order without noticing impact

Evangelink (approved):

Issue #5072, opened 2025-02-19
error TESTERROR:        GetDownload (53ms): Error Message: Error calling Test Cleanup method for test class Aiursoft.Tracer.Tests.IntegrationTests.BasicTests: System.MethodAccessException: Attempt by method


Anduin2017:
After upgrading:

```log
Exit code is 0

Last runner error: Process /usr/lib/dotnet/dotnet:192904 exited with code '0'.

--- EXCEPTION #1/1 [LoggerException]
Message = “Process /usr/lib/dotnet/dotnet:192904 exited with code '0'.”
StackTraceString = “
  at JetBrains.Util.ILoggerEx.LogMessage(ILogger logger, LoggingLevel level, String message) in LogMessage.il:line IL_002E mvid 48B4
     at JetBrains.Util.ILoggerEx.Error(ILogger logger, String message) in Error.il:line IL_0000 mvid 48B4
     at JetBrains.ReSharper.UnitTestFramework.Execution.TestRunner.TestRunnerRunStrategy.AnalyseProcessCrash(ProcessExitedUnexpectedlyException ex, IUnitTestRun run) in AnalyseProcessCrash.il:line IL_006B mvid 1598
     at JetBrains.ReSharper.UnitTestFramework.Execution.TestRunner.TestRunnerRunStrategy.Run(IUnitTestRun run) in Run.il:line IL_01C8 mvid 1598
     at System.Runtime.CompilerServices.AsyncTaskMethodBuilder1.AsyncStateMachineBox1.ExecutionContextCallback(Object s) in ExecutionContextCallback.il:line IL_0000 mvid C608
     at System.Threading.ExecutionContext.RunInternal(ExecutionContext executionContext, ContextCallback callback, Object state) in RunInternal.il:line IL_0040 mvid C608
     at System.Runtime.CompilerServices.AsyncTaskMethodBuilder1.AsyncStateMachineBox1.MoveNext(Thread threadPoolThread) in MoveNext.il:line IL_0000 mvid C608 or MoveNext.il:line IL_0040 mvid C608 or MoveNext.il:line IL_004E mvid C608
     at System.Runtime.CompilerServices.AsyncTaskMethodBuilder1.AsyncStateMachineBox1.MoveNext() in MoveNext.il:line IL_0000 mvid C608 or MoveNext.il:line IL_0040 mvid C608 or MoveNext.il:line IL_004E mvid C608
[...]

PR #5073, opened 2025-02-19
Fix IVT from TestFramework.Extensions to TestAdapter

Youssef1313:
Potential fix for #5072, but needs to be confirmed.


Youssef1313:
This is actually not changing anything.

PR #5079, opened 2025-02-19
Respect EditorConfig spec about comments

Youssef1313:
https://spec.editorconfig.org/#no-inline-comments

> *Changed in version 0.15.0.*
>
> A `;` or `#` anywhere other than at the beginning of a line does not start a comment, but is part of the text of that line.


Evangelink:
That sucks! I saw this pattern on a repo and found it way better than the line by line comment :(

Evangelink (approved):

PR #3703, opened 2024-08-27
Manual resolution of the muxer for `dotnet test` MSBuild extension

MarcoRossignoli:
closes https://github.com/microsoft/testfx/issues/3689

PR #5087, opened 2025-02-20
Fix typo in determining `_TestArchitecture`

Youssef1313:
@Evangelink Worth backporting? Not sure if it can affect some arm64 scenarios

### DIFF
--- a/src/Platform/Microsoft.Testing.Platform.MSBuild/buildMultiTargeting/Microsoft.Testing.Platform.MSBuild.targets
+++ b/src/Platform/Microsoft.Testing.Platform.MSBuild/buildMultiTargeting/Microsoft.Testing.Platform.MSBuild.targets
@@ -340,7 +340,7 @@
     <!--
     NETCoreSdkRuntimeIdentifier is used by DefaultAppHostRuntimeIdentifier to figure out the architecture, and is defined also for .NET Framework.
     -->
-    <_TestArchitecture Condition="('$(_TestArchitecture)' == '' or '$(_TestArchitecture)' == 'AnyCpu') and '$(NETCoreSdkRuntimeIdentifier)' != ''">$([System.Text.RegularExpressions.Regex]::Replace($(DefaultAppHostRuntimeIdentifier), '.*-', ''))</_TestArchitecture>
+    <_TestArchitecture Condition="('$(_TestArchitecture)' == '' or '$(_TestArchitecture)' == 'AnyCpu') and '$(NETCoreSdkRuntimeIdentifier)' != ''">$([System.Text.RegularExpressions.Regex]::Replace($(NETCoreSdkRuntimeIdentifier), '.*-', ''))</_TestArchitecture>
     <!-- If everyting else fails, fallback to x64. -->
     <_TestArchitecture Condition="'$(_TestArchitecture)' == '' or '$(_TestArchitecture)' == 'AnyCpu'">x64</_TestArchitecture>
     <TestingPlatformShowTestsFailure Condition=" '$(TestingPlatformShowTestsFailure)' == '' ">False</TestingPlatformShowTestsFailure>


src/Platform/Microsoft.Testing.Platform.MSBuild/buildMultiTargeting/Microsoft.Testing.Platform.MSBuild.targets
```diff
@@ -340,7 +340,7 @@
     <!--
     NETCoreSdkRuntimeIdentifier is used by DefaultAppHostRuntimeIdentifier to figure out the architecture, and is defined also for .NET Framework.
     -->
-    <_TestArchitecture Condition="('$(_TestArchitecture)' == '' or '$(_TestArchitecture)' == 'AnyCpu') and '$(NETCoreSdkRuntimeIdentifier)' != ''">$([System.Text.RegularExpressions.Regex]::Replace($(DefaultAppHostRuntimeIdentifier), '.*-', ''))</_TestArchitecture>
+    <_TestArchitecture Condition="('$(_TestArchitecture)' == '' or '$(_TestArchitecture)' == 'AnyCpu') and '$(NETCoreSdkRuntimeIdentifier)' != ''">$([System.Text.RegularExpressions.Regex]::Replace($(NETCoreSdkRuntimeIdentifier), '.*-', ''))</_TestArchitecture>
```
Youssef1313, line 343: This is a copy from line 339, but when it was copy-pasted, the condition was modified but the actual value is not. The code is meaningless on its current form.

src/Platform/Microsoft.Testing.Platform.MSBuild/buildMultiTargeting/Microsoft.Testing.Platform.MSBuild.targets
```diff
@@ -340,7 +340,7 @@
...
     <!-- If everyting else fails, fallback to x64. -->
     <_TestArchitecture Condition="'$(_TestArchitecture)' == '' or '$(_TestArchitecture)' == 'AnyCpu'">x64</_TestArchitecture>
```
Youssef1313, line 345: @Evangelink This looks suspicious. I don't know how we can reach this "fallback" case. But if we do, I think we should use whatever `dotnet` we find from `PATH` and not force a specific architecture.

Evangelink: @MarcoRossignoli do you recall?

Youssef1313: https://github.com/microsoft/testfx/issues/5091 suggests a larger change to how InvokeTestingPlatformTask work.

Youssef1313: Merging for now as-is and we can address this concern later if #5091 isn't going to happen soon.

nohwnd: I believe it was me who wrote some of the code in test anywhere repo and lot of the comments :) see 330, the _testArchitecture is used for log naming, we don't use it to influence which architecture will run, instead we just try to guess what is the architecture of the exe we run. 

So when user builds for arm64 and runs they would see _arm64 in the log names, and not _x64.

MarcoRossignoli: If I recall well @nohwnd did this update, @nohwnd do you recall?

nohwnd: just posted above 1 min ago :)

Youssef1313: > we don't use it to influence which architecture will run

The current implementation *does* that.

https://github.com/microsoft/testfx/blob/cd21e24a1b4531389be892bbd1fbbd387085bd64/src/Platform/Microsoft.Testing.Platform.MSBuild/Tasks/InvokeTestingPlatformTask.cs#L172

It tries to resolve `dotnet` by the `_TestArchitecture` defined.

nohwnd: https://github.com/microsoft/testfx/pull/3703 that was introduced here, long time after the comment.  So I stand corrected.

Youssef1313: Makes sense. Anyways, I think #5091 would be the ultimate path forward

Issue #5088, opened 2025-02-20
Update CI to run MacOS ARM64


Youssef1313:
Currently, our MacOS jobs are x64. We should either add new jobs that run on MacOS ARM64, or simply move the current jobs from x64 to ARM64.

Testing ARM64 is super relevant to us, especially the MTP MSBuild extension that has logic around the architecture.

nohwnd:
https://helix.dot.net/#Test-External-macOS this lists e.g. osx.15.arm64.open, can that not be used?

nohwnd:
I see no build part there, and I am signed into that page 😁

PR #5089, opened 2025-02-20
Use MacOS ARM64

Youssef1313:
Fixes https://github.com/microsoft/testfx/issues/5088

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -194,7 +194,7 @@ stages:
         timeoutInMinutes: 90
         pool:
           name: Azure Pipelines
-          vmImage: macos-14-arm64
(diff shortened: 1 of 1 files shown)


azure-pipelines.yml
```diff
@@ -194,7 +194,7 @@ stages:
         timeoutInMinutes: 90
         pool:
           name: Azure Pipelines
-          vmImage: macos-latest
+          image: macos-14-arm64
```
Youssef1313, line 197: Was copying from https://github.com/microsoft/sbom-tool/blob/20f43605fa20db121f86de73366284970c22f0a9/pipelines/sbom-tool-pr-build.yaml#L55-L57 but it's not working. 😕

PR #5093, opened 2025-02-21
Resolve types in test binary formatter binder

nohwnd:
Resolve type in binder based on the exception. Null binders are not allowed.

### DIFF
--- a/test/UnitTests/TestFramework.UnitTests/Exceptions/BinaryFormatterExceptionSerializationTests.cs
+++ b/test/UnitTests/TestFramework.UnitTests/Exceptions/BinaryFormatterExceptionSerializationTests.cs
@@ -35,7 +35,7 @@ public sealed class BinaryFormatterExceptionSerializationTests : TestContainer
             var mem = new MemoryStream();
             var formatter = new BinaryFormatter
             {
-                Binder = new FormatterBinder(),
+                Binder = new FormatterBinder(ex.GetType()),
             };
             formatter.Serialize(mem, ex);
             mem.Position = 0;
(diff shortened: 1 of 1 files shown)


Youssef1313 (approved):

Youssef1313 (changes requested):
Actually I think we can do better.

The binder class can be implemented like this:

```csharp
    /// <summary>
    /// This is for compliance, usage of BinaryFormatter without binder is not allowed.
    /// </summary>
    private class FormatterBinder : SerializationBinder
    {
        private readonly Type _type;

        public FormatterBinder(Type type)
            => _type = type;

        public override Type BindToType(string assemblyName, string typeName)
            => assemblyName == _type.Assembly.ToString() && typeName == _type.FullName
            ? _type
            : throw new InvalidOperationException();
    }
```

And `VerifySerialization` should have a new parameter, `Type type` that's being passed to `FormatterBinder` constructor.

Then individual test methods should call `VerifySerialization` with the correct type.

```csharp
    public void AssertFailedExceptionCanBeSerializedAndDeserialized()
        => VerifySerialization(Assert.Fail, typeof(AssertFailedException));

    public void AssertInconclusiveExceptionCanBeSerializedAndDeserialized()
        => VerifySerialization(Assert.Inconclusive, typeof(AssertInconclusiveException));

    public void InternalTestFailureExceptionCanBeSerializedAndDeserialized()
        => VerifySerialization(() => throw new InternalTestFailureException("Some internal error."), typeof(InternalTestFailureException));
```

nohwnd:
Yeah that is better, implemented, using the type from `ex`. Confirmed it still fails the test when the serialization ctor is removed.

test/UnitTests/TestFramework.UnitTests/Exceptions/BinaryFormatterExceptionSerializationTests.cs
```diff
@@ -35,7 +35,7 @@ private void VerifySerialization(Action actionThatThrows)
             var mem = new MemoryStream();
             var formatter = new BinaryFormatter
             {
-                Binder = new FormatterBinder(),
+                Binder = new FormatterBinder(ex.GetType()),
```
Youssef1313, line 38: Nice! That's a better idea than my suggestion to add `Type` as a parameter of `VerifySerialization`. Thanks!!

Youssef1313 (approved):

test/UnitTests/TestFramework.UnitTests/Exceptions/BinaryFormatterExceptionSerializationTests.cs
nohwnd: Let's hope it makes the analyzer happy finally :D

PR #5094, opened 2025-02-21
Update InvokeTestingPlatformTask to prefer running Exe over `dotnet exec dll`

Youssef1313:
Manually verified this works as expected. For https://github.com/microsoft/PowerToys/pull/37001, tests run successfully when running the Exe directly but not when using `dotnet exec dll`. I wasn't able to set up projects in a similar way as PowerToys to repro the issue and have a test. But I think it makes sense to run directly, and potentially even faster. We probably have enough tests to know if this change is problematic for a main stream scenario.

### DIFF
--- a/src/Platform/Microsoft.Testing.Platform.MSBuild/Tasks/InvokeTestingPlatformTask.cs
+++ b/src/Platform/Microsoft.Testing.Platform.MSBuild/Tasks/InvokeTestingPlatformTask.cs
@@ -373,6 +373,18 @@ public class InvokeTestingPlatformTask : Build.Utilities.ToolTask, IDisposable
     /// <inheritdoc />
     public override bool Execute()
     {
+        var list = new List<string>();
+        foreach (DictionaryEntry entry in new SystemEnvironment().GetEnvironmentVariables())
+        {
+            if (entry.Key is string key && key.StartsWith("DOTNET_ROOT", StringComparison.OrdinalIgnoreCase))
+            {
+                list.Add($"{key}={entry.Value}");
+            }
+        }
+
+        // This should be done before the base.Execute() call.
+        EnvironmentVariables = list.ToArray();
+
         bool returnValue = base.Execute();
         if (_toolCommand is not null)
         {
--- a/src/Platform/Microsoft.Testing.Platform.MSBuild/Tasks/InvokeTestingPlatformTask.cs
+++ b/src/Platform/Microsoft.Testing.Platform.MSBuild/Tasks/InvokeTestingPlatformTask.cs
@@ -93,7 +93,7 @@ public class InvokeTestingPlatformTask : Build.Utilities.ToolTask, IDisposable
     /// <summary>
     /// Gets or sets the value of MSBuild property _NativeExecutableExtension.
     /// </summary>
-    public ITaskItem NativeExecutableExtension { get; set; }
+    public ITaskItem? NativeExecutableExtension { get; set; }
 
     // -------- END the previous properties shouldn't be used. See https://github.com/microsoft/testfx/issues/5091 --------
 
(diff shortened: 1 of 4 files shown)


src/Platform/Microsoft.Testing.Platform.MSBuild/buildMultiTargeting/Microsoft.Testing.Platform.MSBuild.targets
```diff
@@ -352,8 +352,16 @@
...
+    <!-- Passing UseAppHost, IsExecutable, TargetDir, AssemblyName, and _NativeExecutableExtensions is mostly workaround -->
+    <!-- until we implement https://github.com/microsoft/testfx/issues/5091 properly -->
+    <!-- For now, we construct the correct "RunCommand" via these properties for a specific scenario that's known to be broken -->
     <CallTarget Targets="_TestRunStart" />
     <InvokeTestingPlatformTask TargetPath="$(TargetPath)"
+                               UseAppHost="$(UseAppHost)"
+                               IsExecutable="$(_IsExecutable)"
+                               TargetDir="$(TargetDir)"
+                               AssemblyName="$(AssemblyName)"
+                               NativeExecutableExtension="$(_NativeExecutableExtension)"
```
Youssef1313, line 370: @baronfel Could you please confirm if it's okay to use `_NativeExecutableExtension` and `_IsExecutable`? We want to avoid using things that may potentially break in future.

src/Platform/Microsoft.Testing.Platform.MSBuild/Tasks/InvokeTestingPlatformTask.cs
```diff
@@ -202,6 +242,32 @@ protected override string ToolName
     private bool IsCurrentProcessArchitectureCompatible() =>
         _currentProcessArchitecture == EnumPolyfill.Parse<Architecture>(TestArchitecture.ItemSpec, ignoreCase: true);
 
+    private string? TryGetRunCommand()
+    {
+        if (!RuntimeInformation.IsOSPlatform(OSPlatform.Windows))
+        {
+            // Currently fails on Linux and macOS.
+            // Not yet investigated.
```
Evangelink: Please add a ticket to the backlog.

Evangelink (dismissed):

src/Platform/Microsoft.Testing.Platform.MSBuild/buildMultiTargeting/Microsoft.Testing.Platform.MSBuild.targets
baronfel: Both are set by the .NET SDK in props files and so should be available. They _are_ _-prefixed properties, which means we don't like to encourage their direct use, but I don't think we make this data available in any other way at this time. cc @dsplaisted for heads-up that more folks are taking data dependencies on SDK-private members.

src/Platform/Microsoft.Testing.Platform.MSBuild/Tasks/InvokeTestingPlatformTask.cs
```diff
@@ -202,6 +242,32 @@ protected override string ToolName
...
+            return null;
+        }
+
+        // This condition specifically handles this part:
+        // https://github.com/dotnet/sdk/blob/5846d648f2280b54a54e481f55de4d9eea0e6a0e/src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Sdk.targets#L1152-L1155
+        // The more correct logic is implementing https://github.com/microsoft/testfx/issues/5091
+        if (IsNetCoreApp &&
```
nohwnd, line 253: Please add the comment about the reason that we saw in the app, where it looked at the current process to figure out the path, and then it failed to find a resource. 

Because when running exe we were in the current folder where the resource is, but if we run dotnet.exe exec theapp.dll it thinks we are in program files/dotnet and crashes.

src/Platform/Microsoft.Testing.Platform.MSBuild/Tasks/InvokeTestingPlatformTask.cs
Youssef1313: Removed that code.

src/Platform/Microsoft.Testing.Platform.MSBuild/Tasks/InvokeTestingPlatformTask.cs
Youssef1313: Done

Evangelink (approved):

PR #5095, opened 2025-02-21
Refactor F# compile order

Youssef1313:
I think it's a better idea to avoid having dependencies between the self-registered generation and the entry point generation. We let the target run on whatever order, and we still respect the compile order by using `CompileAfter` for the entry point.


Evangelink (approved):
It's better than today but I think it'd be nice if the self-registered extensions were before so that main (usually last file of the project) can use the regular `Compile`

PR #5124, opened 2025-02-25
Fix serializing special characters in Jsonite

nohwnd:


src/Platform/Microsoft.Testing.Platform/ServerMode/JsonRpc/Json/Jsonite/Jsonite.cs
```diff
@@ -973,9 +973,7 @@ private void WriteString(string text)
                             writer.Write('t');
                             break;
                         default:
-                            ArgumentGuard.Ensure(c >= ' ', nameof(text), $"Invalid control character '{EscapeChar(c)}' found in string");
-
-                            if (IsHighSurrogate(c) || IsLowSurrogate(c))
+                            if (c < ' ' || IsHighSurrogate(c) || IsLowSurrogate(c))
```
Youssef1313, line 979:
```suggestion
                            // Intentionally diverging from the upstream code from xoofx/jsonite. See https://github.com/microsoft/testfx/issues/5120

                            // Also, https://datatracker.ietf.org/doc/html/rfc4627#section-2.5
                            if (c < ' ' || IsHighSurrogate(c) || IsLowSurrogate(c))
```

Youssef1313 (approved):
LGTM

nohwnd:
> LGTM! It might be interesting to use a `DynamicData` source instead of the foreach so we can replay easily but not blocking on it.

This is addressed in comments. I did not want to emit the data, because then too many serializers need to handle them correctly to get the results. If any of them is broken (dynamic data serializer or the protocol serializer) we get no info about the test in UI, leaving us blind.


 // This could be converted to Data source, but this way we have more control about where in the result message the
 // special characters will be (hopefully nowhere) so in case of failure, we can still serialize the message to IDE
 // even if the serializer does not support special characters.

Evangelink:
> > LGTM! It might be interesting to use a `DynamicData` source instead of the foreach so we can replay easily but not blocking on it.
> 
> This is addressed in comments. I did not want to emit the data, because then too many serializers need to handle them correctly to get the results. If any of them is broken (dynamic data serializer or the protocol serializer) we get no info about the test in UI, leaving us blind.
> 
> // This could be converted to Data source, but this way we have more control about where in the result message the // special characters will be (hopefully nowhere) so in case of failure, we can still serialize the message to IDE // even if the serializer does not support special characters.

Didn't think about it but makes total sense! Thanks!

PR #5156, opened 2025-02-28
Improve ExecutionContext propagation

### LINKED ISSUE #4132: Huge performance degradation between 3.2 and 3.6.3
## Steps To Reproduce
I can't share our code base, but I did do some poking around with windbg:

It looks like all the test class instances are held onto permanently. 
Each of our test classes has member variables that hold several MB of data (This may be dubious practice, but it's how our tests been built)
If these instances are not garbage collected then memory keeps growing.

`ExecutionContextService.InstancesExecutionContexts`  in Microsoft.VisualStudio.TestPlatform.MSTestAdapter.PlatformServices seems to be the culprit. For whatever reason, none of the test instances are being cleaned up - I still have 1000s of test instances rooted.

## Expected behavior
Performance should not degrade

## Actual behavior
A set of tests that previously took 5 minutes now takes 1 hour.

## Additional context
Command used to run tests has a rather large test case filter, in case that is relevant
```
vstest.console test.dll 
/Testcasefilter:"(TestCategory!=A & TestCategory!=B & .. & TestCategory!=Z)" 
/Settings:.runsettings
/inIsolation 
/Logger:trx;LogFileName="Log.trx" 
/ResultsDirectory:"c:\temp"
```

### LINKED ISSUE #5154: CallContext.LogicalSetData not preserved
CallContext.LogicalSetData when set in AssemblyInit or ClassInit is not preserved and CallContext.LogicalGetData in a test method is getting null

## Steps To Reproduce

```xml
<Project Sdk="MSTest.Sdk/3.8.2">
  <PropertyGroup>
    <TargetFramework>net472</TargetFramework>
    <LangVersion>latest</LangVersion>
    <Nullable>enable</Nullable>
  </PropertyGroup>
</Project>
```

```csharp
using System.Runtime.Remoting.Messaging;

namespace TestProject1
{
    [TestClass]
    public sealed class Test1
    {
        [AssemblyInitialize]
        public static void AssemblyInit(TestContext context)
        {
            // This method is called once for the test assembly, before any tests are run.
            CallContext.LogicalSetData("AssemblyInit", "AssemblyInit");
        }

        [ClassInitialize]
        public static void ClassInit(TestContext context)
        {
            // This method is called once for the test class, before any tests of the class are run.
            CallContext.LogicalSetData("ClassInit", "ClassInit");
        }

        [TestInitialize]
        public void TestInit()
        {
            // This method is called before each test method.
            CallContext.LogicalSetData("TestInit", "TestInit");
        }

        [TestMethod]
        public void TestMethod()
        {
            Assert.AreEqual("TestInit", CallContext.LogicalGetData("TestInit"));
            Assert.AreEqual("ClassInit", CallContext.LogicalGetData("ClassInit"));
            Assert.AreEqual("AssemblyInit", CallContext.LogicalGetData("AssemblyInit"));
        }

        [TestMethod]
        public async Task TestMethodAsync()
        {
            Assert.AreEqual("TestInit", CallContext.LogicalGetData("TestInit"));
            Assert.AreEqual("ClassInit", CallContext.LogicalGetData("ClassInit"));
            Assert.AreEqual("AssemblyInit", CallContext.LogicalGetData("AssemblyInit"));
            await Task.FromResult(0);
        }
    }
}

```
## Expected behavior
Test passes
## Actual behavior
Test fails

Youssef1313:
Fixes #5154
Fixes #4132

### When not enabling parallelization

We execute tests in a loop, and at any step, async local changes are visible to all subsequent steps. (assuming no Timeout is involved, because that was previously a very different story).

By "step", I mean assembly initialize, class initialize, ctor, test initialize, test method, test cleanup, Dispose, and DisposeAsync.

### When enabling parallelization

We create N worker threads, each is dequeuing test cases from a common concurrent queue, and then running them.

All tests in each thread are run on a similar fashion as the "non-parallelizing" case. But when parallelization is enabled, a change to an async local at any step, won't be visible from the other thread. That means, a change to an async local in assembly initialize won't be visible to all tests. It will be visible to tests that run on the same worker thread though.

## New behavior from this PR

We have three levels. First is assembly, second is class, third is test.
We propagate from the highest level down, but not the other way around.

That means, a change of an async local in AssemblyInitialize is visible to all ClassInitialize methods regardless of which thread they are on. Similarly, a change in ClassInitialize is visible to all test methods (including initialize and cleanup). Also, a change from test initialize is visible to the test method itself, etc.

**However**, A change of an async local in a TestCleanup is NEVER seen in ClassCleanup. That's what I mean by propagating from highest level down, and not the other way around. This creates reasonable isolation.

This works as follows:

1. After assembly initialize, we capture the execution context and save it as the "assembly-level" context.
2. We execute class initialize on the assembly-level execution context.
3. After running class init, we capture the execution context and save it as the "class-level" context.
4. etc etc

ℹ️ The test methods are **duplicated**, because the first method will already have the correct context. So it would not fail. The second method will restore the context, and if we don't restore it correctly we will fail.


test/IntegrationTests/MSTest.Acceptance.IntegrationTests/GenericTestMethodTests.cs
```diff
@@ -38,18 +38,22 @@ at .+?
             failed ParameterizedMethodSimple \(null\) \(\d+ms\)
               Test method TestClass\.ParameterizedMethodSimple threw exception: 
               System\.InvalidOperationException: The type of the generic parameter 'T' could not be inferred\.
+                at .+?
             failed ParameterizedMethodTwoGenericParametersAndFourMethodParameters \(1,"Hello world",2,3\) \(\d+ms\)
```
nohwnd, line 42: what stack frame is added here?

src/Adapter/MSTestAdapter.PlatformServices/Services/ThreadSafeStringWriter.cs
```diff
@@ -181,6 +181,14 @@ private ThreadSafeStringBuilder GetOrAddStringBuilder()
         }
     }
 
+    internal static void CleanState()
```
nohwnd, line 184: I don't see this used anywhere?

nohwnd (commented):
seeing minimal changes in tests, is the change of how tests are isolated not happening in this PR, or we don't test it?

test/IntegrationTests/MSTest.Acceptance.IntegrationTests/GenericTestMethodTests.cs
Youssef1313: Just ExecutionContext.RunInternal.

src/Adapter/MSTestAdapter.PlatformServices/Services/ThreadSafeStringWriter.cs
Youssef1313: It's used in TestMethodInfo.Invoke to separate potential assembly init or class init from the methods.

Youssef1313:
> seeing minimal changes in tests, is the change of how tests are isolated not happening in this PR, or we don't test it?

There were some existing tests which I extended a bit.

test/IntegrationTests/MSTest.Acceptance.IntegrationTests/ThreadContextTests.cs
```diff
@@ -15,7 +15,7 @@ public async Task ThreadingContext_WhenCultureIsNotSet_TestMethodFails(string tf
     {
         var testHost = TestHost.LocateFrom(AssetFixture.InitToTestProjectPath, TestAssetFixture.InitToTestProjectName, tfm);
         TestHostResult testHostResult = await testHost.ExecuteAsync();
-        testHostResult.AssertOutputContainsSummary(failed: 1, passed: 0, skipped: 0);
+        testHostResult.AssertOutputContainsSummary(failed: 2, passed: 0, skipped: 0);
```
nohwnd, line 18: Test methods are duplicated because the first method has the correct context automatically, but the second method has to restore it. So the first one would not fail, but second one would.

nohwnd: Hence he higher test count.

PR #5161, opened 2025-03-01
Fix TestingPlatformSelfRegisteredExtensions to allow empty RootNamespace

### LINKED ISSUE #5160: Microsoft.Testing.Platform.MSBuild 1.6.x fails to build with RootNamespace error
## Describe the bug

After upgrading to xunit.v3 2.0.0, my test projects fail to build with this error:

> The "TestingPlatformSelfRegisteredExtensions" task was not given a value for the required parameter "RootNamespace".

This is coming from `Microsoft.Testing.Platform.MSBuild.targets(201,5).

Now, [my test project expressly sets RootNamespace to the empty string](https://github.com/AArnott/Nerdbank.MessagePack/blob/b4eae6d62a38c0a9f00827fe37c6abea2863e487/test/Nerdbank.MessagePack.Tests/Nerdbank.MessagePack.Tests.csproj#L8) because Test Explorer sometimes displays xunit tests with namespaces poorly. I've done this for years. But apparently M.T.P. considers that to be an error condition in its latest version.

## Steps To Reproduce

Create a test project that consumes Microsoft.Testing.Platform.MSBuild 1.6.2.
Set `<RootNamespace />` in the test project

## Expected behavior

The build succeeds.

## Actual behavior

The build fails

## Additional context

https://github.com/AArnott/Nerdbank.MessagePack/pull/286

Youssef1313:
Fixes #5160


src/Platform/Microsoft.Testing.Platform.MSBuild/Tasks/TestingPlatformEntryPointTask.cs
```diff
@@ -83,23 +82,23 @@ public override bool Execute()
...
         string entryPointSource = GetEntryPointSourceCode(language, rootNamespace);
         taskLoggingHelper.LogMessage(MessageImportance.Normal, $"Entrypoint source:\n'{entryPointSource}'");
         fileSystem.WriteAllText(testingPlatformEntryPointSourcePath.ItemSpec, entryPointSource);
     }
 
-    private static string GetEntryPointSourceCode(string language, string rootNamespace)
+    private static string GetEntryPointSourceCode(string language, string? rootNamespace)
     {
-        if (language != VBLanguageSymbol && !string.IsNullOrEmpty(rootNamespace))
+        if (language != VBLanguageSymbol && !RoslynString.IsNullOrEmpty(rootNamespace))
```
AArnott, line 94: How will this fix the original problem of disallowing an empty root namespace?

Youssef1313: @AArnott The real fix is removing `[Required]` attribute that was on line 53.
Other changes are side effect of marking the property nullable (as netstandard2.0 doesn't have the nullable annotations on IsNullOrEmpty).

PR #5164, opened 2025-03-03
remove redundant RootNamespace from TestingPlatformSelfRegisteredExtensions

### LINKED ISSUE #5163: Cant have a Project name the same as a file name. Eg Tests
this project https://github.com/SimonCropp/TestFxNamespaceConflict has

 * csproj named Tests
 * class named Tests
 * ref to Microsoft.Testing.Platform.MSBuild v 1.6.2

Wont compile with 

> Error CS0101 : The namespace '\<global namespace>' already contains a definition for 'Tests'

SimonCropp:
fixes #5163


Youssef1313 (changes requested):
This will cause other issues. Like preventing test projects to reference each others, which was causing issues for a good number of projects migrating from VSTest to MTP.

SimonCropp:
> Like preventing test projects to reference each others

that will already happen given root namespace is not required

SimonCropp:
can we use a random namespace?

Youssef1313:
> > Like preventing test projects to reference each others
> 
> that will already happen given root namespace is not required

RootNamespace will match project file name by default. If a user changes it and is having two test projects referencing each other and overwriting the root namespace so that it's the same, then for sure it will fail. But this is an edge case that we can do nothing about it. User has to fixup his projects.

---

> can we use a random namespace?

Nope. There are scenarios where the user will want to disable entry point generation but keep the extension generation one and call it manually. So it has to be deterministic for (at least) this case. But in general, having determinism in compilations is a goal.

SimonCropp:
can we add a suffix to the rootnamespace to reduce the chance of a conflict? eg appending `_TestFx`

SimonCropp:
> User has to fixup his projects

I am sure u mean *their* project

>  If a user changes it and is having two test projects referencing each other and overwriting the root namespace so that it's the same, then for sure it will fail. But this is an edge case that we can do nothing about it. User has to fixup his projects.

it is fine to say "User has to fixup" as long as it fails in a graceful way. And a namespace or type conflict with a code gen'd file, that is is difficult to find, is not graceful

Youssef1313:
> can we add a suffix to the rootnamespace to reduce the chance of a conflict? eg appending `_TestFx`

Technically possible, but I tend to think that `RootNamespace` is the right thing to use, without altering it. I also think that many components out there will use `RootNamespace` for generation. For example, IIRC, there is an Android target that generates resource-related code in RootNamespace, also [`ResxSourceGenerator`](https://www.nuget.org/packages/Microsoft.CodeAnalysis.ResxSourceGenerator) is likely to be using RootNamespace by default as well. So I don't see why MTP would be different here.

If absolutely necessary, we can introduce `TestingPlatformRootNamespace`, which defaults to `RootNamespace` if not specified by user. But in that case if the user will be willing to add `<TestingPlatformRootNamespace></TestingPlatformRootNamespace` in their project file, why wouldn't they set `RootNamespace` instead?

Youssef1313:
Yes, I understand that. Sorry if I wasn't clear. My concern here is that I'd like to avoid the default behavior to be inconsistent with what other components in the ecosystem generally do, and how RootNamespace is expected to be consumed in general. So, I was thinking to have the new behavior opt-in, and keep the default as-is. If we go that way, I'd prefer to use `Microsoft.Testing.Platform.` as prefix. Not `_Testfx` as suffix. I'm reaching out internally to .NET SDK team to see if they have some sort of guidance on how `RootNamespace` is expected to be used.

SimonCropp:
irrespective of recommendations on how RootNamespace  should be used, any injected code should do its best to avoid conflicting with non-injected code.

its why the c# compile injects il type names that can not be represented in c#. eg `<>z__ReadOnlyArray<T>`

SimonCropp:
also note that  `<>z__ReadOnlyArray<T>` has no namespace

Youssef1313:
Hi @SimonCropp
 
Sorry for taking long to get back to you. We have had internal discussions about this, and the way we do it now is the recommended way to handle this scenario. It is unfortunate that it prevents what you want to do. In retrospect, this change was somewhat breaking, and we'd like to avoid making another breaking change by inventing a new special way to handle this scenario.
 
We truly appreciate you taking the time to report the issue and contribute a fix, as well as all your other contributions, but we will not change this behavior at the moment.

PR #5171, opened 2025-03-04
simplify process code

SimonCropp:


src/Platform/Microsoft.Testing.Platform/Helpers/System/IProcessHandler.cs
```diff
@@ -5,9 +5,13 @@ namespace Microsoft.Testing.Platform.Helpers;
 
 internal interface IProcessHandler
 {
-    IProcess GetProcessById(int pid);
```
Youssef1313, line 8: Let's add the new needed APIs without removing the existing ones.

While it's internal, it's used by HangDump via IVT.

This means, if user updates core MTP but didn't update HangDump, it will fail with MissingMethodException. While this has happened multiple times in the past already, let's be careful to avoid it whenever possible.

nohwnd: Also hard Deprecate that old api? It should not fail on runtime, but should not build for new code.

src/Platform/Microsoft.Testing.Platform/Helpers/System/SystemProcessHandler.cs
```diff
@@ -8,10 +8,30 @@ namespace Microsoft.Testing.Platform.Helpers;
 [SuppressMessage("ApiDesign", "RS0030:Do not use banned APIs", Justification = "This is the Process wrapper.")]
 internal sealed class SystemProcessHandler : IProcessHandler
 {
-#pragma warning disable CA1416 // Validate platform compatibility
-    public IProcess GetCurrentProcess() => new SystemProcess(Process.GetCurrentProcess());
+    public async Task KillAsync(int id)
+    {
+        using var process = Process.GetProcessById(id);
+        process.Kill();
```
Youssef1313, line 14: ⚠️ Previously we were calling `Kill(true)` on .NET Core to kill child processes as well

PR #5180, opened 2025-03-06
Batching Method to handle acquiring the Console lock, executing delegates and then releasing the lock

thomhurst:
Alternatives proposal to #5179

This would avoid callers having to handle the acquire, try, finally, release logic. This helper method does it. Just pass a delegate of all the calls you want to batch.


Youssef1313 (commented):
I'm not opposed to the refactor, but note that it doesn't resolve the deadlock.

nohwnd (approved):

nohwnd:
actually won't merge. Let's wait for Youssefs code that actually solves the problem, maybe we can find some patterns to optimize there, and also avoid capturing state in lambdas.

Thanks for the proposal anyway!

Issue #5158, opened 2025-02-28
[MTP] Hanging on v1.6.2


thomhurst:
I've stuck on 1.4.3 for a while now because Rider had an issue where test discovery broke for any versions above this.

However I tried to upgrade to 1.6.2 yesterday and now I'm seeing hanging in Ubuntu and Mac. It happens the majority of the time, but also can be intermittent. Windows seems to always pass?

I've downgraded back to 1.4.3 and the hanging seems to have stopped again.

I collected an ubuntu hang dump file (I can't attach it here as it's too large) but here is the link to the artifact: https://github.com/thomhurst/TUnit/actions/runs/13595551691/artifacts/2672995550

There's also lots of output from the hangdump process that looks like machine instructions that I don't understand, not sure if it's useful to you, but you can see the summary here: https://github.com/thomhurst/TUnit/actions/runs/13595551691

MacOS doesn't produce me a proper hang dump. See #5157.

I'm not experienced with dump files, but I tried loading the ubuntu hangdump into WinDbg and just got an error (is this because it was generated on ubuntu?)

```
************* Preparing the environment for Debugger Extensions Gallery repositories **************
   ExtensionRepository : Implicit
   UseExperimentalFeatureForNugetShare : true
   AllowNugetExeUpdate : true
   NonInteractiveNuget : true
   AllowNugetMSCredentialProviderInstall : true
   AllowParallelInitializationOfLocalRepositories : true
   EnableRedirectToChakraJsProvider : false

   -- Configuring repositories
      ----> Repository : LocalInstalled, Enabled: true
      ----> Repository : UserExtensions, Enabled: true

>>>>>>>>>>>>> Preparing the environment for Debugger Extensions Gallery repositories completed, duration 0.000 seconds

************* Waiting for Debugger Extensions Gallery to Initialize **************

>>>>>>>>>>>>> Waiting for Debugger Extensions Gallery to Initialize completed, duration 0.844 seconds
   ----> Repository : UserExtensions, Enabled: true, Packages count: 0
   ----> Repository : LocalInstalled, Enabled: true, Packages count: 43

Microsoft (R) Windows Debugger Version 10.0.27793.1000 AMD64
Copyright (c) Microsoft Corporation. All rights reserved.


Loading Dump File [C:\Users\Tom\Downloads\HangDumpubuntu-latest\TUnit.Engine.Tests\bin\Release\net9.0\TestResults\hangdump.engine-tests.dmp]
64-bit machine not using 64-bit API

[...]

PR #5182, opened 2025-03-07
Fix possible deadlock when writing to console

Youssef1313:
Hopefully fixes https://github.com/microsoft/testfx/issues/5158

### DIFF
--- a/src/Platform/Microsoft.Testing.Platform/OutputDevice/Terminal/NonAnsiTerminal.cs
+++ b/src/Platform/Microsoft.Testing.Platform/OutputDevice/Terminal/NonAnsiTerminal.cs
@@ -112,6 +112,26 @@ internal sealed class NonAnsiTerminal : ITerminal
         // Console methods will internally lock on Console.Out, so we are locking on the same thing.
         // This locking is the easiest way to get coloring to work correctly while preventing
         // interleaving with user's calls to Console.Write methods.
+        // One extra note:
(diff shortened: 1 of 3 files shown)


src/Platform/Microsoft.Testing.Platform/OutputDevice/Terminal/NonAnsiTerminal.cs
```diff
@@ -90,8 +91,11 @@ public void StartUpdate()
         }
 
         bool lockTaken = false;
-        // SystemConsole.ConsoleOut is set only once in static ctor.
-        // So we are sure we will be doing Monitor.Exit on the same instance.
+
+        // We store Console.Out in a field to make sure we will be doing
+        // the Monitor.Exit call on the same instance.
+        _batchingLock = Console.Out;
```
nohwnd, line 97: as you described https://source.dot.net/#System.Console/System/ConsolePal.Unix.cs,898f4da5a113560c console pal on unix is locking on console.out (so on the object that someone - like TUnit substituted), it would be nice to add that reasoning somewhere so we know what are the rules we need to follow when locking on output. To prevent deadlock and preventing mixed output from our internal services (that use our SystemConsole class) and test framework or other test code that substituted Console.Out.

And also that we know we can stop using that pattern if the underlying implementation changes.

nohwnd:
approved, would be nice to fix the ignored tests later.

PR #5185, opened 2025-03-07
Remove dead code in TerminalTestReporter

Youssef1313:


nohwnd (approved):
this was done for the dotnet test output, but since those two implementations diverged it is indeed not needed anymore.

Issue #3559, opened 2024-08-15
Lookup test configuration from a file


nohwnd:
## Summary

There are many configuration options in runsettings/other files. It would be nice to not have to specify them in commandline, or in VS in UI.

## Background and Motivation

It is hard to recommend people to configure their test base in some non-default way when they don't have an easy way to consume such settings. For example running on commandline with settings requires to always specify them: 

```
dotnet test --settings mysettings.runsettings
```

and in VS they need to be set in UI by every developer.

## Proposed Feature

Lookup settings / configuration file by recursing directories, similarly to `Directory.Build.props`, `nuget.config` or `.editorconfig`, to allow using settings by default without having to specify them on commandline or in UI. 

Offload the lookup to MSBuild when building the test project, and copy the settings next to the test project so we keep the bin folder self-contained and portable.

nohwnd:
When is this useful? 

For example I would like my tests to run with output that shows passed tests, and never uses ansi. I would like to be able to set those settings in my repo, and version them and have them automatically picked up for any test run.

Evangelink:
Thanks @nohwnd! 

I find it useful indeed but I think I would prefer this behavior to not be part of the core platform so it remains 100% deterministic. 

I am wondering if this could be an added feature of VS and dotnet test (could be some kind of shared library or an extension) that does the lookup and force the `--settings path` to the call. This means we would have a different UX when calling the exe directly but I think it's fine.

Evangelink:
Actually as runsettings are coming from the bridge, it's probably find to say the lookup feature is provided by the bridge.

nohwnd:
It does not have to be runsettings specifically, I would much rather have it built in for some configuration that the platform is able to consume. I don't think runsettings should be used in the future as the way to configure test runs.

Evangelink:
>I would much rather have it built in for some configuration that the platform

See related ticket.

thomhurst:
Is this the same as #3499 ?

Evangelink:
Partially, this ticket is about how to find the configuration file (similarly to #3562). Yours is asking about the support of command line options in the config file.

BTW you could already have a look at https://github.com/microsoft/testfx/blob/main/docs/testingplatform/configuration.md to start adding your settings in this file. We will be releasing 1.4 preview today (please use this version if you want to do so).

Evangelink:
Closing as won't fix for now

Issue #3499, opened 2024-08-08
[Testing Platform] Command line arguments file source


thomhurst:
From what I can tell options are only supported currently from the command line.

It might be useful to try to find a well-known file that can store options within it. This way it can be source controlled by teams, and provide a more deterministic result between local development and CI builds without developers having to know which various flags to set.

E.g.

`test-options.json`
```json
{
    "minimum-expected-tests": 200,
    "coverage": true,
    "report-trx": true
}
```

Evangelink:
We have it partially done but that does need to be finished and documented (cc @MarcoRossignoli).

MarcoRossignoli:
> We have it partially done but that does need to be finished and documented (cc @MarcoRossignoli).

Actually is not yet implemented but it's doable and a good idea, maybe a section like

```json
{
    "testingPlatform" :
    {
        "additionalCommandLine" : "--someoption file --anotherone"
    }
}
```
That we can parse and add, we need to understand what to do in case of duplicates, override, merge, throw.

thomhurst:
I'd probably take the file as the base values, and then any actual command line options override them

nohwnd:
> I'd probably take the file as the base values, and then any actual command line options override them

👍 This is how the current vstest works. It takes config file, then answer file, then commandline arguments, then inline settings (settings specified on commandline after ` -- `).

One great addition to that would imho be also applying settings file over a settings file, so we can have slightly different settings in CI, similar to my.configy.json, my.**dev**.config.json

e.g. --settings my-settings.json --settings my-settings.ci.json

>  maybe a section like additionalCommandLine

@MarcoRossignoli  I think there is a good concept hidden in there, IMHO the settings and commandline parameters should be closely related and there should be a definitive step where all settings are finalized and are applied back to a "runsettings" object. This object is then the source of truth for the local run, or a remote run.

In less abstract terms we should be able to write a plugin / execution mode that takes given command line parameters and turns them into runsettings.

MarcoRossignoli:
> In less abstract terms we should be able to write a plugin / execution mode that takes given command line parameters and turns them into runsettings.

My main problem are complex configuration, in command line you cannot express too much and the configuration file in future could change the format like toml. One thing that we can do is maybe override the config with a convention "--overrideconfig-mstest-parallelize true" for scalar value
```json
{
...
 "mstest":
 {
    "parallelize" : false  
 }
}
```

here we have only strings without type safety.

thomhurst:
[...]

PR #5201, opened 2025-03-10
Consider `--results-directory` before configuration

### LINKED ISSUE #5189: Results directory from configuration is preferred over command-line option `--results-directory`
To me, it seems like `--results-directory` should win over configuration files.

Relevant code:

https://github.com/microsoft/testfx/blob/e5345132464e5b6ab26cd6fc2c6f8f44854d548e/src/Platform/Microsoft.Testing.Platform/Configurations/AggregatedConfiguration.cs#L66-L77

This can cause issues for retry extension. If user is originally calling the test executable passing a runsettings files that contains ResultsDirectory entry. Then, retry extension will re-pass the runsettings file [along with `--results-directory` pointing to the attempt directory](https://github.com/microsoft/testfx/blob/e5345132464e5b6ab26cd6fc2c6f8f44854d548e/src/Platform/Microsoft.Testing.Extensions.Retry/RetryOrchestrator.cs#L159-L160). The new run is now using the same ResultsDirectory and not using the attempt folder.

**CONSIDER:** Producing a warning if ResultsDirectory is specified twice (only if they are different? or always?).

AB#2415008

Youssef1313:
Fixes #5189

### DIFF
--- a/src/Platform/Microsoft.Testing.Platform/Configurations/ConfigurationExtensions.cs
+++ b/src/Platform/Microsoft.Testing.Platform/Configurations/ConfigurationExtensions.cs
@@ -35,6 +35,7 @@ public static class ConfigurationExtensions
     /// </summary>
     /// <param name="configuration">The configuration.</param>
     /// <returns>The test host working directory.</returns>
+    [Obsolete("This API is obsolete and will be removed in v2. Use GetCurrentWorkingDirectory instead.", error: true)]
     // TODO: This is only used in unit tests. Should that be removed (maybe in v2)?
     public static string GetTestHostWorkingDirectory(this IConfiguration configuration)
     {
--- a/src/Platform/Microsoft.Testing.Platform/Configurations/ConfigurationExtensions.cs
+++ b/src/Platform/Microsoft.Testing.Platform/Configurations/ConfigurationExtensions.cs
@@ -35,6 +35,7 @@ public static class ConfigurationExtensions
     /// </summary>
     /// <param name="configuration">The configuration.</param>
     /// <returns>The test host working directory.</returns>
+    // TODO: This is only used in unit tests. Should that be removed (maybe in v2)?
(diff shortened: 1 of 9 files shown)


Evangelink (commented):
@MarcoRossignoli I recall we discussed that and I thought we ended up with:
- Env var wins over
- CLI args that wins over
- config file

Do I recall incorrectly?

Youssef1313:
> @MarcoRossignoli I recall we discussed that and I thought we ended up with:
> 
> * Env var wins over
> * CLI args that wins over
> * config file
> 
> Do I recall incorrectly?

Currently, config file is winning over CLI

nohwnd:
Yes, env > cli > config file

So file config can be committed and implicit (https://github.com/microsoft/testfx/issues/3559), overriden by commandline parameters, either from cli or script or from ci yml, 
and then env variables, so they can be used specifically in CI to setup secrets / pipeline overrides.

we also discuss it here: https://github.com/microsoft/testfx/issues/3499#issuecomment-2295876429

src/Platform/Microsoft.Testing.Platform/Configurations/ConfigurationExtensions.cs
```diff
@@ -35,6 +35,7 @@ public static string GetCurrentWorkingDirectory(this IConfiguration configuratio
     /// </summary>
     /// <param name="configuration">The configuration.</param>
     /// <returns>The test host working directory.</returns>
+    // TODO: This is only used in unit tests. Should that be removed (maybe in v2)?
```
nohwnd, line 39: in that case deprecate it?

Youssef1313: Sure.

Youssef1313: Marked as obsolete and cleaned up usages.

nohwnd (approved):

PR #5220, opened 2025-03-12
Ship props/targets of MSTest.TestFramework and MSTest.TestAdapter in both build and buildTransitive

### LINKED ISSUE #5213: Add props/targets to both build and buildTransitive rather than buildTransitive only
> For maximum compatibility, a package can have both `build/` and `buildTransitive/` since `packages.config` only supports `build/`. There are some project types that do not support `<PackageReference />` yet so it can be tricky to know if a particular component can have just `buildTransitive/`
> 
> @nkolev92 might have some guidance as well. 

 _Originally posted by @jeffkl in [#4736](https://github.com/microsoft/testfx/issues/4736#issuecomment-2715379183)_

Youssef1313:
Fixes #5213


nohwnd (dismissed):
double checked all the tfms are aligned.

PR #5235, opened 2025-03-14
Add assembly name

### LINKED ISSUE #5086: `MSBuildResources.Summary` should include assembly name at the end for easier interpretation

nohwnd:
Fix #5086

### DIFF
--- a/eng/Versions.props
+++ b/eng/Versions.props
@@ -1,9 +1,9 @@
 <Project>
   <PropertyGroup Label="Version settings">
     <!-- MSTest version -->
-    <VersionPrefix>3.9.0</VersionPrefix>
+    <VersionPrefix>3.10.0</VersionPrefix>
     <!-- Testing Platform version -->
-    <TestingPlatformVersionPrefix>1.7.0</TestingPlatformVersionPrefix>
+    <TestingPlatformVersionPrefix>1.8.0</TestingPlatformVersionPrefix>
     <PreReleaseVersionLabel>preview</PreReleaseVersionLabel>
   </PropertyGroup>
   <PropertyGroup Label="MSTest prod dependencies - darc updated">
--- a/eng/Versions.props
+++ b/eng/Versions.props
@@ -1,9 +1,9 @@
 <Project>
   <PropertyGroup Label="Version settings">
     <!-- MSTest version -->
-    <VersionPrefix>3.10.0</VersionPrefix>
+    <VersionPrefix>3.9.0</VersionPrefix>
     <!-- Testing Platform version -->
-    <TestingPlatformVersionPrefix>1.8.0</TestingPlatformVersionPrefix>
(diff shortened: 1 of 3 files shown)


eng/Versions.props
```diff
@@ -1,9 +1,9 @@
 <Project>
   <PropertyGroup Label="Version settings">
     <!-- MSTest version -->
-    <VersionPrefix>3.9.0</VersionPrefix>
+    <VersionPrefix>3.10.0</VersionPrefix>
```
nohwnd: will revert after, I think this PR will fail integration tests, so I don't want to revert immediately.

src/Platform/Microsoft.Testing.Platform.MSBuild/Tasks/InvokeTestingPlatformTask.cs
```diff
@@ -13,6 +13,7 @@
 using Microsoft.Testing.Platform.IPC.Serializers;
 using Microsoft.Testing.Platform.Logging;
 using Microsoft.Testing.Platform.MSBuild.Tasks;
+using Microsoft.Testing.Platform.OutputDevice;
```
nohwnd, line 16: target framework parser would be nice to move to some other place. or publish, I saw the same need for simplifying the tfm e.g. in the gitub logger.

Won't do here though.

src/Platform/Microsoft.Testing.Platform.MSBuild/Tasks/InvokeTestingPlatformTask.cs
```diff
@@ -488,6 +489,11 @@ private Task<IResponse> HandleRequestAsync(IRequest request)
                 runSummaryInfoRequest.Total,
                 runSummaryInfoRequest.Duration);
 
+            if (_moduleInfo != null)
+            {
+                summary += $" - {TargetPath.ItemSpec} ({TargetFrameworkParser.GetShortTargetFramework(TargetFramework.ItemSpec)}|{TestArchitecture.ItemSpec})";
```
Youssef1313: I think it will display nicer as just the assembly name?

src/Platform/Microsoft.Testing.Platform.MSBuild/Tasks/InvokeTestingPlatformTask.cs
```diff
@@ -488,6 +489,11 @@ private Task<IResponse> HandleRequestAsync(IRequest request)
                 runSummaryInfoRequest.Total,
                 runSummaryInfoRequest.Duration);
 
+            if (_moduleInfo != null)
```
Youssef1313: Can it be null? 😮

nohwnd: probably but I did not end up using that field.

Youssef1313 (approved):

PR #5236, opened 2025-03-14
Fix TestFailedException outcome not propagating to TestResult outcome

### LINKED ISSUE #5226: MSTest - ClassInitialize - Assert.Inconclusive fails the workflow on MSTest 3.8.2, but not on 3.7.3
## Describe the bug

After upgrading MSTest from 3.7.3 to 3.8.2, unit tests start failing. Inside the ClassInitialize, we use the Assert.Inconclusive() if a certain environment variable / user secret is not present (cause not everything can necessarily run on GitHub). This was working nicely, but after updating MSTest to 3.8.2, this started failing the unit tests instead of marking them as skipped like before.

## Steps To Reproduce

See the following workflow run: [https://github.com/MichielOda/MSTestClassInitializeInconclusive/actions/runs/13831330210](https://github.com/MichielOda/MSTestClassInitializeInconclusive/actions/runs/13831330210)

## Expected behavior
After changing to MSTest 3.8.2, the tests keep on working like normal.

## Actual behavior
The Inconclusive in the ClassInitialize is failing the tests instead of marking them as skipped.

## Additional context

Full test repo: [https://github.com/MichielOda/MSTestClassInitializeInconclusive](https://github.com/MichielOda/MSTestClassInitializeInconclusive)

I went over the changes of the 3.8.x range, but didn't saw something immediately that could be the cause of this issue. Of course, I could have missed it due to terminology that I'm missing.

Youssef1313:
Fixes #5226

Looks like it could be a regression from #4686

### DIFF
--- a/test/IntegrationTests/MSTest.Acceptance.IntegrationTests/InconclusiveTests.cs
+++ b/test/IntegrationTests/MSTest.Acceptance.IntegrationTests/InconclusiveTests.cs
@@ -1,0 +1,235 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using Microsoft.Testing.Platform.Acceptance.IntegrationTests;
+using Microsoft.Testing.Platform.Acceptance.IntegrationTests.Helpers;
+using Microsoft.Testing.Platform.Helpers;
+
+namespace MSTest.Acceptance.IntegrationTests;
+
+[TestClass]
+public sealed class InconclusiveTests : AcceptanceTestBase<InconclusiveTests.TestAssetFixture>
+{
+    public enum Lifecycle
+    {
+        AssemblyInitialize,
+        ClassInitialize,
+        TestInitialize,
+        TestMethod,
+        TestCleanup,
+        ClassCleanup,
+        AssemblyCleanup,
+    }
+
+    [TestMethod]
+    [DataRow(Lifecycle.AssemblyInitialize)]
+    [DataRow(Lifecycle.ClassInitialize)]
+    [DataRow(Lifecycle.TestInitialize)]
+    [DataRow(Lifecycle.TestMethod)]
+    [DataRow(Lifecycle.TestCleanup)]
+    [DataRow(Lifecycle.ClassCleanup)]
+    [DataRow(Lifecycle.AssemblyCleanup)]
+    public async Task TestOutcomeShouldBeRespectedCorrectly(Lifecycle inconclusiveStep)
+    {
+        var testHost = TestHost.LocateFrom(AssetFixture.ProjectPath, TestAssetFixture.ProjectName, TargetFrameworks.NetCurrent);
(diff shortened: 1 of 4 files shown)


test/IntegrationTests/MSTest.Acceptance.IntegrationTests/InconclusiveTests.cs
```diff
@@ -0,0 +1,235 @@
...
+
+    [TestMethod]
+    [DataRow(Lifecycle.AssemblyInitialize)]
+    [DataRow(Lifecycle.ClassInitialize)]
+    [DataRow(Lifecycle.TestInitialize)]
+    [DataRow(Lifecycle.TestMethod)]
+    [DataRow(Lifecycle.TestCleanup)]
+    [DataRow(Lifecycle.ClassCleanup)]
+    [DataRow(Lifecycle.AssemblyCleanup)]
+    public async Task TestOutcomeShouldBeRespectedCorrectly(Lifecycle inconclusiveStep)
```
Evangelink, line 32: It'd be nice to have the "values source" attribute for such case

Evangelink, line 32: Fine as-is but I am wondering if a different test for each case would not be clearer for people reviewing in the future.

test/UnitTests/MSTestAdapter.UnitTests/Execution/TestClassInfoTests.cs
```diff
@@ -207,7 +207,7 @@ public void RunClassInitializeShouldThrowIfTestContextIsNull()
         TestResult result = GetResultOrRunClassInitialize(null);
         var exception = result.TestFailureException as TestFailedException;
         Verify(exception is not null);
-        Verify(result.Outcome == UTF.UnitTestOutcome.Failed);
+        Verify(result.Outcome == UTF.UnitTestOutcome.Error);
```
Evangelink, line 210: It's hard to understand if the change here makes sense or not so I'll approve but please have a quick look.

Evangelink (approved):

test/UnitTests/MSTestAdapter.UnitTests/Execution/TestClassInfoTests.cs
Youssef1313: The unit test passes `null` as TestContext which is something that never happens in production actually. Even if it happens, I think `Error` makes better sense as it's more of MSTest error rather than a "failing" test on user side.

test/IntegrationTests/MSTest.Acceptance.IntegrationTests/InconclusiveTests.cs
Youssef1313: I'll merge as we discussed.

Youssef1313: I'll merge as we discussed.

PR #5260, opened 2025-03-18
Add Azure DevOps extension to report errors

nohwnd:
Azure DevOps extension to report errors in a way that is understood by Azure DevOps user interface.

image

The failures are also be transported to GitHub, so we can see errors directly in PRs.

image

image

User enables the extension by `--report-azdo`, and can reduce the severity of reports to warning by `--report-azdo-severity warning` (this prevents the report to show up in GitHub!).

This is not a replacement for https://github.com/Tyrrrz/GitHubActionsTestLogger, because that extension is working directly with GitHubActions. 

This extension is working with Azure DevOps, and if the standard Azure DevOps integration to GitHub is used it forwards the failures.


src/Platform/Microsoft.Testing.Extensions.AzureDevOps/AzDoEscaper.cs
```diff
@@ -0,0 +1,39 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using Microsoft.Testing.Platform;
+
+namespace Microsoft.Testing.Extensions.TrxReport.Abstractions;
```
Youssef1313: nit: namespace needs to be fixed?

src/Platform/Microsoft.Testing.Extensions.AzureDevOps/AzureDevOpsExtensions.cs
```diff
@@ -0,0 +1,34 @@
...
+    /// </summary>
+    /// <param name="builder">The test application builder.</param>
+    public static void AddAzureDevOpsProvider(this ITestApplicationBuilder builder)
+    {
+        var compositeTestSessionAzDoService =
+           new CompositeExtensionFactory<AzureDevOpsReporter>(serviceProvider =>
+               new AzureDevOpsReporter(
+               serviceProvider.GetCommandLineOptions(),
+               serviceProvider.GetEnvironment(),
+               serviceProvider.GetOutputDevice()));
```
Youssef1313: nit: indentation

src/Platform/Microsoft.Testing.Extensions.AzureDevOps/AzureDevOpsReporter.cs
```diff
@@ -0,0 +1,177 @@
...
+
+        if (isEnabled)
+        {
+            bool found = _commandLine.TryGetOptionArgumentList(AzureDevOpsCommandLineOptions.AzureDevOpsReportSeverity, out string[]? arguments);
+            if (found && arguments?.Length > 0)
+            {
+                if (string.Equals(arguments[0], "warning", StringComparison.OrdinalIgnoreCase))
+                {
+                    _severity = "warning";
+                }
```
Youssef1313: Why only warning is handled here? Should we simplify to use `arguments[0]`, and even require the command-line argument to be provided all lowercase? (i.e, the validation logic in the command line options provider becomes case sensitive)

src/Platform/Microsoft.Testing.Extensions.AzureDevOps/AzureDevOpsReporter.cs
```diff
@@ -0,0 +1,177 @@
...
+        }
+
+        return Task.FromResult(isEnabled);
+    }
+
+    public async Task ConsumeAsync(IDataProducer dataProducer, IData value, CancellationToken cancellationToken)
+    {
+        if (cancellationToken.IsCancellationRequested)
+        {
+            return;
```
Youssef1313, line 80: While this is a pattern we are following in other places, I don't feel like it's correct. I think we should call `ThrowIfCancellationIsRequested`. But whether or not the cancellation exception is handled correctly for now is something I don't really know.

src/Platform/Microsoft.Testing.Extensions.AzureDevOps/AzureDevOpsReporter.cs
```diff
@@ -0,0 +1,177 @@
...
+                case ErrorTestNodeStateProperty error:
+                    await WriteExceptionAsync(error.Explanation, error.Exception);
+                    break;
+                case CancelledTestNodeStateProperty cancelled:
+                    await WriteExceptionAsync(cancelled.Explanation, cancelled.Exception);
+                    break;
+                case TimeoutTestNodeStateProperty timeout:
+                    await WriteExceptionAsync(timeout.Explanation, timeout.Exception);
+                    break;
+            }
```
Youssef1313: I think we may be doing this somewhere else. Maybe time to extract an internal helper?

src/Platform/Microsoft.Testing.Extensions.AzureDevOps/AzureDevOpsReporter.cs
```diff
@@ -0,0 +1,177 @@
...
+
+            return;
+        }
+        catch (OperationCanceledException ex) when (ex.CancellationToken == cancellationToken)
+        {
+            // Do nothing, we're stopping
+        }
```
Youssef1313: Is this catch really reachable? Nothing in the "try" uses the token.

src/Platform/Microsoft.Testing.Extensions.AzureDevOps/AzureDevOpsReporter.cs
```diff
@@ -0,0 +1,177 @@
...
+
+        return;
```
Youssef1313, line 106: Redundant return?

src/Platform/Microsoft.Testing.Extensions.AzureDevOps/AzureDevOpsReporter.cs
```diff
@@ -0,0 +1,177 @@
...
+        }
+
+        string message = explanation ?? exception.Message;
+
+        if (message == null)
+        {
+            return;
+        }
+
+        string stackTrace = exception.StackTrace;
```
Youssef1313, line 123: This seems to be relying on MSTest implementation detail where we trim stack traces

src/Platform/Microsoft.Testing.Extensions.AzureDevOps/Microsoft.Testing.Extensions.AzureDevOps.csproj
```diff
@@ -0,0 +1,52 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFrameworks>netstandard2.0;$(MicrosoftTestingTargetFrameworks)</TargetFrameworks>
+    <RootNamespace>Microsoft.Testing.Extensions.Reporting</RootNamespace>
```
Youssef1313: Any reason we need to set RootNamespace this way?

src/Platform/Microsoft.Testing.Extensions.AzureDevOps/Microsoft.Testing.Extensions.AzureDevOps.csproj
```diff
@@ -0,0 +1,52 @@
...
+  </PropertyGroup>
+
+  <!-- NuGet properties -->
```
Youssef1313, line 11: Do we want to ship it as stable? Or do we need to first wait until we are trying it ourselves for a while?

src/Platform/Microsoft.Testing.Extensions.AzureDevOps/RootFinder.cs
```diff
@@ -0,0 +1,34 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+namespace Microsoft.Testing.Extensions.TrxReport.Abstractions;
```
Youssef1313: nit: Fix namespace

src/Platform/Microsoft.Testing.Extensions.AzureDevOps/Microsoft.Testing.Extensions.AzureDevOps.csproj
```diff
@@ -0,0 +1,52 @@
...
+    <AdditionalFiles Include="PublicAPI/PublicAPI.Unshipped.txt" />
+  </ItemGroup>
+
+  <ItemGroup>
+    <ProjectReference Include="$(RepoRoot)src\Platform\Microsoft.Testing.Platform\Microsoft.Testing.Platform.csproj" />
+  </ItemGroup>
+
+  <ItemGroup>
+    <InternalsVisibleTo Include="Microsoft.Testing.Extensions.VSTestBridge" Key="$(VsPublicKey)" />
+    <InternalsVisibleTo Include="Microsoft.Testing.Extensions.HangDump" Key="$(VsPublicKey)" />
```
Youssef1313: Hopefully redundant?

src/Platform/Microsoft.Testing.Extensions.AzureDevOps/AzureDevOpsReporter.cs
nohwnd: yeah those are the cases I wanted to avoid. I don't know how case sensitive is the azdo syntax. And did not want to fail the option parsing just because user says Warning instead of warning. I could also lowercase the arguments[0].

Only warning is handled here because "error" is the default. I can change this.

src/Platform/Microsoft.Testing.Extensions.AzureDevOps/AzureDevOpsReporter.cs
nohwnd: you mean where we trim Assertion internal stack frames from the stack trace? in that case yup. But I am not sure what to do with that for other frameworks.

src/Platform/Microsoft.Testing.Extensions.AzureDevOps/Microsoft.Testing.Extensions.AzureDevOps.csproj
nohwnd: It was like that in crashdump from where I copied it, will try to remote.🤷‍♀️

src/Platform/Microsoft.Testing.Extensions.AzureDevOps/Microsoft.Testing.Extensions.AzureDevOps.csproj
nohwnd: ok will make it alpha

src/Platform/Microsoft.Testing.Extensions.AzureDevOps/AzureDevOpsReporter.cs
Youssef1313: I'd personally prefer to start with making the option case sensitive and require the user to have it lowercase. We can adjust later if many users complain.

src/Platform/Microsoft.Testing.Extensions.AzureDevOps/AzureDevOpsReporter.cs
Youssef1313: Maybe the first stack frame that points to an existing file?

src/Platform/Microsoft.Testing.Extensions.AzureDevOps/AzureDevOpsReporter.cs
nohwnd: good point, I think you made a separate issue for this, let's solve it there. I don't want to introduce unrelated changes here.

src/Platform/Microsoft.Testing.Extensions.AzureDevOps/AzureDevOpsReporter.cs
nohwnd: There already is TestNodePropertiesCategories.WellKnownTestNodeTestRunOutcomeFailedProperties but that cannot touch the node state, and it is not practical to implement the helper in its terms, because once we have a switch, it is extra work to check if the TestNodePropertiesCategories.WellKnownTestNodeTestRunOutcomeFailedProperties array has the state value.

And I would like those two to not become out of sync.

Moving to new issue.

nohwnd: https://github.com/microsoft/testfx/issues/5266

src/Platform/Microsoft.Testing.Extensions.AzureDevOps/AzureDevOpsReporter.cs
nohwnd: that sounds expensive, maybe first stack frame that points to a path that starts with the repo root?

nohwnd: (assuming users would provide that externally) That could work even if we move the dlls to a separate server, ~but it feels like we are inventing a way to detect what is user code, and imho the test framework is best suited for that, and all frameworks should provide filtered call stack from the assertions.~

I am missing the point in what I wrote above, the extension needs to point at code files, no matter where the exception occurred, so detecting the first stack frame that belongs to the repo path is the best imho.

src/Platform/Microsoft.Testing.Extensions.AzureDevOps/AzureDevOpsReporter.cs
nohwnd: that is not a good UX to me, (enum) options should be case insensitive. Changed it to keep the case insensitive option, and I just lowercase the user provided value (which can be just error or warning).

src/Platform/Microsoft.Testing.Extensions.AzureDevOps/AzureDevOpsReporter.cs
nohwnd: https://github.com/microsoft/testfx/issues/5272 moving here

nohwnd:
Made issues for improving how we find interesting stack frame. And making it part of all extensions. Let's dogfood now.

PR #5264, opened 2025-03-19
Use FileAccess.Read when reading testconfig.json file

### LINKED ISSUE #5257: [MTP] Windows - testconfig.json in use if invoking test suite multiple times concurrently
Feel free to close if you don't deem this an issue. But I didn't get this issue on 1.4.3. I've just upgraded TUnit to 1.6.3 and am seeing this on Windows builds. Mac and Linux are fine. I believe Windows locks files differently to Unix so that's why.

With TUnit, I have a set of "Engine" tests and assert that I get back the correct passed/failed/skipped etc. for certain scenarios.

I do this by using the treenode filter and targeting specific classes etc. and these will run concurrently. 

So my tests actually spawn new `dotnet` processes.

I'm often seeing:

`Error: Unhandled exception. System.IO.IOException: The process cannot access the file 'D:\a\TUnit\TUnit\TUnit.TestProject\bin\Release\net9.0\TUnit.TestProject.testconfig.json' because it is being used by another process.`

It'd be nice if we could somehow avoid the lock (not sure if that's possible or not?) - Or if not maybe a retry mechanism.

Youssef1313:
Potentially fixes #5257

### DIFF
--- a/test/UnitTests/Microsoft.Testing.Platform.UnitTests/Configuration/ConfigurationManagerTests.cs
+++ b/test/UnitTests/Microsoft.Testing.Platform.UnitTests/Configuration/ConfigurationManagerTests.cs
@@ -64,7 +64,7 @@ public sealed class ConfigurationManagerTests
         ConfigurationManager configurationManager = new(fileSystem.Object, testApplicationModuleInfo);
         configurationManager.AddConfigurationSource(() =>
             new JsonConfigurationSource(testApplicationModuleInfo, fileSystem.Object, null));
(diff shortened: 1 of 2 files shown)


test/UnitTests/Microsoft.Testing.Platform.UnitTests/Configuration/ConfigurationManagerTests.cs
```diff
@@ -1,6 +1,8 @@
 ﻿// Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
+using System.Text.Json;
```
Youssef1313, line 5:
```suggestion
#if !NETFRAMEWORK
using System.Text.Json;
#endif
```

PR #5267, opened 2025-03-19
Unblock run when cancellation was requested 

nohwnd:
Ensure that we can complete the run when non-graceful cancellation is requested (e.b. ctrl+C on commandline), by making sure that the test runner does not wait for all currently running tests to finish.

### DIFF
--- a/src/Adapter/MSTest.TestAdapter/Execution/TestExecutionManager.cs
+++ b/src/Adapter/MSTest.TestAdapter/Execution/TestExecutionManager.cs
@@ -471,7 +471,11 @@ public class TestExecutionManager
 
                 try
                 {
-                    await Task.WhenAll(tasks);
+                    // Ensure we can abandon currently running tasks on cancellation, rather than waiting for them to complete.
+                    // They will still run on background but we won't await them.
+                    var runCancelled = new TaskCompletionSource<object>();
+                    _testRunCancellationToken?.Register(() => runCancelled.TrySetCanceled());
+                    await Task.WhenAny(Task.WhenAll(tasks), runCancelled.Task);
                 }
                 catch (Exception ex)
                 {


src/Adapter/MSTest.TestAdapter/Execution/TestExecutionManager.cs
```diff
@@ -471,7 +471,11 @@ private async Task ExecuteTestsInSourceAsync(IEnumerable<TestCase> tests, IRunCo
 
                 try
                 {
-                    await Task.WhenAll(tasks);
+                    // Ensure we can abandon currently running tasks on cancellation, rather than waiting for them to complete.
+                    // They will still run on background but we won't await them.
+                    var runCancelled = new TaskCompletionSource<object>();
+                    _testRunCancellationToken?.Register(() => runCancelled.TrySetCanceled());
+                    await Task.WhenAny(Task.WhenAll(tasks), runCancelled.Task);
```
nohwnd: Cancellation token _testRunCancellationToken is not a CancellationToken, so we cannot use `Task.WhenAll(tasks, ct)` here.

test/IntegrationTests/MSTest.Acceptance.IntegrationTests/AbortionTests.cs
```diff
@@ -25,17 +37,45 @@ public async Task AbortWithCTRLPlusC_CancellingTests(string tfm)
...
+                    <Scope>MethodLevel</Scope>
+                </Parallelize>
+                </MSTest>
+            </RunSettings>
+            """);
+            parameters = $"--settings {runSettingsPath}";
+        }
+
         string fileCreationPath = Path.Combine(testHost.DirectoryName, "fileCreation");
-        File.WriteAllText(fileCreationPath, string.Empty);
```
nohwnd, line 29: This was unblocking the loop below, so as soon as we started the app we cancelled tests before any test ever had a chance to run.

So this test was testing almost nothing.

test/IntegrationTests/MSTest.Acceptance.IntegrationTests/AbortionTests.cs
```diff
@@ -25,17 +37,45 @@ public async Task AbortWithCTRLPlusC_CancellingTests(string tfm)
...
+        if (parallelize)
+        {
+            testHostResult.AssertOutputContains("Test Parallelization enabled for");
+        }
+        else
+        {
+            testHostResult.AssertOutputDoesNotContain("Test Parallelization enabled for");
+        }
 
-        testHostResult.AssertOutputMatchesRegex("Canceling the test session.*");
```
nohwnd, line 38: Now we abort so fast that this does not have a chance to write to screen.

src/Adapter/MSTest.TestAdapter/Execution/TestExecutionManager.cs
```diff
@@ -485,11 +489,13 @@ private async Task ExecuteTestsInSourceAsync(IEnumerable<TestCase> tests, IRunCo
             // Queue the non parallel set
             if (nonParallelizableTestSet != null)
             {
+                // What to do?
                 await ExecuteTestsWithTestRunnerAsync(nonParallelizableTestSet, frameworkHandle, source, sourceLevelParameters, testRunner, usesAppDomains);
             }
         }
         else
         {
+            // What to do?
```
nohwnd: Here and few lines above we run the tests when parallelization is not enabled or test is marked with do not parallelize, this blocks the cancellation until the test finishes (and fails the non-parallel tests), and advice here how elegantly solve this? (it won't block if we have parallelism with 1 worker, because that takes the other parallel path.

Youssef1313: @nohwnd How does it block cancellation? `ExecuteTestsWithTestRunnerAsync` calls `ThrowIfCancellationRequested`

nohwnd: Yes but only before it starts running the test. if we cancel while the test is running we don't have a way to abandon it to background and continue to test run teardown code. Like writing test summary to screen. We still have 1 tests in progress and wait for it to complete.

Youssef1313: You can still do the same Task.WhenAny trick right?

nohwnd: yeah 😀 I must be tired

src/Adapter/MSTest.TestAdapter/Execution/TestExecutionManager.cs
```diff
@@ -471,7 +471,9 @@ private async Task ExecuteTestsInSourceAsync(IEnumerable<TestCase> tests, IRunCo
 
                 try
                 {
-                    await Task.WhenAll(tasks);
+                    // Ensure we can abandon currently running tasks on cancellation, rather than waiting for them to complete.
+                    // They will still run on background but we won't await them.
+                    await Task.WhenAny(Task.WhenAll(tasks), _testRunCancellationToken.AsTask());
```
Evangelink: I would make the `Task.WhenAny(task, cancellationToken.AsTask())` part as the helper instead. Also, could we - if not already done - get some blessing from runtime? I saw multiple different suggestions for this online.

src/Adapter/MSTest.TestAdapter/Execution/TestExecutionManager.cs
```diff
@@ -515,6 +520,10 @@ private async Task ExecuteTestsWithTestRunnerAsync(
         UnitTestRunner testRunner,
         bool usesAppDomains)
     {
+        // This method does a lot of work synchronously and cannot be easily cancelled. Make sure we run asynchronously and
+        // can abandon the work here (by calling Task.WhenAny higher in the call stack) when user cancels the test run.
+        await Task.Yield();
```
Youssef1313, line 523: We are already calling ThrowIfCancellationRequested just before we call this method. I'm failing to understand why we need Task.Yeild here

src/Adapter/MSTest.TestAdapter/Execution/TestExecutionManager.cs
```diff
@@ -487,16 +490,18 @@ private async Task ExecuteTestsInSourceAsync(IEnumerable<TestCase> tests, IRunCo
             // Queue the non parallel set
             if (nonParallelizableTestSet != null)
             {
+                _testRunCancellationToken?.ThrowIfCancellationRequested();
```
Youssef1313, line 491: nit: As you have the same line in the if/else, consider moving it once before the `if`

src/Adapter/MSTest.TestAdapter/Execution/TestExecutionManager.cs
nohwnd: ThrowIfCancellationRequested will throw when cancellation was requested before we see that line of code. It has no impact on code that is already running. 

So if we cancel before entering this method, all is good and we will cancel the process and will be able to leave the Main method and teardown the process.

But if we enter this method, it will run **synchronously**, and we are forced to await its completion. So as long as there is a test running  (e.g. test that is hanging) we won't be able to leave that stack and won't be able to complete the Main method. The Task.WhenAny that we called this method from will not be effective because the method is running synchronously.

This Yield makes sure that this method runs asynchronously, so we can avoid awaiting its completion.

Another approach would be to merge how tests are run when parallel is enabled and when parallel is disabled. When parallel is enabled (even if there is single worker allowed = serial execution) we offload the work to a Task in a queue, which is what this does as well.

nohwnd: But I think I am just implementing a partially graceful cancellation here, because we need cooperation from the code that runs the tests (not the test themselves, but the code that runs them), so NUnit would also have to go and make sure they  cancel in a way that allows them to finish the process.

Instead I should maybe lift this approach higher, and yield in place where we start the tests in platform. So platform ran re-gain control on ctrl+c, without any cooperation with the test framework.

src/Adapter/MSTest.TestAdapter/Execution/TestExecutionManager.cs
```diff
@@ -485,12 +488,18 @@ private async Task ExecuteTestsInSourceAsync(IEnumerable<TestCase> tests, IRunCo
             // Queue the non parallel set
             if (nonParallelizableTestSet != null)
             {
-                await ExecuteTestsWithTestRunnerAsync(nonParallelizableTestSet, frameworkHandle, source, sourceLevelParameters, testRunner, usesAppDomains);
+                _testRunCancellationToken?.ThrowIfCancellationRequested();
+                // Ensure we can abandon currently running tasks on cancellation, rather than waiting for them to complete.
+                // They will still run on background but we won't await them.
+                await Task.WhenAny(ExecuteTestsWithTestRunnerAsync(nonParallelizableTestSet, frameworkHandle, source, sourceLevelParameters, testRunner, usesAppDomains), completeWhenTestRunIsCanceled);
```
Youssef1313, line 494: The more I think of it, the more I think it's probably not the right thing. Cancellation will be requested, then we just proceed while the task of `ExecuteTestsWithTestRunnerAsync` could still be running.

MarcoRossignoli: I agree...you don't know what the dangling tasks will do and when...at this point it's better `Env.Exit(abort code)` the result of the run cannot be trusted and some other crash could happen for race for unexpected app state after this return the behavior will be always undefined.

Or you cooperate or you don't cooperate or user should decide what to do `--disable-cooperative-cancellation` and `Env.Exit(abort code)`, by default cooperative and wait.

Evangelink:
@nohwnd could you please summarize the use case you are trying to fix here?

I seem to see 2 cases:

1. We want to exit but we want to allow cleanups (platform, extensions, framework...) in which case current implementation (no modification) is the best and this does match cooperative cancellation. Yes it's possible that you end up with a super long cleanup, it's possible that we have a bug in the platform or some extension that does not play well with cooperative and so needs to be fixed.

2. We want a fast exit, in which case we should end the process (Env fail fast for example) BUT we cannot guarantee cleanup.

From looking up online, it seems that MOST apps/people expect the CTRL+C to be an abortion and not a cancellation so maybe the only change we should do is move the CTRL+C handler to move to the fast exit path.

Youssef1313:
@Evangelink There is also one thing I saw in MSBuild when I was looking recently:

https://github.com/dotnet/msbuild/blob/88496b9c8a4e9b4a84bee8889e47cf106dd7d96d/src/MSBuild/XMake.cs#L1146-L1150

nohwnd:
Summarized in internal document. let's discuss there, it has code samples and the whole plan, but is not short. I am happy to summarize it here once we agree there. Or post it as whole, there is no secret in it.

I think control break is interesting as an abort option, but personally I don't even know where Break is on my keyboard 😁

nohwnd:
> 2. We want a fast exit, in which case we should end the process (Env fail fast for example) BUT we cannot guarantee cleanup.

This is what VSTest does under VS, and in console, even though there is a cancel gesture, and I don't think it is right. We should give the test framework a chance to do at least a minimal cleanup and chance to the plugins to finalize their work.

I would like the cancellation to become the norm (e.g. thanks to better reporting of the cancellation process), and abort without cleanup being the last resort. And if we now implement ctrl+c to kill the process, then there is no way back.

Evangelink:
@nohwnd are you going to complete this soon? Otherwise let's close it.

Issue #5282, opened 2025-03-20
[MTP] Trace logging can be a significant source of memory allocations


Youssef1313:
Profiling our `Microsoft.Testing.Extensions.UnitTests.exe`:

![Image](https://github.com/user-attachments/assets/ceccf340-c130-4333-84a4-4f46c5197852)

These allocations seems significant?

Possible ideas:

- Pool string builders (we can borrow Roslyn's implementation of PooledStringBuilder)
- Create StringBuilders with higher initial capacity to avoid potential resizes
- Don't make trace logging default (https://github.com/microsoft/testfx/pull/5283), and consider the perf hit as "non-issue" as trace logging should only be enabled while investigations of issues.

Youssef1313:
Trace is already documented as:

https://github.com/microsoft/testfx/blob/66b23c38620c1ad74ce3f42d3555defdd2703177/src/Platform/Microsoft.Testing.Platform/Logging/LogLevel.cs#L11-L15

So I think we may not need to really invest in improving perf here, and just decrease the default logging level.

Youssef1313:
Closing. See the linked PR.

PR #5283, opened 2025-03-21
Decrease default logging level to Debug

Youssef1313:
I feel like Trace logging shouldn't be the default, esp that it can have a perf impact (https://github.com/microsoft/testfx/issues/5282).

Also:

https://github.com/microsoft/testfx/blob/66b23c38620c1ad74ce3f42d3555defdd2703177/src/Platform/Microsoft.Testing.Platform/Logging/LogLevel.cs#L11-L15

### DIFF
--- a/src/Platform/Microsoft.Testing.Platform/Builder/TestApplication.cs
+++ b/src/Platform/Microsoft.Testing.Platform/Builder/TestApplication.cs
@@ -291,7 +291,7 @@ public sealed class TestApplication : ITestApplication
             return new(logLevel, result);
         }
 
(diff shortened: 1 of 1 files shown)


Evangelink (changes requested):
I think it's current (before this pr) behavior is correct. When using --diagnostic option I want to use the trace level by default as I run there to have max info about a problem. You can change level if for some reason you want to do something else

Youssef1313:
My bad. I thought it's default. We are passing --diagnostic via TestingPlatformCommandLineArgs.

PR #5291, opened 2025-03-21
Fix TypeCache re-calculating info when running in parallel

Youssef1313:

### DIFF
--- a/src/Adapter/MSTest.TestAdapter/Execution/TypeCache.cs
+++ b/src/Adapter/MSTest.TestAdapter/Execution/TypeCache.cs
@@ -145,6 +145,8 @@ internal sealed class TypeCache : MarshalByRefObject
         DebugEx.Assert(testMethod != null, "test method is null");
 
         string typeName = testMethod.FullClassName;
+        // Using GetOrAdd to ensure we calculate only once when this is called by different threads in parallel.
+        // Using a static lambda to ensure we don't capture.
         return _classInfoCache.GetOrAdd(typeName, static (typeName, tuple) =>
         {
             TestMethod testMethod = tuple.testMethod;
@@ -376,6 +378,8 @@ internal sealed class TypeCache : MarshalByRefObject
     /// <param name="assembly"> The assembly to get its info. </param>
     /// <returns> The <see cref="TestAssemblyInfo"/> instance. </returns>
     private TestAssemblyInfo GetAssemblyInfo(Assembly assembly)
+        // Using GetOrAdd to ensure we calculate only once when this is called by different threads in parallel.
+        // Using a static lambda to ensure we don't capture.
         => _testAssemblyInfoCache.GetOrAdd(assembly, static (assembly, @this) =>
             {
                 var assemblyInfo = new TestAssemblyInfo(assembly);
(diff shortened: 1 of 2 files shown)


src/Adapter/MSTest.TestAdapter/Execution/TypeCache.cs
```diff
@@ -145,9 +145,11 @@ public IEnumerable<TestAssemblyInfo> AssemblyInfoListWithExecutableCleanupMethod
         DebugEx.Assert(testMethod != null, "test method is null");
 
         string typeName = testMethod.FullClassName;
-
-        if (!_classInfoCache.TryGetValue(typeName, out TestClassInfo? classInfo))
+        return _classInfoCache.GetOrAdd(typeName, static (typeName, tuple) =>
```
nohwnd, line 150: Is this the crux of the whole PR? to avoid race condition in the info cache by using proper concurrent api rather than TryGetValue?

src/Adapter/MSTest.TestAdapter/Execution/TypeCache.cs
```diff
@@ -168,13 +170,8 @@ public IEnumerable<TestAssemblyInfo> AssemblyInfoListWithExecutableCleanupMethod
...
             // Get the classInfo
-            classInfo = CreateClassInfo(type, testMethod);
-
-            // Use the full type name for the cache.
-            classInfo = _classInfoCache.GetOrAdd(typeName, classInfo);
-        }
-
-        return classInfo;
+            return @this.CreateClassInfo(type);
+        }, (this, testMethod));
```
nohwnd, line 176: are we passing in this to avoid capturing it, or it has deeper meaning?

src/Adapter/MSTest.TestAdapter/Execution/TypeCache.cs
```diff
@@ -36,7 +36,7 @@ internal sealed class TypeCache : MarshalByRefObject
     /// <summary>
     /// ClassInfo cache.
     /// </summary>
-    private readonly ConcurrentDictionary<string, TestClassInfo> _classInfoCache = new();
+    private readonly ConcurrentDictionary<string, TestClassInfo?> _classInfoCache = new();
```
nohwnd, line 39: Why can the value be null now?

src/Adapter/MSTest.TestAdapter/Execution/TypeCache.cs
Youssef1313: Yup.

src/Adapter/MSTest.TestAdapter/Execution/TypeCache.cs
Youssef1313: Only to avoid capturing. That's why I explicitly added static modifier to the lambda.

src/Adapter/MSTest.TestAdapter/Execution/TypeCache.cs
Youssef1313: Previously, we were not caching null values, but the method that returns the info had a code path that returns null (it was an un-cached code path)

PR #5293, opened 2025-03-21
Fix test property not considering the test class correctly

### LINKED ISSUE #5249: `TestPropertyAttribute` doesn't correctly consider the test class and only considers the "declaring type"
https://github.com/microsoft/testfx/blob/eba5fbbe867519fe2304ec606e76bc2e6d0eefa5/src/Adapter/MSTest.TestAdapter/Execution/TypeCache.cs#L791

This logic is incorrect. `testMethodInfo.TestMethod.DeclaringType` isn't necessarily the same as `testMethodInfo.Parent.ClassType`.

The correct one should be `testMethodInfo.Parent.ClassType`.

This affects the case where there a `TestMethod` is *declared* in abstract class, and the derived test class has `TestProperty` attribute. In that case, the property doesn't correctly flow to `TestContext.Properties`. In this case, the DeclaringType is the abstract class, while the ClassType is really the concrete test class.

Youssef1313:
Fixes #5249


Evangelink (approved):
Could you please follow-up with a test to avoid regression?

Issue #740, opened 2020-10-25
Proposal: UIDataTestMethod / Support for DataRows in UITestMethods


marcelwgn:
The DataTestMethod attribute is really helpful and makes writing tests a lot easier. However when developing tests that need to run on the UI thread, we have to use the UITestMethod attribute. This however leaves us with having to implement data driven tests ourselfes. Having a UIDataTestMethod attribute would be great to have to fill the gap between UI tests and non UI tests.

[AB#1408166](https://devdiv.visualstudio.com/0bdbc590-a062-4c3f-b0f6-9383f67865ee/_workitems/edit/1408166)

nohwnd:
Maybe you could PR it?

chingucoding:
Where would this actually be contributed to? Does UITestMethod even live in this repository?

RosarioPulella:
@chingucoding it does live in this repo.
https://github.com/microsoft/testfx/blob/ec18af6f90c272f68f48d9d0b94c71b8e276c7a0/src/TestFramework/Extension.UWP/UITestMethodAttribute.cs#L12-L43

Also on a similar note, I am working on the [Windows Community Toolkit](aka.ms/uwptoolkit) and we make heavy use of `UITestMethod` either a solution like this would be very useful to be able to use `DataRow` or maybe a better solution would be to make using `UITestMethod` and `DataRow` together work, rather than make a new attribute?

Currently using those attributes together causes this rather obscure [error](https://stackoverflow.com/questions/58410805/parameter-count-mismatch-in-unit-test-uwp-project/66462980#66462980).

chingucoding:
Oh that's interesting, thanks for pointing that out @RosarioPulella!

I'm also not sure what the best move forward would be. I think having a UIDataTestMethod would be best for API simplicity since it extends on an existing pattern. I also got the issue you linked often and was confused what was happening. This error message should definitely be updated.

chingucoding:
@nohwnd What do you think is the best way forward here?

RosarioPulella:
> I think having a UIDataTestMethod would be best for API simplicity since it extends on an existing pattern.

I am not in tune with a lot of the patterns of this API, but I see a lot of test methods decorated with `DataRow` and a simple `TestMethod` and it works fine. It also seems like there is an existing pattern to compose `DataRow` with other attributes than `DataTestMethod`.

chingucoding:
[...]

PR #5297, opened 2025-03-22
Allow async test methods for UITestMethod on UWP and WinUI

### LINKED ISSUE #5306: Support async test methods for UITestMethodAttribute (for UWP/WinUI)
This is currently not supported due to the "sync" nature of test execution in MSTest, which would normally lead to deadlock in this case. This issue is a split of https://github.com/microsoft/testfx/issues/740

Youssef1313:
Fixes #5306


src/Adapter/MSTest.TestAdapter/Execution/TestMethodInfo.cs
```diff
@@ -422,11 +431,33 @@ private TestResult ExecuteInternal(object?[]? arguments, ExecutionContext? execu
                     if (RunTestInitializeMethod(_classInstance, result, ref executionContext, timeoutTokenSource))
                     {
                         hasTestInitializePassed = true;
-                        FixtureMethodRunner.RunOnContext(executionContext, () =>
+                        var tcs = new TaskCompletionSource<object?>();
+#pragma warning disable VSTHRD101 // Avoid unsupported async delegates
+                        FixtureMethodRunner.RunOnContext(executionContext, async () =>
```
Youssef1313: TODO: Refactor to `RunOnContextAsync` and move these helper out of `FixtureMethodRunner` to a new helper `ExecutionContextHelpers`

src/Adapter/MSTest.TestAdapter/Execution/TestMethodInfo.cs
```diff
@@ -1084,7 +1115,8 @@ private TestResult ExecuteInternalWithTimeout(object?[]? arguments, ExecutionCon
 
                 try
                 {
-                    return ExecuteInternal(arguments, executionContext, timeoutTokenSource);
+                    // TODO: Same comment as in ExecuteAsyncAction below.
```
Youssef1313: Outdated. I'll remove

samples/Playground/Tests.cs
Youssef1313, line 1: TODO: Revert

marcelwgn:
Uhm, how does this address https://github.com/microsoft/testfx/issues/740? The former is about being able to use datarows with UITests, this PR on the other hand allows using async on tests right?

Youssef1313:
@marcelwgn The original issue description is inaccurate. Based on the comments/discussions there, the real issue is really just UITestMethodAttribute deadlocking.

marcelwgn:
So the following test should work then?
```
[UITestMethod]
[DataRow(0)]
[DataRow(1)]
[DataRow(2)]
public void TestMethod2(int index)
{
    var grid = new Grid
    {
        MinWidth = index
    };

    grid.UpdateLayout();

    Assert.AreEqual(index, grid.ActualWidth);
}
```

Because when I run it in Visual Studio, I get the following issue:
```
    Cannot run test method 'App8.UnitTest1.TestMethod2': Test data doesn't match method parameters. Either the count or types are different.
Test expected 1 parameter(s), with types 'Int32',
but received 0 argument(s), with types ''.
```

The thread in that issue evolved to providing a workaround to that missing capability using async tests that would essentially implement the behaviour that UITestMethod should support, but that requires async which is not supported. I disagree that the original issue description is misleading, the above sample does not work so there still is no support for this. ALlowing async would just allow for a workaround to work.

Youssef1313:
@marcelwgn This should be working if you are using the latest MSTest. I could repro only with really old versions of MSTest that are not supported.

test/IntegrationTests/MSTest.Acceptance.IntegrationTests/GenericTestMethodTests.cs
```diff
@@ -22,59 +22,59 @@ public async Task TestDifferentGenericMethodTestCases()
...
-                at .+?
+                .+?
             failed ParameterizedMethodSimple \(2\) \(\d+ms\)
               Assert\.Fail failed\. Test method 'ParameterizedMethodSimple' did run with parameter '2' and type 'System\.Int32'\.
-                at .+?
+                .+?
             failed ParameterizedMethodSimple \("Hello world"\) \(\d+ms\)
               Assert\.Fail failed\. Test method 'ParameterizedMethodSimple' did run with parameter 'Hello world' and type 'System\.String'\.
-                at .+?
+                .+?
```
Youssef1313, line 37: Some of these are becoming `--- End of stack trace from previous location ---`

The stack trace was previously not so helpful anyways, so I think the change is fine.

src/Adapter/MSTest.TestAdapter/Execution/TestMethodInfo.cs
```diff
@@ -420,11 +427,33 @@ private TestResult ExecuteInternal(object?[]? arguments, ExecutionContext? execu
                     if (RunTestInitializeMethod(_classInstance, result, ref executionContext, timeoutTokenSource))
                     {
                         hasTestInitializePassed = true;
-                        FixtureMethodRunner.RunOnContext(executionContext, () =>
+                        var tcs = new TaskCompletionSource<object?>();
```
Youssef1313, line 430: The core of the fix is here.
`InvokeAsSynchronousTask` would have previously blocked the UI thread, preventing the continuation from running.

nohwnd (approved):

PR #5298, opened 2025-03-23
Use new version of .NET SDK and re-enable new dotnet test experience

Youssef1313:
We previously made changes in the protocol used by dotnet test to add `InstanceId` which was breaking. So, we temporarily switched back to Arcade's way of testing. Now that we have an updated version of SDK that knows the new protocol, I'm switching back to `dotnet test`.


Evangelink (commented):
This is also changing the pipeline yamls

Evangelink (approved):

Youssef1313:
Updated the title and description to clarify why I'm updating YML.

PR #5314, opened 2025-03-25
Add RetryContext.FirstRunResults

### LINKED ISSUE #5224: RetryAttribute: Allow custom filtering on TestRun to determine whether or not the test should be retried
## Summary

It'd be useful if there was a way to provide custom filtering on the test run to determine if the test should be retried or not.

## Background and Motivation

I have integration tests that are flakey when making requests to some services and I'd like to only retry when a specific exception happens during those flakey requests. (For additional context, those flakey requests happen in the first 5 minutes but the tests can take an hour and it'd be nice if I can only retry when the flakeyness happens)

## Proposed Feature

Make `RetryAttribute` inheritable (no longer sealed) and make `IsAcceptableResultForRetry` in `RetryBaseAttribute`  overridable/no longer static/no longer internal.

#### MyCustomRetryAttribute.cs (example of a custom implementation) 
```csharp
    public class MyCustomRetryAttribute : RetryAttribute  {
        protected override bool IsAcceptableResultForRetry(TestResult[] results)
        {
            foreach (TestResult result in results)
            {
                UnitTestOutcome outcome = result.Outcome;
                if (outcome is UnitTestOutcome.Failed && result.TestFailureException != null && result.TestFailureException!.Message.Contains("id not found"))
                {
                    return false;
                }
            }

            return true;
        }
    }
```

## Alternative Designs

Allow for a type to be passed into a RetryAttribute constructor and call a method from that type when determining if a test should be retried.

#### RetryAttribute.cs (new constructor)
```csharp
    public RetryAttribute(int retryCount, Type retryFilter)
    {
        if (!typeof(BaseRetryFilter).IsAssignableFrom(retryFilter))
        {
            throw new ArgumentException($"{retryFilter.Name} must inherit {nameof(BaseRetryFilter)}");
        }

        RetryFilter = (BaseRetryFilter)Activator.CreateInstance<RetryFilter>();
    }

```
#### RetryBaseAttribute.cs (new property + update method)

```csharp
   // new property
   protected BaseRetryFilter RetryFilter = new BaseRetryFilter ();

    // update method (no longer static and now calls a method from `RetryFilter`
    internal bool IsAcceptableResultForRetry(TestResult[] results, Type[] allowedExceptions)
    {
        foreach (TestResult result in results)
        {
            if (!RetryFilter.FilterFunction(result)) {
              return false;
            }
         }
         return true;
    }
```
#### new class
```csharp
    public class BaseRetryFilter
    {
        public virtual bool FilterFunction(TestResult result)
        {
            UnitTestOutcome outcome = result.Outcome;
            if (outcome is UnitTestOutcome.Failed or UnitTestOutcome.Timeout)
            {
                return false;
            }

            return true;
        }
    }
```

#### custom class
```csharp
    public class CustomRetryFilter : BaseRetryFilter
    {
        public override bool FilterFunction(TestResult result)
        {
            // only retry when exception message contains `id not found`
            return base.FilterFunction(result) && result.TestFailureException != null && result.TestFailureException!.Message.Contains("id not found");
        }
    }
```
#### example 
```csharp
[TestMethod]
[Retry(2, typeof(CustomRetryFilter))]
public void FlakeyTestMethod1() {
 //
}
```

Youssef1313:
Fixes #5224

### DIFF
--- a/src/TestFramework/TestFramework/Attributes/TestMethod/RetryContext.cs
+++ b/src/TestFramework/TestFramework/Attributes/TestMethod/RetryContext.cs
@@ -9,10 +9,10 @@ namespace Microsoft.VisualStudio.TestTools.UnitTesting;
 [Experimental("MSTESTEXP", UrlFormat = "https://aka.ms/mstest/diagnostics#{0}")]
 public readonly struct RetryContext
 {
-    internal RetryContext(Func<Task<TestResult[]>> executeTaskGetter, TestResult[] originalResults)
+    internal RetryContext(Func<Task<TestResult[]>> executeTaskGetter, TestResult[] firstRunResults)
     {
         ExecuteTaskGetter = executeTaskGetter;
-        OriginalResults = originalResults;
+        FirstRunResults = firstRunResults;
     }
 
     /// <summary>
@@ -23,5 +23,5 @@ public readonly struct RetryContext
     /// <summary>
     /// Gets the test results of the initial run that failed.
     /// </summary>
-    public TestResult[] OriginalResults { get; }
+    public TestResult[] FirstRunResults { get; }
 }
--- a/src/TestFramework/TestFramework/Attributes/TestMethod/RetryContext.cs
+++ b/src/TestFramework/TestFramework/Attributes/TestMethod/RetryContext.cs
@@ -9,11 +9,19 @@ namespace Microsoft.VisualStudio.TestTools.UnitTesting;
 [Experimental("MSTESTEXP", UrlFormat = "https://aka.ms/mstest/diagnostics#{0}")]
 public readonly struct RetryContext
 {
-    internal RetryContext(Func<Task<TestResult[]>> executeTaskGetter)
-        => ExecuteTaskGetter = executeTaskGetter;
+    internal RetryContext(Func<Task<TestResult[]>> executeTaskGetter, TestResult[] originalResults)
+    {
+        ExecuteTaskGetter = executeTaskGetter;
+        OriginalResults = originalResults;
+    }
 
     /// <summary>
     /// Gets the function that will execute the test asynchronously.
     /// </summary>
     public Func<Task<TestResult[]>> ExecuteTaskGetter { get; }
+
+    /// <summary>
+    /// Gets the test results of the initial run that failed.
+    /// </summary>
+    public TestResult[] OriginalResults { get; }
 }
--- a/src/TestFramework/TestFramework/PublicAPI/PublicAPI.Unshipped.txt
+++ b/src/TestFramework/TestFramework/PublicAPI/PublicAPI.Unshipped.txt
@@ -3,4 +3,4 @@ Microsoft.VisualStudio.TestTools.UnitTesting.AssertFailedException.AssertFailedE
 Microsoft.VisualStudio.TestTools.UnitTesting.AssertInconclusiveException.AssertInconclusiveException(System.Runtime.Serialization.SerializationInfo! info, System.Runtime.Serialization.StreamingContext context) -> void
 Microsoft.VisualStudio.TestTools.UnitTesting.InternalTestFailureException.InternalTestFailureException(System.Runtime.Serialization.SerializationInfo! info, System.Runtime.Serialization.StreamingContext context) -> void
 Microsoft.VisualStudio.TestTools.UnitTesting.UnitTestAssertException.UnitTestAssertException(System.Runtime.Serialization.SerializationInfo! info, System.Runtime.Serialization.StreamingContext context) -> void
-[MSTESTEXP]Microsoft.VisualStudio.TestTools.UnitTesting.RetryContext.OriginalResults.get -> Microsoft.VisualStudio.TestTools.UnitTesting.TestResult![]!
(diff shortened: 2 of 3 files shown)


src/TestFramework/TestFramework/Attributes/TestMethod/RetryContext.cs
```diff
@@ -9,11 +9,19 @@ namespace Microsoft.VisualStudio.TestTools.UnitTesting;
...
 
     /// <summary>
     /// Gets the function that will execute the test asynchronously.
     /// </summary>
     public Func<Task<TestResult[]>> ExecuteTaskGetter { get; }
+
+    /// <summary>
+    /// Gets the test results of the initial run that failed.
+    /// </summary>
+    public TestResult[] OriginalResults { get; }
```
Youssef1313: ⚠️ This is a public API. @nohwnd Are you fine with the name?

nohwnd: I am not sure from the name what this represents, is this all previous attempts? If so how do I know after how many tries the run succeeded? is that by checking the count in the array, and their order? 

Is the mapping between test case and test result 1:1  now? so I will get separate test result of every test case on a test?

nohwnd: I would call it either `PreviousResults` or `FailedRunAttempts`? or something like that name?

Youssef1313: @nohwnd

> is this all previous attempts?

There is no *previous attempts*. There is only **one previous (initial/original) attempt**

Youssef1313: Maybe `FirstRunResults`?

nohwnd: FirstRunResults seems the best, after re-reading the issue it fixes.

src/TestFramework/TestFramework/Attributes/TestMethod/RetryContext.cs
```diff
@@ -9,11 +9,19 @@ namespace Microsoft.VisualStudio.TestTools.UnitTesting;
 [Experimental("MSTESTEXP", UrlFormat = "https://aka.ms/mstest/diagnostics#{0}")]
 public readonly struct RetryContext
 {
-    internal RetryContext(Func<Task<TestResult[]>> executeTaskGetter)
-        => ExecuteTaskGetter = executeTaskGetter;
+    internal RetryContext(Func<Task<TestResult[]>> executeTaskGetter, TestResult[] originalResults)
+    {
+        ExecuteTaskGetter = executeTaskGetter;
+        OriginalResults = originalResults;
```
Youssef1313:
```suggestion
        FirstRunResults = firstRunResults;
```

src/TestFramework/TestFramework/Attributes/TestMethod/RetryContext.cs
```diff
@@ -9,11 +9,19 @@ namespace Microsoft.VisualStudio.TestTools.UnitTesting;
 [Experimental("MSTESTEXP", UrlFormat = "https://aka.ms/mstest/diagnostics#{0}")]
 public readonly struct RetryContext
 {
-    internal RetryContext(Func<Task<TestResult[]>> executeTaskGetter)
-        => ExecuteTaskGetter = executeTaskGetter;
+    internal RetryContext(Func<Task<TestResult[]>> executeTaskGetter, TestResult[] originalResults)
```
Youssef1313:
```suggestion
    internal RetryContext(Func<Task<TestResult[]>> executeTaskGetter, TestResult[] firstRunResults)
```

src/TestFramework/TestFramework/Attributes/TestMethod/RetryContext.cs
Youssef1313:
```suggestion
    public TestResult[] FirstRunResults { get; }
```

src/TestFramework/TestFramework/PublicAPI/PublicAPI.Unshipped.txt
```diff
@@ -3,3 +3,4 @@ Microsoft.VisualStudio.TestTools.UnitTesting.AssertFailedException.AssertFailedE
 Microsoft.VisualStudio.TestTools.UnitTesting.AssertInconclusiveException.AssertInconclusiveException(System.Runtime.Serialization.SerializationInfo! info, System.Runtime.Serialization.StreamingContext context) -> void
 Microsoft.VisualStudio.TestTools.UnitTesting.InternalTestFailureException.InternalTestFailureException(System.Runtime.Serialization.SerializationInfo! info, System.Runtime.Serialization.StreamingContext context) -> void
 Microsoft.VisualStudio.TestTools.UnitTesting.UnitTestAssertException.UnitTestAssertException(System.Runtime.Serialization.SerializationInfo! info, System.Runtime.Serialization.StreamingContext context) -> void
+[MSTESTEXP]Microsoft.VisualStudio.TestTools.UnitTesting.RetryContext.OriginalResults.get -> Microsoft.VisualStudio.TestTools.UnitTesting.TestResult![]!
```
Youssef1313:
```suggestion
[MSTESTEXP]Microsoft.VisualStudio.TestTools.UnitTesting.RetryContext.FirstRunResults.get -> Microsoft.VisualStudio.TestTools.UnitTesting.TestResult![]!
```

nohwnd (approved):

Issue #5265, opened 2025-03-19
Open up traits for consumption by extensions


Evangelink:
As of now, when using traits with MTP you have 2 options:
- MetadataProperty - used by TUnit, xUnit v3, TATF
- the internal serializable kvp property - used by the bridge

When an extension wants to act on this property it currently can do it only on MetadataProperty which won't work for VSTest bridge extensions (MSTest, NUnit...). The bridge should be changed to no longer use the serializable property but instead rely on the MetadataProperty.

Ideally, in order to reduce memory consumption, these properties should not always be emitted by the test framework and we should have a capability like we have for TRX that would allow clients or extensions to declare they are interested about such properties so that test framework can conditionally emit them.

martincostello:
Any updates here? This issue is the only blocker I have for moving to MTP as it's needed to move https://github.com/Tyrrrz/GitHubActionsTestLogger/pull/43 along.

Youssef1313:
@martincostello There are no blockers for moving `GitHubActionsTestLogger`, except if you are using NUnit as VSTestBridge doesn't have a way to produce `TestMethodIdentifierProperty` for NUnit. The property is properly produced for MSTest, xUnit, and TUnit though.

PR #5316, opened 2025-03-25
VSTestBridge: Add traits as TestMetadataProperty

Youssef1313:
Partially addressing #5265

From the original issue:

> Ideally, in order to reduce memory consumption, these properties should not always be emitted by the test framework and we should have a capability like we have for TRX that would allow clients or extensions to declare they are interested about such properties so that test framework can conditionally emit them.

This is something I'm not doing yet in this PR. So will keep the issue open to address it.

### DIFF
--- a/src/Platform/Microsoft.Testing.Extensions.VSTestBridge/ObjectModel/ObjectModelConverters.cs
+++ b/src/Platform/Microsoft.Testing.Extensions.VSTestBridge/ObjectModel/ObjectModelConverters.cs
@@ -46,7 +46,6 @@ internal static class ObjectModelConverters
     private static void CopyVSTestProperties(IEnumerable<TestProperty> testProperties, TestNode testNode, TestCase testCase, Func<TestProperty, object?> getPropertyValue,
         bool isTrxEnabled, IClientInfo client)
     {
-        List<KeyValuePair<string, string>>? testNodeTraits = null;
         foreach (TestProperty property in testProperties)
         {
             testNode.Properties.Add(new VSTestProperty(property, testCase));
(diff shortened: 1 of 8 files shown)


test/Utilities/Microsoft.Testing.TestInfrastructure/Microsoft.Testing.TestInfrastructure.csproj
```diff
@@ -13,6 +13,7 @@
 
   <ItemGroup>
     <ProjectReference Include="$(RepoRoot)src\Platform\Microsoft.Testing.Platform\Microsoft.Testing.Platform.csproj" />
+    <ProjectReference Include="$(RepoRoot)src\Platform\Microsoft.Testing.Extensions.VSTestBridge\Microsoft.Testing.Extensions.VSTestBridge.csproj" />
```
Youssef1313, line 16: ⚠️ This is needed because VSTestBridge was consuming types via IVT.

Now, newer versions of MTP are not compatible with older versions of VSTestBridge, as VSTestBridge would be trying to access something that no longer exists.

For our tests, at least Microsoft.Testing.Platform.UnitTests project will be running using the current build of MTP. But will use the VSTestBridge via transitive dependency of MSTest version we use for testing, which is not compatible. By adding this ProjectReference, we ensure we also use the current build of VSTestBridge.

src/Platform/Microsoft.Testing.Extensions.VSTestBridge/ObjectModel/ObjectModelConverters.cs
```diff
@@ -90,38 +89,29 @@ private static void CopyVSTestProperties(IEnumerable<TestProperty> testPropertie
...
+            if (property.Id == "TestObject.Traits"
+                && getPropertyValue(property) is KeyValuePair<string, string>[] traits && traits.Length > 0)
+            {
+                foreach (KeyValuePair<string, string> trait in traits)
                 {
-#pragma warning disable SA1010 // Opening square brackets should be spaced correctly
-                    testNodeTraits ??= [];
-#pragma warning restore SA1010 // Opening square brackets should be spaced correctly
-                    testNodeTraits.AddRange(traits);
+                    testNode.Properties.Add(new TestMetadataProperty(trait.Key, trait.Value));
```
nohwnd, line 100: Doing this it looks like we will not be able to separate traits from other test metadata anymore? This might be unexpected for VS. Have you tested what we see in TE after this change in the traits dropdown / selection ?

nohwnd: Looking at the json below in tests `"traits":[{"testmetadata-key":"testmetadata-value"}]`

we see testMetadataProperty  collected here in the trait key. Looks like we would have a separate object for each key, so we could change to:
```
"traits":[
    {
        "testmetadata-key":"testmetadata-value",
        "test-metadata-type": "platform-metadata"
    },
]
```

But I am unsure it that is needed.

Youssef1313: @nohwnd I'm not following. TestMetadataProperty and traits are the same thing. The only thing that we should have on MTP is just `TestMetadataProperty`. I see nothing to separate

nohwnd: I thought that what we shown in VS was the data produced by Traits / Category however each framework calls that attribute. And in MTP traits is bigger group of all possible properties and test metadata you can put on a test. In TATF there was a KVP Property on the test. So if I wanted to add a description to a test I would add "description"="this is my tests and I like it described", but that is not a category of a test that should show up in TE (in my opinion).

Youssef1313: I don't know exactly what TATF was doing. But purely from MTP point of view, there is only one concept, "traits". I'm not seeing what problem could be introduced by the PR. The change done here should only affect external extensions so that they can retrieve the traits via `TestMetadataProperty`, while previously they had no way.

nohwnd: Ok. I will leave it up to your judgement :)

nohwnd (approved):

PR #5323, opened 2025-03-27
[Source Breaking (only for framework authors)]: Support test artifacts in VS

### LINKED ISSUE #4934: Is TestNodeFileArtifact the way we report attachments in MTP?

Youssef1313:
Fixes #4934

This change is a breaking change for framework authors. Framework authors will now need to add attachments to TestNode properties as `TestFileArtifactProperty`. There can be multiple of such property, or none at all.

Publishing `TestNodeFileArtifact` is no longer the way to report artifacts. The type is marked obsolete but is kept for binary compatibility.

Data consumers who used to consume `TestNodeFileArtifact` will instead need to consume `TestNodeUpdateMessage` and find the artifacts in the properties of the test node.

Tested in Playground:

code:

```csharp
[TestClass]
public class TestClass
{
    public TestContext TestContext { get; set; }

    [TestMethod]
    [DataRow(0, UnfoldingStrategy = TestDataSourceUnfoldingStrategy.Fold)]
    [DataRow(1, UnfoldingStrategy = TestDataSourceUnfoldingStrategy.Fold)]
    public void TestMethod1(int x)
    {
        for (int i = 0; i < 5; i++)
        {
            TestContext.AddResultFile($"File #{i} for test case {x}");
        }
    }
}
```

image

@drognanar The order of attachments is reversed compared to VSTest, I think. But that seems to be on Test Explorer side.

FYI @bradwilson @thomhurst 

Also @OsirisTerje. As you are relying on VSTestBridge, you will only need to bump MTP version so that attachments start to show correctly in Test Explorer.


src/Platform/Microsoft.Testing.Platform/Messages/TestNodeProperties.cs
```diff
@@ -372,6 +374,15 @@ public record StandardOutputProperty(string StandardOutput) : IProperty;
...
 public record StandardErrorProperty(string StandardError) : IProperty;
 
+/// <summary>
+/// Property that represents multiple artifacts/attachments to associate with a test node.
+/// </summary>
+/// <param name="SessionUid">The session UID.</param>
+/// <param name="FileInfo">The file information.</param>
+/// <param name="DisplayName">The display name.</param>
+/// <param name="Description">The description.</param>
+public record TestFileArtifactProperty(SessionUid SessionUid, FileInfo FileInfo, string DisplayName, string? Description = null) : IProperty;
```
MarcoRossignoli: maybe for some time we could ship it as [Experimental("TPEXP", UrlFormat = "https://aka.ms/testingplatform/diagnostics#{0}")]

MarcoRossignoli: Why not simply `FileArtifact`?  Test is adding nothing useful could be any file or?

cc: @Evangelink

MarcoRossignoli (dismissed):
Nit on name LGTM

src/Platform/Microsoft.Testing.Platform/Messages/TestNodeProperties.cs
Youssef1313: @MarcoRossignoli We always have `Property` suffix for properties, and there is already `FileArtifact` class (non-property that's used for TRX). So I assume you mean `FileArtifactProperty`? I'm fine with either name.

MarcoRossignoli: `FileArtifactProperty` yep not bad imo

nohwnd:
Do you know who we are breaking with this? Is this all nunit, xunit, mstest and tunit?

Youssef1313:
> Do you know who we are breaking with this? Is this all nunit, xunit, mstest and tunit?

xUnit and TUnit. For MSTest, we are always using the current version of VSTestBridge which accounts for the break in this PR. For NUnit, they are also good as they will just update VSTestBridge and that's it.

nohwnd:
Does sound reasonable to me. :) (how about Expecto?)

nohwnd (approved):

Evangelink:
Expecto also relies on the bridge so it's also only a matter of bumping version.

thomhurst:
Happy to make this change.

Session Artifacts will remain as they are today?

Youssef1313:
@thomhurst Yes, session artifacts are still the same. This only affects `TestNodeFileArtifact` which previously didn't show up in VS at all.

bradwilson:
Is this change pushed to NuGet yet? If not, is there a planned release vehicle & timing?

Youssef1313:
@bradwilson It will be available in 1.7. We may want to release a preview of 1.7 sometime soon.

bradwilson:
Available in xUnit.net v3 `3.0.0-pre.4` https://xunit.net/docs/using-ci-builds

PR #5330, opened 2025-03-28
Move System.Threading.Tasks.Extensions to TestFramework

Youssef1313:
Scenario:

A test project uses MSTest 3.8.3, and upgrades `System.Threading.Tasks.Extensions` to 4.6.0 or later.
MSTest.TestAdapter already depends on 4.5.4.

Because the user is using a newer version, the requested assembly version by MSTest.TestAdapter won't be found.
**Normally**, binding redirects should be generated and are considered by VSTest. However, the tooling only consider assemblies in `Reference` item for binding redirect generation. As MSTest.TestAdapter NuGet package is kinda special and the adapter dll isn't added to `Reference` item, binding redirect is not generated.

### DIFF
--- a/src/Adapter/MSTest.TestAdapter/MSTest.TestAdapter.nuspec
+++ b/src/Adapter/MSTest.TestAdapter/MSTest.TestAdapter.nuspec
@@ -3,10 +3,13 @@
   <metadata>
     $CommonMetadataElements$
     <dependencies>
-      <!-- WARNING! MSTest.TestAdapter shouldn't have dependencies unless they are used by MTP only -->
-      <!-- If you add a dependency here, and user decided to upgrade it in their own project, it won't be considered -->
-      <!-- by GenerateBindingRedirects target (because adapter isn't add as "Reference" item) -->
-      <!-- This can then cause the assembly to fail loading. -->
+      <!-- 
+        WARNING! MSTest.TestAdapter shouldn't have dependencies unless they are used by MTP only.
+        MSTest.TestAdapter dll is not referenced by the test project, unless it uses MTP. So the dependency 
+        won't be considered by GenerateBindingRedirects target. When user references the same package and
+        upgrades it to version that is newer than what we reference, they will get assembly load failures, because
+        the assembly version is mismatched, and there is not binding redirect to correct it.
+      -->
       <group targetFramework="netstandard2.0">
         <dependency id="Microsoft.Testing.Extensions.VSTestBridge" version="$TestingPlatformVersion$" />
         <dependency id="Microsoft.Testing.Platform.MSBuild" version="$TestingPlatformVersion$" />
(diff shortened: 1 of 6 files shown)


src/Adapter/MSTest.TestAdapter/MSTest.TestAdapter.nuspec
```diff
@@ -3,18 +3,19 @@
   <metadata>
     $CommonMetadataElements$
     <dependencies>
+      <!-- WARNING! MSTest.TestAdapter shouldn't have dependencies unless they are used by MTP only -->
+      <!-- If you add a dependency here, and user decided to upgrade it in their own project, it won't be considered -->
+      <!-- by GenerateBindingRedirects target (because adapter isn't add as "Reference" item) -->
+      <!-- This can then cause the assembly to fail loading. -->
       <group targetFramework="netstandard2.0">
```
nohwnd: nitpick, and rewrite to be hopefully more understandable to layman:

```suggestion
      <!-- 
        WARNING! MSTest.TestAdapter shouldn't have dependencies unless they are used by MTP only.
        MSTest.TestAdapter dll is not referenced by the test project, unless it uses MTP. So the dependency 
        won't be considered by GenerateBindingRedirects target. When user references the same package and
        upgrades it to version that is newer than what we reference, they will get assembly load failures, because
        the assembly version is mismatched, and there is not binding redirect to correct it.
      -->
      <group targetFramework="netstandard2.0">
```

src/TestFramework/TestFramework.Extensions/MSTest.TestFramework.NonWindows.nuspec
```diff
@@ -5,6 +5,7 @@
     <dependencies>
       <group targetFramework="netstandard2.0">
         <dependency id="MSTest.Analyzers" version="$Version$" />
+        <dependency id="System.Threading.Tasks.Extensions" version="$SystemThreadingTasksExtensionsVersion$" />
```
nohwnd, line 12: group for dependencies that would normally belong to adapter?

src/TestFramework/TestFramework.Extensions/TestFramework.Extensions.csproj
```diff
@@ -59,6 +59,7 @@
 
   <ItemGroup>
     <PackageReference Condition=" '$(TargetFramework)' == '$(WinUiMinimum)' " Include="Microsoft.WindowsAppSDK" Version="$(MicrosoftWindowsAppSDKVersion)" />
+    <PackageReference Include="System.Threading.Tasks.Extensions" Condition=" '$(TargetFramework)' == 'netstandard2.0' OR '$(TargetFramework)' == '$(NetFrameworkMinimum)' OR '$(TargetFramework)' == '$(UwpMinimum)' " />
```
nohwnd, line 66: group for items that would normally belong to adapter?

nohwnd (commented):
I wish there was more info in the description, once we forget why this change was made :)

nohwnd (approved):

Issue #5366, opened 2025-04-03
Allow running on wasm platform


jjonescz:
## Describe the bug

When I run MTP on wasi-wasm runtime, it fails.

## Steps To Reproduce

Create wasi-wasm app with .NET 10 Preview 2

```csproj
<Project Sdk="Microsoft.NET.Sdk">
  <PropertyGroup>
    <TargetFramework>net10.0</TargetFramework>
    <RuntimeIdentifier>wasi-wasm</RuntimeIdentifier>
    <OutputType>Exe</OutputType>
    <PublishTrimmed>true</PublishTrimmed>
  </PropertyGroup>

  <ItemGroup>
    <PackageReference Include="TUnit" Version="0.19.32" />
  </ItemGroup>
</Project>
```

Install wasi workload

```ps1
dotnet workload install wasi-experimental
```

Build it

```ps1
dotnet build
```

Run it

```ps1
cd bin\Debug\net10.0\wasi-wasm\AppBundle
wasmtime run --wasi http --dir . -- dotnet.wasm App1
```

## Expected behavior

Runs.

## Actual behavior

```
Unhandled Exception:
System.PlatformNotSupportedException: Arg_PlatformNotSupported
   at System.Runtime.InteropServices.PosixSignalRegistration.Register(PosixSignal signal, Action`1 handler)
   at System.Runtime.InteropServices.PosixSignalRegistration.Create(PosixSignal signal, Action`1 handler)
   at System.Console.add_CancelKeyPress(ConsoleCancelEventHandler value)
   at Microsoft.Testing.Platform.Helpers.SystemConsole.add_CancelKeyPress(ConsoleCancelEventHandler value)
   at Microsoft.Testing.Platform.Services.CTRLPlusCCancellationTokenSource..ctor(IConsole console, ILogger logger)
   at Microsoft.Testing.Platform.Hosts.TestHostBuilder.BuildAsync(ApplicationLoggingState loggingState, TestApplicationOptions testApplicationOptions, IUnhandledExceptionsHandler unhandledExceptionsHandler, DateTimeOffset createBuilderStart)
   at Microsoft.Testing.Platform.Builder.TestApplicationBuilder.BuildAsync()
   at TestingPlatformEntryPoint.Main(String[] args)
   at TestingPlatformEntryPoint.<Main>(String[] args)
```

## Additional context

This particular error would likely go away if WASM was added to this exclusion list:

[...]

PR #5367, opened 2025-04-03
Mark APIs not supported in wasi

Youssef1313:
There are likely more needed changes to properly support wasi, but this is a good step at least.

Related to #5366, but likely more changes will be needed.

cc @jjonescz

### DIFF
--- a/src/Platform/Microsoft.Testing.Extensions.HotReload/HotReloadHandler.cs
+++ b/src/Platform/Microsoft.Testing.Extensions.HotReload/HotReloadHandler.cs
@@ -51,6 +51,7 @@ internal sealed class HotReloadHandler
         => RuntimeInformation.IsOSPlatform(OSPlatform.Create("ANDROID")) ||
             RuntimeInformation.IsOSPlatform(OSPlatform.Create("IOS")) ||
             RuntimeInformation.IsOSPlatform(OSPlatform.Create("TVOS")) ||
+            RuntimeInformation.IsOSPlatform(OSPlatform.Create("WASI")) ||
             RuntimeInformation.IsOSPlatform(OSPlatform.Create("BROWSER"));
 
     // Called automatically by the runtime through the MetadataUpdateHandlerAttribute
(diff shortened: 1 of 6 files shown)


src/Platform/Microsoft.Testing.Extensions.HotReload/HotReloadHandler.cs
```diff
@@ -51,6 +51,7 @@ private static bool IsCancelKeyPressNotSupported()
         => RuntimeInformation.IsOSPlatform(OSPlatform.Create("ANDROID")) ||
             RuntimeInformation.IsOSPlatform(OSPlatform.Create("IOS")) ||
             RuntimeInformation.IsOSPlatform(OSPlatform.Create("TVOS")) ||
+            RuntimeInformation.IsOSPlatform(OSPlatform.Create("WASI")) ||
```
jjonescz, line 54: Are you sure the keyword is `WASI` and not `WASM`? I have no idea honestly, just asking.

Youssef1313: Based on .NET runtime code, I think yes it's WASI.

https://github.com/dotnet/runtime/blob/15872212c29cecc8d82da4548c3060f2614665f7/src/libraries/System.Private.CoreLib/src/System/Runtime/InteropServices/RuntimeInformation.cs#L26

https://github.com/dotnet/runtime/blob/15872212c29cecc8d82da4548c3060f2614665f7/src/libraries/System.Private.CoreLib/src/System/OperatingSystem.cs#L127

https://github.com/dotnet/runtime/blob/15872212c29cecc8d82da4548c3060f2614665f7/src/libraries/System.Private.CoreLib/src/System/OperatingSystem.cs#L14-L43

jjonescz: Oh yeah, makes sense, it's `wasi` for `wasi-wasm` and `browser` for `browser-wasm`.

nohwnd (approved):

Issue #5176, opened 2025-03-05
Warnings are not written out


nohwnd:
## Describe the bug

When class fails to discover there is no warning.

## Steps To Reproduce

The project below is broken, when you will run it using vstest console from nuget package it fails to load system.valuetuple, because of the binding redirect in app.config.

This will prevent classes to be discovered, but all the details are swallowed and reported neither to screen, nor to tptrace.

Looks like eqTrace might not be getting the real settings because we in appdomain (this is netframework project). 

And the warnings I did not investigate yet.
```
S:\t\mstest229> dotnet build
S:\t\mstest229> $env:vstest_host_debug_attachvs=0; C:\Users\jajares\.nuget\packages\microsoft.testplatform\17.13.0\tools\net462\Common7\IDE\Extensions\TestPlatform\vstest.console.exe bin\Debug\net472\mstest229.dll  /Diag:logs/vstest.console.diagnostics.TpTrace.log /inisolation
```

[mstest229.zip](https://github.com/user-attachments/files/19091520/mstest229.zip)

Note that we would be glad if you could enclose a repro (zip or link to some GitHub project).
-->

## Expected behavior

I am told what was the cause of the issue, and that some classes were not discovered.

## Actual behavior

![Image](https://github.com/user-attachments/assets/cd776bda-4911-40a2-ab1a-4fc9d91080b7)

## Additional context

-[ ] check if we handle warnings correctly in mtp

nohwnd:
![Image](https://github.com/user-attachments/assets/513489b3-59f7-46a1-a520-52d25f5d69ad)

This is where we get the messages logged including the details of the assembly load failure, but the messages don't end up anywhere. eqtrace is set to false.

**make sure you run /inisolation to use attach vs (or use the one that attaches to runner)**

warnings also look like they don't propagate back from the appdomain, where we pass an instance of list<string>, not 100% sure, I thought the marshalling is clever enough to handle that.

![Image](https://github.com/user-attachments/assets/aab555cf-e170-4c5a-bfe5-6fc48eec63ac)

nohwnd:
this 

`vstest.console.exe .\TestProject8\bin\Debug\net481\TestProject8.exe  -- RunConfiguration.DisableAppDomain=true`

and this

`vstest.console.exe .\TestProject8\bin\Debug\net481\TestProject8.exe  `

should show the same result.

nohwnd:
It also looks like we 
-[ ] don't write them out from MTP at all

PR #5371, opened 2025-04-04
Write warnings outside of appdomain

nohwnd:
Related #5176 

Figuring out a reliable way to test this, will follow up with other PR. Ensured manually that it works.


src/Adapter/MSTest.TestAdapter/Discovery/AssemblyEnumeratorWrapper.cs
```diff
@@ -114,6 +114,10 @@ private static ICollection<UnitTestElement> GetTestsInIsolation(string fullFileP
             PlatformServiceProvider.Instance.AdapterTraceLogger.LogWarning(Resource.OlderTFMVersionFound);
         }
 
-        return assemblyEnumerator.EnumerateAssembly(fullFilePath, warnings);
+        // This method runs inside of appdomain, when appdomains are available and enabled.
+        // Don't move the 'warning' parameter to be written into by reference, it will not return the warnings from appdomain.
+        AssemblyEnumerationResult result = assemblyEnumerator.EnumerateAssembly(fullFilePath);
+        warnings.AddRange(result.Warnings);
```
Youssef1313: nit: return AssemblyEnumerationResult here and get rid of the `warnings` parameter of `GetTestsInIsolation`. The caller is always passing a fresh list anyways, so I think it's easier to follow to just return the result here. The caller can add extra warnings to the returned list if needed

Youssef1313 (approved):

src/Adapter/MSTest.TestAdapter/Discovery/AssemblyEnumeratorWrapper.cs
nohwnd: done

Youssef1313 (approved):

test/IntegrationTests/MSTest.Acceptance.IntegrationTests/ThreadingTests.cs
```diff
@@ -225,7 +225,7 @@ public sealed class TestAssetFixture() : TestAssetFixtureBase(AcceptanceFixture.
         public const string STAThreadProjectName = "STATestThreading";
         public const string LifecycleAttributesVoidProjectName = "LifecycleAttributesVoid";
         public const string LifecycleAttributesTaskProjectName = "LifecycleAttributesTask";
-        public const string LifecycleWithParallelAttributesTaskProjectName = "LifecycleAttributesTask";
+        public const string LifecycleWithParallelAttributesTaskProjectName = "LifecycleWithParallelAttributesTask";
```
nohwnd, line 228: This name was the same as above so we ended up writing into the same folder and mixing sources, added explicit check for that so it does not happen, and each asset gets its own folder that did not exist before. The asset ids need to be unique only within the same asset group (produced by GetAssetsToGenerate method), not within the whole solution.

nohwnd:
I've re-run it on the merge, and not here. AAARGh!

fhnaseer (approved):

PR #5378, opened 2025-04-06
remove redundant Empty constructor and mark as error

SimonCropp:

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -683,6 +683,9 @@ csharp_style_unused_value_expression_statement_preference = discard_variable:sil
 # IDE0290: Use primary constructor
 csharp_style_prefer_primary_constructors = false
 
+# Empty constructor
+resharper_empty_constructor_highlighting = error
+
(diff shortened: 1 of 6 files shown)


.editorconfig
```diff
@@ -683,6 +683,9 @@ csharp_style_unused_value_expression_statement_preference = discard_variable:sil
 # IDE0290: Use primary constructor
 csharp_style_prefer_primary_constructors = false
 
+# Empty constructor
+resharper_empty_constructor_highlighting = error
```
Evangelink: We usually prefer warning only to allow moving faster in IDE and we have warnings as errors in CLI to ensure (when no bug) we have the expected "style". We don't work with ReSharper so I don't know if it works differently.

Youssef1313: Yeah I'm concerned of introducing code style preferences that can break IDE and they can't be enforced in CI.

Youssef1313 (approved):

Evangelink (approved):

PR #5383, opened 2025-04-06
use null propagation and mark as warning in editorconfig

SimonCropp:

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -151,7 +151,7 @@ csharp_style_throw_expression = true:warning
 dotnet_style_coalesce_expression = true:warning
 
 # IDE0031: Use null propagation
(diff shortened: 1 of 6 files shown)


.editorconfig
```diff
@@ -151,7 +151,7 @@ csharp_style_throw_expression = true:warning
 dotnet_style_coalesce_expression = true:warning
 
 # IDE0031: Use null propagation
-dotnet_style_null_propagation = true:warning
+dotnet_style_null_propagation = true:error
```
Youssef1313: We treat warnings as errors in CI. But for local builds we prefer warnings.

Youssef1313 (dismissed):
LGTM with one small comment

PR #5394, opened 2025-04-06
fix nullability of InvokeTestingPlatformTask _outputFileName

SimonCropp:


Evangelink (approved):

Youssef1313 (approved):
LGTM. We are losing the compiler warning here as we suppress it at the beginning of the file for other false positives for properties that will be set by MSBuild.

PR #5398, opened 2025-04-06
fix nullability in FileLoggerProvider

SimonCropp:
since `FileLogger` is never assigned to null

### DIFF
--- a/src/Platform/Microsoft.Testing.Platform/Logging/FileLoggerProvider.cs
+++ b/src/Platform/Microsoft.Testing.Platform/Logging/FileLoggerProvider.cs
@@ -47,8 +47,7 @@ internal sealed class FileLoggerProvider(
         // If custom directory is provided for the log file, we don't WANT to move the log file
         // We won't betray the users expectations.
         if (_customDirectory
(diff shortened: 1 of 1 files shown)


Youssef1313 (dismissed):

src/Platform/Microsoft.Testing.Platform/Logging/FileLoggerProvider.cs
```diff
@@ -73,15 +72,10 @@ public ILogger CreateLogger(string categoryName)
...
 #if NETCOREAPP
-    public async ValueTask DisposeAsync()
-    {
-        if (FileLogger is not null)
-        {
-            await FileLogger.DisposeAsync();
-        }
-    }
+    public ValueTask DisposeAsync()
+        => FileLogger.DisposeAsync();
```
Youssef1313: I wouldn't be concerned about any perf benefits here and would favor preserving stack traces for exceptions here.

```suggestion
    public async ValueTask DisposeAsync()
        => await FileLogger.DisposeAsync();
```

nohwnd (approved):

PR #5401, opened 2025-04-06
fix nullability of AssemblyEnumerator.GetTypes

SimonCropp:
* `ReflectionTypeLoadException.Types` returns `Type?[]` https://learn.microsoft.com/en-us/dotnet/api/System.Reflection.ReflectionTypeLoadException.Types?view=net-9.0 to return type of `AssemblyEnumerator.GetTypes` is incorrect
* `ReflectionTypeLoadException.LoaderExceptions` returns a `Exception?[]` https://learn.microsoft.com/en-us/dotnet/api/system.reflection.reflectiontypeloadexception.loaderexceptions?view=net-9.0  so the null check is a no-op. i assume the intent is to "log a warning and continue with loaded types **if there are LoaderExceptions**"
* fixed a bug where `GetAssemblyInfo` would redundantly attempt to `IsDerivedAttributeDefined` for a null value, and then catch the result null ref excpetion


src/Adapter/MSTest.TestAdapter/Execution/TypeCache.cs
```diff
@@ -384,14 +384,19 @@ private TestAssemblyInfo GetAssemblyInfo(Assembly assembly)
...
-                Type[] types = AssemblyEnumerator.GetTypes(assembly, assembly.FullName!, null);
+                Type?[] types = AssemblyEnumerator.GetTypes(assembly, assembly.FullName!, null);
 
-                foreach (Type t in types)
+                foreach (Type? type in types)
                 {
+                    if (type == null)
+                    {
+                        continue;
+                    }
```
Youssef1313, line 394: This is a change in behavior.

Previously if `type` was null, we will go through the `try`, which will fail, then we get into the `catch` that will fail due to `t.FullName` (it will throw NullReferenceException).

The original intent is to probably ignore types that failed to load, but that as well wasn't a good design decision.

I think it's best in v4 if we just not catch any `ReflectionTypeLoadException` and let everything fail.

SimonCropp:
going to re-do this one due to the conflicts

PR #5404, opened 2025-04-06
Show disk info when hang dump fails

Youssef1313:
For context: https://github.com/dotnet/diagnostics/issues/2215

### DIFF
--- a/src/Platform/Microsoft.Testing.Extensions.HangDump/HangDumpProcessLifetimeHandler.cs
+++ b/src/Platform/Microsoft.Testing.Extensions.HangDump/HangDumpProcessLifetimeHandler.cs
@@ -191,12 +191,43 @@ internal sealed class HangDumpProcessLifetimeHandler : ITestHostProcessLifetimeH
             await _logger.LogDebugAsync($"Connected to the test host server pipe '{_namedPipeClient.PipeName}'");
 
             // Keep the custom thread to avoid to waste one from thread pool.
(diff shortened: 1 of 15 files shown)


src/Platform/Microsoft.Testing.Extensions.HangDump/HangDumpProcessLifetimeHandler.cs
```diff
@@ -191,14 +191,45 @@ public async Task OnTestHostProcessStartedAsync(ITestHostProcessInformation test
...
 
+    private static string GetDiskInfo()
+    {
+        var builder = new StringBuilder();
+        DriveInfo[] allDrives = DriveInfo.GetDrives();
+
+        foreach (DriveInfo d in allDrives)
+        {
+            builder.AppendLine(CultureInfo.InvariantCulture, $"Drive {d.Name}");
+            if (d.IsReady)
```
fhnaseer, line 222: Print something if it is not ready? Like it is locked by bitlocker or some other condition,

Evangelink (approved):

src/Platform/Microsoft.Testing.Extensions.HangDump/HangDumpProcessLifetimeHandler.cs
Youssef1313: Will do in a follow-up PR.

nohwnd (approved):
Approved, but I would put getting the extra info into try catch of its own so we don’t fail in case the info is not available to us

nohwnd:
I am also not sure why we report on all drives, not just the one that should write the dump. I assume there can be also network drives, etc. And it seems like too much info.

PR #5409, opened 2025-04-07
VSTestBridge+MSTest: Use TestMethodIdentifierProperty and stop sending VSTest-specifics

### LINKED ISSUE #5406: VSTestBridge doesn't use `TestMethodIdentifierProperty`
This property can be used by extensions to know test information.

The only thing we add in VSTestBridge is `vstest.TestCase.FullyQualifiedName` which is even done only for VS and isn't publicly accessible.

We need a reliable way to correctly add `TestMethodIdentifierProperty`, but maybe we are limited by VSTest there.

Ideally, we would get rid of VSTestBridge completely and have an MSTest implementation completely based on MTP, but NUnit is also using the bridge so it would be good if we can fix the bridge as well

Youssef1313:
Fixes #5406

@drognanar Any concerns?

- MSTest won't use the capabilities implementation of the bridge so that it removes vstestProvider (we still keep it for NUnit and Expecto)
- When a framework of the VSTestBridge uses vstestProvider, everything is the same as today.
- When a framework of the VSTestBridge doesn't use vstestProvider (i.e, MSTest for now), we expect ManagedType/ManagedMethod to be present. In that case, we parse them via AdapterUtilities to construct TestMethodIdentifierProperty.
- We stop serializing location.namespace completely. And include both the namespace and type name of TestMethodIdentifierProperty in `location.type`.


src/Platform/Microsoft.Testing.Extensions.VSTestBridge/ObjectModel/ObjectModelConverters.cs
```diff
@@ -238,22 +215,88 @@ internal static void FixUpTestCase(this TestCase testCase, string? testAssemblyP
...
+        ReadOnlySpan<char> managedType,
+        ReadOnlySpan<char> managedMethod,
+        [NotNullWhen(true)] out TestMethodIdentifierProperty? methodIdentifierProperty)
+    {
+        string assemblyFullName = string.Empty;
+        string @namespace;
+        string typeName;
+        string methodName;
+        string[] parameterTypeFullNames = Array.Empty<string>();
+        string returnTypeFullName = string.Empty;
```
SimonCropp: this variable doesnt seem to be used. perhaps just inline it into the `new TestMethodIdentifierProperty` call?

src/Platform/Microsoft.Testing.Extensions.VSTestBridge/ObjectModel/ObjectModelConverters.cs
```diff
@@ -238,22 +215,88 @@ internal static void FixUpTestCase(this TestCase testCase, string? testAssemblyP
...
+    private static bool TryGetMethodIdentifierPropertyFromManagedTypeAndManagedMethod(
+        ReadOnlySpan<char> managedType,
+        ReadOnlySpan<char> managedMethod,
+        [NotNullWhen(true)] out TestMethodIdentifierProperty? methodIdentifierProperty)
+    {
+        string assemblyFullName = string.Empty;
+        string @namespace;
+        string typeName;
+        string methodName;
+        string[] parameterTypeFullNames = Array.Empty<string>();
```
SimonCropp: avoid redundantly assigning the variable. and instead assign it in the else of the below if

SimonCropp: perhaps use a collection expression instead of `Array.Empty<string>();`

src/Platform/Microsoft.Testing.Extensions.VSTestBridge/ObjectModel/ObjectModelConverters.cs
```diff
@@ -238,22 +215,88 @@ internal static void FixUpTestCase(this TestCase testCase, string? testAssemblyP
...
     {
-        fullyQualifiedType = null;
+        string? managedType = testCase.GetPropertyValue<string>(ManagedTypeProperty, defaultValue: null);
+        string? managedMethod = testCase.GetPropertyValue<string>(ManagedMethodProperty, defaultValue: null);
+        // NOTE: ManagedMethod, in case of MSTest, will have the parameter types.
+        // So, we prefer using it to display the parameter types in Test Explorer.
+        return !RoslynString.IsNullOrEmpty(managedType) && !RoslynString.IsNullOrEmpty(managedMethod)
+            ? TryGetMethodIdentifierPropertyFromManagedTypeAndManagedMethod(managedType, managedMethod, out methodIdentifierProperty)
+            : TryGetMethodIdentifierPropertyFromFullyQualifiedName(testCase.FullyQualifiedName, out methodIdentifierProperty);
+    }
```
SimonCropp, line 300: this would read better if the postive case goes first

```
return RoslynString.IsNullOrEmpty(managedType) &&
               RoslynString.IsNullOrEmpty(managedMethod)
            ? TryGetMethodIdentifierPropertyFromFullyQualifiedName(testCase.FullyQualifiedName, out methodIdentifierProperty)
            : TryGetMethodIdentifierPropertyFromManagedTypeAndManagedMethod(managedType, managedMethod, out methodIdentifierProperty);
```

src/Platform/Microsoft.Testing.Extensions.VSTestBridge/ObjectModel/ObjectModelConverters.cs
```diff
@@ -238,22 +215,88 @@ internal static void FixUpTestCase(this TestCase testCase, string? testAssemblyP
...
+    {
+        string assemblyFullName = string.Empty;
+        string @namespace;
+        string typeName;
+        string methodName;
+        string[] parameterTypeFullNames = Array.Empty<string>();
+        string returnTypeFullName = string.Empty;
+
+        int indexOfParen = managedMethod.IndexOf('(');
+        if (indexOfParen != -1)
```
SimonCropp: invert so the postive case goes first

src/Platform/Microsoft.Testing.Extensions.VSTestBridge/ObjectModel/ObjectModelConverters.cs
```diff
@@ -238,22 +215,88 @@ internal static void FixUpTestCase(this TestCase testCase, string? testAssemblyP
...
+        // Get type name
+        int lastIndexOfDot = managedType.LastIndexOf('.');
+        if (lastIndexOfDot == -1)
+        {
+            typeName = managedType.ToString();
+            @namespace = string.Empty;
+        }
+        else
+        {
+            typeName = managedType.Slice(lastIndexOfDot + 1).ToString();
```
SimonCropp:
```suggestion
            typeName = managedType[(lastIndexOfDot + 1)..].ToString();
```

src/Platform/Microsoft.Testing.Extensions.VSTestBridge/ObjectModel/ObjectModelConverters.cs
```diff
@@ -238,22 +215,88 @@ internal static void FixUpTestCase(this TestCase testCase, string? testAssemblyP
...
+            @namespace = managedType.Slice(0, lastIndexOfDot).ToString();
```
SimonCropp:
```suggestion
            @namespace = managedType[..lastIndexOfDot].ToString();
```

src/Platform/Microsoft.Testing.Extensions.VSTestBridge/ObjectModel/ObjectModelConverters.cs
```diff
@@ -238,22 +215,88 @@ internal static void FixUpTestCase(this TestCase testCase, string? testAssemblyP
...
+        string @namespace;
+        string typeName;
+        string methodName;
+        string[] parameterTypeFullNames = Array.Empty<string>();
+        string returnTypeFullName = string.Empty;
+
+        int indexOfParen = managedMethod.IndexOf('(');
+        if (indexOfParen != -1)
+        {
+            parameterTypeFullNames = GetParameterTypes(managedMethod.Slice(indexOfParen + 1));
```
SimonCropp:
```suggestion
            parameterTypeFullNames = GetParameterTypes(managedMethod[(indexOfParen + 1)..]);
```

src/Platform/Microsoft.Testing.Extensions.VSTestBridge/ObjectModel/ObjectModelConverters.cs
```diff
@@ -238,22 +215,88 @@ internal static void FixUpTestCase(this TestCase testCase, string? testAssemblyP
...
+            methodName = managedMethod.Slice(0, indexOfParen).ToString();
```
SimonCropp:
```suggestion
            methodName = managedMethod[..indexOfParen].ToString();
```

src/Platform/Microsoft.Testing.Extensions.VSTestBridge/ObjectModel/ObjectModelConverters.cs
```diff
@@ -238,22 +215,88 @@ internal static void FixUpTestCase(this TestCase testCase, string? testAssemblyP
...
+        string? managedType = testCase.GetPropertyValue<string>(ManagedTypeProperty, defaultValue: null);
+        string? managedMethod = testCase.GetPropertyValue<string>(ManagedMethodProperty, defaultValue: null);
+        // NOTE: ManagedMethod, in case of MSTest, will have the parameter types.
+        // So, we prefer using it to display the parameter types in Test Explorer.
+        return !RoslynString.IsNullOrEmpty(managedType) && !RoslynString.IsNullOrEmpty(managedMethod)
+            ? TryGetMethodIdentifierPropertyFromManagedTypeAndManagedMethod(managedType, managedMethod, out methodIdentifierProperty)
+            : TryGetMethodIdentifierPropertyFromFullyQualifiedName(testCase.FullyQualifiedName, out methodIdentifierProperty);
+    }
+
+    private static bool TryGetMethodIdentifierPropertyFromManagedTypeAndManagedMethod(
```
SimonCropp: method always returns true. may as well directly return the TestMethodIdentifierProperty instance

src/Platform/Microsoft.Testing.Extensions.VSTestBridge/ObjectModel/ObjectModelConverters.cs
```diff
@@ -238,22 +215,88 @@ internal static void FixUpTestCase(this TestCase testCase, string? testAssemblyP
...
+
+    private static bool TryGetMethodIdentifierPropertyFromFullyQualifiedName(ReadOnlySpan<char> fullyQualifiedName, [NotNullWhen(true)] out TestMethodIdentifierProperty? methodIdentifierProperty)
+    {
+        int indexOfParen = fullyQualifiedName.IndexOf('(');
+
+        int lastIndexOfDotBeforeParen = indexOfParen == -1
             ? fullyQualifiedName.LastIndexOf('.')
-            : fullyQualifiedName.LastIndexOf('.', openBracketIndex - 1);
-        if (lastDotIndexBeforeOpenBracket <= 0)
+            : fullyQualifiedName.Slice(0, indexOfParen).LastIndexOf('.');
```
SimonCropp:
```suggestion
            : fullyQualifiedName[..indexOfParen].LastIndexOf('.');
```

src/Platform/Microsoft.Testing.Extensions.VSTestBridge/ObjectModel/ObjectModelConverters.cs
```diff
@@ -238,22 +215,88 @@ internal static void FixUpTestCase(this TestCase testCase, string? testAssemblyP
...
+        if (lastIndexOfDotBeforeParen == -1)
         {
+            methodIdentifierProperty = null;
             return false;
         }
 
-        fullyQualifiedType = fullyQualifiedName[..lastDotIndexBeforeOpenBracket];
-        return true;
+        return TryGetMethodIdentifierPropertyFromManagedTypeAndManagedMethod(
+            managedType: fullyQualifiedName.Slice(0, lastIndexOfDotBeforeParen),
```
SimonCropp:
```suggestion
            managedType: fullyQualifiedName[..lastIndexOfDotBeforeParen],
```

src/Platform/Microsoft.Testing.Extensions.VSTestBridge/ObjectModel/ObjectModelConverters.cs
```diff
@@ -238,22 +215,88 @@ internal static void FixUpTestCase(this TestCase testCase, string? testAssemblyP
...
+            managedMethod: fullyQualifiedName.Slice(lastIndexOfDotBeforeParen + 1),
```
SimonCropp:
```suggestion
            managedMethod: fullyQualifiedName[(lastIndexOfDotBeforeParen + 1)..],
```

src/Platform/Microsoft.Testing.Extensions.VSTestBridge/ObjectModel/ObjectModelConverters.cs
```diff
@@ -238,22 +215,88 @@ internal static void FixUpTestCase(this TestCase testCase, string? testAssemblyP
...
+            out methodIdentifierProperty);
+    }
+
+    private static string[] GetParameterTypes(ReadOnlySpan<char> afterOpenParen)
+    {
+        if (afterOpenParen[afterOpenParen.Length - 1] != ')')
```
SimonCropp:
```suggestion
if (afterOpenParen[^1] != ')')
```

src/Platform/Microsoft.Testing.Extensions.VSTestBridge/ObjectModel/ObjectModelConverters.cs
```diff
@@ -238,22 +215,88 @@ internal static void FixUpTestCase(this TestCase testCase, string? testAssemblyP
...
+        {
+            // TODO: Maybe better to throw?
+            return Array.Empty<string>();
+        }
+
+        afterOpenParen = afterOpenParen.Slice(0, afterOpenParen.Length - 1).Trim();
+        return afterOpenParen.Length == 0
+            ? Array.Empty<string>()
+            : afterOpenParen.ToString().Split(',', StringSplitOptions.None);
```
SimonCropp:
```suggestion
            : afterOpenParen.ToString().Split(',');
```

src/Platform/Microsoft.Testing.Extensions.VSTestBridge/ObjectModel/ObjectModelConverters.cs
```diff
@@ -238,22 +215,88 @@ internal static void FixUpTestCase(this TestCase testCase, string? testAssemblyP
...
+
+    private static string[] GetParameterTypes(ReadOnlySpan<char> afterOpenParen)
+    {
+        if (afterOpenParen[afterOpenParen.Length - 1] != ')')
+        {
+            // TODO: Maybe better to throw?
+            return Array.Empty<string>();
+        }
+
+        afterOpenParen = afterOpenParen.Slice(0, afterOpenParen.Length - 1).Trim();
```
SimonCropp:
```suggestion
        afterOpenParen = afterOpenParen.Slice(0, afterOpenParen.Length - 1).Trim();
```

src/Platform/Microsoft.Testing.Extensions.VSTestBridge/ObjectModel/ObjectModelConverters.cs
```diff
@@ -238,22 +215,88 @@ internal static void FixUpTestCase(this TestCase testCase, string? testAssemblyP
...
+        return afterOpenParen.Length == 0
+            ? Array.Empty<string>()
```
SimonCropp:
```suggestion
            ? []
```

src/Platform/Microsoft.Testing.Platform/ServerMode/ClientHelpers.cs
```diff
@@ -0,0 +1,20 @@
...
+
+namespace Microsoft.Testing.Platform.ServerMode;
+
+internal static class ClientHelpers
+{
+    internal static string? ClientName { get; set; }
+
+    internal static Version? ClientVersion { get; set; }
+
+    private static readonly Version? MinimumVisualStudioClientVersionWithFixedLocationImplementation = new("1.0.1");
```
SimonCropp:
```suggestion
    private static readonly Version MinimumVisualStudioClientVersionWithFixedLocationImplementation = new("1.0.1");
```

test/UnitTests/Microsoft.Testing.Extensions.VSTestBridge.UnitTests/ObjectModel/ObjectModelConvertersTests.cs
```diff
@@ -17,17 +16,14 @@ namespace Microsoft.Testing.Extensions.VSTestBridge.UnitTests.ObjectModel;
 [TestClass]
 public sealed class ObjectModelConvertersTests
```
Youssef1313, line 19: Initially I deleted some tests here when I was removing vstestProvider completely. In the current PR status, some logic is brought back, and I need to plug the tests back per the updated logic (we add the vstest properties under vstestProvider named capability + JsonRpc server mode).

src/Adapter/MSTest.TestAdapter/TestingPlatformAdapter/TestApplicationBuilderExtensions.cs
```diff
@@ -34,7 +45,7 @@ public static void AddMSTest(this ITestApplicationBuilder testApplicationBuilder
         testApplicationBuilder.AddRunSettingsEnvironmentVariableProvider(extension);
         testApplicationBuilder.RegisterTestFramework(
             serviceProvider => new TestFrameworkCapabilities(
-                new VSTestBridgeExtensionBaseCapabilities(),
+                new MSTestCapabilities(),
```
fhnaseer, line 48: Do we really need this? Removing VSTestBridge will not be enough?

Youssef1313: @fhnaseer We still use the bridge, but only don't want to use vstestProvider.

src/Platform/Microsoft.Testing.Extensions.VSTestBridge/ObjectModel/ObjectModelConverters.cs
```diff
@@ -34,7 +49,13 @@ public static TestNode ToTestNode(this TestCase testCase, bool isTrxEnabled, ICl
...
         };
 
-        CopyVSTestProperties(testCase.Properties, testNode, testCase, testCase.GetPropertyValue, isTrxEnabled, client);
+        // This will be false for Expecto and NUnit currently, as they don't provide ManagedType/ManagedMethod.
+        if (TryGetMethodIdentifierProperty(testCase, out TestMethodIdentifierProperty? methodIdentifierProperty))
+        {
+            testNode.Properties.Add(methodIdentifierProperty);
+        }
+
+        CopyVSTestProperties(testCase.Properties, testNode, testCase, testCase.GetPropertyValue, isTrxEnabled, ShouldAddVSTestProviderProperties(serviceProvider));
```
fhnaseer: Suggestion: 

```
if (TryGetMethodIdentifierProperty(testCase, out TestMethodIdentifierProperty? methodIdentifierProperty))
    testNode.Properties.Add(methodIdentifierProperty);
else if (ShouldAddVSTestProviderProperties(serviceProvider)) // check here about vstest properties rather than inside CopyVSTestProperties function
   CopyVSTestProperties(testCase.Properties, testNode, testCase, testCase.GetPropertyValue, isTrxEnabled)
else
  throw exception?,
```

Maybe a check as well. If test framework sends both managed and vstest properties, then we can stop or warn them,

src/Platform/Microsoft.Testing.Extensions.VSTestBridge/ObjectModel/ObjectModelConverters.cs
```diff
@@ -44,12 +65,10 @@ public static TestNode ToTestNode(this TestCase testCase, bool isTrxEnabled, ICl
     }
 
     private static void CopyVSTestProperties(IEnumerable<TestProperty> testProperties, TestNode testNode, TestCase testCase, Func<TestProperty, object?> getPropertyValue,
-        bool isTrxEnabled, IClientInfo client)
+        bool isTrxEnabled, bool addVSTestProviderProperties)
```
fhnaseer: Not like the idea of having `bool addVSTestProviderProperties` here. By just looking at the name `CopyVSTestProperties`, this function should not do anything if `addVSTestProviderProperties = false`. So, why someone should call it with `false`?

I would prefer to have 
```
if (addVstestProperties)
   CopyVSTestProperties(...);
```

rather than

```
CopyVSTestProperties(..., false);
```

test/UnitTests/Microsoft.Testing.Platform.UnitTests/ServerMode/FormatterUtilitiesTests.cs
```diff
@@ -188,7 +188,7 @@ private static void AssertSerialize(Type type, string instanceSerialized)
...
+            Assert.AreEqual("""{"node":{"uid":"uid","display-name":"DisplayName","traits":[{"testmetadata-key":"testmetadata-value"}],"array-key":["1","2"],"standardError":"textProperty2","standardOutput":"textProperty","time.start-utc":"2023-01-01T01:01:01.0000000+00:00","time.stop-utc":"2023-01-01T01:01:01.0000000+00:00","time.duration-ms":0,"location.type":"namespace.typeName","location.method":"methodName(param1,param2)","location.file":"filePath","location.line-start":1,"location.line-end":2,"key":"value","node-type":"action","execution-state":"failed","error.message":"sample","error.stacktrace":"","assert.actual":"","assert.expected":""},"parent":"parent-uid"}""".Replace(" ", string.Empty), instanceSerialized, because);
```
fhnaseer, line 191: "namespace.typeName" => is it part of protocol?

src/Platform/Microsoft.Testing.Extensions.VSTestBridge/ObjectModel/ObjectModelConverters.cs
Youssef1313: `MSTestDiscoverer.TestCategory` needs to be copied regardless. It could be separated out of CopyVSTestProperties though, but we do it while we are looping already

src/Platform/Microsoft.Testing.Extensions.VSTestBridge/ObjectModel/ObjectModelConverters.cs
Youssef1313: Same here.

`MSTestDiscoverer.TestCategory` needs to be copied regardless.

test/UnitTests/Microsoft.Testing.Platform.UnitTests/ServerMode/FormatterUtilitiesTests.cs
Youssef1313: That's how `location.type` is implemented on TE, so we are following it now.

src/Platform/Microsoft.Testing.Extensions.VSTestBridge/ObjectModel/ObjectModelConverters.cs
fhnaseer: A suggestion then. Maybe we can rename CopyVSTestProperties to something generic CopyProperties and then pass different flags `ShouldAddVSTestProviderProperties`, `ShouldCopySomething`.

Youssef1313: The properties are still "VSTest" properties though. I went ahead and extracted the VSTest properties that we always copy (i.e, those not specific to vstestProvider capabilitiy) outside of this method, renamed the method to CopyVSTestProviderProperties, and removed the `addVSTestProviderProperties` parameter and made it a caller check. I don't know if it makes it harder to review though.

src/Platform/Microsoft.Testing.Extensions.VSTestBridge/Capabilities/VSTestBridgeExtensionBaseCapabilities.cs
```diff
@@ -9,6 +9,10 @@ namespace Microsoft.Testing.Extensions.VSTestBridge.Capabilities;
 /// <summary>
 /// The VSTest bridged test framework capabilities.
 /// </summary>
+// NOTE: MSTest no longer uses this, as we don't want to use the vstestProvider.
+// Only NUnit and Expecto use this.
+// https://github.com/nunit/nunit3-vs-adapter/blob/3d0f824243aaaeb85621d3c7dddc92e7a7c45097/src/NUnitTestAdapter/TestingPlatformAdapter/TestApplicationBuilderExtensions.cs#L20
+// https://github.com/YoloDev/YoloDev.Expecto.TestSdk/blob/0d1a3eadd65b605f61bb01d302f28382be76b8ac/src/YoloDev.Expecto.TestSdk/TestApplicationHelpers.fs#L16
 public sealed class VSTestBridgeExtensionBaseCapabilities : ITrxReportCapability, IVSTestFlattenedTestNodesReportCapability, INamedFeatureCapability
```
Evangelink, line 16: Maybe mark it as obsolete? Expecto and NUnit auhtors are super friendly so if we do the PR they would update.

src/Platform/Microsoft.Testing.Extensions.VSTestBridge/ObjectModel/FrameworkHandlerAdapter.cs
```diff
@@ -30,14 +29,14 @@ internal sealed class FrameworkHandlerAdapter : IFrameworkHandle
...
-    private readonly IClientInfo _clientInfo;
     private readonly CancellationToken _cancellationToken;
     private readonly bool _isTrxEnabled;
     private readonly MessageLoggerAdapter _comboMessageLogger;
     private readonly string _testAssemblyPath;
+    private readonly IServiceProvider _serviceProvider;
 
-    public FrameworkHandlerAdapter(VSTestBridgedTestFrameworkBase adapterExtensionBase, TestSessionContext session, IClientInfo clientInfo, string[] testAssemblyPaths,
-        ITestApplicationModuleInfo testApplicationModuleInfo, ILoggerFactory loggerFactory, IMessageBus messageBus, IOutputDevice outputDevice,
+    public FrameworkHandlerAdapter(VSTestBridgedTestFrameworkBase adapterExtensionBase, TestSessionContext session, string[] testAssemblyPaths,
```
Evangelink: My idea was to avoid (when possible) to pas the `IServiceProvider` because it's hiding the real deps (I don't know what to mock and what are the deps of the object looking at the ctor).

src/Platform/Microsoft.Testing.Extensions.VSTestBridge/ObjectModel/ObjectModelConverters.cs
```diff
@@ -43,84 +82,85 @@ public static TestNode ToTestNode(this TestCase testCase, bool isTrxEnabled, ICl
...
-            // ID is defined on TraitCollection but is internal so again we copy the string here.
-            if (property.Id == "TestObject.Traits"
-                && getPropertyValue(property) is KeyValuePair<string, string>[] traits && traits.Length > 0)
+    private static void CopyVSTestProviderProperties(IEnumerable<TestProperty> testProperties, TestNode testNode, Func<TestProperty, object?> getPropertyValue)
+    {
+        foreach (TestProperty property in testProperties)
+        {
+            // If vstestProvider is enabled (only known to be true for NUnit and Expecto so far), and we are running server mode in IDE (not dotnet test),
+            // we add these stuff.
+            // Once NUnit and Expecto allow us to move forward and remove vstestProvider, we can remove this logic and get rid of the whole vstestProvider capability.
```
Evangelink, line 118: It's implicit but I think it would deserve to be made more explicit - you are not relying on client being VS/VSCode but you use the capability because we pass different properties for the client based check.

src/Platform/Microsoft.Testing.Extensions.VSTestBridge/ObjectModel/ObjectModelConverters.cs
```diff
@@ -43,84 +82,85 @@ public static TestNode ToTestNode(this TestCase testCase, bool isTrxEnabled, ICl
...
+            {
+                testNode.Properties.Add(new SerializableKeyValuePairStringProperty("vstest.original-executor-uri", originalExecutorUri.AbsoluteUri));
             }
 
-            // TPv2 is doing some special handling for MSTest... we should probably do the same.
-            // See https://github.com/microsoft/vstest/blob/main/src/Microsoft.TestPlatform.Extensions.TrxLogger/Utility/Converter.cs#L66-L70
-            else if (property.Id == "MSTestDiscoverer.TestCategory"
-                && getPropertyValue(property) is string[] mstestCategories && mstestCategories.Length > 0)
+            // The TP object holding the hierarchy property is defined on adapter utilities and we don't want to enforce that dependency
+            // so instead I use the string ID copied from TP.
```
Evangelink: You have added the adapter utilities deps

src/Platform/Microsoft.Testing.Extensions.VSTestBridge/ObjectModel/ObjectModelConverters.cs
```diff
@@ -43,84 +82,85 @@ public static TestNode ToTestNode(this TestCase testCase, bool isTrxEnabled, ICl
...
+        var testNode = testResult.TestCase.ToTestNode(isTrxEnabled, serviceProvider, testResult.DisplayName);
+
+        CopyCategoryAndTraits(testResult, testNode, isTrxEnabled);
+
+        bool addVSTestProviderProperties = ShouldAddVSTestProviderProperties(serviceProvider);
+        if (addVSTestProviderProperties)
+        {
+            // TODO: This call might be unnecessary.
+            // All the relevant properties should be on TestCase, not TestResult.
+            // And properties on TestCase where already copied as part of ToTestNode call above.
```
Evangelink: Nit:

```suggestion
            // And properties on TestCase were already copied as part of ToTestNode call above.
```

src/Platform/Microsoft.Testing.Extensions.VSTestBridge/ObjectModel/ObjectModelConverters.cs
```diff
@@ -238,6 +286,45 @@ internal static void FixUpTestCase(this TestCase testCase, string? testAssemblyP
...
+        parameterTypes ??= [];
+
+        ManagedNameParser.ParseManagedTypeName(managedType, out string @namespace, out string typeName);
+
+        // In the context of the VSTestBridge where we only have access to VSTest object model, we cannot determine ReturnTypeFullName.
+        // For now, we lose this bit of information.
+        // If really needed in the future, we can introduce a VSTest property to hold this info.
+        // But the eventual goal should be to stop using the VSTestBridge altogether.
+        // TODO: For AssemblyFullName, can we use Assembly.GetEntryAssembly().FullName?
+        // Or alternatively, does VSTest object model expose the assembly full name somewhere?
```
Evangelink, line 321: @nohwnd can you comment here pls?

src/Platform/Microsoft.Testing.Extensions.VSTestBridge/ObjectModel/ObjectModelConverters.cs
```diff
@@ -238,6 +286,45 @@ internal static void FixUpTestCase(this TestCase testCase, string? testAssemblyP
...
+        }
+
+        parameterTypes ??= [];
+
+        ManagedNameParser.ParseManagedTypeName(managedType, out string @namespace, out string typeName);
+
+        // In the context of the VSTestBridge where we only have access to VSTest object model, we cannot determine ReturnTypeFullName.
+        // For now, we lose this bit of information.
+        // If really needed in the future, we can introduce a VSTest property to hold this info.
+        // But the eventual goal should be to stop using the VSTestBridge altogether.
```
Evangelink, line 319: I think it's currently unused so not big deal

src/Platform/Microsoft.Testing.Platform/Messages/TestNodeProperties.cs
```diff
@@ -314,7 +314,7 @@ public sealed record TestFileLocationProperty(string FilePath, LinePositionSpan
 /// </summary>
 /// <param name="AssemblyFullName">Assembly full name.</param>
 /// <param name="Namespace">Namespace.</param>
-/// <param name="TypeName">Type name.</param>
+/// <param name="TypeName">Type name in metadata format, not including the namespace. Generics are represented by backtick followed by arity. Nested types are represented by <c>+</c>.</param>
```
Evangelink, line 317: Please add a link to the spec

src/Platform/Microsoft.Testing.Platform/ServerMode/JsonRpc/Json/Json.cs
```diff
@@ -163,8 +163,14 @@ public Json(Dictionary<Type, JsonSerializer>? serializers = null, Dictionary<Typ
...
                 if (property is TestMethodIdentifierProperty testMethodIdentifierProperty)
                 {
-                    properties.Add(("location.namespace", testMethodIdentifierProperty.Namespace));
-                    properties.Add(("location.type", testMethodIdentifierProperty.TypeName));
+                    string locationType = testMethodIdentifierProperty.TypeName;
+
+                    if (!RoslynString.IsNullOrEmpty(testMethodIdentifierProperty.Namespace))
+                    {
+                        locationType = $"{testMethodIdentifierProperty.Namespace}.{testMethodIdentifierProperty.TypeName}";
+                    }
```
Evangelink:
```suggestion
                    string locationType = RoslynString.IsNullOrEmpty(testMethodIdentifierProperty.Namespace)
                        ? testMethodIdentifierProperty.TypeName
                        : $"{testMethodIdentifierProperty.Namespace}.{testMethodIdentifierProperty.TypeName}";
```

src/Platform/Microsoft.Testing.Platform/ServerMode/JsonRpc/Json/Json.cs
```diff
@@ -163,8 +163,14 @@ public Json(Dictionary<Type, JsonSerializer>? serializers = null, Dictionary<Typ
...
+
+                    properties.Add(("location.type", locationType));
```
Evangelink:
```suggestion
                    properties.Add(("location.type", RoslynString.IsNullOrEmpty(testMethodIdentifierProperty.Namespace)
                        ? testMethodIdentifierProperty.TypeName
                        : $"{testMethodIdentifierProperty.Namespace}.{testMethodIdentifierProperty.TypeName}"));
```

src/Platform/Microsoft.Testing.Platform/ServerMode/JsonRpc/SerializerUtilities.cs
```diff
@@ -225,8 +225,14 @@ static SerializerUtilities()
...
-                        properties["location.namespace"] = testMethodIdentifierProperty.Namespace;
-                        properties["location.type"] = testMethodIdentifierProperty.TypeName;
+                        string locationType = testMethodIdentifierProperty.TypeName;
+
+                        if (!RoslynString.IsNullOrEmpty(testMethodIdentifierProperty.Namespace))
+                        {
+                            locationType = $"{testMethodIdentifierProperty.Namespace}.{testMethodIdentifierProperty.TypeName}";
+                        }
+
+                        properties["location.type"] = locationType;
```
Evangelink: Same suggestion as above (depending on the one you picked)

src/Platform/Microsoft.Testing.Extensions.VSTestBridge/ObjectModel/FrameworkHandlerAdapter.cs
Youssef1313: Ok will revert back to explicit interfaces

src/Platform/Microsoft.Testing.Extensions.VSTestBridge/ObjectModel/ObjectModelConverters.cs
Youssef1313: I'm not following, but my idea is that this shouldn't be specific to VS/VSCode

src/Platform/Microsoft.Testing.Extensions.VSTestBridge/ObjectModel/ObjectModelConverters.cs
Youssef1313: Yup I added it after I wrote this comment so that we can parse managed type/method 😢

src/Platform/Microsoft.Testing.Platform/Messages/TestNodeProperties.cs
Youssef1313: I want to avoid linking to VSTest stuff. The current documentation for TypeName seems reasonable to me though.

src/Platform/Microsoft.Testing.Extensions.VSTestBridge/Capabilities/VSTestBridgeExtensionBaseCapabilities.cs
Youssef1313: The eventual goal for me at least is to deprecate the whole VSTestBridge.

For now we can't mark as obsolete because of NUnit at least. I have an open issue there, so until it's resolved we will keep this.

Youssef1313: The eventual goal for me at least is to deprecate the whole VSTestBridge.

For now we can't mark as obsolete because of NUnit at least. I have an open issue there, so until it's resolved we will keep this.

Youssef1313:
@Evangelink I addressed your review comments.

src/Platform/Microsoft.Testing.Extensions.VSTestBridge/ObjectModel/ObjectModelConverters.cs
nohwnd: We don't have assembly full name in the data, the test cases don't have to be produced from the current process or source.  

If we want to populate that data, it might be better to ask for the actual type in the namespace, and grab the assembly information from there. Because then we don't assume the entry assembly is non-null, and we don't assume it is the same as the source of the test, we just assume that the type is available, and can decide to fail if it is not, or not fail and do something else.

Youssef1313: > ask for the actual type in the namespace

I can't see a good way to do it. I think passing null should be sufficient for now, and hopefully we may be able to drop VSTestBridge at some point later.

nohwnd (approved):

PR #5424, opened 2025-04-09
Condition all logic of dotnet test support under TestingPlataformDotnetTestSupport

### LINKED ISSUE #5271: `Microsoft.Testing.Platform.MSBuild.CustomTestTarget.targets` shouldn't be imported if MTP not enabled
If MTP isn't opted-in, `Microsoft.Testing.Platform.MSBuild.CustomTestTarget.targets` shouldn't be imported.

A user defining `Test` target for all their projects via props will break when we import our CustomTestTarget. The target has a condition that will evaluate to false, then MSBuild will complain:

```
error MSB4057: The target "Test" does not exist in the project.
```

Youssef1313:
Fixes #5271


Youssef1313:
Part of the logic here is related to `-t:Test` and not `dotnet test`, which is currently supported even without TestingPlatformDotnetTestSupport. So this is a risky change on its current form.

PR #5429, opened 2025-04-10
Conditional expression can be rewritten as null-coalescing﻿ and mark as warning

SimonCropp:


Youssef1313 (approved):
LGTM. Thanks @SimonCropp

PR #5438, opened 2025-04-10
Remove redudnant internals visible to in ms test.source generation

SimonCropp:


Evangelink (commented):
I am not a super big fan of having public types when it's not needed - we have too many issues and problems linked to public types where we don't know if they are used preventing us from doing any change.

I also think IVTs kinda suck but from source to test it feels ok to me - at least preferable than public.

nohwnd:
Keeping internal, and replacing by #5561

PR #5441, opened 2025-04-10
Dont generate documentation for tests and samples

SimonCropp:
no value and slows down the build


nohwnd (approved):

Youssef1313:
> slows down the build

I'm fine not generating documentation files for samples and tests, but also curious if you have some numbers or profiling results, or what exactly is slow here? Is it the `SA0001` analyzer being slow? Or is it really the compiler that's being slower?

SimonCropp:
@Youssef1313 i tried to get some numbers. but the compiler does not break out how much time generating docs takes. and there is too much variance between runs to give number with that approach.

PR #5454, opened 2025-04-11
Use `GetFileNameWithoutExtension` for crashdump file name to be consistent with hangdump

### LINKED ISSUE #5036: Inconsistent dump file name between crash dump and hang dump
https://github.com/microsoft/testfx/blob/13b583305c70b6c0959431e9a3b90a17120c50e3/src/Platform/Microsoft.Testing.Extensions.CrashDump/CrashDumpEnvironmentVariableProvider.cs#L117

https://github.com/microsoft/testfx/blob/13b583305c70b6c0959431e9a3b90a17120c50e3/src/Platform/Microsoft.Testing.Extensions.HangDump/HangDumpProcessLifetimeHandler.cs#L88

One is using `Path.GetFileName` while the other is using `Path.GetFileNameWithoutExtension`

Youssef1313:
Fixes #5036


Evangelink (approved):
LGTM note that this could be considered as breaking change as any tool relying on the expected dump name would be broken.

nohwnd (approved):

PR #5459, opened 2025-04-13
Remove some redundant casts and mark as a warning for rider and R#

SimonCropp:


src/Platform/Microsoft.Testing.Extensions.Retry/RetryExecutionFilterFactory.cs
```diff
@@ -38,7 +38,7 @@ public Task<bool> IsEnabledAsync()
         _retryFailedTestsLifecycleCallbacks = _serviceProvider.GetRequiredService<RetryLifecycleCallbacks>();
         if (_retryFailedTestsLifecycleCallbacks.FailedTestsIDToRetry?.Length > 0)
         {
-            return (true, (ITestExecutionFilter?)new TestNodeUidListFilter(_retryFailedTestsLifecycleCallbacks.FailedTestsIDToRetry
+            return (true, new TestNodeUidListFilter(_retryFailedTestsLifecycleCallbacks.FailedTestsIDToRetry
```
Youssef1313, line 41: @CyrusNajmabadi I was expecting that we get IDE0004 diagnostics here per:

https://github.com/microsoft/testfx/blob/8d07ced3e8cf5ec03127c4ef9f401713ca4023ed/.editorconfig#L169-L170

Are the cases here "by-design" for some reason? Or is it a false negative bug?

Youssef1313 (approved):
Approving, but surprised this is not caught by IDE0004

PR #5460, opened 2025-04-13
Fix some nullability in tests

SimonCropp:

### DIFF
--- a/test/IntegrationTests/MSTest.IntegrationTests/Utilities/TestCaseFilterFactory.cs
+++ b/test/IntegrationTests/MSTest.IntegrationTests/Utilities/TestCaseFilterFactory.cs
@@ -13,9 +13,9 @@ namespace DiscoveryAndExecutionTests.Utilities;
 
 internal static class TestCaseFilterFactory
 {
-    private static readonly MethodInfo CachedGetMultiValueMethod = typeof(TestCaseFilterFactory).GetMethod(nameof(GetMultiValue), BindingFlags.Static | BindingFlags.NonPublic);
-    private static readonly MethodInfo CachedEqualsComparerMethod = typeof(TestCaseFilterFactory).GetMethod(nameof(EqualsComparer), BindingFlags.Static | BindingFlags.NonPublic);
-    private static readonly MethodInfo CachedContainsComparerMethod = typeof(TestCaseFilterFactory).GetMethod(nameof(ContainsComparer), BindingFlags.Static | BindingFlags.NonPublic);
+    private static readonly MethodInfo CachedGetMultiValueMethod = typeof(TestCaseFilterFactory).GetMethod(nameof(GetMultiValue), BindingFlags.Static | BindingFlags.NonPublic)!;
+    private static readonly MethodInfo CachedEqualsComparerMethod = typeof(TestCaseFilterFactory).GetMethod(nameof(EqualsComparer), BindingFlags.Static | BindingFlags.NonPublic)!;
+    private static readonly MethodInfo CachedContainsComparerMethod = typeof(TestCaseFilterFactory).GetMethod(nameof(ContainsComparer), BindingFlags.Static | BindingFlags.NonPublic)!;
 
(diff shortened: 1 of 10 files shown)


test/IntegrationTests/MSTest.IntegrationTests/Utilities/TestCaseFilterFactory.cs
```diff
@@ -13,9 +13,9 @@ namespace DiscoveryAndExecutionTests.Utilities;
 
 internal static class TestCaseFilterFactory
 {
-    private static readonly MethodInfo CachedGetMultiValueMethod = typeof(TestCaseFilterFactory).GetMethod(nameof(GetMultiValue), BindingFlags.Static | BindingFlags.NonPublic);
-    private static readonly MethodInfo CachedEqualsComparerMethod = typeof(TestCaseFilterFactory).GetMethod(nameof(EqualsComparer), BindingFlags.Static | BindingFlags.NonPublic);
-    private static readonly MethodInfo CachedContainsComparerMethod = typeof(TestCaseFilterFactory).GetMethod(nameof(ContainsComparer), BindingFlags.Static | BindingFlags.NonPublic);
+    private static readonly MethodInfo CachedGetMultiValueMethod = typeof(TestCaseFilterFactory).GetMethod(nameof(GetMultiValue), BindingFlags.Static | BindingFlags.NonPublic)!;
+    private static readonly MethodInfo CachedEqualsComparerMethod = typeof(TestCaseFilterFactory).GetMethod(nameof(EqualsComparer), BindingFlags.Static | BindingFlags.NonPublic)!;
+    private static readonly MethodInfo CachedContainsComparerMethod = typeof(TestCaseFilterFactory).GetMethod(nameof(ContainsComparer), BindingFlags.Static | BindingFlags.NonPublic)!;
```
Youssef1313, line 18: The suppressions are redundant here as we target .NET Framework where `GetMethod` isn't annotated.

test/IntegrationTests/TestAssets/DynamicDataTestProject/DynamicDataTests.cs
```diff
@@ -234,7 +234,7 @@ private static IEnumerable<object[]> GetExampleTestCases()
               }
             ]
             """;
-        return JsonConvert.DeserializeObject<ExampleTestCase[]>(json)
+        return JsonConvert.DeserializeObject<ExampleTestCase[]>(json)!
```
Youssef1313, line 237: Same here, it's redundant. I would avoid using `!` when it effectively does nothing (i.e, it doesn't suppress a warning)

test/UnitTests/Microsoft.Testing.Extensions.VSTestBridge.UnitTests/ObjectModel/ObjectModelConvertersTests.cs
```diff
@@ -141,7 +141,7 @@ public void ToTestNode_WhenTestResultHasFullyQualifiedTypeAndTrxEnabled_TestNode
 
         var testNode = testResult.ToTestNode(true, TestClient);
 
-        Assert.AreEqual(testNode.Properties.OfType<TrxExceptionProperty>()?.Length, 1);
+        Assert.AreEqual(testNode.Properties.OfType<TrxExceptionProperty>().Length, 1);
```
Youssef1313, line 144: After the removal of `?`, this is now producing:

> test/UnitTests/Microsoft.Testing.Extensions.VSTestBridge.UnitTests/ObjectModel/ObjectModelConvertersTests.cs(144,9): error MSTEST0017: (NETCORE_ENGINEERING_TELEMETRY=Build) Assertion arguments should be passed in the correct order. 'actual' and 'expected'/'notExpected' arguments have been swapped. (https://learn.microsoft.com/dotnet/core/testing/mstest-analyzers/mstest0017)

I believe this is a false negative case of the analyzer. I will open an issue to track this. Meanwhile, you can safely swap the order.

PR #5461, opened 2025-04-13
remoev some Multiple enumeration

SimonCropp:

### DIFF
--- a/src/Platform/Microsoft.Testing.Extensions.VSTestBridge/ObjectModel/ContextAdapterBase.cs
+++ b/src/Platform/Microsoft.Testing.Extensions.VSTestBridge/ObjectModel/ContextAdapterBase.cs
@@ -40,7 +40,7 @@ internal abstract class ContextAdapterBase
     // NOTE: Implementation is borrowed from VSTest
     // MSTest relies on this method existing and access it through reflection: https://github.com/microsoft/testfx/blob/main/src/Adapter/MSTest.TestAdapter/TestMethodFilter.cs#L115
     public ITestCaseFilterExpression? GetTestCaseFilter(
-        IEnumerable<string>? supportedProperties,
+        ICollection<string>? supportedProperties,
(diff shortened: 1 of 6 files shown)


src/Platform/Microsoft.Testing.Extensions.VSTestBridge/ObjectModel/ContextAdapterBase.cs
```diff
@@ -40,7 +40,7 @@ protected ContextAdapterBase(ICommandLineOptions commandLineOptions, IRunSetting
     // NOTE: Implementation is borrowed from VSTest
     // MSTest relies on this method existing and access it through reflection: https://github.com/microsoft/testfx/blob/main/src/Adapter/MSTest.TestAdapter/TestMethodFilter.cs#L115
     public ITestCaseFilterExpression? GetTestCaseFilter(
-        IEnumerable<string>? supportedProperties,
```
Youssef1313, line 43: This is a risky change to make given we do this:

https://github.com/microsoft/testfx/blob/c06959644eaf8132067c80d38d948f6df6c21327/src/Adapter/MSTest.TestAdapter/TestMethodFilter.cs#L124

src/Platform/Microsoft.Testing.Platform/Messages/PropertyBag.cs
```diff
@@ -230,11 +230,11 @@ public TProperty Single<TProperty>()
             throw new InvalidOperationException($"Could not find a property of type '{typeof(TProperty)}'.");
         }
 
-        IEnumerable<TProperty> matchingValues = _property is null ? Array.Empty<TProperty>() : _property.OfType<TProperty>();
+        TProperty[] matchingValues = _property is null ? [] : _property.OfType<TProperty>().ToArray();
```
Youssef1313, line 233: Can we have a benchmark for this? I'd personally avoid changing it without a benchmark showing clear benefit.

PR #5505, opened 2025-04-23
Avoid overwriting TestFailureException

### LINKED ISSUE #5165: Exception from class init can be swallowed if class cleanup also fails
Run the following test class:

```csharp
[TestClass]
public class TestClass
{
    [ClassInitialize]
    public static void ClassInit(TestContext _)
    {
        throw new Exception("Exception from class init");
    }

    [TestMethod]
    public void Test()
    {
    }

    [ClassCleanup]
    public static void ClassCleanup(TestContext _)
    {
        throw new Exception("Exception from class cleanup");
    }
}
```

It fails with:

```
Class Cleanup method TestClass.ClassCleanup failed. Error Message: System.Exception: Exception from class cleanup.
```

This is very likely the root cause of some flakiness we have in CI.

AB#2404442

Youssef1313:
Fixes #5165

### DIFF
--- a/src/Adapter/MSTest.TestAdapter/Execution/TestMethodInfo.cs
+++ b/src/Adapter/MSTest.TestAdapter/Execution/TestMethodInfo.cs
@@ -751,76 +751,23 @@ public class TestMethodInfo : ITestMethod
             return;
         }
 
-        // If the exception is already a `TestFailedException` we throw it as-is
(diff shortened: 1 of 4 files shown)


nohwnd:
Code looks good, still some failing tests though.

nohwnd (approved):

PR #5481, opened 2025-04-17
Enable azdo gh plugin

nohwnd:
Enables AzureDevOps error reporting to get errors directly in code where they happened for easier debugging.

Files that are changed:

image

Files that are unchanged, and multiple line messages:

image

Fixes #5518

PR #5513, opened 2025-04-24
Few improvements to AzDO extension

Youssef1313:
@nohwnd Can you please validate that this still works as expected?

We may also need at least few basic unit tests.

### DIFF
--- a/src/Platform/Microsoft.Testing.Extensions.AzureDevOps/AzureDevOpsReporter.cs
+++ b/src/Platform/Microsoft.Testing.Extensions.AzureDevOps/AzureDevOpsReporter.cs
@@ -119,10 +119,8 @@ internal sealed class AzureDevOpsReporter :
         }
 
         string stackTrace = exception.StackTrace;
-        int index = stackTrace.IndexOfAny(NewlineCharacters);
-        string firstLine = index == -1 ? stackTrace : stackTrace.Substring(0, index);
-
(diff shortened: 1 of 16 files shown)


src/Platform/Microsoft.Testing.Extensions.AzureDevOps/AzureDevOpsReporter.cs
```diff
@@ -119,10 +119,8 @@ private async Task WriteExceptionAsync(string? explanation, Exception? exception
         }
 
         string stackTrace = exception.StackTrace;
-        int index = stackTrace.IndexOfAny(NewlineCharacters);
-        string firstLine = index == -1 ? stackTrace : stackTrace.Substring(0, index);
-
-        (string Code, string File, int LineNumber)? location = GetStackFrameLocation(firstLine);
+        string[] lines = stackTrace.Split(NewlineCharacters, StringSplitOptions.RemoveEmptyEntries);
+        (string Code, string File, int LineNumber)? location = lines.Select(GetStackFrameLocation).FirstOrDefault(location => location is not null);
```
nohwnd, line 123: This will (in GetStackFrameLocation) find all files that exist (and will check for every line in stack trace), and then it will take just the first one found. 

2 issues with that:
1) we do more checks of the filesystem than necessary
2) won't work with deterministic build because the paths won't ever exist.

but both are okay, I will merge this PR for the other changes, and fix it in #5481 where I already handle deterministic build.

nohwnd (approved):

PR #5522, opened 2025-04-27
Enable HangDumpTests on macOS

Youssef1313:


nohwnd:
Is this to repro the problem with the netclient from dotnet/diagnostics?

Youssef1313:
@nohwnd Yup

Youssef1313:
@nohwnd What we could be missing here is https://github.com/microsoft/vstest/blob/f93fcd2943f39c2cbb65d25e3949e21c3190184c/src/Microsoft.TestPlatform.Extensions.BlameDataCollector/BlameCollector.cs#L157-L167

I'll give it a try later.

nohwnd:
Do we see the issue on net6 and older only?

According to the note in the link below that env variable is no longer necessary, and I did not find other mentions of it that would document it with the DOTNET_ prefix, which is what all new env variables do.

 https://github.com/dotnet/runtime/blob/main/docs/design/coreclr/botr/xplat-minidump-generation.md#macos

Youssef1313:
Hmm, indeed. 😕

PR #5528, opened 2025-04-28
Add TestMethodIdentifierProperty constructor with arity parameter

### LINKED ISSUE #5516: Obsolete `TestMethodIdentifierProperty` constructor and add new constructor with arity parameter
Today, `TestMethodIdentifierProperty` is declared as:

```csharp
public sealed record TestMethodIdentifierProperty(string AssemblyFullName, string Namespace, string TypeName, string MethodName, string[] ParameterTypeFullNames, string ReturnTypeFullName) : IProperty
```

This doesn't have a good place to specify arity. I suggest obsoleting the current constructor (but keeping it for binary compatibility), and add another constructor with int arity parameter.

So, for a method like:

```csharp
namespace MyNamespace1.MyNamespace2
{
    public class C<T>
    {
         public void M<T2>() { }
    }
}
```

- Namespace will be `MyNamespace1.MyNamespace2`
- TypeName will be ``C`1``
- MethodName will be `M`
- Arity will be 1

Youssef1313:
Fixes #5516


src/Platform/Microsoft.Testing.Platform/ServerMode/JsonRpc/Json/Json.cs
```diff
@@ -164,6 +164,14 @@ public Json(Dictionary<Type, JsonSerializer>? serializers = null, Dictionary<Typ
                     properties.Add(("location.method", testMethodIdentifierProperty.ParameterTypeFullNames.Length > 0
                         ? $"{testMethodIdentifierProperty.MethodName}({string.Join(",", testMethodIdentifierProperty.ParameterTypeFullNames)})"
                         : testMethodIdentifierProperty.MethodName));
+
+                    if (testMethodIdentifierProperty.MethodArity != 0)
+                    {
+                        // Avoid serializing location.method-arity if it's zero for performance.
+                        // Clients should assume the value is 0 when the property is not present.
+                        properties.Add(("location.method-arity", testMethodIdentifierProperty.MethodArity));
```
nohwnd: Do we do this in other places? we might be getting minimal benefit by doing it just here, at the price of making it hard to figure out the real content of the object from the sent data. Overall this feels like it should be a global setting that is handshaken by the protocol and should apply to any number or null consistently.

Youssef1313: @nohwnd I was thinking it may be too much to repeatedly send `location.method-arity` when it's 0 in 99% of cases. We don't do this in other places because none of the other properties we send today has that benefit. I'm fine to always send it though and only do the optimization if we really got perf data showing benefits. Do you prefer always sending it for now?

nohwnd: Yes, please send it, and add an issue for omitting default values from serialization please. I've done some measurement on this in the past (on vstest), and locally it was never an issue that would translate to huge time savings. I expect it might be different over network, but I would rather optimize centrally, than on case-by-case basis.

Youssef1313: @nohwnd Done

Youssef1313: @nohwnd Done

nohwnd (approved):

PR #5535, opened 2025-04-28
Show colors in AzDo and GH actions

nohwnd:
image


test/IntegrationTests/Microsoft.Testing.Platform.Acceptance.IntegrationTests/ExitOnProcessExitTests.cs
```diff
@@ -83,7 +84,7 @@ public sealed class TestAssetFixture() : TestAssetFixtureBase(AcceptanceFixture.
 using Microsoft.Testing.Platform.Extensions.Messages;
 using Microsoft.Testing.Platform.Extensions.TestFramework;
 
-if (args.Length == 0)
+if (!args.Contains("--exit-on-process-exit"))
```
nohwnd, line 87: This test assumed no args, but we now set --no-progress --no-ansi, so it never got setup, and was just sitting there stuck.

test/IntegrationTests/Microsoft.Testing.Platform.Acceptance.IntegrationTests/ExitOnProcessExitTests.cs
```diff
@@ -93,7 +94,7 @@ public sealed class TestAssetFixture() : TestAssetFixtureBase(AcceptanceFixture.
     Environment.SetEnvironmentVariable("WaitTestHost", mutexName);
     ProcessStartInfo processStartInfo = new();
     processStartInfo.FileName = currentEntryPoint;
-    processStartInfo.Arguments = $"--exit-on-process-exit {currentPid}";
+    processStartInfo.Arguments = $"--exit-on-process-exit {currentPid} --no-progress --no-ansi";
```
nohwnd, line 97: Adding the parameters here so we don't end up writing into the output.

test/IntegrationTests/Microsoft.Testing.Platform.Acceptance.IntegrationTests/ExitOnProcessExitTests.cs
```diff
@@ -48,7 +48,8 @@ public void ExitOnProcessExit_Succeed(string tfm)
         startTime = Stopwatch.StartNew();
         while (!process.HasExited)
         {
-            if (startTime.Elapsed.TotalSeconds > 55)
+            Thread.Sleep(1000);
```
nohwnd, line 51: Avoid the tight loop that checks for process exit to save cpu.

nohwnd:
So far I am only butchering the ansi output, I will need to:
- [ ] extract the progress rendering from non-ansi output
- [ ] extract progress rendering from ansi ouitput
- [ ] add parameter to ansi output to use appropriate progress renderer
- [ ] add test for the output
- [ ] add some test to write the output to azdo for quicker development and visual check
- [ ] do this also to dotnet test implementation in sdk so we can consume it here

nohwnd:
image
The colors render which is good.

src/Platform/Microsoft.Testing.Platform/OutputDevice/Terminal/AnsiTerminalTestProgressFrame.cs
```diff
@@ -37,7 +37,7 @@ public void AppendTestWorkerProgress(TestProgressState progress, RenderedProgres
 
         terminal.Append('[');
         charsTaken++;
-        terminal.SetColor(TerminalColor.Green);
+        terminal.SetColor(TerminalColor.DarkGreen);
```
fhnaseer, line 40: Suggestion: What if we have `Constants.OurOwnGreen = TerminalColor.DarkGreen` and then use `OurOwnGreen` in all places. For some strange reason if we have to change `TerminalColor.DarkGreen` to `TerminalColor.LightGreen` in future, then we can simply change just `OurOwnGreen` value,

src/Platform/Microsoft.Testing.Platform/OutputDevice/Terminal/NonAnsiTerminal.cs
```diff
@@ -2,151 +2,55 @@
...
-            throw new InvalidOperationException(PlatformResources.ConsoleIsAlreadyInBatchingMode);
-        }
-
-        bool lockTaken = false;
-
-        // We store Console.Out in a field to make sure we will be doing
-        // the Monitor.Exit call on the same instance.
-        _batchingLock = Console.Out;
+        _colorNotSupported ??= RuntimeInformation.IsOSPlatform(OSPlatform.Create("ANDROID")) ||
+                    RuntimeInformation.IsOSPlatform(OSPlatform.Create("IOS")) ||
```
fhnaseer: nit: indentation,

nohwnd (commented):
merge after snap to 3.9

src/Platform/Microsoft.Testing.Platform/OutputDevice/Terminal/AnsiTerminalTestProgressFrame.cs
nohwnd: I hear you, but I would like to skip this for now. I don't see a reason why we would change the color. If we would unify to 1 color identifier, I would rather do it as a public option that sets the color for success, warning and error, and similarly progress / result indicators (like +, x, ?) for the same thing. There has been previous asks for distinguishing inner tests from outer tests in the output.

Evangelink (approved):

PR #5548, opened 2025-04-30
Fix false positive of Assert.IsTrue(x == null) when x is a pointer type

### LINKED ISSUE #5543: MSTEST0037 fires for "unsafe" pointer types
## Describe the bug
The following code will generate MSTEST0037 which is incorrect, worse, the recommended solution (Created by the "fixer") is a SYNTAX error. 
``` C#
unsafe
{
    byte* foo = null;
    Assert.IsTrue(foo is null);
}
```

## Steps To Reproduce
Add the above code to any test method

## Expected behavior
No errors/warnings/recommendations to change things. Ideally, `Assert.IsNull(foo)` would just work barring that, the workaround should NOT trigger more warnings requiring suppressions.

## Additional context
This is broken functionality where a workaround for that is causing more issues.

Youssef1313:
Fixes #5543

### DIFF
--- a/src/Analyzers/MSTest.Analyzers/UseProperAssertMethodsAnalyzer.cs
+++ b/src/Analyzers/MSTest.Analyzers/UseProperAssertMethodsAnalyzer.cs
@@ -185,69 +185,82 @@ internal sealed class UseProperAssertMethodsAnalyzer : DiagnosticAnalyzer
         }
     }
 
-    private static bool IsIsNullPattern(IOperation operation, [NotNullWhen(true)] out SyntaxNode? expressionUnderTest)
+    private static bool IsIsNullPattern(IOperation operation, [NotNullWhen(true)] out SyntaxNode? expressionUnderTest, out ITypeSymbol? typeOfExpressionUnderTest)
     {
         if (operation is IIsPatternOperation { Pattern: IConstantPatternOperation { Value: { } constantPatternValue } } isPatternOperation &&
             constantPatternValue.WalkDownConversion() is ILiteralOperation { ConstantValue: { HasValue: true, Value: null } })
         {
             expressionUnderTest = isPatternOperation.Value.Syntax;
+            typeOfExpressionUnderTest = isPatternOperation.Value.WalkDownConversion().Type;
             return true;
         }
 
         expressionUnderTest = null;
+        typeOfExpressionUnderTest = null;
         return false;
     }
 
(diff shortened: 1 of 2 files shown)


src/Analyzers/MSTest.Analyzers/UseProperAssertMethodsAnalyzer.cs
```diff
@@ -185,82 +185,102 @@ private static void AnalyzeInvocationOperation(OperationAnalysisContext context,
...
 
         expressionUnderTest = null;
+        typeOfExpressionUnderTest = null;
         return false;
     }
 
-    private static NullCheckStatus RecognizeNullCheck(IOperation operation, /*We cannot express this is not null when NullCheckStatis is not Unknown*/ out SyntaxNode? expressionUnderTest)
+    private static NullCheckStatus RecognizeNullCheck(
+        IOperation operation,
+        /*We cannot express this is not null when NullCheckStatis is not Unknown*/ out SyntaxNode? expressionUnderTest,
```
nohwnd: Too many nots in this comment for me to understand. What did you mean there?

(also typo in NullCheckStatus)

test/UnitTests/MSTest.Analyzers.UnitTests/UseProperAssertMethodsAnalyzerTests.cs
```diff
@@ -59,6 +61,105 @@ await VerifyCS.VerifyCodeFixAsync(
             fixedCode);
     }
 
+    [TestMethod]
+    public async Task WhenPointerTypes()
```
nohwnd: This test needs a more descriptive name? Or no "When" ?

src/Analyzers/MSTest.Analyzers/UseProperAssertMethodsAnalyzer.cs
Youssef1313: @nohwnd What I mean is:

If the returned NullCheckStatus is something other than Unknown, then `expressionUnderTest` is guaranteed to be non-null. But with the current attributes used for nullability analysis, we cannot express it. So the parameter is declared to be nullable. Ideally we would want something like `[NotNullWhenNot(NullCheckStatus.Unknown)]`, but that's not possible in C# today.

nohwnd: That is way more understandable. Can you put it there please?

nohwnd (approved):

PR #5549, opened 2025-04-30
Reference trimmer

nohwnd:
Replacing #5009


nohwnd:
Looks like this is not very far from being able to build, I would like to see comment on every NoWarn that was added to say why that is necessary, so we don't end up adding noWarn automatically every time the trimmer complains, defeating the purpose of the trimmer.

stan-sz:
The purpose of this change was to help ensure minimal dependencies are shipped with MSTest. Do you plan to re-start this work from scratch?

PR #5556, opened 2025-05-02
Set _TestingPlatformDiagnostics_ in VS

Youssef1313:

### DIFF
--- a/eng/build.ps1
+++ b/eng/build.ps1
@@ -51,6 +51,9 @@ if ($vs) {
     # Disable .NET runtime signature validation errors which errors for local builds
     $env:VSDebugger_ValidateDotnetDebugLibSignatures=0;
 
+    # Enables the logginc of Json RPC messages if diagnostic logging for Test Explorer is enabled in Visual Studio.
+    $env:_TestingPlatformDiagnostics_=1;
+
     # Launch Visual Studio with the locally defined environment variables
     & "$PSScriptRoot\..\TestFx.sln"
 


eng/build.ps1
```diff
@@ -51,6 +51,9 @@ if ($vs) {
     # Disable .NET runtime signature validation errors which errors for local builds
     $env:VSDebugger_ValidateDotnetDebugLibSignatures=0;
 
+    # Enables the logginc of Json RPC messages if diagnostic logging for Test Explorer is enabled in Visual Studio.
+    $env:_TestingPlatformDiagnostics_=1;
```
drognanar, line 55: the one thing I'd double check is the perf impact of this, because this will start logging all RPC messages to the output window

which is why we've made it an environment variable, to avoid flooding the output window, unless you're investigating protocol failure

Youssef1313: This is only for us when running locally, so I assume we are fine paying a perf impact here?

@nohwnd Do you care if things are slower for us, only locally when running through `open-vs.cmd`? At least temporarily until we figure out the processes that remain open?

nohwnd: I don't use open-vs.cmd, and I probably have Trace logging enabled in VS all the time (which IIRC are tied to this as well?), so no impact on me.

Youssef1313: @nohwnd the env variable is still required to log rpc messages, even for trace logging (if I understand correctly). But given you don't use `open-vs.cmd` then you won't be impacted.

I think there is no much risk merging it this way.

nohwnd (approved):
Good for me if we plan to revert it eventually or parametrize the script.

PR #5561, opened 2025-05-05
Add public api files

nohwnd:
Add public api files to projects that were missing the. We are trying to keep the public apis minimal, to avoid users accidentally depending on api that we don't want to keep backwards compatible.


nohwnd:
Probably better approach to https://github.com/microsoft/testfx/pull/5438 to keep the api internal if we can.

Evangelink (approved):

PR #5580, opened 2025-05-13
Update Fakes dependency to 17.14.1

stan-sz:


Evangelink:
Oh right not yet fully done! Let's include this please.

nohwnd:
https://github.com/microsoft/testfx/blob/rel/3.9/Directory.Packages.props#L26 

it is done on the rel branch from which we release, I am just waiting for appropriate moment to release, and then I will consolidate with main.

nohwnd (approved):

PR #5587, opened 2025-05-19
OSConditionAttribute: Make ignore message depend on ConditionMode

Youssef1313:

### DIFF
--- a/src/TestFramework/TestFramework/Attributes/TestMethod/OSConditionAttribute.cs
+++ b/src/TestFramework/TestFramework/Attributes/TestMethod/OSConditionAttribute.cs
@@ -32,7 +32,9 @@ public sealed class OSConditionAttribute : ConditionBaseAttribute
         : base(mode)
     {
         _operatingSystems = operatingSystems;
-        IgnoreMessage = $"Test is only supported on {operatingSystems}";
+        IgnoreMessage = mode == ConditionMode.Include
+            ? $"Test is only supported on {operatingSystems}"
+            : $"Test is not supported on {operatingSystems}";
     }
 
     /// <summary>


src/TestFramework/TestFramework/Attributes/TestMethod/OSConditionAttribute.cs
```diff
@@ -32,7 +32,9 @@ public OSConditionAttribute(ConditionMode mode, OperatingSystems operatingSystem
         : base(mode)
     {
         _operatingSystems = operatingSystems;
-        IgnoreMessage = $"Test is only supported on {operatingSystems}";
+        IgnoreMessage = mode == ConditionMode.Include
```
Evangelink, line 35: I am wondering if we should do some split/join to have something like `Test is only supported on the following OSes: 'Windows', 'MacOS'`

Youssef1313: In case of `[OSCondition(ConditionMode.Exclude, OperatingSystems.Windows)]`, I find it better to read as:

```
Test is not supported on Windows
```

rather than

```
Test is only supported on the following OSes: 'Linux', 'OSX', 'FreeBSD'
```

Evangelink: Sorry looks like I wasn't clear :) I like to have 2 messages, I am just wondering what's being display when you bitwise the enum.

Youssef1313: `ToString` will already comma-separate the named values:

image

Evangelink: Oh cool, I wasn't aware of it. Approving then.

Evangelink (approved):

PR #5589, opened 2025-05-19
Add implicit using even when not using MSTest.Sdk

Youssef1313:

### DIFF
--- a/eng/verify-nupkgs.ps1
+++ b/eng/verify-nupkgs.ps1
@@ -20,7 +20,7 @@ function Confirm-NugetPackages {
     Write-Verbose "Starting Confirm-NugetPackages."
     $expectedNumOfFiles = @{
         "MSTest.Sdk"                                  = 15
-        "MSTest.TestFramework"                        = 154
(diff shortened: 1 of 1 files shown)


Evangelink (changes requested):
You moved it only under `winui+uwp` targets

Youssef1313:
> You moved it only under `winui+uwp` targets

I thought it's just a naming issue on our side and the same file is used for everything, as we seem to include it for all TFMs in nuspec, for example:

https://github.com/microsoft/testfx/blob/a6502391016300f0e4ec299038ccc9775ea75fa9/src/TestFramework/TestFramework.Extensions/MSTest.TestFramework.nuspec#L89

But there are test failures so there is an issue with my change. I'll spend some time to hunt it down.

Evangelink:
My bad! There was some refactoring and there is a single targets left (would be good to cleanup the name).

Evangelink:
I have not followed all changes but the `buildTransitive` is not defined for all supported TFMs which is likely causing the issue above:

image

PR #5592, opened 2025-05-19
Add open-code.cmd

Youssef1313:


Evangelink (dismissed):
It would probably be easier to make the code part of `eng/build.ps1` where we have similar code for VS as it would reduce code duplication but I'll approve and let you decide what you want to do.

Youssef1313:
I think it's generally useful for all Arcade-enabled repos, so it might be good to actually even be under `eng/common`.

Youssef1313:
@Evangelink I went ahead and simplified it per your feedback

Evangelink (dismissed):

eng/build.ps1
```diff
@@ -54,8 +55,12 @@ if ($vs) {
     # Enables the logginc of Json RPC messages if diagnostic logging for Test Explorer is enabled in Visual Studio.
     $env:_TestingPlatformDiagnostics_=1;
 
-    # Launch Visual Studio with the locally defined environment variables
-    & "$PSScriptRoot\..\TestFx.sln"
+    if ($vs) {
+        # Launch Visual Studio with the locally defined environment variables
+        & "$PSScriptRoot\..\TestFx.sln"
+    } else {
+        & code "$PSScriptRoot\.."
```
Evangelink: You way want to check if `code` exists, in case people (like me) have only `code-insiders`.

nohwnd: probably easiest way to do that is via `Get-Command code -ErrorAction Ignore`

Youssef1313: Thanks @Evangelink and @nohwnd. I'm now checking installation for both code and code-insiders.

eng/build.ps1
```diff
@@ -54,8 +55,19 @@ if ($vs) {
...
+    if ($vs) {
+        # Launch Visual Studio with the locally defined environment variables
+        & "$PSScriptRoot\..\TestFx.sln"
+    } else {
+        if (Get-Command code -ErrorAction Ignore) {
+            & code "$PSScriptRoot\.."
+        } elseif (Get-Command code-insiders -ErrorAction Ignore) {
+            & code-insiders "$PSScriptRoot\.."
+        } else {
+            Write-Host "VS Code not found. Please install it from https://code.visualstudio.com/"
```
nohwnd: nit: Write-Error or throw instead?

Youssef1313: Sure, I'm not PowerShell expert and followed Copilot completion :smile:

```suggestion
            Write-Error "VS Code not found. Please install it from https://code.visualstudio.com/"
```

Evangelink (approved):

PR #5599, opened 2025-05-20
Don't publish final package builds to feeds

nohwnd:


azure-pipelines-official.yml
```diff
@@ -181,13 +182,16 @@ extends:
...
 
-            - task: 1ES.PublishNuget@1
-              displayName: 'Publish NuGet packages to test-tools feed'
-              inputs:
-                # Do not push symbol packages nor Microsoft.Testing.Platform package
-                packageParentPath: '$(Build.SourcesDirectory)/artifacts/packages/$(_BuildConfig)'
-                packagesToPush: '$(Build.SourcesDirectory)/artifacts/packages/$(_BuildConfig)/**/*.nupkg;!$(Build.SourcesDirectory)/artifacts/packages/$(_BuildConfig)/**/*.symbols.nupkg;!$(Build.SourcesDirectory)/artifacts/packages/$(_BuildConfig)/NonShipping/Microsoft.Testing.Platform.*.nupkg'
-                publishVstsFeed: 'public/test-tools'
+              # Final builds should not go into test-tools package feed, so we can keep re-building them until we are ready to ship to nuget.org.
+              # This has to depend on the parameter and not on variable, variables are not defined early enough to be used in templating condition.
```
Youssef1313, line 186:
```suggestion
              # This has to depend on the parameter and not on variable, variables are not defined early enough to be used in templating condition.
              # We still need the final builds on test-tools so that internal repos can consume it, so we will publish it manually.
```

mariam-abdulla (approved):

PR #5627, opened 2025-05-21
Remove BuildTimeUTC

Youssef1313:
I don't think I ever found it useful. It only adds burden IMO.


src/Platform/Microsoft.Testing.Platform/PublicAPI/PublicAPI.Unshipped.txt
```diff
@@ -1 +1,2 @@
 #nullable enable
+*REMOVED*[TPEXP]Microsoft.Testing.Platform.Services.IPlatformInformation.BuildDate.get -> System.DateTimeOffset?
```
Youssef1313, line 2: Note: While this is public API, it's experimental and I think we should be fine breaking it. I checked and currently it's used only by MSTest. It's not used by TUnit or xUnit.

Youssef1313:
@nohwnd Aside from the failures in CI caused by MissingMethodException for the removed API (because we use last MSTest from test-tools but we use MTP from the current build), are you generally okay with this change?

nohwnd:
Yes it is fine for me, I was not even aware of it being there, and there is no motivation in the original issue to include the time, so it was not a customer asking to do this.

PR #5644, opened 2025-05-26
VSTestBridge: Handle TestPropertyAttributes.Trait instead of special casing specific properties

### LINKED ISSUE #5642: VSTestBridge should handle `TestPropertyAttributes.Trait`
See https://github.com/nunit/nunit3-vs-adapter/issues/1258#issuecomment-2910097757

Youssef1313:
Fixes #5642

### DIFF
--- a/src/Platform/Microsoft.Testing.Extensions.VSTestBridge/ObjectModel/ObjectModelConverters.cs
+++ b/src/Platform/Microsoft.Testing.Extensions.VSTestBridge/ObjectModel/ObjectModelConverters.cs
@@ -34,18 +34,6 @@ internal static class ObjectModelConverters
         valueType: typeof(string),
         owner: typeof(TestCase));
 
-    private static readonly TestProperty TestCategoryProperty = TestProperty.Register(
-        id: "MSTestDiscoverer.TestCategory",
-        label: "TestCategory",
-        valueType: typeof(string[]),
-        owner: typeof(TestCase));
-
-    private static readonly TestProperty TraitsProperty = TestProperty.Register(
-        id: "TestObject.Traits",
-        label: "Traits",
-        valueType: typeof(KeyValuePair<string, string>[]),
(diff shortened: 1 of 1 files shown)


src/Platform/Microsoft.Testing.Extensions.VSTestBridge/ObjectModel/ObjectModelConverters.cs
```diff
@@ -82,27 +70,33 @@ public static TestNode ToTestNode(this TestCase testCase, bool isTrxEnabled, INa
 
     private static void CopyCategoryAndTraits(TestObject testCaseOrResult, TestNode testNode, bool isTrxEnabled)
     {
-        // TPv2 is doing some special handling for MSTest... we should probably do the same.
-        // See https://github.com/microsoft/vstest/blob/main/src/Microsoft.TestPlatform.Extensions.TrxLogger/Utility/Converter.cs#L66-L70
```
Youssef1313, line 86: @nohwnd FYI this looks like a TRX bug on VSTest where categories are only added for `MSTestDiscoverer.TestCategory` and won't work for NUnit.

nohwnd: Iirc fixed in 17.14, that mstest cannot update to.

Youssef1313: https://github.com/microsoft/vstest/pull/14905 is this what you are referring to?
It seems like it's fixing `Properties`, not `TestCategory`

nohwnd: yes, my mistake then.

PR #5658, opened 2025-05-28
Allow framework authors to use TestCase.FullyQualifiedName as the TestNodeUid

Youssef1313:
@nohwnd What do you think?

### DIFF
--- a/src/Platform/Microsoft.Testing.Extensions.VSTestBridge/ObjectModel/ObjectModelConverters.cs
+++ b/src/Platform/Microsoft.Testing.Extensions.VSTestBridge/ObjectModel/ObjectModelConverters.cs
@@ -49,9 +49,16 @@ internal static class ObjectModelConverters
     /// <summary>
     /// Converts a VSTest <see cref="TestCase"/> to a Microsoft Testing Platform <see cref="TestNode"/>.
     /// </summary>
-    public static TestNode ToTestNode(this TestCase testCase, bool isTrxEnabled, INamedFeatureCapability? namedFeatureCapability, ICommandLineOptions commandLineOptions, IClientInfo clientInfo, string? displayNameFromTestResult = null)
+    public static TestNode ToTestNode(
+        this TestCase testCase,
+        bool isTrxEnabled,
+        bool useFQNAsUid,
+        INamedFeatureCapability? namedFeatureCapability,
+        ICommandLineOptions commandLineOptions,
+        IClientInfo clientInfo,
+        string? displayNameFromTestResult = null)
     {
-        string testNodeUid = testCase.Id.ToString();
+        string testNodeUid = useFQNAsUid ? testCase.FullyQualifiedName : testCase.Id.ToString();
 
         TestNode testNode = new()
         {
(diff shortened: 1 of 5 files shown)


Youssef1313:
@nohwnd Do you prefer the `ITestNodeUidProviderCapability` approach?

Youssef1313:
> Yeah, that definitely seems more dynamic. If you don't see other problems with that let's do it this way. 👍

I don't see other problems. The only problem I initially saw was building the filter but it's solvable with `GetFilterPropertyName`.

Youssef1313:
@nohwnd Thinking more. This may cause issues with Test Explorer due to a behavior that I *think* is a bug.

When vstestProvider is used, Test Explorer always sends FullyQualifiedName as TestNodeUid when running a selected test in Test Explorer. Allowing full customization will mean that test framework can use arbitrary test node uids and arbitrary property name for filtering. Then, test explorer will end up sending FQN, but we will attempt to use it for the arbitrary filter property name defined by the framework.

In general, I think that the current Test Explorer behavior is not correct. Fixing Test Explorer will then mean breaking old versions of NUnit.

src/Platform/Microsoft.Testing.Extensions.VSTestBridge/ObjectModel/ObjectModelConverters.cs
```diff
@@ -49,9 +49,16 @@ internal static class ObjectModelConverters
     /// <summary>
     /// Converts a VSTest <see cref="TestCase"/> to a Microsoft Testing Platform <see cref="TestNode"/>.
     /// </summary>
-    public static TestNode ToTestNode(this TestCase testCase, bool isTrxEnabled, INamedFeatureCapability? namedFeatureCapability, ICommandLineOptions commandLineOptions, IClientInfo clientInfo, string? displayNameFromTestResult = null)
+    public static TestNode ToTestNode(
+        this TestCase testCase,
+        bool isTrxEnabled,
+        bool useFQNAsUid,
```
Evangelink: Let's avoid abbreviations.

Evangelink (dismissed):

src/Platform/Microsoft.Testing.Extensions.VSTestBridge/ObjectModel/ObjectModelConverters.cs
Youssef1313: Renamed.

Youssef1313:
> as for the test window behaviour, because in TPv2 TCIDs are GUIDs and in MTP UIDs are strings, they've not been used for identity. we can update the TestExplorer behavior, but we need to prepare both VS and MTP to send/receive old and new data. so if we could look at any additions to the message payload (like adding parametrized test data) it would be nice to make all changes in one go

From what I observed, Test Explorer will already send the correct TestNodeUid if vstestProvider isn't used. In that case, it correctly passes the given string even if it's not a guid.

nohwnd:
> This may cause issues with Test Explorer due to a behavior that I think is a bug.

In that case please revert to your original approach.  Sorry for making it harder.

Youssef1313:
> > This may cause issues with Test Explorer due to a behavior that I think is a bug.
> 
> In that case please revert to your original approach. Sorry for making it harder.

I didn't change the original approach in this PR. The updated approach was done separately in https://github.com/microsoft/testfx/pull/5664 and closed that PR after seeing the potential issues with it.

nohwnd (approved):

PR #5660, opened 2025-05-28
Cleanup dead code in MTP

Youssef1313:


src/Platform/Microsoft.Testing.Platform/ServerMode/JsonRpc/MessageHandlerFactory.cs
```diff
@@ -39,52 +39,21 @@ public MessageHandlerFactory(
 
         public string Description => nameof(MessageHandlerFactory);
 
-        public Task<IMessageHandler> CreateMessageHandlerAsync(CancellationToken cancellationToken)
-            => _host is not null
```
Youssef1313, line 43: Note to make it easier for review.

`_host` is never null. That means `StartTestPlatformServerAsync` is dead. So kept only the logic of `ConnectToTestPlatformClientAsync` but inlined right away in `CreateMessageHandlerAsync`.

src/Platform/Microsoft.Testing.Platform/ServerMode/JsonRpc/ServerModeManager.cs
```diff
@@ -20,34 +18,10 @@ internal IMessageHandlerFactory Build(IServiceProvider serviceProvider)
             ? int.Parse(clientPortArgs[0], CultureInfo.InvariantCulture)
             : throw new InvalidOperationException(PlatformResources.MissingClientPortFoJsonRpc);
 
-        string? clientHostName;
-        clientHostName = commandLineService.TryGetOptionArgumentList(PlatformCommandLineProvider.ClientHostOptionKey, out string[]? clientHostArgs)
+        string clientHostName = commandLineService.TryGetOptionArgumentList(PlatformCommandLineProvider.ClientHostOptionKey, out string[]? clientHostArgs)
             ? clientHostArgs[0]
             : "localhost";
 
-        if (CommunicationProtocol is not null)
```
Youssef1313, line 28: Note to make it easier for review.

We never set `CommunicationProtocol`, and never instantiated `JsonRpcTcpServerToSingleClient`. So cleaned all that up.

The cleanup made it such that this method (`Build`), can be made static. So made it static, and consequently cleaned up the no longer needed interface `IServerModeManager` along with the `ServerMode` property that is no longer needed as the `Build` method is accessed directly from `ServerModeManager`

PR #5671, opened 2025-05-30
Use local SDK in global.json

Youssef1313:


global.json
```diff
@@ -24,6 +24,7 @@
   },
   "sdk": {
     "version": "10.0.100-preview.5.25269.23",
+    "paths": [ ".dotnet", "$host$" ],
```
Youssef1313, line 27: I'm seeing the following in CI:

```
2025-05-30T07:58:27.6583168Z D:\a\_work\1\s\artifacts\bin\MSTest.Acceptance.IntegrationTests\Release\net10.0\MSTest.Acceptance.IntegrationTests.exe (net10.0|) failed (0ms)
2025-05-30T07:58:27.6588369Z Exit code: -2147450730
2025-05-30T07:58:27.6589476Z   Error output: You must install or update .NET to run this application.
2025-05-30T07:58:27.6590320Z   App: D:\a\_work\1\s\artifacts\bin\MSTest.Acceptance.IntegrationTests\Release\net10.0\MSTest.Acceptance.IntegrationTests.exe
2025-05-30T07:58:27.6590869Z   Architecture: x64
2025-05-30T07:58:27.6591348Z   Framework: 'Microsoft.NETCore.App', version '10.0.0-preview.5.25265.101' (x64)
2025-05-30T07:58:27.6591829Z   .NET location: C:\Program Files\dotnet\
2025-05-30T07:58:27.6592256Z   The following frameworks were found:
2025-05-30T07:58:27.6593235Z     6.0.36 at [C:\Program Files\dotnet\shared\Microsoft.NETCore.App]
2025-05-30T07:58:27.6593725Z     6.0.37 at [C:\Program Files\dotnet\shared\Microsoft.NETCore.App]
2025-05-30T07:58:27.6594218Z     8.0.13 at [C:\Program Files\dotnet\shared\Microsoft.NETCore.App]
2025-05-30T07:58:27.6594690Z     9.0.2 at [C:\Program Files\dotnet\shared\Microsoft.NETCore.App]
2025-05-30T07:58:27.6595105Z   Learn more:
2025-05-30T07:58:27.6595531Z   https://aka.ms/dotnet/app-launch-failed
2025-05-30T07:58:27.6595993Z   To install missing framework, download:
2025-05-30T07:58:27.6596537Z   https://aka.ms/dotnet-core-applaunch?framework=Microsoft.NETCore.App&framework_version=10.0.0-preview.5.25265.101&arch=x64&rid=win-x64&os=win10
```

This is kinda unexpected for me.

Youssef1313: @jaredpar Is there any additional work needed on SDK side for `dotnet test`?

We simply just run the test executable in `dotnet test` [here](https://github.com/dotnet/sdk/blob/7a7cb3e80b1231276993c5a5fd3ef60994e543d5/src/Cli/dotnet/Commands/Test/TestApplication.cs#L45-L48) (and a parent directory of the executable has `global.json` pointing to the locally installed SDK).

Youssef1313: Running locally:

image

- `dotnet --list-sdks`: Recognizes the SDK from global.json
- `dotnet --list-runtimes`: Doesn't recognize runtimes from the locally installed SDK
- `.dotnet/dotnet.exe --list-runtimes`: Recognizes runtimes from the locally installed SDK

Not sure if this is by-design or if there is something missing with the local SDK feature, and if it's by-design, how can we improve the experience there for `dotnet test` at least (if we should do anything at all)

nohwnd: I would start from `corehost_trace=1` to understand what is happening. Typically I see dotnet_root set to define where the local dotnets are, dotnet_multilevel_lookup set to 0 to prevent going to global dotnet installation (program files), and then also .dotnet.dotnet.exe being added to PATH to ensure we resolve the local dotnet exe for SDK builds.

Youssef1313: Nothing is interesting in the trace logs it seems :(

> Using dotnet root path [C:\Program Files\dotnet\]

It's not considering the local SDK at all. I'm not sure if that's intended.

Youssef1313: https://github.com/dotnet/designs/blob/main/accepted/2025/local-sdk-global-json.md#dotnet-exec:

```
This proposal only impacts how .NET SDK commands do runtime discovery. The command dotnet exec is not an .NET SDK command but instead a way to invoke the app directly using the runtime installed with dotnet.

It is reasonable for complex builds to build and use small tools. For example building a tool for linting the build, running complex validation, etc ... To work with local SDK discovery these builds need to leverage dotnet run to execute such tools instead of dotnet exec.
```

Looks like probably by-design? 😕

jaredpar: @agocke

agocke: Correct, by design. This is an SDK feature, not a runtime feature. If you run “app.exe” it will not look for a global.json.

Youssef1313: So, from user point of view, they are invoking `dotnet test`, so it's SDK. But as SDK currently simply runs the test executable directly, we hit this issue (in simple words, we are moving from SDK-world to runtime but losing the info about runtimes installed in the local SDK).

So I guess there might be something to do for the implementation of `dotnet test` in that regard? What exactly should we do for `dotnet test`?

Youssef1313: Adding @elinor-fung to the conversation as well.

azure-pipelines.yml
```diff
@@ -219,11 +217,10 @@ stages:
           # Because the build step is using -ci flag, restore is done in a local .packages directory.
           # We need to pass NUGET_PACKAGES so that when dotnet test is doing evaluation phase on the projects, it can resolve .props/.targets from packages and import them.
           # Otherwise, props/targets are not imported. It's important that they are imported so that IsTestingPlatformApplication ends up being set.
-          - script: $(Build.SourcesDirectory)/.dotnet/dotnet test --solution NonWindowsTests.slnf -c $(_BuildConfig) --no-build -bl:$(BUILD.SOURCESDIRECTORY)\artifacts\log\$(_BuildConfig)\TestStep.binlog
+          - script: dotnet test --solution NonWindowsTests.slnf -c $(_BuildConfig) --no-build -bl:$(BUILD.SOURCESDIRECTORY)\artifacts\log\$(_BuildConfig)\TestStep.binlog
             name: Test
             displayName: Test
             env:
-              DOTNET_ROOT: $(Build.SourcesDirectory)/.dotnet
               NUGET_PACKAGES: $(Build.SourcesDirectory)/.packages
```
Youssef1313:
```suggestion
              DOTNET_ROOT: $(Build.SourcesDirectory)/.dotnet
              NUGET_PACKAGES: $(Build.SourcesDirectory)/.packages
```

azure-pipelines.yml
```diff
@@ -162,11 +161,10 @@ stages:
...
```
Youssef1313:
```suggestion
              DOTNET_ROOT: $(Build.SourcesDirectory)/.dotnet
              NUGET_PACKAGES: $(Build.SourcesDirectory)/.packages
```

azure-pipelines.yml
```diff
@@ -98,11 +98,10 @@ stages:
           # Because the build step is using -ci flag, restore is done in a local .packages directory.
           # We need to pass NUGET_PACKAGES so that when dotnet test is doing evaluation phase on the projects, it can resolve .props/.targets from packages and import them.
           # Otherwise, props/targets are not imported. It's important that they are imported so that IsTestingPlatformApplication ends up being set.
-          - script: $(Build.SourcesDirectory)/.dotnet/dotnet test -c $(_BuildConfig) --no-build -bl:$(BUILD.SOURCESDIRECTORY)\artifacts\log\$(_BuildConfig)\TestStep.binlog
+          - script: dotnet test -c $(_BuildConfig) --no-build -bl:$(BUILD.SOURCESDIRECTORY)\artifacts\log\$(_BuildConfig)\TestStep.binlog
             name: Test
             displayName: Test
             env:
-              DOTNET_ROOT: $(Build.SourcesDirectory)/.dotnet
               NUGET_PACKAGES: $(Build.SourcesDirectory)/.packages
```
Youssef1313:
```suggestion
              DOTNET_ROOT: $(Build.SourcesDirectory)/.dotnet
              NUGET_PACKAGES: $(Build.SourcesDirectory)/.packages
```

global.json
Youssef1313: Tracking this by https://github.com/dotnet/sdk/issues/48060 which we will leave for now until there is a strong reason to address it.

agocke: Where is the design doc on how dotnet test should work with regards to local runtimes, global runtimes, etc? I think we can give guidance on how to implement your design once we see it.

Youssef1313: We don't have a design doc so far. I talked to @nohwnd and based on previous history, Jakub thinks it's not worth bothering with setting DOTNET_ROOT ourselves etc and he is okay keeping the current behavior (always using default .NET installation, i.e, simply starting the test apphost directly). I have no **strong** opinion even though I'm leaning towards that we *need* to improve the experience. I'm more than happy to set a meeting up for all of us if needed, or if you strongly think that the current behavior is not correct.

nohwnd: Yes, I would like to avoid changing how the runtime / sdk is resolved, and lean only on the built in functionality. There have been several changes in how runtimes are resolved (e.g. if multilevel lookup is done or not), and vstest always had to try hard to keep the behavior the same as before. 

Instead we would just do it the way the current sdk/runtime does it, without playing catch up with the changes.

agocke: > Jakub thinks it's not worth bothering with setting DOTNET_ROOT ourselves etc and he is okay keeping the current behavior (always using default .NET installation, i.e, simply starting the test apphost directly)

So if someone doesn't have the SDK globally installed, dotnet test won't work?

Youssef1313: @agocke If the needed *runtime* isn't found globally (usually `C:\Program Files\dotnet\shared\Microsoft.NETCore.App` and there are some registry keys that affect the resolution AFAIK, etc), yes it will error and show "You must install or update .NET to run this application". That's why we currently need to set `DOTNET_ROOT` in our YML, even when using local SDK in global.json.

nohwnd (approved):

PR #5676, opened 2025-06-01
Allow serialization of DateOnly/TimeOnly for parameterized tests

Youssef1313:


src/Adapter/MSTest.TestAdapter/Helpers/DataSerializationHelper.cs
```diff
@@ -93,6 +105,10 @@ internal static class DataSerializationHelper
             }
 
             DataContractJsonSerializer serializer = GetSerializer(assemblyQualifiedName);
+#if NET7_0_OR_GREATER // TODO: Docs say it's available on net6.0, but that's not what I see here :/
```
Youssef1313: @nohwnd Do you see anything here please?

Evangelink: image

Youssef1313: @Evangelink Yes that's what I'm seeing, but docs say otherwise.

Evangelink: Yes I have also seen the docs, I think they are wrong because uncompiling net6 dll shows the method is not present.

src/Adapter/MSTest.TestAdapter/Helpers/DataSerializationHelper.cs
```diff
@@ -13,6 +18,10 @@ internal static class DataSerializationHelper
         UseSimpleDictionaryFormat = true,
         EmitTypeInformation = System.Runtime.Serialization.EmitTypeInformation.Always,
         DateTimeFormat = new System.Runtime.Serialization.DateTimeFormat("O", CultureInfo.InvariantCulture),
+#if NETFRAMEWORK
+        DataContractSurrogate = SerializationSurrogateProvider.Instance,
```
drognanar, line 22: are DateOnly/TimeOnly structs only available on .NET core? or can then be used in netfx code via some nuget package, which is why we're adding the netfx code?

Youssef1313: They are not available for .NET Framework.

`DataContractSurrogate` will currently do "nothing" on .NET Framework, but I'm setting everything up in place to easily allow us to add more types if needed in future.

Issue #5677, opened 2025-06-02
Treenode Filter: KeyValuePairStringProperty vs TestMetadataProperty


thomhurst:
In order to filter with properties using the treenode filter, these properties must be defined as a `KeyValuePairStringProperty`: https://github.com/microsoft/testfx/blob/ab9e1494a285006703cc3703be7f4ed4268cff23/src/Platform/Microsoft.Testing.Platform/Requests/TreeNodeFilter/TreeNodeFilter.cs#L553

However, in order for Visual Studio to be able to use properties for filtering, they must be defined as a `TestMetadataProperty`.

So one works in Visual Studio but not the treenode filter, and one works in the treenode filter but not Visual Studio.

Seems `TestMetadataProperty` should work for both?

I think the simplest solution is that `TestMetadataProperty` inherits from `KeyValuePairStringProperty` so the `is` check in the treenode filter succeeds for a `TestMetadataProperty` ?

PR #5679, opened 2025-06-02
Enable filtering with `TestMetadataProperty` in the treenode filter

thomhurst:
Fix: #5677

### DIFF
--- a/src/Platform/Microsoft.Testing.Platform/Messages/TestNodeProperties.cs
+++ b/src/Platform/Microsoft.Testing.Platform/Messages/TestNodeProperties.cs
@@ -416,4 +416,6 @@ public record StandardErrorProperty(string StandardError) : IProperty;
 /// <param name="Description">The description.</param>
 public record FileArtifactProperty(FileInfo FileInfo, string DisplayName, string? Description = null) : IProperty;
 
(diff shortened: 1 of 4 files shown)


Youssef1313 (changes requested):
This seems to be a binary breaking change.

Could we instead deprecate `KeyValuePairStringProperty`, change the filter to type-check for both properties, and in a future release (probably v2?), take the breaking change of removing `KeyValuePairStringProperty`?

thomhurst:
How's this?

Youssef1313:
> How's this?

The `Key` and `Value` properties are now on the base class, and not on `TestMetadataProperty`. So, if a library (e.g, TUnit) is compiled against old version of the platform where `TestMetadataProperty` contains `Key` property, and the user updates to a newer version of the platform, where the property is now moved up to the base class, I expect to see a MissingMethodException saying that `TestMetadataProperty.get_Key` isn't found.

Youssef1313 (approved):
LGTM. I'll let @nohwnd take another look as well.

Please update the PR title to reflect the current shape of the fix.

nohwnd (dismissed):
LGMT, will need to re-approve when tests pass.

PR #5684, opened 2025-06-03
Fix MSTEST0042 (duplicate data row) false positive with Zero/NegativeZero

### LINKED ISSUE #5683: MSTEST0042 FP for -0.0 and +0.0
## Describe the bug

For `double`, `float` and `half` zero (0.0) and negative (-0.0) are bitwise _different_ values, which can be seen using `BitConverter`, but they are considered equal when using `==` or `Equals`.

This detail seems to trip MSTEST0042, which reports `[DataRow(-0.0)]` and `[DataRow(+0.0)]` as duplicate rows, even though they are different test cases.

```
Do not duplicate 'DataRow' attributes. This is usually a copy/paste error. The attribute indices are '0' and '1'.
```

## Steps To Reproduce

```cs
[DataRow(-0.0)]
[DataRow(+0.0)] // <-- MSTEST0042 is triggered
[TestMethod]
public void TestMethod1(double d)
{
    Assert.AreEqual(0.0, d);
}
```

Note that we would be glad if you could enclose a repro (zip or link to some GitHub project).
-->

## Expected behavior

`-0-0` and `0.0` should be considered distinct values

## Actual behavior

MSTEST0042 reports that `-0-0` and `0.0` are duplicate values.

## Additional context

The use case is that we have a method to omit the leading minus when stringifying `-0.0`, such that both `0.0` and `-0.0` are stringified to `0`.
So we have a unit test similar to this, for which MSTEST0042 is now reported.

```cs
[DataRow(-0.0, "0")]
[DataRow(+0.0, "0")]
[DataRow(1.0, "1")]
[DataRow(1.5, "1.5")]
[TestMethod]
public void TestMethod1(double d, string expected)
{
    Assert.AreEqual(expected, ToString(d));
}

/// <summary>
/// Converts a double to a string representation, but treats -0.0 and +0.0 as "0.0".
/// </summary>
private static string ToString(double d) => d == 0.0 ? "0" : d.ToString(CultureInfo.InvariantCulture);
```

Youssef1313:
Fixes #5683

### DIFF
--- a/src/Analyzers/MSTest.Analyzers/DuplicateDataRowAnalyzer.cs
+++ b/src/Analyzers/MSTest.Analyzers/DuplicateDataRowAnalyzer.cs
@@ -126,6 +126,21 @@ public sealed class DuplicateDataRowAnalyzer : DiagnosticAnalyzer
                 return TypedConstantArrayComparer.Instance.Equals(typedConstant1.Values, typedConstant2.Values);
             }
 
+            if (typedConstant1.Kind == TypedConstantKind.Primitive)
+            {
+                // object.Equals(float.NegativeZero, 0.0f) will return true.
+                // But we don't want to consider it as "equal" as the test case can yield different results.
+                // Behavior difference between zero and negative zero can be observed via BitConverter or ToString.
+                if (typedConstant1.Value is float float1 && typedConstant2.Value is float float2)
+                {
+                    return BitConverter.GetBytes(float1).SequenceEqual(BitConverter.GetBytes(float2));
+                }
+                else if (typedConstant1.Value is double double1 && typedConstant2.Value is double double2)
+                {
+                    return BitConverter.GetBytes(double1).SequenceEqual(BitConverter.GetBytes(double2));
+                }
+            }
+
             // At this point, the type is matching and the kind is matching and is not array.
             return object.Equals(typedConstant1.Value, typedConstant2.Value);
         }
--- a/test/UnitTests/MSTest.Analyzers.UnitTests/DuplicateDataRowAnalyzerTests.cs
+++ b/test/UnitTests/MSTest.Analyzers.UnitTests/DuplicateDataRowAnalyzerTests.cs
@@ -177,4 +177,116 @@ public sealed class DuplicateDataRowAnalyzerTests
 
         await VerifyCS.VerifyAnalyzerAsync(code);
     }
+
+    [TestMethod]
+    public async Task WhenZeroAndNegativeZero_NoDiagnostic()
+    {
+        string code = """
+            using Microsoft.VisualStudio.TestTools.UnitTesting;
+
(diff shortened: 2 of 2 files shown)


src/Analyzers/MSTest.Analyzers/DuplicateDataRowAnalyzer.cs
```diff
@@ -126,6 +126,21 @@ private static bool AreTypedConstantEquals(TypedConstant typedConstant1, TypedCo
...
             }
 
+            if (typedConstant1.Kind == TypedConstantKind.Primitive)
+            {
+                // object.Equals(float.NegativeZero, 0.0f) will return true.
+                // But we don't want to consider it as "equal" as the test case can yield different results.
+                // Behavior difference between zero and negative zero can be observed via BitConverter or ToString.
+                if (typedConstant1.Value is float float1 && typedConstant2.Value is float float2)
+                {
+                    return BitConverter.GetBytes(float1).SequenceEqual(BitConverter.GetBytes(float2));
```
jnyrup, line 137: Can we just reinterpret to `int`s using `SingleToInt32Bits` and compare those two ints?
```cs
return BitConverter.SingleToInt32Bits(float1) == BitConverter.SingleToInt32Bits(float2);
```

Youssef1313: Analyzers target netstandard2.0 where `SingleToInt32Bits` isn't available.

Youssef1313: We *could* rewrite it to the following though:

```csharp
if (typedConstant1.Value is float float1 && typedConstant2.Value is float float2)
{
    return *(int*)&float1 == *(int*)&float2;
}
else if (typedConstant1.Value is double double1 && typedConstant2.Value is double double2)
{
    return *(long*)&double1 == *(long*)&double2;
}
```

Youssef1313: @nohwnd Any preferences?

jnyrup: Roger.
`BitConverter.DoubleToInt64Bits` seems to be available in netstandard2.0, that could perhaps be used for the `double` case below.

(didn't see the follow-up with raw cast before writing this)

Youssef1313: > Roger. `BitConverter.DoubleToInt64Bits` seems to be available in netstandard2.0, that could perhaps be used for the `double` case below.

Yup, `DoubleToInt64Bits` is available. It's likely faster, but unlikely to matter in real-world scenarios. But I switched to it anyways.

PR #5694, opened 2025-06-04
Avoid loading System.Threading.Tasks.Extensions when not needed

Youssef1313:
This works-around cases where binding redirects are not properly generated in consumer projects, either due to missing Microsoft.NET.Test.Sdk dependency, or due to reasons not yet known.

### DIFF
--- a/src/Adapter/MSTest.TestAdapter/Execution/TestMethodInfo.cs
+++ b/src/Adapter/MSTest.TestAdapter/Execution/TestMethodInfo.cs
@@ -432,32 +432,10 @@ public class TestMethodInfo : ITestMethod
 
                         if (executionContext is null)
                         {
-                            object? invokeResult = TestMethod.GetInvokeResult(_classInstance, arguments);
-                            if (invokeResult is null)
+                            Task? invokeResult = TestMethod.GetInvokeResultAsync(_classInstance, arguments);
+                            if (invokeResult is not null)
                             {
-                                // Do nothing.
-                                // Don't remove this if condition as we don't want to go in the "else" in this case.
-                            }
-                            else if (invokeResult is Task task)
-                            {
-                                await task;
-                            }
-                            else
-                            {
-                                await TryAwaitValueTaskAsync(invokeResult);
-
-                                // Avoid loading System.Threading.Tasks.Extensions if not needed.
-                                // Note: .NET runtime will load all types once it's entering the method.
-                                // So, moving this out of the method will load System.Threading.Tasks.Extensions
-                                // Even when invokeResult is null or Task.
-                                [MethodImpl(MethodImplOptions.NoInlining)]
-                                static async Task TryAwaitValueTaskAsync(object invokeResult)
-                                {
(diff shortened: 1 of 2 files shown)


src/Adapter/MSTest.TestAdapter/Execution/TestMethodInfo.cs
```diff
@@ -452,13 +467,28 @@ private async Task<TestResult> ExecuteInternalAsync(object?[]? arguments, Cancel
...
                                         await task;
                                     }
-                                    else if (invokeResult is ValueTask valueTask)
+                                    else
                                     {
-                                        await valueTask;
+                                        await TryAwaitValueTaskAsync(invokeResult);
+
+                                        // Avoid loading System.Threading.Tasks.Extensions if not needed.
+                                        [MethodImpl(MethodImplOptions.NoInlining)]
```
nohwnd: I assume here we rely on the behavior of JITter, that will load types mentioned in a method just before entering the method. Is that correct? Might be worth explaining that a little futher in the comment here (and the one below)?

Youssef1313: @nohwnd Yes. I'm not sure also if `NoInlining` is required or not, but having it "just in case".

Maybe @jkotas knows.

src/Adapter/MSTest.TestAdapter/Execution/TestMethodInfo.cs
```diff
@@ -433,13 +433,31 @@ private async Task<TestResult> ExecuteInternalAsync(object?[]? arguments, Cancel
                         if (executionContext is null)
                         {
                             object? invokeResult = TestMethod.GetInvokeResult(_classInstance, arguments);
-                            if (invokeResult is Task task)
+                            if (invokeResult is null)
```
mariam-abdulla: We can do some refactoring here as we have this code snippet multiple times.

Youssef1313: It was actually buggy the original logic where I'm awaiting `valueTask` inside of an `async` method as it will lose any `AsyncLocal`s set in test method (i.e, they won't be seen in cleanup). Refactored to also address that issue.

src/Adapter/MSTest.TestAdapter/Execution/TestMethodInfo.cs
jkotas: The file not found exceptions during inlining abort the inlining attempt, so it would still work without `NoInling` annotation, but you would see exception thrown and ignored under debugger. `NoInling` annotation avoids these thrown and ignored exceptions.

Youssef1313: Thanks for the info. This is good to know!

Evangelink (approved):

Youssef1313:
I'll get to writing a test for this later as few important stuff are showing up.

PR #5713, opened 2025-06-09
Move files from adapter to platform services project

Evangelink:
While being technically a breaking change, by adding type forwarding and looking at public usages we should not have any issue.

### DIFF
--- a/eng/Version.Details.xml
+++ b/eng/Version.Details.xml
@@ -1,21 +1,21 @@
 <?xml version="1.0" encoding="utf-8"?>
 <Dependencies>
   <ToolsetDependencies>
-    <Dependency Name="Microsoft.DotNet.Arcade.Sdk" Version="10.0.0-beta.25309.2">
+    <Dependency Name="Microsoft.DotNet.Arcade.Sdk" Version="10.0.0-beta.25310.2">
       <Uri>https://github.com/dotnet/arcade</Uri>
-      <Sha>add492db56ebc80df604ba5a30723beef14ed3dd</Sha>
+      <Sha>d237be4630776450573bc588eb995bb8f9ebc04f</Sha>
     </Dependency>
-    <Dependency Name="Microsoft.DotNet.Build.Tasks.Templating" Version="10.0.0-beta.25309.2">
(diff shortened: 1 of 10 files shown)


src/Adapter/MSTest.TestAdapter/Execution/TcmTestPropertiesProvider.cs
```diff
@@ -1,50 +0,0 @@
-﻿// Copyright (c) Microsoft Corporation. All rights reserved.
```
Evangelink, line 1: Wrongly detected as removed + added, it was moved.

Youssef1313: I think it's seen that way due to the Constants -> EngineConstants change which affects many lines. But that's fine 👍

test/UnitTests/TestFramework.UnitTests/Assertions/AssertTests.IsInstanceOfTypeTests.cs
Youssef1313, line 1: Should the changes here be reverted? Or is there some analyzer showing warning to simplify the code?

Youssef1313 (dismissed):

test/UnitTests/TestFramework.UnitTests/Assertions/AssertTests.IsInstanceOfTypeTests.cs
Evangelink: Sorry missed it! I did call the simplify analyzer fixer after the rename of `Constants` and asked to fix on all sln that's probably where this happened.

Youssef1313 (dismissed):

Evangelink:
> Can you please link an issue, or add more context about why this change is desirable? :)

This is a first-step toward being able to abstract away VSTest objects so we can ship a new NuGet package that will be MTP only allowing to get rid of all VSTest deps, and remove all requirements for custom properties.

Youssef1313 (approved):

Youssef1313 (approved):

PR #5715, opened 2025-06-09
Fix UnitTestRunner leaking some test class instances

### LINKED ISSUE #5711: UnitTestRunner "leaks"  Test Class instances
## Describe the bug
We're running 5k tests and suffering from memory leaks.
Part of the problem is that we have a long-established pattern of storing lots of transient data on the test class instance.
While the test is running (or being cleaned-up) the data is required. But once the test is finished, it can all be collected.

Unfortunately, one instance of each test class appears to be held in `UnitTestRunner._fixtureTests`, and I don't see any code paths that remove items from this concurrent dictionary. 
Shouldn't there be some mechanism for clearing these up once all the fixtures in a class have run?

Youssef1313:
Fixes #5711


src/Adapter/MSTest.TestAdapter/Execution/UnitTestRunner.cs
```diff
@@ -22,7 +22,8 @@ namespace Microsoft.VisualStudio.TestPlatform.MSTest.TestAdapter.Execution;
 /// </summary>
 internal sealed class UnitTestRunner : MarshalByRefObject
 {
-    private readonly ConcurrentDictionary<string, TestMethodInfo> _fixtureTests = new();
+    private readonly ConcurrentDictionary<string, TestAssemblyInfo> _assemblyFixtureTests = new();
```
fhnaseer, line 25: How is this fixing the issue? Are you cleaning _classFixtureTests after class cleanup somewhere?

Youssef1313: @fhnaseer We are no longer holding into test class instances. The test class instance is held by `TestMethodInfo`, but not by `TestAssemblyInfo` or `TestClassInfo`.

Issue #5292, opened 2025-03-21
`Microsoft.Testing.Extensions.Retry` can't move files from TestResults when running from `dotnet test` on Windows


CheezeCS:
## Describe the bug

Hello, when running tests on Windows with `dotnet test` when using the ‘retry’ parameter, the test run terminates with an error, due to moving a file that is used by another process

## Steps To Reproduce

1. Create simple unit test
2. Run it through `dotnet test -- --retry-failed-tests 1` on Windows
3. error run failed: Tests failed

## Expected behavior
Test run completes without errors
## Actual behavior
Test run failed
## Additional context

TestResults:
```
=====================
Tests suite completed successfully in 1 attempts
=====================

=====================
Moving last attempt asset files to the default result directory
=====================

Moving file 'C:\RiderProjects\Allure.Testing.Platform\TestProject1\bin\Debug\net8.0\TestResults\Retries\qNqn3\1\TestProject1_net8.0_x64.log' to 'C:\RiderProjects\Allure.Testing.Platform\TestProject1\bin\Debug\net8.0\TestResults\TestProject1_net8.0_x64.log'
Unhandled exception. System.IO.IOException: The process cannot access the file because it is being used by another process.
   at System.IO.FileSystem.MoveFile(String sourceFullPath, String destFullPath, Boolean overwrite)
   at Microsoft.Testing.Extensions.Policy.RetryOrchestrator.OrchestrateTestHostExecutionAsync() in /_/src/Microsoft.Testing.Extensions.Retry/RetryOrchestrator.cs:line 320
   at Microsoft.Testing.Platform.Hosts.TestHostOrchestratorHost.RunAsync() in /_/src/Platform/Microsoft.Testing.Platform/Hosts/TestHostOchestratorHost.cs:line 30
   at Microsoft.Testing.Platform.Builder.TestApplication.RunAsync() in /_/src/Platform/Microsoft.Testing.Platform/Builder/TestApplication.cs:line 244
   at Xunit.Runner.InProc.SystemConsole.TestingPlatform.TestPlatformTestFramework.RunAsync(String[] args, Action`2 extensionRegistration) in /_/src/xunit.v3.runner.inproc.console/TestingPlatform/TestPlatformTestFramework.cs:line 270
   at XunitAutoGeneratedEntryPoint.Main(String[] args) in C:\.nuget\packages\xunit.v3.core\1.1.0\_content\EntryPoint-mtp.cs:line 11

=== COMMAND LINE ===
[...]

PR #5717, opened 2025-06-09
Add lifecycle callbacks to test host orchestrator

Youssef1313:
Fixes https://github.com/microsoft/testfx/issues/5292


src/Platform/Microsoft.Testing.Platform/TestHostOrcherstrator/ITestHostOrchestratorApplicationLifetime.cs
```diff
@@ -0,0 +1,25 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+namespace Microsoft.Testing.Platform.Extensions.TestHostOrchestrator;
+
+// NOTE: The equivalent of this for "test host" is ITestApplicationLifecycleCallbacks, which is an unfortunate naming.
```
Evangelink, line 6: We could ship one for TestHostControllers, one for TestHost (with similar naming pattern), and deprecate the current `ITestApplicationLifecycleCallbacks` (suggesting to use `ITestHostApplicationLifetime` instead). That would put us in a better shape for v2.

src/Platform/Microsoft.Testing.Platform/TestHostOrcherstrator/TestHostOrchestratorManager.cs
```diff
@@ -48,4 +49,37 @@ public async Task<TestHostOrchestratorConfiguration> BuildAsync(ServiceProvider
...
+        foreach (Func<IServiceProvider, ITestHostOrchestratorApplicationLifetime> testHostOrchestratorApplicationLifetimeFactory in _testHostOrchestratorApplicationLifetimeFactories)
+        {
+            ITestHostOrchestratorApplicationLifetime service = testHostOrchestratorApplicationLifetimeFactory(serviceProvider);
+
+            // Check if we have already extensions of the same type with same id registered
+            if (lifetimes.Any(x => x.Uid == service.Uid))
+            {
+                ITestHostOrchestratorApplicationLifetime currentRegisteredExtension = lifetimes.Single(x => x.Uid == service.Uid);
+                throw new InvalidOperationException(string.Format(CultureInfo.InvariantCulture, PlatformResources.ExtensionWithSameUidAlreadyRegisteredErrorMessage, service.Uid, currentRegisteredExtension.GetType()));
+            }
```
Evangelink, line 71: Why not simply `Where` to collect all potential duplicates (although we should not have any).

src/Platform/Microsoft.Testing.Platform/TestHostOrcherstrator/TestHostOrchestratorManager.cs
```diff
@@ -48,4 +49,37 @@ public async Task<TestHostOrchestratorConfiguration> BuildAsync(ServiceProvider
...
+    {
+        List<ITestHostOrchestratorApplicationLifetime> lifetimes = [];
+        foreach (Func<IServiceProvider, ITestHostOrchestratorApplicationLifetime> testHostOrchestratorApplicationLifetimeFactory in _testHostOrchestratorApplicationLifetimeFactories)
+        {
+            ITestHostOrchestratorApplicationLifetime service = testHostOrchestratorApplicationLifetimeFactory(serviceProvider);
+
+            // Check if we have already extensions of the same type with same id registered
+            if (lifetimes.Any(x => x.Uid == service.Uid))
+            {
+                ITestHostOrchestratorApplicationLifetime currentRegisteredExtension = lifetimes.Single(x => x.Uid == service.Uid);
```
Evangelink, line 69: I am never a big fan of `Single` API because when it throws it's hard to debug, you don't know what was there.

Evangelink (approved):

src/Platform/Microsoft.Testing.Platform/TestHostOrcherstrator/TestHostOrchestratorManager.cs
Youssef1313: I don't know. This is a copy of https://github.com/microsoft/testfx/blob/b3b81f44e18c4f1c5acdbb3974791b3023e158d1/src/Platform/Microsoft.Testing.Platform/TestHost/TestHostManager.cs#L104-L109 😄

src/Platform/Microsoft.Testing.Platform/TestHostOrcherstrator/TestHostOrchestratorManager.cs
Youssef1313: Same comment, it's a copy of https://github.com/microsoft/testfx/blob/b3b81f44e18c4f1c5acdbb3974791b3023e158d1/src/Platform/Microsoft.Testing.Platform/TestHost/TestHostManager.cs#L104-L109 😄

src/Platform/Microsoft.Testing.Platform/TestHostOrcherstrator/TestHostOrchestratorManager.cs
Evangelink: Oh right ok, not a big fan but let's use the same style then :)

Youssef1313: I'll create an issue to fix it in both places.

PR #5744, opened 2025-06-12
Exclude folders

nohwnd:
Exclude docs and infra from full PR build for faster iteration.


Youssef1313 (approved):

Youssef1313:
I feel like it may end up being an Azure bug (see https://github.com/microsoft/testfx/pull/5745), but we can try.

nohwnd:
YEah I would suspect the same, the exclude rules are super fiddly.

PR #5747, opened 2025-06-12
Revert "Exclude folders"

nohwnd:
Reverts microsoft/testfx#5744


nohwnd:
Reverting because it does not work, the main microsoft.testfx check is correctly skipped, but the child checks are still scheduled for some reason.

Youssef1313 (approved):

Youssef1313:
Looks like an Azure regression :)

PR #5750, opened 2025-06-12
Fix TraceListenerManager thread safety issue

### LINKED ISSUE #2616: `TraceListnerManager` is not thread safe
`TraceListenerManager` can be created in parallel and in the constructor set the static console `TextWriter` without any lock. So when for instance one test complete and restore the original `TextWriter` another concurrent `Console.WriteLine` will pickup wrong writer.

https://github.com/microsoft/testfx/blob/261430037d9e526d398e6d5fa5aac3c88eb6597c/src/Adapter/MSTestAdapter.PlatformServices/Services/TraceListenerManager.cs#L42-L43

Youssef1313:
Fixes #2616

Instead of repeatedly setting Console.Out/Console.Error and resetting to original in a thread-unsafe way, we now set Conosle.Out/Console.Error once in the UnitTestRunner constructor.

Additionally, the logic of capturing is way simplified now.

Each step of the test lifecycle (asm init, asm cleanup, class init, test, class cleanup, and asm cleanup) will set its current TestContext. The current TestContext is async local, so it's isolated properly when running in parallel. Then when each step creates its test result, it just needs to get the data from the current test context.

The redirected out/error will simply write to the current test context if there is one, or write to the original if there isn't one. This simplifies the logic a lot and makes it much easier to follow.

Note that there is a lot of code that can be deleted with this change, but we cannot do so now as it affects public API. I kept the public API the same. We simply just are not using it in the implementation.

### DIFF
--- a/src/Adapter/MSTestAdapter.PlatformServices/Execution/ConsoleErrorCapturer.cs
+++ b/src/Adapter/MSTestAdapter.PlatformServices/Execution/ConsoleErrorCapturer.cs
@@ -37,4 +37,16 @@ internal sealed class ConsoleErrorCapturer : TextWriter
             _originalConsoleErr.Write(value);
         }
     }
+
+    public override void Write(char[] buffer, int index, int count)
+    {
+        if (TestContextImplementation.CurrentTestContext is { } testContext)
+        {
+            testContext.WriteConsoleErr(buffer, index, count);
+        }
+        else
+        {
+            _originalConsoleErr.Write(buffer, index, count);
+        }
+    }
 }
--- a/src/Adapter/MSTestAdapter.PlatformServices/Execution/ConsoleErrorCapturer.cs
+++ b/src/Adapter/MSTestAdapter.PlatformServices/Execution/ConsoleErrorCapturer.cs
@@ -1,0 +1,40 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using Microsoft.VisualStudio.TestPlatform.MSTestAdapter.PlatformServices;
+
+namespace Microsoft.VisualStudio.TestPlatform.MSTest.TestAdapter.Execution;
+
+internal sealed class ConsoleErrorCapturer : TextWriter
+{
+    private readonly TextWriter _originalConsoleErr;
+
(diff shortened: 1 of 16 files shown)


src/Adapter/MSTestAdapter.PlatformServices/Execution/ConsoleErrorCapturer.cs
```diff
@@ -0,0 +1,52 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using Microsoft.VisualStudio.TestPlatform.MSTestAdapter.PlatformServices;
+
+namespace Microsoft.VisualStudio.TestPlatform.MSTest.TestAdapter.Execution;
+
+internal sealed class ConsoleErrorCapturer : TextWriter
```
Evangelink, line 8: Nit: it feels move like a `Router` or `Forwarder` than a capturer.

src/Adapter/MSTestAdapter.PlatformServices/Execution/TestAssemblyInfo.cs
```diff
@@ -146,15 +140,18 @@ internal void RunAssemblyInitialize(TestContext testContext, out LogMessageListe
                 // Perform a check again.
                 if (!IsAssemblyInitializeExecuted)
                 {
-                    LogMessageListener? logMessageListener = null;
```
Evangelink, line 149: Is this class still used somewhere? If not, can we drop it?

Youssef1313: @Evangelink Not used. But it's public. I don't think any one is using it. So I'm personally fine dropping if you are okay with that. Also, PlatformServices public APIs aren't intended for use by users anyways.

Evangelink: Let's wait for v4, it'll be here soon.

src/Adapter/MSTestAdapter.PlatformServices/Execution/ConsoleErrorCapturer.cs
Youssef1313: I'll follow-up in a new PR (will let Copilot do it)

PR #5771, opened 2025-06-13
[MTP] Few renames around "test host" terminology

Youssef1313:


src/Platform/Microsoft.Testing.Platform/Hosts/IHost.cs
```diff
@@ -0,0 +1,14 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+namespace Microsoft.Testing.Platform.Hosts;
+
+/// <summary>
+/// This represents a host (i.e, a process). It could be a "test host" (the process that actually runs the tests),
+/// a "test host controller" (runs out-of-process extensions like dump extensions which observes the test host), or
+/// a "test host orchestrator" (like Retry extension, which coordinates how test hosts should be run)
```
Youssef1313:
```suggestion
/// a "test host orchestrator" (like Retry extension, which coordinates how test hosts should be run).
```

src/Platform/Microsoft.Testing.Platform/Hosts/IHost.cs
```diff
@@ -0,0 +1,14 @@
...
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+namespace Microsoft.Testing.Platform.Hosts;
+
+/// <summary>
+/// This represents a host (i.e, a process). It could be a "test host" (the process that actually runs the tests),
+/// a "test host controller" (runs out-of-process extensions like dump extensions which observes the test host), or
+/// a "test host orchestrator" (like Retry extension, which coordinates how test hosts should be run).
+/// </summary>
+internal interface IHost
```
Evangelink, line 11: I'm wondering if we have some interest in keeping the specialized host interfaces because a test host is not a test host cotntroller etc and we know that a controller won't have a controller under it.

Youssef1313: @Evangelink I think we can adjust that later if needed?
Today we don't need those extra interfaces, I think.

Evangelink: Sure fine by me.

PR #5778, opened 2025-06-15
Store the actual data to instead of deserializing

Youssef1313:
When we get a run request with a given filter, we do:

1. Discovery
    - We call `UnitTestDiscoverer.DiscoverTestsInSource` which creates `UnitTestElement`s
    - It then converts `UnitTestElement`s to `TestCase`s (VSTest object model), and sends the test cases to discovery sink.
2. Execution
    - We call `ExecuteTestsAsync` with the `TestCase`s.
    - We then convert back `TestCase`s to `UnitTestElement`s that we pass to `UnitTestRunner`.

During discovery, we already have the "real data". We are only starting to lose them when moving from `UnitTestElement`s to `TestCase`s.

This PR makes the following changes:

1. Updates `AssemblyEnumerator` to store the real/original data along with the serialized data.
2. `UnitTestDiscoverer.SendTestCases` can read the data set in the previous step, and store that in CWT that maps `TestCase` to the actual data. Note that the `SendTestCases` step is exactly when we are starting to go from `UnitTestElement`s world to `TestCase`s world.
3. When converting the `TestCase`s back to `UnitTestElement`s, we set the actual data again from the CWT.
4. `TestMethodRunner` can then read the actual data instead of deserializing (which may not round trip correctly).
5. Note: I'm still keeping the `Deserialize` call if `ActualData` is null. That's probably still reachable when app domains are enabled (execution happens on different app domain). Propagating the info across app domain is probably not possible, because the `object[]` can carry user-defined types that cannot cross app domains.

### DIFF
--- a/test/UnitTests/MSTest.SelfRealExamples.UnitTests/ParameterizedTestSerializationIssue2390.cs
+++ b/test/UnitTests/MSTest.SelfRealExamples.UnitTests/ParameterizedTestSerializationIssue2390.cs
@@ -1,7 +1,6 @@
-// Copyright (c) Microsoft Corporation. All rights reserved.
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
-#if !NETFRAMEWORK // fails because of app domains.
 using Microsoft.VisualStudio.TestTools.UnitTesting;
 
 namespace MSTest.SelfRealExamples.UnitTests;
@@ -18,8 +17,12 @@ public class ParameterizedTestSerializationIssue2390
     {
         object[] array = (object[])nested;
         object actual = Assert.ContainsSingle(array);
-        Assert.AreEqual(expected.GetType(), ((object[])nested)[0].GetType());
+#if NETFRAMEWORK
+        // Buggy behavior, because of app domains.
+        Assert.AreEqual(typeof(int), actual.GetType());
+#else
+        Assert.AreEqual(expected.GetType(), actual.GetType());
         Assert.AreEqual(expected, actual);
+#endif
     }
 }
-#endif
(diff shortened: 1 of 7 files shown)


test/UnitTests/MSTest.SelfRealExamples.UnitTests/ParameterizedTestSerializationIssue2390.cs
```diff
@@ -0,0 +1,25 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
```
Evangelink, line 1: Is it meant to be commited? Is this not supposed to be some acceptance test instead?

Evangelink (dismissed):
I think a regular acceptance test where we could disable appdomain (via runsettings) to check it also work for netfx and a different test (same code) showing it fails when appdomain are enabled would be a better future-proof documentation.

Accepting anyways so you can decide if you merge as-is or not.

test/UnitTests/MSTest.SelfRealExamples.UnitTests/ParameterizedTestSerializationIssue2390.cs
```diff
@@ -0,0 +1,25 @@
...
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+#if !NETFRAMEWORK // fails because of app domains.
```
Youssef1313: I'm re-thinking and actually surprised this is running with app domains. This is an MTP test project and we should have app domains disabled by default

Youssef1313: It's not disabled by default in MTP.

Evangelink (approved):

PR #5779, opened 2025-06-16
Refactor ClassCleanupManager

Youssef1313:


src/Adapter/MSTestAdapter.PlatformServices/Execution/ClassCleanupManager.cs
```diff
@@ -11,56 +11,91 @@ namespace Microsoft.VisualStudio.TestPlatform.MSTest.TestAdapter.Execution;
...
+    }
+
+    /// <summary>
+    /// Gets the class cleanup lifecycle for the class, if set.
+    /// </summary>
+    /// <param name="classInfo">The class to inspect.</param>
+    /// <returns>Returns <see cref="ClassCleanupBehavior"/> if provided, otherwise <c>null</c>.</returns>
+    private ClassCleanupBehavior GetClassCleanupBehavior(TestClassInfo classInfo)
+    {
+        // TODO: not discovery related but seems expensive and unnecessary, because we do inheritance lookup, and to put the method into the stack we've already did this lookup before?
```
Evangelink, line 68: Only done in Debug so never for our users. This is more of a safeguard to ensure we use things correctly.

Evangelink (dismissed):
There's a small change of behavior that if for some reason the same class was calling twice the cleanup we could reach negative number while currently we would return

src/Adapter/MSTestAdapter.PlatformServices/Execution/ClassCleanupManager.cs
Youssef1313: Not sure what you mean here

Youssef1313: Note: the comment and logic was moved from `ReflectHelper`, but I don't actually understand what the comment tries to say. This code path should be unreachable during discovery and is used only for running.

Maybe historically this was reachable during discovery but is no longer the case?

nohwnd: I think the comment refers to the code below (on line 87). We have a stack of class cleanup methods there, which we probably identified by calling code similar to what is on line 87, and now we take each of those methods and again go through atributes to re-check that it still has the class cleanup attribute and what is the mode on that.

Instead it probably be better to put the method + cleanup behavior on the stack and do the lookup once.

Youssef1313: Ahh I see. I think it can be addressed separately?

nohwnd: Yes sure.

Evangelink (approved):
There's a small change of behavior that if for some reason the same class was calling twice the cleanup we could reach negative number while currently we would return

PR #5799, opened 2025-06-17
Cleanup RunSettingsPatcher

Youssef1313:
`DisableAppDomain` is already `false` by default. In MSTest v4, if we decided to change the default, we want to make sure the bridge is following.


nohwnd (approved):
should be okay, imho I wrote the original code, to set disable appdomain to true when not specified, when optimizing performance for xunit2 and other adapters we were wrapping at that time.

Then it got changed to not-disable appdomains because we had problems with mstest users relying on it.

So now it is okay to remove altogether because it does not achieve the goal of setting the faster default.

PR #5805, opened 2025-06-17
Prefer async over blocking

Youssef1313:

### DIFF
--- a/test/UnitTests/MSTestAdapter.PlatformServices.UnitTests/Execution/TestMethodInfoTests.cs
+++ b/test/UnitTests/MSTestAdapter.PlatformServices.UnitTests/Execution/TestMethodInfoTests.cs
@@ -82,7 +82,7 @@ public class TestMethodInfoTests : TestContainer
 
     #region TestMethod invoke scenarios
 
-    public void TestMethodInfoInvokeShouldWaitForAsyncTestMethodsToComplete()
+    public async Task TestMethodInfoInvokeShouldWaitForAsyncTestMethodsToComplete()
     {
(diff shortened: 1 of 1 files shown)


src/Adapter/MSTestAdapter.PlatformServices/Execution/TestMethodInfo.cs
```diff
@@ -671,13 +671,22 @@ private static TestFailedException HandleMethodException(Exception ex, Exception
     /// <param name="executionContext">The execution context to run on.</param>
     /// <param name="timeoutTokenSource">The timeout token source.</param>
     [SuppressMessage("Microsoft.Design", "CA1031:DoNotCatchGeneralExceptionTypes", Justification = "Requirement is to handle all kinds of user exceptions and message appropriately.")]
-    private void RunTestCleanupMethod(TestResult result, ExecutionContext? executionContext, CancellationTokenSource? timeoutTokenSource)
+    private
+#if NET6_0_OR_GREATER
+        async
+#endif
+        Task
```
nohwnd, line 678: why this? (too stupid to figure it out myself :D)

Youssef1313: @nohwnd The only async call is `DisposeAsync` which is net6.0+ only.

Youssef1313: We could keep it `void` on non-net6.0, but then callers need to be conditional as well.

PR #5806, opened 2025-06-18
Attempt to fix race condition building MSTest.GlobalConfigsGenerator

Youssef1313:
I think `dotnet run` could be invoking a build which is racing with the main build of GlobalConfigsGenerator. Adding a `ProjectReference` to force GlobalConfigsGenerator to be built before MSTest.Analyzers.Package. Then invoking `dotnet run` with `--no-build`.


Youssef1313:
Replaced with https://github.com/microsoft/testfx/pull/5807

PR #5807, opened 2025-06-18
Ensure MSTest.GlobalConfigsGenerator is built before MSTest.Analyzers.Package

Youssef1313:
I think `dotnet run` could be invoking a build which is racing with the main build of GlobalConfigsGenerator. Adding a `ProjectReference` to force GlobalConfigsGenerator to be built before MSTest.Analyzers.Package. Then invoking `dotnet run` with `--no-build`.


TestFx.sln
```diff
@@ -163,6 +163,9 @@ EndProject
 Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "MSTest.Analyzers.CodeFixes", "src\Analyzers\MSTest.Analyzers.CodeFixes\MSTest.Analyzers.CodeFixes.csproj", "{462B0201-1C26-4951-97C9-722C2A58EF8C}"
 EndProject
 Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "MSTest.Analyzers.Package", "src\Analyzers\MSTest.Analyzers.Package\MSTest.Analyzers.Package.csproj", "{DC068986-7549-4B75-8EFC-A9958FD5CF88}"
+	ProjectSection(ProjectDependencies) = postProject
+		{A85AA656-6DB6-4A0B-AA80-CBB4058B3DDB} = {A85AA656-6DB6-4A0B-AA80-CBB4058B3DDB}
+	EndProjectSection
```
Youssef1313, line 168: We only use `.sln` on Windows currently. So this won't work for Linux/macOS.

https://github.com/microsoft/testfx/blob/7bf53d17d85a1e99a61130febeacc1581aff8a4c/eng/Build.props#L8-L10

PR #5811, opened 2025-06-18
Deprecates `That` property and suggest `Instance` instead

Evangelink:
This moves will allows to introduce the following fluent `Assert.That(x).IsYYY` pattern

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -76,7 +76,15 @@ stages:
             filePath: ./eng/install-procdump.ps1
             failOnStderr: true
             showWarnings: true
-
+        - task: PowerShell@2
+          displayName: 'Enable local dumps'
+          inputs:
+            targetType: 'inline'
+            script: |
+              New-Item -Path $(Build.SourcesDirectory)\artifacts\CrashDumps -ItemType Directory -Force
+              Get-ChildItem "HKLM:\SOFTWARE\Microsoft\Windows\Windows Error Reporting\LocalDumps"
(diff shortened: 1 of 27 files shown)


Youssef1313 (dismissed):

Evangelink:
@cagrin your syntax is a little bit different `Assert.That.Contains(x, collection)` which is equivalent to `Assert.Contains(x, collection)` vs `Assert.That(collection).Contains(x)`. Moreoever, we have implemented `Assert.That(expression)` feature in #5991. We have been thinking about a couple of alternatives for the name but this one felt the more natural.

Looking up to GitHub repos using MSTest `That`, we saw only handful usage so it felt like it would not be too painful to update`. 

How many instances does this change impact on your codebase?

cagrin:
Yes, I agree, the name That will be best for this type of assertion extension. I'm waiting for new methods. I'll try to adapt my extensions to the new Instance name or find another convinient way.

mdrexel:
Hello, I saw that 3.10.1 had released and went to check the changelog, and I just wanted to chime in that deprecating the `.That` extension point is very painful... we have tens of thousands of usages via an internal helper library 😢 
Screenshot 2025-08-06 143459

orbitalbits:
Seconding what @mdrexel just wrote here. We have tens of thousands of uses of the `.That` extension point. Deprecating this extension is going to incur a very significant negative impact on us. Please reconsider this change.

Youssef1313:
> Great call @Youssef1313 and thanks everyone for the feedback! If extension methods aren't working we can go back to a different naming like Assert.Subject, Assert.ThatSubject, Assert.Expression ...

Either a different name, or with "extension members" (C# 14).

Issue #1285, opened 2022-09-27
MSTest v4 Breaking Changes


Evangelink:
## Description

The goal of this ticket is to announce the upcoming breaking changes in MSTest v4.

### Planned

- https://github.com/microsoft/testfx/issues/1284

Nirmal4G:
I also propose a new Shared Framework of testing tools and frameworks in .NET SDK along with a new Test SDK for the Project (*not just a simple rename as requested in #2693*). May be not for v4 but for v5 maybe? .NET 10? ***Basically, a built-in Testing support from the .NET SDK.***

See #3254 for the proposal and discussion.

PR #5830, opened 2025-06-19
Add CallerFilePath/CallerLineNumber to TestMethodAttribute

### LINKED ISSUE #5500: [MSTest v4][breaking] Use CallerFilePath and CallerLineNumber in TestMethodAttribute
This would avoid us to do any magic to resolve source information, and should hopefully work for F# as well. No more PDB processing is required.

I would also remove the constructor that takes `displayName` and convert it to a property.

Youssef1313:
Closes #5500
Related to #1285

### DIFF
--- a/src/Adapter/MSTestAdapter.PlatformServices/ObjectModel/UnitTestElement.cs
+++ b/src/Adapter/MSTestAdapter.PlatformServices/ObjectModel/UnitTestElement.cs
@@ -64,9 +64,9 @@ internal sealed class UnitTestElement
     /// </summary>
     public string? DisplayName { get; set; }
 
-    internal string? CallerFilePath { get; set; }
+    internal string? DeclaringFilePath { get; set; }
 
-    internal int? CallerLineNumber { get; set; }
+    internal int? DeclaringLineNumber { get; set; }
 
     /// <summary>
     /// Gets or sets the compiler generated type name for async test method.
@@ -203,8 +203,8 @@ internal sealed class UnitTestElement
             }
         }
 
-        testCase.LineNumber = CallerLineNumber ?? -1;
-        testCase.CodeFilePath = CallerFilePath;
+        testCase.LineNumber = DeclaringLineNumber ?? -1;
+        testCase.CodeFilePath = DeclaringFilePath;
 
         SetTestCaseId(testCase, testFullName);
 
--- a/src/TestFramework/TestFramework/Attributes/TestMethod/TestMethodAttribute.cs
+++ b/src/TestFramework/TestFramework/Attributes/TestMethod/TestMethodAttribute.cs
@@ -38,8 +38,8 @@ public class TestMethodAttribute : Attribute
     /// </summary>
     public TestMethodAttribute([CallerFilePath] string callerFilePath = "", [CallerLineNumber] int callerLineNumber = -1)
     {
-        CallerFilePath = callerFilePath;
-        CallerLineNumber = callerLineNumber <= 0 ? null : callerLineNumber;
+        DeclaringFilePath = callerFilePath;
+        DeclaringLineNumber = callerLineNumber <= 0 ? null : callerLineNumber;
     }
 
     /// <summary>
@@ -50,12 +50,12 @@ public class TestMethodAttribute : Attribute
     /// <summary>
     /// Gets the file path that declares the attribute.
     /// </summary>
-    public string CallerFilePath { get; }
+    public string DeclaringFilePath { get; }
 
     /// <summary>
-    /// Gets the line number within <see cref="CallerFilePath"/> that declares the attribute.
+    /// Gets the line number within <see cref="DeclaringFilePath"/> that declares the attribute.
     /// </summary>
-    public int? CallerLineNumber { get; }
+    public int? DeclaringLineNumber { get; }
 
     /// <summary>
     /// Executes a test method.
--- a/src/Adapter/MSTestAdapter.PlatformServices/Discovery/TypeEnumerator.cs
+++ b/src/Adapter/MSTestAdapter.PlatformServices/Discovery/TypeEnumerator.cs
@@ -193,8 +193,8 @@ internal class TypeEnumerator
 
         // get DisplayName from TestMethodAttribute (or any inherited attribute)
         testElement.DisplayName = testMethodAttribute?.DisplayName ?? method.Name;
-        testElement.CallerFilePath = testMethodAttribute?.CallerFilePath;
-        testElement.CallerLineNumber = testMethodAttribute?.CallerLineNumber;
+        testElement.DeclaringFilePath = testMethodAttribute?.DeclaringFilePath;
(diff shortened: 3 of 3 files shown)


src/Adapter/MSTestAdapter.PlatformServices/ObjectModel/UnitTestElement.cs
```diff
@@ -64,6 +64,10 @@ public UnitTestElement(TestMethod testMethod)
     /// </summary>
     public string? DisplayName { get; set; }
 
+    internal string? CallerFilePath { get; set; }
+
+    internal int CallerLineNumber { get; set; } = -1;
```
Evangelink: Shall it be nullable instead?

src/Adapter/MSTestAdapter.PlatformServices/ObjectModel/UnitTestElement.cs
```diff
@@ -64,6 +64,10 @@ public UnitTestElement(TestMethod testMethod)
     /// </summary>
     public string? DisplayName { get; set; }
 
+    internal string? CallerFilePath { get; set; }
```
Evangelink: Shall we use more general names like `DeclaringFilePath`?

src/Adapter/MSTestAdapter.PlatformServices/ObjectModel/UnitTestElement.cs
Evangelink: Same question here about naming

src/Adapter/MSTestAdapter.PlatformServices/RunConfigurationSettings.cs
```diff
@@ -127,19 +117,6 @@ private static RunConfigurationSettings ToSettings(XmlReader reader)
                 string elementName = reader.Name.ToUpperInvariant();
                 switch (elementName)
                 {
-                    case "COLLECTSOURCEINFORMATION":
```
Evangelink, line 130: Shall we emit a warning to say this is no longer supported so that people can cleanup their runsettings? I don't know if that's changing anything in VSTest side.

src/Adapter/MSTestAdapter.PlatformServices/RunConfigurationSettings.cs
```diff
@@ -173,17 +150,9 @@ internal static RunConfigurationSettings SetRunConfigurationSettingsFromConfig(I
         // Expected format of the json is: -
         // "mstest" : {
         //  "execution": {
-        //    "collectSourceInformation": true,
         //    "executionApartmentState": "STA"
         //  }
         // }
-        if (bool.TryParse(configuration["mstest:execution:collectSourceInformation"], out bool collectSourceInformation))
```
Evangelink, line 180: Same comment here about warning.

src/TestFramework/TestFramework/Attributes/TestMethod/TestMethodAttribute.cs
```diff
@@ -36,25 +36,26 @@ public class TestMethodAttribute : Attribute
...
-    /// Initializes a new instance of the <see cref="TestMethodAttribute"/> class.
+    /// Gets or sets display name for the test.
     /// </summary>
-    /// <param name="displayName">Display name for the test.</param>
-    public TestMethodAttribute(string? displayName)
-    {
-        DisplayName = displayName;
-        UseAsync = GetType() == typeof(TestMethodAttribute);
-    }
+    public string? DisplayName { get; set; }
```
Evangelink: Maybe `init` instead of settable? I don't know if that works for attributes.

src/Adapter/MSTestAdapter.PlatformServices/RunConfigurationSettings.cs
Youssef1313: Visual Studio will always pass it. A warning here will be too annoying/aggressive. I think it's safe to simply ignore. The usage of this setting was intended to be mostly for perf (i.e, set it to false and let TE try to figure out on its own by FQN and/or ManagedType/ManagedMethod)

src/TestFramework/TestFramework/Attributes/TestMethod/TestMethodAttribute.cs
Youssef1313: I don't have a strong opinion. But sure, I can go ahead with that.

src/Adapter/MSTestAdapter.PlatformServices/ObjectModel/UnitTestElement.cs
Youssef1313: Hmm, I think that might indeed be a better idea. Good point.

src/Adapter/MSTestAdapter.PlatformServices/ObjectModel/UnitTestElement.cs
Youssef1313: @Evangelink This comment applies to the public properties on the attribute as well?

src/Adapter/MSTestAdapter.PlatformServices/ObjectModel/UnitTestElement.cs
Youssef1313: Changing to `int?` causes:

> src/TestFramework/TestFramework/Attributes/TestMethod/TestMethodAttribute.cs(34,14): error CS3015: (NETCORE_ENGINEERING_TELEMETRY=Build) 'TestMethodAttribute' has no accessible constructors which use only CLS-compliant types

I'm not sure if it's okay to suppress or not (and is it going to work for F# for example?).

Youssef1313: I tried and doesn't work for F# if nullable.

```
C:\Users\ygerges\source\repos\TestProject3\TestProject3\Test1.fs(13,7): error FS0073: internal error: The type 'System.Nullable`1' may not be used as a custom attribute value
```

Youssef1313: I kept the properties as nullable, but only attribute constructor uses `int`. If 0 or negative, we set the property to null.

Evangelink: We should probably remove CLS-compliance support as part of v4 as it's no longer something Roslyn and runtime push for. We should mention that for v4.

Youssef1313: @Evangelink It breaks F# though, maybe not CLS compliance in general, but at least in this specific case.

Evangelink (approved):

src/Adapter/MSTestAdapter.PlatformServices/ObjectModel/UnitTestElement.cs
Evangelink: Yes

PR #5843, opened 2025-06-20
Rename assemblies and don't fix version to 14

### LINKED ISSUE #1607: Microsoft.VisualStudio.TestPlatform.MSTestAdapter.PlatformServices errors due to path length not handled well on UWP
## Description

If you use `MSTest.TestAdapter` on a UWP project, the very long name of the `Microsoft.VisualStudio.TestPlatform.MSTestAdapter.PlatformServices` component can cause the resource generation step to fail because the Windows maximum path length is exceeded. The error message reported in this case does not make it at all obvious that path lengths are the problem.

I am using MSTest.TestAdapter 3.0.2 on Visual Studio 2022 17.4.5

## Steps To Reproduce

In Visual Studio, create a UWP test runner project using `MSTest.TestAdapter` (or clone a repo that contains one) such that the fully-qualified name of the folder is longer than 80 characters. For example, the name of the project folder in which I encountered this problem was:

`C:\Users\matth\source\repos\reactive\Rx.NET\Source\tests\Tests.System.Reactive.Uwp.DeviceRunner`

That folder path is 95 characters long. This is more than enough to trigger the problem.

Building the project causes this error:

```
1>C:\Program Files (x86)\Microsoft SDKs\UWPNuGetPackages\microsoft.net.uwpcoreruntimesdk\2.2.14\tools\CoreRuntime\Microsoft.Net.CoreRuntime.targets(195,9): error : Framework resource extraction failed. Could not find a part of the path 'C:\Users\matth\source\repos\reactive\Rx.NET\Source\tests\Tests.System.Reactive.Uwp.DeviceRunner\obj\x86\Debug\Microsoft.VisualStudio.TestPlatform.MSTestAdapter.PlatformServices\en\Microsoft.VisualStudio.TestPlatform.MSTestAdapter.PlatformServices.Resources.Resource.resw'.
```

The path in the error message is 270 characters long, 10 over the traditional Windows maximum path length. As is common for modern Windows systems, this particular machine is configured to allow files longer than that old limit. But paths that long cause problems for code that does not know how to work with long filenames.

## Expected behavior

Ideally, it should just be able to work with longer filenames.

Failing that, it should report that this is a path length issue.

Furthermore, it would be worth considering giving this component a much shorter name and/or making the name of the embedded resource shorter (does it really need to end with ".Resource.Resource"?) so that it is less likely to trigger this sort of problem. The current names of these components mean that this will add 175 characters to your project folder ("\obj\x86\Debug\Microsoft.VisualStudio.TestPlatform.MSTestAdapter.PlatformServices\en\Microsoft.VisualStudio.TestPlatform.MSTestAdapter.PlatformServices.Resources.Resource.resw") or more for Release builds (because "Release" is longer than "Debug")

## Actual behavior

The error "Could not find a part of the path" is not helpful when trying to diagnose the problem.

## Additional context

Very long component names tend to cause path length problems, so it would be great it there was any way to make the names shorter.

### LINKED ISSUE #5690: [Breaking][v4] Assembly version shouldn't be fixed to 14.0.0.0
Both test framework and adapter set `UseAssemblyVersion14` to true.

https://github.com/microsoft/testfx/blob/7b0d5d1c5049a2a485586b45bce5a98305e3d300/Directory.Build.targets#L10-L11

We shouldn't do that anymore, and MSTest assemblies should be properly versioned, similar to MTP assemblies.

This means that in v4, we will have to use a lower version (4.x) instead of 14.x, which may also be problematic that a newer version of the assembly is actually using an older version number. Thus, I also think it's safest to rename the assembly altogether.

- Instead of `Microsoft.VisualStudio.TestPlatform.TestFramework`, we simply use `MSTest.TestFramework`
- Instead of `Microsoft.VisualStudio.TestPlatform.MSTest.TestAdapter`, we simply use `MSTest.TestAdapter`.

*NOTE:* I'm suspecting that fixing the version to 14.0.0.0 may be causing issues on .NET Framework when dealing with [GAC](https://learn.microsoft.com/en-us/dotnet/framework/app-domains/gac).

Youssef1313:
Closes #5690
Closes #1607

Related to #1285


Evangelink (approved):

src/Adapter/MSTest.TestAdapter/MSTest.TestAdapter.nuspec
```diff
@@ -131,11 +131,11 @@
       deps to it in the "engine" project and keep them only under the TestingPlatformAdapter folder.
       See https://github.com/microsoft/testfx/issues/4426
     -->
-    <file src="net462\Microsoft.VisualStudio.TestPlatform.TestFramework.dll" target="buildTransitive\net462\" />
+    <file src="net462\MSTest.TestFramework.dll" target="buildTransitive\net462\" />
 
     <!-- Localization -->
     <!-- All TFMs share the same resx, copy only TestAdapter TestAdapter.PlatformServices, do NOT copy TestFramework and other dependency resource dlls. -->
-    <file src="net462\**\Microsoft.VisualStudio.TestPlatform.*Adapter*.resources.dll" target="\buildTransitive\_localization\" />
+    <file src="net462\**\*Adapter*.resources.dll" target="\buildTransitive\_localization\" />
```
Youssef1313:
```suggestion
    <file src="net462\**\MSTest.TestAdapter.resources.dll" target="\buildTransitive\_localization\" />
    <file src="net462\**\MSTestAdapter.PlatformServices.resources.dll" target="\buildTransitive\_localization\" />
```

Evangelink (approved):

PR #5845, opened 2025-06-20
Make types in platform services internal

Youssef1313:


Evangelink (approved):

dotMorten:
This change would break my custom test runners for mobile devices.

dotMorten:
Sure will do. Need to work through them all, as part of moving up to latest. This was just the first I hit and it pointed to this PR.

Issue #2365, opened 2024-02-16
Assert.AreEqual when types `A` and `B` implements `IEquatable<A>` no longer works


Evangelink:
If type `A` implements `IEquatable<A>` and type `B` also implements `IEquatable<A>` (yes of `A`), then you could call `Assert.AreEqual(instanceOfA, instanceOfB)` before and it would work, now `T` is expected to be either both `A` or both `B`, even though they could be "equatabled".

I mean it's questionable to compare two different types, but I haven't really seen this scenario discussed as breaking, so I thought I mention it here. Guess, I'll have to convert on the test, rather than inside the Equals functions.

```csharp
class A : IEquatable<A>
{
    public bool Equals(A other) { ... }
}

class B : IEquatable<A>
{
    public bool Equals(A other) { ... }
}

// ...

[TestMethod]
public void Test()
{
    var instanceOfA = new A();
    var instanceOfB = new B();

    Assert.AreEqual(instanceOfA, instanceOfB);
}
```

_Originally posted by @eXpl0it3r in https://github.com/microsoft/testfx/issues/254#issuecomment-1948829980_

Evangelink:
@eXpl0it3r Thanks for the report, I definitely didn't consider this case. I guess we could simply add a new overload as `public static void AreEqual<T>(IEquatable<T>? expected, IEquatable<T>? actual)` and it would fix this case without breaking some other scenario (from what I can think of).

Evangelink:
This is interesting case, I have never done this kind of equality. So with your example B.Equals(A) is true but A.Equals(B) is false. At least with the dumb implementation of having `Equals(object obj)` calling `Equals(T t)`.

Evangelink:
We will be shipping the bugfix release this week (hopefully Wednesday).

Youssef1313:
@eXpl0it3r I don't think the fix done here was the right thing to do, and we are going to remove the `IEquatable<T>` overload in MSTest v4. Do you have many occurrences of this pattern that will make the migration harder for you? The needed code change will simply be changing `Assert.AreEqual(..., ...)` to `Assert.AreEqual<object>(..., ...)`

PR #5850, opened 2025-06-22
Remove Assert.AreEqual for IEquatable<T>

Youssef1313:
I don't think it was a good idea to introduce this. The APIs were added to fix https://github.com/microsoft/testfx/issues/2365, but I think it should be by-design.


Youssef1313:
@Evangelink I'm going ahead and merging this based on our discussion.

PR #5851, opened 2025-06-22
Add CallerArgumentExpression overloads to Assert.IsTrue/IsFalse/IsNull/IsNotNull

Youssef1313:


src/TestFramework/TestFramework/Assertions/Assert.IsNull.cs
```diff
@@ -163,7 +145,7 @@ public static void IsNull(object? value, [InterpolatedStringHandlerArgument(name
     /// <exception cref="AssertFailedException">
     /// Thrown if <paramref name="value"/> is not null.
     /// </exception>
-    public static void IsNull(object? value, string? message)
+    public static void IsNull(object? value, [CallerArgumentExpression(nameof(value))] string message = "")
```
Evangelink: We can discuss about the design, what I had in mind was to still allow the user to set a custom message but for us to use the capture mechanism to enrich the default message (always displayed - or with some verbosity setting in case).

Evangelink: I don't know if we should have an analyzer to ensure people don't set this value or if we should use the trick from @sharwell (bait and switch).

src/TestFramework/TestFramework/Assertions/Assert.IsNull.cs
```diff
@@ -160,14 +142,17 @@ public static void IsNull(object? value, [InterpolatedStringHandlerArgument(name
     /// The message to include in the exception when <paramref name="value"/>
     /// is not null. The message is shown in test results.
     /// </param>
+    /// <param name="valueExpression">
+    /// The syntactic expression of value as given by the compiler via caller argument expression.
```
Evangelink, line 145: Let's add a comment saying this should not be set by the user (I know it's implied by the sentence).

src/TestFramework/TestFramework/Assertions/Assert.IsNull.cs
Youssef1313: I could be missing something, but I don't see how that trick can prevent users from passing the value. If the overload is available in the assembly passed to `Csc`, then it's available.

The trick is useful if we want to preserve binary compatibility and keep `Assert.IsNull(object)` but let the compiler prefer `Assert.IsNull(object, /*caller arg expression*/ string = "")`, by passing an assembly that only has the latter to `Csc`, but at run time, we keep the assembly that has both overloads.

src/TestFramework/TestFramework/Assertions/Assert.IsNull.cs
Youssef1313: Done.

Evangelink (approved):
Not a big fan of the current message but let's merge as-is and I'll suggest a bigger assertion message rename.

Issue #5614, opened 2025-05-20
Drop support for < net8.0 in MSTest and MTP


Youssef1313:
As VSTest latest version supports only net8.0+ and net462, I think we should take the same step in MSTest and MTP.

Evangelink:
You would want this to be part of v4 as it's a technical breaking change.

Youssef1313:
@Evangelink The reason I'm putting it in 3.10 is that a similar breaking change was taken already in VSTest in a minor release. https://www.nuget.org/packages/Microsoft.NET.Test.Sdk/17.14.0 is net8.0+ while https://www.nuget.org/packages/Microsoft.NET.Test.Sdk/17.13.0 is netcoreapp3.1+.

Also, at least in theory, older TFMs should still work for the case of MSTest and MTP via netstandard2.0. There could be some problems we are not anticipating, but they will probably be fixable?

Evangelink:
Yes and I am (was) against the change but VSTest is following VS versioning scheme so they don't guarantee any semantic meaning while we do. I think it's important to respect that, I do expect still a lot of teams to use MSTest with non-supported TFMs

nohwnd:
I don't see how netstandard2.0 would produce a runnable exe for MTP, while technically you can run it, it won't work out of the box, and so it will be a breaking change.

On the other hand it seems like self-imposed limitation to have to make big breaking changes in mstest v4 and mtp v2, maybe those can move to "vNext" and we can follow normal semantic versioning with smaller breaking changes in v4, v5 etc.

Youssef1313:
Moving to v4 then. We will then have to be careful not to update VSTest dependency to 17.14. 😕

Romfos:
Any plans for V4? Maybe release date?

pavelhorak:
We're currently aligning internally on the release timeline for V4. We expect to have a confirmed date by the end of the month and will share an update as soon as it's finalized.

PR #5852, opened 2025-06-22
Drop unsupported TFMs

Youssef1313:
The MSTest part of https://github.com/microsoft/testfx/issues/5614

Related to #1285

### DIFF
--- a/test/UnitTests/MSTestAdapter.PlatformServices.UnitTests/Execution/TestMethodInfoTests.cs
+++ b/test/UnitTests/MSTestAdapter.PlatformServices.UnitTests/Execution/TestMethodInfoTests.cs
@@ -331,11 +331,11 @@ public class TestMethodInfoTests : TestContainer
 
         Verify(exception is not null);
         Verify(exception.StackTraceInformation is not null);
-        // NOTE: On net8.0 and later, the first frame is  System.Reflection.RuntimeConstructorInfo.Invoke(BindingFlags invokeAttr, Binder binder, Object[] parameters, CultureInfo culture)
+        // NOTE: On net8.0 and later, the first frame is System.Reflection.MethodBaseInvoker.ThrowTargetParameterCountException()
         // So, we do a consistent Contains check to check that the stack trace is there.
         Verify(
             exception.StackTraceInformation.ErrorStackTrace.Contains(
-            "   TestMethodInfoInvokeShouldSetStackTraceInformationIfTestClassConstructorThrowsWithoutInnerException", StringComparison.Ordinal));
+            "   at System.Reflection.RuntimeConstructorInfo.Invoke(BindingFlags invokeAttr, Binder binder, Object[] parameters, CultureInfo culture)", StringComparison.Ordinal));
     }
 
     public async Task TestMethodInfoInvokeShouldSetResultFilesIfTestContextHasAttachments()


Evangelink (approved):
Great!

Directory.Build.props
```diff
@@ -24,10 +24,10 @@
     <!-- The windows SDK version of UWP, Modern UWP and WinUI could be aligned -->
     <UwpMinimum>uap10.0.16299</UwpMinimum>
     <ModernUwpMinimum>net9.0-windows10.0.17763.0</ModernUwpMinimum>
-    <WinUiMinimum>net6.0-windows10.0.18362.0</WinUiMinimum>
+    <WinUiMinimum>net8.0-windows10.0.18362.0</WinUiMinimum>
 
-    <MicrosoftTestingTargetFrameworks>net6.0;net7.0;net8.0;net9.0</MicrosoftTestingTargetFrameworks>
-    <SupportedNetFrameworks>netcoreapp3.1;net6.0;net7.0;net8.0;net9.0</SupportedNetFrameworks>
+    <MicrosoftTestingTargetFrameworks>net8.0;net9.0</MicrosoftTestingTargetFrameworks>
+    <SupportedNetFrameworks>net8.0;net9.0</SupportedNetFrameworks>
```
Evangelink, line 30: Could we do a follow-up PR merging these 2 set of properties?

Directory.Build.props
```diff
@@ -24,10 +24,10 @@
     <!-- The windows SDK version of UWP, Modern UWP and WinUI could be aligned -->
     <UwpMinimum>uap10.0.16299</UwpMinimum>
     <ModernUwpMinimum>net9.0-windows10.0.17763.0</ModernUwpMinimum>
-    <WinUiMinimum>net6.0-windows10.0.18362.0</WinUiMinimum>
```
Evangelink, line 27: @Sergio0694 I think you said we could align UWP and modern UWP ending part of the version. Do you know if this can be also aligned with winui?

Directory.Build.props
Youssef1313: @Evangelink, sure. This assumes that MSTest will always be matching the TFMs of MTP though. Do we want to make that assumption going forward?

Evangelink: We can introduce back 2 properties if that were to happen. Given we support what's supported by MS, I feel like the .NET version would always match.

Directory.Build.props
Sergio0694: WinUI (WindowsAppSDK) is currently targeting `net6.0-windows10.0.17763.0` as minimum, though given .NET 6 is out of support I don't think it should be a concern here. I'd recommend using the latest LTS release with a matching Windows SDK version for the minimum WinUI target. So eg. that'd be `net8.0-windows10.0.17763.0`.

Also note that UWP .NET 9 needs to target that STS release right now just because that's the first one where we added UWP support. Once .NET 10 rolls around, you could switch the minimum version for modern UWP to also target the latest LTS release, meaning once .NET 8 goes out of support, then both UWP and WinUI could then target the same exact TFM here.

Hope this helps! 🙂

Youssef1313: > meaning once .NET 8 goes out of support, then both UWP and WinUI could then target the same exact TFM here.

Eh, that might require more infra changes on our side if we will use the same TFM. Our build output simply follows Arcade structure so it's `artifacts/bin/<ProjectName>/<Configuration>/<Tfm>`. We will want to have the output of uwp and winui builds in different directories. But we can think of that later when it will be time to use the same TFM for both.

Youssef1313:
@Sergio0694 I'm getting:

> D:\a\_work\1\s\.dotnet\sdk\10.0.100-preview.6.25309.101\Sdks\Microsoft.NET.Sdk\targets\Microsoft.NET.Sdk.targets(305,5): error NETSDK1206: Found version-specific or distribution-specific runtime identifier(s): win10-arm64, win10-x64, win10-x86. Affected libraries: Microsoft.WindowsAppSDK. In .NET 8.0 and higher, assets for version-specific and distribution-specific runtime identifiers will not be found by default. See https://aka.ms/dotnet/rid-usage for details. [D:\a\_work\1\s\src\TestFramework\TestFramework.Extensions\TestFramework.Extensions.csproj]

However, `RuntimeIdentifiers` is set to `win-x86;win-x64;win-arm64`:

image

src/Adapter/MSTest.TestAdapter/MSTest.TestAdapter.csproj
```diff
@@ -13,6 +13,15 @@
...
     <IsAotCompatible Condition="$([MSBuild]::IsTargetFrameworkCompatible('$(TargetFramework)', 'net7.0'))">$(EnableIotAnalyzers)</IsAotCompatible>
   </PropertyGroup>
+  
+  <PropertyGroup Condition=" '$(TargetFramework)' == '$(WinUiMinimum)' ">
+    <TargetPlatformVersion>10.0.18362.0</TargetPlatformVersion>
+    <TargetPlatformMinVersion>10.0.17763.0</TargetPlatformMinVersion>
+    <SupportedOSPlatformVersion>$(TargetPlatformMinVersion)</SupportedOSPlatformVersion>
+    <RuntimeIdentifiers>win-x86;win-x64;win-arm64</RuntimeIdentifiers>
+    <DefineConstants>$(DefineConstants);WIN_UI</DefineConstants>
+  </PropertyGroup>
```
Evangelink: Is it expected that this part was added?

src/Adapter/MSTestAdapter.PlatformServices/MSTestAdapter.PlatformServices.csproj
```diff
@@ -24,7 +24,7 @@
     <TargetPlatformVersion>10.0.18362.0</TargetPlatformVersion>
     <TargetPlatformMinVersion>10.0.17763.0</TargetPlatformMinVersion>
     <SupportedOSPlatformVersion>$(TargetPlatformMinVersion)</SupportedOSPlatformVersion>
-    <RuntimeIdentifiers>win10-x86;win10-x64;win10-arm64</RuntimeIdentifiers>
+    <RuntimeIdentifiers>win-x86;win-x64;win-arm64</RuntimeIdentifiers>
```
Evangelink, line 27: Is it an expected change?

src/Adapter/MSTest.TestAdapter/MSTest.TestAdapter.csproj
Youssef1313: At least for RuntimeIdentifiers, yes. The `win10-<arch>` no longer work in .NET 8. The breaking change is documented [here](https://learn.microsoft.com/en-us/dotnet/core/compatibility/sdk/8.0/rid-graph)

src/Adapter/MSTestAdapter.PlatformServices/MSTestAdapter.PlatformServices.csproj
Youssef1313: This is also linked to the breaking change documented [here](https://learn.microsoft.com/en-us/dotnet/core/compatibility/sdk/8.0/rid-graph)

Evangelink: Oh great, I didn't know about this change!

Sergio0694:
@Youssef1313 mmh that's weird. Do you not see `win10-` anywhere in the binlog? 🤔
Can you try to narrow it down and eg. only build/publish with UWP .NET 9 and WinUI 3 separately?

src/TestFramework/TestFramework.Extensions/MSTest.TestFramework.nuspec
```diff
@@ -73,46 +64,19 @@
...
-    <file src="$RepoRoot$src\TestFramework\TestFramework.Extensions\buildTransitive\net6.0AndLater\MSTest.TestFramework.targets" target="buildTransitive\net8.0\" />
+    <file src="$RepoRoot$src\TestFramework\TestFramework.Extensions\buildTransitive\net8.0AndLater\MSTest.TestFramework.targets" target="buildTransitive\net8.0\" />
     <file src="net8.0\MSTest.TestFramework.dll" target="lib\net8.0\" />
     <file src="net8.0\MSTest.TestFramework.xml" target="lib\net8.0\" />
     <file src="net8.0\**\MSTest.TestFramework.resources.dll" target="lib\net8.0\" />
     <file src="net8.0\MSTest.TestFramework.Extensions.dll" target="buildTransitive\net8.0\" />
     <file src="net8.0\MSTest.TestFramework.Extensions.xml" target="buildTransitive\net8.0\" />
     <file src="$RepoRoot$src\TestFramework\TestFramework.Extensions\build\net8.0\MSTest.TestFramework.targets" target="build\net8.0\" />
+    <file src="net8.0-windows10.0.18362.0\MSTest.TestFramework.Extensions.dll" target="buildTransitive\net8.0\winui\" />
+    <file src="net8.0-windows10.0.18362.0\MSTest.TestFramework.Extensions.xml" target="buildTransitive\net8.0\winui\" />
```
Youssef1313, line 76: Downloading the package from artifacts, looks like these are not packed

image

Not sure why exactly.

Youssef1313: I cannot repro this locally. Waiting for the current build to see

Evangelink (approved):

PR #5856, opened 2025-06-23
Refactor STA attributes logic outside of adapter

### LINKED ISSUE #5693: [MSTest][v4] STAThreadAttribute should be sealed
We should seal the attribute, and no longer make it special cased by the adapter. Instead, the attribute should override `Execute` (or `ExecuteAsync`) to invoke the test in an STA thread.

Youssef1313:
Closes #5693

Related to #1285

### DIFF
--- a/src/TestFramework/TestFramework/Attributes/TestMethod/STATestMethodAttribute.cs
+++ b/src/TestFramework/TestFramework/Attributes/TestMethod/STATestMethodAttribute.cs
@@ -19,8 +19,14 @@ public class STATestMethodAttribute : TestMethodAttribute
     {
     }
 
-    public STATestMethodAttribute(TestMethodAttribute? testMethodAttribute, string callerFilePath, int callerLineNumber)
-        : base(callerFilePath, callerLineNumber)
+    /// <summary>
+    /// Initializes a new instance of the <see cref="STATestMethodAttribute"/> class.
+    /// This constructor is intended to be called by <see cref="STATestClassAttribute"/> (or derived) to wrap an existing test method attribute.
+    /// It can also be called by a derived <see cref="STATestMethodAttribute"/> which will be called by a derived <see cref="STATestClassAttribute"/>.
+    /// </summary>
+    /// <param name="testMethodAttribute">The wrapped test method.</param>
+    public STATestMethodAttribute(TestMethodAttribute testMethodAttribute)
+        : base(testMethodAttribute.DeclaringFilePath, testMethodAttribute.DeclaringLineNumber ?? -1)
         => _testMethodAttribute = testMethodAttribute;
 
     /// <summary>
--- a/src/TestFramework/TestFramework/Attributes/TestMethod/STATestClassAttribute.cs
+++ b/src/TestFramework/TestFramework/Attributes/TestMethod/STATestClassAttribute.cs
@@ -10,6 +10,6 @@ namespace Microsoft.VisualStudio.TestTools.UnitTesting;
 public class STATestClassAttribute : TestClassAttribute
 {
     /// <inheritdoc />
-    public override TestMethodAttribute? GetTestMethodAttribute(TestMethodAttribute? testMethodAttribute)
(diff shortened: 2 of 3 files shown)


src/TestFramework/TestFramework/Attributes/TestMethod/STATestMethodAttribute.cs
```diff
@@ -9,11 +9,52 @@ namespace Microsoft.VisualStudio.TestTools.UnitTesting;
...
+
     /// <summary>
     /// Initializes a new instance of the <see cref="STATestMethodAttribute"/> class.
     /// </summary>
     public STATestMethodAttribute([CallerFilePath] string callerFilePath = "", [CallerLineNumber] int callerLineNumber = -1)
         : base(callerFilePath, callerLineNumber)
     {
     }
+
+    internal STATestMethodAttribute(TestMethodAttribute? testMethodAttribute, string callerFilePath, int callerLineNumber)
```
Evangelink: Maybe this should be `protected internal` so that it's possible to call when inheriting.

Youssef1313: My theory is that this constructor needs to only ever be called by `STATestClassAttribute`. Why would a derived class need to use it?

Youssef1313:
> Do you want to also simplify logic from

Hmm, I'm not sure exactly what to simplify there as this is more of a global setting unrelated to the attributes.

src/TestFramework/TestFramework/Attributes/TestMethod/STATestMethodAttribute.cs
Evangelink: `MySTATestClassAttribute` that would call `MySTATestMethodAttribute` exactly how you are doing now.

Evangelink:
> > Do you want to also simplify logic from
> 
> Hmm, I'm not sure exactly what to simplify there as this is more of a global setting unrelated to the attributes.

True, if we don't care much about perf, we could force `STATEstMethod` on all test methods in such context. This is obviously less performant than having a single STA move.

src/TestFramework/TestFramework/Attributes/TestMethod/STATestMethodAttribute.cs
Youssef1313: @Evangelink In that case it would be `public` not `protected internal`?

Evangelink: I made the assumption their 2 attributes would be defined in the same assembly so we need `internal` to have class attribute calling method attribute and we need `protected` to call base from the method one.

Youssef1313: I'm not sure I follow. If it's `protected internal` it means it's accessible in MSTest.TestFramework assembly **or** in derived types in any assembly. So, for users (outside of MSTest.TestFramework) it means that it's only accessible in derived types. So, a custom STATestClassAttribute can only call the constructor if it's public.

Youssef1313:
> > > Do you want to also simplify logic from
> > 
> > 
> > Hmm, I'm not sure exactly what to simplify there as this is more of a global setting unrelated to the attributes.
> 
> True, if we don't care much about perf, we could force `STATEstMethod` on all test methods in such context. This is obviously less performant than having a single STA move.

The logic in the attribute checks if we are already in STA thread, and in that case we don't create a thread and we run directly.
So, if the option is globally enabled, it doesn't matter much whether or not the attribute is used as it will almost do nothing special.

src/TestFramework/TestFramework/Attributes/TestMethod/STATestMethodAttribute.cs
Evangelink: As per our discussions both approaches are valid. Let's go with `public` to give me flexibility to users.

Youssef1313:
```suggestion
    public STATestMethodAttribute(TestMethodAttribute? testMethodAttribute, string callerFilePath, int callerLineNumber)
```

Evangelink (approved):

PR #5857, opened 2025-06-23
Fix TestContext properties nullability

### LINKED ISSUE #1403: MSTest v4 TestContext properties should not be nullable
## Describe the bug
MSTest 3.0.0-preview-20221122-01
`TestContext.TestName` is nullable, which is technically correct, but makes no sense to me.

## Expected behavior
"All" (those which make sense) Properties of `TestContext` should not be nullable and throw an exception if the underlying property dictionary has those properties removed or replaced by nulls.

Example:
`public virtual string TestName => GetProperty<string>("TestName") ?? throw new InvalidOperationException("Missing TestName");`

Having nullable on all those properties just makes tests much more verbose. And when I expect a `TestRunDirectory`, havin null there is an exceptional case anyway. Plus if I really needed an optional `TestRunDirectory` I could retrieve it via `GetProperty()`.

## Actual behavior
https://github.com/microsoft/testfx/blob/8161dfbba093957e2cd1e6aabaf099a10e2915d3/src/TestFramework/TestFramework.Extensions/TestContext.cs#L122

Youssef1313:
Closes #1403

Related to #1285

### DIFF
--- a/src/TestFramework/TestFramework.Extensions/TestContext.cs
+++ b/src/TestFramework/TestFramework.Extensions/TestContext.cs
@@ -141,22 +141,26 @@ public abstract class TestContext
     /// in the test results. Users can benefit from messages that include the fully-qualified
     /// class name in addition to the name of the test method currently being executed.
     /// </remarks>
-    public virtual string FullyQualifiedTestClassName => GetProperty<string>(FullyQualifiedTestClassNameLabel) ?? throw new InvalidOperationException();
+    public virtual string FullyQualifiedTestClassName => GetProperty<string>(FullyQualifiedTestClassNameLabel)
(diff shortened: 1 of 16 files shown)


src/TestFramework/TestFramework.Extensions/TestContext.cs
```diff
@@ -141,22 +141,22 @@ public abstract class TestContext
     /// in the test results. Users can benefit from messages that include the fully-qualified
     /// class name in addition to the name of the test method currently being executed.
     /// </remarks>
-    public virtual string? FullyQualifiedTestClassName => GetProperty<string>(FullyQualifiedTestClassNameLabel);
+    public virtual string FullyQualifiedTestClassName => GetProperty<string>(FullyQualifiedTestClassNameLabel) ?? throw new InvalidOperationException();
```
Evangelink: Let's add the localized exception message, something along the lines of `TestContext properties related to the current test are not available during Assembly or Class Initialize execution.`.

PR #5858, opened 2025-06-23
Remove CssIteration, CssProjectStructure, and Description

### LINKED ISSUE #5787: [MSTest v4] Consider removing `CssIterationAttribute`, `CssProjectStructureAttribute`, and `DescriptionAttribute`
I don't see these being used for anything useful. We should probably remove them.

Youssef1313:
Closes #5787

Related to #1285


Evangelink (approved):

avivanoff:
We actually heavily use description attributes to provide detailed information about the tests.

avivanoff:
@Evangelink, at a minimum it is part of TRX schema, see `%VSINSTALLDIR%\Xml\Schemas\vstst.xsd`.

Youssef1313:
@avivanoff Thanks for bringing this up. It's indeed specified in the TRX schema, but it seems like VSTest never included this info in TRX before?

There was a feature request here https://github.com/microsoft/vstest/issues/98 but it was closed and not implemented. So far, I don't see anything we are losing with the removal of the attribute. To me, this sounds like it will be a new feature to implement on MTP TrxReport, and support it in MSTest and VSTestBridge. And then we can reuse System.ComponentModel.DescriptionAttribute.

avivanoff:
@Youssef1313, I really see a value in the attribute. How else would you specify what unit test does?

We have tens of thousands of unit tests that use description attribute. We would have to touch thousands of files when upgrading to MSTest 4. If I am reading you right, we should switch to System.ComponentModel.DescriptionAttribute and file a feature request to support it.

Youssef1313:
> We would have to touch thousands of files when upgrading to MSTest 4.

To make it easier, you can add a global using for `System.ComponentModel` so that you don't have to touch thousands of files.

> If I am reading you right, we should switch to System.ComponentModel.DescriptionAttribute and file a feature request to support it.

Yes. I think `DescriptionAttribute` was never supported properly before as it doesn't end up being written in TRX today already.

avivanoff:
@Youssef1313, not everyone is lucky enough to be on C# 10 or later, our products still support .NET Framework 4.8.

Youssef1313:
@nohwnd You meant latest not default?

nohwnd:
(and we do use Latest in our projects, with mostly success, but we also might be more used to having our projects not build after update, and that might not be what OP wants)

avivanoff:
@nohwnd, everything I find online about using LangVersion=default warns that it might lead to runtime errors.

Youssef1313:
@avivanoff We will add back `DescriptionAttribute`, and potentially support it in MTP TRX.
For `CssIterationAttribute` and `CssProjectStructureAttribute`, I think they are not widely used. We are obsoleting them in 3.10 https://github.com/microsoft/testfx/pull/5981.

nohwnd:
> We do not have the luxury to rely on YouTube videos, we need a clear statement in Microsoft's documentation.

Can you please tell me where the documentation says that you should not use `default`? Because in my original response I was trying to find the best way to do this, according to the documentation, rather than relying on my experience.

PR #5868, opened 2025-06-24
Remove ExpectedExceptionAttribute

### LINKED ISSUE #135: Deprecate ExpectedException attribute.
## Description
I can still write tests that use ExpectedExceptionAttribute.

## Steps to reproduce
Write this test
[TestMethod]
[ExpectedException(typeof(ArgumentException))]
public void TestMyStuff()
{
// arrange
var instanceUnderTest = new YourFunkyClass();
DoSomethingThatThrowsArgumentException();

// act // assert
Assert.ThrowsException<ArgumentException>(()=>instanceUnderTest.DoStuff("badargument");}
}
## Expected behavior
I understand backwards compatibility, but ExpectedException is WIDELY recognized to be a source of tests passing when they should be failing. Any test with ExpectedException is effectively useless unless there are absolutely no other calls in the test. You just don't know if it's failing for the reason you expect it to fail, or if it's even getting to the line you want it to fail on.

At a minimum, there needs to be a way to set the test runner to reject it so that if someone brings it in, it fails. By MSTest 3, the attribute should go away entirely.

## Actual behavior
I can still use it.

## Environment
plain Win10, VS2017.

[AB#1934068](https://devdiv.visualstudio.com/0bdbc590-a062-4c3f-b0f6-9383f67865ee/_workitems/edit/1934068)

Youssef1313:
Closes #135

Related to #1285

### DIFF
--- a/test/UnitTests/MSTestAdapter.PlatformServices.UnitTests/Execution/TestMethodInfoTests.cs
+++ b/test/UnitTests/MSTestAdapter.PlatformServices.UnitTests/Execution/TestMethodInfoTests.cs
@@ -1092,20 +1092,6 @@ public class TestMethodInfoTests : TestContainer
         Verify(!testCleanupCalled);
     }
 
-    public async Task TestMethodInfoInvokeShouldSetResultAsFailedWhenExceptionIsExpectedButIsNotThrown()
-    {
-        DummyTestClass.TestMethodBody = o => { };
-        var testMethodInfo = new TestMethodInfo(_methodInfo, _testClassInfo, _testContextImplementation)
-        {
-            TimeoutInfo = TimeoutInfo.FromTimeout(3600 * 1000),
-            Executor = _testMethodAttribute,
-        };
-        TestResult result = await testMethodInfo.InvokeAsync(null);
-        Verify(result.Outcome == UTF.UnitTestOutcome.Failed);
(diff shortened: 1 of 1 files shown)


src/TestFramework/TestFramework.Extensions/TestContext.cs
```diff
@@ -100,15 +100,6 @@ public abstract class TestContext
...
-    /// <summary>
-    /// Gets the Fully-qualified name of the class containing the test method currently being executed.
-    /// </summary>
-    /// <remarks>
-    /// This property can be useful in attributes derived from ExpectedExceptionBaseAttribute.
-    /// Those attributes have access to the test context, and provide messages that are included
-    /// in the test results. Users can benefit from messages that include the fully-qualified
-    /// class name in addition to the name of the test method currently being executed.
-    /// </remarks>
     public virtual string FullyQualifiedTestClassName => GetProperty<string>(FullyQualifiedTestClassNameLabel)
```
Youssef1313, line 106:
```suggestion
    /// <summary>
    /// Gets the Fully-qualified name of the class containing the test method currently being executed.
    /// </summary>
    public virtual string FullyQualifiedTestClassName => GetProperty<string>(FullyQualifiedTestClassNameLabel)
```

PR #5882, opened 2025-06-25
Drop netstandard2.0 from adapter

### LINKED ISSUE #5645: [Breaking][v4] MSTest.TestAdapter (and consequently MSTest metapackage) should drop netstandard2.0
MSTest.TestAdapter is intended to be used in actual test projects, not in helper projects/libraries. Actual test projects aren't supposed to be `netstandard2.0`.

The only thing we lose by this is incompatibility with Xamarin projects (e.g, MonoAndroid12.0). But as that's out of support already, I think a break here makes sense.

https://dotnet.microsoft.com/en-us/platform/support/policy/xamarin#microsoft-support

> Xamarin support ended on May 1, 2024 for all Xamarin SDKs including Xamarin.Forms.

The only way to have a test project that can run and targets netstandard2.0 is https://github.com/microsoft/testfx/issues/3984#issuecomment-2459084580 which I think is uncommon and isn't something we should support.

Youssef1313:
Closes #5645

Related to #1285

### DIFF
--- a/src/Adapter/MSTest.Engine/MSTest.Engine.csproj
+++ b/src/Adapter/MSTest.Engine/MSTest.Engine.csproj
@@ -1,7 +1,7 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFrameworks>$(MicrosoftTestingTargetFrameworks)</TargetFrameworks>
+    <TargetFrameworks>netstandard2.0;$(MicrosoftTestingTargetFrameworks)</TargetFrameworks>
     <RootNamespace>Microsoft.Testing.Framework</RootNamespace>
     <!-- Cancel default license (closed-source) and set license as packaged license file -->
     <PackageLicenseExpressionInternal></PackageLicenseExpressionInternal>
--- a/test/UnitTests/MSTest.Engine.UnitTests/MSTest.Engine.UnitTests.csproj
+++ b/test/UnitTests/MSTest.Engine.UnitTests/MSTest.Engine.UnitTests.csproj
@@ -1,7 +1,7 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFrameworks>$(MicrosoftTestingTargetFrameworks)</TargetFrameworks>
+    <TargetFrameworks>$(MicrosoftTestingTargetFrameworks);net462</TargetFrameworks>
     <RootNamespace>Microsoft.Testing.Framework.UnitTests</RootNamespace>
     <NoMoq>true</NoMoq>
     <NoFluentAssertions>true</NoFluentAssertions>


Romfos:
Make sense also drop it everywhere. Anyway you cannot run tests for NS. You must use platform TFM

Youssef1313:
@Romfos It's too aggressive to drop it from TestFramework IMO. You can have a test utilities library that target netstandard2.0 that you then reference in your test projects. For adapter, the story is different, because adapter isn't intended to be used in test utility libraries and is only intended to be used in actual test projects that shouldn't target netstandard2.0.

test/UnitTests/MSTestAdapter.PlatformServices.UnitTests/Services/FileOperationsTests.cs
```diff
@@ -37,7 +37,7 @@ public void LoadAssemblyShouldLoadAssemblyInCurrentContext()
         _fileOperations.LoadAssembly(filePath, false);
     }
 
-#if !WIN_UI
+#if !WIN_UI && !NETFRAMEWORK
```
Evangelink: Isn't this a change of behavior?

test/UnitTests/MSTest.Engine.UnitTests/MSTest.Engine.UnitTests.csproj
```diff
@@ -1,7 +1,7 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFrameworks>$(MicrosoftTestingTargetFrameworks);net462</TargetFrameworks>
```
Evangelink: Looks like a bug

test/UnitTests/MSTestAdapter.PlatformServices.UnitTests/Services/FileOperationsTests.cs
Youssef1313: The whole file now is actually under `!NETFRAMEWORK`

Youssef1313:
```suggestion
#if !WIN_UI
```

src/Adapter/MSTest.Engine/MSTest.Engine.csproj
```diff
@@ -1,7 +1,7 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFrameworks>netstandard2.0;$(MicrosoftTestingTargetFrameworks)</TargetFrameworks>
+    <TargetFrameworks>$(MicrosoftTestingTargetFrameworks)</TargetFrameworks>
```
Youssef1313:
```suggestion
    <TargetFrameworks>netstandard2.0;$(MicrosoftTestingTargetFrameworks)</TargetFrameworks>
```

test/UnitTests/MSTest.Engine.UnitTests/MSTest.Engine.UnitTests.csproj
```diff
@@ -1,7 +1,7 @@
...
+    <TargetFrameworks>$(MicrosoftTestingTargetFrameworks)</TargetFrameworks>
```
Youssef1313:
```suggestion
    <TargetFrameworks>$(MicrosoftTestingTargetFrameworks);net462</TargetFrameworks>
```

PR #5892, opened 2025-06-27
Support passing arguments to DynamicData methods

### LINKED ISSUE #1510: Add support for passing in arguments for methods used to store dynamic data
## Summary

Hello team, hope you're all well. I'd like to request for a feature to be able to pass in arguments to methods used for storing dynamic data.

## Background and Motivation

I was working on some complicated test scenarios with many variations and found myself having to write methods that are almost identical to each other for generating a list of list of N booleans:

For example:
```
private static IEnumerable<object[]> GenerateBoolListWith3Flags()
{
    bool[] boolArray = new bool[] { true, false };

    var testData = new List<object[]>();

    foreach (var flag0 in boolArray)
    {
        foreach (var flag1 in boolArray)
        {
            foreach (var flag2 in boolArray)
            {
                testData.Add(new object[] { flag0, flag1, flag2 });
            }
        }
    }

    return testData;
}

private static IEnumerable<object[]> GenerateBoolListWith6Flags()
{
    bool[] boolArray = new bool[] { true, false };

    var testData = new List<object[]>();

    foreach (var flag0 in boolArray)
    {
        foreach (var flag1 in boolArray)
        {
            foreach (var flag2 in boolArray)
            {
                foreach (var flag3 in boolArray)
                {
                    foreach (var flag4 in boolArray)
                    {
                        foreach (var flag5 in boolArray)
                        {
                            testData.Add(new object[] { flag0, flag1, flag2, flag3, flag4, flag5 });
                        }
                    }
                }
            }
        }
    }

    return testData;
}
```

Then use it as follows:
```
[TestMethod]
[DynamicData(nameof(GenerateBoolListWith3Flags), DynamicDataSourceType.Method)]
public void Test1(
    bool flag0,
    bool flag1,
    bool flag2)
{
    ...
}

[TestMethod]
[DynamicData(nameof(GenerateBoolListWith6Flags), DynamicDataSourceType.Method)]
public void Test2(
    bool flag0,
    bool flag1,
    bool flag2,
    bool flag3,
    bool flag4,
    bool flag5)
{
    ...
}
```

If I happen to have a bunch of tests with many different numbers of bool parameters for different test variations, then I currently am writing a bunch of GenerateBoolListWith*Flags methods.

I was hoping to refactor and combine the dynamic data source methods into one, but I realized I can only achieve this if I am able to pass in an argument

## Proposed Feature

A new attribute such as `DynamicDataMethodArgumentAttribute`:
```
[TestMethod]
[DynamicData(nameof(GenerateBoolListWithNFlags), DynamicDataSourceType.Method)]
[DynamicDataMethodArgument(3)]
public void Test1(
    bool flag0,
    bool flag1,
    bool flag2)
{
    ...
}

[TestMethod]
[DynamicData(nameof(GenerateBoolListWithNFlags), DynamicDataSourceType.Method)]
[DynamicDataMethodArgument(6)]
public void Test2(
    bool flag0,
    bool flag1,
    bool flag2,
    bool flag3,
    bool flag4,
    bool flag5)
{
    ...
}

private static IEnumerable<object[]> GenerateBoolListWithNFlags(int n)
{
    bool[] boolArray = new bool[] { true, false };

    var testData = new List<object[]>();

    // Generate test data with n elements per row

    return testData;
}
```

I understand this may be a very specific usecase, but I think it may also be useful if we somehow want to change up the dynamic data returned depending on passed arguments.

Please let me know if there are any workarounds, other suggestions, or if this is feasible. Thank you!

## Alternative Designs

n/a

Youssef1313:
Fixes #1510

- [x] Revisit DynamicDataShouldBeValidAnalyzer


test/UnitTests/MSTest.SelfRealExamples.UnitTests/DynamicDataMethodArgumentsTests.cs
```diff
@@ -0,0 +1,22 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+
+namespace MSTest.SelfRealExamples.UnitTests;
+
+[TestClass]
+public class DynamicDataMethodArgumentsTests
```
Evangelink: It'd be great to have a few more tests: different types and multiples arguments. From the code implementation, it's also not clear if `GetData(params object[] args)` would work.

Youssef1313: Extended the test. What I have though is that I don't special case params. Nothing special is really done. If it exists in the method signature and user want to pass something, the user needs to wrap that in an array. I don't have a strong opinion though for introducing difference in behavior between `params object[]` and `object[]`.

test/UnitTests/MSTest.SelfRealExamples.UnitTests/DynamicDataMethodArgumentsTests.cs
```diff
@@ -0,0 +1,34 @@
...
+{
+    [TestMethod]
+    [DynamicData(nameof(GetData1), 4)]
+    public void TestMethod(int a)
+        => Assert.IsInRange(4, 6, a);
+
+    [TestMethod]
+    [DynamicData(nameof(GetData2), [new int[] { 4, 5, 6 }])]
```
drognanar: it might be good to add tests for the multiple parameter case and also for object[] parameter case

test/UnitTests/MSTest.Analyzers.UnitTests/DynamicDataShouldBeValidAnalyzerTests.cs
```diff
@@ -1641,4 +1641,29 @@ public void TestMethod2(char c)
...
+
+    [TestMethod]
+    public async Task WhenDataWithArgument_NoDiagnostic()
+        => await VerifyCS.VerifyAnalyzerAsync(
+            """            
+            using System;
+            using System.Collections.Generic;
+            using Microsoft.VisualStudio.TestTools.UnitTesting;
+            [TestClass]
+            public class MyTestClass
```
drognanar, line 1653: is there a test for a failing scenario? where we do pass in the incorrect number of parameters?

test/UnitTests/MSTest.SelfRealExamples.UnitTests/DynamicDataMethodArgumentsTests.cs
Evangelink: Looking at DataRow tests, we probably want:
- 1 value
- few values
- tuple
- string[]
- object[]

And a case where the data member declares parameters with `params` (whether or not we support it).

It's probably also good if we have an acceptance test asserting logs + test didn't run for invalid input cases (too few, too many, invalid type args).

Youssef1313: @Evangelink Just to be clear, do we want to make it special case `params` (`ParamArrayAttribute`)? Similar to [here](https://github.com/microsoft/testfx/blob/94360dfccacfc549d3a385300dba6ab40ca4736d/src/Adapter/MSTestAdapter.PlatformServices/Execution/TestMethodInfo.cs#L191)? Or just treat it as `object[]`, in that case it's no different from any other parameter? Or probably block the usage of `params` until we make a decision?

Evangelink: Whatever is easier. I am not a big fan of params but I know some people are using it. Given this is a new feature, we are quite free and I am perfectly fine to wait for feedback. On the other hand, if that's "easy" to do, it's probably better to have similarity in all places.

Youssef1313: I'll disallow params for now.

mariam-abdulla (approved):

PR #5893, opened 2025-06-27
projects should inherit defined constants

SimonCropp:
this was one of the blockers when i was working on https://github.com/microsoft/testfx/pull/4368

since those csproj ignored solution level constants

### DIFF
--- a/test/IntegrationTests/TestAssets/TestProject/TestProjectForDiscovery.csproj
+++ b/test/IntegrationTests/TestAssets/TestProject/TestProjectForDiscovery.csproj
@@ -5,7 +5,7 @@
   </PropertyGroup>
 
   <PropertyGroup>
-    <DefineConstants>TRACE</DefineConstants>
+    <DefineConstants>$(DefineConstants);TRACE</DefineConstants>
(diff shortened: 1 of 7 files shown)


test/IntegrationTests/TestAssets/TestProject/TestProjectForDiscovery.csproj
```diff
@@ -5,7 +5,7 @@
   </PropertyGroup>
 
   <PropertyGroup>
-    <DefineConstants>TRACE</DefineConstants>
```
Youssef1313, line 8: I think this can even be completely removed. TRACE is already defined by SDK: https://github.com/dotnet/sdk/blob/53b18f981377748f863fdce2716613b22c2bc53f/src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Sdk.CSharp.props#L22

SimonCropp: @Youssef1313 so u want a new PR that removes it from all the projects?

PR #5904, opened 2025-06-29
check logging level in BridgedTraceLogger

### LINKED ISSUE #5903: MTP logger creates strings even if given log level isn't enabled
https://github.com/microsoft/testfx/blob/82f0d684f2a9cacb6f9101aa1494dd6359a18fc9/src/Adapter/MSTest.TestAdapter/TestingPlatformAdapter/BridgedTraceLogger.cs#L18-L25

Here, we call `string.Format` unconditionally.

SimonCropp:
fixes #5903


Youssef1313 (approved):
Thanks @SimonCropp!

PR #5906, opened 2025-06-29
update polyfill

SimonCropp:
related: https://github.com/microsoft/testfx/discussions/5674

polyfill v9 is stable and passes all tests. the only reason it is a beta is because it targets c#14  and sdk10


Evangelink (commented):
I have just started mirroring process.

Evangelink (approved):

PR #5911, opened 2025-06-29
Avoid keeping instance of process in DotnetMuxerLocator

SimonCropp:
the `_currentProcess` was not being disposed. but rather than make `DotnetMuxerLocator` implement `IDisposable`, i changed it to keep the `_currentProcessFileName` instead of the process


Evangelink (approved):

Youssef1313 (approved):
LGTM

PR #5913, opened 2025-06-29
Add ProcessId to IEnvironment

SimonCropp:
Avoid instantiating a new process to get the current process id

this also fixes 3 instances where the current process instance is not being disposed of


Youssef1313 (approved):
LGTM. Thanks @SimonCropp!

PR #5915, opened 2025-06-29
use some string interpolation

SimonCropp:


Youssef1313 (approved):
There might be a small semantic change where we always produce `\r\n` when it used to be `Environment.NewLine` before, but I'm very fine with that. It's unlikely to affect anything.

For most VSTestBridge changes, these are strings that I was expecting to be localized actually. But we can still just merge as-is.

PR #5917, opened 2025-06-30
Fix display name format method

Evangelink:

### DIFF
--- a/test/IntegrationTests/Microsoft.Testing.Platform.Acceptance.IntegrationTests/MSBuildTests.Test.cs
+++ b/test/IntegrationTests/Microsoft.Testing.Platform.Acceptance.IntegrationTests/MSBuildTests.Test.cs
@@ -9,7 +9,10 @@ public class MSBuildTests_Test : AcceptanceTestBase<NopAssetFixture>
     private const string AssetName = "MSBuildTests";
 
     public static string? FormatBuildMatrixEntry(MethodInfo method, object?[]? data)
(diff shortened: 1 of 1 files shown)


test/IntegrationTests/Microsoft.Testing.Platform.Acceptance.IntegrationTests/MSBuildTests.Test.cs
```diff
@@ -9,7 +9,10 @@ public class MSBuildTests_Test : AcceptanceTestBase<NopAssetFixture>
     private const string AssetName = "MSBuildTests";
 
     public static string? FormatBuildMatrixEntry(MethodInfo method, object?[]? data)
-        => $"{data![0]},{(Equals(TargetFrameworks.All.ToMSBuildTargetFrameworks(), data[1]) ? "multitfm" : data[1])},{data[2]},{((bool)data[3]! ? "Succeeded" : "Failed")}";
```
Evangelink, line 12: As pointed out by @Youssef1313, this might be related to a fix that was recently done for the support of tuples. We need to investigate if that's a proble.=m.

Youssef1313: It could potentially be that we need a fix similar to https://github.com/microsoft/testfx/pull/5013. I need to investigate further.

Issue #5920, opened 2025-06-30
Global runsettings thread apartment might not work if an async method is executed in-between


Youssef1313:
> @Youssef1313  Sorry if I wasn't clear I was refering to my sample with the 3 tests methods, not from the same method:
> ```
> [TestClass]
> [STATestClass]
> public sealed class Test1
> {
>     [TestMethod]
>     public void TestMethod1()
>     {
>         _ = new Window();
>     }
> 
>     [TestMethod]
>     public async Task  TestMethod2()
>     {
>         await Task.Yield();
>     }
> 
>     [TestMethod]
>     public void TestMethod3()
>     {
>         _ = new Window();
>     }
> }
> ``` 

 _Originally posted by @fforjanAVEVA in [#5729](https://github.com/microsoft/testfx/issues/5729#issuecomment-3019658384)_

PR #5922, opened 2025-07-01
Fix STA regression when async method completes asynchronously between sync methods

Youssef1313:
This is one possible fix for #5920, and it's the "easy" fix.

We could go with adding our own synchronization context, which actually will ensure that async sta test methods will also continue on STA, which isn't (and was never) the case.

Maybe we can fix it better in v4 and add our custom synchronization context for both STATestMethodAttribute and the global setting as well?

### DIFF
--- a/test/IntegrationTests/MSTest.Acceptance.IntegrationTests/ThreadingTests.cs
+++ b/test/IntegrationTests/MSTest.Acceptance.IntegrationTests/ThreadingTests.cs
@@ -342,12 +342,17 @@ public class UnitTest1
     public async Task TestMethod2()
     {
         AssertCorrectThreadApartmentState();
+        // Ensure that we continue on a thread pool thread after this await.
         await Task.Yield();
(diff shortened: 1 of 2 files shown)


test/IntegrationTests/MSTest.Acceptance.IntegrationTests/ThreadingTests.cs
```diff
@@ -342,7 +342,7 @@ public void TestMethod1()
     public async Task TestMethod2()
     {
         AssertCorrectThreadApartmentState();
-        await Task.CompletedTask;
+        await Task.Yield();
```
Evangelink, line 346: Might be nice to add a comment to help future readers

Evangelink (approved):

test/IntegrationTests/MSTest.Acceptance.IntegrationTests/ThreadingTests.cs
Youssef1313: Done.

Issue #5762, opened 2025-06-12
MTP v2 Breaking Changes


Youssef1313:
## Planned

- [x] #5733 (Done in #5993)
- [x] https://github.com/microsoft/testfx/pull/5993
- [x] #5614 (NOTE: This is common for MTP and MSTest)
- [x] #5517
- [x] #5342
- [x] #3356
- [x] #3354
- [x] TrxReport should use TestNodeUid as-is if it's a Guid already. (https://github.com/microsoft/testfx/pull/5936)
- [x] #5722
- [x] #3630
- [x] Enable again EnableCodeCoverage which is temporarily being disabled in https://github.com/microsoft/testfx/pull/5995, as well as adding CC as a dependency of MSTest metapackage back, and MSTest.Sdk as well. PR: https://github.com/microsoft/testfx/pull/6343

## Postponed

- [ ] #3473
- [ ] We have two `TestSessionContext` classes. One in `TestHost` namespace and one in `Services`. We should discuss removing the one in TestHost namespace. https://github.com/microsoft/testfx/pull/6344

## To discuss

- [x] https://github.com/microsoft/testfx/issues/4294

## Cancelled

- [x] #5453
- [x] #5532
- [x] #3474
- [x] #5474
- [x] #5697 (**NOTE:** Not needed in v2 if we remove the whole thing in v2 ;))
- [x] #5524

Youssef1313:
This will mostly affect non-VSTestBridge users. As for VSTestBridge, we will be addressing these ourselves in the VSTestBridge package.

cc @thomhurst @bradwilson

bradwilson:
What's the proposed release timeframe? I assume it's aligned with .NET 10...?

Youssef1313:
Not yet known: https://github.com/microsoft/testfx/issues/5614#issuecomment-2984871466
It's unlikely to be *before* .NET 10. But whether it's aligned with .NET 10 or some time after is something we don't know yet.

PR #5936, opened 2025-07-04
Few adjustments for test id

Youssef1313:
Related to #1285 and #5762

### DIFF
--- a/src/Platform/Microsoft.Testing.Extensions.TrxReport/Hashing/XxHashShared.cs
+++ b/src/Platform/Microsoft.Testing.Extensions.TrxReport/Hashing/XxHashShared.cs
@@ -324,7 +324,8 @@ internal static unsafe class XxHashShared
                 Debug.Assert(sourceIndex < source.Length, "Expected sourceIndex to be less than source.Length");
                 Debug.Assert(source.Length - sourceIndex <= InternalBufferLengthBytes, "Expected remaining.Length to be less than or equals InternalBufferLengthBytes");
                 Debug.Assert(state.BufferedCount == 0, "Expected BufferedCount to be zero.");
-                Buffer.MemoryCopy(sourcePtr + sourceIndex, buffer, source.Length - sourceIndex, source.Length);
+                Buffer.MemoryCopy(sourcePtr + sourceIndex, buffer, source.Length - sourceIndex, source.Length - sourceIndex);
+                state.BufferedCount = (uint)(source.Length - sourceIndex);
 #endif
             }
 #pragma warning restore SA1519 // Braces should not be omitted from multi-line child statement
@@ -434,6 +435,7 @@ internal static unsafe class XxHashShared
                 new ReadOnlySpan<byte>(buffer, (int)state.BufferedCount).CopyTo(new Span<byte>(lastStripe + catchupSize, (int)state.BufferedCount));
 #else
                 Buffer.MemoryCopy(buffer + InternalBufferLengthBytes - catchupSize, lastStripe, StripeLengthBytes, catchupSize);
+                Buffer.MemoryCopy(buffer, lastStripe + catchupSize, (int)state.BufferedCount, (int)state.BufferedCount);
 #endif
                 accumulateData = lastStripe;
             }


test/UnitTests/Microsoft.Testing.Extensions.UnitTests/TrxTests.cs
```diff
@@ -527,7 +527,7 @@ public async Task TrxReportEngine_GenerateReportAsync_WithMetadataProperties_Trx
         AssertTrxOutcome(xml, "Completed");
         string trxContent = xml.ToString();
         string trxContentsPattern = @"
-    <UnitTest name=""TestMethod"" storage=""testapppath"" id=""b1e0b10f-442a-7875-e431-96fc1c27316b"" priority=""5"">
+    <UnitTest name=""TestMethod"" storage=""testapppath"" id=""59992374-3255-493d-bb45-b618ee00973a"" priority=""5"">
```
Youssef1313:
```suggestion
    <UnitTest name=""TestMethod"" storage=""testapppath"" id=""eaf49958-7df6-876c-f786-9026367c1ae6"" priority=""5"">
```

test/UnitTests/Microsoft.Testing.Extensions.UnitTests/TrxTests.cs
```diff
@@ -527,7 +527,7 @@ public async Task TrxReportEngine_GenerateReportAsync_WithMetadataProperties_Trx
         AssertTrxOutcome(xml, "Completed");
         string trxContent = xml.ToString();
         string trxContentsPattern = @"
-    <UnitTest name=""TestMethod"" storage=""testapppath"" id=""b1e0b10f-442a-7875-e431-96fc1c27316b"" priority=""5"">
+    <UnitTest name=""TestMethod"" storage=""testapppath"" id=""eaf49958-7df6-876c-f786-9026367c1ae6"" priority=""5"">
```
Youssef1313:
```suggestion
    <UnitTest name=""TestMethod"" storage=""testapppath"" id=""39697f97-07bd-1f42-c69a-32f372f41ef4"" priority=""5"">
```

src/Platform/Microsoft.Testing.Extensions.TrxReport/Hashing/XxHash128.cs
```diff
@@ -0,0 +1,423 @@
...
+        if (BitConverter.IsLittleEndian)
+        {
+            low = BinaryPrimitives.ReverseEndianness(low);
+            high = BinaryPrimitives.ReverseEndianness(high);
+        }
+
+#if NET
+        ref byte dest0 = ref MemoryMarshal.GetReference(destination);
+        Unsafe.WriteUnaligned(ref dest0, high);
+        Unsafe.WriteUnaligned(ref Unsafe.AddByteOffset(ref dest0, new IntPtr(sizeof(ulong))), low);
```
Youssef1313, line 223: This is the original logic from dotnet/runtime.

On TFMs where we don't have these APIs, I'm doing it as:

```csharp
        fixed (byte* dest0 = destination)
        {
            *(ulong*)(void*)dest0 = high;
            *((ulong*)(void*)dest0 + 1) = low;
        }
```

@jkotas Could you please help us verify if the semantics are still always correct here?

src/Platform/Microsoft.Testing.Extensions.TrxReport/Hashing/XxHashShared.cs
```diff
@@ -0,0 +1,790 @@
...
+            else
+            {
+                byte* lastStripe = stackalloc byte[StripeLengthBytes];
+                int catchupSize = StripeLengthBytes - (int)state.BufferedCount;
+
+#if NET
+                new ReadOnlySpan<byte>(buffer + InternalBufferLengthBytes - catchupSize, catchupSize).CopyTo(new Span<byte>(lastStripe, StripeLengthBytes));
+                new ReadOnlySpan<byte>(buffer, (int)state.BufferedCount).CopyTo(new Span<byte>(lastStripe + catchupSize, (int)state.BufferedCount));
+#else
+                Buffer.MemoryCopy(buffer + InternalBufferLengthBytes - catchupSize, lastStripe, StripeLengthBytes, catchupSize);
```
Youssef1313, line 437: Original logic from dotnet/runtime is:

```csharp
                new ReadOnlySpan<byte>(buffer + InternalBufferLengthBytes - catchupSize, catchupSize).CopyTo(new Span<byte>(lastStripe, StripeLengthBytes));
                new ReadOnlySpan<byte>(buffer, (int)state.BufferedCount).CopyTo(new Span<byte>(lastStripe + catchupSize, (int)state.BufferedCount));
```

I hope `Buffer.MemoryCopy(buffer + InternalBufferLengthBytes - catchupSize, lastStripe, StripeLengthBytes, catchupSize);` is the same?

@jkotas Can you confirm please?

src/Platform/Microsoft.Testing.Extensions.TrxReport/Hashing/XxHash128.cs
```diff
@@ -0,0 +1,423 @@
...
+
+    private static Hash128 HashLengthOver240(byte* source, uint length, ulong seed)
+    {
+        Debug.Assert(length > 240, "Length was expected to be greater than 240.");
+
+#if NET
+        fixed (byte* defaultSecret = &MemoryMarshal.GetReference(DefaultSecret))
+#else
+        fixed (byte* defaultSecret = DefaultSecret)
+#endif
```
Youssef1313, line 412: In dotnet/runtime, `DefaultSecret` is always `ReadOnlySpan<byte>`. In this PR, I'm making it `byte[]` where `ReadOnlySpan<byte>` isn't available. Then, I'm using `fixed (byte* defaultSecret = DefaultSecret)` when ROS isn't available for us.

src/Platform/Microsoft.Testing.Extensions.TrxReport/Hashing/XxHashShared.cs
```diff
@@ -0,0 +1,790 @@
...
+    private static T ReadUnaligned<T>(void* source)
+#if NET
+        => Unsafe.ReadUnaligned<T>(source);
+#else
+        // ldarg.0
+        // unaligned. 0x1
+        // ldobj !!T
+        // ret
+#pragma warning disable CS8500 // This takes the address of, gets the size of, or declares a pointer to a managed type
+        => *(T*)source;
```
Youssef1313: @jkotas Please confirm if this is good for `ReadUnaligned` when `Unsafe.ReadUnaligned` isn't available for us.

src/Platform/Microsoft.Testing.Extensions.TrxReport/Hashing/XxHashShared.cs
```diff
@@ -0,0 +1,790 @@
...
+#if NET
+        => Unsafe.WriteUnaligned<T>(destination, value);
+#else
+    // ldarg .0
+    // ldarg .1
+    // unaligned. 0x01
+    // stobj !!T
+    // ret
+#pragma warning disable CS8500 // This takes the address of, gets the size of, or declares a pointer to a managed type
+    => *(T*)destination = value;
```
Youssef1313: Same here @jkotas

src/Adapter/MSTestAdapter.PlatformServices/ObjectModel/UnitTestElement.cs
```diff
@@ -197,10 +198,14 @@ internal TestCase ToTestCase()
     private void SetTestCaseId(TestCase testCase, string testFullName)
         => testCase.Id = GenerateSerializedDataStrategyTestId(testFullName);
 
-    private Guid GenerateSerializedDataStrategyTestId(string testFullName)
+    internal static Guid GuidFromString(string data)
     {
-        var idProvider = new TestIdProvider();
+        byte[] hash = XxHash128.Hash(Encoding.Unicode.GetBytes(data));
+        return new Guid(hash);
```
nohwnd, line 207: This should form a v8 uuid https://www.rfc-editor.org/rfc/rfc9562.html#name-uuid-version-8 and I think it would be reasonable to reserve 4 bits for the version / origin.

src/Adapter/MSTestAdapter.PlatformServices/ObjectModel/UnitTestElement.cs
```diff
@@ -221,19 +226,25 @@ private Guid GenerateSerializedDataStrategyTestId(string testFullName)
             // In case path contains invalid characters.
         }
 
-        idProvider.AppendString(fileNameOrFilePath);
-        idProvider.AppendString(testFullName);
+        // TODO: Pool?
+        StringBuilder builder = new();
```
nohwnd: This implementation seems to be doing different things than the test idProvider / xxhasher. 

Those two "append" string, but the string is not really appended (we don't consume more and more memory), instead we take the current hash and transform it to be the hash + the string, getting a new hash. This is more memory efficient and (based on my measurements in the past) faster, than creating the string, and then hashing it in 1 go.

src/Platform/Microsoft.Testing.Extensions.TrxReport/Hashing/BinaryPrimitives.cs
```diff
@@ -0,0 +1,52 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+#if !NET
+using System.Numerics;
+
+namespace System.Buffers.Binary;
```
nohwnd, line 9: Maybe: change the namespaces of the code we are copying in?

src/Platform/Microsoft.Testing.Extensions.TrxReport/Hashing/XxHash128.cs
```diff
@@ -0,0 +1,423 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+// Based on the XXH128 implementation from https://github.com/Cyan4973/xxHash.
```
jkotas, line 4: Do you need to include this in https://github.com/microsoft/testfx/blob/main/THIRD-PARTY-NOTICES.TXT ?

src/Platform/Microsoft.Testing.Extensions.TrxReport/Hashing/XxHash128.cs
```diff
@@ -0,0 +1,423 @@
...
+
+/// <summary>Provides an implementation of the XXH128 hash algorithm for generating a 128-bit hash.</summary>
+/// <remarks>
+/// For methods that persist the computed numerical hash value as bytes,
+/// the value is written in the Big Endian byte order.
+/// </remarks>
+#if NET
+[SkipLocalsInit]
+#endif
+public sealed unsafe class XxHash128
```
jkotas: Should this be internal so that it does not collide with the official implementation when somebody references this package?

src/Platform/Microsoft.Testing.Extensions.TrxReport/Hashing/XxHash128.cs
jkotas: > On TFMs where we don't have these APIs,

Why can't you reference the https://www.nuget.org/packages/System.IO.Hashing/ package for those TFMs?

src/Adapter/MSTestAdapter.PlatformServices/ObjectModel/UnitTestElement.cs
nohwnd: E.g. you can see it here where I use xxhash64 in vstest to replace sha1 in TestIdProvider. https://github.com/microsoft/vstest/compare/main...nohwnd:vstest:xxhash

src/Platform/Microsoft.Testing.Extensions.TrxReport/Hashing/BinaryPrimitives.cs
nohwnd: https://github.com/microsoft/vstest/blob/main/scripts/update-nuget-frameworks.ps1 for inspiration this is what we do in vstest to grab nuget.frameworks as source so we don't have to depend on the package. It does not update often because the functionality is pretty generic, but might help here if we do want to update.

src/Platform/Microsoft.Testing.Extensions.TrxReport/Hashing/XxHashShared.cs
jkotas: This is not a full fidelity replacement of Unsafe.ReadUnaligned. For example, it can crash for some Ts on Arm32.

It is probably going to work for the targets you care about (what are those?), but you would have to test on all of them to be sure. This depends on undefined behavior.

nohwnd (commented):
I would like to see way more tests, especially some that compare  guids based on real system.io.hashin.xx128hash with our patched version, both for making sure that the hashing function is not broken with the code changes, and for future reference, because same as sha1 hashes, the way the hashes are generated should not change among the stated version in guid, to not break testId consumers.

src/Platform/Microsoft.Testing.Extensions.TrxReport/Hashing/XxHash128.cs
Youssef1313: We want to avoid forcing dependencies on MTP users in most cases.

src/Platform/Microsoft.Testing.Extensions.TrxReport/Hashing/XxHashShared.cs
Youssef1313: > it can crash for some Ts on Arm32

I think the `T`s we use here are ulong and uint only. Is that part of the "some Ts"?
We want to be covering any possible target.

src/Platform/Microsoft.Testing.Extensions.TrxReport/Hashing/XxHash128.cs
Youssef1313: Yes, thanks for catching!

src/Platform/Microsoft.Testing.Extensions.TrxReport/Hashing/XxHashShared.cs
jkotas: > We want to be covering any possible target.

If you want to do that, you should use MemoryCopy. MemoryCopy is guaranteed to handle misaligned memory on every target.

Youssef1313: @jkotas I'm confused. This is `ReadUnaligned`, it doesn't move/copy anything in memory. Was this comment intended to be on the `WriteUnaligned` implementation? Note that we want to really guarantee the same behavior as System.IO.Hashing under any scenario (be AOT, CoreCLR, .NET Framework, all architectures etc).

Youssef1313: Oh, maybe you mean creating a local variable, and do MemoryCopy from original pointer to the local variable? Also, is that different from `Marshal.PtrToStructure`?

jkotas: > Oh, maybe you mean creating a local variable, and do MemoryCopy from original pointer to the local variable? 

Yes.

> Also, is that different from Marshal.PtrToStructure?

Marshal.PtrToStructure would be like 10+x slower than MemoryCopy. Othewise, Marshal.PtrToStructure would be functionally same for primitive numerical cases (not for other Ts though).

Youssef1313: I think we care only about primitives as that's what we use here. I'll try with MemoryCopy.

src/Adapter/MSTestAdapter.PlatformServices/ObjectModel/UnitTestElement.cs
Youssef1313: @nohwnd I'm not following here.

nohwnd: You get "random" bytes and mash them into a guid. That is not what should be done, even though the current implementation in vstest and mstest does that. 

Instead it should form a correct v8 guid that uses the data and embeds our version, so we can recognize in the future by which hashing algorithm we generated that given guid, when replacing it with newer version for example. Or with new way of generating the guid.

src/Platform/Microsoft.Testing.Extensions.TrxReport/Hashing/XxHash128.cs
Youssef1313: Added. Thanks!

src/Platform/Microsoft.Testing.Extensions.TrxReport/Hashing/XxHashShared.cs
jkotas: This should get the same treatment as ReadUnaligned

Youssef1313: @jkotas Thanks. I switched to `Buffer.MemoryCopy(&value, destination, sizeof(T), sizeof(T));`. Does it sound correct?

src/Platform/Microsoft.Testing.Extensions.TrxReport/Hashing/XxHashShared.cs
Youssef1313: I switched to

```csharp
        T t;
        Buffer.MemoryCopy(source, &t, sizeof(T), sizeof(T));
        return t;
```

Hopefully it looks correct?

src/Adapter/MSTestAdapter.PlatformServices/ObjectModel/UnitTestElement.cs
Youssef1313: I wrote a benchmark:

```csharp
using System;
using System.Security.Cryptography;
using System.Text;
using BenchmarkDotNet.Attributes;
using BenchmarkDotNet.Running;
using Microsoft.TestPlatform.AdapterUtilities;

namespace MyBenchmarks;

[MemoryDiagnoser]
public class Hashing
{
    [Benchmark]
    public byte[] VSTest_SHA1()
    {
        var idProvider = new TestIdProvider();
        idProvider.AppendString("MyProjectName.UnitTests");
        idProvider.AppendString("MyTestNamespace.MyTestClassName.MyTestMethodName");
        return idProvider.GetHash();
    }

    [Benchmark]
    public byte[] SystemIOHashing_OneString()
    {
        var stringBuilder = new StringBuilder();
        stringBuilder.Append("MyProjectName.UnitTests");
        stringBuilder.Append("MyTestNamespace.MyTestClassName.MyTestMethodName");
        var bytes = Encoding.Unicode.GetBytes(stringBuilder.ToString());
        return System.IO.Hashing.XxHash128.Hash(bytes);
    }

    [Benchmark]
    public byte[] SystemIOHashing_Append()
    {
        var hash = new System.IO.Hashing.XxHash128();
        hash.Append(Encoding.Unicode.GetBytes("MyProjectName.UnitTests"));
        hash.Append(Encoding.Unicode.GetBytes("MyTestNamespace.MyTestClassName.MyTestMethodName"));
        return hash.GetCurrentHash();
    }
}

public class Program
{
    public static void Main(string[] args)
    {
        var summary = BenchmarkRunner.Run<Hashing>();
    }
}

```

| Method                    | Mean      | Error     | StdDev     | Median      | Gen0   | Gen1   | Allocated |
|-------------------------- |----------:|----------:|-----------:|------------:|-------:|-------:|----------:|
| VSTest_SHA1               | 992.83 ns | 45.059 ns | 132.858 ns | 1,048.52 ns | 0.0410 | 0.0401 |     520 B |
| SystemIOHashing_OneString | 102.09 ns | 11.037 ns |  32.543 ns |    81.94 ns | 0.0587 |      - |     736 B |
| SystemIOHashing_Append    |  77.58 ns |  1.609 ns |   3.665 ns |    76.19 ns | 0.0637 | 0.0001 |     800 B |

I agree using `Append` of the hashing algorithm is faster, but I'm not sure if it's worth the extra complexity of porting the additional logic of `Append` from dotnet/runtime. I think it will be negligible in real-world.

Youssef1313: For completeness, the benchmark results on .NET Framework is:

| Method                    | Mean       | Error     | StdDev    | Median     | Gen0   | Gen1   | Allocated |
|-------------------------- |-----------:|----------:|----------:|-----------:|-------:|-------:|----------:|
| VSTest_SHA1               | 3,272.4 ns | 252.38 ns | 744.16 ns | 3,628.0 ns | 0.1183 | 0.0591 |     746 B |
| SystemIOHashing_OneString |   327.7 ns |  11.00 ns |  32.25 ns |   337.2 ns | 0.1173 |      - |     738 B |
| SystemIOHashing_Append    |   237.3 ns |   1.40 ns |   1.17 ns |   237.1 ns | 0.1273 |      - |     802 B |

Youssef1313: I added it anyways.

src/Adapter/MSTestAdapter.PlatformServices/ObjectModel/UnitTestElement.cs
Youssef1313: @nohwnd I'm not sure exactly what we should put into the "version" part of the Guid.

src/Platform/Microsoft.Testing.Extensions.TrxReport/Hashing/XxHashShared.cs
jkotas: Is it surprising that you are replacing two copy operations with just one copy operation. Is there a bug?

Youssef1313: Indeed. Let me see if the tests that I just ported from dotnet/runtime will catch this.

Youssef1313: Fixed, and looks like the tests in dotnet/runtime has good coverage to fail by this missing write :)

```csharp
#if NET
                new ReadOnlySpan<byte>(buffer + InternalBufferLengthBytes - catchupSize, catchupSize).CopyTo(new Span<byte>(lastStripe, StripeLengthBytes));
                new ReadOnlySpan<byte>(buffer, (int)state.BufferedCount).CopyTo(new Span<byte>(lastStripe + catchupSize, (int)state.BufferedCount));
#else
                Buffer.MemoryCopy(buffer + InternalBufferLengthBytes - catchupSize, lastStripe, StripeLengthBytes, catchupSize);
                Buffer.MemoryCopy(buffer, lastStripe + catchupSize, (int)state.BufferedCount, (int)state.BufferedCount);
#endif
```

(Hopefully the semantics is always the same between the two, regardless of architecture or OS differences etc)

src/Adapter/MSTestAdapter.PlatformServices/ObjectModel/UnitTestElement.cs
nohwnd: right now that would be 1, then once we make changes to the data that we put into the guid it should be 2, or when we change hash. basically a way for us to map the bits to some table and be able to tell that they are not the same version, so in theory azdo can map them from 1 to the other.

test/UnitTests/MSTestAdapter.PlatformServices.UnitTests/XxHash128Tests.cs
```diff
@@ -0,0 +1,1732 @@
...
+        return new Hash128(high, low);
+    }
+
+    private record struct Hash128(ulong High, ulong Low)
+    {
+        public override readonly string ToString()
+            => $"Hash128 {{ High = 0x{High:x16}, Low = 0x{Low:x16} }}";
+    }
+
+    private static IEnumerable<(ulong HashHigh, ulong HashLow, long Seed, string Ascii)> TestCases()
```
nohwnd, line 188: Beautiful 😊

src/Adapter/MSTestAdapter.PlatformServices/ObjectModel/UnitTestElement.cs
nohwnd: Thanks for testing that!

src/Adapter/MSTestAdapter.PlatformServices/ObjectModel/UnitTestElement.cs
nohwnd: moved to https://github.com/microsoft/testfx/issues/5957

PR #5943, opened 2025-07-07
update Polyfill and use Process.Kill

SimonCropp:


src/Platform/Microsoft.Testing.Platform/Helpers/System/SystemProcess.cs
```diff
@@ -43,13 +43,8 @@ public void WaitForExit()
     public Task WaitForExitAsync()
         => _process.WaitForExitAsync();
 
-#if NETCOREAPP
     public void Kill()
         => _process.Kill(true);
```
Youssef1313, line 47: I find it a bit misleading (IMO) to polyfill in this case as the boolean isn't used. I would prefer to see the semantics I have clearly even if it's additional annoying `#if`s.

SimonCropp: seems that decisions has already been made, given the same misleading experience happens for all consumers of `IProcess`

Youssef1313: Indeed. Maybe it's better idea for code clarity if `IProcess` has `Kill(bool)` overload, and always throw if `true` is provided on .NET Framework, and then callsites has to be adjusted to use `true` or false` depending on the TFM. It's "more code" but more clear IMO.

@Evangelink @MarcoRossignoli Was this considered in initial designs of the interface? Luckily, `IProcess` is internal so we can easily make changes.

Also @nohwnd , what do you think?

Directory.Packages.props
```diff
@@ -54,7 +54,7 @@
     <PackageVersion Include="Microsoft.TestPlatform.TranslationLayer" Version="$(MicrosoftNETTestSdkVersion)" />
     <PackageVersion Include="Microsoft.Win32.Registry" Version="5.0.0" />
     <PackageVersion Include="Microsoft.WindowsAppSDK" Version="1.0.0" />
-    <PackageVersion Include="Polyfill" Version="9.0.0-beta.4" />
+    <PackageVersion Include="Polyfill" Version="9.0.0-beta.7" />
```
Youssef1313: I just started a mirror of latest version. It may take 30mins to complete.

```suggestion
    <PackageVersion Include="Polyfill" Version="9.0.0-beta.9" />
```

SimonCropp: i am releasing 9.0.0-beta.10 in 15 min

src/Platform/Microsoft.Testing.Platform/Helpers/System/SystemProcess.cs
SimonCropp: @Youssef1313 i think i was a bit too dismissive in my previous comment.

 i did put considerable thought in to if i should polyfill `Process.Kill(bool)` to call `Process.Kill()`. the reason i did it is IMO that is the best outcome for a codebase whose intent is for `Process.Kill(bool)` to be called. any cross platform and cross framework codebase that wants to kill a tree would have to do the if-def like this codebase does. 

i did also look at the potential of properly doing a kill tree, and IMO the required code is not something anyone would want to pull in to their codebase, nor something that is feasible to properly polyfill

> Maybe it's better idea for code clarity if IProcess has Kill(bool)

i would prefer `IProcess.KillTree`. 

As for what for what do do in this PR. how is this for a compromise?

```
    // Note that Process.Kill(true) is polyfilled to Process.Kill() in legacy frameworks
    public void KillTree() => _process.Kill(true);
```

It still calls out the redirect, and is less verbose

Evangelink: I don't recall precisely but the idea was to say (this is only internal abstraction not meant to be exposed at all) we always want to kill the tree, it was just not available before.

Youssef1313 (approved):

Evangelink (approved):

PR #5945, opened 2025-07-07
Ensure TestRun id for retries is the same in TRX

Youssef1313:

### DIFF
--- a/src/Platform/Microsoft.Testing.Extensions.TrxReport/TrxReportEngine.cs
+++ b/src/Platform/Microsoft.Testing.Extensions.TrxReport/TrxReportEngine.cs
@@ -148,7 +148,12 @@ internal sealed partial class TrxReportEngine
             // create the xml doc
             var document = new XDocument(new XDeclaration("1.0", "UTF-8", null));
             var testRun = new XElement(_namespaceUri + "TestRun");
-            testRun.SetAttributeValue("id", Guid.NewGuid());
+            if (!Guid.TryParse(_environment.GetEnvironmentVariable(EnvironmentVariableConstants.TESTINGPLATFORM_TRX_TESTRUN_ID), out Guid testRunId))
+            {
+                testRunId = Guid.NewGuid();
+            }
+
+            testRun.SetAttributeValue("id", testRunId);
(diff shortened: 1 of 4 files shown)


src/Platform/Microsoft.Testing.Extensions.TrxReport/TrxReportEngine.cs
```diff
@@ -148,7 +148,12 @@ public TrxReportEngine(IFileSystem fileSystem, ITestApplicationModuleInfo testAp
             // create the xml doc
             var document = new XDocument(new XDeclaration("1.0", "UTF-8", null));
             var testRun = new XElement(_namespaceUri + "TestRun");
-            testRun.SetAttributeValue("id", Guid.NewGuid());
+            if (!Guid.TryParse(_environment.GetEnvironmentVariable(EnvironmentVariableConstants.TESTINGPLATFORM_TRX_TESTRUN_ID), out Guid testRunId))
```
Evangelink, line 151: What's the impact if 2 different (and separate) runs have the same ID? It seems possible that if you run tests on 2 projects from 2 different repos at "the same time" they would pickup each other ID.

Youssef1313: How would they pickup each other ID? I think `SetEnvironmentVariable` will only set the env variable for the current process (the orchestrator), but then the child processes (each test host that's running the current attempt) will inherit the env variable from orchestrator. But a different project cannot see the same value, unless it's just that the `Guid.NewGuid` is producing a collision which is unlikely to happen in practice.

Youssef1313: So far, no tools I know of are doing anything with the `id` attribute on `TestRun` element. For supporting AzDO retry, I think it's best to have the same ID for all attempts so that we can relate the TRX files together.

Evangelink: I don't know what I assumed but you are right!

PR #5953, opened 2025-07-09
use ReadAllTextAsync from polyfill

SimonCropp:

### DIFF
--- a/Directory.Packages.props
+++ b/Directory.Packages.props
@@ -54,7 +54,7 @@
     <PackageVersion Include="Microsoft.TestPlatform.TranslationLayer" Version="$(MicrosoftNETTestSdkVersion)" />
     <PackageVersion Include="Microsoft.Win32.Registry" Version="5.0.0" />
     <PackageVersion Include="Microsoft.WindowsAppSDK" Version="1.0.0" />
-    <PackageVersion Include="Polyfill" Version="9.0.0-beta.10" />
+    <PackageVersion Include="Polyfill" Version="9.0.0-beta.12" />
     <PackageVersion Include="System.Threading.Tasks.Extensions" Version="$(SystemThreadingTasksExtensionsVersion)" />
   </ItemGroup>
   <ItemGroup Label="Test dependencies">


src/Platform/Microsoft.Testing.Platform/Helpers/System/SystemFileSystem.cs
```diff
@@ -17,14 +17,5 @@ internal sealed class SystemFileSystem : IFileSystem
 
     public string ReadAllText(string path) => File.ReadAllText(path);
 
-#if NETCOREAPP
     public Task<string> ReadAllTextAsync(string path) => File.ReadAllTextAsync(path);
```
Youssef1313, line 20: I think the polyfill doesn't use `FileOptions.SequentialScan`, so this potentially introduces performance difference.

SimonCropp: i did look at that. it is recommended for large files. and is slower on non windows for smaller files. 


so thoughts? should polyfill move to sequential?

SimonCropp: or more specifically what kind of perf gains did testfx achieve by using sequential?

Youssef1313: I think we might have been just replicating what `File.ReadAllTextAsync` API does in .NET Core. https://github.com/dotnet/runtime/blob/c1523bbb379371a75ff28a31748186700396ec1a/src/libraries/System.Private.CoreLib/src/System/IO/File.cs#L1077-L1082

So probably it wasn't measured and was used like that from the beginning.

So far, it seems to only be used by VSTestBridge to read runsettings.

If it can really be slower, then maybe we shouldn't even be using `File.ReadAllTextAsync` at all and use the lower-level StreamReader directly.

SimonCropp: @Youssef1313 those are some good points.

i will adjust `FilePolyfill.Read*` to more accurately reflect the .net implementation by using Sequential and detect encoding  https://github.com/SimonCropp/Polyfill/pull/378 

this will be in Polyfill 9.0.0-beta.12

Youssef1313: I started a mirror of beta.12 which will roughly complete in half an our. Then you can update the package version in the PR.

Youssef1313 (approved):

PR #5955, opened 2025-07-09
Simplify IProcess.MainModule

SimonCropp:
in netclassic Process MainModule can still return null https://learn.microsoft.com/en-us/dotnet/api/system.diagnostics.process.mainmodule?view=netframework-4.8.1#property-value

### DIFF
--- a/global.json
+++ b/global.json
@@ -1,6 +1,6 @@
 {
   "tools": {
-    "dotnet": "10.0.100-preview.6.25315.102",
+    "dotnet": "10.0.100-preview.7.25359.101",
     "runtimes": {
       "dotnet": [
         "3.1.32",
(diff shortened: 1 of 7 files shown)


Youssef1313 (approved):
Very nice cleanup!

Thanks @SimonCropp

SimonCropp:
@Youssef1313  thanks. fixed

src/Platform/Microsoft.Testing.Platform/Services/CurrentTestApplicationModuleInfo.cs
```diff
@@ -95,7 +95,7 @@ public string GetProcessPath()
 #else
     {
         using IProcess currentProcess = process.GetCurrentProcess();
-        return currentProcess.MainModule.FileName;
+        return currentProcess.MainModule?.FileName;
```
Youssef1313: Let's respect the `throwOnNull` in this code path as well.

```suggestion
        string processPath = currentProcess.MainModule?.FileName;
        ApplicationStateGuard.Ensure(processPath is not null || !throwOnNull);
        return processPath;
```

Youssef1313 (approved):

Evangelink (approved):

PR #5969, opened 2025-07-10
Remove un used magic bytes in try get dotnet path by architecture

SimonCropp:

### DIFF
--- a/src/Platform/Microsoft.Testing.Platform.MSBuild/Tasks/DotnetMuxerLocator.cs
+++ b/src/Platform/Microsoft.Testing.Platform.MSBuild/Tasks/DotnetMuxerLocator.cs
@@ -407,21 +407,17 @@ internal sealed class DotnetMuxerLocator
             // skip magicBytes by moving forward 4 bytes
             headerReader.Position += 4;
 
-            byte[] cpuInfoBytes = new byte[4];
-#pragma warning disable CA2022 // Avoid inexact read with 'Stream.Read'
-            headerReader.Read(cpuInfoBytes, 0, cpuInfoBytes.Length);
-#pragma warning restore CA2022 // Avoid inexact read with 'Stream.Read'
+            var cpuInfoBytes = new byte[4];
+            headerReader.ReadExactly(cpuInfoBytes, 0, cpuInfoBytes.Length);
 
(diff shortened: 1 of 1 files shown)


src/Platform/Microsoft.Testing.Platform.MSBuild/Tasks/DotnetMuxerLocator.cs
```diff
@@ -401,24 +401,21 @@ public bool TryGetDotnetPathByArchitecture(
         try
         {
             using var headerReader = new FileStream(path, FileMode.Open, FileAccess.Read);
-            byte[] magicBytes = new byte[4];
+
+            // skip magicBytes by moving forward 4 bytes
+            headerReader.Position += 4;
```
MarcoRossignoli, line 406: Maybe it's better check it and throw if it's not what we expect, we should fix it also here https://github.com/microsoft/vstest/blob/ece7f1653908d4eed603ae4cf0122a54d6df6a74/src/Microsoft.TestPlatform.CoreUtilities/Helpers/DotnetHostHelper.cs#L410

@nohwnd I recall you wrote this part do you agree?

nohwnd: I don't think I wrote this code in vstest, think it was you. but I do agree that checking for the correct values would be nice, we also do it in checking the PE header on windows.

probably just by checking all the possible values via OR. so we don't overcomplicate it with architecture and endiannes:

>For big-endian binaries (as in, the architecture uses big endian), magic number for 32-bit code is 0xfeedface while the magic number for 64-bit architectures is 0xfeedfacf. For little-endian binaries, it will be 0xcefaedfe for 32-bit and 0xcffaedfe for 64-bit. These latter two are just the former but with inverted endianness.

https://en.wikipedia.org/wiki/Mach-O#:~:text=The%20Mach%2DO%20header%20tells,used%20as%20a%20redundancy%20check.

nohwnd: we ended up using the magic bytes. https://github.com/microsoft/testfx/issues/6032

PR #5974, opened 2025-07-10
Set Guid version to 1

Youssef1313:

### DIFF
--- a/src/Adapter/MSTestAdapter.PlatformServices/ObjectModel/UnitTestElement.cs
+++ b/src/Adapter/MSTestAdapter.PlatformServices/ObjectModel/UnitTestElement.cs
@@ -207,7 +207,7 @@ internal sealed class UnitTestElement
         return new Guid(hash);
     }
 
-    private Guid GenerateSerializedDataStrategyTestId(string testFullName)
+    private unsafe Guid GenerateSerializedDataStrategyTestId(string testFullName)
     {
         // Below comment is copied over from Test Platform.
         // If source is a file name then just use the filename for the identifier since the file might have moved between
@@ -246,7 +246,21 @@ internal sealed class UnitTestElement
             hash.Append(CloseBracket);
         }
 
-        return new Guid(hash.GetCurrentHash());
+        byte[] hashBytes = hash.GetCurrentHash();
+
+        // https://www.rfc-editor.org/rfc/rfc9562.html#name-uuid-version-8
+        var guid = new Guid(hashBytes);
+        // We get the address of the 6th byte (_c in Guid) which stores the version.
+        // The 4 most significant bits of the value (interpreted as **short**, which is relevant for endianness) are the version.
+        // So we set those 4 MSBs to be 0001.
+        short* addressOfC = (short*)((byte*)Unsafe.AsPointer(ref guid) + 6);
+        *addressOfC = (short)((*addressOfC & 0b0000_1111_1111_1111) | 0b0001_0000_0000_0000);
+
+#if NET9_0_OR_GREATER
+        // Version property is only available on .NET 9 and later.
+        Debug.Assert(guid.Version == 1, "Expected Guid version to be 1");
+#endif
+        return guid;
     }
 
     private string GetDisplayName() => StringEx.IsNullOrWhiteSpace(DisplayName) ? TestMethod.Name : DisplayName;


src/Adapter/MSTestAdapter.PlatformServices/ObjectModel/UnitTestElement.cs
```diff
@@ -246,7 +240,26 @@ private Guid GenerateSerializedDataStrategyTestId(string testFullName)
...
+        // We get the address of the 6th byte (_c in Guid) which stores the version.
+        // The 4 most significant bits of the value (interpreted as **short**, which is relevant for endianness) are the version.
+        // So we set those 4 MSBs to be 0001.
+        short* addressOfC = (short*)((byte*)&guid + 6);
+        *addressOfC = (short)((*addressOfC & 0b0000_1111_1111_1111) | 0b1000_0000_0000_0000);
+
+        // Here we set our "own" version to "1". We are using the custom_a part of the guid to store that.
+        // We reserved 4-bits for that.
+        int* addressOfA = (int*)(byte*)&guid;
+        *addressOfA = (*addressOfA & 0b0000_1111_1111_1111_1111_1111_1111_1111) | 0b0001_0000_0000_0000_0000_0000_0000_0000;
```
Youssef1313: @tannergooding Does this look correct to you please?

Evangelink (approved):
Approving but I'll let you merge once you get the confirmation.

src/Adapter/MSTestAdapter.PlatformServices/ObjectModel/UnitTestElement.cs
```diff
@@ -246,7 +240,26 @@ private Guid GenerateSerializedDataStrategyTestId(string testFullName)
             hash.Append(CloseBracket);
         }
 
-        return new Guid(hash.GetCurrentHash());
+        byte[] hashBytes = hash.GetCurrentHash();
+
+        // https://www.rfc-editor.org/rfc/rfc9562.html#name-uuid-version-8
+        var guid = new Guid(hashBytes);
```
nohwnd: Your comment on the PR says:

In addition, XxHash128 output is big endian, so it's best if we do new Guid(hash, bigEndian: true) instead of new Guid(hash).


```suggestion
        // XxHash128 output is big endian
        var guid = new Guid(hashBytes, bigEndian: true);
```

Youssef1313: `bigEndian` parameter wasn't available everywhere, so I preferred to use the same code across all TFMs.

nohwnd (changes requested):
Will rewrite in a way that touches single bytes.

nohwnd:
- [ ] Still missing: setting the variant to `0b10` https://www.rfc-editor.org/rfc/rfc9562.html#name-uuid-version-8

src/Adapter/MSTestAdapter.PlatformServices/ObjectModel/UnitTestElement.cs
```diff
@@ -246,7 +241,32 @@ private Guid GenerateSerializedDataStrategyTestId(string testFullName)
...
-        return new Guid(hash.GetCurrentHash());
+        byte[] hashBytes = hash.GetCurrentHash();
+
+        return VersionedGuidFromHash(hashBytes, hashVersion);
+    }
+
+    internal /* for testing */ static Guid VersionedGuidFromHash(byte[] hashBytes, byte hashVersion)
+    {
+        const int firstByte = 0;
+        const int versionByte = 6;
```
Youssef1313: Shouldn't this be `7`?

src/Adapter/MSTestAdapter.PlatformServices/ObjectModel/UnitTestElement.cs
```diff
@@ -246,7 +241,32 @@ private Guid GenerateSerializedDataStrategyTestId(string testFullName)
...
 
-        return new Guid(hash.GetCurrentHash());
+        byte[] hashBytes = hash.GetCurrentHash();
+
+        return VersionedGuidFromHash(hashBytes, hashVersion);
+    }
+
+    internal /* for testing */ static Guid VersionedGuidFromHash(byte[] hashBytes, byte hashVersion)
+    {
+        const int firstByte = 0;
```
Youssef1313: I would go with `3` so that when you do the `ToString()`, it's the first hex character that is representing our own version.

src/Adapter/MSTestAdapter.PlatformServices/ObjectModel/UnitTestElement.cs
nohwnd: Corrected for that in code.

src/Adapter/MSTestAdapter.PlatformServices/ObjectModel/UnitTestElement.cs
nohwnd: This depends on whether we are on big endian system or not, detailed in comments in code.

src/Adapter/MSTestAdapter.PlatformServices/ObjectModel/UnitTestElement.cs
nohwnd: this also ends up depending on whether or not we are on big endian system, detailed in code.

src/Adapter/MSTestAdapter.PlatformServices/ObjectModel/UnitTestElement.cs
```diff
@@ -246,7 +241,55 @@ private Guid GenerateSerializedDataStrategyTestId(string testFullName)
...
+        // We set the first 2 bits of nineth byte to 0b10. In the guid this is stored as byte, so we don't care about endiannes.
+        int variantByte = 8;
+        hashBytes[variantByte] = (byte)((hashBytes[variantByte] & 0b0011_1111) | 0b1000_0000);
+
+#if !NET8_0_OR_GREATER
+
+        if (!BitConverter.IsLittleEndian)
+        {
+            // On .NET Framework and UWP this always returns true, so unless we add more runtimes we should never reach this place.
+            ApplicationStateGuard.Unreachable();
```
Youssef1313: This is kinda scary.

Could we go with our own `Guid` struct that follows the same layout, and we add the `OurOwnGuid(byte[], bool)` constructor, then we unsafe cast it?

Youssef1313: We can even create our own struct as `VersionedGuid` and handle that part in its constructor. That might be much easier?

nohwnd: Why is this scary, the only possibility for this is that we add more runtimes that are not .NET8 or newer, and this will blow up. The code is built just for netframework, uwp and net8+.

This check can also be handled by the binary math logic below, just in different order if you prefer that, but it probably won't be ever used.

nohwnd: Image
 This for non-little endian systems that are not net8 or newer.

Youssef1313: What about running .NET Framework using Mono on non-Windows machine (even if it's discouraged). Do we know for sure in this scenario that this is still unreachable?

nohwnd: image

In that case let's do it like this?

Youssef1313: Fine for me. But wouldn't creating our own struct simplify all those bit operations?

nohwnd: It is almost the same code (but with more assignments), and the struct should validate that the array is not null and has the right lenght, like Guid does. 

But if that is how you want to do it I can do it.

src/Adapter/MSTestAdapter.PlatformServices/ObjectModel/UnitTestElement.cs
```diff
@@ -246,7 +241,55 @@ private Guid GenerateSerializedDataStrategyTestId(string testFullName)
...
+                (short)((hashBytes[4] << 8) | hashBytes[5]),
+                (short)((hashBytes[6] << 8) | hashBytes[7]),
+                hashBytes[8], hashBytes[9], hashBytes[10], hashBytes[11], hashBytes[12], hashBytes[13], hashBytes[14], hashBytes[15]);
+        }
+        else
+        {
+            guid = new Guid(
+                (hashBytes[3] << 24) | (hashBytes[2] << 16) | (hashBytes[1] << 8) | hashBytes[0],
+                (short)((hashBytes[5] << 8) | hashBytes[4]),
+                (short)((hashBytes[7] << 8) | hashBytes[6]),
```
Youssef1313: `Guid.Version` implementation is:

```
public int Version => (ushort)_c >>> 12;
```

- In case of little endian, `_c` is `(hashBytes[6] << 8) | hashBytes[7]`
- In case of big endian, `_c` is `(hashBytes[7] << 8) | hashBytes[6]`

`_c >>> 12` is always the 4 most significant bits, and we added the version (`8`) in the 4 most significant bits of `hashBytes[6]`.

Something feels off here on big endian?

Youssef1313: Or I could easily be missing something too :)

nohwnd: I am trying to setup an emulator to be able to try it :D better than guessing.

nohwnd: Yes you were right my thinking was wrong.

src/Adapter/MSTestAdapter.PlatformServices/ObjectModel/UnitTestElement.cs
```diff
@@ -277,7 +277,8 @@ private Guid GenerateSerializedDataStrategyTestId(string testFullName)
 
 #if NET9_0_OR_GREATER
         // Version property is only available on .NET 9 and later.
-        Debug.Assert(guid.Version == 8, "Expected Guid version to be 8");
+        Debug.Assert(guid.Version == 8, $"Expected Guid version to be 8, but it was {guid.Version}");
+        Debug.Assert(guid.Variant == 8, $"Expected Guid variant to be 8, but it was {guid.Variant}");
```
Youssef1313: Isn't variant supposed to be 2 (`0b10`)?

Youssef1313: Hmm, docs say:

> This corresponds to the most significant 4 bits of the 8th byte

Let me read the spec again as I'm confused if Variant is two bits or four bits.

Youssef1313: Spec:

> The 2-bit variant field as defined by [Section 4.1](https://www.rfc-editor.org/rfc/rfc9562.html#variant_field), set to 0b10. Occupies bits 64 and 65 of octet 8.

.NET implementation is so weird then. 👀

nohwnd: Yeah, the doc kinda sucks as well if you don't know what you are doing. `The "don't-care" bits are not masked out.` Editing that.

nohwnd: https://github.com/dotnet/runtime/pull/118987

PR #5977, opened 2025-07-10
Add TryGetOptionArgument

SimonCropp:
So usages of TryGetOptionArgumentList are problematic

 * some usages use the first arg and ignore the rest
 * some usages check if there is only one and dont use the first if there is more than one
 * some usages assume there is 1 and will throw if that is not correct

so this PR adds TryGetOptionArgument. which will formalize the usages. it will:

 * return false if there is no arg
 * return false if there is no items in the arg
 * throw if there is more than 1

### DIFF
--- a/src/Platform/Microsoft.Testing.Platform/CommandLine/ICommandLineOptions.cs
+++ b/src/Platform/Microsoft.Testing.Platform/CommandLine/ICommandLineOptions.cs
@@ -28,7 +28,7 @@ internal static class CommandLineOptionsExtensions
 {
     public static bool TryGetOptionArgument(this ICommandLineOptions options, string optionName, [NotNullWhen(true)] out string? argument)
     {
-        if (!options.TryGetOptionArgumentList(optionName, out var arguments) ||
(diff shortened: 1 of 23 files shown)


src/Platform/Microsoft.Testing.Platform/CommandLine/ICommandLineOptions.cs
```diff
@@ -23,3 +23,24 @@ public interface ICommandLineOptions
...
+internal static class CommandLineOptionsExtensions
+{
+    public static bool TryGetOptionArgument(this ICommandLineOptions options, string optionName, [NotNullWhen(true)] out string? argument)
+    {
+        if (!options.TryGetOptionArgumentList(optionName, out string[]? arguments) ||
+            arguments.Length == 0)
+        {
+            argument = null;
+            return false;
+        }
```
Youssef1313, line 36: I would `return false` only if `TryGetOptionArgumentList` returned false. Otherwise, throw if `arguments.Length != 1`, and I would probably throw Unreachable, since validation should have happened early and by the point we are calling this, we should be expecting us to be in a valid state.

PR #5984, opened 2025-07-11
suppress TPEXP warning for solution

SimonCropp:


Evangelink (commented):
Makes sense to me, wdyt @Youssef1313?

Youssef1313:
Only downside for me is that we don't clearly see places where we use our experimental APIs, but that is fine for me.

Youssef1313 (approved):

PR #5985, opened 2025-07-11
Fix AppDomain base directory regression in v4

Youssef1313:
Both MSTest and VSTest use DisableAppDomain under RunConfiguration.

- VSTest behavior:
    - DisableAppDomain default is false.
    - The behavior of "false" actually does **not** create app domain, then the "default" app domain will have its directory set to the same directory as `testhost.netXX.exe` which is from VS installation or SDK installation (depends on where you are running).
    - The behavior of "true" **does** create a `TestHostAppDomain` app domain with the expected directory.
    - Relevant parts of VSTest implementation:
        - https://github.com/microsoft/vstest/blob/b9c33d9d01f64a4c63b6dcc28c5fd6835fb99ef0/src/Microsoft.TestPlatform.TestHostProvider/Hosting/DefaultTestHostManager.cs#L203-L208
        - https://github.com/microsoft/vstest/blob/b9c33d9d01f64a4c63b6dcc28c5fd6835fb99ef0/src/testhost.x86/Program.cs#L71-L86
- MSTest behavior:
    - DisableAppDomain default was false in v3, but is now true.
    - The behavior of "false" is that we create our own app domain, with the right base directory.
    - The behavior of "true" is that we do nothing, following what VSTest hands to us.

The bug manifests when VSTest treats DisableAppDomain as "false" (thus, we get unexpected base directory), and MSTest treats DisableAppDomain as "true" (thus, we use what we get from VSTest).

I confirmed the test without the fix fails:

>    Assert.AreEqual failed. Expected:<D:\a\_work\1\s\artifacts\tmp\Debug\testsuite\gwlME\AppDomainTests\bin\Debug\net462>. Actual:<D:\a\_work\1\s\.dotnet\sdk\10.0.100-preview.7.25359.101\TestHostNetFramework\>.

### DIFF
--- a/src/Adapter/MSTest.TestAdapter/VSTestAdapter/AppDomainEngineInvoker.cs
+++ b/src/Adapter/MSTest.TestAdapter/VSTestAdapter/AppDomainEngineInvoker.cs
@@ -22,7 +22,6 @@ internal sealed class AppDomainEngineInvoker : IDisposable
     public AppDomainEngineInvoker(string testSourcePath)
         => _appDomain = CreateNewAppDomain(testSourcePath);
 
-
     /// <summary>
     /// Invokes the Engine.
     /// </summary>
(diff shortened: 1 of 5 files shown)


Youssef1313:
The issue happening here is blocked on https://github.com/microsoft/vstest/issues/15220

Youssef1313:
Thinking more, we might be able to workaround it on our side as well, if it's not possible on vstest side. I think we can simply just replicate vstest behavior of creating an appdomain at the very beginning when app domain is disabled..

PR #5991, opened 2025-07-11
Feat: Assert.That expression

Evangelink:
Add `Assert.That(expression)` as an alternative way to assert expressions. This is providing a unified API that will parse the expression tree to display a useful failure message.

There are some limitations around dynamics, ref etc but this is a good alternative.

Overall this new API si not meant to replace the previous ones but to offer an alternative to assertions.


src/TestFramework/TestFramework/Assertions/Assert.That.cs
```diff
@@ -0,0 +1,786 @@
...
+        }
+
+        if (condition.Compile()())
+        {
+            return;
+        }
+
+        var sb = new StringBuilder();
+        string expressionText = conditionExpression
+            ?? $"() => {CleanExpressionText(condition.Body.ToString())}";
```
Youssef1313: Is the null check here a safe guard for old compiler versions that don't support `CallerArgumentExpression`?

src/TestFramework/TestFramework/Assertions/Assert.That.cs
```diff
@@ -0,0 +1,786 @@
...
+            "UInt32" => "uint",
+            "UInt64" => "ulong",
+            "UInt16" => "ushort",
+            "Single" => "float",
+            "Double" => "double",
+            "Decimal" => "decimal",
+            "Boolean" => "bool",
+            "String" => "string",
+            "Char" => "char",
+            "Object" => "object",
```
Youssef1313, line 582: It feels biased towards C# specifically but fine for me. I'm not sure though if we could be messing up with the unlikely case of users declaring types with the same name as those in BCL but in a different namespace?

src/TestFramework/TestFramework/Assertions/Assert.That.cs
Evangelink: Yes, I wasn't sure if needed or not.

src/TestFramework/TestFramework/Assertions/Assert.That.cs
Evangelink: Definitely biased but I would prefer to have complaints before complexifying codebase with conditions for VB and F# here.

src/TestFramework/TestFramework/Assertions/Assert.That.cs
Youssef1313: I didn't look at the tests yet, but if not tested already, it would be good to explicitly pass null to exercise that code path.

Issue #3356, opened 2024-07-26
Review all exposed extensibility points and substitute the couple `SessionId` and `CancellationToken` with `Microsoft.Testing.Platform.Services.ITestSessionContext`


MarcoRossignoli:
When https://github.com/microsoft/testfx/issues/3355 will be completed we should review the exposed extension and where we need the session id or the cancellation token we should simplify the api using the new exposed `Microsoft.Testing.Platform.Services.ITestSessionContext`

PR #5995, opened 2025-07-12
Use ITestSessionContext instead of SessionUid and CancellationToken

Youssef1313:
Related to #5762

Closes https://github.com/microsoft/testfx/issues/3356

- [ ] We have two `TestSessionContext` classes. One in `TestHost` namespace and one in `Services`. We should discuss removing the one in `TestHost` namespace. I'll do in a follow-up


src/Platform/Microsoft.Testing.Platform/Services/ITestSessionContext.cs
```diff
@@ -5,9 +5,19 @@
...
+/// Interface representing the context of a test session.
+/// It's used in ITestSessionLifetimeHandler to provide information about the session.
+/// </summary>
+public interface ITestSessionContext
 {
-    SessionUid SessionId { get; }
+    /// <summary>
+    /// Gets a unique identifier for the test session.
+    /// </summary>
+    SessionUid SessionId { get; } // TODO: Should this rename to SessionUid?
```
Youssef1313: I'd prefer to address this TODO in this PR and not leave it once we decide.

Evangelink: Makes sense to me, it's clearer.

Cc @MarcoRossignoli

Evangelink (approved):

src/Platform/Microsoft.Testing.Platform/Services/ITestSessionContext.cs
Youssef1313:
```suggestion
    SessionUid SessionUid { get; }
```

Youssef1313: Renamed.

PR #5997, opened 2025-07-12
Add CancellationToken to more MTP APIs

### LINKED ISSUE #5342: [MTP v2] Add CancellationToken parameter to more async methods.
The one I see for now is `IOutputDevice.DisplayAsync`. I think it's worth having CancellationToken parameter there as a breaking change in v2.

Then, internally, `IServerTestHost.PushDataAsync` will have cancellation token parameter as well which will be used in concrete implementation `ServerTestHost.PushDataAsync` instead of the concrete implementation retrieving `ITestApplicationCancellationTokenSource` from service provider.

Youssef1313:
Related to #5762

Closes #5342

### DIFF
--- a/src/Platform/Microsoft.Testing.Platform/Hosts/ServerTestHost.cs
+++ b/src/Platform/Microsoft.Testing.Platform/Hosts/ServerTestHost.cs
@@ -28,7 +28,6 @@ internal sealed partial class ServerTestHost : CommonTestHost, IServerTestHost, 
 
     private readonly IMessageHandlerFactory _messageHandlerFactory;
     private readonly TestFrameworkManager _testFrameworkManager;
-    private readonly ITestApplicationCancellationTokenSource _testApplicationCancellationTokenSource;
     private readonly TestHostManager _testSessionManager;
     private readonly ServerTelemetry _telemetryService;
     private readonly IAsyncMonitor _messageMonitor;
@@ -51,7 +50,6 @@ internal sealed partial class ServerTestHost : CommonTestHost, IServerTestHost, 
     // Whenever a client responds with a result or an error, the requests
     // get completed.
     private ConcurrentDictionary<int, RpcInvocationState> _serverToClientRequests;
-    private int _serverToClientRequestId;
     private IMessageHandler? _messageHandler;
     private TestHost.ClientInfo? _client;
     private IClientInfo? _clientInfoService;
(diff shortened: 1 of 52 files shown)


src/Platform/Microsoft.Testing.Platform/Hosts/ServerTestHost.cs
```diff
@@ -516,7 +513,9 @@ await ExecuteRequestAsync(
             // Verify request cancellation, above the chain the exception will be
             // catch and propagated as correct json rpc error
             perRequestTestSessionContext.CancellationToken.ThrowIfCancellationRequested();
-            await SendTestUpdateCompleteAsync(args.RunId).ConfigureAwait(false);
+
+            // TODO: Should this be ServiceProvider.GetTestApplicationCancellationTokenSource().CancellationToken? Or perRequestTestSessionContext.CancellationToken?
```
Youssef1313: Not sure here.

src/Platform/Microsoft.Testing.Platform/Hosts/ServerTestHost.cs
```diff
@@ -553,7 +552,8 @@ await ExecuteRequestAsync(
                     testNodeUpdateProcessor.GetTestNodeStatistics().TotalDiscoveredTests)
                 : throw new NotImplementedException($"Request not implemented '{method}'");
 
-        await _telemetryService.LogEventAsync(TelemetryEvents.TestsRunEventName, metadata).ConfigureAwait(false);
+        // TODO: Should this be ServiceProvider.GetTestApplicationCancellationTokenSource()? Or perRequestTestSessionContext.CancellationToken?
```
Youssef1313: Not sure here.

src/Platform/Microsoft.Testing.Platform/Hosts/ServerTestHost.cs
MarcoRossignoli: Telemetry and update I think it should be "the global one" the per session only to "cancel single call"

src/Platform/Microsoft.Testing.Platform/Hosts/ServerTestHost.cs
MarcoRossignoli: Telemetry and update I think it should be "the global one" the per session only to "cancel single call"

Evangelink (approved):

PR #6002, opened 2025-07-14
Support `--filter-uid` on console

### LINKED ISSUE #6001: [MTP] Support `--filter-uid`
Today, we have already `TestNodeUidListFilter`, but that's only ever created by ServerTestHost (when running with JsonRPC protocol, e.g, with Test Explorer), and is also used by Retry extension.

I think it's a good idea to expose `--filter-uid`. The idea is to allow tools to arbitrarily run specific tests given the uid, without forcing usage of JsonRPC protocol.

Youssef1313:
Fixes #6001

### DIFF
--- a/src/Platform/Microsoft.Testing.Platform/Resources/PlatformResources.resx
+++ b/src/Platform/Microsoft.Testing.Platform/Resources/PlatformResources.resx
@@ -708,7 +708,7 @@ Takes one argument as string in the format &lt;value&gt;[h|m|s] where 'value' is
     <value>The current test framework does not implement 'IGracefulStopTestExecutionCapability' which is required for '--maximum-failed-tests' feature.</value>
   </data>
   <data name="PlatformCommandLineFilterUidOptionDescription" xml:space="preserve">
-    <value>Specifies a list of test node uids to filter.</value>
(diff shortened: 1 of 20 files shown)


src/Platform/Microsoft.Testing.Platform/Resources/PlatformResources.resx
```diff
@@ -707,4 +707,10 @@ Takes one argument as string in the format &lt;value&gt;[h|m|s] where 'value' is
   <data name="AbortForMaxFailedTestsCapabilityNotAvailable" xml:space="preserve">
     <value>The current test framework does not implement 'IGracefulStopTestExecutionCapability' which is required for '--maximum-failed-tests' feature.</value>
   </data>
+  <data name="PlatformCommandLineFilterUidOptionDescription" xml:space="preserve">
+    <value>Specifies a list of test node uids to filter.</value>
```
Evangelink: I would reword into one of the following:

- `Provides a list of test node UIDs to filter by.`
- `Defines which test node UIDs should be included.`
- `Specifies which test node UIDs to include in the filter.`

Evangelink (approved):

PR #6003, opened 2025-07-14
Cleanup available Assert API overloads

Evangelink:
Reduce number of available APIs, this will help with adding the `CallerArgumentExpression`

Related to #1285


src/TestFramework/TestFramework/Assertions/Assert.Contains.cs
```diff
@@ -381,32 +253,23 @@ public static void DoesNotContain<T>(T expected, IEnumerable<T> collection, IEqu
     /// <param name="collection">The collection.</param>
     /// <param name="comparer">An equality comparer to compare values.</param>
     /// <param name="message">The message to display when the assertion fails.</param>
-    public static void DoesNotContain<T>(T expected, IEnumerable<T> collection, IEqualityComparer<T> comparer, string? message)
+    public static void DoesNotContain<T>(T expected, IEnumerable<T> collection, IEqualityComparer<T>? comparer = null, string message = "")
```
Youssef1313: This makes it not possible to do `Assert.DoesNotContain(t, arrayOfT, "MyMessage")` without using named parameters. Probably not very ideal.

Maybe we should also add `DoesNotContain<T>(T expected, IEnumerable<T> collection, string message = "")`

test/UnitTests/MSTest.Analyzers.UnitTests/AssertionArgsShouldAvoidConditionalAccessAnalyzerTests.cs
```diff
@@ -750,7 +750,7 @@ public class MyTestClass
                 [TestMethod]
                 public void Compliant()
                 {
-                    Assert.AreEqual(new object(), new object(), new A().S?.Length.ToString());
```
Youssef1313: This test was added because the analyzer incorrectly produced a diagnostic for null-conditional in message. Let's remove the `#nullable enable` instead.

src/TestFramework/TestFramework/Assertions/Assert.StartsWith.cs
```diff
@@ -4,11 +4,10 @@
 namespace Microsoft.VisualStudio.TestTools.UnitTesting;
 
 #pragma warning disable RS0026 // Do not add multiple public overloads with optional parameters
+#pragma warning disable RS0027 // API with optional parameter(s) should have the most parameters amongst its public overloads
 
 public sealed partial class Assert
 {
-    #region StartsWith
```
Youssef1313, line 10: Intentional to remove the regions? It's fine for me but just making sure it's intentional.

Youssef1313 (approved):

src/TestFramework/TestFramework/Assertions/Assert.StartsWith.cs
Evangelink: It is intentional, we have only few APIs so it's easy to read. We may want to add back regions if we have more APIs

PR #6008, opened 2025-07-15
remove un-used filesystem in OrchestrateTestHostExecutionAsync

SimonCropp:

### DIFF
--- a/src/Platform/Microsoft.Testing.Extensions.Retry/RetryOrchestrator.cs
+++ b/src/Platform/Microsoft.Testing.Extensions.Retry/RetryOrchestrator.cs
@@ -133,7 +133,6 @@ internal sealed class RetryOrchestrator : ITestHostOrchestrator, IOutputDeviceDa
 
         List<int> exitCodes = [];
         IOutputDevice outputDevice = _serviceProvider.GetOutputDevice();
-        IFileSystem fileSystem = _serviceProvider.GetFileSystem();
 
         int attemptCount = 0;
         List<string> finalArguments = [];


src/Platform/Microsoft.Testing.Extensions.Retry/RetryOrchestrator.cs
```diff
@@ -133,7 +133,6 @@ public async Task<int> OrchestrateTestHostExecutionAsync()
 
         List<int> exitCodes = [];
         IOutputDevice outputDevice = _serviceProvider.GetOutputDevice();
-        IFileSystem fileSystem = _serviceProvider.GetFileSystem();
```
Youssef1313, line 136: There is `File.Move` call that is done by the orchestrator. We should probably have been using the `fileSystem` instance to do it, instead of calling `File.Move`

SimonCropp: what is the purpose pf that abstraction. it is not used constantly, not used for mocking, and adds non trivial complexity

Youssef1313: @SimonCropp We do use it for mocking in some places

e.g, https://github.com/microsoft/testfx/blob/3d3ad857d537add532b1b0918b6f5547a61192c7/test/UnitTests/Microsoft.Testing.Platform.UnitTests/Telemetry/TelemetryManagerTests.cs#L35

We should be using it constantly. If there are places in MTP where we touch file system without this abstraction, it's likely not intended and is something we will want to fix.

PR #6009, opened 2025-07-15
use escapedClassFullName in AppendAssemblyTestNodeBuilderContent

SimonCropp:
should escapedClassFullName be used in AppendAssemblyTestNodeBuilderContent?

or should the variable just be deleted?


src/Analyzers/MSTest.SourceGeneration/Generators/TestNodesGenerator.cs
```diff
@@ -159,7 +159,7 @@ private static void AppendAssemblyTestNodeBuilderContent(IndentedStringBuilder s
             foreach (TestTypeInfo testClassInfo in namespaceClasses)
             {
                 string escapedClassFullName = TestNodeHelpers.GenerateEscapedName(testClassInfo.FullyQualifiedName);
```
Youssef1313, line 161: Let's remove this variable instead.

SimonCropp: ok. will do a new PR

PR #6011, opened 2025-07-15
fix nullability of TestProgressState.DiscoveredTests

SimonCropp:

### DIFF
--- a/src/Platform/Microsoft.Testing.Platform/OutputDevice/Terminal/TestProgressState.cs
+++ b/src/Platform/Microsoft.Testing.Platform/OutputDevice/Terminal/TestProgressState.cs
@@ -45,7 +45,7 @@ internal sealed class TestProgressState
 
     public long Version { get; internal set; }
 
-    public List<(string? DisplayName, string? UID)> DiscoveredTests { get; internal set; } = [];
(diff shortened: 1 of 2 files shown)


src/Platform/Microsoft.Testing.Platform/OutputDevice/Terminal/TestProgressState.cs
```diff
@@ -45,7 +45,7 @@ public TestProgressState(long id, string assembly, string? targetFramework, stri
 
     public long Version { get; internal set; }
 
-    public List<(string? DisplayName, string? UID)> DiscoveredTests { get; internal set; } = [];
+    public List<(string DisplayName, string UID)> DiscoveredTests { get; internal set; } = [];
```
SimonCropp, line 48: @Youssef1313 i dont think `UID` is used anywhere. should it be?

Youssef1313: I think it's fine to remove it and switch it to `List<string>` instead

Evangelink: Merging as-is, let's do a follow-up

PR #6019, opened 2025-07-15
remove Uid from DiscoveredTests

SimonCropp:

### DIFF
--- a/src/Platform/Microsoft.Testing.Platform/OutputDevice/Terminal/TestProgressState.cs
+++ b/src/Platform/Microsoft.Testing.Platform/OutputDevice/Terminal/TestProgressState.cs
@@ -45,7 +45,7 @@ internal sealed class TestProgressState
 
     public long Version { get; internal set; }
 
-    public List<string> DiscoveredTests { get; internal set; } = [];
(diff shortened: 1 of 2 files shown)


src/Platform/Microsoft.Testing.Platform/OutputDevice/Terminal/TestProgressState.cs
```diff
@@ -45,7 +45,7 @@ public TestProgressState(long id, string assembly, string? targetFramework, stri
 
     public long Version { get; internal set; }
 
-    public List<(string DisplayName, string UID)> DiscoveredTests { get; internal set; } = [];
+    public List<string> DiscoveredTests { get; internal set; } = [];
```
Evangelink: Good as-is, I think it would be ideal to also rename the property to make it explicit it's the display names (e.g. `DiscoveredTestDisplayNames`

SimonCropp: done :)

Evangelink (approved):

PR #6025, opened 2025-07-15
reduce some linq alloc in AppendTestDiscoverySummary

SimonCropp:

### DIFF
--- a/src/Platform/Microsoft.Testing.Platform/OutputDevice/Terminal/TerminalTestReporter.cs
+++ b/src/Platform/Microsoft.Testing.Platform/OutputDevice/Terminal/TerminalTestReporter.cs
@@ -887,7 +887,7 @@ internal sealed partial class TerminalTestReporter : IDisposable
     {
         terminal.AppendLine();
 
-        var assemblies = _assemblies.Select(asm => asm.Value).OrderBy(a => a.Assembly).Where(a => a is not null).ToList();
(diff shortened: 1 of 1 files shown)


src/Platform/Microsoft.Testing.Platform/OutputDevice/Terminal/TerminalTestReporter.cs
```diff
@@ -887,7 +887,7 @@ public void AppendTestDiscoverySummary(ITerminal terminal)
     {
         terminal.AppendLine();
 
-        var assemblies = _assemblies.Select(asm => asm.Value).OrderBy(a => a.Assembly).Where(a => a is not null).ToList();
+        var assemblies = _assemblies.Values.OrderBy(a => a.Assembly).ToList();
```
Youssef1313, line 889: @nohwnd We only ever have one assembly right? I think it might be worth if we clean this up.

Youssef1313 (approved):

src/Platform/Microsoft.Testing.Platform/OutputDevice/Terminal/TerminalTestReporter.cs
nohwnd: yes, there are probably more optimizations that can be done in the terminal logger, e.g. getting rid of the flatException type (if you did not do that already)

PR #6026, opened 2025-07-15
fix some nullability in ShowTelemetryBannerFirstNoticeAsync

SimonCropp:

### DIFF
--- a/src/Platform/Microsoft.Testing.Platform/Telemetry/TelemetryManager.cs
+++ b/src/Platform/Microsoft.Testing.Platform/Telemetry/TelemetryManager.cs
@@ -89,12 +89,10 @@ internal sealed class TelemetryManager : ITelemetryManager, IOutputDeviceDataPro
 
         ITestApplicationModuleInfo testApplicationModuleInfo = serviceProvider.GetTestApplicationModuleInfo();
 #pragma warning disable RS0030 // Do not use banned APIs - There is no easy way to disable it for all members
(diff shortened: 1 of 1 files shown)


src/Platform/Microsoft.Testing.Platform/Telemetry/TelemetryManager.cs
```diff
@@ -89,12 +89,10 @@ private async Task ShowTelemetryBannerFirstNoticeAsync(ServiceProvider servicePr
 
         ITestApplicationModuleInfo testApplicationModuleInfo = serviceProvider.GetTestApplicationModuleInfo();
 #pragma warning disable RS0030 // Do not use banned APIs - There is no easy way to disable it for all members
-        string? directory = environment.GetFolderPath(Environment.SpecialFolder.LocalApplicationData, Environment.SpecialFolderOption.Create);
+        string directory = environment.GetFolderPath(Environment.SpecialFolder.LocalApplicationData, Environment.SpecialFolderOption.Create);
```
Youssef1313, line 92: I think this will be null in some unit tests that mock `IEnvironment`.

Youssef1313: The impacted tests will likely be those in https://github.com/microsoft/testfx/blob/f0f7edc0cc72d7dc1b3364e2961f9b3df095669b/test/UnitTests/Microsoft.Testing.Platform.UnitTests/Telemetry/TelemetryManagerTests.cs

PR #6029, opened 2025-07-15
update Polyfill and simplify SHA256 usage

SimonCropp:
also addresses a missing using in TrxReportEngine


Evangelink (approved):

src/Platform/Microsoft.Testing.Extensions.TrxReport/TrxReportEngine.cs
```diff
@@ -16,6 +16,8 @@
 using Microsoft.Testing.Platform.Messages;
 using Microsoft.Testing.Platform.Services;
 
+using Polyfills;
+
```
Evangelink:
```suggestion
```

PR #6030, opened 2025-07-15
Avoid using records for public APIs

Youssef1313:


src/Platform/Microsoft.Testing.Platform/Messages/TestNodeProperties.cs
```diff
@@ -53,327 +89,559 @@ public sealed record PassedTestNodeStateProperty(string? Explanation = null) : T
...
 /// </summary>
-public sealed record FailedTestNodeStateProperty : TestNodeStateProperty
+public sealed class FailedTestNodeStateProperty : TestNodeStateProperty
 {
     /// <summary>
-    /// Initializes a new instance of the <see cref="FailedTestNodeStateProperty"/> class.
+    /// Initializes a new instance of the <see cref="FailedTestNodeStateProperty"/> class with no explanation or exception.
     /// </summary>
     public FailedTestNodeStateProperty()
-        : base(default(string))
```
Youssef1313, line 74: @CyrusNajmabadi, FYI this was triggered by the analyzer to simplify `default` expressions (IDE0034), and wasn't triggered for records. I'm not sure if it's an analyzer bug, or missing callbacks from the compiler for records. We started getting diagnostics only after switching from record to class

PR #6040, opened 2025-07-15
Use some more Polyfill APIs

SimonCropp:
* Range
 * Index
 * char base overload

and remove the pragmas that previously stopped the IDE from suggesting these features

### DIFF
--- a/src/Adapter/MSTestAdapter.PlatformServices/Execution/TestClassInfo.cs
+++ b/src/Adapter/MSTestAdapter.PlatformServices/Execution/TestClassInfo.cs
@@ -765,9 +765,7 @@ public class TestClassInfo
                 TestFailedException? ex = ExecuteClassCleanup(testContext.Context);
                 if (ex is not null && results.Length > 0)
                 {
-#pragma warning disable IDE0056 // Use index operator
-                    TestResult lastResult = results[results.Length - 1];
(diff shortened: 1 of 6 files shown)


src/Adapter/MSTestAdapter.PlatformServices/Execution/TestClassInfo.cs
```diff
@@ -765,9 +765,7 @@ void DoRun()
                 TestFailedException? ex = ExecuteClassCleanup(testContext.Context);
                 if (ex is not null && results.Length > 0)
                 {
-#pragma warning disable IDE0056 // Use index operator
-                    TestResult lastResult = results[results.Length - 1];
```
Youssef1313, line 769: It's sometimes tricky/risky to use polyfills via IVT in test adapter. We have in the past seen weird cases where the adapter and framework being loaded at runtime are not the same TFM, thus, causing MissingMethodException for polyfills that are used via IVT.

Ideal solution IMO is that we use Polyfill **both** in adapter and framework, instead of relying on it only in framework and flowing to adapter via IVT. This can be achieved with `EmbeddedAttribute`.

Evangelink:
@SimonCropp I am closing the PR for now, let's revisit once the `EmbeddedAttribute` is done so we can ensure we won't risk a break because of our IVTs

PR #6046, opened 2025-07-16
remove un-used producer in IPlatformOutputDevice

SimonCropp:

### DIFF
--- a/src/Platform/Microsoft.Testing.Platform/OutputDevice/IPlatformOutputDevice.cs
+++ b/src/Platform/Microsoft.Testing.Platform/OutputDevice/IPlatformOutputDevice.cs
@@ -2,7 +2,6 @@
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
 using Microsoft.Testing.Platform.Extensions;
-using Microsoft.Testing.Platform.Extensions.OutputDevice;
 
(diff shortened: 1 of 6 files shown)


src/Platform/Microsoft.Testing.Platform/OutputDevice/IPlatformOutputDevice.cs
```diff
@@ -14,7 +13,7 @@ internal interface IPlatformOutputDevice : IExtension
 
     Task DisplayAfterSessionEndRunAsync();
 
-    Task DisplayAsync(IOutputDeviceDataProducer producer, IOutputDeviceData data);
+    Task DisplayAsync(IOutputDeviceData data);
```
Youssef1313, line 16: While `IPlatformOutputDevice` is internal today. I think we will want to expose `IOutputDeviceDataProducer` if we will make it public, even if it's unused in today's scenarios. I recall having some discussion around this topic few months back with @Evangelink and @MarcoRossignoli.

Evangelink: That's correct. It's possible that you would want to have a specific handling based on who pushed the data and not only the kind of data. It's similar to even with `object sender`.

PR #6047, opened 2025-07-16
Mark TestHostControllersTestHost.InternalRunAsync as Unsupported browser

SimonCropp:
so we can remove the repetitive `pragma warning disable CA1416`

### DIFF
--- a/src/Platform/Microsoft.Testing.Platform/Hosts/TestHostControllersTestHost.cs
+++ b/src/Platform/Microsoft.Testing.Platform/Hosts/TestHostControllersTestHost.cs
@@ -60,6 +60,7 @@ internal sealed class TestHostControllersTestHost : CommonTestHost, ITestHost, I
 
     public Task<bool> IsEnabledAsync() => Task.FromResult(false);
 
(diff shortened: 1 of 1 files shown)


src/Platform/Microsoft.Testing.Platform/Hosts/TestHostControllersTestHost.cs
```diff
@@ -60,6 +60,7 @@ public TestHostControllersTestHost(TestHostControllerConfiguration testHostsInfo
 
     public Task<bool> IsEnabledAsync() => Task.FromResult(false);
 
+    [UnsupportedOSPlatform("browser")]
```
Youssef1313: This is likely going to hide the warnings up the chain, because the platform compat analyzer will only see that we call `InternalRunAsync` on `CommonTestHost` and is unlikely to produce warning in that case. The pragma in this case is easier to find so we can easily find out what parts we need to do something about.

Let's see what others think as well.

SimonCropp:
@Youssef1313 done

PR #6050, opened 2025-07-16
suppress MSTESTEXP

SimonCropp:
image


SimonCropp:
i am using rider. but yes it looks like a bug in rider

SimonCropp:
yep logging a bug with jetbrains now. and it is not a blocker for me, so dont merge this

PR #6051, opened 2025-07-16
fix Coxtext typo

SimonCropp:

### DIFF
--- a/test/UnitTests/MSTestAdapter.PlatformServices.UnitTests/Execution/TestClassInfoTests.cs
+++ b/test/UnitTests/MSTestAdapter.PlatformServices.UnitTests/Execution/TestClassInfoTests.cs
@@ -16,9 +16,9 @@ namespace Microsoft.VisualStudio.TestPlatform.MSTestAdapter.UnitTests.Execution;
 
 public class TestClassInfoTests : TestContainer
 {
-    private sealed class TestTestCoxtextImpl : ITestContext
+    private sealed class TestContextImpl : ITestContext
     {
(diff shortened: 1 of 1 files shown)


test/UnitTests/MSTestAdapter.PlatformServices.UnitTests/Execution/TestClassInfoTests.cs
```diff
@@ -16,9 +16,9 @@ namespace Microsoft.VisualStudio.TestPlatform.MSTestAdapter.UnitTests.Execution;
 
 public class TestClassInfoTests : TestContainer
 {
-    private sealed class TestTestCoxtextImpl : ITestContext
+    private sealed class TestContextImpl : ITestContext
```
Youssef1313: I don't think it's a typo. The name is indicating that this instance is an implementation of "TestContext" but is used for testing purposes only.

It's like having `class TestComponent : IComponent`, so you prefix the component being tested with `Test`. Here the component is `TestContext` and we are creating a fake test implementation of it.

SimonCropp:
um. Coxtext?

Youssef1313:
Oh sorry I didn't notice that part. I thought it's only removing the "duplicate" `Test` word

test/UnitTests/MSTestAdapter.PlatformServices.UnitTests/Execution/TestClassInfoTests.cs
Youssef1313: Let's then keep the duplicate "Test" word and fix the "Context" typo.

PR #6097, opened 2025-07-19
Produce test binlogs for official pipeline

Youssef1313:

### DIFF
--- a/azure-pipelines-official.yml
+++ b/azure-pipelines-official.yml
@@ -142,11 +142,11 @@ extends:
 
             - ${{ if eq(parameters.SkipTests, False) }}:
               # -ci is allowing to import some environment variables and some required configurations
-              # -nobl avoids overwriting build binlog with binlog from tests
               - script: Test.cmd
                   -configuration $(_BuildConfig)
                   -ci
                   -nobl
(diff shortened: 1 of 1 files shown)


azure-pipelines-official.yml
```diff
@@ -142,11 +142,11 @@ extends:
 
             - ${{ if eq(parameters.SkipTests, False) }}:
               # -ci is allowing to import some environment variables and some required configurations
-              # -nobl avoids overwriting build binlog with binlog from tests
               - script: Test.cmd
                   -configuration $(_BuildConfig)
                   -ci
                   -nobl
+                  -bl:$(BUILD.SOURCESDIRECTORY)\artifacts\TestResults\$(_BuildConfig)\TestStep.binlog
```
Youssef1313: @ViktorHofer IIRC you mentioned this trick somewhere, but it causes:

```
D:\a\_work\1\s\eng\build.ps1 : Cannot process argument transformation on parameter 'binaryLog'. Cannot convert value 
"System.String" to type "System.Management.Automation.SwitchParameter". Boolean parameters accept only Boolean values 
and numbers, such as $True, $False, 1 or 0.

```

Youssef1313: Let me try with `/bl` instead of `-bl` maybe that will work.

Youssef1313: Worked

Evangelink (approved):

PR #6105, opened 2025-07-21
Attempt to stabilize MTP AbortionTests

### LINKED ISSUE #5785: `AbortWithCTRLPlusC_TestHost_Succeeded` is too flaky
This test is flaky in CI.

Youssef1313:
Fixes #5785

Today's situation:

1. StopPoliciesService listens to cancellation [here](https://github.com/microsoft/testfx/blob/860abb97944315891d960bcf58c5909bc46c47a2/src/Platform/Microsoft.Testing.Platform/Services/StopPoliciesService.cs#L22). This registration happens first. It happens during `BuildAsync` (we are still building the test host). As part of this abort, we call TerminalTestReporter.StartCancelling.
2. `ExecuteRequestAsync` uses cancellation token (in the case of this specific test, it's passed to Task.Delay, which will internally also register for cancellation notification)
3. In .NET, cancellation callbacks are run in the reverse order of registration (likely an undocumented implementation detail that no one should rely on). But that means `Task.Delay` is first notified about the cancellation.
4. Now, Ctrl+C signal is received (we receive it on a background thread). `Task.Delay` receives cancellation (on main thread in most cases), then `StartCancelling` is called (on background thread).
5. These two threads are now racing. The `await Task.Delay` will throw OperationCanceledException that we catch and causes us to start writing the test summary. If we started writing the test summary before the background thread calls `StartCancelling`, the test fails.

### DIFF
--- a/src/Platform/Microsoft.Testing.Platform/Hosts/CommonTestHost.cs
+++ b/src/Platform/Microsoft.Testing.Platform/Hosts/CommonTestHost.cs
@@ -148,6 +148,14 @@ internal abstract class CommonHost(ServiceProvider serviceProvider) : IHost
             // Do nothing we're canceled
         }
 
+        if (testSessionCancellationToken.IsCancellationRequested &&
+            serviceProvider.GetService<IStopPoliciesService>() is { } stopPoliciesService)
+        {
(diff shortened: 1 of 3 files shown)


Youssef1313:
Will open an issue to track the design of abort callbacks and will open another PR as a quicker fix

PR #6117, opened 2025-07-21
ConditionBaseAttribute.ShouldRun -> IsConditionMet

### LINKED ISSUE #6086: Consider renaming `ConditionBaseAttribute.ShouldRun` to `IsConditionMet`/`IsConditionSatisfied`
The current name `ShouldRun` may indicate that the implementation should take into account `ConditionMode` (`Include` vs `Exclude`). But in fact, the implementation shouldn't check the `ConditionMode` as the adapter itself inverts the `ShouldRun` based on `ConditionMode`.

The API seems to be confusing on its current form.

Youssef1313:
Related to #1285

Closes #6086


Evangelink (commented):
As discussed offline, it would probably be a good opportunity to transform the property into an async method so we avoid the need for a breaking change like we did for test method.

Evangelink (approved):

PR #4387, opened 2024-12-18
Allow AssemblyCleanup/ClassCleanup to have TestContext parameter

Youssef1313:
Fixes #4233

PR #6126, opened 2025-07-22
Add analyzer to flow TestContext.CTS.Token

### LINKED ISSUE #5878: Add analyzer to flow TestContext.CancellationToken in tests

Youssef1313:
Fixes #5878
Replaces #5941

- Codefix doesn't handle correctly if a TestContext property or parameter needs to be added. This will require a custom fix all provider.


Youssef1313:
> > We do report. Both can optionally accept a TestContext.
> 
> Have we done this update???? I thought we discussed about it but we didn't do it. That's awesome!

https://github.com/microsoft/testfx/pull/4387

PR #6131, opened 2025-07-22
Move some MTP properties from TestNode to TestNodeUpdateMessage

### LINKED ISSUE #5524: [MTP v2][Breaking] Revise MTP properties and where they do belong (test node vs update message)
Many (almost all?) properties are expected to be added today to TestNode, while some of them feel to me they should belong to TestNodeUpdateMessage instead. Examples:

- [x] TimingProperty
    - This doesn't feel like a property of TestNode itself, but rather a success/fail/error/timeout/cancel update message.
    - Fixed in #6124
- [ ] TestNodeStateProperty (and derived)
    - The name gives impression that it's property of the test node, but I also tend to think it belongs to TestNodeUpdateMessage. Take the case of folded tests in MSTest where we will report multiple results for the same TestNode. To me that's an indicator that the property belongs to the update message not the node itself.
- [x] StandardOutputProperty/StandardErrorProperty
    - Fixed in #6124
- [x] FileArtifactProperty
    - Fixed in #6124

On the other hand, the following properties are present on TestNode today and are correct:

- TestFileLocationProperty
- TestMethodIdentifierProperty
- TestMetadataProperty

In addition, we should also revise Trx-specific properties:

https://github.com/microsoft/testfx/blob/0e87d5aafb41692309b310704c6fa858b146919c/src/Platform/Microsoft.Testing.Extensions.TrxReport.Abstractions/TrxReportProperties.cs

- [ ] `TrxExceptionProperty`, likely needs to be moved to `TestNodeUpdateMessage`
- [x] `TrxFullyQualifiedTypeNameProperty` good on `TestNode`.
- [ ] `TrxMessagesProperty`, likely needs to be moved to `TestNodeUpdateMessage`
- [x] `TrxCategoriesProperty` good on `TestNode`

Youssef1313:
Closes #5524

Related to #5524

### DIFF
--- a/test/IntegrationTests/Microsoft.Testing.Platform.Acceptance.IntegrationTests/MSBuildTests.GenerateEntryPoint.cs
+++ b/test/IntegrationTests/Microsoft.Testing.Platform.Acceptance.IntegrationTests/MSBuildTests.GenerateEntryPoint.cs
@@ -226,8 +226,10 @@ internal sealed class DummyTestFramework : ITestFramework, IDataProducer
 
     public async Task ExecuteRequestAsync(ExecuteRequestContext context)
     {
-        await context.MessageBus.PublishAsync(this, new TestNodeUpdateMessage(context.Request.Session.SessionUid,
-            new TestNode() { Uid = "1", DisplayName = "DummyTest", Properties = new(PassedTestNodeStateProperty.CachedInstance) }));
+        var message = new TestNodeUpdateMessage(context.Request.Session.SessionUid,
+            new TestNode() { Uid = "1", DisplayName = "DummyTest" });
+        message.Properties.Add(PassedTestNodeStateProperty.CachedInstance);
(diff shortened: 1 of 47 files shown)


Evangelink (approved):

test/IntegrationTests/Microsoft.Testing.Platform.Acceptance.IntegrationTests/MSBuildTests.GenerateEntryPoint.cs
```diff
@@ -441,6 +444,7 @@ member _.IsEnabledAsync() = Task.FromResult true
                 dataProducer,
                 TestNodeUpdateMessage(
                     context.Request.Session.SessionUid,
+                    // TODO!
```
Youssef1313: Reminder to self.

PR #6132, opened 2025-07-22
Add GlobalTestInitializeAttribute and GlobalTestCleanupAttribute

### LINKED ISSUE #2673: Allow before/after TestMethod code with registering global callback
the semantic should be something similar as below:

```
public class BeforeAfterTestMethodGlobalHook
{
    [GlobalBeforeTestMethod]
    public async Task BeforeTestMethodAsync(TestContext testContext)
    {
        // ...
    }

    [GlobalAfterTestMethod]
    public async Task AfterTestMethodAsync(TestContext testContext)
    {
        // ...
    }
}

### PR DESCRIPTION
**Pull request by Youssef1313**
Fixes #2673

TODOs:

- [x] `TestClassShouldHaveTestMethodAnalyzer` shouldn't report a diagnostic if a static test class only has method attributed with GlobalTestInitializeAttribute/GlobalTestCleanupAttribute.
- [x] Update `UnusedParameterSuppressor`
- [x] Update `UseAsyncSuffixTestFixtureMethodSuppressor`
- [x] Add new analyzer to ensure correct usage/signature of methods attributed with the new attributes

### CODE REVIEW

File: test/IntegrationTests/MSTest.Acceptance.IntegrationTests/LifecycleTests.cs [hunk 1]
```diff
@@ -0,0 +1,249 @@
...
+        //     - ClassInitialize
+        //     - foreach test:
+        //         - ctor
+        //         - TestContext property setter
+        //         - BeforeEveryTestMethod
+        //         - TestInitialize
+        //         - TestMethod
+        //         - TestCleanup
+        //         - AfterEveryTestMethod
+        //         - Dispose
```
Youssef1313: I'm not sure if this is the best order. Or if it's better to have it like:

- BeforeEveryTestMethod
- ctor
- TestContext property setter
- TestInitialize
- TestMethod
- TestCleanup
- Dispose
- AfterEveryTestMethod

Evangelink: It's probably better as the callback are static and cannot access the "instance" state they might have more value  there.

One note, if we give access to the test context somehow (static instance of flowing as parameter) it's possible to have access to some of the "state" of the test in which case it might be weird to be outside.

src/Adapter/MSTestAdapter.PlatformServices/Execution/TestMethodInfo.cs
```diff
@@ -394,6 +394,11 @@ private async Task<TestResult> ExecuteInternalAsync(object?[]? arguments, Cancel
         {
             try
             {
+                foreach ((MethodInfo method, TimeoutInfo? timeoutInfo) in Parent.Parent.GlobalTestInitializations)
```
Evangelink, line 399: It's documented by the tests but I would make it also explicit in the code that we want to run the global init/cleanup before and after everything. It's not like a regular init/cleanup.

src/TestFramework/TestFramework/PublicAPI/PublicAPI.Unshipped.txt
```diff
@@ -1,12 +1,12 @@
...
 Microsoft.VisualStudio.TestTools.UnitTesting.Assert.AssertNonStrictThrowsInterpolatedStringHandler<TException>.AssertNonStrictThrowsInterpolatedStringHandler(int literalLength, int formattedCount, System.Func<object?>! action, out bool shouldAppend) -> void
-Microsoft.VisualStudio.TestTools.UnitTesting.BeforeEveryTestMethodAttribute
-Microsoft.VisualStudio.TestTools.UnitTesting.BeforeEveryTestMethodAttribute.BeforeEveryTestMethodAttribute() -> void
 Microsoft.VisualStudio.TestTools.UnitTesting.CIConditionAttribute
 Microsoft.VisualStudio.TestTools.UnitTesting.CIConditionAttribute.CIConditionAttribute(Microsoft.VisualStudio.TestTools.UnitTesting.ConditionMode mode) -> void
 Microsoft.VisualStudio.TestTools.UnitTesting.ConditionBaseAttribute.Mode.get -> Microsoft.VisualStudio.TestTools.UnitTesting.ConditionMode
+Microsoft.VisualStudio.TestTools.UnitTesting.GlobalTestCleanupAttribute
+Microsoft.VisualStudio.TestTools.UnitTesting.GlobalTestCleanupAttribute.GlobalTestCleanupAttribute() -> void
+Microsoft.VisualStudio.TestTools.UnitTesting.GlobalTestInitializeAttribute
+Microsoft.VisualStudio.TestTools.UnitTesting.GlobalTestInitializeAttribute.GlobalTestInitializeAttribute() -> void
```
Evangelink, line 9: I think these methods need to have a test context instance, otherwise how come a user access it?

Evangelink (commented):
Please also add some acceptance test

src/TestFramework/TestFramework/PublicAPI/PublicAPI.Unshipped.txt
Youssef1313: @Evangelink This is the attribute constructor. The method on which the attribute is applied is passed a TestContext.

Youssef1313:
> Please also add some acceptance test

for which scenario?

src/TestFramework/TestFramework/PublicAPI/PublicAPI.Unshipped.txt
Evangelink: Ofc! Sorry let me redo a proper review, not sure what I have been thinking about.

Evangelink:
> > Please also add some acceptance test
> 
> for which scenario?

I think my page was not refreshed and so I had no acceptance test showing up. All good as is

Evangelink (approved):

src/Adapter/MSTestAdapter.PlatformServices/Execution/TestMethodInfo.cs
```diff
@@ -394,6 +392,18 @@ private async Task<TestResult> ExecuteInternalAsync(object?[]? arguments, Cancel
         {
             try
             {
+                // We invoke global test initialize methods before creating the test class instance.
+                // We consider the test class constructor as a "local" initialization.
+                // We want to invoke first the global initializations, then local ones, then test method.
+                // After that, we invoke local cleanups (including Dispose) and finally global cleanups at last.
+                foreach ((MethodInfo method, TimeoutInfo? timeoutInfo) in Parent.Parent.GlobalTestInitializations)
+                {
+                    InvokeGlobalInitializeMethod(method, ref executionContext, timeoutInfo, timeoutTokenSource);
```
Youssef1313, line 401: TODO: I need to double check what happens in case of exception in global initialize, how is it impacting the running of the global cleanup etc.

src/Adapter/MSTestAdapter.PlatformServices/Execution/TestMethodInfo.cs
```diff
@@ -394,6 +392,18 @@ private async Task<TestResult> ExecuteInternalAsync(object?[]? arguments, Cancel
...
+                }
+
+                // TODO remove dry violation with TestMethodRunner
+                _classInstance = CreateTestClassInstance(result);
```
Youssef1313: We need to execute this (and SetTestContext as well) on the executionContext produced by `InvokeGlobalInitializeMethod`. The only impact this will have is if the global initialize method has non-cooperative timeout.

src/Adapter/MSTestAdapter.PlatformServices/Execution/TestMethodInfo.cs
```diff
@@ -394,12 +392,47 @@ private async Task<TestResult> ExecuteInternalAsync(object?[]? arguments, Cancel
...
+                // We want to invoke first the global initializations, then local ones, then test method.
+                // After that, we invoke local cleanups (including Dispose) and finally global cleanups at last.
+                foreach ((MethodInfo method, TimeoutInfo? timeoutInfo) in Parent.Parent.GlobalTestInitializations)
+                {
+                    InvokeGlobalInitializeMethod(method, ref executionContext, timeoutInfo, timeoutTokenSource);
+                }
+
+                // TODO remove dry violation with TestMethodRunner
+                bool setTestContextSucessful = false;
+                if (executionContext is null)
```
Evangelink, line 406: Please add a small comment (can be a follow-up) to explain here that the Invoke might change execution context so we can "understand" from here without drilling down.

Evangelink (commented):
Ideally we would also add tests to https://github.com/microsoft/testfx/blob/main/test/IntegrationTests/MSTest.Acceptance.IntegrationTests/TimeoutTests.cs for these attributes. I am fine if you want to merge as-is and follow up.

Issue #6136, opened 2025-07-23
CS8618 (nullability warning) for TestContext is not suppressed when a constructor is explicitly declared


Youssef1313:
```csharp
#nullable enable

[TestClass]
public class C
{
    // public C() { }

    // Normally, the compiler produces a nullability warning here.
    // We have a diagnostic suppressor to suppress it.
    // However, once the constructor above is commented out, the
    // compiler starts producing the warning on the constructor declaration instead.
    // In this case, we fail to suppress.
    public string S { get; set; }

    [TestMethod]
    public void MyTest() { }
}
```

PR #6139, opened 2025-07-23
Fix nullability suppressor for TestContext when explicit constructor is declared

Youssef1313:
Fixes https://github.com/microsoft/testfx/issues/6136

### DIFF
--- a/src/Analyzers/MSTest.Analyzers/NonNullableReferenceNotInitializedSuppressor.cs
+++ b/src/Analyzers/MSTest.Analyzers/NonNullableReferenceNotInitializedSuppressor.cs
@@ -41,19 +41,25 @@ public sealed class NonNullableReferenceNotInitializedSuppressor : DiagnosticSup
 
         foreach (Diagnostic diagnostic in context.ReportedDiagnostics)
         {
-            // The diagnostic is reported on the test method
-            if (diagnostic.Location.SourceTree is not { } tree)
+            // The main diagnostic location isn't always pointing to the TestContext property.
+            // It can point to the constructor.
+            // The additional locations will have the right thing.
+            // Per https://github.com/dotnet/roslyn/issues/79188#issuecomment-3017087900,
+            // it seems like it was an intentional design to include the additional locations specifically for DiagnosticSuppressor scenarios.
+            // So it is safe to use AdditionalLocations here. We are not relying on an implementation detail here.
+            Location location = diagnostic.AdditionalLocations[0];
+            if (location.SourceTree is not { } tree)
             {
                 continue;
             }
 
             SyntaxNode root = tree.GetRoot(context.CancellationToken);
-            SyntaxNode node = root.FindNode(diagnostic.Location.SourceSpan, getInnermostNodeForTie: true);
+            SyntaxNode node = root.FindNode(location.SourceSpan, getInnermostNodeForTie: true);
 
             SemanticModel semanticModel = context.GetSemanticModel(tree);
             ISymbol? declaredSymbol = semanticModel.GetDeclaredSymbol(node, context.CancellationToken);
             if (declaredSymbol is IPropertySymbol property
-                && string.Equals(property.Name, "TestContext", StringComparison.OrdinalIgnoreCase)
+                && string.Equals(property.Name, "TestContext", StringComparison.Ordinal)
                 && SymbolEqualityComparer.Default.Equals(testContextSymbol, property.GetMethod?.ReturnType)
                 && property.ContainingType.GetAttributes().Any(attr => attr.AttributeClass.Inherits(testClassAttributeSymbol)))
             {
--- a/src/Analyzers/MSTest.Analyzers/NonNullableReferenceNotInitializedSuppressor.cs
+++ b/src/Analyzers/MSTest.Analyzers/NonNullableReferenceNotInitializedSuppressor.cs
@@ -44,10 +44,13 @@ public sealed class NonNullableReferenceNotInitializedSuppressor : DiagnosticSup
             // The main diagnostic location isn't always pointing to the TestContext property.
             // It can point to the constructor.
             // The additional locations will have the right thing.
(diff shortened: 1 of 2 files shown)


test/UnitTests/MSTest.Analyzers.UnitTests/NonNullableReferenceNotInitializedSuppressorTests.cs
```diff
@@ -54,56 +42,55 @@ public async Task TestContextPropertyOnNonTestClass_DiagnosticIsNotSuppressed()
...
+        };
 
-            yield return new NonNullableReferenceNotInitializedSuppressor();
+        foreach ((int location, bool isSuppressed) in diagnostics)
+        {
+            test.ExpectedDiagnostics.Add(DiagnosticResult.CompilerError("CS8618")
+                .WithLocation(location)
+                .WithOptions(DiagnosticOptions.IgnoreAdditionalLocations)
+                .WithArguments("property", "TestContext")
+                .WithIsSuppressed(isSuppressed));
```
Youssef1313, line 91: We need to update all other suppressor tests similar to here. I'll track that separately.

On the old form, the tests are not asserting anything, and commented out the `context.ReportSuppression` call will still pass all tests.

src/Analyzers/MSTest.Analyzers/NonNullableReferenceNotInitializedSuppressor.cs
```diff
@@ -41,19 +41,25 @@ public override void ReportSuppressions(SuppressionAnalysisContext context)
...
 
             SyntaxNode root = tree.GetRoot(context.CancellationToken);
-            SyntaxNode node = root.FindNode(diagnostic.Location.SourceSpan, getInnermostNodeForTie: true);
+            SyntaxNode node = root.FindNode(location.SourceSpan, getInnermostNodeForTie: true);
 
             SemanticModel semanticModel = context.GetSemanticModel(tree);
             ISymbol? declaredSymbol = semanticModel.GetDeclaredSymbol(node, context.CancellationToken);
             if (declaredSymbol is IPropertySymbol property
-                && string.Equals(property.Name, "TestContext", StringComparison.OrdinalIgnoreCase)
+                && string.Equals(property.Name, "TestContext", StringComparison.Ordinal)
```
Evangelink, line 65: I think I did ignore case because the property actually supports different casing (but I would need to double check).

Youssef1313: It's case sensitive. I just tried it.

The setter breakpoint isn't hit:

image

And we also produce a warning that it's invalid.

src/Analyzers/MSTest.Analyzers/NonNullableReferenceNotInitializedSuppressor.cs
```diff
@@ -41,19 +41,25 @@ public override void ReportSuppressions(SuppressionAnalysisContext context)
...
         {
-            // The diagnostic is reported on the test method
-            if (diagnostic.Location.SourceTree is not { } tree)
+            // The main diagnostic location isn't always pointing to the TestContext property.
+            // It can point to the constructor.
+            // The additional locations will have the right thing.
+            // Per https://github.com/dotnet/roslyn/issues/79188#issuecomment-3017087900,
+            // it seems like it was an intentional design to include the additional locations specifically for DiagnosticSuppressor scenarios.
+            // So it is safe to use AdditionalLocations here. We are not relying on an implementation detail here.
+            Location location = diagnostic.AdditionalLocations[0];
```
drognanar: is it always guaranteed to have at least one additional location? or can this be empty?

test/UnitTests/MSTest.Analyzers.UnitTests/NonNullableReferenceNotInitializedSuppressorTests.cs
```diff
@@ -54,56 +42,55 @@ public async Task TestContextPropertyOnNonTestClass_DiagnosticIsNotSuppressed()
...
-            context.RegisterSymbolAction(AnalyzeSymbol, SymbolKind.Property);
-        }
+    public TestContext TestContext { get; set; }
+}
+";
 
-        private void AnalyzeSymbol(SymbolAnalysisContext context)
-        {
-        }
+        await VerifySingleSuppressionAsync(code, true);
```
drognanar: nit: use named parameter?

drognanar (dismissed):

src/Analyzers/MSTest.Analyzers/NonNullableReferenceNotInitializedSuppressor.cs
Youssef1313: As explained in the comment above, it seems that Roslyn will always produce it and it's not an implementation detail.

But tagging @RikkiGibson for confirmation as well.

RikkiGibson: See https://github.com/dotnet/roslyn/pull/58096#discussion_r761898594

In my opinion, it would be best to be defensive still, in case we have a bug that causes the additional location to be dropped. For example, you could just adjust the pattern below to be something like `if (diagnostic.AdditionalLocations is not [{ SourceTree: { } tree } location }, ..]) continue;`.

Then, if for some reason Roslyn is accidentally dropping the AdditionalLocation, which has happened before such as in dotnet/roslyn#66037, your suppressor at least will not crash/be completely suspended from running, but will have a bug where the symptom is something that was supposed to be suppressed, isn't getting suppressed. (and, where the fix would be on the Roslyn side.)

Youssef1313: Thanks @RikkiGibson.

I updated the logic to use the first additional location if present, or fallback to the primary diagnostic location otherwise (to at least still work in case of constructor not explicitly declared).

Evangelink (approved):

PR #6171, opened 2025-07-25
Drop stacktrace trimming

### LINKED ISSUE #4266: Revise Regex usage in TrimStackTrace

Evangelink:
Fix #4266


src/Adapter/MSTestAdapter.PlatformServices/Extensions/ExceptionExtensions.cs
```diff
@@ -1,6 +1,8 @@
 ﻿// Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
+using System.Diagnostics;
```
Evangelink:
```suggestion
```

test/IntegrationTests/MSTest.Acceptance.IntegrationTests/GenericTestMethodTests.cs
```diff
@@ -20,58 +20,62 @@ public async Task TestDifferentGenericMethodTestCases()
...
-              Cannot create an instance of T\[] because Type\.ContainsGenericParameters is true\.
-                .+?
-            failed ParameterizedMethodWithNestedGeneric \(System\.Collections\.Generic\.List`1\[System.String],System\.Collections\.Generic\.List`1\[System.String]\) \(\d+ms\)
-              Assert\.Fail failed\. Test method 'ParameterizedMethodWithNestedGeneric' did run with first list \[Hello, World] and second list \[Unit, Testing]
-                .+?
-            failed ParameterizedMethodWithNestedGeneric \(System\.Collections\.Generic\.List`1\[System.Int32],System\.Collections\.Generic\.List`1\[System.Int32]\) \(\d+ms\)
-              Assert\.Fail failed\. Test method 'ParameterizedMethodWithNestedGeneric' did run with first list \[0, 1] and second list \[2, 3]
-                .+?
-            """, RegexOptions.Singleline);
+failed AMethodWithBadConstraints \(0\) \((?:\d+s\s?)?(?:\d+ms)\)
```
Youssef1313: I personally prefer the old indentation

Youssef1313 (approved):

test/IntegrationTests/MSTest.Acceptance.IntegrationTests/GenericTestMethodTests.cs
Evangelink: Moving to the left was helping me to diff where things got wrong, we were losing too much screen space. I can reset now if you prefer.

test/IntegrationTests/MSTest.Acceptance.IntegrationTests/GenericTestMethodTests.cs
```diff
@@ -20,57 +20,61 @@ public async Task TestDifferentGenericMethodTestCases()
         testHostResult.AssertExitCodeIs(ExitCodes.AtLeastOneTestFailed);
         testHostResult.AssertOutputMatchesRegex(
             """
-            failed AMethodWithBadConstraints \(0\) \((\d+s )?\d+ms\)
+            failed AMethodWithBadConstraints \(0\) \((?:\d+s\s?)?(?:\d+ms)\)
```
Youssef1313: Isn't the previous regex more readable? The part of an optional "seconds" component of duration is already fixed. So I guess the changes here are unnecessary.

test/IntegrationTests/MSTest.Acceptance.IntegrationTests/GenericTestMethodTests.cs
```diff
@@ -20,57 +20,61 @@ public async Task TestDifferentGenericMethodTestCases()
...
               GenericArguments\[0], 'System\.Int32', on 'Void AMethodWithBadConstraints\[T]\(T\)' violates the constraint of type 'T'\.
-                .+?
-            failed NonParameterizedTestMethod \((\d+s )?\d+ms\)
+            .+?
```
Youssef1313: There are spaces removed here.

Evangelink: Yes it was causing problems with multi-lines stack traces

test/IntegrationTests/MSTest.Acceptance.IntegrationTests/GenericTestMethodTests.cs
Evangelink: Fine by me, reverting this part. Note that it was done before you fixed something (PR is open for a bit now)

test/IntegrationTests/MSTest.Acceptance.IntegrationTests/GenericTestMethodTests.cs
Youssef1313: How so? We use RegexOptions.SingleLine here which means the `.` will match new lines. The removed whitespaces here was asserting that at least the start of the stack trace is indented correctly.

If it's failing now, that means there is a change in behavior in stack trace indentation, which would be a bug.

Evangelink: Let me undo and see on CI. Pretty sure there was some indeterminism in the stacktrace shown

test/IntegrationTests/MSTest.Acceptance.IntegrationTests/GenericTestMethodTests.cs
```diff
@@ -36,20 +36,24 @@ public async Task TestDifferentGenericMethodTestCases()
               Assert\.Fail failed\. Test method 'ParameterizedMethodSimple' did run with parameter 'Hello world' and type 'System\.String'\.
                 .+?
             failed ParameterizedMethodSimple \(null\) \((\d+s )?\d+ms\)
-              Test method TestClass\.ParameterizedMethodSimple threw exception: 
+              Test method TestClass\.ParameterizedMethodSimple threw exception:
```
Youssef1313, line 39: There is an extra whitespace here at the end (that I wasn't aware it exists!) and it's removed now. It originates from here:

https://github.com/microsoft/testfx/blob/d879f669ec5a48dddc72641b9fbcc3710a7c4061/src/Adapter/MSTestAdapter.PlatformServices/Resources/Resource.resx#L469

So I guess that might cause the test to fail now

Evangelink: Let me fix the resx!

PR #6197, opened 2025-07-28
Move MSTest.Analyzers.UnitTests to net472 instead of net462

Youssef1313:


Evangelink (commented):
LGTM - I would add a comment to say why we cannot test with net462

Evangelink (approved):

PR #6222, opened 2025-07-29
Update public samples to 3.10

Youssef1313:

### DIFF
--- a/samples/public/Directory.Build.props
+++ b/samples/public/Directory.Build.props
@@ -5,8 +5,8 @@
     <CodeCoverageVersion>17.14.2</CodeCoverageVersion>
     <MSTestVersion>3.10.0</MSTestVersion>
     <MSTestAOTVersion>1.0.0-alpha.25379.8</MSTestAOTVersion>
-    <PlaywrightVersion>1.50.0</PlaywrightVersion>
-    <TestingPlatformVersion>1.7.0</TestingPlatformVersion>
+    <PlaywrightVersion>1.54.0</PlaywrightVersion>
+    <TestingPlatformVersion>1.8.0</TestingPlatformVersion>
     <VSTestVersion>17.13.0</VSTestVersion>
     <GenerateDocumentationFile>false</GenerateDocumentationFile>
     <NoWarn>$(NoWarn);SA0001;EnableGenerateDocumentationFile</NoWarn>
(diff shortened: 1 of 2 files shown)


samples/public/Directory.Build.props
```diff
@@ -3,8 +3,8 @@
   <PropertyGroup>
     <AspireTestingVersion>9.0.0</AspireTestingVersion>
     <CodeCoverageVersion>17.14.2</CodeCoverageVersion>
-    <MSTestVersion>3.9.2</MSTestVersion>
-    <MSTestAOTVersion>1.0.0-alpha.25256.6</MSTestAOTVersion>
+    <MSTestVersion>3.10.0</MSTestVersion>
+    <MSTestAOTVersion>1.0.0-alpha.25379.8</MSTestAOTVersion>
     <PlaywrightVersion>1.50.0</PlaywrightVersion>
     <TestingPlatformVersion>1.7.0</TestingPlatformVersion>
```
Evangelink: This needs to be updated too

samples/public/Directory.Build.props
```diff
@@ -3,8 +3,8 @@
   <PropertyGroup>
     <AspireTestingVersion>9.0.0</AspireTestingVersion>
     <CodeCoverageVersion>17.14.2</CodeCoverageVersion>
-    <MSTestVersion>3.9.2</MSTestVersion>
-    <MSTestAOTVersion>1.0.0-alpha.25256.6</MSTestAOTVersion>
+    <MSTestVersion>3.10.0</MSTestVersion>
+    <MSTestAOTVersion>1.0.0-alpha.25379.8</MSTestAOTVersion>
     <PlaywrightVersion>1.50.0</PlaywrightVersion>
```
Evangelink: Have you ensured there is no new versions before releasing?

Youssef1313: My bad, should be 1.54.0 :/

Updating version used by MSTest.Sdk in https://github.com/microsoft/testfx/pull/6224, and will update sample in this PR.

I should have updated the version used by MSTest.Sdk before releasing 3.10 😕

Evangelink: That's fine! I am wondering if we could add some special check when we do final release to ensure there is no new version avialable.

Evangelink (approved):

PR #6233, opened 2025-07-29
Allow to opt-out of Microsoft.NET.Test.Sdk

Evangelink:
When a repo is fully onboarded to MTP and dotnet test, we should give the opportunity for developers to stop relying on VSTEst package. This should not be done by default, but as an opt-in, as some extra tools might be broken.


src/Package/MSTest.Sdk/Sdk/Runner/ClassicEngine.targets
```diff
@@ -48,6 +48,7 @@
   <!-- Core (for test applications - not for test libraries) -->
   <ItemGroup Condition=" '$(IsTestApplication)' == 'true' ">
     <PackageReference Include="Microsoft.NET.Test.Sdk" Sdk="MSTest"
+                      Condition=" '$(ExcludeMicrosoftNetTestSdk)' != 'true' "
```
Evangelink, line 51: Follows naming in arcade but maybe we would prefer to follow the `EnableXXX` pattern we have for the extensions.

Youssef1313 (approved):

PR #6234, opened 2025-07-29
Integration tests are using MSTest

Evangelink:
Increase usage of MSTest where we know there will be no conflict. Also favor using MSTest assertion APIs over AwesomeAssertions as it helps with dogfooding and understanding missing APIs.

### DIFF
--- a/test/IntegrationTests/MSTest.VstestConsoleWrapper.IntegrationTests/MSTest.VstestConsoleWrapper.IntegrationTests.csproj
+++ b/test/IntegrationTests/MSTest.VstestConsoleWrapper.IntegrationTests/MSTest.VstestConsoleWrapper.IntegrationTests.csproj
@@ -2,14 +2,26 @@
 
   <PropertyGroup>
     <TargetFramework>$(NetFrameworkMinimum)</TargetFramework>
-    <UseInternalTestFramework>true</UseInternalTestFramework>
-    <!-- Don't generate MSTest using -->
-    <UseMSTest>false</UseMSTest>
+    <EnableMSTestRunner>true</EnableMSTestRunner>
+    <UseMSTestFromSource>true</UseMSTestFromSource>
     <OutputType>Exe</OutputType>
   </PropertyGroup>
 
   <ItemGroup>
     <ProjectReference Include="$(RepoRoot)test\Utilities\Automation.CLI\Automation.CLI.csproj" />
+    <ProjectReference Include="$(RepoRoot)src\Adapter\MSTest.TestAdapter\MSTest.TestAdapter.csproj" />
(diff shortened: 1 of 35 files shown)


test/Utilities/Automation.CLI/Automation.CLI.csproj
```diff
@@ -1,19 +1,18 @@
...
-    <PackageReference Include="AwesomeAssertions" />
     <PackageReference Include="Microsoft.TestPlatform" />
     <PackageReference Include="Microsoft.TestPlatform.TranslationLayer" />
     <PackageReference Include="Polyfill" PrivateAssets="all" />
   </ItemGroup>
 
   <ItemGroup>
-    <ProjectReference Include="$(RepoRoot)src\TestFramework\TestFramework\TestFramework.csproj" PrivateAssets="all" />
-    <ProjectReference Include="$(RepoRoot)test\Utilities\TestFramework.ForTestingMSTest\TestFramework.ForTestingMSTest.csproj" />
+    <ProjectReference Include="$(RepoRoot)src\Adapter\MSTest.TestAdapter\MSTest.TestAdapter.csproj" />
```
Youssef1313: Why does this need adapter? It looks like it's a utility project?

test/IntegrationTests/MSTest.VstestConsoleWrapper.IntegrationTests/MSTest.VstestConsoleWrapper.IntegrationTests.csproj
```diff
@@ -2,14 +2,26 @@
...
-    <!-- Don't generate MSTest using -->
-    <UseMSTest>false</UseMSTest>
+    <EnableMSTestRunner>true</EnableMSTestRunner>
+    <UseMSTestFromSource>true</UseMSTestFromSource>
     <OutputType>Exe</OutputType>
   </PropertyGroup>
 
   <ItemGroup>
     <ProjectReference Include="$(RepoRoot)test\Utilities\Automation.CLI\Automation.CLI.csproj" />
+    <ProjectReference Include="$(RepoRoot)src\Adapter\MSTest.TestAdapter\MSTest.TestAdapter.csproj" />
```
Youssef1313, line 12: I guess the original reason why we had an internal test framework is that we don't want to use whatever we have from source for testing. Because if it's broken in some way, then the tests may be testing what it should.

Evangelink: Although what you say is a valid concern, I think we have better ways around that (min number of tests, in the past we had a txt file that was listing all tests passing and was checked at the end of the session).

The original reason was around around modification of the states of MSTest internal and that's why I think we don't have much way around for unit tests.

As we discussed in past, the only viable option would be to rely on NuGet packages of the last insertion (or last daily) but we would still need to find a way to isolation classes from MSTest local to MSTest package.

PR #6251, opened 2025-07-30
Fail on warn as error in discovery

### LINKED ISSUE #5992: `TreatDiscoveryWarningsAsErrors` should throw instead of sending message
https://github.com/microsoft/testfx/blob/96f6b1cc6138cd5a3166355b54547bae6df0c2b4/src/Adapter/MSTestAdapter.PlatformServices/Discovery/UnitTestDiscoverer.cs#L71

To guarantee that the error will end up failing the test run, we should throw an exception.

Evangelink:
Replaces #6202 and fixes #5992

### DIFF
--- a/src/TestFramework/TestFramework/Exceptions/MSTestException.cs
+++ b/src/TestFramework/TestFramework/Exceptions/MSTestException.cs
@@ -1,0 +1,57 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System.ComponentModel;
+using System.Runtime.Serialization;
+
+namespace Microsoft.VisualStudio.TestTools.UnitTesting;
+
+/// <summary>
+/// Base class for Framework Exceptions.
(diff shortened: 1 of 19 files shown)


src/TestFramework/TestFramework/Exceptions/MSTestException.cs
```diff
@@ -0,0 +1,57 @@
...
+using System.ComponentModel;
+using System.Runtime.Serialization;
+
+namespace Microsoft.VisualStudio.TestTools.UnitTesting;
+
+/// <summary>
+/// Base class for Framework Exceptions.
+/// </summary>
+[Serializable]
+public class MSTestException : Exception
```
Youssef1313, line 13: Intentionally public?

Youssef1313 (approved):

src/TestFramework/TestFramework/Exceptions/MSTestException.cs
Evangelink: Yes I want to reuse this for an open issue we have

PR #6253, opened 2025-07-30
Add Moq.Analyzers

Evangelink:


Directory.Packages.props
```diff
@@ -34,6 +34,7 @@
     <PackageVersion Include="Microsoft.CodeAnalysis.BannedApiAnalyzers" Version="$(MicrosoftCodeAnalysisBannedApiAnalyzersVersion)" />
     <PackageVersion Include="Microsoft.CodeAnalysis.PublicApiAnalyzers" Version="$(MicrosoftCodeAnalysisPublicApiAnalyzersVersion)" />
     <PackageVersion Include="Microsoft.VisualStudio.Threading.Analyzers" Version="17.14.15" />
+    <PackageVersion Include="Moq.Analyzers" Version="0.3.0" />
```
Youssef1313, line 37: Note: This package is BSD-3-Clause license. I'm not super familiar with that but thought it's worth noting.

Evangelink: I have seen other projects using it so we should be fine. The only question is whether we need to specify that in the third-party licenses given this is only for testing and not shipped.

Youssef1313: I think we need it in third-party-notices indeed.

test/UnitTests/MSTestAdapter.PlatformServices.UnitTests/Deployment/AssemblyLoadWorkerTests.cs
```diff
@@ -22,7 +22,7 @@ public void GetFullPathToDependentAssembliesShouldReturnV1FrameworkAssembly()
             GetReferencedAssembliesSetter = () => [v1AssemblyName],
         };
 
-        var mockAssemblyUtility = new Mock<IAssemblyUtility>();
+        var mockAssemblyUtility = new Mock<IAssemblyUtility>(MockBehavior.Loose);
```
Youssef1313, line 25: I feel this analyzer is too annoying. There is nothing very wrong with not specifying the "default" `MockBehavior.Loose`.

Evangelink: It's always about making explicit choice, we have something similar about parallelization in MSTest. Ideally, we should only mock with strict level to ensure no other API is called (another analyzer is set for that).

I am happy to undo changes and reduce severity of this.

Youssef1313: > Ideally, we should only mock with strict level

I do agree, but I think we are too far from reaching there.

I'm fine as-is then, even though I personally don't see a point when specifying Loose everywhere. The advantage I can see here is for tests that will be written in future. This might ring a bell for us to not rely on Loose and have the new tests going into the right direction.

Youssef1313 (dismissed):
There are few more analyzer warnings, but LGTM.

Youssef1313 (changes requested):
Blocking on license question to avoid accidental merge

PR #6254, opened 2025-07-30
Made changes to dev-guide

AtolagbeMuiz:
updated the dev-guide.md to include common issue I faced while building and loading the project repository on my local machine.


docs/dev-guide.md
```diff
@@ -166,3 +166,14 @@ Note that `-test` allows to run the unit tests and `-integrationTest` allows to
...
 
 Inside Visual Studio, all projects can be built normally. All but acceptance tests can be tested directly from Visual Studio. The acceptance tests will always use the version of the NuGet packages produced in the `artifacts/packages/shipping` folder so if you have made some changes and run these tests, it's likely that the changes will not be applied.
+
+## Common Build issues while loading project with Visual Studio
+NB: If working with Visual Studio, this project repository contains a Visual Studio solution file of extension type .slnx, which is the new XML-based solution file format in Visual Studio (i.e. TestFx.slnx). This solution file can only be opened or loaded successfully using Visual Studio 2022 17.13 or higher.
+
+If you attempt to build using the `Build.cmd` script in the repository root or run `build.cmd -configuration release` in the terminal with a Visual Studio version other than Visual Studio 2022 17.13 or higher installed on your machine, you will encounter a build error as shown below.
```
Youssef1313: This part is surprising to me. I thought support for building slnx from command-line comes through .NET SDK (>= 9.0.200). And we already pin to .NET 10 in global.json.

@baronfel Is the VS installed on machine having impact here when building from command-line with regards with slnx support?

docs/dev-guide.md
```diff
@@ -166,3 +166,14 @@ Note that `-test` allows to run the unit tests and `-integrationTest` allows to
 If you are working with Visual Studio, we recommend opening it through the `open-vs.cmd` script at the repo root. This script will set all the required environment variables required so that Visual Studio picks up the locally downloaded version of the .NET SDK. If you prefer to use your machine-wide configuration, you can open Visual Studio directly.
 
 Inside Visual Studio, all projects can be built normally. All but acceptance tests can be tested directly from Visual Studio. The acceptance tests will always use the version of the NuGet packages produced in the `artifacts/packages/shipping` folder so if you have made some changes and run these tests, it's likely that the changes will not be applied.
+
+## Common Build issues while loading project with Visual Studio
```
Youssef1313:
```suggestion
## Visual Studio version requirement
```

docs/dev-guide.md
```diff
@@ -166,3 +166,14 @@ Note that `-test` allows to run the unit tests and `-integrationTest` allows to
...
+NB: If working with Visual Studio, this project repository contains a Visual Studio solution file of extension type .slnx, which is the new XML-based solution file format in Visual Studio (i.e. TestFx.slnx). This solution file can only be opened or loaded successfully using Visual Studio 2022 17.13 or higher.
```
Youssef1313:
```suggestion

If working with Visual Studio, this repository uses the new, modern, XML-based slnx solution file format (`TestFx.slnx`). This solution file can only be opened or loaded successfully using Visual Studio 2022 17.13 or higher.
```

docs/dev-guide.md
AtolagbeMuiz: yes, you are right.. I had to install .NET SDK 9.0.200 but still doesn't allow other version Visual Studio other than Visual Studio 2022 17.13 or higher to load the TestFx.snlx  solution file.

docs/dev-guide.md
```diff
@@ -166,3 +166,16 @@ Note that `-test` allows to run the unit tests and `-integrationTest` allows to
 If you are working with Visual Studio, we recommend opening it through the `open-vs.cmd` script at the repo root. This script will set all the required environment variables required so that Visual Studio picks up the locally downloaded version of the .NET SDK. If you prefer to use your machine-wide configuration, you can open Visual Studio directly.
 
 Inside Visual Studio, all projects can be built normally. All but acceptance tests can be tested directly from Visual Studio. The acceptance tests will always use the version of the NuGet packages produced in the `artifacts/packages/shipping` folder so if you have made some changes and run these tests, it's likely that the changes will not be applied.
+
+## Visual Studio version requirement
+
+
```
Youssef1313: Extra blank line here

```suggestion
```

docs/dev-guide.md
Youssef1313: > I had to install .NET SDK 9.0.200

That's surprising to me. Our global.json forces specific preview of .NET 10. Were you deleting `global.json` when trying to build locally?
You should be letting the build and restore scripts to install what is required in global.json. In that case, that should always work in command-line.

AtolagbeMuiz: I didn't delete the `global.json`, If the `global.json` installs the required packages, it still requires VS 2022 17.13 or higher to load the project if any person is working with Visual Studio.

Evangelink: Our build script will download the .NET 10 SDK we use under `.dotnet` folder at the root, but you are right we still need recent version of VS and that's worth documenting.

Youssef1313: My point is, if building from command-line, it should always work regardless of VS version? I'm only *guessing* here. So I want to understand if you really saw issues when using .NET 10 SDK specified by global.json?

Evangelink: My understanding is that `build.cmd` alone should work but then you won't be able to open the solution in VS before 17.13. In CLI, our deps would be 17.8 according to https://github.com/microsoft/testfx/blob/main/global.json#L22

Youssef1313: @Evangelink Interesting point about global.json. I think that part of global.json is only used by Arcade to determine msbuild that needs to be used for building, so maybe that's what causes issues?

Youssef1313: I think with https://github.com/microsoft/testfx/pull/6296, maybe we will always be able to build correctly from command-line, regardless of the installed VS version. In that case, the part about building from command-line could be dropped maybe?

Evangelink: Not sure, let's try to merge your PR and see if that's helping somehow.

Youssef1313: I would remove the part about building from command line for now as I think it probably should start work now that #6296 is merged (or if you can try it locally @AtolagbeMuiz and let us know). Other than that, this looks good to me!

AtolagbeMuiz: @Youssef1313  if I get you right, you want me to try it locally now.. if I can load the project irrespective of the installed Visual Studio version right with the [#6296](https://github.com/microsoft/testfx/pull/6296)...  If so, I will try it and revert

Youssef1313: Yes, that will be very much appreciated. I'm expecting that **command-line** builds using `build.cmd` will work starting with #6296, regardless of the installed VS version.

However, if you want to open in Visual Studio, for sure 17.13 or higher is required.

AtolagbeMuiz: I have just executed the `build.cmd` locally and the build was successful.

And as you stated, 17.13 will be required to open via Visual Studio

Youssef1313: Thanks for confirming. In that case let's remove the paragraph about `Build.cmd`. Then this should be ready to merge 🎉

Youssef1313:
Thanks for the contribution @AtolagbeMuiz. This is very appreciated!

PR #6281, opened 2025-08-01
Cleanup TerminalTestReporter

Evangelink:
- Reduce nesting
- factorize code

### DIFF
--- a/src/Platform/Microsoft.Testing.Platform/OutputDevice/Terminal/TerminalTestReporter.cs
+++ b/src/Platform/Microsoft.Testing.Platform/OutputDevice/Terminal/TerminalTestReporter.cs
@@ -548,64 +548,68 @@ internal sealed partial class TerminalTestReporter : IDisposable
     private void AppendAssemblyLinkTargetFrameworkAndArchitecture(ITerminal terminal)
     {
         terminal.AppendLink(_assembly, lineNumber: null);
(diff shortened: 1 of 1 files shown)


src/Platform/Microsoft.Testing.Platform/OutputDevice/Terminal/TerminalTestReporter.cs
```diff
@@ -548,64 +548,68 @@ private static void FormatStandardAndErrorOutput(ITerminal terminal, string? sta
...
 
-            if (_architecture != null)
-            {
-                terminal.Append(_architecture);
-            }
+        terminal.Append(" (");
+        if (_targetFramework != null)
+        {
+            terminal.Append(_targetFramework);
+            terminal.Append('|');
```
Youssef1313, line 560: Unrelated to your change, but maybe we should only append `|` if both TFM and arch are non-null. Otherwise, we can end up with `(net8.0|)`

Youssef1313 (approved):

src/Platform/Microsoft.Testing.Platform/OutputDevice/Terminal/TerminalTestReporter.cs
Evangelink: Let me do a follow-up.

Issue #662, opened 2019-11-21
IsSubset should return a better default AssertMessage


LokiMidgard:
## Description
The `CollectionAssert.IsSubsetOf` error message could have more information.

Currently it only says it failed. But what elements are missing is actually ommitted.

https://github.com/microsoft/testfx/blob/2166170f73c9aa13124286c34f675a5cbebee988/src/TestFramework/MSTest.Core/Assertions/CollectionAssert.cs#L451-L459

The helper method seems to have the information but will not give that to the caller. (It takes a shourtcut on an error not searching for mor problems but even one element that is missing is an information for the tester)

https://github.com/microsoft/testfx/blob/2166170f73c9aa13124286c34f675a5cbebee988/src/TestFramework/MSTest.Core/Assertions/CollectionAssert.cs#L1270-L1306

## Steps to reproduce

Write a Method that uses `Assert.IsSubsetOf` with an input that will fail

## Expected behavior
Information what elements are missing in the collection.

## Actual behavior
The Message only says something with the collection is wrong. `CollectionAssert.IsSubsetOf failed.`

## Environment
* MSTest.TestAdapter (1.4.0)
* MSTest.TestFramework (1.4.0)

jayaranigarg:
@LokiMidgard: Thank you for your feedback. This certainly makes sense. We will be happy to take contribution for the same if you would like to pick this up.

LokiMidgard:
I'm not sure when I could get to this. This would propably be spontanius.

I have written an extension method for now to use in my project. 

```c#
        public static void IsSubsetOf<T>(ICollection<T> subset, ICollection<T> superset, string? message)
        {
            var elementsLeft = subset.Except(superset).Select(x => $" * ({x})");
            var missingText = string.Join("\n", elementsLeft);

            if (missingText != string.Empty)
            {
                if (message is null)
                    HandleFail("\n"+missingText);
                else
                    HandleFail($"{message}\n{missingText}");
            }
        }
```
Compared with the method currently used the above propably has a problem with double values.

If I would implement this, I would tend to use the mechanic already present and write the output like this:
```
Assert Failed. x Elements were missing.
 [* Value1]
 [* Value2]
 [* Value3]
 [* More values were ommited.]
The collection contaed x elements.
 [* Value 1
 * Value 2
 * Value 3]
```

I would print up to 3 values that where not in the superset and all values for the superset if the collection does not contain more then 5.

I'm open for better suggestions.

[...]

PR #6292, opened 2025-08-03
Is subsetof return non subset values of the superset (relates to #662)

AtolagbeMuiz:
This PR fixes [IsSubset should return a better default AssertMessage](https://github.com/microsoft/testfx/issues/662) #662 

This implementation allows The CollectionAssert.IsSubsetOf default message to have more information regarding the message being returned. This message returns value(s) that is/are not a subset of a given superset.

### Current Behaviour
Currently, The Message only says `CollectionAssert.IsSubsetOf failed.`

### Expected Behaviour
With this Implementation, the message becomes more explanatory by giving more Information on what elements are missing in the collection. The message that will be returned is `CollectionAssert.IsSubsetOf: values {"test1", "test2"} is/are not a subset of the superset`


src/TestFramework/TestFramework/TestFramework.csproj
```diff
@@ -25,6 +25,7 @@
 
   <ItemGroup>
     <PackageReference Include="Polyfill" PrivateAssets="all" />
+    <PackageReference Include="System.ValueTuple" />
```
Youssef1313: Adding dependencies is often tricky. I would avoid it.

AtolagbeMuiz: I will review this alongside other warnings that caused failure in the PR checklist.

AtolagbeMuiz: > Adding dependencies is often tricky. I would avoid it.

The dependencies have been uninstalled and its references have been avoided as requested in the a new release and the PR has been updated.

src/TestFramework/TestFramework/Assertions/CollectionAssert.cs
```diff
@@ -418,9 +418,23 @@ public static void IsSubsetOf([NotNull] ICollection? subset, [NotNull] ICollecti
     {
         Assert.CheckParameterNotNull(subset, "CollectionAssert.IsSubsetOf", "subset", string.Empty);
         Assert.CheckParameterNotNull(superset, "CollectionAssert.IsSubsetOf", "superset", string.Empty);
-        if (!IsSubsetOfHelper(subset, superset))
+        Tuple<bool, ICollection<object?>> isSubsetValue = IsSubsetOfHelper(subset, superset);
+        if (!isSubsetValue.Item1)
         {
-            Assert.ThrowAssertFailed("CollectionAssert.IsSubsetOf", Assert.BuildUserMessage(message, parameters));
+            string returnedSubsetValueMessage = string.Empty;
```
Youssef1313: Consider using a StringBuilder instead of concatenating in a loop

src/TestFramework/TestFramework/Assertions/CollectionAssert.cs
```diff
@@ -418,9 +418,23 @@ public static void IsSubsetOf([NotNull] ICollection? subset, [NotNull] ICollecti
...
+                returnedSubsetValueMessage = returnedSubsetValueMessage + " " + Convert.ToString(item, CultureInfo.InvariantCulture) + ",";
+            }
+
+            // Remove the last trailing comma
+            if (!string.IsNullOrEmpty(returnedSubsetValueMessage))
+            {
+                returnedSubsetValueMessage = returnedSubsetValueMessage.TrimEnd(',').Trim();
+            }
+
+            returnedSubsetValueMessage = "Element(s) {" + returnedSubsetValueMessage.Trim() + "} is/are not a subset of the superset";
```
Youssef1313: This string needs to be localized. To do that, you need to add it to resx file. After that, when you build, the build will automatically update Xlf files and mark them as need translation. The actual translation will be brought in an automated PR later on (you don't need to care about the actual translations).

Youssef1313 (approved):
Overall, LGTM!

Youssef1313 (approved):
Thanks @AtolagbeMuiz.
One final thing to consider, it would be very great if you can add some unit tests to assert the message

AtolagbeMuiz:
> Thanks @AtolagbeMuiz. One final thing to consider, it would be very great if you can add some unit tests to assert the message

This has now been done, and the PR updated. unit tests were written to assert the message

src/TestFramework/TestFramework/Assertions/CollectionAssert.cs
```diff
@@ -418,9 +418,25 @@ public static void IsSubsetOf([NotNull] ICollection? subset, [NotNull] ICollecti
...
+                sb.Append(' ').Append(Convert.ToString(item, CultureInfo.InvariantCulture)).Append(',');
+            }
+
+            string returnedSubsetValueMessage = sb.ToString();
+
+            // Remove the last trailing comma
+            if (!string.IsNullOrEmpty(returnedSubsetValueMessage))
+            {
+                returnedSubsetValueMessage = returnedSubsetValueMessage.TrimEnd(',').Trim();
+            }
```
nohwnd: Are we re-implementing String.Join here (or pre-filtered collection, and then string.join)?

src/TestFramework/TestFramework/Assertions/CollectionAssert.cs
```diff
@@ -1438,21 +1455,30 @@ public static void AreNotEqual(ICollection? notExpected, ICollection? actual, [N
     /// True if <paramref name="subset"/> is a subset of
     /// <paramref name="superset"/>, false otherwise.
     /// </returns>
-    internal static bool IsSubsetOfHelper(ICollection subset, ICollection superset)
+    internal static Tuple<bool, ICollection<object?>> IsSubsetOfHelper(ICollection subset, ICollection superset)
```
nohwnd, line 1446: I would find it more obvious to use named tuple here, or if that is not possible then an actual type. Value1 and Value2 make it harder to understand what data I am getting out.

src/TestFramework/TestFramework/Resources/xlf/FrameworkMessages.cs.xlf
```diff
@@ -283,6 +283,11 @@ Skutečnost: {2}</target>
         <target state="translated">Rozdílný počet elementů.</target>
         <note></note>
       </trans-unit>
+      <trans-unit id="ReturnedSubsetValueMessage">
+        <source>Element(s) &lt;{0}&gt; is/are not a subset of the superset</source>
```
nohwnd:
```suggestion
        <source>Element(s) &lt;{0}&gt; is/are not present in the collection</source>
```
Alternative wording. I find it hard to process what happened when thinking about subset of superset. but very easy to understand that the values I was looking for were not in the collection.

test/UnitTests/TestFramework.UnitTests/Assertions/CollectionAssertTests.cs
```diff
@@ -502,12 +543,16 @@ private static List<object> GenerateDeeplyNestedCollection(int depth)
...
 
+    private ICollection? GetSubsetCollection() => new[] { "iem", "a", "b" };
+
     private object? GetMatchingElement() => "item";
 
     private object? GetNotMatchingElement() => "not found";
 
     private ICollection? GetMatchingSuperSet() => new[] { "item", "item2" };
 
+    private ICollection? GetSuperSetCollection() => new[] { "item", "item2", "c", "d" };
```
nohwnd: the typo was already there, but I think it should be changed to Superset everywhere, not SuperSet.

nohwnd (commented):
Hello, found few things that I would suggest revisiting. Please let me know if you have any concerns, or need to discuss it further.

Youssef1313 (approved):

Issue #6317, opened 2025-08-06
MSTest performance


Youssef1313:
This issue tracks few ideas for improving the performance of MSTest.

- We shouldn't do any serialization for parameterized tests if appdomain is disabled.
- ManagedNameHelper from VSTest is extremely slow and memory consuming
    - Note: on a single test class with 10k tests.
        
        <img width="2386" height="943" alt="Image" src="https://github.com/user-attachments/assets/f703b9fc-3615-4c37-b61b-ba28b1c946cf" />

        <img width="2399" height="685" alt="Image" src="https://github.com/user-attachments/assets/6d4ec70c-8b41-4464-92c5-cdd8db592031" />

- `Dictionary<string, string>` used in `UpdateInfoIfTestInitializeOrCleanupMethod`:
   - is used to act more as of HashSet. So arguably a HashSet will be simply faster.
   - is consuming more time and memory for providing zero functionality if the test class inherits directly from `object`.
   - Its overall value is too low (prevents considering base initializes if they are shadowed in concrete test class). Should we get rid of it altogether?

nohwnd:
graphed by copilot, since my brain does not comprehend the numbers without being in a table.

<img width="937" height="553" alt="Image" src="https://github.com/user-attachments/assets/eacc2b21-b198-4e00-806e-a8421fc19b77" />

Youssef1313:
> Consider moving this as documentation instead of keeping an issue ticket.

There is still a bit of work needs to be done here.

PR #6318, opened 2025-08-06
Performance: Reduce unnecessary logic in UpdateInfoIfTestInitializeOrCleanupMethod

Youssef1313:
Related to https://github.com/microsoft/testfx/issues/6317

### DIFF
--- a/src/Adapter/MSTestAdapter.PlatformServices/Execution/TypeCache.cs
+++ b/src/Adapter/MSTestAdapter.PlatformServices/Execution/TypeCache.cs
@@ -314,7 +314,7 @@ internal sealed class TypeCache : MarshalByRefObject
         // which is used to decide whether TestInitialize/TestCleanup methods
         // present in the base type should be used or not. They are not used if
         // the method is overridden in the derived type.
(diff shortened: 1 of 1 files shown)


src/Adapter/MSTestAdapter.PlatformServices/Execution/TypeCache.cs
```diff
@@ -621,16 +621,16 @@ private void UpdateInfoIfTestInitializeOrCleanupMethod(
         TestClassInfo classInfo,
         MethodInfo methodInfo,
         bool isBase,
-        Dictionary<string, string?> instanceMethods)
+        HashSet<string>? instanceMethods)
```
drognanar, line 624: is this special casing the case, where the base class is object, thus we do not need to track the `instanceMethods`?

Youssef1313: @drognanar Yup. I'm passing null when base class is object as we don't want to track anything. The tracking is only relevant when we call `UpdateInfoIfTestInitializeOrCleanupMethod` again with a base class.

drognanar (approved):

PR #6344, opened 2025-08-11
[rel/4.0]: Unify TestSessionContext

Youssef1313:
Related to #5762


src/Platform/Microsoft.Testing.Platform/Services/TestSessionContext.cs
```diff
@@ -5,9 +5,24 @@
 
 namespace Microsoft.Testing.Platform.Services;
 
-internal sealed class TestSessionContext(CancellationToken cancellationToken) : ITestSessionContext
```
Youssef1313, line 8: NOTE: Previously this was internal (in Services namespace), and another public TestSessionContext was in TestHost namespace.

I'm not really sure if it makes more sense to expose in Services namespace or TestHost namespace.

The notable public APIs with `TestSessionContext` in signature is `TestExecutionRequest.Session` (well, the property name isn't super good IMO but that's a different discussion). The TestExecutionRequest is kinda testhost-specific I think.

PR #6352, opened 2025-08-12
Cleanup TestContextImplementation

Youssef1313:


src/Adapter/MSTestAdapter.PlatformServices/Execution/UnitTestRunner.cs
```diff
@@ -154,8 +155,7 @@ internal async Task<TestResult[]> RunSingleTestAsync(TestMethod testMethod, IDic
 
         try
         {
-            var properties = new Dictionary<string, object?>(testContextProperties);
```
nohwnd, line 157: This looks like it might be done on purpose to "clone" the original dictionary, so edits to it won't spill to other tests?

Youssef1313: @nohwnd In this code path, the dictionary is only passed to `TestContextImplementation` constructor, which already clones it. So we were double cloning.

nohwnd (approved):

PR #6367, opened 2025-08-13
Perf: Reduce allocations from ManagedNameHelper

Youssef1313:
Scenario: 10k tests in a test class

Memory allocations before: ~227MB
Memory allocations after: ~208MB

### DIFF
--- a/src/Adapter/MSTestAdapter.PlatformServices/Execution/TypeCache.cs
+++ b/src/Adapter/MSTestAdapter.PlatformServices/Execution/TypeCache.cs
@@ -154,16 +154,6 @@ internal sealed class TypeCache : MarshalByRefObject
 
             // Load the class type
             Type? type = LoadType(typeName, testMethod.AssemblyName);
-
-            // VSTest managed feature is not working properly and ends up providing names that are not fully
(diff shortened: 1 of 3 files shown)


src/Adapter/MSTestAdapter.PlatformServices/Execution/TypeCache.cs
```diff
@@ -154,16 +154,6 @@ public IEnumerable<TestAssemblyInfo> AssemblyInfoListWithExecutableCleanupMethod
...
             Type? type = LoadType(typeName, testMethod.AssemblyName);
-
-            // VSTest managed feature is not working properly and ends up providing names that are not fully
-            // unescaped causing reflection to fail loading. For the cases we know this is happening, we will
-            // try to manually unescape the type name and load the type again.
-            if (type == null
-                && TryGetUnescapedManagedTypeName(testMethod, out string? unescapedTypeName))
-            {
-                type = LoadType(unescapedTypeName, testMethod.AssemblyName);
-            }
```
Youssef1313, line 165: I need to validate if there are any F# issues here with types containing spaces or not.

PR #6378, opened 2025-08-14
Don't store fixture tests when the feature isn't used

Youssef1313:
String concatenation in this code path is responsible for 1.7% of allocations on current main.

image

### DIFF
--- a/src/Adapter/MSTestAdapter.PlatformServices/Execution/UnitTestRunner.cs
+++ b/src/Adapter/MSTestAdapter.PlatformServices/Execution/UnitTestRunner.cs
@@ -172,8 +172,11 @@ internal sealed class UnitTestRunner : MarshalByRefObject
                 DebugEx.Assert(testMethodInfo is not null, "testMethodInfo should not be null.");
 
                 // Keep track of all non-runnable methods so that we can return the appropriate result at the end.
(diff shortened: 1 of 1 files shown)


src/Adapter/MSTestAdapter.PlatformServices/Execution/UnitTestRunner.cs
```diff
@@ -172,8 +172,11 @@ internal async Task<TestResult[]> RunSingleTestAsync(TestMethod testMethod, IDic
                 DebugEx.Assert(testMethodInfo is not null, "testMethodInfo should not be null.");
 
                 // Keep track of all non-runnable methods so that we can return the appropriate result at the end.
-                _assemblyFixtureTests.TryAdd(testMethod.AssemblyName, testMethodInfo.Parent.Parent);
-                _classFixtureTests.TryAdd(testMethod.AssemblyName + testMethod.FullClassName, testMethodInfo.Parent);
+                if (MSTestSettings.CurrentSettings.ConsiderFixturesAsSpecialTests)
```
drognanar, line 175: do we need a new test for this?

Youssef1313: There is no behavior change to test. The dictionary is only read when ConsiderFixturesAsSpecialTests is true. So we avoid adding to the dictionary (and allocating strings) that are never read.

PR #6405, opened 2025-08-18
Implement analyzer for TestMethodAttribute inheritance

Youssef1313:
Related to #1285

### DIFF
--- a/eng/Version.Details.xml
+++ b/eng/Version.Details.xml
@@ -1,17 +1,17 @@
 <?xml version="1.0" encoding="utf-8"?>
 <Dependencies>
   <ToolsetDependencies>
-    <Dependency Name="Microsoft.DotNet.Arcade.Sdk" Version="10.0.0-beta.25415.5">
+    <Dependency Name="Microsoft.DotNet.Arcade.Sdk" Version="10.0.0-beta.25418.4">
       <Uri>https://github.com/dotnet/arcade</Uri>
-      <Sha>e58b086a85a1df5762c8980308f7e4e330edd0aa</Sha>
+      <Sha>5bb3e45c964be0076b22510ce60218284c7bba53</Sha>
     </Dependency>
-    <Dependency Name="Microsoft.DotNet.Build.Tasks.Templating" Version="10.0.0-beta.25415.5">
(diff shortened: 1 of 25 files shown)


src/Analyzers/MSTest.Analyzers.CodeFixes/TestMethodAttributeShouldPropagateSourceInformationFixer.cs
```diff
@@ -0,0 +1,250 @@
...
+            context.RegisterCodeFix(
+                CodeAction.Create(
+                    title: CodeFixResources.AddCallerInfoParametersFix,
+                    createChangedDocument: c => AddCallerInfoParametersAsync(context.Document, constructorDeclaration, root),
+                    equivalenceKey: nameof(TestMethodAttributeShouldPropagateSourceInformationFixer)),
+                diagnostic);
+        }
+        else if (identifierToken.Parent is BaseTypeDeclarationSyntax typeDeclarationSyntax)
+        {
+            // There is no explicit constructor. We only have the implicit parameterless constructor. We will the constructor.
```
nohwnd: second sentence, word missing? we will <what?> the constructor.

nohwnd (commented):
Looks great. Out of curiosity, is there some way that the user won't have the attributes available but would still be able to use mstest v4? (e.g. using an old language verison, outdated compiler that still can compile our minimal TFM, or something similar).

src/Analyzers/MSTest.Analyzers.CodeFixes/TestMethodAttributeShouldPropagateSourceInformationFixer.cs
Youssef1313: "add" :)
Will fix.

Youssef1313:
```suggestion
            // There is no explicit constructor. We only have the implicit parameterless constructor. We will add the constructor with the needed parameters.
```

Evangelink (approved):
LGTM - it could be good to add one more test with caller file path defined after caller line number just to ensure we don't rely on their order (maybe even with another optional parameter in between).

PR #6416, opened 2025-08-20
MSTEST0037: Recognize IsTrue(collection.[Count|Length] [==|>|!=] 0) patterns and cleanup

### LINKED ISSUE #6312: Update MSTEST0037 to propose Assert.IsNotEmpty for more cases
MSTEST0037 could also propose to use [Assert.IsNotEmpty](https://learn.microsoft.com/en-us/dotnet/api/microsoft.visualstudio.testtools.unittesting.assert.isnotempty?view=mstest-net-3.10) for the following code:
```c#
Assert.IsTrue(myList.Count > 0); // Assert.IsGreaterThan is currently proposed
Assert.IsTrue(myList.Count != 0); // Assert.AreNotEqual is currently proposed
```

Ideally, this would be reported for both generic and non-generic collections.

Youssef1313:
Replaces https://github.com/microsoft/testfx/pull/6341

Fixes #6312

### DIFF
--- a/test/UnitTests/MSTest.Analyzers.UnitTests/UseProperAssertMethodsAnalyzerTests.cs
+++ b/test/UnitTests/MSTest.Analyzers.UnitTests/UseProperAssertMethodsAnalyzerTests.cs
@@ -2537,14 +2537,19 @@ public sealed class UseProperAssertMethodsAnalyzerTests
                 {
                     var list = new List<int> { 1, 2, 3 };
                     Assert.IsNotEmpty(list);
+                    Assert.IsNotEmpty(list);
                 }
             }
             """;
 
         await VerifyCS.VerifyCodeFixAsync(
             code,
-            // /0/Test0.cs(11,9): info MSTEST0037: Use 'Assert.IsNotEmpty' instead of 'Assert.IsFalse'
-            VerifyCS.DiagnosticIgnoringAdditionalLocations().WithLocation(0).WithArguments("IsNotEmpty", "IsFalse"),
+            [
+                // /0/Test0.cs(11,9): info MSTEST0037: Use 'Assert.IsNotEmpty' instead of 'Assert.IsFalse'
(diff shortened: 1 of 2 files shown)


test/UnitTests/MSTest.Analyzers.UnitTests/UseProperAssertMethodsAnalyzerTests.cs
```diff
@@ -2160,4 +2160,465 @@ public void MyTestMethod()
...
+            using System.Collections.Generic;
+
+            [TestClass]
+            public class MyTestClass
+            {
+                [TestMethod]
+                public void MyTestMethod()
+                {
+                    var list = new List<int> { 1, 2, 3 };
+                    // This should use the generic equality logic, not IsNotEmpty
```
nohwnd: This comment seems wrong, and I see the same comment in multiple places.

Youssef1313: @nohwnd It seems correct to me?
The assert is doing `!= 5`, so we can't use `IsEmpty` or `IsNotEmpty` as it's different semantics.

nohwnd (commented):
description mentions collection.Count == 0 but it is not tested (afaik), is that on purpose?

test/UnitTests/MSTest.Analyzers.UnitTests/UseProperAssertMethodsAnalyzerTests.cs
nohwnd: Is the comment usable only for the fixed code in that case?

What confuses me is the comment being above a line that does not use either of the things that it mentions. Maybe it is better to just remove it?

Youssef1313: Removed the comment. The test is obvious enough without it.

nohwnd (approved):

Issue #6184, opened 2025-07-26
Collection assert methods should accept IEnumerable collections


avivanoff:
## Describe the bug

Collection assert methods, like Assert.IsEmpty, Assert.HasCount, Assert.Contains, etc. should accept IEnumerable collections.

## Steps To Reproduce

1. Create a test project that targets 3.10 preview MSTest SDK.
2. Set `MSTestAnalysisMode` to `All`.
3. Add a test that uses one of the aforementioned methods.
4. Observe the errors.

[Sample](https://github.com/user-attachments/files/21441668/MSTEST0037.zip) project.

AtolagbeMuiz:
Hi @Youssef1313 

I have been playing around this issue to understand the full context as well as the closed linked duplicate issue... so what i found out from the sample project attached to this issue is that, `Assert.Contains` doesn't allow non-generic collections. So I tried implementing another overload method for `Assert.Contains` that allows non-generic collection, the implementation works for non-generic collections (like ArrayList and Stack)  except the ones that are key/value pairs like HashTable.. 

- is that of key/value pair going be implemented separately as different method overload?
- also I saw you kind of bypassed this issue with your PR #6300 , where does this issue stand if the implementation is done?

looping @Evangelink

PR #6417, opened 2025-08-20
Implement assert contains overload to accept non-generic collection

AtolagbeMuiz:
This Pull Request fixes [Collection assert methods should accept IEnumerable collections](https://github.com/microsoft/testfx/issues/6184) #6184 

This implementation involves an overload of `Assert.Contains` which allows non-generic IEnumerable collection like ArrayList and Stack.

**Current Behaviour**
Currently, the `Assert.Contains` function doesn't allow non-generic collection like ArrayList as this throws compile time error as seen below

image

**Expected Behaviour**
With this Implementation, `Assert.Contains` will allow non-generic collection as seen below;

image

image

### DIFF
--- a/src/TestFramework/TestFramework/Assertions/Assert.Contains.cs
+++ b/src/TestFramework/TestFramework/Assertions/Assert.Contains.cs
@@ -712,12 +712,10 @@ public sealed partial class Assert
         {
             if (predicate(item))
             {
-                return;
+                string userMessage = BuildUserMessage(message);
+                ThrowAssertDoesNotContainPredicateFailed(userMessage);
             }
         }
-
-        string userMessage = BuildUserMessage(message);
-        ThrowAssertDoesNotContainPredicateFailed(userMessage);
     }
 
     /// <summary>


src/TestFramework/TestFramework/Assertions/Assert.Contains.cs
```diff
@@ -205,6 +213,32 @@ public static void Contains<T>(T expected, IEnumerable<T> collection, string? me
...
     }
 
+    /// <summary>
+    /// Tests whether the specified collection contains the given element.
+    /// </summary>
+    /// <param name="expected">The expected item.</param>
+    /// <param name="collection">The collection.</param>
+    /// <param name="message">The message format to display when the assertion fails.</param>
+    /// <param name="parameters">The parameters to format the message.</param>
+    public static void Contains(object expected, IEnumerable collection, string? message, params object?[]? parameters)
```
Youssef1313: I wouldn't add this overload. We are removing the `params` overloads in MSTest v4 anyways.

src/TestFramework/TestFramework/PublicAPI/PublicAPI.Shipped.txt
```diff
@@ -642,6 +642,8 @@ static Microsoft.VisualStudio.TestTools.UnitTesting.Assert.Contains(string! subs
 static Microsoft.VisualStudio.TestTools.UnitTesting.Assert.Contains(string! substring, string! value, System.StringComparison comparisonType, string? message) -> void
 static Microsoft.VisualStudio.TestTools.UnitTesting.Assert.Contains(string! substring, string! value, System.StringComparison comparisonType) -> void
 static Microsoft.VisualStudio.TestTools.UnitTesting.Assert.Contains(string! substring, string! value) -> void
+static Microsoft.VisualStudio.TestTools.UnitTesting.Assert.Contains(object! expected, System.Collections.IEnumerable! collection) -> void
```
Youssef1313: Changes should happen to PublicAPI.Unshipped.txt, not PublicAPI.Shipped.txt.

test/UnitTests/TestFramework.UnitTests/Assertions/AssertTests.Contains.cs
```diff
@@ -362,6 +362,36 @@ public void Contains_ValueExpected_ItemDoesNotExist_ThrowsException()
...
+    /// <summary>
+    /// Tests the Contains method (value overload) when the expected item is present.
+    /// </summary>
+    public void Contains_InNonGenericCollection_ValueExpected_ItemDoesNotExist_ThrowsException()
+    {
+        // Arrange
+        var collection = new ArrayList { 5, 10, "a" };
+
+        // Act
+        Action action = () => Assert.Contains(20, collection, "Item {0} not found", 20);
```
Youssef1313: Please also add a test for `Assert.Contains(5, collection)` to ensure it's handled correctly and doesn't throw.

src/TestFramework/TestFramework/Assertions/Assert.Contains.cs
AtolagbeMuiz: @Youssef1313  you want me to remove the `params object?[]? parameters` argument?? the method should be left with just 3 arguments `object expected, IEnumerable collection, string? message`

test/UnitTests/TestFramework.UnitTests/Assertions/AssertTests.Contains.cs
AtolagbeMuiz: This test has been added and added to the Pull Request

src/TestFramework/TestFramework/PublicAPI/PublicAPI.Shipped.txt
AtolagbeMuiz: this has been corrected.. the declaration moved to PublicAPI.Unshipped.txt..

src/TestFramework/TestFramework/Assertions/Assert.Contains.cs
Youssef1313: Yes, just `expected`, `collection`, and `message` should be enough.

src/TestFramework/TestFramework/Assertions/Assert.Contains.cs
```diff
@@ -178,6 +178,14 @@ public static T ContainsSingle<T>(Func<T, bool> predicate, IEnumerable<T> collec
     public static void Contains<T>(T expected, IEnumerable<T> collection)
         => Contains(expected, collection, string.Empty, null);
 
+    /// <summary>
+    /// Tests whether the specified non-generic collection contains the given element.
+    /// </summary>
+    /// <param name="expected">The expected item.</param>
+    /// <param name="collection">The non-generic collection (like ArrayList).</param>
+    public static void Contains(object expected, IEnumerable collection)
```
Youssef1313: I think we need similar non-generic overloads also for `IsEmpty` and `HasCount`. Then MSTEST0037 analyzer need to be adjusted as well to account for the existence of the non-generic overloads:

https://github.com/microsoft/testfx/blob/8f99258760961972b88c1e60efc81d31d1464af6/src/Analyzers/MSTest.Analyzers/UseProperAssertMethodsAnalyzer.cs#L464-L471

AtolagbeMuiz: Yes, I intend to implement this for IsEmpty etc. as well.. still yet to have a good understanding of the Analyzer part.

Youssef1313: Once you add the overload for `HasCount`, this analyzer test will fail:

https://github.com/microsoft/testfx/blob/8f99258760961972b88c1e60efc81d31d1464af6/test/UnitTests/MSTest.Analyzers.UnitTests/UseProperAssertMethodsAnalyzerTests.cs#L2134

The test will need to be updated as following (note: I'm writing it directly on GitHub so there might be small mistakes):

```csharp
    [TestMethod]
    public async Task WhenAssertAreEqualWithCollectionCountUsingNonGenericCollection()
    {
        string code = """
            using System;
            using System.Collections;

            [TestClass]
            public class MyTestClass
            {
                [TestMethod]
                public void MyTestMethod()
                {
                    var x = new Hashtable();
                    {|#0:Assert.AreEqual(4, x.Count)|};
                }
            }
            """;

        string fixedCode = """
            using System;
            using System.Collections;

            [TestClass]
            public class MyTestClass
            {
                [TestMethod]
                public void MyTestMethod()
                {
                    var x = new Hashtable();
                    Assert.HasCount(4, x);
                }
            }
            """;

        await VerifyCS.VerifyCodeFixAsync(
            code,
            // /0/Test0.cs(13,9): info MSTEST0037: Use 'Assert.HasCount' instead of 'Assert.AreEqual'
            VerifyCS.DiagnosticIgnoringAdditionalLocations().WithLocation(0).WithArguments("HasCount", "AreEqual"),
            fixedCode);
    }
```

The analyzer implementation then will need to check `System_Collections_Generic_IEnumerable` instead of `System_Collections_Generic_IEnumerable_T`

Note: because we use `AllInterfaces`, and `IEnumerable<T>` is `IEnumerable`, we need only `IEnumerable` check. We don't need both `IEnumerable<T>` and `IEnumerable`.
Note 2: `OriginalDefinition` access will become redundant, because we are no longer dealing with generics.

AtolagbeMuiz: i will revert back to this once I want to implement for `HasCount`

src/TestFramework/TestFramework/Assertions/Assert.Contains.cs
AtolagbeMuiz: this has been done and the PR updated..

src/TestFramework/TestFramework/Assertions/Assert.Contains.cs
Evangelink: We may want to allow nullable object. It's been a super long time since I have worked with non-generic collection but I believe we have have nulls

AtolagbeMuiz: you want me to make `object expected` nullable?

Youssef1313: @AtolagbeMuiz Yes. So if user calls `Assert.Contains(someNullableObject, collection)` they don't get a warning.

AtolagbeMuiz: this has been implemented to allow nullable expected value i.e. `object? expected`.. i also wrote unit tests to pass nulls as expected value into the `Assert.Contains`..

Youssef1313: Thanks @AtolagbeMuiz.

Are you planning to continue that soon? It's best to have all the overloads in the same PR as the analyzer is likely handling all these methods in the same way (even if we have only test coverage for `HasCount`)

AtolagbeMuiz: Hi @Youssef1313, should the implementation for other Assert methods like HasCount be in this same PR and branch and not as a different branch??

was intending to work on it on a different branch.. I should commence work on it this week.

Youssef1313: Yes, I prefer if it's the same PR in this case. Thanks @AtolagbeMuiz!

AtolagbeMuiz: Hi @Youssef1313 , will start implementation on this today.. sorry for the delay, my PC is been malfuncting so just got a new one delivered on wednesday.

AtolagbeMuiz: Hi @Youssef1313 from your statement above as quoted "The analyzer implementation then will need to check `System_Collections_Generic_IEnumerable` instead of `System_Collections_Generic_IEnumerable_T`"..

tried to make changes to the analyzer and i found out `System_Collections_Generic_IEnumerable` doesn't exist in `SpecialType`.. do you mean `SpecialType.System_Collections_IEnumerable` instead??

AtolagbeMuiz: updated the PR with the changes for non-generic overload on Assert.HasCount.. will need your review of these changes @Youssef1313

AtolagbeMuiz: changes have been made to the PR to update the Analyzer.. @Youssef1313

AtolagbeMuiz: @Youssef1313  i have also implemented non-generic overload for `IsNotEmpty`.. this is now done for `HasCount`, `IsEmpty` and `IsNotEmpty`.. the build failure in the pipeline is due to the flaky integration test.

src/Analyzers/MSTest.Analyzers/UseProperAssertMethodsAnalyzer.cs
```diff
@@ -470,7 +470,7 @@ private static bool IsBCLCollectionType(ITypeSymbol type, INamedTypeSymbol objec
         // Check if the type implements IEnumerable<T> (but is not string)
         // Note: Assert.Contains/IsEmpty/HasCount for collections accept IEnumerable<T>, but not IEnumerable.
```
Youssef1313: This comment is now outdated.

```suggestion
        // Check if the type implements IEnumerable (but is not string)
```

src/TestFramework/TestFramework/Assertions/Assert.Contains.cs
```diff
@@ -205,6 +213,31 @@ public static void Contains<T>(T expected, IEnumerable<T> collection, string? me
         }
     }
 
+    /// <summary>
+    /// Tests whether the specified collection contains the given element.
+    /// </summary>
+    /// <param name="expected">The expected item.</param>
+    /// <param name="collection">The collection.</param>
+    /// <param name="message">The message format to display when the assertion fails.</param>
+    public static void Contains(object? expected, IEnumerable collection, string? message)
```
Youssef1313, line 222: I think we should also add overloads that accept a `Func<object, bool>` as predicate?

Also there are DoesNotContain overloads that need to be added as well.

src/TestFramework/TestFramework/Assertions/Assert.Count.cs
```diff
@@ -209,6 +224,23 @@ public static void IsNotEmpty<T>(IEnumerable<T> collection, [StringSyntax(String
         ThrowAssertIsNotEmptyFailed(userMessage);
     }
 
+    /// <summary>
+    /// Tests that the collection is not empty.
+    /// </summary>
+    /// <param name="collection">The collection.</param>
+    /// <param name="message">The message format to display when the assertion fails.</param>
+    /// <param name="parameters">The parameters to format the message.</param>
+    public static void IsNotEmpty(IEnumerable collection, [StringSyntax(StringSyntaxAttribute.CompositeFormat)] string? message, params object?[]? parameters)
```
Youssef1313: I would avoid adding new APIs with `params object[]`

src/TestFramework/TestFramework/Assertions/Assert.Count.cs
```diff
@@ -251,6 +300,16 @@ public static void HasCount<T>(int expected, IEnumerable<T> collection, [Interpo
...
         => HasCount("HasCount", expected, collection, message, parameters);
 
+    /// <summary>
+    /// Tests whether the collection has the expected count/length.
+    /// </summary>
+    /// <param name="expected">The expected count.</param>
+    /// <param name="collection">The collection.</param>
+    /// <param name="message">The message format to display when the assertion fails.</param>
+    /// <param name="parameters">The parameters to format the message.</param>
+    public static void HasCount(int expected, IEnumerable collection, [StringSyntax(StringSyntaxAttribute.CompositeFormat)] string? message, params object?[]? parameters)
```
Youssef1313: I would avoid adding new APIs with `params object[]`

src/TestFramework/TestFramework/Assertions/Assert.Count.cs
```diff
@@ -289,6 +363,15 @@ public static void IsEmpty<T>(IEnumerable<T> collection, [InterpolatedStringHand
     public static void IsEmpty<T>(IEnumerable<T> collection, [StringSyntax(StringSyntaxAttribute.CompositeFormat)] string? message, params object?[]? parameters)
         => HasCount("IsEmpty", 0, collection, message, parameters);
 
+    /// <summary>
+    /// Tests that the collection is empty.
+    /// </summary>
+    /// <param name="collection">The collection.</param>
+    /// <param name="message">The message format to display when the assertion fails.</param>
+    /// <param name="parameters">The parameters to format the message.</param>
+    public static void IsEmpty(IEnumerable collection, [StringSyntax(StringSyntaxAttribute.CompositeFormat)] string? message, params object?[]? parameters)
```
Youssef1313: Same here about params.

src/TestFramework/TestFramework/Assertions/Assert.Count.cs
```diff
@@ -301,6 +384,23 @@ private static void HasCount<T>(string assertionName, int expected, IEnumerable<
         ThrowAssertCountFailed(assertionName, expected, actualCount, userMessage);
     }
 
+    private static void HasCount(string assertionName, int expected, IEnumerable collection, [StringSyntax(StringSyntaxAttribute.CompositeFormat)] string? message, params object?[]? parameters)
```
Youssef1313: Same here about `params`

test/UnitTests/MSTest.Analyzers.UnitTests/UseProperAssertMethodsAnalyzerTests.cs
```diff
@@ -2156,30 +2156,169 @@ public void MyTestMethod()
     public async Task WhenAssertAreEqualWithCollectionCountUsingNonGenericCollection()
     {
         string code = """
-            using System;
-            using System.Collections;
-            using System.Collections.Generic;
-            using Microsoft.VisualStudio.TestTools.UnitTesting;
+        using System;
```
Youssef1313: Let's keep the original indentation please

src/TestFramework/TestFramework/Assertions/Assert.Count.cs
AtolagbeMuiz: forgot we are doing away with params for new APIs.. I will correct this.

src/TestFramework/TestFramework/Assertions/Assert.Contains.cs
AtolagbeMuiz: any other methods aside DoesNotContain that needs the non-generic overload to be added?? @Youssef1313

Youssef1313: Other than the predicate overloads and DoesNotContain overloads, I don't currently see anything else.

test/UnitTests/MSTest.Analyzers.UnitTests/UseProperAssertMethodsAnalyzerTests.cs
AtolagbeMuiz: identation fixed on the analyzer tests.. I have updated the pull request..

src/TestFramework/TestFramework/Assertions/Assert.Count.cs
AtolagbeMuiz: params overload removed for `Assert.HasCount`

src/TestFramework/TestFramework/Assertions/Assert.Count.cs
AtolagbeMuiz: params overload removed for `Assert.IsEmpty` ...I have updated pull request @Youssef1313

src/TestFramework/TestFramework/Assertions/Assert.Count.cs
AtolagbeMuiz: PR updated.. params overload removed for `Assert.IsNotEmpty`

src/TestFramework/TestFramework/Assertions/Assert.Contains.cs
AtolagbeMuiz: i have added implementation for non-generic collection overload to `Assert.DoesNotContain`... I also added unit tests as well.. the pull requests has been updated..

what is left for me to work on is the method overloads that takes predicate alongside non-generic collection overload for both `Assert.DoesNotContain` and `Assert.Contain`.. @Youssef1313

Youssef1313: Thanks for your work here @AtolagbeMuiz. I'll do a final review once you finish the predicate overloads. Thanks again!

AtolagbeMuiz: I have updated the pull request with the implementation for `Assert.Contains` and `Assert.DoesNotContain` to accept predicate functions and non-generic collection. I also wrote unit tests to cover for these methods. @Youssef1313

AtolagbeMuiz:

Youssef1313:
@AtolagbeMuiz If you don't mind, I have pushed few small additional changes to your PR.

Youssef1313 (approved):

PR #6423, opened 2025-08-21
Modify internal GetRequiredService call of GetOutputDevice/GetClientInfo APIs

filzrev:
This PR modify existing `GetOutputDevice`/ `GetClientInfo`  extension method APIs to use build-in `GetRequiredService` instead of `GetRequiredServiceInternal` ,

**Background**
I want to wrap following two providers to single `IServiceProvider`.
1. MTP's `IServiceProvider` 
2. Custom IServiceProvider that created by `Microsoft.Extensions.Dependency`.

But  currently `GetOutputDevice`/ `GetClientInfo`  extension method assume IServiceProvider is `ServiceProvider` instance.
And throw exception when using these API via **wrapped** IServiceProvider.

As far as I've confirmed.
`GetRequiredService`/`GetRequiredServiceInternal` APIs difference is `skipInternalOnlyExtensions` flag only.
And it's not affect result when resolving `GetOutputDevice`/ `GetClientInfo`.

### DIFF
--- a/src/Platform/Microsoft.Testing.Platform/Services/ServiceProviderExtensions.cs
+++ b/src/Platform/Microsoft.Testing.Platform/Services/ServiceProviderExtensions.cs
@@ -93,7 +93,7 @@ public static class ServiceProviderExtensions
     /// <param name="serviceProvider">The service provider.</param>
     /// <returns>The output device.</returns>
     public static IOutputDevice GetOutputDevice(this IServiceProvider serviceProvider)
-        => serviceProvider.GetRequiredServiceInternal<IOutputDevice>();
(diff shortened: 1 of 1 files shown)


Youssef1313 (changes requested):
It's unclear to me what the benefit is.

`GetOutputDevice` should only ever be called on MTP's service provider. It seems to me you might be calling the extension method on a different service provider implementation.

filzrev:
> I don't think mixing stuff that way is a good idea.

I agree that it's not a good design to mix IServiceProviders.

But these extension methods are exposed as **public**.
So, I though it should not impose restriction unless it's not strictly required.
(Existing other public methods are using GetRequiresServide except for `IOutputDevice`/`IClientInfo`)

PR #6430, opened 2025-08-24
Implement assert contains key and contains value for keypair non generic collections

AtolagbeMuiz:
This Pull Request fixes [Collection assert methods should accept IEnumerable collections](https://github.com/microsoft/testfx/issues/6184) #6184 

This implementation creating a new of `Assert` member which are `Assert.ContainsKey` and `Assert.ContainsValue`  which accepts non-generic IEnumerable key/value pair collection like HashTable.

**Expected Behaviour of `Assert.ContainsKey`**
These screenshots are examples of how `Assert.ContainsKey` can be used for non-generic key/value pair collection like HashTable

image

image

image

**Expected Behaviour of `Assert.ContainsValue`**
These screenshots are examples of how `Assert.ContainsValue` can be used for non-generic key/value pair collection like HashTable

image

image

image

### DIFF
--- a/src/TestFramework/TestFramework/Assertions/Assert.Contains.cs
+++ b/src/TestFramework/TestFramework/Assertions/Assert.Contains.cs
@@ -179,12 +179,20 @@ public sealed partial class Assert
         => Contains(expected, collection, string.Empty, null);
 
     /// <summary>
-    /// Tests whether the specified non-generic collection contains the given element.
+    /// Tests whether the specified non-generic key/value pair collection contains the given key.
     /// </summary>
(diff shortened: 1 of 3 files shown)


Youssef1313 (changes requested):
I think in most cases users can simply use `Assert.Contains(expected, collection.[Keys|Values])`.

This feels to me low value to add, but I can discuss with rest of the team. Meanwhile, marking "requesting changes" to prevent accidental merge.

Youssef1313:
@AtolagbeMuiz Sorry, it's automation comment. The PR so far is waiting on us deciding whether we want to add this API.

PR #6435, opened 2025-08-25
Don't set insert_final_newline for xlf/resx files

Youssef1313:
Currently it applies to everything which is annoying. For example, when hand-editing resx files

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -5,7 +5,6 @@ root = true
 # All files
 [*]
 indent_style = space
-insert_final_newline = true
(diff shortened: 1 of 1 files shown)


.editorconfig
```diff
@@ -63,6 +62,7 @@ indent_size = 4
 charset = utf-8-bom
 indent_size = 4
 tab_width = 4
+insert_final_newline = true
```
Evangelink: You probably want it for more (like csproj etc)

PR #6450, opened 2025-08-27
Fix some bugs with TestMethodShouldBeValidCodeFix

Evangelink:

### DIFF
--- a/src/Analyzers/MSTest.Analyzers.CodeFixes/TestMethodShouldBeValidCodeFix.cs
+++ b/src/Analyzers/MSTest.Analyzers.CodeFixes/TestMethodShouldBeValidCodeFix.cs
@@ -93,7 +93,14 @@ public sealed class TestMethodShouldBeValidCodeFixProvider : CodeFixProvider
         SyntaxToken publicModifier = newMethodDeclaration.Modifiers.FirstOrDefault(m => m.IsKind(SyntaxKind.PublicKeyword));
         if (publicModifier == default)
         {
-            newMethodDeclaration = newMethodDeclaration.WithModifiers(SyntaxFactory.TokenList(SyntaxFactory.Token(SyntaxKind.PublicKeyword)));
+            IEnumerable<SyntaxToken> modifiersWithoutAccessModifiers = newMethodDeclaration.Modifiers.Where(m =>
+                !m.IsKind(SyntaxKind.PrivateKeyword) &&
+                !m.IsKind(SyntaxKind.ProtectedKeyword) &&
+                !m.IsKind(SyntaxKind.InternalKeyword));
+
+            newMethodDeclaration = newMethodDeclaration.WithModifiers(
+                new SyntaxTokenList(SyntaxFactory.Token(SyntaxKind.PublicKeyword))
+                    .AddRange(modifiersWithoutAccessModifiers));
         }
 
         // Ensure the method returns void or Task/ValueTask.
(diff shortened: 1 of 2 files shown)


src/Analyzers/MSTest.Analyzers.CodeFixes/TestMethodShouldBeValidCodeFix.cs
```diff
@@ -93,7 +93,14 @@ private static async Task<Solution> FixTestMethodAsync(Document document, Method
         SyntaxToken publicModifier = newMethodDeclaration.Modifiers.FirstOrDefault(m => m.IsKind(SyntaxKind.PublicKeyword));
         if (publicModifier == default)
         {
-            newMethodDeclaration = newMethodDeclaration.WithModifiers(SyntaxFactory.TokenList(SyntaxFactory.Token(SyntaxKind.PublicKeyword)));
+            IEnumerable<SyntaxToken> modifiersWithoutAccessModifiers = newMethodDeclaration.Modifiers.Where(m =>
+                !m.IsKind(SyntaxKind.PrivateKeyword) &&
+                !m.IsKind(SyntaxKind.ProtectedKeyword) &&
+                !m.IsKind(SyntaxKind.InternalKeyword));
```
Youssef1313, line 99: I think SyntaxGenerator handles it automagically but this will be a bigger rewrite. I'm fine with this for now.

src/Analyzers/MSTest.Analyzers.CodeFixes/TestMethodShouldBeValidCodeFix.cs
```diff
@@ -122,7 +129,7 @@ private static async Task<Solution> FixTestMethodAsync(Document document, Method
         bool asyncModifier = newMethodDeclaration.Modifiers.Any(m => m.IsKind(SyntaxKind.AsyncKeyword));
         if (asyncModifier && newMethodDeclaration.ReturnType != null && newMethodDeclaration.ReturnType.IsVoid())
         {
-            // Change the return type to Task
+            // Change the return type to Task. We have a space after Task to ensure we have a trailing trivia (space) after Task.
             newMethodDeclaration = newMethodDeclaration.WithReturnType(SyntaxFactory.ParseTypeName("Task "));
```
Youssef1313, line 133: Did you try `SyntaxFactory.IdentifierName("Task")` instead?

I also think we have another bug here if the `using` for System.Threading.Tasks isn't present already.

Youssef1313: Feel free to leave that for a future PR though.

src/Analyzers/MSTest.Analyzers.CodeFixes/TestMethodShouldBeValidCodeFix.cs
Evangelink: Thanks I wasn't aware of this API, that's better but as you said we would need to rewrite it all to be consistent, so I am postponing this.

src/Analyzers/MSTest.Analyzers.CodeFixes/TestMethodShouldBeValidCodeFix.cs
Evangelink: Merging as-is and will test and do a follow-up if that works.

Issue #6277, opened 2025-08-01
Revise testhost controller and retry orchestrator handling of running child processes under .NET Framework


Youssef1313:
### Test host controller

First we do

https://github.com/microsoft/testfx/blob/799da019f3ce9d03fdc785e24cf90abcd15df200/src/Platform/Microsoft.Testing.Platform/Hosts/TestHostControllersTestHost.cs#L85-L90

then

https://github.com/microsoft/testfx/blob/799da019f3ce9d03fdc785e24cf90abcd15df200/src/Platform/Microsoft.Testing.Platform/Hosts/TestHostControllersTestHost.cs#L104-L108

then 

https://github.com/microsoft/testfx/blob/799da019f3ce9d03fdc785e24cf90abcd15df200/src/Platform/Microsoft.Testing.Platform/Hosts/TestHostControllersTestHost.cs#L111-L113

This is looking like a bug to me under .NET Framework specifically as this doesn't seem to properly escape the command-line.

### Retry

It's very similar also for retry.

https://github.com/microsoft/testfx/blob/799da019f3ce9d03fdc785e24cf90abcd15df200/src/Platform/Microsoft.Testing.Extensions.Retry/RetryOrchestrator.cs#L102

https://github.com/microsoft/testfx/blob/799da019f3ce9d03fdc785e24cf90abcd15df200/src/Platform/Microsoft.Testing.Extensions.Retry/RetryOrchestrator.cs#L159

https://github.com/microsoft/testfx/blob/799da019f3ce9d03fdc785e24cf90abcd15df200/src/Platform/Microsoft.Testing.Extensions.Retry/RetryOrchestrator.cs#L181-L188

This also doesn't escape under .NET Framework.

In addition, the following are also suspicious.

https://github.com/microsoft/testfx/blob/799da019f3ce9d03fdc785e24cf90abcd15df200/src/Platform/Microsoft.Testing.Extensions.Retry/RetryOrchestrator.cs#L165

https://github.com/microsoft/testfx/blob/799da019f3ce9d03fdc785e24cf90abcd15df200/src/Platform/Microsoft.Testing.Extensions.Retry/RetryOrchestrator.cs#L170

**NOTE:** The reason why the concern is .NET Framework only is that we are using `ArgumentList` on Core. The issue happens only when using the `Arguments` property, not `ArgumentList`.

.NET runtime will already escape `ArgumentList`.

See around:
[...]

PR #6462, opened 2025-08-29
Fix command-line arguments escaping when retry and test host controllers start child processes

Youssef1313:
Fixes https://github.com/microsoft/testfx/issues/6277


Evangelink (approved):
I think it would be worth adding a comment in code about why you need the special handling before/after.

PR #6477, opened 2025-09-01
Don't keep MTP process alive if pipe disconnects

### LINKED ISSUE #6475: Test executable can hang forever when losing pipe connection
If `dotnet.exe` gets killed at the point after MTP sent a request, but before `dotnet.exe` replies, the MTP process will stay alive forever with this state:

Image

Image

Digging further, we are simply stuck in this `while (true)` loop:

https://github.com/microsoft/testfx/blob/9d2e740bc8be46f45c6d69b33be1f5a913bf6231/src/Platform/Microsoft.Testing.Platform/IPC/NamedPipeClient.cs#L138

Image

We should definitely handle when `currentReadBytes` is zero.

Youssef1313:
Fixes #6475

### DIFF
--- a/src/Platform/Microsoft.Testing.Platform/IPC/NamedPipeClient.cs
+++ b/src/Platform/Microsoft.Testing.Platform/IPC/NamedPipeClient.cs
@@ -3,11 +3,11 @@
 
 #if NETCOREAPP
 using System.Buffers;
-
-using Microsoft.Testing.Platform.Helpers;
 #endif
 
 using System.IO.Pipes;
+
+using Microsoft.Testing.Platform.Helpers;
 
 #if NET
 using Microsoft.Testing.Platform.Resources;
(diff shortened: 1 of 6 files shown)


mariam-abdulla (approved):

src/Platform/Microsoft.Testing.Platform/IPC/NamedPipeClient.cs
```diff
@@ -144,6 +151,16 @@ public async Task<TResponse> RequestReplyAsync<TRequest, TResponse>(TRequest req
...
 #endif
 
+                if (currentReadBytes == 0)
+                {
+                    // We are reading a message response.
+                    // If we cannot get a response, there is no way we can recover and continue executing.
+                    // This can happen if the other processes gets killed or crashes while while it's sending the response.
+                    // This is especially important for 'dotnet test', where the user can simply kill the dotnet.exe process themselves.
+                    // In that case, we want the MTP process to also die.
+                    _environment.FailFast("[NamedPipeClient] Connection lost with the other side.");
```
KirillOsenkov, line 161: what does this do? Does this actually call Environment.FailFast()? Beware that this may generate a dump of the current process, if the LocalDumps registry key is enabled:
https://github.com/KirillOsenkov/dotfiles/blob/main/LocalDumps.reg
or Azure Watson might be listening too.

Not sure if this is desired, perhaps just do Environment.Exit()?

PR #6479, opened 2025-09-01
Fix MSTEST0002 codefix to not change modifier order

### LINKED ISSUE #6468: Fix for mstest0002 ignores formatting and swaps modifiers
## Describe the bug
```csharp
namespace NonPublicClass
{
    [TestClass]
    internal sealed class Test1
    {
        [TestMethod]
        public void TestMethod1()
        {
        }

        [TestMethod]
        public void TestMethod2()
        {
        }

        [TestMethod]
        public void TestMethod3()
        {
        }
    }
}
```
class warns on 0002 that is should have correct shape, but the fix looks like this:

Image

which fixes the problem, but makes to code incorrectly formatted and the modifiers are swapped.

version 3.10.3 of mstest.

Youssef1313:
Fixes #6468

### DIFF
--- a/src/Analyzers/MSTest.Analyzers.CodeFixes/TestClassShouldBeValidFixer.cs
+++ b/src/Analyzers/MSTest.Analyzers.CodeFixes/TestClassShouldBeValidFixer.cs
@@ -9,7 +9,6 @@ using Analyzer.Utilities;
 using Microsoft.CodeAnalysis;
 using Microsoft.CodeAnalysis.CodeActions;
 using Microsoft.CodeAnalysis.CodeFixes;
-using Microsoft.CodeAnalysis.CSharp;
 using Microsoft.CodeAnalysis.CSharp.Syntax;
 using Microsoft.CodeAnalysis.Editing;
 using Microsoft.CodeAnalysis.Text;


test/UnitTests/MSTest.Analyzers.UnitTests/TestClassShouldBeValidAnalyzerTests.cs
```diff
@@ -214,7 +215,7 @@ private class {|#0:MyTestClass|}
             public class A
             {
                 [TestClass]
-                internal class MyTestClass
+                public class MyTestClass
```
Evangelink, line 218: Could we extract the change of behavior to a different PR as the target here is to fix the accessibility issue not to change the scope of the rule.

I'm thinking that if I have my full test suite setup with internal, it all works well and I apply the code fix on sln level because of just 1 hit and now it changes the full test suite I could be upset.

It might be worth adding some configuration for that, wdyt?

Evangelink (commented):
LGTM - 1 nitpick but fine if you merge as-is.

test/UnitTests/MSTest.Analyzers.UnitTests/TestClassShouldBeValidAnalyzerTests.cs
Youssef1313: > I'm thinking that if I have my full test suite setup with internal, it all works well and I apply the code fix on sln level because of just 1 hit and now it changes the full test suite I could be upset.

It's not going to change all your codebase to `internal`. Only if the accessibility was completely incorrect (originally it's nested private class in this case), we change it to public.

Evangelink: Oh right I am dumb it only applies to flagged instances. Approving.

Issue #6467, opened 2025-08-29
mstest0001 rule does not work?


nohwnd:
## Describe the bug

I create project from mstest template, comment out the assembly level attribute, and I get no info message or warning from the analyzer.

<img width="1566" height="370" alt="Image" src="https://github.com/user-attachments/assets/4d99fc18-11f5-431e-af31-cedf7e4cd3a9" />

## Steps To Reproduce

```
dotnet new mstest
```
comment out the code in mstestsettings.cs

rebuild, see no info in the error list.

nohwnd:
testted on 3.10.3

Youssef1313:
We probably need to add `IsTestAdapterReferenced` MSBuild property to adapter props, make it a compiler visible property, and read that in the analyzer.

PR #6480, opened 2025-09-01
Fix MSTEST0001 (parallelization analyzer) to work with VSTest

Youssef1313:
Fixes https://github.com/microsoft/testfx/issues/6467


Evangelink (approved):
Thanks @Youssef1313

Youssef1313:
This PR is already part of 3.11.
What probably affects you is that it's now enabled by default. You can turn it off by adding `<NoWarn>$(NoWarn);MSTEST0001</NoWarn>`, or set it off in `.globalconfig`, or alternatively keep it on and address the warning.

jeroenjanssen-cpp:
Can you please add to the v4.0.0 release notes/documentation that it is enabled by default? (so other people also know about this)

It is also not mentioned in the migration guide from v3 to v4 at https://learn.microsoft.com/en-us/dotnet/core/testing/unit-testing-mstest-migration-v3-v4

Youssef1313:
Updating the migration guide in https://github.com/dotnet/docs/pull/48969

PR #6499, opened 2025-09-03
Add --debug-attach option to MTP

### LINKED ISSUE #5531: Add --debug parameter
## Summary

--debug is used in dotnet , to pause execution, until we attach debugger and continue.

## Background and Motivation

When I want to quickly debug a test in exe that is failing I need to open it in VS, and I need to do it via less known approach: open project / solution, select the exe, debug.

An easier, more discoverable approach is to add a documented --debug parameter that will do the following:

Stop execution and print the ID of the process.

Wait for debugger attach, or keypress, to resume execution.

I've been using this with great success in dotnet --debug test etc.

## Proposed Feature

Above.

## Alternative Designs

Using attachVS or similar built in tool. Using env variables, as we are doing now.

Evangelink:
Fixes #5531

I went with `--debug-attach` as it seems more natural and closer to reality than `--debug` to me. This is also aligned with naming on sdk [--debug: attach](https://github.com/dotnet/templating/blob/main/docs/Reserved-Aliases.md)

Another option would be to use `--debug attach` so we could add more debug modes in the future.

### DIFF
--- a/src/Platform/Microsoft.Testing.Platform/Builder/TestApplication.cs
+++ b/src/Platform/Microsoft.Testing.Platform/Builder/TestApplication.cs
@@ -78,6 +78,11 @@ public sealed class TestApplication : ITestApplication
         // First step is to parse the command line from where we get the second input layer.
         // The first one should be the env vars handled autonomously by extensions and part of the test platform.
         CommandLineParseResult parseResult = CommandLineParser.Parse(args, systemEnvironment);
+        if (parseResult.IsOptionSet(PlatformCommandLineProvider.DebugAttachOptionKey))
+        {
+            WaitForDebuggerToAttach(systemEnvironment, systemConsole, systemProcess);
+        }
+
         TestHostControllerInfo testHostControllerInfo = new(parseResult);
         CurrentTestApplicationModuleInfo testApplicationModuleInfo = new(systemEnvironment, systemProcess);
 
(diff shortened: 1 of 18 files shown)


src/Platform/Microsoft.Testing.Platform/Builder/TestApplication.cs
```diff
@@ -78,6 +78,11 @@ public static async Task<ITestApplicationBuilder> CreateBuilderAsync(string[] ar
         // First step is to parse the command line from where we get the second input layer.
         // The first one should be the env vars handled autonomously by extensions and part of the test platform.
         CommandLineParseResult parseResult = CommandLineParser.Parse(args, systemEnvironment);
+        if (parseResult.IsOptionSet(PlatformCommandLineProvider.DebugAttachOptionKey))
```
Youssef1313, line 81: On a first glance, I would interpret `--debug-attach` as "it attaches the debugger" (i.e, it does `Debugger.Launch`) not that "it waits for the debugger to be attached"

Evangelink: `--debug-wait-attach`?

Youssef1313: Works for me

Youssef1313 (approved):

nohwnd:
Would you reconsider the name to be --debug? That is what dotnet sdk uses, even though they just emit it in debug builds.

https://github.com/dotnet/sdk/blob/main/src/Cli/dotnet/Extensions/ParseResultExtensions.cs#L232-L239

--debug:attach is used just by templating engine.

Also the I've grown to like the implementation on msbuild that does readline, and unblocks even if there is no debugger. Makes it nicer when you run with the parameter / env variable by accident and just want to continue. 

On the other hand, it is annoying to have to press the key to unblock, so imho awaiting the debugger or keypress would be the best.

https://github.com/dotnet/msbuild/blob/53a2d6ba9edc932e724dc80a5a56eeb9943863cf/src/MSBuild/XMake.cs#L589-L605

PR #6536, opened 2025-09-08
Avoid long path issues for pipes

### LINKED ISSUE #6535: Potential long pipe name issues
See https://github.com/xunit/xunit/issues/3397

This might be affecting the new `dotnet test` for .NET 10 as well!

FYI @mariam-abdulla

Youssef1313:
Fixes #6535

### DIFF
--- a/src/Platform/Microsoft.Testing.Platform/IPC/NamedPipeServer.cs
+++ b/src/Platform/Microsoft.Testing.Platform/IPC/NamedPipeServer.cs
@@ -22,6 +22,8 @@ internal sealed class NamedPipeServer : NamedPipeBase, IServer
         ;
 #pragma warning restore CA1416 // Validate platform compatibility
 
+    private static bool IsUnix => Path.DirectorySeparatorChar == '/';
+
     private readonly Func<IRequest, Task<IResponse>> _callback;
     private readonly IEnvironment _environment;
     private readonly NamedPipeServerStream _namedPipeServerStream;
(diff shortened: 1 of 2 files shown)


src/Platform/Microsoft.Testing.Platform/IPC/PipeNameDescription.cs
```diff
@@ -3,32 +3,11 @@
...
-            {
-                Directory.Delete(Path.GetDirectoryName(Name)!, true);
-            }
-            catch (IOException)
-            {
-                // This folder is created inside the temp directory and will be cleaned up eventually by the OS
-            }
-        }
-
-        _disposed = true;
```
Youssef1313, line 32: What we were doing here is:

1. We used to create a subdirectory ourselves, called ".p" (see `NamedPipeServer.cs` changes)
2. Then when disposing, we used to get the parent directory of `.p` and delete it.

I think all this might have been extra unnecessary complexity. When looking through the history in dotnet/runtime, the closest related change I found is https://github.com/dotnet/runtime/commit/f10e8356434047e8f63e039a46467bee6a402f1e (see commit description).

@stephentoub Would you know if all of that is related to some history on BCL implementation that's no longer applicable today (potentially linked to the commit I pointed to above)?

To give more context on what we do, in many cases we need communication across test processes. For example, when running tests with hang dump enabled, one process (we call test host controller) observes another process (we call test host, which actually runs the tests), and there is some inter-process communication going between those two processes via named pipes.

stephentoub: > @stephentoub Would you know if all of that is related to some history on BCL implementation that's no longer applicable today (potentially linked to the commit I pointed to above)?

Sorry, I'm not quite following... what is the question?

Youssef1313: @stephentoub When working with named pipes on Unix, do we need to manually delete any files/directories today? Was it ever needed in the past in old .NET versions?

Our current implementation used to `Path.Combine(path_dependent_on_pipe_name, ".p")`. Then we get the parent directory of that (which is path_dependent_on_pipe_name) and we delete it manually.

stephentoub: No. Though there may be some tmp files left behind, e.g. if the process crashes, and you rely on the OS to clean up tmp as is the purpose of tmp.

Youssef1313: Perfect. Thanks @stephentoub for the confirmation.

Evangelink (approved):

src/Platform/Microsoft.Testing.Platform/IPC/NamedPipeServer.cs
```diff
@@ -22,6 +22,8 @@ internal sealed class NamedPipeServer : NamedPipeBase, IServer
         ;
 #pragma warning restore CA1416 // Validate platform compatibility
 
+    private static bool IsUnix => Path.DirectorySeparatorChar == '/';
```
drognanar, line 25: why have we moved to a simple separator check, rather than RuntimeInformation.IsOSPlatform?

Youssef1313: I preferred to follow what Roslyn does which we know works very well for them.
But it's probably equivalent.

DirectorySeparatorChar on dotnet/runtime is hard coded on Windows to `\` and on Unix to `/`.

PR #4324, opened 2024-12-12
Create draft proposal for running dotnet test using Microsoft.Testing.Platform

mariam-abdulla:

PR #6552, opened 2025-09-10
Remove failfast in named pipe client

### LINKED ISSUE #6510: Revise Environment.FailFast usage in pipe implementation
what does this do? Does this actually call Environment.FailFast()? Beware that this may generate a dump of the current process, if the LocalDumps registry key is enabled:
https://github.com/KirillOsenkov/dotfiles/blob/main/LocalDumps.reg
or Azure Watson might be listening too.

Not sure if this is desired, perhaps just do Environment.Exit()?

_Originally posted by @KirillOsenkov in https://github.com/microsoft/testfx/pull/6477#discussion_r2320009406_

mariam-abdulla:
This pull request introduces an RFC document proposing support for running `dotnet test` using the Microsoft Testing Platform, and updates the IPC error handling logic in `NamedPipeClient`. The RFC outlines a configurable approach that maintains backward compatibility, while the code change ensures more controlled termination in case of IPC failures.

**SDK Testing Platform RFC:**

* Added `docs/RFCs/011-Sdk-Testing-Platform.md` describing a plan to allow users to opt-in to the Microsoft Testing Platform via a `global.json` configuration, with clear defaults and backward compatibility.

**Error Handling Improvement:**

* Updated `NamedPipeClient.RequestReplyAsync` to use `_environment.Exit(1)` for IPC connection loss instead of `_environment.FailFast`, ensuring the process exits with a specific code for abnormal termination.

Fixes #6510


src/Platform/Microsoft.Testing.Platform/IPC/NamedPipeClient.cs
```diff
@@ -164,7 +164,8 @@ public async Task<TResponse> RequestReplyAsync<TRequest, TResponse>(TRequest req
                     // This can happen if the other processes gets killed or crashes while while it's sending the response.
                     // This is especially important for 'dotnet test', where the user can simply kill the dotnet.exe process themselves.
                     // In that case, we want the MTP process to also die.
-                    _environment.FailFast("[NamedPipeClient] Connection lost with the other side.");
+                    // Exit code 1 indicates abnormal termination due to IPC connection loss.
+                    _environment.Exit(1);
```
Youssef1313: I'm wondering if we should have a special exit code for that in https://github.com/microsoft/testfx/blob/main/src/Platform/Microsoft.Testing.Platform/Helpers/ExitCodes.cs ?

Or, if we won't have a special exit code, we can use ExitCodes.GenericFailure instead of hardcoding its value (1)

docs/RFCs/011-Sdk-Testing-Platform.md
```diff
@@ -0,0 +1,41 @@
+# RFC 011 - Run dotnet test with Microsoft.Testing.Platform
```
Youssef1313: It seems this was what you worked on in https://github.com/microsoft/testfx/pull/4324. I think it can be deleted for the purpose of this PR.

PR #6564, opened 2025-09-14
Include parameter types in discovered test message

Youssef1313:


Evangelink (commented):
I wonder if the PR should be moved to `rel/4.0` as there is breaking change in this message.

Evangelink (approved):
My bad I thought it was the identifier property not the IPC msg. LGTM

Issue #4187, opened 2024-11-27
Collect all children processes dump in case of hang


Evangelink:
From @MarcoRossignoli:

Current extension takes the dump only for the test host process. If hang of it is related to children hang looks complex understand the reason (we have only the tests that's hanging).
We should try to dump all children too.

nohwnd:
vstest does this and the code is here: https://github.com/microsoft/vstest/blob/main/src/Microsoft.TestPlatform.Extensions.BlameDataCollector/WindowsHangDumper.cs#L38 

https://github.com/microsoft/vstest/blob/main/src/Microsoft.TestPlatform.Extensions.BlameDataCollector/NetClientHangDumper.cs#L18

using minidump write dump on windows is preferable because it gives dumps also of processes that are not .net.

PR #6574, opened 2025-09-18
Dump child process on hang

nohwnd:
Collect tree of processes and dump them from the bottom up, skipping werFault (dump reporter on windows) and conhost (console host on windows). 

Report all dumps that were collected, together with the tree of processes that we will dump.

Reduce verbosity of the dumper, to not show thousands of lines per dump, seeing memory addresses etc. is not usually useful, and makes it really hard to see what actually happened when stuff goes wrong, e.g. when the process completes after the hang timeout, and so we cannot dump it anymore. (We cannot suspend the processes while dumping them - on .net core dumper it prevents the dumper from working, on .net framework, we don't have documented api to do this, and so we get flagged by compliance).

The way dumps are created depends on the testhost process, if it is .NET it will use the .NET dumper and will assume all the children are .NET.

If it if .NET Framework it will use windows minidump and will try to dump all children using that. Limitation of windows minidump is that it can only produce usable dumps when the dumping process (testhost controller) is the same architecture as the dumped process (testhost). In other cases it will produce a dump that is not usable, and we cannot work around it without shipping additional executables, which would broke the premise of MTP, that all stuff is done via the same process.

Fix https://github.com/microsoft/testfx/issues/4187

### DIFF
--- a/src/Platform/Microsoft.Testing.Extensions.HangDump/HangDumpProcessLifetimeHandler.cs
+++ b/src/Platform/Microsoft.Testing.Extensions.HangDump/HangDumpProcessLifetimeHandler.cs
@@ -194,7 +194,7 @@ internal sealed class HangDumpProcessLifetimeHandler : ITestHostProcessLifetimeH
                 {
                     try
                     {
-                        await ActivityTimerAsync().ConfigureAwait(false);
+                        await ActivityTimerAsync(cancellation).ConfigureAwait(false);
                     }
                     catch (Exception e)
                     {
@@ -246,11 +246,11 @@ internal sealed class HangDumpProcessLifetimeHandler : ITestHostProcessLifetimeH
         }
     }
 
-    private async Task ActivityTimerAsync()
+    private async Task ActivityTimerAsync(CancellationToken cancellationToken)
     {
         _logger.LogDebug("Wait for mutex name from the test host");
 
-        if (!_mutexNameReceived.Wait(TimeoutHelper.DefaultHangTimeSpanTimeout))
+        if (!_mutexNameReceived.Wait(TimeoutHelper.DefaultHangTimeSpanTimeout, cancellationToken))
         {
             throw new InvalidOperationException(string.Format(CultureInfo.InvariantCulture, ExtensionResources.MutexNameReceptionTimeoutErrorMessage, TimeoutHelper.DefaultHangTimeoutSeconds));
         }
@@ -338,20 +338,19 @@ internal sealed class HangDumpProcessLifetimeHandler : ITestHostProcessLifetimeH
 
         if (timeoutFired)
         {
-            await TakeDumpOfTreeAsync().ConfigureAwait(false);
-            // await TakeDumpAsync().ConfigureAwait(false);
+            await TakeDumpOfTreeAsync(cancellationToken).ConfigureAwait(false);
         }
     }
 
-    private async Task TakeDumpOfTreeAsync()
+    private async Task TakeDumpOfTreeAsync(CancellationToken cancellationToken)
     {
         ApplicationStateGuard.Ensure(_testHostProcessInformation is not null);
 
         await _logger.LogInformationAsync($"Hang dump timeout({_activityTimerValue}) expired.").ConfigureAwait(false);
         await _outputDisplay.DisplayAsync(new ErrorMessageOutputDeviceData(string.Format(CultureInfo.InvariantCulture, ExtensionResources.HangDumpTimeoutExpired, _activityTimerValue))).ConfigureAwait(false);
 
-        IProcess process = _processHandler.GetProcessById(_testHostProcessInformation.PID);
-        var processTree = (await process.GetProcessTreeAsync(_logger, _outputDisplay).ConfigureAwait(false)).Where(p => p.Process?.Name is not null and not "conhost" and not "WerFault").ToList();
+        using IProcess process = _processHandler.GetProcessById(_testHostProcessInformation.PID);
+        var processTree = (await process.GetProcessTreeAsync(_logger, _outputDisplay, cancellationToken).ConfigureAwait(false)).Where(p => p.Process?.Name is not null and not "conhost" and not "WerFault").ToList();
         IEnumerable<IProcess> bottomUpTree = processTree.OrderByDescending(t => t.Level).Select(t => t.Process).OfType<IProcess>();
 
         try
--- a/src/Platform/Microsoft.Testing.Extensions.HangDump/HangDumpProcessLifetimeHandler.cs
+++ b/src/Platform/Microsoft.Testing.Extensions.HangDump/HangDumpProcessLifetimeHandler.cs
@@ -427,7 +427,9 @@ internal sealed class HangDumpProcessLifetimeHandler : ITestHostProcessLifetimeH
 
         try
         {
+
 #if NETCOREAPP
+            try {
             DiagnosticsClient diagnosticsClient = new(process.Id);
             DumpType dumpType = _dumpType.ToLowerInvariant().Trim() switch
             {
@@ -446,6 +448,11 @@ internal sealed class HangDumpProcessLifetimeHandler : ITestHostProcessLifetimeH
             }
 
             diagnosticsClient.WriteDump(dumpType, finalDumpFileName, true);
+            }
+            catch  {
+                // this is problematic, when processes depend on each other, some of them will inevitably close before others and exit when we are dumping them
+                // on windows we could possibly pause them, but there is no supported api for that that we are allowed to use
+            }
 #else
             MiniDumpWriteDump.MiniDumpTypeOption miniDumpTypeOption = _dumpType.ToLowerInvariant().Trim() switch
             {
@@ -468,7 +475,7 @@ internal sealed class HangDumpProcessLifetimeHandler : ITestHostProcessLifetimeH
             {
                 try
                 {
-                    process.Kill();
+                    process.KillWithoutKillingChildProcesses();
                 }
                 catch (ArgumentException)
                 {
--- a/src/Platform/Microsoft.Testing.Extensions.HangDump/HangDumpProcessLifetimeHandler.cs
+++ b/src/Platform/Microsoft.Testing.Extensions.HangDump/HangDumpProcessLifetimeHandler.cs
@@ -351,7 +351,7 @@ internal sealed class HangDumpProcessLifetimeHandler : ITestHostProcessLifetimeH
         await _outputDisplay.DisplayAsync(this, new ErrorMessageOutputDeviceData(string.Format(CultureInfo.InvariantCulture, ExtensionResources.HangDumpTimeoutExpired, _activityTimerValue))).ConfigureAwait(false);
 
         IProcess process = _processHandler.GetProcessById(_testHostProcessInformation.PID);
-        var processTree = process.GetProcessTree().Where(p => p.Process?.Name is not null and not "conhost" and not "WerFault").ToList();
+        var processTree = process.GetProcessTree(_logger).Where(p => p.Process?.Name is not null and not "conhost" and not "WerFault").ToList();
         IEnumerable<IProcess> bottomUpTree = processTree.OrderByDescending(t => t.Level).Select(t => t.Process).OfType<IProcess>();
 
         try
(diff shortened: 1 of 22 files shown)


src/Platform/Microsoft.Testing.Extensions.HangDump/HangDumpProcessLifetimeHandler.cs
```diff
@@ -340,19 +338,98 @@ private async Task ActivityTimerAsync()
 
         if (timeoutFired)
         {
-            await TakeDumpAsync().ConfigureAwait(false);
+            await TakeDumpOfTreeAsync().ConfigureAwait(false);
+            // await TakeDumpAsync().ConfigureAwait(false);
```
Youssef1313: Remove comment?

src/Platform/Microsoft.Testing.Extensions.HangDump/HangDumpProcessLifetimeHandler.cs
```diff
@@ -340,19 +338,98 @@ private async Task ActivityTimerAsync()
...
+    private async Task TakeDumpOfTreeAsync()
     {
         ApplicationStateGuard.Ensure(_testHostProcessInformation is not null);
-        ApplicationStateGuard.Ensure(_dumpType is not null);
 
         await _logger.LogInformationAsync($"Hang dump timeout({_activityTimerValue}) expired.").ConfigureAwait(false);
-        await _outputDisplay.DisplayAsync(this, new ErrorMessageOutputDeviceData(string.Format(CultureInfo.InvariantCulture, ExtensionResources.HangDumpTimeoutExpired, _activityTimerValue))).ConfigureAwait(false);
+        await _outputDisplay.DisplayAsync(new ErrorMessageOutputDeviceData(string.Format(CultureInfo.InvariantCulture, ExtensionResources.HangDumpTimeoutExpired, _activityTimerValue))).ConfigureAwait(false);
+
+        IProcess process = _processHandler.GetProcessById(_testHostProcessInformation.PID);
```
Youssef1313: Do we need to dispose this?

src/Platform/Microsoft.Testing.Extensions.HangDump/Helpers/IProcessExtensions.cs
```diff
@@ -0,0 +1,187 @@
...
+            .ToList();
+
+        var acc = new List<ProcessTreeNode>();
+        foreach (Process c in childProcesses)
+        {
+            try
+            {
+                int parentId = await GetParentPidAsync(c, logger, outputDisplay).ConfigureAwait(false);
+
+                // c.ParentId = parentId;
```
Youssef1313: Delete comment?

src/Platform/Microsoft.Testing.Extensions.HangDump/Helpers/IProcessExtensions.cs
```diff
@@ -0,0 +1,187 @@
...
+                acc.Add(new ProcessTreeNode { ParentId = parentId, Process = new SystemProcess(c) });
+            }
+            catch (Exception e)
+            {
+                logger.LogError($"Failed to get parent for process {c.Id} - {c.ProcessName}", e);
+                await outputDisplay.DisplayAsync(new ErrorMessageOutputDeviceData($"Failed to get parent for process {c.Id} - {c.ProcessName}, {e}.")).ConfigureAwait(false);
```
Youssef1313: This needs to be localized as it's user-facing.

src/Platform/Microsoft.Testing.Extensions.HangDump/Helpers/IProcessExtensions.cs
```diff
@@ -0,0 +1,187 @@
...
+            await Task.Run(ps.WaitForExit, new CancellationTokenSource(timeout).Token).ConfigureAwait(false);
+        }
+#endif
+
+        string o = output.ToString();
+        string e = err.ToString();
+
+        int parent = int.TryParse(o.Trim(), out int ppid) ? ppid : InvalidProcessId;
+
+        if (err.ToString() is string error && !RoslynString.IsNullOrWhiteSpace(error))
```
Youssef1313:
```suggestion
        if (!RoslynString.IsNullOrWhiteSpace(e))
```

src/Platform/Microsoft.Testing.Platform/Helpers/System/SystemProcess.cs
```diff
@@ -38,8 +40,14 @@ public void WaitForExit()
     public Task WaitForExitAsync()
         => _process.WaitForExitAsync();
 
+    /// <summary>
+    /// Kills the process and all child processes.
+    /// </summary>
     public void Kill()
-        => _process.Kill(true);
+        => _process.Kill(entireProcessTree: true);
```
Youssef1313, line 44: This is not guaranteed that it kills the entire tree. On TFMs where the API isn't available, we get this polyfilled where the bool is ignored.

https://github.com/SimonCropp/Polyfill/blob/54ade056ec5059665d12e8cfaba964f94484dcb0/src/Polyfill/Polyfill_Process.cs#L24-L25

One reason I prefer explicit `#if`s over polyfills in certain cases..

Not sure if this is an *ok* behavior for your needs in hang dump, but thought it's worth pointing it out. At least the doc comment needs to be less misleading.

src/Platform/Microsoft.Testing.Platform/Helpers/System/SystemProcess.cs
```diff
@@ -38,8 +40,14 @@ public void WaitForExit()
...
+
+    public void KillWithoutKillingChildProcesses()
+        => _process.Kill(entireProcessTree: false);
```
nohwnd:
```suggestion
```

src/Platform/Microsoft.Testing.Extensions.HangDump/Helpers/IProcessExtensions.cs
```diff
@@ -0,0 +1,187 @@
...
+namespace Microsoft.Testing.Extensions.Diagnostics.Helpers;
+
+/// <summary>
+/// Helper functions for IProcess.
+/// </summary>
+internal static class IProcessExtensions
+{
+    private const int InvalidProcessId = -1;
+
+    public static async Task<List<ProcessTreeNode>> GetProcessTreeAsync(this IProcess process, ILogger logger, OutputDeviceWriter outputDisplay)
```
fhnaseer: Add CancellationToken?

src/Platform/Microsoft.Testing.Platform/Helpers/System/SystemProcess.cs
nohwnd: 👍rolling back the whole file, we don't rely on the tree being killed, and I did not use the api I added here.

nohwnd: wrong, I do need the start time. So rolling back to just that.

src/Platform/Microsoft.Testing.Extensions.HangDump/HangDumpProcessLifetimeHandler.cs
nohwnd: disposed

src/Platform/Microsoft.Testing.Extensions.HangDump/Helpers/IProcessExtensions.cs
```diff
@@ -0,0 +1,187 @@
...
+    }
+
+    /// <summary>
+    /// Returns the parent id of a process or -1 if it fails.
+    /// </summary>
+    /// <param name="process">The process to find parent of.</param>
+    /// <param name="logger">The logger.</param>
+    /// <param name="outputDisplay">The output display.git.</param>
+    /// <returns>The pid of the parent process.</returns>
+    internal static async Task<int> GetParentPidAsync(Process process, ILogger logger, OutputDeviceWriter outputDisplay)
```
fhnaseer, line 57: Suggestion: Make these methods (GetPid*) non async and the caller method creates a task instead and `awaits` for it.

nohwnd: why should I do that?

nohwnd: It all already runs as part of long running task, so the less threads we involve into this is probably better?

fhnaseer: It seems like these methods are `async` just for the sake of it and eventually do `Task.FromResult()`. Unless they are used in other classes, move `Task.FromResult()` to the top.

nohwnd: it is the output display in GetParentPidMacOsAsync that forces it to be async (to be aligned). these methods are not used in other classes, and will complete synchronously. Imho it is okay as is.

src/Platform/Microsoft.Testing.Extensions.HangDump/Helpers/IProcessExtensions.cs
```diff
@@ -0,0 +1,188 @@
...
+/// <summary>
+/// Helper functions for IProcess.
+/// </summary>
+internal static class IProcessExtensions
+{
+    private const int InvalidProcessId = -1;
+
+    public static async Task<List<ProcessTreeNode>> GetProcessTreeAsync(this IProcess process, ILogger logger, OutputDeviceWriter outputDisplay, CancellationToken cancellationToken)
+    {
+        var childProcesses = Process.GetProcesses()
```
Youssef1313, line 21: Should we use the `IProcess` here to get the processes instead of using the actual `Process` implementation?

nohwnd: https://github.com/microsoft/testfx/blob/1406f4d1f1831b11dd0b5e00be540b2007614425/src/Platform/Microsoft.Testing.Platform/Helpers/System/IProcessHandler.cs#L6 implementation of this interface you mean?

src/Platform/Microsoft.Testing.Extensions.HangDump/Helpers/IProcessExtensions.cs
```diff
@@ -0,0 +1,188 @@
...
+            : RuntimeInformation.IsOSPlatform(OSPlatform.Linux)
+                ? await GetParentPidLinuxAsync(process).ConfigureAwait(false)
+                : RuntimeInformation.IsOSPlatform(OSPlatform.OSX) ?
+                    await GetParentPidMacOsAsync(process, logger, outputDisplay).ConfigureAwait(false)
+                    : throw new PlatformNotSupportedException();
+
+    internal static Task<int> GetParentPidWindowsAsync(Process process)
+    {
+        IntPtr handle = process.Handle;
+        int res = NtQueryInformationProcess(handle, 0, out ProcessBasicInformation pbi, Marshal.SizeOf<ProcessBasicInformation>(), out int _);
```
Youssef1313, line 69: Another approach used by Roslyn is:

https://github.com/dotnet/roslyn/blob/9b48f158ea39c63cd3be662993c7d0dc978bacf1/src/Tools/Source/RunTests/ProcessUtil.cs#L23-L32

But I'm not sure what is best.

The same linked class also seems to have helpers on getting the process tree

nohwnd: I'd guess the native call we do is faster and better.

src/Platform/Microsoft.Testing.Extensions.HangDump/Helpers/IProcessExtensions.cs
Youssef1313: Yup

Youssef1313: Simply similar to other static methods of `Process` that we abstract away with that interface. It seems like we have been consistently using easy-to-mock abstractions, even if we are not currently unit testing or mocking it. So I would love to be more consistent.

src/Platform/Microsoft.Testing.Extensions.HangDump/Helpers/IProcessExtensions.cs
nohwnd: roslyn thing is also more incorrect in not checking the start time of the process, as PIDs are reused (were more reused on win10 an older where the numbering range was one digit smaller), and so we should only considered processes started after us.

And only looks one level deeper (direct children) while we look for all children to make a full tree.

src/Platform/Microsoft.Testing.Extensions.HangDump/Helpers/IProcessExtensions.cs
nohwnd: Seems fine, we test that we are getting the children process dumps, we don't need to unit test this, which is the main reason to have those abstractions over everything,.

PR #6582, opened 2025-09-20
More async [WIP]

Youssef1313:


nohwnd:
Please add some description on what is this PR solving, and if there are any risks :)

nohwnd:
You most likely did a lot of thinking here, and it is a waste to lose all that context.

Youssef1313:
> > Also it's a discouraged practice to do GetAwaiter().GetResult() so good anyway to get rid of.
> 
> Where is that happening? I see about 10 places where this is added, not removed.

The places where it's added now were already doing it before but more deeper in our code.

Evangelink (approved):

PR #6588, opened 2025-09-22
Fix typo in protocol introduction

nohwnd:


docs/mstest-runner-protocol/001-protocol-intro.md
```diff
@@ -1,7 +1,7 @@
 # 001 - MSTest Runner Protocol
 
 MSTest Runner projects builds into a self-contained executable that can be invoked to run tests.
-The protocol describes the communication between the client (IDE/CLI/CI) any the MSTest Runner executable (also refered to as the server).
+The protocol describes the communication between the client (IDE/CLI/CI) and the MSTest Runner executable (also refered to as the server).
```
Youssef1313, line 4: As MTP is already available for all frameworks now. I think we should also get rid of "mstest runner" naming when we actually refer to MTP.

Youssef1313 (approved):

docs/mstest-runner-protocol/001-protocol-intro.md
nohwnd: k, will make different PR.

PR #6597, opened 2025-09-24
Follow-up on CollectionAssert change

Youssef1313:
Follow-up to #6292. We lost the user message in that PR.

### DIFF
--- a/src/TestFramework/TestFramework/Resources/FrameworkMessages.resx
+++ b/src/TestFramework/TestFramework/Resources/FrameworkMessages.resx
@@ -371,6 +371,6 @@ Actual: {2}</value>
     <value>CollectionAssert.ReferenceEquals should not be used for Assertions. Please use CollectionAssert methods or Assert.AreSame &amp; overloads instead.</value>
   </data>
   <data name="ReturnedSubsetValueMessage" xml:space="preserve">
-    <value>Element(s) &lt;{0}&gt; is/are not present in the collection</value>
+    <value>Element(s) &lt;{0}&gt; is/are not present in the collection.</value>
   </data>
 </root>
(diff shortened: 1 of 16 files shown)


src/TestFramework/TestFramework/Resources/FrameworkMessages.resx
```diff
@@ -371,6 +371,6 @@ Actual: {2}</value>
     <value>CollectionAssert.ReferenceEquals should not be used for Assertions. Please use CollectionAssert methods or Assert.AreSame &amp; overloads instead.</value>
   </data>
   <data name="ReturnedSubsetValueMessage" xml:space="preserve">
-    <value>Element(s) &lt;{0}&gt; is/are not present in the collection</value>
+    <value>Element(s) &lt;{0}&gt; is/are not present in the collection.</value>
```
Evangelink, line 374: I'm not sure but I wonder if the enclosing message isn't already adding the final "."

Evangelink (approved):
It would be great to have one test impacted by the change of the message.

src/TestFramework/TestFramework/Resources/FrameworkMessages.resx
Youssef1313: This message replaces the `{1}` place holder in:

https://github.com/microsoft/testfx/blob/9086d5ea5fce06fdd36dc44e5c84d282783d79f8/src/TestFramework/TestFramework/Resources/FrameworkMessages.resx#L203-L205

Youssef1313:
@Evangelink We already have tests that should fail. So far the Linux and MacOS jobs are green which is suspicious as this seem to indicate that assertion tests might be run on Windows only (`TestFramework.UnitTests` project specifically - not sure if this was ever intended?)

Evangelink (approved):

PR #6641, opened 2025-10-01
Run mark-shipped

Youssef1313:


src/Adapter/MSTest.Engine/PublicAPI/net/PublicAPI.Shipped.txt
```diff
@@ -1 +1 @@
-#nullable enable
```
nohwnd, line 1: You probably run the script from windows powershell, and not from modern powershell, which adds bom, because old powershell utf8 means **with bom**, and in new utf8 means **without bom**.

So quickfix to not add more changes than necessary, run it from modern powershell.

nohwnd: https://github.com/microsoft/testfx/pull/6642

nohwnd (approved):

PR #6642, opened 2025-10-01
Add requires to script to make it work constistently

nohwnd:
Utf8 encoding means utf8 with BOM in the old windows powershell, modern powershell makes that mean utf8 without BOM.

Require powershell 7 to run the script.


Youssef1313:
Thanks @nohwnd. I installed PowerShell 7 and the diff is much easier to see now.
Also it looks like the sorting is different in some cases between Windows PowerShell and PowerShell 7 :)

PR #6677, opened 2025-10-06
Filter Extensibility

### LINKED ISSUE #3590: [Testing Platform] Open up filters for custom implementations

thomhurst:
Fixes #3590


Youssef1313 (changes requested):
Thanks for the PR @thomhurst 

I recall last time I discussed this with @Evangelink, we had concerns/doubts/questions about exposing filters for public extensibility, and we were unsure about how the final design shape should look like.

This will need more thinking and design work on our side.

PR #6699, opened 2025-10-08
Fix Assert.That to not evaluate expression twice

### LINKED ISSUE #6690: Assert.That evaluates expression twice
## Describe the bug

Assert.That evaluates the expression twice (or more times) to extract the values. This can lead to confusing behavior when any of the parts have state or are dynamic / random.

## Steps To Reproduce

```csharp

namespace mstest280;

[TestClass]
public sealed class Test2
{
    [TestMethod]
    public void TestMethod1()
    {
        var box = new Box();
        var expected = 2;
        Assert.That(() => box.GetNumber() == expected);
    }
}

public class Box
{
    int _c = 0;
    public int GetNumber()
    {
        _c++;
        return _c;
    }
}
```

Will fail the assertion, but will report that the values are the same:

Image

This is happening because evaluation happens once to figure out the result of the whole expression:

https://github.com/microsoft/testfx/blob/adb8914cbdedbc243350a4c0ae7b50251df53281/src/TestFramework/TestFramework/Assertions/Assert.That.cs#L26

And then it happens again when getting the details of each detail within the expression:

https://github.com/microsoft/testfx/blob/adb8914cbdedbc243350a4c0ae7b50251df53281/src/TestFramework/TestFramework/Assertions/Assert.That.cs#L768

To solve this problem it needs to happen in one step, invoking while collecting the values.

## Expected behavior

It is correctly reported that GetNumber returned 1 when the  == evaluation happened.

## Actual behavior

It is incorrectly reported that GetNumber returned 2, when the == evaluation happened, showing the user a message that the values were the same.

Evangelink:
Fixes #6690

### DIFF
--- a/src/TestFramework/TestFramework/Assertions/Assert.That.cs
+++ b/src/TestFramework/TestFramework/Assertions/Assert.That.cs
@@ -10,6 +10,29 @@ namespace Microsoft.VisualStudio.TestTools.UnitTesting;
 /// </summary>
 public static partial class AssertExtensions
 {
+    // Constants for standardized display values
+    private const string FailedToEvaluate = "<Failed to evaluate>";
+    private const string NullDisplay = "null";
+    private const string NullAngleBrackets = "<null>";
+
+    // Constants for indexer method names
+    private const string GetItemMethodName = "get_Item";
+    private const string GetMethodName = "Get";
+
+    // Constants for compiler-generated patterns
+    private const string AnonymousTypePrefix = "<>f__AnonymousType";
+    private const string ValueWrapperPattern = "value(";
+    private const string ArrayLengthWrapperPattern = "ArrayLength(";
+    private const string NewKeyword = "new ";
+    private const string ActionTypePrefix = "Action`";
+    private const string FuncTypePrefix = "Func`";
+
+    // Constants for collection type patterns
+    private const string ListInitPattern = "`1()";
+
+    // Constants for parenthesis limits
+    private const int MaxConsecutiveParentheses = 2;
+
     /// <summary>
     /// Provides That extension to Assert class.
     /// </summary>
@@ -32,9 +55,12 @@ public static partial class AssertExtensions
                 throw new ArgumentNullException(nameof(condition));
             }
 
-            // Cache to store evaluated expression values to avoid re-evaluation
+            // Cache to store evaluated expression values to avoid re-evaluation.
+            // This is critical for expressions with side effects - we evaluate each sub-expression
+            // only once and reuse the cached result throughout the assertion process.
             var evaluationCache = new Dictionary<Expression, object?>();
 
+            // Evaluate the condition expression and cache all sub-expression values
             bool result = EvaluateExpression(condition.Body, evaluationCache);
 
             if (result)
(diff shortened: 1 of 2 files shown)


src/TestFramework/TestFramework/Assertions/Assert.That.cs
```diff
@@ -10,6 +10,29 @@ namespace Microsoft.VisualStudio.TestTools.UnitTesting;
 /// </summary>
 public static partial class AssertExtensions
 {
+    // Constants for standardized display values
+    private const string FailedToEvaluate = "<Failed to evaluate>";
```
Youssef1313: Do we want this to be localized?

src/TestFramework/TestFramework/Assertions/Assert.That.cs
```diff
@@ -56,17 +88,238 @@ public static void That(Expression<Func<bool>> condition, string? message = null
...
+        EvaluateAllSubExpressions(expr, cache);
+
+        // The root expression should now be cached
+        if (cache.TryGetValue(expr, out object? result))
+        {
+            return (bool)result!;
+        }
+
+        // Fallback - this should not happen if EvaluateAllSubExpressions works correctly
+        return false;
```
nohwnd: Consider throwing here if we should not reach it.

src/TestFramework/TestFramework/Assertions/Assert.That.cs
```diff
@@ -610,6 +966,10 @@ private static string CleanTypeName(string typeName)
             _ => typeName,
         };
 
+    /// <summary>
+    /// Removes parentheses that wrap the entire expression and cleans up excessive
+    /// consecutive parentheses (e.g., "(((x)))" becomes "x", "((x) && (y))" becomes "(x) && (y)").
```
Evangelink:
```suggestion
    /// consecutive parentheses (e.g., "(((x)))" becomes "x", "((x) &amp;&amp; (y))" becomes "(x) &amp;&amp; (y)").

```

src/TestFramework/TestFramework/Assertions/Assert.That.cs
Evangelink: Fine by me. FYI it wasn't localized before.

test/UnitTests/TestFramework.UnitTests/Assertions/AssertTests.That.cs
```diff
@@ -1042,4 +1042,50 @@ public void That_WithNullConstantAndVariable_OnlyIncludesVariable()
...
+        var box = new Box();
+
+        // Compare to 0 to force failure.
+        Action act = () => Assert.That(() => box.GetNumber() + box.GetNumber() == 0);
+
+        act.Should().Throw<AssertFailedException>()
+            .WithMessage("""
+            Assert.That(() => box.GetNumber() + box.GetNumber() == 0) failed.
+            Details:
+              box.GetNumber() = 1
```
nohwnd: This test passes but imho should not, we get no information about the second call to box.GetNumber, leaving us stumped why it failed. Especially if we use expected = 1, not super easy to debug :)


Image

PR #6720, opened 2025-10-10
Analyzer implementation to enhance assertions

### LINKED ISSUE #6360: MSTEST0037 analyzer should suggest `Contains`/`DoesNotContain`/`ContainsSingle` for overloads taking a predicate
## Summary
It would be nice if the MSTEST0037 analyzer could suggest changing Linq statements with predicates that can be instead written with `Assert.Contains`, `Assert.DoesNotContain` or `Assert.ContainsSingle`.

## Background and Motivation
Cleaner code.

## Proposed Feature
The MSTEST0037 analyzer (and code fix) should be extended to cover these cases:

```cs
[TestClass]
public sealed class Test1
{
    private readonly IEnumerable<int> _enumerable = [];

    [TestMethod]
    public void ShouldSuggestContainsSingle()
    {
        // Expected analyzer suggestion:
        //Assert.ContainsSingle(x => x == 1, _enumerable);

        Assert.AreEqual(1, _enumerable.Where(x => x == 1).Count());
        Assert.AreEqual(1, _enumerable.Count(x => x == 1));
    }

    [TestMethod]
    public void ShouldSuggestContains()
    {
        // Expected analyzer suggestion:
        //Assert.Contains(x => x == 1, _enumerable);

        Assert.IsTrue(_enumerable.Where(x => x == 1).Any());
        Assert.IsTrue(_enumerable.Any(x => x == 1));

        Assert.AreNotEqual(0, _enumerable.Where(x => x == 1).Count());
        Assert.AreNotEqual(0, _enumerable.Count(x => x == 1));

        Assert.IsTrue(_enumerable.Where(x => x == 1).Count() > 0);
        Assert.IsTrue(_enumerable.Count(x => x == 1) > 0);

        // I added those only because the analyzer currently suggests them instead of the 'Assert.IsTrue' above.
        Assert.IsGreaterThan(0, _enumerable.Where(x => x == 1).Count());
        Assert.IsGreaterThan(0, _enumerable.Count(x => x == 1));
    }

    [TestMethod]
    public void ShouldSuggestDoesNotContain()
    {
        // Expected analyzer suggestion:
        //Assert.DoesNotContain(x => x == 1, _enumerable);

        Assert.IsFalse(_enumerable.Where(x => x == 1).Any());
        Assert.IsFalse(_enumerable.Any(x => x == 1));

        Assert.AreEqual(0, _enumerable.Where(x => x == 1).Count());
        Assert.AreEqual(0, _enumerable.Count(x => x == 1));

        Assert.IsFalse(_enumerable.Where(x => x == 1).Count() > 0);
        Assert.IsFalse(_enumerable.Count(x => x == 1) > 0);

        // I added those only because the analyzer currently suggests them instead of the 'Assert.IsFalse' above.
        Assert.IsLessThanOrEqualTo(0, _enumerable.Where(x => x == 1).Count());
        Assert.IsLessThanOrEqualTo(0, _enumerable.Count(x => x == 1));
    }
}
```

AtolagbeMuiz:
This Pull request fixes #6360 

This implementation involves changes to the Analyzer to suggest a much cleaner approach to write Assertions that involves the use of Predicate functions. for example 

```
Assert.IsTrue(_enumerable.Where(x => x == 1).Any());
Assert.IsTrue(_enumerable.Any(x => x == 1));
Assert.IsTrue(_enumerable.Count(x => x == 1) > 0);
```

Analyzer will suggest (Expected Behaviour) 
`Assert.Contains(x => x == 1, _enumerable);`

Example 2:
```
Assert.IsFalse(_enumerable.Where(x => x == 1).Any());
Assert.IsFalse(_enumerable.Any(x => x == 1));
Assert.IsFalse(_enumerable.Count(x => x == 1) > 0);
```

Analyzer will suggest (Expected Behaviour) 
`Assert.DoesNotContain(x => x == 1, _enumerable);`


src/Analyzers/MSTest.Analyzers/UseProperAssertMethodsAnalyzer.cs
```diff
@@ -125,6 +125,15 @@ private enum CountCheckStatus
         HasCount,
     }
 
+    private enum LinqPredicateCheckStatus
+    {
+        Unknown,
+        Any,
+        Count,
+        WhereAny,
```
stan-sz, line 133: Consider adding SingleAny and SingleCount for the `.Single()` Linq extension method.

AtolagbeMuiz: is this just to add it as part of the enums because, there is no use of it for now in the codebase except if we have an issue that wants the analyzer to check for predicate that uses Single Linq


@Youssef1313 thoughts on this?

stan-sz: I meant full support of `.Single()` in this analyzer.

AtolagbeMuiz: ohh okay.. please can you provide sample code of how it can be used and the expected suggestion code we want analyzer to give..

just to make sure we are on the same page. thank you.

cremor: I can see a few cases, but I'm not sure if they make sense. In all those cases the thrown exception would change at least in one of the possible failure cases. (And those old assertions are not good to start with.)
That's also why I didn't include those when I initially wrote the issue.

```cs
// Expected analyzer suggestion:
//Assert.ContainsSingle(x => x == 1, _enumerable);

// Questionable:
Assert.IsNotNull(_enumerable.Where(x => x == 1).SingleOrDefault());
Assert.IsNotNull(_enumerable.SingleOrDefault(x => x == 1));

// Even more questionable:
Assert.IsNotNull(_enumerable.Where(x => x == 1).Single());
Assert.IsNotNull(_enumerable.Single(x => x == 1));
```

```cs
// Expected analyzer suggestion:
//Assert.DoesNotContain(x => x == 1, _enumerable);

// Questionable:
Assert.IsNull(_enumerable.Where(x => x == 1).SingleOrDefault());
Assert.IsNull(_enumerable.SingleOrDefault(x => x == 1));
```

AtolagbeMuiz: the pull request is now updated with the requested changes.. analyzer has been updated for Single and SingleOrDefault @stan-sz @cremor

src/Analyzers/MSTest.Analyzers/UseProperAssertMethodsAnalyzer.cs
```diff
@@ -519,6 +582,146 @@ private static ComparisonCheckStatus RecognizeComparisonCheck(
         return ComparisonCheckStatus.Unknown;
     }
 
+    private static LinqPredicateCheckStatus RecognizeLinqPredicateCheck(
+    IOperation operation,
+    out SyntaxNode? collectionExpression,
+    out SyntaxNode? predicateExpression,
+    out IOperation? countOperation)
```
Evangelink:
```suggestion
    private static LinqPredicateCheckStatus RecognizeLinqPredicateCheck(
        IOperation operation,
        out SyntaxNode? collectionExpression,
        out SyntaxNode? predicateExpression,
        out IOperation? countOperation)
```

test/UnitTests/MSTest.Analyzers.UnitTests/UseProperAssertMethodsAnalyzerTests.cs
```diff
@@ -2908,4 +2908,566 @@ await VerifyCS.VerifyCodeFixAsync(
...
+            using Microsoft.VisualStudio.TestTools.UnitTesting;
+
+            [TestClass]
+            public class TestClass
+            {
+                [TestMethod]
+                public void TestMethod()
+                {
+                    var enumerable = new List<int>();
+                    {|#0:Assert.IsNull(enumerable.SingleOrDefault(x => x == 1))|};
```
Evangelink: This test would fail because `int` is a struct` so `SingleOrDefault` would return `default(int)` and not `null`.

This could be a case we want to catch in the always true/false analyzer (cc @Youssef1313)

test/UnitTests/MSTest.Analyzers.UnitTests/UseProperAssertMethodsAnalyzerTests.cs
```diff
@@ -2908,4 +2908,566 @@ await VerifyCS.VerifyCodeFixAsync(
...
+            using Microsoft.VisualStudio.TestTools.UnitTesting;
+
+            [TestClass]
+            public class TestClass
+            {
+                [TestMethod]
+                public void TestMethod()
+                {
+                    var enumerable = new List<int>();
+                    {|#0:Assert.IsNull(enumerable.Where(x => x == 1).SingleOrDefault())|};
```
Evangelink: Same remark about the test failing because struct can't be null.

test/UnitTests/MSTest.Analyzers.UnitTests/UseProperAssertMethodsAnalyzerTests.cs
Youssef1313: What happens for `enumerable.SingleOrDefault(x => x == null)` ? This would mean "either zero match of nulls in the collection, or exactly one". But in this case, the `DoesNotContain` suggestion will be wrong.

I would probably avoid reporting in this case completely, especially that the lambda can be really complex to analyze (it could be calling into another method for example)

test/UnitTests/MSTest.Analyzers.UnitTests/UseProperAssertMethodsAnalyzerTests.cs
```diff
@@ -2908,4 +2908,566 @@ await VerifyCS.VerifyCodeFixAsync(
...
+            using Microsoft.VisualStudio.TestTools.UnitTesting;
+
+            [TestClass]
+            public class TestClass
+            {
+                [TestMethod]
+                public void TestMethod()
+                {
+                    var enumerable = new List<int>();
+                    {|#0:Assert.IsNotNull(enumerable.Single(x => x == 1))|};
```
Youssef1313: This suggestion might be problematic as well. If `enumerable` is a collection that can contain nulls, this switching to `ContainsSingle` is making the assertion less strict. If we want to report a diagnostic in this case, I think it should be `ContainsSingle` followed by `Assert.IsNotNull`. So just expanding this line to two lines.

src/Analyzers/MSTest.Analyzers/UseProperAssertMethodsAnalyzer.cs
```diff
@@ -519,6 +582,146 @@ private static ComparisonCheckStatus RecognizeComparisonCheck(
...
+            whereSingleOrDefaultInvocation.Arguments.Length == 1 &&
+            whereSingleOrDefaultInvocation.Arguments[0].Value is IInvocationOperation whereInvocation4 &&
+            whereInvocation4.TargetMethod.Name == "Where" &&
+            whereInvocation4.TargetMethod.ContainingType?.ToDisplayString() == "System.Linq.Enumerable" &&
+            whereInvocation4.Arguments.Length == 2)
+        {
+            collectionExpression = whereInvocation4.Arguments[0].Value.Syntax;
+            predicateExpression = whereInvocation4.Arguments[1].Value.Syntax;
+            return LinqPredicateCheckStatus.WhereSingleOrDefault;
+        }
```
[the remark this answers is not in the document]
AtolagbeMuiz: refactor of RecognizeLinqPredicateCheck is done.. and pull request updated

src/Analyzers/MSTest.Analyzers/UseProperAssertMethodsAnalyzer.cs
```diff
@@ -268,6 +281,56 @@ private static void AnalyzeInvocationOperation(OperationAnalysisContext context,
...
+        LinqPredicateCheckStatus linqStatus = RecognizeLinqPredicateCheck(
+            argument,
+            out SyntaxNode? linqCollectionExpr,
+            out SyntaxNode? predicateExpr,
+            out _);
+
+        if (linqStatus is LinqPredicateCheckStatus.Single or
+                         LinqPredicateCheckStatus.SingleOrDefault or
+                         LinqPredicateCheckStatus.WhereSingle or
+                         LinqPredicateCheckStatus.WhereSingleOrDefault &&
```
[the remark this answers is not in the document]
AtolagbeMuiz: the if-statement has been parenthesized as suggested by copilot.. the pull request now updated

test/IntegrationTests/MSTest.IntegrationTests/OutputTests.cs
```diff
@@ -84,7 +84,8 @@ private static void ValidateOutputIsNotMixed(IEnumerable<TestResult> testResults
         Assert.Contains(methodName, message.Text);
         Assert.Contains("TestInitialize", message.Text);
         Assert.Contains("TestCleanup", message.Text);
-        Assert.IsFalse(shouldNotContain.Any(message.Text.Contains));
+        // Assert.IsFalse(shouldNotContain.Any(message.Text.Contains));
+        Assert.DoesNotContain(message.Text.Contains, shouldNotContain);
```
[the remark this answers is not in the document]
AtolagbeMuiz: chnages implemented, pull request updated

src/Analyzers/MSTest.Analyzers/UseProperAssertMethodsAnalyzer.cs
```diff
@@ -624,6 +857,54 @@ private static void AnalyzeIsTrueOrIsFalseInvocation(OperationAnalysisContext co
...
+                                countPredicateExpr.GetLocation(),
+                                countCollectionExpr.GetLocation()),
+                            properties: properties.ToImmutable(),
+                            properAssertMethod,
+                            isTrueInvocation ? "IsTrue" : "IsFalse"));
+
+                    return;
+                }
+            }
+        }
```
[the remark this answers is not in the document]
AtolagbeMuiz: implemented this to reuse `RecognizeLinqPredicateCheck` for the validation

test/UnitTests/MSTest.Analyzers.UnitTests/UseProperAssertMethodsAnalyzerTests.cs
AtolagbeMuiz: hmm, I didn't consider this edge case.. will review the analyzer again to avoid reporting `SingleOrDefault` and `Single`

test/UnitTests/MSTest.Analyzers.UnitTests/UseProperAssertMethodsAnalyzerTests.cs
AtolagbeMuiz: pull request updated to avoid reporting/suggesting for `SingleOrDefault ` and `Single`

Evangelink (approved):

src/Analyzers/MSTest.Analyzers/UseProperAssertMethodsAnalyzer.cs
```diff
@@ -529,6 +538,68 @@ private static ComparisonCheckStatus RecognizeComparisonCheck(
...
+        predicateExpression = null;
+        countOperation = null;
+
+        if (operation is not IInvocationOperation invocation)
+        {
+            return LinqPredicateCheckStatus.Unknown;
+        }
+
+        string methodName = invocation.TargetMethod.Name;
+        string? containingType = invocation.TargetMethod.ContainingType?.ToDisplayString();
```
Youssef1313, line 557: This is not a good practice, and might be costly performance wise. It's better to retrieve symbols early and do symbol comparison instead. I'll open an issue and address it later.

Issue #6732, opened 2025-10-12
Assert.ContainsSingle should accept non-generic collections


AtolagbeMuiz:
## Summary

`Assert.ContainsSingle` should accept non-generic IEnumerable collections like ArrayList, Stacks e.t.c. just the way `Assert.Contains` and `Assert.DoesNotContain` accept non-generic collection overload.

## Current Behaviour
`Assert.ContainsSingle` throws compile time error because it only accepts generic collections.

<img width="475" height="187" alt="Image" src="https://github.com/user-attachments/assets/6e3416ef-fcdd-4979-8941-042fcbcb0b4b" />

PR #6736, opened 2025-10-13
implemented non-generic overload for Assert.ContainsSingle

AtolagbeMuiz:
This Pull Request fixes [Assert.ContainsSingle should accept non-generic collections](https://github.com/microsoft/testfx/issues/6732) #6732 

This Implementation involves an overload being added to the `Assert.ContainsSingle` which allows non-generic collections like ArrayList.

### Current Behaviour
`Assert.ContainsSingle` throws compile time error because it only accepts generic collections.
Screenshot 2025-10-12 230729

### Expected Behaviour

With this Implementation, A`ssert.ContainsSingle` will allow non-generic collection as seen below;

image

image


AtolagbeMuiz:
removed the `AssertSingleInterpolatedStringHandler ` for non-generic implemnation based because we didnt implement if for `Assert.Count` as well.... 

@Youssef1313

test/UnitTests/TestFramework.UnitTests/Assertions/AssertTests.Contains.cs
```diff
@@ -311,6 +341,35 @@ public void ContainsSingle_InterpolatedHandler_WithMultipleElements_ThrowsExcept
...
+    public void ContainsSingle_InNonGenericCollection_NoMessage_WithEmptyCollection_ReturnsNoElement()
+    {
+        // Arrange
+        var collection = new ArrayList();
+
+        // Act
+        Action action = () => Assert.ContainsSingle(collection);
+
+        // Assert
+        action.Should().Throw<AssertFailedException>();
```
Evangelink: Let's add full expected message.

test/UnitTests/TestFramework.UnitTests/Assertions/AssertTests.Contains.cs
```diff
@@ -311,6 +341,35 @@ public void ContainsSingle_InterpolatedHandler_WithMultipleElements_ThrowsExcept
...
+        object? result = Assert.ContainsSingle(collection);
+
+        // Assert
+        result.Should().Be(null);
+    }
+
+    /// <summary>
+    /// Tests the ContainsSingle method without message parameters where the collection has a single element.
+    /// </summary>
+    public void ContainsSingle_InNonGenericCollection_NoMessage_WithEmptyCollection_ReturnsNoElement()
```
Evangelink: Let's do another test similar to this one with a custom message (assert custom message is there).

Evangelink (commented):
For the 2 new APIs, we could instead call the generic version by doing `.Cast<object>()`. There is a potential small performance hit but I think it's ok.

WDYT @Youssef1313

test/UnitTests/TestFramework.UnitTests/Assertions/AssertTests.Contains.cs
AtolagbeMuiz: i have added another test case similar to this but assert the actual custom exception error message thrown @Evangelink

test/UnitTests/TestFramework.UnitTests/Assertions/AssertTests.Contains.cs
AtolagbeMuiz: this has been added and the pull request updated

test/UnitTests/TestFramework.UnitTests/Assertions/AssertTests.Contains.cs
```diff
@@ -311,6 +341,50 @@ public void ContainsSingle_InterpolatedHandler_WithMultipleElements_ThrowsExcept
...
+    {
+        // Arrange
+        var collection = new ArrayList();
+
+        // Act
+        Action action = () => Assert.ContainsSingle(collection);
+
+        // Assert
+        action.Should().Throw<AssertFailedException>();
+    }
```
Evangelink: I'd remove this test as it has no value compared to test below (besides test comment is wrong).

test/UnitTests/TestFramework.UnitTests/Assertions/AssertTests.Contains.cs
```diff
@@ -311,6 +341,50 @@ public void ContainsSingle_InterpolatedHandler_WithMultipleElements_ThrowsExcept
...
+    {
+        // Arrange
+        var collection = new ArrayList();
+
+        // Act
+        Action action = () => Assert.ContainsSingle(collection);
+
+        // Assert
+        action.Should().Throw<AssertFailedException>().WithMessage("Assert.ContainsSingle failed. Expected collection to contain exactly one element but found 0 element(s). 'collection' expression: 'collection'.");
+    }
```
Evangelink, line 387: Please add another test case like:

```
 public void ContainsSingle_InNonGenericCollection_AssertCustomMessage_WithEmptyCollection_ThrowsException()
    {
        // Arrange
        var collection = new ArrayList();

        // Act
        Action action = () => Assert.ContainsSingle(collection, "my custom message");

        // Assert
        action.Should().Throw<AssertFailedException>().WithMessage("Assert.ContainsSingle failed. Expected collection to contain exactly one element but found 0 element(s). 'collection' expression: 'collection'.");
Comment view    }

**Review by Evangelink — COMMENTED** [review]
Reviving my earlier comment, I would personally use `Cast<object>` to avoid to have to maintain both implementations, the perf cost is IMO neglictible.

**Comment by Youssef1313** [comment]
@Evangelink I would personally avoid `.Cast<object>`. It might fail scenarios like:

```csharp
unsafe
{
    var x = new int*[] { (int*)1 };
    IEnumerable y = x;
    Assert.ContainsSingle(y);
}
```

It might also be worth adding such test :)

src/TestFramework/TestFramework/Assertions/Assert.Contains.cs
```diff
@@ -168,6 +211,52 @@ public static T ContainsSingle<T>(Func<T, bool> predicate, IEnumerable<T> collec
...
+    /// Users shouldn't pass a value for this parameter.
+    /// </param>
+    /// <param name="collectionExpression">
+    /// The syntactic expression of collection as given by the compiler via caller argument expression.
+    /// Users shouldn't pass a value for this parameter.
+    /// </param>
+    /// <returns>The item that matches the predicate.</returns>
+    public static object ContainsSingle(Func<object, bool> predicate, IEnumerable collection, string? message = "", [CallerArgumentExpression(nameof(predicate))] string predicateExpression = "", [CallerArgumentExpression(nameof(collection))] string collectionExpression = "")
+    {
+        var matchingElements = new List<object>();
```
Youssef1313: Are we creating a list unnecessary? We only care about first match and acount.

src/TestFramework/TestFramework/Assertions/Assert.Contains.cs
```diff
@@ -135,6 +135,49 @@ public static T ContainsSingle<T>(IEnumerable<T> collection, string? message = "
...
+    /// Tests whether the specified collection contains exactly one element.
+    /// </summary>
+    /// <param name="collection">The collection.</param>
+    /// <param name="message">The message to display when the assertion fails.</param>
+    /// <param name="collectionExpression">
+    /// The syntactic expression of collection as given by the compiler via caller argument expression.
+    /// Users shouldn't pass a value for this parameter.
+    /// </param>
+    /// <returns>The item.</returns>
+    public static object ContainsSingle(IEnumerable collection, string? message = "", [CallerArgumentExpression(nameof(collection))] string collectionExpression = "")
```
Youssef1313, line 136: This method can be written in terms of the predicate method, where the predicate is just `static _ => true`.

test/UnitTests/TestFramework.UnitTests/Assertions/AssertTests.Contains.cs
AtolagbeMuiz: i have replaced the initial test with this.. and pull request has been updated

test/UnitTests/TestFramework.UnitTests/Assertions/AssertTests.Contains.cs
AtolagbeMuiz: this is removed, and replaced with the suggested unit test..

src/TestFramework/TestFramework/Assertions/Assert.Contains.cs
AtolagbeMuiz: the unncessary list has been removed... i also noticed I missed writing unit tests that covers this  `ContainsSingle(Func<object, bool> predicate, IEnumerable collection, string? message = "", [CallerArgumentExpression(nameof(predicate))] string predicateExpression = "", [CallerArgumentExpression(nameof(collection))] string collectionExpression = "")`

these unit tests has been added and pull request updated..

src/TestFramework/TestFramework/Assertions/Assert.Contains.cs
AtolagbeMuiz: this has been written as a predicate method and changes pushed to Pull Request @Youssef1313

test/UnitTests/TestFramework.UnitTests/Assertions/AssertTests.Contains.cs
```diff
@@ -311,6 +341,50 @@ public void ContainsSingle_InterpolatedHandler_WithMultipleElements_ThrowsExcept
...
+    public void ContainsSingle_InNonGenericCollection_NoMessage_WithEmptyCollection_ReturnsNoElement()
+    {
+        // Arrange
+        var collection = new ArrayList();
+
+        // Act
+        Action action = () => Assert.ContainsSingle(collection);
+
+        // Assert
+        action.Should().Throw<AssertFailedException>();
```
Evangelink:
```suggestion
        action.Should().Throw<AssertFailedException>().WithMessage("Assert.ContainsSingle failed. Expected collection to contain exactly one element but found 0 element(s). 'collection' expression: 'collection'.");
```

test/UnitTests/TestFramework.UnitTests/Assertions/AssertTests.Contains.cs
```diff
@@ -311,6 +341,50 @@ public void ContainsSingle_InterpolatedHandler_WithMultipleElements_ThrowsExcept
...
+        object? result = Assert.ContainsSingle(collection);
+
+        // Assert
+        result.Should().Be(null);
+    }
+
+    /// <summary>
+    /// Tests the ContainsSingle method without message parameters where the collection has a single element.
+    /// </summary>
+    public void ContainsSingle_InNonGenericCollection_NoMessage_WithEmptyCollection_ReturnsNoElement()
```
Evangelink:
```suggestion
    public void ContainsSingle_InNonGenericCollection_NoMessage_WithEmptyCollection_ThrowsException()
```

PR #6768, opened 2025-10-20
Improve dev guide

MarcoRossignoli:
This pull request adds a new troubleshooting section to the developer documentation to help resolve build errors related to missing Universal Windows Platform (UWP) tooling targets. The documentation now guides users on how to properly set up their environment when building with Visual Studio.

### DIFF
--- a/docs/dev-guide.md
+++ b/docs/dev-guide.md
@@ -171,7 +171,6 @@ Inside Visual Studio, all projects can be built normally. All but acceptance tes
 
 ### Missing UAP tooling targets
 
-
 If you encounter errors similar to:
 
 ```text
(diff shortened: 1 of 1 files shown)


docs/dev-guide.md
```diff
@@ -167,6 +167,25 @@ If you are working with Visual Studio, we recommend opening it through the `open
...
+
+### Missing UAP tooling targets
+
+If you encounter errors similar to:
+
+```text
+C:\Users\<username>\.nuget\packages\msbuild.sdk.extras\3.0.44\Build\LanguageTargets\CheckMissing.targets(44,5): error : The specified language targets for uap10.0.16299 is missing. Ensure correct tooling is installed for 'uap'. Missing: 'C:\Program Files\Microsoft Visual Studio\xx\xxxx\MSBuild\Microsoft\WindowsXaml\vXX.0\Microsoft.Windows.UI.Xaml.CSharp.targets'
+```
+
+This indicates that MSBuild cannot locate the required Universal Windows Platform (UWP) build targets. To resolve this issue, ensure you run the build using the **Visual Studio Developer Command Prompt** instead of a regular command prompt or PowerShell session. The Developer Command Prompt sets up the necessary environment variables and paths for MSBuild to find all required tooling.
```
Youssef1313: Can we reword this a bit for more accurate description of the issue?

This is all about either:

- Installing the required workloads into the correct Visual Studio.
  - When using non-VS command-line, that is the "latest" VS.
  - When using VS command-line, that is the version of that specific VS.
- or; if the required workloads are already installed into a specific VS that is not latest, use the VS command-line corresponding to it.

I don't know how to best word/describe that info.

cc @nohwnd

nohwnd: looks like we should be able to provide requirements to VS installation, so vswhere can locate the correct version, if there is any.

And then this can be reworded based on that error. looking into that.

MarcoRossignoli: Feel free to takeover the PR if you have more precise instructions thanks!

nohwnd: Might be replaced by https://github.com/microsoft/testfx/pull/6769, then we should see other error.

Issue #6738, opened 2025-10-14
Deployment item doesn't work well when app domain is disabled


Youssef1313:
This is a bug dating back to the oldest stable MSTest version on nuget.org. The two features never worked correctly together.

[TestProject44.zip](https://github.com/user-attachments/files/22901374/TestProject44.zip)

Run with `dotnet test -- RunConfiguration.DisableAppDomain=true`:

- **All versions**: The test assembly is loaded twice in the same app domain. Once from bin and once from deployment directory

    <img width="1006" height="206" alt="Image" src="https://github.com/user-attachments/assets/c2b43b40-8ed3-4755-9277-b371f11729ba" />

- Up to **3.10.5**: The test fails with InvalidCastException:

  ```
      Example (10s 862ms): Error Message: Test method TestProject44.Test1.Example threw exception:
      System.InvalidCastException: [A]ExampleDataType cannot be cast to [B]ExampleDataType. Type A originates from 'TestProject44, Version=1.0.0.0, Culture=neutral, Pub
      licKeyToken=null' in the context 'LoadFrom' at location 'C:\Users\ygerges\source\repos\TestProject44\TestProject44\TestResults\Deploy_ygerges 2025-10-14 11_24_27\
      Out\TestProject44.dll'. Type B originates from 'TestProject44, Version=1.0.0.0, Culture=neutral, PublicKeyToken=null' in the context 'Default' at location 'C:\Use
      rs\ygerges\source\repos\TestProject44\TestProject44\bin\Debug\net48\TestProject44.dll'.
      Stack Trace:
          at TestProject44.Test1.ExampleDataTypeTypeConverter.ConvertTo(ITypeDescriptorContext context, CultureInfo culture, Object value, Type destinationType) in C:\U
      sers\ygerges\source\repos\TestProject44\TestProject44\Test1.cs:line 33
         at TestProject44.Test1.Example() in C:\Users\ygerges\source\repos\TestProject44\TestProject44\Test1.cs:line 17
  ```

- **3.11.0** and **4.0.0**: the test passes, even if the assembly is still loaded twice. That **might** get broken as it was before again when https://github.com/microsoft/testfx/pull/6718 is merged.

Youssef1313:
This is critical for MTP as appdomain is disabled by default now. This needs to be fixed ASAP. We need to decide if we will make a 4.0.2 for this or go directly to 4.1.

[...]

PR #6772, opened 2025-10-20
Fix DeploymentItem when appdomain is disabled or not available

Youssef1313:
Fixes https://github.com/microsoft/testfx/issues/6738

### DIFF
--- a/src/Adapter/MSTestAdapter.PlatformServices/Deployment/TestRunDirectories.cs
+++ b/src/Adapter/MSTestAdapter.PlatformServices/Deployment/TestRunDirectories.cs
@@ -32,13 +32,23 @@ internal sealed class TestRunDirectories
     /// Initializes a new instance of the <see cref="TestRunDirectories"/> class.
     /// </summary>
     /// <param name="rootDirectory">The root directory path.</param>
-    public TestRunDirectories(string rootDirectory)
+    /// <param name="firstTestSource">
+    /// The path to the test assembly of the first test case. In most cases, all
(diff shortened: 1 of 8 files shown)


src/Adapter/MSTestAdapter.PlatformServices/Deployment/TestRunDirectories.cs
```diff
@@ -32,13 +32,23 @@ internal sealed class TestRunDirectories
...
+    /// <param name="isAppDomainCreationDisabled">Whether or not app domain is disabled.</param>
+    public TestRunDirectories(string rootDirectory, string firstTestSource, bool isAppDomainCreationDisabled)
     {
         DebugEx.Assert(!StringEx.IsNullOrEmpty(rootDirectory), "rootDirectory");
 
         RootDeploymentDirectory = rootDirectory;
         InDirectory = Path.Combine(RootDeploymentDirectory, DeploymentInDirectorySuffix);
-        OutDirectory = Path.Combine(RootDeploymentDirectory, DeploymentOutDirectorySuffix);
+
+        OutDirectory = isAppDomainCreationDisabled
```
Youssef1313: If appdomain is disabled, we keep deployment out dir as as the assembly being tested.

There is one pitfall here, we could have been given multiple assemblies to run. In that case, there is no really single directory that we are supposed to use. That makes the whole deployment item feature design broken. This should have never been implemented for .NET (Core), and should have never been supported when appdomain is disabled (well, it already never works correctly today in those cases). But breaking it completely is a big breaking change.
This leaves us potentially with this hack as our only option, unless others have better ideas.

PR #6777, opened 2025-10-21
Add AI support to the `Microsoft.Testing.Platform`

MarcoRossignoli:
Add AI support to the `Microsoft.Testing.Platform`

### DIFF
--- a/src/Platform/Microsoft.Testing.Platform/AI/ChatClientManager.cs
+++ b/src/Platform/Microsoft.Testing.Platform/AI/ChatClientManager.cs
@@ -10,7 +10,7 @@ internal sealed class ChatClientManager : IChatClientManager
 {
     private Func<IServiceProvider, object>? _chatClientProviderFactory;
 
-    public void SetChatClientProviderFactory(Func<IServiceProvider, object> chatClientProviderFactory)
+    public void AddChatClientProvider(Func<IServiceProvider, object> chatClientProviderFactory)
     {
         if (chatClientProviderFactory is null)
         {
@@ -25,7 +25,7 @@ internal sealed class ChatClientManager : IChatClientManager
         _chatClientProviderFactory = chatClientProviderFactory;
     }
 
-    public void InstantiateChatClientProvider(ServiceProvider serviceProvider)
+    public void BuildChatClients(ServiceProvider serviceProvider)
     {
         if (_chatClientProviderFactory is not null)
         {
--- a/src/Platform/Microsoft.Testing.Platform/AI/IChatClientManager.cs
+++ b/src/Platform/Microsoft.Testing.Platform/AI/IChatClientManager.cs
@@ -7,7 +7,7 @@ namespace Microsoft.Testing.Platform.AI;
 
 internal interface IChatClientManager
 {
-    void SetChatClientProviderFactory(Func<IServiceProvider, object> chatClientProviderFactory);
+    void AddChatClientProvider(Func<IServiceProvider, object> chatClientProviderFactory);
 
-    void InstantiateChatClientProvider(ServiceProvider serviceProvider);
+    void BuildChatClients(ServiceProvider serviceProvider);
 }
--- a/src/Platform/Microsoft.Testing.Platform.AI/ChatClientProviderExtensions.cs
+++ b/src/Platform/Microsoft.Testing.Platform.AI/ChatClientProviderExtensions.cs
@@ -25,7 +25,7 @@ public static class ChatClientProviderExtensions
             throw new InvalidOperationException(PlatformResources.InvalidTestApplicationBuilderTypeForAI);
         }
 
-        builder.ChatClientManager.SetChatClientProviderFactory(chatClientProvider);
+        builder.ChatClientManager.AddChatClientProvider(chatClientProvider);
     }
 
     /// <summary>
--- a/src/Platform/Microsoft.Testing.Platform/Hosts/TestHostBuilder.cs
+++ b/src/Platform/Microsoft.Testing.Platform/Hosts/TestHostBuilder.cs
@@ -314,7 +314,7 @@ internal sealed class TestHostBuilder(IFileSystem fileSystem, IRuntimeFeature ru
         serviceProvider.AddService(testApplicationResult);
 
         // Add Chat Client if AI capabilities are enabled
-        ChatClientManager.InstantiateChatClientProvider(serviceProvider);
+        ChatClientManager.BuildChatClients(serviceProvider);
 
         // ============= SETUP COMMON SERVICE USED IN ALL MODES END ===============//
 


src/Platform/Microsoft.Testing.Platform.AI/AIExtensions.cs
```diff
@@ -0,0 +1,38 @@
...
+using Microsoft.Testing.Platform.Builder;
+using Microsoft.Testing.Platform.Resources;
+using Microsoft.Testing.Platform.Services;
+
+namespace Microsoft.Testing.Platform.AI;
+
+/// <summary>
+/// Extension methods for AI-related functionality.
+/// </summary>
+public static class AIExtensions
```
MarcoRossignoli: Rename to `ChatClientFactoryExtensions`

src/Platform/Microsoft.Testing.Platform.AI/IChatClientFactory.cs
```diff
@@ -0,0 +1,18 @@
...
+/// <summary>
+/// Factory interface for creating chat clients.
+/// </summary>
+public interface IChatClientFactory
+{
+    /// <summary>
+    /// Creates a new instance of <see cref="IChatClient"/>.
+    /// </summary>
+    /// <returns>An instance of <see cref="IChatClient"/>.</returns>
+    IChatClient CreateChatClient();
```
nohwnd: Factory implementations in MTP are all over the place, but to me it feels like this should at least be async and take a cancellation token, so when we want to integrate with other ways to get credentials we can call their async apis without doing tricks.

https://grep.app/search?f.path=src%2FPlatform%2F&f.repo.pattern=testfx&regexp=true&q=interface.*Factory

E.g. in the past we've been asked by Playwright service team to support azure authentication providers, which I can see being relevant here.

https://learn.microsoft.com/en-us/dotnet/api/azure.identity.defaultazurecredential?view=azure-dotnet

MarcoRossignoli: It's fine we can change the shape, it's a quick draft to show the idea

TestFx.slnx
```diff
@@ -23,16 +23,17 @@
...
     <File Path="eng/TestingPlatformRunner/TestingPlatformRunner.targets" />
   </Folder>
   <Folder Name="/src/">
     <File Path="src/.editorconfig" />
     <File Path="src/Directory.Build.props" />
   </Folder>
-  <Folder Name="/src/1 - Platform and Extensions/" Id="4272bbd9-efdb-ae85-3883-a1ab2fbd20bd">
+  <Folder Name="/src/1 - Platform and Extensions/">
     <File Path="src/Platform/Directory.Build.props" />
+    <Project Path="src/Platform/Microsoft.Testing.Extensions.AzFoundry/Microsoft.Testing.Extensions.AzFoundry.csproj" Id="60c660b9-6d0c-4967-b21c-e1c34257cc75" />
```
Youssef1313:
```suggestion
    <Project Path="src/Platform/Microsoft.Testing.Extensions.AzFoundry/Microsoft.Testing.Extensions.AzFoundry.csproj" />
```

TestFx.slnx
```diff
@@ -43,19 +44,20 @@
     <Project Path="src/Platform/Microsoft.Testing.Extensions.TrxReport.Abstractions/Microsoft.Testing.Extensions.TrxReport.Abstractions.csproj" />
     <Project Path="src/Platform/Microsoft.Testing.Extensions.TrxReport/Microsoft.Testing.Extensions.TrxReport.csproj" />
     <Project Path="src/Platform/Microsoft.Testing.Extensions.VSTestBridge/Microsoft.Testing.Extensions.VSTestBridge.csproj" />
+    <Project Path="src/Platform/Microsoft.Testing.Platform.AI/Microsoft.Testing.Platform.AI.csproj" Id="53db5319-243a-4b47-91f1-9b450b4f2917" />
```
Youssef1313:
```suggestion
    <Project Path="src/Platform/Microsoft.Testing.Platform.AI/Microsoft.Testing.Platform.AI.csproj" />
```

src/Platform/Microsoft.Testing.Extensions.AzureFoundry/Microsoft.Testing.Extensions.AzureFoundry.csproj
```diff
@@ -0,0 +1,16 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
```
Youssef1313, line 1: Can we add this to MTP slnf please?

src/Platform/Microsoft.Testing.Extensions.AzFoundry/OpenAIChatClientFactory.cs
```diff
@@ -0,0 +1,47 @@
...
+/// Factory for creating Azure OpenAI chat clients.
+/// </summary>
+public sealed class AzureOpenAIChatClientFactory : IChatClientFactory
+{
+    /// <inheritdoc />
+    public IChatClient CreateChatClient()
+    {
+        string? endpoint = Environment.GetEnvironmentVariable("AZURE_OPENAI_ENDPOINT");
+        string? deploymentName = Environment.GetEnvironmentVariable("AZURE_OPENAI_DEPLOYMENT_NAME");
+        string? apiKey = Environment.GetEnvironmentVariable("AZURE_OPENAI_API_KEY");
```
Youssef1313: Should we be given `IEnvrionment` here?

src/Platform/Microsoft.Testing.Platform.AI/IChatClientFactory.cs
Youssef1313: Seems like it's a public API? So breaking in the future could be problematic.
If we anticipate that we might need it to be async in future, I think we should do now.

src/Platform/Microsoft.Testing.Platform.AI/Microsoft.Testing.Platform.AI.csproj
```diff
@@ -0,0 +1,15 @@
+<Project Sdk="Microsoft.NET.Sdk">
```
Youssef1313, line 1: Please add to MTP slnf

src/Platform/Microsoft.Testing.Platform.AI/IChatClientFactory.cs
MarcoRossignoli: Yep will be public I'll update with ct and async.

src/Platform/Microsoft.Testing.Extensions.AzureFoundry/Microsoft.Testing.Extensions.AzureFoundry.csproj
MarcoRossignoli: Didn't know about custom slnx will do

src/Platform/Microsoft.Testing.Extensions.AzFoundry/OpenAIChatClientFactory.cs
MarcoRossignoli: I think it's fine, it's our extension we can flow our IEnv

MarcoRossignoli:
> Any tests possible for this?

At the moment nope it's experimental for now...I need to play a bit with the usage to understand, I would not ship the core ai and the extension lib for now, the plumbing is internal and should not affect the normal run if not registered.

We can try to add some real test when complete...but not sure if we will be able to test the AI interaction, but let's see maybe we will do it in some way.

MarcoRossignoli:
> Should we avoid publishing it publicly for now until we implement something that uses it and we have more confidence of the design? Maybe we should set <IsShipping>false</IsShipping> in the new projects?

wonder if we can ship it inside not official nuget feed but in the dogfooding testing one so I can pick it up to do some experiment outside the testfx repo.

We can always delist there in case.

Youssef1313:
I'm fine publishing to test-tools feed and not nuget.org. But I want to ensure we don't end up publishing to nuget.org accidentally. I think setting `IsShipping` to false is not going to prevent publishing to test-tools feed as long as the pattern [here](https://github.com/microsoft/testfx/blob/82daaae2be64c7d3058a15656781bbf790e4e628/azure-pipelines-official.yml#L196) is adjusted to match NonShipping.

Microsoft.Testing.Platform.slnf
```diff
@@ -13,6 +14,7 @@
       "src\\Platform\\Microsoft.Testing.Extensions.TrxReport.Abstractions\\Microsoft.Testing.Extensions.TrxReport.Abstractions.csproj",
       "src\\Platform\\Microsoft.Testing.Extensions.TrxReport\\Microsoft.Testing.Extensions.TrxReport.csproj",
       "src\\Platform\\Microsoft.Testing.Extensions.VSTestBridge\\Microsoft.Testing.Extensions.VSTestBridge.csproj",
+      "src\\Platform\\Microsoft.Testing.Platform.AI\\Microsoft.Testing.Platform.AI.csproj",
```
Evangelink, line 17: Shall we rename to .Copilot? You know why

src/Platform/Microsoft.Testing.Extensions.AzureFoundry/OpenAIChatClientProvider.cs
```diff
@@ -0,0 +1,67 @@
...
+/// </summary>
+public sealed class AzureOpenAIChatClientProvider : IChatClientProvider
+{
+    private readonly IServiceProvider _serviceProvider;
+
+    /// <summary>
+    /// Initializes a new instance of the <see cref="AzureOpenAIChatClientProvider"/> class.
+    /// </summary>
+    /// <param name="serviceProvider">The service provider.</param>
+    public AzureOpenAIChatClientProvider(IServiceProvider serviceProvider)
```
Evangelink: Let's take IEnvironment instead as it's the only service we need

src/Platform/Microsoft.Testing.Platform.AI/Microsoft.Testing.Platform.AI.csproj
```diff
@@ -0,0 +1,20 @@
...
+  <ItemGroup>
+    <ProjectReference Include="$(RepoRoot)src\Platform\Microsoft.Testing.Platform\Microsoft.Testing.Platform.csproj" />
+  </ItemGroup>
+
+  <ItemGroup>
+    <PackageReference Include="Microsoft.Extensions.AI" />
+  </ItemGroup>
+
+  <ItemGroup>
+    <InternalsVisibleTo Include="Microsoft.Testing.Extensions.AzureFoundry" />
```
Evangelink: It's missing the public key property, isn't it?

src/Platform/Microsoft.Testing.Platform/AI/ChatClientManager.cs
```diff
@@ -0,0 +1,35 @@
...
+
+        if (_chatClientProvider is not null)
+        {
+            throw new InvalidOperationException(PlatformResources.ChatClientProviderAlreadyRegistered);
+        }
+
+        _chatClientProvider = chatClientProvider;
+    }
+
+    public void RegisterChatClientProvider(ServiceProvider serviceProvider)
```
Evangelink: It's not entirely clear what's the difference between Register and Add

Microsoft.Testing.Platform.slnf
MarcoRossignoli: I think this has nothing to do with Copilot...Copilot will be eventually a "chat provider"(like AzureFoundry) this is more like M.E.AI

src/Platform/Microsoft.Testing.Platform.AI/Microsoft.Testing.Platform.AI.csproj
MarcoRossignoli: mmm it's building fine

src/Platform/Microsoft.Testing.Platform/AI/ChatClientManager.cs
MarcoRossignoli: Add is used to save the ref to the lambda factory, register used in the builder phase to "register" the factory inside the service provider.

src/Platform/Microsoft.Testing.Platform.AI/Microsoft.Testing.Platform.AI.csproj
MarcoRossignoli: Also I don't think it's actually needed, let me check

Youssef1313: Hmm, I'm also surprised this is building fine.

src/Platform/Microsoft.Testing.Extensions.AzureFoundry/OpenAIChatClientProvider.cs
Youssef1313: This is quite concerning for me. It's a public API. What if we wanted to have additional services in future? :/

MarcoRossignoli: I'll make it an extension method with all internals, the api is experimental not meant for public yet

MarcoRossignoli: also by design the fact that we ship everything together we won't have buildability issues.

src/Platform/Microsoft.Testing.Platform.AI/Microsoft.Testing.Platform.AI.csproj
MarcoRossignoli: not needed removed

src/Platform/Microsoft.Testing.Extensions.AzureFoundry/Microsoft.Testing.Extensions.AzureFoundry.csproj
```diff
@@ -0,0 +1,32 @@
...
+    <Content Include="buildMultiTargeting/**">
+      <Pack>true</Pack>
+      <PackagePath>buildMultiTargeting</PackagePath>
+    </Content>
+    <TfmSpecificPackageFile Include="buildTransitive/**">
+      <PackagePath>buildTransitive/$(TargetFramework)</PackagePath>
+    </TfmSpecificPackageFile>
+    <TfmSpecificPackageFile Include="build/**">
+      <PackagePath>build/$(TargetFramework)</PackagePath>
+    </TfmSpecificPackageFile>
```
Youssef1313, line 20: I don't think we need these to be tfm-specific for now.

src/Platform/Microsoft.Testing.Extensions.AzureFoundry/Microsoft.Testing.Extensions.AzureFoundry.csproj
```diff
@@ -0,0 +1,32 @@
...
+
+  <PropertyGroup>
+    <TargetFrameworks>netstandard2.0;$(SupportedNetFrameworks)</TargetFrameworks>
+    <IsShipping>false</IsShipping>
+  </PropertyGroup>
+
+  <!-- NuGet package layout -->
+  <!-- NuGet folders https://learn.microsoft.com/nuget/create-packages/creating-a-package#from-a-convention-based-working-directory -->
+  <ItemGroup>
+    <Content Include="buildMultiTargeting/**">
```
Youssef1313, line 11: Should this be `None` instead?

src/Platform/Microsoft.Testing.Extensions.AzureFoundry/Microsoft.Testing.Extensions.AzureFoundry.csproj
MarcoRossignoli: I took what we do in other shipped extensions

src/Platform/Microsoft.Testing.Extensions.AzureFoundry/Microsoft.Testing.Extensions.AzureFoundry.csproj
MarcoRossignoli: I took what we do in other shipped extensions

MarcoRossignoli: we can update bulk after if it's not affecting the packaging

src/Platform/Microsoft.Testing.Extensions.AzureFoundry/Microsoft.Testing.Extensions.AzureFoundry.csproj
MarcoRossignoli: we can update bulk after if it's not affecting the packaging

docs/microsoft.testing.platform/001-AI-Extensibility.md
```diff
@@ -0,0 +1,296 @@
...
+    /// A provider may be registered but not available if required configuration
+    /// (e.g., environment variables) is missing.
+    /// </summary>
+    bool IsAvailable { get; }
+
+    /// <summary>
+    /// Gets a value indicating whether the chat client supports tool calling
+    /// (e.g., MCP tools or functions).
+    /// </summary>
+    bool SupportsToolCalling { get; }
```
Evangelink: Maybe `HasToolCapability`

docs/microsoft.testing.platform/001-AI-Extensibility.md
```diff
@@ -0,0 +1,296 @@
...
+
+##### 2. ChatClientManager (Internal)
+
+The platform includes an internal manager that handles provider registration:
+
+```csharp
+internal interface IChatClientManager
+{
+    void AddChatClientProvider(Func<IServiceProvider, object> chatClientProvider);
+    void RegisterChatClientProvider(ServiceProvider serviceProvider);
```
Evangelink: It's not easy to grasp difference between APIs here.

src/Platform/Microsoft.Testing.Extensions.AzureFoundry/Resources/ExtensionResources.resx
```diff
@@ -0,0 +1,129 @@
...
+  </resheader>
+  <data name="AzureOpenAIEndpointNotSet" xml:space="preserve">
+    <value>AZURE_OPENAI_ENDPOINT environment variable is not set.</value>
+  </data>
+  <data name="AzureOpenAIDeploymentNameNotSet" xml:space="preserve">
+    <value>AZURE_OPENAI_DEPLOYMENT_NAME environment variable is not set.</value>
+  </data>
+  <data name="AzureOpenAIApiKeyNotSet" xml:space="preserve">
+    <value>AZURE_OPENAI_API_KEY environment variable is not set.</value>
+  </data>
```
Evangelink: Maybe reverse order in the message and create a single resx entry: `Environment variable '{0}' is not set`

src/Platform/Microsoft.Testing.Extensions.AzureFoundry/OpenAIChatClientProvider.cs
```diff
@@ -0,0 +1,72 @@
...
+    internal AzureOpenAIChatClientProvider(IServiceProvider serviceProvider)
+    {
+        _environment = serviceProvider.GetRequiredService<IEnvironment>();
+    }
+
+    /// <inheritdoc />
+    public bool IsAvailable =>
+        !string.IsNullOrEmpty(_environment.GetEnvironmentVariable("AZURE_OPENAI_ENDPOINT")) &&
+        !string.IsNullOrEmpty(_environment.GetEnvironmentVariable("AZURE_OPENAI_DEPLOYMENT_NAME")) &&
+        !string.IsNullOrEmpty(_environment.GetEnvironmentVariable("AZURE_OPENAI_API_KEY"));
```
Evangelink: Use the string helper (not sure to recall the name in testfx repo).

src/Platform/Microsoft.Testing.Extensions.AzureFoundry/OpenAIChatClientProvider.cs
```diff
@@ -0,0 +1,72 @@
...
+/// </summary>
+internal sealed class AzureOpenAIChatClientProvider : IChatClientProvider
+{
+    private readonly IEnvironment _environment;
+
+    /// <summary>
+    /// Initializes a new instance of the <see cref="AzureOpenAIChatClientProvider"/> class.
+    /// </summary>
+    /// <param name="serviceProvider">The service provider.</param>
+    internal AzureOpenAIChatClientProvider(IServiceProvider serviceProvider)
```
Evangelink: I would take IEnvironment here. It's easier to grasp what the service needs.

src/Platform/Microsoft.Testing.Platform.AI/IChatClientProvider.cs
```diff
@@ -0,0 +1,35 @@
...
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using Microsoft.Extensions.AI;
+
+namespace Microsoft.Testing.Platform.AI;
+
+/// <summary>
+/// Provider interface for creating and configuring chat clients.
+/// </summary>
+public interface IChatClientProvider
```
Evangelink, line 12: Shall we make it Experimental to give us time to refine?

docs/microsoft.testing.platform/001-AI-Extensibility.md
MarcoRossignoli: in LLM space the feature is called "tool calling", tool capability looks pretty generic imo

docs/microsoft.testing.platform/001-AI-Extensibility.md
MarcoRossignoli: Propose a name pls it's all internal, register is "to register to the service provider" looks clear to me, but open to different naming

src/Platform/Microsoft.Testing.Extensions.AzureFoundry/OpenAIChatClientProvider.cs
MarcoRossignoli: It's all internal but it's fine I'll update the ctor

src/Platform/Microsoft.Testing.Platform.AI/IChatClientProvider.cs
MarcoRossignoli: We won't ship in public for now, I'll anyway add experimental.

src/Platform/Microsoft.Testing.Platform/AI/IChatClientManager.cs
```diff
@@ -0,0 +1,13 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using Microsoft.Testing.Platform.Services;
+
+namespace Microsoft.Testing.Platform.AI;
+
+internal interface IChatClientManager
+{
+    void SetChatClientProviderFactory(Func<IServiceProvider, object> chatClientProviderFactory);
```
Evangelink: To match other API it's probably better as `AddChatClientProvider`, ideally having a marker interface instead of `object`.

src/Platform/Microsoft.Testing.Platform/AI/IChatClientManager.cs
```diff
@@ -0,0 +1,13 @@
...
+
+    void InstantiateChatClientProvider(ServiceProvider serviceProvider);
```
Evangelink: Maybe `BuildChatClients` and to match other it would probably be async but given it's internal we can make async later IMO.

src/Platform/Microsoft.Testing.Platform/AI/IChatClientManager.cs
MarcoRossignoli: done, not clear what's the goal of a marker interface...the AI lib is part of the core...so cannot be a wrong "value" and have a marker add one more interface to inherit for users(also if without any member), I would for now keep as-is, we cannot break ourselves shipping everything together, the separation is only to avoid deps on the core

src/Platform/Microsoft.Testing.Platform/AI/IChatClientManager.cs
MarcoRossignoli: we will make async when will be needed.

Evangelink (approved):

PR #6783, opened 2025-10-22
No nuget access

nohwnd:
Pipeline setting ensures that nuget.org access is completely blocked. To avoid restoring packages from there, even in tests.

To try this locally you can add this to hosts file (C:\Windows\System32\drivers\etc\hosts

```
127.0.0.1 api.nuget.org
```

The reference assemblies default restore will cache the restored packages in users temp, debug into ResolveAsync method, you will see where it grabs it from in ResolveAsyncCore.

Changes are not observable on the normal PR pipeline, but I've run build on internal pipeline. Unfortunately there is no public infra for 1es so we won't be able to get this into PR build as well, but it should be rare that we see this fail.

### DIFF
--- a/test/Utilities/Microsoft.Testing.TestInfrastructure/TestAsset.cs
+++ b/test/Utilities/Microsoft.Testing.TestInfrastructure/TestAsset.cs
@@ -109,16 +109,12 @@ public class TestAsset : IDisposable
     {
         string publicFeedsFragment = addPublicFeeds
             ? """
-                    <add key="nuget.org" value="https://api.nuget.org/v3/index.json" />
-                    <add key="test-tools" value="https://pkgs.dev.azure.com/dnceng/public/_packaging/test-tools/nuget/v3/index.json" />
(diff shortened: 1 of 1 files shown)


nohwnd:
The build : https://dnceng.visualstudio.com/internal/_build/results?buildId=2822110&view=results

Youssef1313 (approved):

azure-pipelines-official.yml
```diff
@@ -75,6 +75,8 @@ extends:
       name: $(DncEngInternalBuildPool)
       image: windows.vs2022preview.amd64
       os: windows
+    settings:
+      networkIsolationPolicy: Permissive,CFSClean
```
Youssef1313, line 79: What is the default when this is not set?

Why is it not sufficient to only update the tests to not use nuget.org?

nohwnd: This is what the compliance guidance says, I can link you to the documentation. My guess is that they require the setting to be explicitly set to be able to on-glance identify the compliant repos, rather than identify them based on behavior.

This setting will setup the pipeline in a way that blocks access to api.nuget.org completely.

PR #6794, opened 2025-10-24
Fix diff marker in Assert.AreEqual when strings are compared to not point at '...'

### LINKED ISSUE #6784: Visual indicator for string AreEqual doesn't work well for mismatches at the end of long strings
## Summary
Good dehavior with enough characters at the end
```
Assert.AreEqual("01234567890123456789012345678901234567890123X", "01234567890123456789012345678901234567890123456789");

Assert.AreEqual failed. Expected string length 45 but was 50. 'expected' expression: '"01234567890123456789012345678901234567890123X"', 'actual' expression: '"01234567890123456789012345678901234567890123456789"'.
Expected: "...23456789012345678901234567890123X"
But was:  "...23456789012345678901234567890123456789"
----------------------------------------------^
```

Bad behavior with not enough characters at the end
```
Assert.AreEqual("01234567890123456789012345678901234567890123X", "01234567890123456789012345678901234567890123456");

 Assert.AreEqual failed. Expected string length 45 but was 47. 'expected' expression: '"01234567890123456789012345678901234567890123X"', 'actual' expression: '"01234567890123456789012345678901234567890123456"'.
 Expected: "...7890123456789012345678901234567890123X"
 But was:  "...78901234567890123456789012345678901..."
 ---------------------------------------------------^
```
## Background and Motivation

While verifying new behavior for mismatches I noticed that I wasn't able to see the mismatch at the end of a long string.

## Proposed Feature

Similar to how when there are 5 characters after the mismatch and the end of the longer string is shown it should not produce ellipsis to hide the mismatch for only 2 additional characters.

nohwnd:
Fix #6784

### DIFF
--- a/test/UnitTests/TestFramework.UnitTests/Assertions/AssertTests.AreEqualTests.cs
+++ b/test/UnitTests/TestFramework.UnitTests/Assertions/AssertTests.AreEqualTests.cs
@@ -1594,7 +1594,6 @@ public partial class AssertTests : TestContainer
             stringPreview);
     }
 
-
     public void CreateStringPreviews_MakeSureWeDontPointToEndEllipsis_WhenLongerStringOneCharLargerThanPreviewWindow()
     {
         // We will mask last 3 chars of the string, so we need to make sure that the diff index is not pointing to the end ellipsis.


test/UnitTests/TestFramework.UnitTests/Assertions/AssertTests.AreEqualTests.cs
```diff
@@ -1628,4 +1707,19 @@ private static string DigitString(int length, int diffIndex)
 
         return result.ToString();
     }
+
+    private void StringPreviewsAreEqual(string expected, string actual)
```
nohwnd, line 1711: awesome assertions shorten the text and replace new lines, that makes it much harder to see the diff. This just shows it as is.

test/UnitTests/TestFramework.UnitTests/Assertions/AssertTests.AreEqualTests.cs
```diff
@@ -1537,11 +1539,12 @@ public void CreateStringPreviews_DiffPointsToCorrectPlaceInShortenedStringWithEn
         int length = preview + 10;
         int diffIndex = 0;
         string stringPreview = FormatStringPreview(StringPreviewHelper.CreateStringPreviews(DigitString(length, diffIndex), DigitString(length, diffIndex), diffIndex, preview));
-        stringPreview.Should().Be("""
+        StringPreviewsAreEqual(
+            """
             "X12345..."
```
Youssef1313, line 1544: Not sure if this should be a single ellipsis char `…` instead of three dots `...`, but that's a different topic.

nohwnd: Single ellipsis might not render everywhere.

PR #6795, opened 2025-10-24
Fix assert that test

nohwnd:
Test fails locally because my culture is not en-us.


Evangelink:
@Youssef1313 This makes me think it would be good to add a new leg for PR building with a different culture (maybe just Windows as it's where we have most tests).

PR #6796, opened 2025-10-24
Codefix for MSTEST0023

### LINKED ISSUE #6585: Implement a codefix for MSTEST0023
Currently, [MSTEST0023](https://learn.microsoft.com/dotnet/core/testing/mstest-analyzers/mstest0023) doesn't have an associated CodeFixProvider to help users fix all violations easily. We should add that.

AtolagbeMuiz:
This PR fixes #6585 

This Implementation is to modify the `DoNotNegateBoolean` Analyzer code for to suggest better and cleaner code when developers negate boolean arguments. 

For example; 

When a code is wrriten as `Assert.IsTrue(!condition)` , then the Analyzer would suggest `Assert.IsFalse(condition)`

`Assert.IsFalse(!condition)` Analyzer would suggest `Assert.IsTrue(condition)`.

other edge cases were also considered, for example;
`Assert.IsTrue(!(a && b))` Analyzer would suggest `Assert.IsFalse(a && b)`

`Assert.IsFalse(!(x == y))`, Analyzer would suggest `Assert.IsTrue(x == y)`

`Assert.IsTrue(!!condition)` Analyzer would suggest `Assert.IsFalse(condition)`


src/Analyzers/MSTest.Analyzers.CodeFixes/DoNotNegateBooleanAssertionFixer.cs
```diff
@@ -0,0 +1,264 @@
...
+[Shared]
+public sealed class DoNotNegateBooleanAssertionFixer : CodeFixProvider
+{
+    /// <inheritdoc />
+    public override ImmutableArray<string> FixableDiagnosticIds { get; }
+        = ImmutableArray.Create(DiagnosticIds.DoNotNegateBooleanAssertionRuleId);
+
+    /// <inheritdoc />
+    public override FixAllProvider GetFixAllProvider()
+        => FixAll.Instance;
```
Evangelink: Please add a comment as to why you are not using the BatchFixer

src/Analyzers/MSTest.Analyzers.CodeFixes/DoNotNegateBooleanAssertionFixer.cs
```diff
@@ -0,0 +1,264 @@
...
+        editor.ReplaceNode(invocation, newInvocation);
+        return editor.GetChangedDocument();
+    }
+
+    private static ExpressionSyntax RemoveAllNegations(ExpressionSyntax expression)
+    {
+        // Walk down parentheses and negations until we find the core expression
+        ExpressionSyntax current = expression;
+
+        while (true)
```
Evangelink: Nit: I personally prefer to avoid `while (true)` anf favor having explicit stop condition.
I'll let @Youssef1313 decides what he prefers here.

Evangelink (commented):
Let's replace all calls to `VerifyAnalyzerAsync` from existing tests please

src/Analyzers/MSTest.Analyzers.CodeFixes/DoNotNegateBooleanAssertionFixer.cs
Youssef1313: I think batch fixer should actually be fine here?

src/Analyzers/MSTest.Analyzers.CodeFixes/DoNotNegateBooleanAssertionFixer.cs
Youssef1313: I'm not completely sure if this logic is correct on its current form, and I think the bug here is already demonstrated by the test `WhenAssertIsTrueWithDoubleNegation_Diagnostic` where we are changing semantics.

src/Analyzers/MSTest.Analyzers.CodeFixes/DoNotNegateBooleanAssertionFixer.cs
```diff
@@ -0,0 +1,264 @@
...
+
+        InvocationExpressionSyntax newInvocation = invocation
+            .WithExpression(newMemberAccess)
+            .WithArgumentList(newArgumentList);
+
+        editor.ReplaceNode(invocation, newInvocation);
+        return editor.GetChangedDocument();
+    }
+
+    private static ExpressionSyntax RemoveAllNegations(ExpressionSyntax expression)
```
Youssef1313: Instead of figuring out the syntax node that we need here, it's a better idea IMO to modify the analyzer itself to add that node as an additionalLocation. The codefix can then retrieve the node via the diagnostic additional locations.

src/Analyzers/MSTest.Analyzers.CodeFixes/DoNotNegateBooleanAssertionFixer.cs
AtolagbeMuiz: > I'm not completely sure if this logic is correct on its current form, and I think the bug here is already demonstrated by the test `WhenAssertIsTrueWithDoubleNegation_Diagnostic` where we are changing semantics.


my approach to resolving this is double negation is treeated just like a single negation.. so should be treated as one and get all the negations removed.. and fixes the code to `Assert.IsFalse`.. i.e. `Assert.IsTrue(!!condition)` ==> `Assert.IsFalse(condition)`...

similar to the same way `Assert.IsTrue((!condition))` ==> `Assert.IsFalse(condition)`...

or is this wrong?? If so, what how should the double negation be treated? @Youssef1313

src/Analyzers/MSTest.Analyzers.CodeFixes/DoNotNegateBooleanAssertionFixer.cs
AtolagbeMuiz: updated the pull request to use BatchFixer. @Evangelink @Youssef1313

src/Analyzers/MSTest.Analyzers.CodeFixes/DoNotNegateBooleanAssertionFixer.cs
Youssef1313: I wouldn't handle double negation as it's not a common case. Multiple iterations of the codefix will resolve it though.

AtolagbeMuiz: > I wouldn't handle double negation as it's not a common case. Multiple iterations of the codefix will resolve it though.

this was what i did initially but the test case failed for double negation.. and i implemented the test to handle it iteratively but the test case for double negation keeps failing.. because `IsFalse`/`IsTrue` changes to `IsTrue`/`IsFalse` on each iteration… which will make `Assert.IsTrue(!!condition)` change to `Assert.IsFalse(!condition)` in the first iteration and becomes `Assert.IsTrue(condition)`on second iteration.

@Youssef1313

AtolagbeMuiz: @Youssef1313 what do you conclude on this?

Youssef1313: Sorry I missed your comment.

> which will make Assert.IsTrue(!!condition) change to Assert.IsFalse(!condition) in the first iteration and becomes Assert.IsTrue(condition)on second iteration.

That sounds like the desirable behavior to me.

AtolagbeMuiz: okay, I will make the necessary corrections.

src/Analyzers/MSTest.Analyzers.CodeFixes/DoNotNegateBooleanAssertionFixer.cs
AtolagbeMuiz: > Instead of figuring out the syntax node that we need here, it's a better idea IMO to modify the analyzer itself to add that node as an additionalLocation. The codefix can then retrieve the node via the diagnostic additional locations.

this has been done and pull request updated @Youssef1313

src/Analyzers/MSTest.Analyzers.CodeFixes/DoNotNegateBooleanAssertionFixer.cs
AtolagbeMuiz: > Sorry I missed your comment.
> 
> > which will make Assert.IsTrue(!!condition) change to Assert.IsFalse(!condition) in the first iteration and becomes Assert.IsTrue(condition)on second iteration.
> 
> That sounds like the desirable behavior to me.

 I have implmenated this to handle multiple negations by iteration till all negations are removed.. PR has been updated as well.. also added another unit test for triple negation @Youssef1313

src/Analyzers/MSTest.Analyzers.CodeFixes/DoNotNegateBooleanAssertionFixer.cs
```diff
@@ -0,0 +1,368 @@
...
+        // Recursively unwrap all unnecessary parentheses
+        while (expression is ParenthesizedExpressionSyntax parenthesized)
+        {
+            expression = parenthesized.Expression;
+        }
+
+        return expression;
+    }
+
+    private sealed class FixAll : DocumentBasedFixAllProvider
```
Evangelink: I think this is unused and can be removed.

test/UnitTests/MSTest.Analyzers.UnitTests/DoNotNegateBooleanAssertionAnalyzerTests.cs
```diff
@@ -75,6 +75,619 @@ public void TestMethod()
...
+
+                private bool GetBoolean() => true;
+            }
+            """;
+
+        await VerifyCS.VerifyCodeFixAsync(code, fixedCode);
+    }
+
+    [TestMethod]
+    public async Task WhenAssertIsTrueWithNegation_Diagnostic()
```
Evangelink: I think this test case can be dropped as it's already covered above.

test/UnitTests/MSTest.Analyzers.UnitTests/DoNotNegateBooleanAssertionAnalyzerTests.cs
```diff
@@ -75,6 +75,619 @@ public void TestMethod()
...
+                    Assert.IsFalse(condition);
+                }
+            }
+            """;
+
+        await VerifyCS.VerifyCodeFixAsync(code, fixedCode);
+    }
+
+    [TestMethod]
+    public async Task WhenAssertIsFalseWithNegation_Diagnostic()
```
Evangelink: Same remark for this test case.

test/UnitTests/MSTest.Analyzers.UnitTests/DoNotNegateBooleanAssertionAnalyzerTests.cs
AtolagbeMuiz: this test case is now removed and pull request updated @Evangelink

test/UnitTests/MSTest.Analyzers.UnitTests/DoNotNegateBooleanAssertionAnalyzerTests.cs
AtolagbeMuiz: this test case is removed as well.. @Evangelink

src/Analyzers/MSTest.Analyzers.CodeFixes/DoNotNegateBooleanAssertionFixer.cs
AtolagbeMuiz: the unused `FixAll` removed

src/Analyzers/MSTest.Analyzers.CodeFixes/DoNotNegateBooleanAssertionFixer.cs
```diff
@@ -0,0 +1,368 @@
...
+
+        if (conditionArgumentIndex == -1)
+        {
+            return;
+        }
+
+        // Get the actual argument from the invocation
+        conditionArgument = invocation.ArgumentList.Arguments[conditionArgumentIndex];
+
+        string title = string.Format(System.Globalization.CultureInfo.InvariantCulture, Resources.DoNotNegateBooleanAssertionFix, properAssertMethodName);
```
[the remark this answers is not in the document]
AtolagbeMuiz: unnessary System.Globalization removed

Evangelink (commented):
@Youssef1313 please double check the code fixer.

src/Analyzers/MSTest.Analyzers.CodeFixes/DoNotNegateBooleanAssertionFixer.cs
```diff
@@ -0,0 +1,234 @@
...
+    /// <inheritdoc />
+    public override async Task RegisterCodeFixesAsync(CodeFixContext context)
+    {
+        Diagnostic diagnostic = context.Diagnostics[0];
+        SyntaxNode? root = await context.Document.GetSyntaxRootAsync(context.CancellationToken).ConfigureAwait(false);
+
+        if (root is null)
+        {
+            return;
+        }
```
Youssef1313:
```suggestion
        SyntaxNode root = await context.Document.GetRequiredSyntaxRootAsync(context.CancellationToken).ConfigureAwait(false);
```

src/Analyzers/MSTest.Analyzers/DoNotNegateBooleanAssertionAnalyzer.cs
```diff
@@ -64,7 +66,24 @@ private static void AnalyzeOperation(OperationAnalysisContext context, INamedTyp
...
+            string properAssertMethodName = currentMethodName == "IsTrue" ? "IsFalse" : "IsTrue";
+
+            // Add the proper method name to diagnostic properties for the code fix
+            ImmutableDictionary<string, string?>.Builder properties = ImmutableDictionary.CreateBuilder<string, string?>();
+            properties.Add(ProperAssertMethodNameKey, properAssertMethodName);
+
+            Location? conditionArgumentLocation = conditionArgument.Syntax.GetLocation();
+            ImmutableArray<Location> additionalLocations = conditionArgumentLocation != null
+                ? ImmutableArray.Create(conditionArgumentLocation)
+                : ImmutableArray<Location>.Empty;
```
Youssef1313: I think the null check here might be redundant.

src/Analyzers/MSTest.Analyzers/Resources.resx
```diff
@@ -684,6 +684,11 @@ The type declaring these methods should also respect the following rules:
   <data name="IgnoreStringMethodReturnValueDescription" xml:space="preserve">
     <value>Methods like Contains, StartsWith, and EndsWith return boolean values that indicate whether the condition was met. Ignoring these return values is likely a mistake.</value>
   </data>
+  <data name="DoNotNegateBooleanAssertionFix" xml:space="preserve">
+    <value>Use 'Assert.{0}' instead</value>
+  </data>
+  <data name="DoNotNegateBooleanAssertionFixAllTitle" xml:space="preserve">
+    <value>Use proper Assert methods for all negated assertions</value>
```
Youssef1313, line 691:
```suggestion
    <value>Use proper Assert methods for all negated assertions</value>
  </data>
```

src/Analyzers/MSTest.Analyzers.CodeFixes/DoNotNegateBooleanAssertionFixer.cs
```diff
@@ -0,0 +1,234 @@
...
+            }
+        }
+
+        if (conditionArgumentIndex == -1)
+        {
+            return;
+        }
+
+        // Get the actual argument from the invocation
+        conditionArgument = invocation.ArgumentList.Arguments[conditionArgumentIndex];
```
Youssef1313: What is the purpose of this? We already have the `ArgumentSyntax`. Why would it be different here?

test/UnitTests/MSTest.Analyzers.UnitTests/DoNotNegateBooleanAssertionAnalyzerTests.cs
```diff
@@ -75,6 +75,615 @@ public void TestMethod()
...
+            [TestClass]
+            public class MyTestClass
+            {
+                [TestMethod]
+                public void TestMethod()
+                {
+                    Assert.IsFalse(/* some comment */ false /* some other comment */);
+                }
+            }
+            """;
```
Youssef1313, line 687:
```suggestion
    [TestMethod]
    public async Task WhenAssertIsTrueWithNegation_PreservesComments()
    {
        string code = """
            using Microsoft.VisualStudio.TestTools.UnitTesting;

            [TestClass]
            public class MyTestClass
            {
                [TestMethod]
                public void TestMethod()
                {
                    // Comment on assert call
                    [|Assert.IsTrue(/* some comment */ !false /* some other comment */)|];
                }
            }
            """;

        string fixedCode = """
            using Microsoft.VisualStudio.TestTools.UnitTesting;

            [TestClass]
            public class MyTestClass
            {
                [TestMethod]
                public void TestMethod()
                {
                    // Comment on assert call
                    Assert.IsFalse(/* some comment */ false /* some other comment */);
                }
            }
            """;
```

src/Analyzers/MSTest.Analyzers.CodeFixes/DoNotNegateBooleanAssertionFixer.cs
```diff
@@ -0,0 +1,234 @@
...
+
+        // Get the proper assert method name from diagnostic properties
+        if (!diagnostic.Properties.TryGetValue(DoNotNegateBooleanAssertionAnalyzer.ProperAssertMethodNameKey, out string? properAssertMethodName) ||
+            properAssertMethodName is null)
+        {
+            return;
+        }
+
+        // Retrieve the condition argument from the additional locations provided by the analyzer
+        if (diagnostic.AdditionalLocations.Count == 0)
```
Youssef1313: Can this happen?

test/UnitTests/MSTest.Analyzers.UnitTests/DoNotNegateBooleanAssertionAnalyzerTests.cs
```diff
@@ -75,6 +75,615 @@ public void TestMethod()
...
+
+            [TestClass]
+            public class MyTestClass
+            {
+                [TestMethod]
+                public void TestMethod()
+                {
+                    bool b = true;
+            
+                    Assert.IsFalse(true);
```
Youssef1313, line 89: Do we have tests that use named arguments? e.g, `Assert.IsTrue(condition: !GetBoolean());` ?

If not, can we add one?

src/Analyzers/MSTest.Analyzers.CodeFixes/DoNotNegateBooleanAssertionFixer.cs
AtolagbeMuiz: that part is redundant, removed and pull request updated

src/Analyzers/MSTest.Analyzers/DoNotNegateBooleanAssertionAnalyzer.cs
AtolagbeMuiz: null check removed

src/Analyzers/MSTest.Analyzers.CodeFixes/DoNotNegateBooleanAssertionFixer.cs
AtolagbeMuiz: looking at the analyzer, it cant.. i removed that part..

test/UnitTests/MSTest.Analyzers.UnitTests/DoNotNegateBooleanAssertionAnalyzerTests.cs
AtolagbeMuiz: now added test cases for named arguments

PR #6831, opened 2025-10-30
Simplify MTP packaging

Youssef1313:
- In many packages, we shouldn't need `buildMultitargeting` because `buildMultitargeting` is only relevant for the "outer" build. TestingPlatformBuilderHook is only relevant for the inner builds.
- Instead of packing props/targets for each TFM, we can simply just pack in build/buildTransitive directly. We don't have any TFM-specific logic.

### DIFF
--- a/src/Platform/Microsoft.Testing.Extensions.HotReload/Microsoft.Testing.Extensions.HotReload.csproj
+++ b/src/Platform/Microsoft.Testing.Extensions.HotReload/Microsoft.Testing.Extensions.HotReload.csproj
@@ -32,6 +32,7 @@ This package extends Microsoft Testing Platform to provide Hot Reload support.]]
   <ItemGroup>
     <None Include="buildTransitive/**" PackagePath="buildTransitive" Pack="true" />
     <None Include="build/**" PackagePath="build" Pack="true" />
+    <None Include="$(RepoRoot)/LICENSE.PLATFORMTOOLS.txt" Pack="true" PackagePath="License.txt" />
   </ItemGroup>
 
   <ItemGroup>


src/Platform/Microsoft.Testing.Extensions.AzureDevOpsReport/Microsoft.Testing.Extensions.AzureDevOpsReport.csproj
```diff
@@ -16,16 +16,8 @@ This package extends Microsoft Testing Platform to provide a Azure DevOps report
...
-    <Content Include="buildMultiTargeting/**">
-      <Pack>true</Pack>
-      <PackagePath>buildMultiTargeting</PackagePath>
-    </Content>
-    <TfmSpecificPackageFile Include="buildTransitive/**">
-      <PackagePath>buildTransitive/$(TargetFramework)</PackagePath>
-    </TfmSpecificPackageFile>
-    <TfmSpecificPackageFile Include="build/**">
-      <PackagePath>build/$(TargetFramework)</PackagePath>
-    </TfmSpecificPackageFile>
```
Youssef1313, line 28: @nohwnd Here we were packing per-TFM. The updated logic doesn't.

Youssef1313:
> also if we don't have tfm specific folders ,it will make restore the package for any tfm.

Hmm, I thought it would fail if no compatible TFM is found in lib. Also note that for all modified packages, we target `netstandard2.0` which is almost compatible with the whole world (small exceptions are netcoreapp1.0, netcoreapp1.1, net45, MonoAndroid70, etc).

nohwnd:
hmm then I don't think there is a good solution. 
image

I am not 100% sure without trying (since it is an "edge-case"), but imho this will make the package restore for everything, we had this problem with one of vstest packages.

Secondly the netstandard2.0 compatibility is unfortunate, but you can do this what system.io.hashing does (dotnet/runtime  does that via some custom target generator) .

https://nuget.info/packages/System.IO.Hashing/10.0.0-rc.2.25502.107

```xml
<Project InitialTargets="NETStandardCompatError_System_IO_Hashing_net8_0">
  <Target Name="NETStandardCompatError_System_IO_Hashing_net8_0"
          Condition="'$(SuppressTfmSupportBuildWarnings)' == ''">
    <Warning Text="System.IO.Hashing 10.0.0-rc.2.25502.107 doesn't support $(TargetFramework) and has not been tested with it. Consider upgrading your TargetFramework to net8.0 or later. You may also set &lt;SuppressTfmSupportBuildWarnings&gt;true&lt;/SuppressTfmSupportBuildWarnings&gt; in the project file to ignore this warning and attempt to run in this unsupported configuration at your own risk." />
  </Target>
</Project>
```

they do that in buildTransitive/netstandard2.0 target.

I am not sure if we can combine this with actually supporting netstandard2.0 (unless we want to add netcore1.0 to tfms :D)

PR #6852, opened 2025-11-01
Unskip tests on macOS

Youssef1313:


Youssef1313:
It's supposed to work if we run under dotnet executable instead of the apphost. I want to have better understanding of what's going on.

PR #6861, opened 2025-11-02
Improve MTP unit tests

Youssef1313:

### DIFF
--- a/test/UnitTests/Microsoft.Testing.Platform.UnitTests/Helpers/SystemAsyncMonitorTests.cs
+++ b/test/UnitTests/Microsoft.Testing.Platform.UnitTests/Helpers/SystemAsyncMonitorTests.cs
@@ -24,7 +24,7 @@ public sealed class SystemAsyncMonitorTests
 
         await Task.WhenAll([.. tasks]);
 
-        Assert.IsGreaterThan(3000, stopwatch.ElapsedMilliseconds);
+        Assert.IsGreaterThanOrEqualTo(3000, stopwatch.ElapsedMilliseconds);
 
         async Task TestLock()
         {


test/UnitTests/Microsoft.Testing.Platform.UnitTests/Helpers/SystemAsyncMonitorTests.cs
```diff
@@ -17,15 +17,19 @@ public async Task AsyncMonitor_ShouldCorrectlyLock()
...
-        }
-
-        await Task.WhenAll([.. tasks]);
+            for (int i = 0; i < 3; i++)
+            {
+                tasks.Add(Task.Run(TestLock, TestContext.CancellationToken));
+            }
 
-        // Give more time to be above 3s
-        Thread.Sleep(500);
```
Youssef1313, line 28: This is an unnecessary delay. The test seems reliable without it.
If we found the test to be flaky in future, what we can do instead is adjust the assert below to use `Assert.IsGreaterThan(2500, stopwatch.ElapsedMilliseconds)`

test/UnitTests/Microsoft.Testing.Platform.UnitTests/Helpers/SystemAsyncMonitorTests.cs
```diff
@@ -17,15 +17,19 @@ public async Task AsyncMonitor_ShouldCorrectlyLock()
...
+            await Task.WhenAll([.. tasks]);
+        }
+        finally
+        {
+            stopwatch.Stop();
```
Youssef1313: The stopwatch previously wasn't stopped. I think it's better practice to stop it.

test/UnitTests/Microsoft.Testing.Platform.UnitTests/Helpers/TaskExtensionsTests.cs
```diff
@@ -72,49 +72,45 @@ public async Task CancellationAsync_NonCanceledWithArgument_Succeeds()
 
     [TestMethod]
     public async Task CancellationAsync_ObserveException_Succeeds()
-        => await RetryHelper.RetryAsync(
```
Youssef1313, line 75: I'm avoiding Retry here. If this test is flaky, we would want to know about it instead of masking a potential issue :)

test/UnitTests/Microsoft.Testing.Platform.UnitTests/Helpers/TaskExtensionsTests.cs
```diff
@@ -72,49 +72,45 @@ public async Task CancellationAsync_NonCanceledWithArgument_Succeeds()
...
-                            throw new InvalidOperationException();
-                        }, TestContext.CancellationToken).WithCancellationAsync(new CancellationTokenSource(TimeSpan.FromSeconds(1)).Token));
+    {
+        ManualResetEvent waitException = new(false);
+        CancellationToken token = new CancellationTokenSource(TimeSpan.FromSeconds(1)).Token;
+        OperationCanceledException ex = await Assert.ThrowsAsync<OperationCanceledException>(async ()
+            => await Task.Run(
+                async () =>
+                {
+                    await Task.Delay(TimeSpan.FromSeconds(5), TestContext.CancellationToken);
```
Youssef1313, line 82: Going down a bit with the delay from 10 seconds to only 5 seconds.

test/UnitTests/Microsoft.Testing.Platform.UnitTests/Helpers/TaskExtensionsTests.cs
```diff
@@ -72,49 +72,47 @@ public async Task CancellationAsync_NonCanceledWithArgument_Succeeds()
...
+                    waitException.Set();
+                    throw new InvalidOperationException();
+                }, TestContext.CancellationToken).WithCancellationAsync(token));
+#if !NETFRAMEWORK // Polyfill bug in Task.WhenAsync implementation :/
```
Youssef1313: FYI @SimonCropp

I also think the whole implementation here is not correct. It feels to me like if cancellation was requested in the middle of the original task, it won't immediately return a completed task.

test/UnitTests/Microsoft.Testing.Platform.UnitTests/Helpers/SystemAsyncMonitorTests.cs
SimonCropp: but there is nothing counting under the covers. stop is redundant unless u want to use the timespan some time after the stop. if u just need the timespan at that instant, then Stop is redundant code

test/UnitTests/Microsoft.Testing.Platform.UnitTests/Helpers/SystemAsyncMonitorTests.cs
```diff
@@ -17,15 +17,19 @@ public async Task AsyncMonitor_ShouldCorrectlyLock()
...
-
-        await Task.WhenAll([.. tasks]);
+            for (int i = 0; i < 3; i++)
+            {
+                tasks.Add(Task.Run(TestLock, TestContext.CancellationToken));
+            }
 
-        // Give more time to be above 3s
-        Thread.Sleep(500);
+            await Task.WhenAll([.. tasks]);
```
SimonCropp: I know this was not change in the PR. but cant tasks be passed in directly?

test/UnitTests/Microsoft.Testing.Platform.UnitTests/Helpers/TaskExtensionsTests.cs
SimonCropp: @Youssef1313 thanks. will take a look tonight

test/UnitTests/Microsoft.Testing.Platform.UnitTests/Helpers/TaskExtensionsTests.cs
```diff
@@ -72,49 +72,47 @@ public async Task CancellationAsync_NonCanceledWithArgument_Succeeds()
...
 
     [TestMethod]
     public async Task CancellationAsyncWithReturnValue_ObserveException_Succeeds()
-        => await RetryHelper.RetryAsync(
-            async () =>
+    {
+        ManualResetEvent waitException = new(false);
+        CancellationToken token = new CancellationTokenSource(TimeSpan.FromSeconds(1)).Token;
+        OperationCanceledException ex = await Assert.ThrowsAsync<OperationCanceledException>(async ()
+            => await Task.Run(async () =>
```
SimonCropp, line 98: can we elide the asyncs here?

test/UnitTests/Microsoft.Testing.Platform.UnitTests/Helpers/SystemAsyncMonitorTests.cs
Youssef1313: Indeed. Let me clean that up back again

test/UnitTests/Microsoft.Testing.Platform.UnitTests/Helpers/SystemAsyncMonitorTests.cs
Youssef1313: Probably it can if `tasks` switched from `List<Task>` instead `Task[]`

test/UnitTests/Microsoft.Testing.Platform.UnitTests/Helpers/TaskExtensionsTests.cs
Youssef1313: I don't think it brings much value so I prefer to keep the test as-is.

test/UnitTests/Microsoft.Testing.Platform.UnitTests/Helpers/TaskExtensionsTests.cs
SimonCropp: what method? since `Task.WhenAsync` isnt a thing

SimonCropp: or better yet. can u share a test that illustrates the problem

Youssef1313: Sorry, meant `WaitAsync`.

https://github.com/SimonCropp/Polyfill/blob/3f9ccd279a04bfa25d592e576cc1099fd56a36ef/src/Polyfill/Polyfill_Task.cs#L18

SimonCropp: can u give me a failing unit test?

Youssef1313: I think something along the lines of:

```csharp
var cts = new CancellationTokenSource(TimeSpan.FromSeconds(1));
var stopwatch = Stopwatch.StartNew();
try
{    
    await Task.Delay(10000).WaitAsync(cts.Token);
}
catch (TaskCanceledException ex)
{
    // ex.CancellationToken == cts.Token should be true.
}

// stopwatch.ElapsedMilliseconds should be around the 1 second. Probably just a little bit larger. But should be much less than 10000.
```

PR #6863, opened 2025-11-03
Fix CI after image update to a stable VS

Youssef1313:

### DIFF
--- a/test/IntegrationTests/Microsoft.Testing.Platform.Acceptance.IntegrationTests/Helpers/AcceptanceTestBase.cs
+++ b/test/IntegrationTests/Microsoft.Testing.Platform.Acceptance.IntegrationTests/Helpers/AcceptanceTestBase.cs
@@ -160,12 +160,13 @@ public class UnitTest1
             string installationVersion = await RunAndGetSingleLineStandardOutputAsync(vswherePath, "-property installationVersion", cancellationToken);
             string instanceId = await RunAndGetSingleLineStandardOutputAsync(vswherePath, "-property instanceId", cancellationToken);
             var version = Version.Parse(installationVersion);
-            string settingsFilePath = Path.Combine(
+            string directory = Path.Combine(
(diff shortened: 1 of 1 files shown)


test/IntegrationTests/Microsoft.Testing.Platform.Acceptance.IntegrationTests/Helpers/AcceptanceTestBase.cs
```diff
@@ -151,8 +151,27 @@ public static void ClassCleanup()
...
+        string installationVersion = await RunAndGetSingleLineStandardOutputAsync(vswherePath, "-property installationVersion", cancellationToken);
+        string instanceId = await RunAndGetSingleLineStandardOutputAsync(vswherePath, "-property instanceId", cancellationToken);
+        var version = Version.Parse(installationVersion);
+        string settingsFilePath = Path.Combine(
+            Environment.GetFolderPath(Environment.SpecialFolder.LocalApplicationData),
+            "Microsoft",
+            "VisualStudio",
+            version.Major + ".0_" + instanceId,
+            "sdk.txt");
+        File.WriteAllText(settingsFilePath, "UsePreviews=True");
```
nohwnd: this will keep adding that config over and over on my local computer?

Youssef1313: If you use stable VS, yes :/

Youssef1313:
Alternatively, we need to refactor our pipeline to avoid calling into `eng\common\CIBuild.cmd` directly so that we have a chance using our own scripts that does similar thing as Roslyn, and do it only when `-ci` is used.

Youssef1313:
@nohwnd I hacked it for now so that it's CI-only.

Eventually I would like one of the following to happen:

1. SDK is able to tell from global.json that it should allow prerelease.
2. This hack is integrated into Arcade, so we eventually remove it.

PR #6865, opened 2025-11-03
Unskip HangDumpProcessTreeTests on macos

Youssef1313:


Evangelink (approved):

test/IntegrationTests/Microsoft.Testing.Platform.Acceptance.IntegrationTests/HangDumpProcessTreeTests.cs
```diff
@@ -52,7 +58,6 @@ public sealed class TestAssetFixture() : TestAssetFixtureBase(AcceptanceFixture.
   <PropertyGroup>
     <TargetFrameworks>$TargetFrameworks$</TargetFrameworks>
     <OutputType>Exe</OutputType>
-    <UseAppHost>true</UseAppHost>
```
nohwnd, line 55: why is this removed? the global property should override it. Is that to keep it "os specific" if the build does / will do that?

Youssef1313: UseAppHost already defaults to true. It was redundant.

I'm only setting it in one place (global prop), if needed, for clarity.

PR #6866, opened 2025-11-03
Cleanup MSTestAdapter.PlatformServices.UnitTests

Youssef1313:


Evangelink (commented):
I think we are not standard but the idea was often to use source namespace + `.UnitTests`.

Issue #3933, opened 2024-10-11
Standard output is provided in two different properties on test node


nohwnd:
In https://github.com/microsoft/testfx/pull/3932 we put standard output into two properties on test node, those are potentially both serialized when calling a client, check if that is the case, and make sure that newer TE client can consume the platform provided property so we don't have to send both forever. 

See this conversation for details ( https://github.com/microsoft/testfx/pull/3932#discussion_r1797016815 ) 

[nohwnd](https://github.com/nohwnd) [1 hour ago](https://github.com/microsoft/testfx/pull/3932#discussion_r1797016815)
These are potentially much larger than the rest of the message, is it good idea to have 2 copies of the text in the object? Is this the only way to do this? can "vstest.TestCase.StandardError" be dropped in favor of the new Platform provided property?

Member
@[MarcoRossignoli](https://github.com/MarcoRossignoli) MarcoRossignoli [1 hour ago](https://github.com/microsoft/testfx/pull/3932#discussion_r1797022689) • 
We cannot we need 2 way because the SerializableKeyValuePairStringProperty is an internal prop to accomodate VS. VS TE is working in 2 different mode where the values are taken using different key, value.
Under VS protocol we declare vstest capablity and that value is the expected one.

Adding StandardErrorProperty like xunit did is to have the new model and avoid the need to go with an untyped "key name" that's also used in the protocol atm as vstest capability. Otherwise the core platform should embed code related to the protocol for the vstest capability.

Member
@[MarcoRossignoli](https://github.com/MarcoRossignoli) MarcoRossignoli [1 hour ago](https://github.com/microsoft/testfx/pull/3932#discussion_r1797026918)
My idea was to drop the vstest capability, but looks like at time wasn't good for TE. cc: @drognanar

Member
@[nohwnd](https://github.com/nohwnd) nohwnd [1 hour ago](https://github.com/microsoft/testfx/pull/3932#discussion_r1797045150)
So the solution here is fix the VS client, and then remove the vstest.* property, yeah?

Member
@[nohwnd](https://github.com/nohwnd) nohwnd [52 minutes ago](https://github.com/microsoft/testfx/pull/3932#discussion_r1797049667)
Do you know if we end up serializing both when calling TE or some other client? If this is not the case then the overhead of having two "copies" of the same string on the node is minimal, but I doubt that is the case, and we actually send it twice.

If you don't know off hand just say it , I will make an issue for investigating it.

Evangelink:
[...]

PR #6872, opened 2025-11-04
VSTestBridge: Don't add TRX properties when TRX not enabled

Youssef1313:
Related to https://github.com/microsoft/testfx/issues/3933

### DIFF
--- a/src/Adapter/MSTest.TestAdapter/TestingPlatformAdapter/TestApplicationBuilderExtensions.cs
+++ b/src/Adapter/MSTest.TestAdapter/TestingPlatformAdapter/TestApplicationBuilderExtensions.cs
@@ -3,6 +3,7 @@
 
 #if !WINDOWS_UWP
 using Microsoft.Testing.Extensions.TrxReport.Abstractions;
+using Microsoft.Testing.Extensions.VSTestBridge.Capabilities;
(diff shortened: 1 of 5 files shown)


src/Adapter/MSTest.TestAdapter/TestingPlatformAdapter/TestApplicationBuilderExtensions.cs
```diff
@@ -19,13 +20,14 @@ public static class TestApplicationBuilderExtensions
 {
     // NOTE: We intentionally use this class and not VSTestBridgeExtensionBaseCapabilities because
     // we don't want MSTest to use vstestProvider capability
```
nohwnd, line 22: does this comment still apply, it seems a bit confusing when the new interface has "vstest" in name.

Youssef1313: Yes. The comment still applies.

vstestProvider affects more things, and we don't want to use it with MSTest+MTP.

nohwnd:
Okay makes sense. I thought it was something like that. I have bunch of such "tricks" in vstest, and I really don't like them. One of the classes is private so no problem there, and the other is public but tied to vstest bridge it seems. So we should not have hard time removing that interface.

PR #6887, opened 2025-11-05
Avoid BlockingCollection in AsyncConsumerDataProcessor

### LINKED ISSUE #6858: `AsyncConsumerDataProcessor.ConsumeAsync` should avoid running on threadpool
Currently, we run `AsyncConsumerDataProcessor.ConsumeAsync` on threadpool. We should avoid that.

User test code might end up blocking all threadpool threads, causing the platform to not be able to make any further progress.

We should also run some tests with `DOTNET_ThreadPool_ForceMinWorkerThreads` and `DOTNET_ThreadPool_ForceMaxWorkerThreads` environment variables to 1. Such tests could also test for a user code that gets a threadpool thread and block it. These tests should also consider running with some of the extensions we have, e.g, trx, hangdump, etc...

Youssef1313:
Fixes #6858

We are still running on threadpool, but we are avoiding blocking it via `BlockingCollection`.

There are mainly two things going on here:

1. The intended fix -> avoid blocking threadpool threads. This can be seen when there is number of data consumers higher than Environment.ProcessorCount, and then we are exhausting the threadpool and slowing any other operations that need to run on the threadpool (i.e, waiting for the TP to be able to inject more threads).
2. General perf improvement unrelated to the fact that we are blocking threadpool (this was unintended :smile:)

For 2, I'm testing two channel implementations.

- Channel1 -> the implementation in this PR.
- Channel2 -> a fake implementation that is wrapping BlockingCollection, simulating what we have in MTP today.
- .NET channel -> using System.Threading.Channels 9.0.10 (but we cannot use it because we don't want to have any dependencies).

```csharp
// --- First metric: time taken to write and read a number of items. ---
// First scenario, writing 50 millions items in a tight loop.
// Channel1: 2.3 seconds.
// Channel2: 9 seconds.
// .NET channel: 1.4 seconds.
//
// Second scenario, writing 20 millions items, with Thread.Yield in between.
// Channel1: 7.3 seconds.
// Channel2: 8 seconds
// .NET channel: 6.1 seconds

// --- Second metric: how many items we can process per second ---
// First scenario, writing infinite items in a tight loop.
// Channel1: 15million/sec
// Channel2: 5million/sec
// .NET channel: 25million/sec
//
// Second scenario, writing infinite items, with Thread.Yield in between.
// Channel1: 2.5million/sec
// Channel2: 2.3million/sec
// .NET channel: 3million/sec
```

Channel2 (BlockingCollection) seems to be the worse in all tested scenarios.

App used for measuring:

[ConsoleApp25.zip](https://github.com/user-attachments/files/23394779/ConsoleApp25.zip)

### DIFF
--- a/test/IntegrationTests/Microsoft.Testing.Platform.Acceptance.IntegrationTests/DataConsumerThroughputTests.cs
+++ b/test/IntegrationTests/Microsoft.Testing.Platform.Acceptance.IntegrationTests/DataConsumerThroughputTests.cs
@@ -13,7 +13,7 @@ public class DataConsumerThroughputTests : AcceptanceTestBase<DataConsumerThroug
     public async Task MultipleDataConsumers_ShouldCompleteInReasonableTime(string tfm)
     {
         var testHost = TestInfrastructure.TestHost.LocateFrom(AssetFixture.TargetAssetPath, AssetName, tfm);
-        
+
         var stopwatch = System.Diagnostics.Stopwatch.StartNew();
         TestHostResult testHostResult = await testHost.ExecuteAsync(cancellationToken: TestContext.CancellationToken);
         stopwatch.Stop();


src/Platform/Microsoft.Testing.Platform/Messages/ConsumingEnumerableConsumerDataProcessor.cs
```diff
@@ -36,30 +37,42 @@ public AsyncConsumerDataProcessor(IDataConsumer dataConsumer, ITask task, Cancel
...
+        _signal.Release();
         return Task.CompletedTask;
     }
 
     private async Task ConsumeAsync()
     {
         try
         {
-            foreach ((IDataProducer dataProducer, IData data) in _payloads.GetConsumingEnumerable(_cancellationToken))
+            while (!_isAddingCompleted || _signal.CurrentCount > 0 || !_payloads.IsEmpty)
```
Youssef1313: What if we are in this situation:

1. Thread A: is about to evaluate this condition.
2. Thread B: is about to call complete adding.
3. Thread C: is about to publish.
4. Thread C got past the _isAddingCompleted check, but didn't yet add to the queue.
5. Thread B called complete adding and set the flag to true.
6. Thread A evaluates this to false, because adding is completed, and the queues are empty and no signals are there.

That means that something is getting published but not consumed. That probably isn't very expected?
Previously, BlockingCollection was synchronizing all these operations internally.

I think we need to think about the proper mechanisms to synchronize this correctly.

Also not sure if `volatile` makes sense for `_isAddingCompleted` flag?

src/Platform/Microsoft.Testing.Platform/Messages/SingleConsumerUnboundedChannel.cs
```diff
@@ -0,0 +1,112 @@
...
+    private TaskCompletionSource<bool>? _waitingReader;
+
+    private CancellationTokenRegistration? _cancellationRegistration;
+
+    // A flag indicating whether or not complete has been called.
+    private bool _completed;
+
+    // It's safe to lock on _items rather than a dedicated lock object.
+    // It's a private field and we own it.
+    // There is no need to extra an extra object for no value.
```
nohwnd:
```suggestion
    // There is no need to add an extra object for no value.
```

src/Platform/Microsoft.Testing.Platform/Messages/AsyncConsumerDataProcessor.netstandard.cs
```diff
@@ -36,59 +33,58 @@ public AsyncConsumerDataProcessor(IDataConsumer dataConsumer, ITask task, Cancel
...
+                        try
+                        {
+                            await DataConsumer.ConsumeAsync(item.DataProducer, item.Data, _cancellationToken).ConfigureAwait(false);
+                        }
+
+                        // We let the catch below to handle the graceful cancellation of the process
+                        catch (Exception ex) when (ex is not OperationCanceledException)
+                        {
+                            // If we're draining before to increment the _totalPayloadProcessed we need to signal that we should throw because
+                            // it's possible we have a race condition where the payload check at line 106 return false and the current task is not yet in a
```
nohwnd, line 69: "at line 106" did not age well. Do you see what this was referring to?

src/Platform/Microsoft.Testing.Platform/Messages/SingleConsumerUnboundedChannel.cs
```diff
@@ -0,0 +1,112 @@
...
+                _cancellationRegistration = null;
+                waitingReader.TrySetResult(true);
+            }
+        }
+    }
+
+    public bool TryRead(out T item)
+        => _items.TryDequeue(out item);
+
+    public Task<bool> WaitToReadAsync(CancellationToken cancellationToken)
```
nohwnd, line 58: should this check the cancellation token being already cancelled and act on that?

src/Platform/Microsoft.Testing.Platform/Messages/SingleConsumerUnboundedChannel.cs
```diff
@@ -0,0 +1,112 @@
...
+            }
+
+            if (_waitingReader is not null)
+            {
+                // Two calls happened to WaitToReadAsync without anything being written in between.
+                // This indicates multiple consumers (or single consumer calling this twice without await), which should never happen.
+                throw new UnreachableException();
+            }
+
+            _waitingReader = new TaskCompletionSource<bool>(TaskCreationOptions.RunContinuationsAsynchronously);
```
nohwnd, line 98: having it explained why this option is necessary would be nice for the future.

src/Platform/Microsoft.Testing.Platform/Messages/SingleConsumerUnboundedChannel.cs
```diff
@@ -0,0 +1,112 @@
...
+            _cancellationRegistration = cancellationToken.Register(static (tcs, ct) => ((TaskCompletionSource<bool>)tcs!).TrySetCanceled(ct), _waitingReader);
```
nohwnd: You said this is polyfilled, maybe  it would be worth it to expand that to the full code manually, and explain that we cannot avoid the allocation from capturing the values?

So we are not "surprised" by it in the future.

src/Platform/Microsoft.Testing.Platform/Messages/SingleConsumerUnboundedChannel.cs
```diff
@@ -0,0 +1,112 @@
...
+    private TaskCompletionSource<bool>? _waitingReader;
+
+    private CancellationTokenRegistration? _cancellationRegistration;
+
+    // A flag indicating whether or not complete has been called.
+    private bool _completed;
+
+    // It's safe to lock on _items rather than a dedicated lock object.
+    // It's a private field and we own it.
+    // There is no need to add an extra object for no value.
```
nohwnd:
```suggestion
    // This avoids allocating additional object that adds no value.
```

src/Platform/Microsoft.Testing.Platform/Messages/SingleConsumerUnboundedChannel.cs
Youssef1313: Done.

src/Platform/Microsoft.Testing.Platform/Messages/AsyncConsumerDataProcessor.netstandard.cs
Youssef1313: I'll try to figure it out.

src/Platform/Microsoft.Testing.Platform/Messages/SingleConsumerUnboundedChannel.cs
Youssef1313: I agree.
tbh, I think Polyfill is doing much more than it should, to the point that we can end up with hidden issues due to it.

I might even consider finding a way to "whitelist" what APIs we use from polyfill.

src/Platform/Microsoft.Testing.Platform/Messages/SingleConsumerUnboundedChannel.cs
Youssef1313: Added a comment.

src/Platform/Microsoft.Testing.Platform/Messages/AsyncConsumerDataProcessor.netstandard.cs
Youssef1313: I have no idea. But for history: https://github.com/microsoft/testanywhere/pull/1439

src/Platform/Microsoft.Testing.Platform/Messages/ConsumingEnumerableConsumerDataProcessor.cs
```diff
@@ -10,17 +10,14 @@ namespace Microsoft.Testing.Platform.Messages;
 
 internal sealed class AsyncConsumerDataProcessor : IDisposable
```
Youssef1313: TODO: Hide the previous implementation under env variable, just in case something breaks.

src/Platform/Microsoft.Testing.Platform/Messages/AsyncConsumerDataProcessor.netstandard.cs
Youssef1313, line 1: Also TODO: add a test for the crafted scenario in https://github.com/microsoft/testfx/pull/6887#issuecomment-3494047668 that verifies it works as expected with the change.

Youssef1313: @copilot Please add a test for me.

Youssef1313:
> We don't have also that amount of consumers shipped at the moment so we don't see this perf issue.

We actually see it on CI (in the form of flaky MSTest test where we expect cancellation to happen in a timely manner but the callback is likely delayed due to this) where the machines are 4 core only and we enable many extensions (trx, azdo report, hangdump, DotnetTestDataConsumer, and we have at least two data consumers that are always plugged: TerminalOutputDevice and `TestApplicationResult`).

PR #6901, opened 2025-11-07
Run pipeline on all PRs regardless of branches

Youssef1313:
In Copilot PRs that target other branches that I work on, I would expect the pipeline to be run.


Evangelink (approved):

Youssef1313:
It's not running at all in this PR. I'll need to look later into why. It's not important so far.

Youssef1313:
Actually I think I know. Let me try again

PR #6973, opened 2025-11-22
Analyzer suggestion for duplicate test methods

### LINKED ISSUE #6813: [Analyzer suggestion]: Duplicate test methods
I have seen it few times that test methods get copied without changes, unintentionally.

I think it might be a good idea to try to implement something for detecting duplicate test methods where only the method name differs, possibly only in the same test class for perf reasons.

AtolagbeMuiz:
This pull request fixes #6813 

This new analyzer implementation allows similar unit test methods with different names within the same test class to be detected as duplicates.

**NB**: Because this is a new analyzer feature, I declared a new DIagnosticId value of `MSTEST0059` as `DoNotDuplicateTestMethodRuleId`

### DIFF
--- a/src/Analyzers/MSTest.Analyzers/DoNotDuplicateTestMethodAnalyzer.cs
+++ b/src/Analyzers/MSTest.Analyzers/DoNotDuplicateTestMethodAnalyzer.cs
@@ -13,7 +13,7 @@ using MSTest.Analyzers.Helpers;
 namespace MSTest.Analyzers;
 
 /// <summary>
-/// MSTEST0059: <inheritdoc cref="Resources.DoNotDuplicateTestMethodTitle"/>.
+/// MSTEST0062: <inheritdoc cref="Resources.DoNotDuplicateTestMethodTitle"/>.
 /// Detects test methods with different names but identical or very similar implementations.
 /// </summary>
 [DiagnosticAnalyzer(LanguageNames.CSharp, LanguageNames.VisualBasic)]
--- a/src/Analyzers/MSTest.Analyzers/AnalyzerReleases.Unshipped.md
+++ b/src/Analyzers/MSTest.Analyzers/AnalyzerReleases.Unshipped.md
@@ -6,7 +6,7 @@
 Rule ID | Category | Severity | Notes
 --------|----------|----------|-------
 MSTEST0058 | Usage | Info | AvoidAssertsInCatchBlocksAnalyzer, [Documentation](https://learn.microsoft.com/dotnet/core/testing/mstest-analyzers/mstest0058)
-MSTEST0059 | Usage | Warning | DoNotDuplicateTestMethodAnalyzer, [Documentation](https://learn.microsoft.com/dotnet/core/testing/mstest-analyzers/mstest0059)
 MSTEST0059 | Usage | Warning | DoNotUseParallelizeAndDoNotParallelizeTogetherAnalyzer, [Documentation](https://learn.microsoft.com/dotnet/core/testing/mstest-analyzers/mstest0059)
 MSTEST0060 | Usage | Warning | DuplicateTestMethodAttributeAnalyzer, [Documentation](https://learn.microsoft.com/dotnet/core/testing/mstest-analyzers/mstest0060)
 MSTEST0061 | Usage | Info | UseOSConditionAttributeInsteadOfRuntimeCheckAnalyzer, [Documentation](https://learn.microsoft.com/dotnet/core/testing/mstest-analyzers/mstest0061)
+MSTEST0062 | Usage | Warning | DoNotDuplicateTestMethodAnalyzer, [Documentation](https://learn.microsoft.com/dotnet/core/testing/mstest-analyzers/mstest0062)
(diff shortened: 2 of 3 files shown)


src/Analyzers/MSTest.Analyzers/DoNotDuplicateTestMethodAnalyzer.cs
```diff
@@ -0,0 +1,269 @@
...
+
+using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.Diagnostics;
+
+using MSTest.Analyzers.Helpers;
+
+namespace MSTest.Analyzers;
+
+/// <summary>
+/// MSTEST0033: <inheritdoc cref="Resources.DoNotDuplicateTestMethodTitle"/>.
```
Youssef1313:
```suggestion
/// MSTEST0059: <inheritdoc cref="Resources.DoNotDuplicateTestMethodTitle"/>.
```

src/Analyzers/MSTest.Analyzers/DoNotDuplicateTestMethodAnalyzer.cs
```diff
@@ -0,0 +1,269 @@
...
+            if (compilationContext.Compilation.TryGetOrCreateTypeByMetadataName(WellKnownTypeNames.MicrosoftVisualStudioTestToolsUnitTestingTestClassAttribute, out INamedTypeSymbol? testClassAttributeSymbol)
+                && compilationContext.Compilation.TryGetOrCreateTypeByMetadataName(WellKnownTypeNames.MicrosoftVisualStudioTestToolsUnitTestingTestMethodAttribute, out INamedTypeSymbol? testMethodAttributeSymbol))
+            {
+                ConcurrentBag<(IMethodSymbol Symbol, SyntaxNode Syntax)> testMethodsFound = [];
+
+                compilationContext.RegisterSymbolAction(
+                    context => CollectTestMethod(context, testClassAttributeSymbol, testMethodAttributeSymbol, testMethodsFound),
+                    SymbolKind.Method);
+
+                compilationContext.RegisterCompilationEndAction(
```
Youssef1313, line 60: Instead of being compilation end, can it instead of a symbol start/end analyzer? You first `RegisterSymbolStartAction` for named types, then `RegisterOperationBlockAction` to analyze the individual methods, and then `RegisterSymbolEndAction` to do the final analysis of the collected methods?

src/Analyzers/MSTest.Analyzers/DoNotDuplicateTestMethodAnalyzer.cs
```diff
@@ -0,0 +1,269 @@
...
+            currentType = currentType.BaseType;
+        }
+
+        return false;
+    });
+
+    private static SyntaxNode? GetMethodBody(SyntaxNode methodNode)
+    {
+        // Try block body first
+        System.Reflection.PropertyInfo? bodyProperty = methodNode.GetType().GetProperty("Body");
```
Youssef1313, line 91: Using reflection here doesn't seem like a good idea. This also actually might fail for VB which the analyzer claims to support

src/Analyzers/MSTest.Analyzers/DoNotDuplicateTestMethodAnalyzer.cs
```diff
@@ -0,0 +1,269 @@
...
+            var methods = typeGroup
+                .OrderBy(m => m.Symbol.Locations.FirstOrDefault()?.SourceTree?.FilePath ?? string.Empty)
+                .ThenBy(m => m.Symbol.Locations.FirstOrDefault()?.SourceSpan.Start ?? int.MaxValue)
+                .ToList();
+
+            // Compare each pair of test methods for duplicate implementations
+            for (int i = 0; i < methods.Count; i++)
+            {
+                for (int j = i + 1; j < methods.Count; j++)
+                {
```
Youssef1313, line 203: This looks like very extensive computations. Probably there is something we can do better here.

Youssef1313: @CyrusNajmabadi if you can suggest something here please. This PR attempts to implement https://github.com/microsoft/testfx/issues/6813

CyrusNajmabadi: yup.  i have major concerns here. 

1. n^2 on number of methods.  Which can be huge.  We def have tests classes with thousands of items.
2. edit-distance on string representation.  Which can also be huge and n^2
3. major allocs for each pair.   Things like `int[,] d = new int[s1.Length + 1, s2.Length + 1];` can be multi-mb allocs per pair.

I don't know of some great solution here off the top of my head.  But i think some research needs to be done in state of teh art duplication detection, and seeing what might apply here.

As an example, if each method was fingerprinted somehow (maybe a vector embedding), then only the methods with the closest embedding value would be comapred.  eetc. etc.

AtolagbeMuiz:
observed another property` DoNotUseParallelizeAndDoNotParallelizeTogetherRuleId` has taken the value `MSTEST0059`.. 
image

I will be changing the DisagnosticId to `MSTEST0062` @Youssef1313

Evangelink:
@AtolagbeMuiz based on the comment from @CyrusNajmabadi, we won't merge the PR as-is and we would need to do some research on state of the art algorithms for distance and computation. Given we don't have much time, I will move by closing this PR.

Thanks for the time invested!

PR #6989, opened 2025-11-26
Report the actual exception from assertion

### LINKED ISSUE #6906: Assert.Throws should print the actual exception.
## Summary

```
 Assert.Throws failed. Expected exception type:<System.ArgumentException>. Actual exception type:<System.InvalidOperationException>. 'action' expression: '() => new MyClass().MyMethod()'.
    at Microsoft.VisualStudio.TestTools.UnitTesting.Assert.<>c__DisplayClass226_1`1.<IsThrowsFailing>b__1(String userMessage) in /_/src/TestFramework/TestFramework/Assertions/Assert.ThrowsException.cs:581
    at Microsoft.VisualStudio.TestTools.UnitTesting.Assert.ThrowsException[TException](Action action, Boolean isStrictType, String message, String actionExpression, String assertMethodName) in /_/src/TestFramework/TestFramework/Assertions/Assert.ThrowsException.cs:330
    at Microsoft.VisualStudio.TestTools.UnitTesting.Assert.Throws[TException](Action action, String message, String actionExpression) in /_/src/TestFramework/TestFramework/Assertions/Assert.ThrowsException.cs:187
    at mstest306.Test1.TestMethod1() in Test1.cs:9
    at System.Reflection.MethodBaseInvoker.InterpretedInvoke_Method(Object obj, IntPtr* args)
    at System.Reflection.MethodBaseInvoker.InvokeWithNoArgs(Object obj, BindingFlags invokeAttr)
bin\Debug\net10.0\mstest306.dll (net10.0|x64) failed with 1 error(s) (566ms)
```

is what I see currently when Assert.Throws fails. This is not giving it all the information I need to find the failure, and forces me to guess where the error happened. Am I seeing the friday error, or other error? what if the problem is not easily replicable?

Instead it would be nice if we reported the whole exception, so I can see the message and stack trace of the unexpected exception, and maybe solve this problem without debugging .

```cs
// file Test1.cs
namespace mstest306;

[TestClass]
public sealed class Test1
{
    [TestMethod]
    public void TestMethod1()
    {
        Assert.Throws<ArgumentException>(() => new MyClass().MyMethod());
    }
}

public sealed class MyClass
{
    public void MyMethod ()
    {
        Helper.HelperMethod();
    }
}

public static class Helper
{
    public static void HelperMethod()
    {

        if (DateTime.Now.DayOfWeek.ToString() == "Friday")
        {
            throw new InvalidOperationException("the error");
        }   
        else
        {
            throw new InvalidOperationException("the other error");
        } 
    }
}

```

nohwnd:
Partially fix #6906 

The data are there, but the formatting is not great.


src/TestFramework/TestFramework/Assertions/Assert.ThrowsException.cs
```diff
@@ -577,7 +578,8 @@ private static ThrowsExceptionState IsThrowsFailing<TException>(Action action, b
                             FrameworkMessages.WrongExceptionThrown,
                             userMessage,
                             typeof(TException),
-                            ex.GetType());
+                            ex.GetType(),
+                            ex.ToString());
```
Youssef1313, line 582: `ex.ToString()` already contains the type right? Should we just have `ex.ToString()` to avoid unnecessarily duplicate info in assertion failure message?

nohwnd: I considered it but the whole message needs more love (I am summarizing that in the linked issue). So let's consider this just the first step there to help investigating the issue you were seeing.

Youssef1313: The issue I was seeing is already fixed, it was just the message update related to caller arg expression breaking the test.

PR #766, opened 2021-02-02
ManagedNames implementation refactored.

Haplois:
- ManagedNames support in TestCase is updated to support the latest preview.
 - Updated TP to `16.10.0-preview-20210204-01`.

PR #7034, opened 2025-12-02
Cleanup unneeded workaround

### LINKED ISSUE #7033: Revise workaround for managed type
https://github.com/microsoft/testfx/pull/1876 introduces a workaround that I would like to validate if it's still needed or not. If not still needed, let's revert it.

Youssef1313:
Fixes #7033

I think the cause of issues there for F# is the fact that we incorrectly plug ManagedType in TestClassName.


src/Adapter/MSTestAdapter.PlatformServices/ObjectModel/UnitTestElement.cs
```diff
@@ -112,13 +112,10 @@ internal TestCase ToTestCase()
         {
             testCase.SetPropertyValue(TestCaseExtensions.ManagedTypeProperty, TestMethod.ManagedTypeName);
             testCase.SetPropertyValue(TestCaseExtensions.ManagedMethodProperty, TestMethod.ManagedMethodName);
-            testCase.SetPropertyValue(EngineConstants.TestClassNameProperty, TestMethod.ManagedTypeName);
```
Youssef1313, line 115: This line looks completely wrong.

It simply means we don't round trip correctly from UnitTestElement -> TestCase -> UnitTestElement back. This is the likely cause of issues for F#. This was introduced way back in #766

PR #7054, opened 2025-12-05
Don't trim trailing whitespace for resx and xlf

Evangelink:

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -15,6 +15,7 @@ vsspell_ignored_words_842f80e2d4aa4288afdcd0e42833eeaf = runsettings|nullable|ar
 
 [*.{xlf,resx}]
 insert_final_newline = unset
+trim_trailing_whitespace = unset
 
 # Xml localization files
 [*.xlf]


.editorconfig
```diff
@@ -15,6 +15,7 @@ vsspell_ignored_words_842f80e2d4aa4288afdcd0e42833eeaf = runsettings|nullable|ar
 
 [*.{xlf,resx}]
 insert_final_newline = unset
+trim_trailing_whitespace = unset
```
Youssef1313, line 18: What's the difference between false and unset in this case? Do you know?

Evangelink: I am not sure if it's better to use `false`. From what I saw:

false: Actively disables the setting. The editor will enforce that the rule is turned off (e.g., it will actively not insert a final newline or not trim trailing whitespace).

unset: Removes the setting entirely, reverting to the default behavior. This allows the setting to be inherited from a parent section or the editor's own defaults. It doesn't enforce anything—it just stops overriding whatever was set before.

PR #7065, opened 2025-12-06
use contains from polyfill

SimonCropp:


Evangelink:
Thank you @SimonCropp

PR #7067, opened 2025-12-06
Assembly.Load cant be null

SimonCropp:
AFAIK `Assembly.Load` and `Assembly.ReflectionOnlyLoad` cant return null?


Evangelink (approved):
Seems to be matching the documentation and compiler nullability check.

Thanks @SimonCropp!

PR #7081, opened 2025-12-08
Report target framework rather than runtime framework

### LINKED ISSUE #7066: Nitpick: `dotnet test` with MTP shows incorrect target framework version
## Describe the bug

Image

You can see here that `dotnet test` is showing this as `net48`, when it's clearly `net472`. I don't know if this a difference between "the target framework you built with" vs. "the framework we're running with" (because you could the same thing for `x64` vs. `AnyCPU`).

This is a nitpick for me. Not sure how important it is for other users.

nohwnd:
Fix #7066

### DIFF
--- a/src/Platform/Microsoft.Testing.Platform/ServerMode/DotnetTest/DotnetTestConnection.cs
+++ b/src/Platform/Microsoft.Testing.Platform/ServerMode/DotnetTest/DotnetTestConnection.cs
@@ -95,7 +95,10 @@ internal sealed class DotnetTestConnection : IPushOnlyProtocol, IDisposable
         {
             { HandshakeMessagePropertyNames.PID, _environment.ProcessId.ToString(CultureInfo.InvariantCulture) },
             { HandshakeMessagePropertyNames.Architecture, RuntimeInformation.ProcessArchitecture.ToString() },
-            { HandshakeMessagePropertyNames.Framework, RuntimeInformation.FrameworkDescription },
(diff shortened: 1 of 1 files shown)


Youssef1313 (changes requested):
I'm concerned about this change. The actual TFM is usually present in the path of the assembly. Reporting the actual running runtime here is an additional piece of info we don't want to lose.

This also just caught a bug in .NET 11 insertion today. https://github.com/microsoft/testfx/pull/7154#issuecomment-3742711091

That bug would have been masked if we report the target framework.

Youssef1313:
@bradwilson

Interesting, for 4.8.1 we should display the full thing:

https://github.com/dotnet/sdk/blob/de50305357a7020997e543e036dfe29ded1832c7/src/Cli/dotnet/Commands/Test/MTP/Terminal/TargetFrameworkParser.cs#L43-L46

Can you double check what `RuntimeInformation.FrameworkDescription` has for your case?

For what happened in .NET 11, it's because .NET 11 SDK used was in an intermediate state and wasn't fully correct. Putting aside that bug, it's still possible that you can run with a newer runtime if you specify `DOTNET_ROLL_FORWARD`, or if you alter runtimeconfig.json manually, or if you explicitly run with `dotnet exec --runtimeconfig path/to/custom/runtimeconfigjson`, or `dotnet exec --roll-forward`, or `dotnet exec --fx-version`.

PR #7098, opened 2025-12-10
Flow assert expected and actual to be printed on CLI or TE

### LINKED ISSUE #6717: Assert.AreEqual string: consider displaying the full actual and expected in addition to the diff
## Summary

Let's take the following error diff:

```
Assert.AreEqual failed. Expected string length 199 but was 139. 'expected' expression: 'Normalize(expected)', 'actual' expression: 'Normalize(actual)'.
Expected: "...n 1; }␊    #endregion␊␊    #region ..."
But was:  "...n 1; }␊    #endregion␊    #endregion␊}"
```

This is nice for spotting the exact difference but it's sometimes interesting to have access to the full context.

For examples:

1. When you are in a scenario where you are changing the output, it's useful to be able to copy the new expected content to replace current expected.

2. When there are more than one issue, you would need to run the assert multiple times when having the full actual/expected might allow you to do an extra check upfront.

Evangelink:
Fix #6717

### DIFF
--- a/src/TestFramework/TestFramework/Assertions/Assert.cs
+++ b/src/TestFramework/TestFramework/Assertions/Assert.cs
@@ -224,15 +224,11 @@ public sealed partial class Assert
     /// <param name="parameterName">
     /// parameter name.
     /// </param>
-    /// <param name="message">
-    /// message for the invalid parameter exception.
-    /// </param>
-    internal static void CheckParameterNotNull([NotNull] object? param, string assertionName, string parameterName, string? message)
+    internal static void CheckParameterNotNull([NotNull] object? param, string assertionName, string parameterName)
     {
         if (param == null)
         {
-            string userMessage = BuildUserMessage(message);
-            string finalMessage = string.Format(CultureInfo.CurrentCulture, FrameworkMessages.NullParameterToAssert, parameterName, userMessage);
+            string finalMessage = string.Format(CultureInfo.CurrentCulture, FrameworkMessages.NullParameterToAssert, parameterName);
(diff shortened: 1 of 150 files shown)


src/TestFramework/TestFramework/Assertions/Assert.cs
```diff
@@ -40,6 +40,70 @@ internal static void ThrowAssertFailed(string assertionName, string? message)
...
+
+        if (HasKnownGoodToString(actual))
+        {
+            exception.Data["assert.actual"] = actual;
+        }
+
+        throw exception;
+    }
+
+    private static bool HasKnownGoodToString([NotNullWhen(true)] object? value)
```
Evangelink, line 82: We could improve the logic to try to detect if the given type overrides `ToString` in which case we call it. We could also allow users to register a method for the display representation of types. 

But all of that can happen later, we are already improving experience as-is.

samples/Playground/Tests.cs
```diff
@@ -15,5 +15,9 @@ public class TestClass
     [TestMethod]
     public void Test1()
     {
+        string commonPart = new('a', 1000);
+        string expected = commonPart + "expected";
+        string actual = commonPart + "actual";
+        Assert.AreEqual(expected, actual, ignoreCase: true);
```
Evangelink:
```suggestion
        // string commonPart = new('a', 1000);
        // string expected = commonPart + "expected";
        // string actual = commonPart + "actual";
        // Assert.AreEqual(expected, actual, ignoreCase: true);
```

PR #7125, opened 2025-12-15
Move to modern markdownlint

Evangelink:

### DIFF
--- a/.github/copilot-instructions.md
+++ b/.github/copilot-instructions.md
@@ -1,4 +1,4 @@
-# Github Copilot Instructions
+# GitHub Copilot Instructions
 
 This is a .NET based repository that contains the MSTest testing framework and Microsoft.Testing.Platform (aka MTP) testing platform. Please follow these guidelines when contributing:
 


nohwnd:
Which repos are those?

nohwnd (approved):
approved, added to exclusions.

.github/workflows/markdownlint-problem-matcher.json
Youssef1313, line 1: Are we now losing the GitHub annotations when there are failures?

Evangelink: I am not aware of it, have you seen some problem?

Youssef1313: Not yet. What the problem matcher used to do is that it will show where the errors are in GitHub Files tab. So not sure how this would work now. It's not a big deal anyways.

Evangelink: We may need to use this formatter https://github.com/DavidAnson/markdownlint-cli2/blob/main/formatter-template/README.md

Issue #2196, opened 2024-01-29
Enable test execution in browser-wasm


idg10:
## Summary

I want to be able to run tests inside a browser-hosted WASM environment. That environment behaves significantly differently from other .NET environments, so it's vital to test thoroughly, but we can't today because there seems to be no support for it.

We were hopeful when the [new test runner](https://devblogs.microsoft.com/dotnet/introducing-ms-test-runner/) was announced that it would address this scenario, since it deals with some other environments that used not to support running tests. But apparently not.

## Background and Motivation

I maintain [Rx.NET (`System.Reactive`)](https://github.com/dotnet/reactive) and we recently had a report of a [problem that occurs in Blazor WASM](https://github.com/dotnet/reactive/issues/2061).

When fixing problems I generally want to add a failing test first to verify that the fix addresses the problem, and to prevent regressions. But I can't do that here because there appears to be no way to run MSTest tests in browser-hosted WASM based .NET code. The problem described occurs only when running in the `browser-wasm` runtime, so for an automated test to reproduce the scenario that triggers this bug, we'd need that test to be able to execute in that runtime.

## Proposed Feature

In an ideal world, we would just be able to add `browser-wasm` (and, since we're talking about WASM also `wasi-wasm`) to the `<RuntimeIdentifiers>` in a test project's `csproj` and for it to automatically run the tests in each listed runtime. But failing that, we'd like some way for `browser-wasm` apps to use the mechanisms that make the new test runner possible.

As far as we can see, at least two things would need to be addressed:

* we'd need some mechanism to capture the output because there's no meaningful standard output in `browser-wasm`
* the mechanism by which configuration is supplied would need to be modified

That second point is the first thing we run into when trying to do this today. Since letting the MSTest tooling generate the app entry point won't work for a `browser-wasm` app, we tried putting code similar to what it generates into our app:

```cs
string[] args = [];
global::Microsoft.Testing.Platform.Builder.ITestApplicationBuilder builder = await 
    global::Microsoft.Testing.Platform.Builder.TestApplication.CreateBuilderAsync(args);
Microsoft.Testing.Platform.MSBuild.TestingPlatformBuilderHook.AddExtensions(builder, args);
Microsoft.Testing.Extensions.Telemetry.TestingPlatformBuilderHook.AddExtensions(builder, args);
Task<ITestApplication> bt = builder.BuildAsync(); // Separating from await because single stepping async in blazor is weird
[...]

PR #7146, opened 2025-12-17
Fix running in browser wasm

Evangelink:
Contributes to #2196.

I have manually validated things are working with the fix:

image

I'll try to work on adding some acceptance test in a follow-up PR.


src/Platform/Microsoft.Testing.Platform/Services/CurrentTestApplicationModuleInfo.cs
```diff
@@ -93,6 +94,17 @@ public string GetProcessPath()
         string? processPath = currentProcess.MainModule?.FileName;
 #endif
 
+        if (processPath is null)
+        {
+            // Fallback for environments where ProcessPath is null (e.g., browser OS)
+            string[] commandLineArgs = environment.GetCommandLineArgs();
```
Youssef1313, line 100: Doesn't GetCommandLineArgs return empty array on browser?

Issue #7162, opened 2026-01-01
`NopFilter` is still marked with `TPEXP`


bradwilson:
## Describe the bug

The `NopFilter` class, which has been in MTP since 1.0.2 and is presumably stable, is still marked with `TPEXP`. It should not be.

## Steps To Reproduce

* Use `NopFilter`
* Get `TPEXP` compiler warning

## Expected behavior

Should be usable without `#pragma warning disable TPEXP`.

## Actual behavior

Can't. 😄

PR #7169, opened 2026-01-08
Remove experimental attr from NopFilter

mariam-abdulla:
This pull request makes a minor change to the `NopFilter` class by removing the `[Experimental("TPEXP", ...)]` attribute from its declaration. The class remains part of the public API.

- Removed the `[Experimental("TPEXP", ...)]` attribute from the `NopFilter` class in `NopFilter.cs`, making it no longer marked as experimental.
- Updated the `PublicAPI.Shipped.txt` to reflect that `NopFilter` is no longer tagged as experimental.

Fixes https://github.com/microsoft/testfx/issues/7162

### DIFF
--- a/src/Platform/Microsoft.Testing.Platform/PublicAPI/PublicAPI.Shipped.txt
+++ b/src/Platform/Microsoft.Testing.Platform/PublicAPI/PublicAPI.Shipped.txt
@@ -45,8 +45,8 @@
 [TPEXP]Microsoft.Testing.Platform.Logging.ILoggingManager.AddProvider(System.Func<Microsoft.Testing.Platform.Logging.LogLevel, System.IServiceProvider!, Microsoft.Testing.Platform.Logging.ILoggerProvider!>! loggerProviderFactory) -> void
 [TPEXP]Microsoft.Testing.Platform.Requests.IExecuteRequestCompletionNotifier
 [TPEXP]Microsoft.Testing.Platform.Requests.IExecuteRequestCompletionNotifier.Complete() -> void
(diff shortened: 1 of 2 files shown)


src/Platform/Microsoft.Testing.Platform/PublicAPI/PublicAPI.Shipped.txt
```diff
@@ -45,8 +45,8 @@
 [TPEXP]Microsoft.Testing.Platform.Logging.ILoggingManager.AddProvider(System.Func<Microsoft.Testing.Platform.Logging.LogLevel, System.IServiceProvider!, Microsoft.Testing.Platform.Logging.ILoggerProvider!>! loggerProviderFactory) -> void
 [TPEXP]Microsoft.Testing.Platform.Requests.IExecuteRequestCompletionNotifier
 [TPEXP]Microsoft.Testing.Platform.Requests.IExecuteRequestCompletionNotifier.Complete() -> void
-[TPEXP]Microsoft.Testing.Platform.Requests.NopFilter
-[TPEXP]Microsoft.Testing.Platform.Requests.NopFilter.NopFilter() -> void
+Microsoft.Testing.Platform.Requests.NopFilter
+Microsoft.Testing.Platform.Requests.NopFilter.NopFilter() -> void
```
Youssef1313: Usually we don't modify `.Shipped.txt` until we are really shipping. So for this case, we would copy these two lines (with TPEXP) to Unshipped with a `*REMOVED*` prefix, and then add them again to unshipped without the `*REMOVED*` prefix.

PR #7170, opened 2026-01-08
Fix `That_StaticPropertyAccess_FailsAsExpected` to not fail every year

Youssef1313:


Evangelink (approved):

test/UnitTests/TestFramework.UnitTests/Assertions/AssertTests.That.cs
```diff
@@ -339,14 +339,14 @@ public void That_NestedMethodCalls_FailsAsExpected()
     public void That_StaticPropertyAccess_FailsAsExpected()
     {
         Action act = () => Assert.That(() => DateTime.Now.Year < 2000, "Current year should be before 2000");
-
+        var year = DateTime.Now.Year;
```
Evangelink:
```suggestion
        int year = DateTime.Now.Year;
```

PR #7173, opened 2026-01-08
Fix AssemblyCleanup running when not all class cleanups might have finished

### LINKED ISSUE #7120: AssemblyCleanup runs before all ClassCleanups finish
## Describe the bug

`AssemblyCleanup` may begin executing while some `ClassCleanup` methods are still running.
This only happens when tests are run in parallel (`ExecutionScope.MethodLevel`).

## Steps To Reproduce

Running the following program:

```csharp
[assembly: Parallelize(Scope = ExecutionScope.MethodLevel)]

namespace TestProjectParallel;

[TestClass]
public static class GlobalSettings
{
    public static int MissingClassCleanups;

    [AssemblyCleanup]
    public static async Task AssemblyCleanup()
    {
        int foundMissingClassCleanups = MissingClassCleanups;
        await Task.Delay(1000);
        if (foundMissingClassCleanups > 0)
        {
            int foundMissingClassCleanupsAfter1Second = MissingClassCleanups;
            throw new InvalidOperationException($"AssemblyCleanup started while ClassCleanup is not finished (MissingClassCleanups: {foundMissingClassCleanups} after 1 sec: {foundMissingClassCleanupsAfter1Second})");
        }
    }
}

[TestClass]
public sealed class Test1
{
    [ClassCleanup]
    public static async Task ClassCleanup()
    {
        await Task.Delay(Random.Shared.Next(1, 100));
        Interlocked.Decrement(ref GlobalSettings.MissingClassCleanups);
    }

    [TestMethod]
    public void TestMethod1()
    {
        Interlocked.Increment(ref GlobalSettings.MissingClassCleanups);
    }
}

[TestClass]
public sealed class Test2
{
    [ClassCleanup]
    public static async Task ClassCleanup()
    {
        await Task.Delay(Random.Shared.Next(1, 100));
        Interlocked.Decrement(ref GlobalSettings.MissingClassCleanups);
    }

    [TestMethod]
    public void TestMethod1()
    {
        Interlocked.Increment(ref GlobalSettings.MissingClassCleanups);
    }
}
```

Given the following csproj:
```xml

  <PropertyGroup>
    <LangVersion>latest</LangVersion>
    <Nullable>enable</Nullable>
    <OutputType>Exe</OutputType>
  </PropertyGroup>

  <ItemGroup>
    <PackageReference Include="MSTest" Version="4.0.2" />
  </ItemGroup>

  <ItemGroup>
    <Using Include="Microsoft.VisualStudio.TestTools.UnitTesting" />
  </ItemGroup>

</Project>
```

Often (try multiple runs) it executes `GlobalSettings.AssemblyCleanup()` while `Test1.ClassCleanup()` or `Test2.ClassCleanup()` is still running.
`System.InvalidOperationException: AssemblyCleanup started while ClassCleanup is not finished (MissingClassCleanups: 1 after 1 sec: 0).`

## Expected behavior

`AssemblyCleanup` runs after all `ClassCleanup` are finished, e.g.

Test2.TestMethod()
Test1.TestMethod()
Test2.ClassCleanup() - start
Test2.ClassCleanup() - finish
Test1.ClassCleanup() - start
Test1.ClassCleanup() - finish
GlobalSettings.AssemblyCleanup() - start
GlobalSettings.AssemblyCleanup() - finish

Note: The order of the `TestMethod`/`ClassCleanup` between `Test1` and `Test2` is not relevant.

## Actual behavior

Test2.TestMethod()
Test1.TestMethod()
Test2.ClassCleanup() - start
Test2.ClassCleanup() - finish
Test1.ClassCleanup() - start
GlobalSettings.AssemblyCleanup() - start
Test1.ClassCleanup() - finish
GlobalSettings.AssemblyCleanup() - finish

## Additional context

This behavior affects users of Reqnroll: https://github.com/reqnroll/reqnroll/issues/951.

Youssef1313:
Fixes #7120

After every test method completes, we marked it as completed.
Then after every class cleanup, we called `RunAssemblyCleanupIfNeededAsync`, which will check if all tests have been executed, and runs assembly cleanup. This is wrong because all tests could have been executed, but there are remaining class cleanups running in parallel.

This PR updates the logic so we have the concept of "marking test as completed" and "marking test class as completed". Previously we only had the former concept.

### DIFF
--- a/src/Adapter/MSTestAdapter.PlatformServices/Execution/UnitTestRunner.cs
+++ b/src/Adapter/MSTestAdapter.PlatformServices/Execution/UnitTestRunner.cs
@@ -210,9 +210,9 @@ internal sealed class UnitTestRunner : MarshalByRefObject
             {
                 await testMethodInfo.Parent.RunClassCleanupAsync(testContextForClassCleanup, result).ConfigureAwait(false);
 
-                // We need to mark the class as complete to allow assembly cleanup to run.
-                // Note that we don't mark the class as complete when we are marking the last method.
-                // We need to only mark the class as complete after we are sure we have executed the class cleanup, to prevent concurrent runs of class cleanup and assembly cleanup.
+                // Mark the class as complete when all class cleanups are complete. When all classes are complete we progress to running assembly cleanup.
+                // Class is not complete until after all class cleanups are done, to prevent running assembly cleanup too early.
+                // Do not mark the class as complete when the last test method in the class completed. That is too early, we need to run class cleanups before marking class as complete.
                 _classCleanupManager.MarkClassComplete(testMethod.FullClassName);
             }
 


src/Adapter/MSTestAdapter.PlatformServices/Execution/UnitTestRunner.cs
```diff
@@ -204,15 +204,23 @@ internal async Task<TestResult[]> RunSingleTestAsync(TestMethod testMethod, IDic
...
-            if (testMethodInfo is not null)
+
+            _classCleanupManager.MarkTestComplete(testMethod, out bool isLastTestInClass);
+            if (isLastTestInClass && testMethodInfo is not null)
             {
-                await testMethodInfo.Parent.RunClassCleanupAsync(testContextForClassCleanup, _classCleanupManager, testMethodInfo, result).ConfigureAwait(false);
+                await testMethodInfo.Parent.RunClassCleanupAsync(testContextForClassCleanup, result).ConfigureAwait(false);
+
+                // We need to mark the class as complete to allow assembly cleanup to run.
+                // Note that we don't mark the class as complete when we are marking the last method.
```
nohwnd: What does this mean? That the class is not complete when the last test method has run, but that we mark it as complete only when ClassCleanup of this class is completed?

test/IntegrationTests/MSTest.Acceptance.IntegrationTests/AssemblyCleanupTests.cs
```diff
@@ -0,0 +1,107 @@
...
+public class TestClass2
+{
+    [TestMethod]
+    public void Test2()
+    {
+    }
+
+    [ClassCleanup]
+    public static void Cleanup2()
+        => Thread.Sleep(2000);
```
nohwnd, line 106: Why is this sleep needed and how it helps to prove that the test class cleanup has completed non-concurrently with assembly cleanup?

Seems that it would have been much easier to output "Starting cleanup 1...", "Sleeping in cleanup 1...", "Completed cleanup 1..." from the class and assembly cleanups and checking that output in the test. That would make it easier to see what is going on in the test if it stop working correctly.

src/Adapter/MSTestAdapter.PlatformServices/Execution/UnitTestRunner.cs
Youssef1313: Yes. Basically, we don't want to mark the class as complete in the state between "last test completed" and "class cleanup" complete.

test/IntegrationTests/MSTest.Acceptance.IntegrationTests/AssemblyCleanupTests.cs
Youssef1313: The sleeps are done in a way that always produces the race, and always fails on main before the change. This is asserted in the assembly cleanup, where we assert that the longer class cleanup has already been completed.

On main, what will happen is, both class cleanups start roughly at the same time. Then assembly cleanup will start right after the shorter class cleanup finishes, which means we executed assembly cleanup while another class cleanup is in progress.

nohwnd: Seeing how many different sleeps were changed in the past, because they were now "flaky" tests I just don't have much trust in writing a test case in a sleeps without adding any additional info. If it fails it gives very little info about what it was doing, apart from "expected true and is now false". I would much rather see some additional output that makes it easier to compare successful run with unsuccessful run.

If I then see:

"Starting cleanup 1...", 
"Sleeping in cleanup 1...",
"Starting assembly cleanup..."
 "Completed cleanup 1..."
 
 I can tell much easier why the assertion is false, and won't go hunting for it with debugger.

Youssef1313: Updated the test

nohwnd: What about this sleep? can that also get an output? or it is not relevant anymore?

Youssef1313: The two cleanups are executed concurrently so an output for both is expected to not be stable. The important thing is that assembly cleanup always remains after both, which is asserted by asserting the output of the assembly cleanup and the output of the longer class cleanup.

src/Adapter/MSTestAdapter.PlatformServices/Execution/UnitTestRunner.cs
```diff
@@ -204,15 +204,23 @@ internal async Task<TestResult[]> RunSingleTestAsync(TestMethod testMethod, IDic
...
+                // We need to only mark the class as complete after we are sure we have executed the class cleanup, to prevent concurrent runs of class cleanup and assembly cleanup.
```
nohwnd:
```suggestion
                // Mark the class as complete when all class cleanups are complete. When all classes are complete we progress to running assembly cleanup.
                // Class is not complete until after all class cleanups are done, to prevent running assembly cleanup too early.
                // Do not mark the class as complete when the last test method in the class completed. That is too early, we need to run class cleanups before marking class as complete.
```

src/Adapter/MSTestAdapter.PlatformServices/Execution/UnitTestRunner.cs
```diff
@@ -204,15 +204,23 @@ internal async Task<TestResult[]> RunSingleTestAsync(TestMethod testMethod, IDic
...
+                _classCleanupManager.MarkClassComplete(testMethod.FullClassName);
             }
 
-            if (testMethodInfo?.Parent.Parent.IsAssemblyInitializeExecuted == true)
+            if (testMethodInfo?.Parent.Parent.IsAssemblyInitializeExecuted == true &&
```
nohwnd, line 219: what is the logic there? seems weird to tie something to assembly init running or not running. Or we are just keeping the previous behavior no matter if it is broken?

Youssef1313: I think this is done so that if you have all tests skipped, you don't run assembly cleanup (and we also won't run assembly init)

test/IntegrationTests/MSTest.Acceptance.IntegrationTests/AssemblyCleanupTests.cs
nohwnd: Yeah, but having no message from this cleanup makes it invisible, while it still impacts the result, So you basically want to assert "Asm cleanup is after all class cleanups", but seem like you are reluctant to do that because you don't have the right assertion for it?

Youssef1313: Yes. It will complicate the test unnecessarily. The test on the current form guards correctly against this bug.

nohwnd: Feels to me like you have to whip out debugger to see why the test fails when this gets again failing.  But your call to make :)

nohwnd (approved):

Issue #7179, opened 2026-01-08
Investigate if OperatingSystem.IsBrowser polyfill that we use for netstandard2.0 is correct


Youssef1313:
We get IsBrowser via Polyfill and is implemented as:

```
        /// <summary>
        /// Indicates whether the current application is running as WASM in a browser.
        /// </summary>
        //Link: https://learn.microsoft.com/en-us/dotnet/api/system.operatingsystem.isbrowser?view=net-10.0
        public static bool IsBrowser() =>
            RuntimeInformation.FrameworkDescription.Contains(".NET WebAssembly");

```

That implementation doesn't sound correct, looking at https://github.com/dotnet/runtime/blob/7a2016c11cfff514e5df4e56931785b2eb0cb832/src/libraries/System.Private.CoreLib/gen/ProductVersionInfoGenerator.cs

PR #7180, opened 2026-01-08
Cleanup TrxReport implementation

Youssef1313:
Follow-up to #7115

### DIFF
--- a/src/Platform/Microsoft.Testing.Extensions.TrxReport/TrxModeHelpers.cs
+++ b/src/Platform/Microsoft.Testing.Extensions.TrxReport/TrxModeHelpers.cs
@@ -1,0 +1,14 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using Microsoft.Testing.Platform.CommandLine;
+
+namespace Microsoft.Testing.Extensions.TrxReport;
+
+internal static class TrxModeHelpers
+{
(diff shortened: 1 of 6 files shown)


src/Platform/Microsoft.Testing.Extensions.TrxReport/TrxModeHelpers.cs
```diff
@@ -0,0 +1,14 @@
...
+using Microsoft.Testing.Platform.CommandLine;
+
+namespace Microsoft.Testing.Extensions.TrxReport;
+
+internal static class TrxModeHelpers
+{
+    [UnsupportedOSPlatformGuard("BROWSER")]
+    public static bool ShouldUseOutOfProcessTrxGeneration(ICommandLineOptions commandLineOptions)
+        => commandLineOptions.IsOptionSet(CrashDumpCommandLineOptions.CrashDumpOptionName) &&
+            !RuntimeInformation.IsOSPlatform(OSPlatform.Create("BROWSER"));
```
Evangelink, line 13: `OperatingSystem.IsBrowser` is easier to read IMO.

Youssef1313: It is. But we get it via polyfill which is likely buggy today. https://github.com/microsoft/testfx/issues/7179

Evangelink: Ohhh right! I have used it in quite a few places.

PR #7189, opened 2026-01-08
Remove ConsiderFixturesAsSpecialTests

### LINKED ISSUE #3376: Fixture start/end time is not correct when using `ConsiderFixturesAsSpecialTests`
## Summary

When enabling the `ConsiderFixturesAsSpecialTests` feature, the fixtures are reported as special tests but the associated start/end time is not correct.

### LINKED ISSUE #3377: Fixtures are not reported following logical order when using ConsiderFixturesAsSpecialTests
## Summary

When enabling the `ConsiderFixturesAsSpecialTests` feature, the fixture methods should be reported in the same logical order as they are returned.

Today, they are all reported last, after all tests.

### LINKED ISSUE #5716: `ConsiderFixturesAsSpecialTests` is broken
In Playground, I have:

```csharp
// Copyright (c) Microsoft Corporation. All rights reserved.

#pragma warning disable VSTHRD200 // Use "Async" suffix for async methods

[assembly: Parallelize(Scope = ExecutionScope.MethodLevel, Workers = 0)]

namespace Playground;

[TestClass]
public class TestClass
{
    [TestMethod]
    [DynamicData(nameof(Data))]
    public void Test3(int a, int b)
    {
    }

    [TestInitialize]
    public void TestInitialize()
    {
        // Initialization code here
    }

    [TestCleanup]
    public void TestCleanup()
    {
        // Cleanup code here
    }

    [ClassInitialize]
    public static void ClassInitialize(TestContext _)
    {
        // Class-level initialization code here
    }

    [ClassCleanup(ClassCleanupBehavior.EndOfClass)]
    {
        // Class-level cleanup code here
    }

    [AssemblyInitialize]
    public static void AssemblyInitialize(TestContext _)
    {
        // Assembly-level initialization code here
    }

    [AssemblyCleanup]
    public static void AssemblyCleanup()
    {
        // Assembly-level cleanup code here
    }

    public static IEnumerable<(int A, int B)> Data
    {
        get
        {
            yield return (1, 2);
            yield return (3, 4);
        }
    }
}

[TestClass]
public class TestClass2
{
    [TestMethod]
    [DynamicData(nameof(Data))]
    public void Test3(int a, int b)
    {
    }

    [TestInitialize]
    public void TestInitialize()
    {
        // Initialization code here
    }

    [TestCleanup]
    public void TestCleanup()
    {
        // Cleanup code here
    }

    [ClassInitialize]
    public static void ClassInitialize(TestContext _)
    {
    }

    [ClassCleanup(ClassCleanupBehavior.EndOfClass)]
    {
        throw new Exception("TestClass2.ClassCleanup failed");
    }

    public static IEnumerable<(int A, int B)> Data
    {
        get
        {
            yield return (1, 2);
            yield return (3, 4);
        }
    }
}
```

If I enable `ConsiderFixturesAsSpecialTests` and try to run the tests in `Playground`, I get this:

Image

The class cleanup isn't associated to the right node.

Youssef1313:
`ConsiderFixturesAsSpecialTests` is an opt-in feature that causes test fixtures (e.g, class/assembly init/cleanup) to be discovered as fake test cases. The code related to this feature is quite big, the feature has been buggy for long, and people are unlikely to be using it. I think it's better to remove it. In future, we might want to implement it differently, in collaboration with Test Explorer so that we have a way to Test Explorer to say this test result is related to a fixture and it gets displayed in a more proper way and have a proper experience in IDE. The impact of removing the feature is low as it only affects the UX.

Closes #3376
Closes #3377 
Closes #5716

### DIFF
--- a/test/UnitTests/MSTestAdapter.PlatformServices.UnitTests/MSTestSettingsTests.cs
+++ b/test/UnitTests/MSTestAdapter.PlatformServices.UnitTests/MSTestSettingsTests.cs
@@ -120,7 +120,7 @@ public class MSTestSettingsTests : TestContainer
 
         var adapterSettings = MSTestSettings.GetSettings(runSettingsXml, MSTestSettings.SettingsNameAlias, _mockMessageLogger.Object);
 
-        _mockMessageLogger.Verify(lm => lm.SendMessage(TestMessageLevel.Warning, It.IsAny<string>()), Times.Exactly(16));
+        _mockMessageLogger.Verify(lm => lm.SendMessage(TestMessageLevel.Warning, It.IsAny<string>()), Times.Exactly(15));
         _mockMessageLogger.Verify(lm => lm.SendMessage(TestMessageLevel.Warning, "Invalid value '3' for runsettings entry 'CooperativeCancellationTimeout', setting will be ignored."), Times.Once);
         _mockMessageLogger.Verify(lm => lm.SendMessage(TestMessageLevel.Warning, "Invalid value 'timeout' for runsettings entry 'TestCleanupTimeout', setting will be ignored."), Times.Once);
         _mockMessageLogger.Verify(lm => lm.SendMessage(TestMessageLevel.Warning, "Invalid value 'timeout' for runsettings entry 'ClassCleanupTimeout', setting will be ignored."), Times.Once);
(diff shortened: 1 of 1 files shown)


Evangelink:
This is used and requested internally by Substrate, please validate with them before dropping it.

PR #7203, opened 2026-01-12
Automate publishing to test-tools feed for RTM, but require explicit manual validation

Youssef1313:


azure-pipelines-official.yml
```diff
@@ -96,7 +96,8 @@ extends:
           enableTelemetry: true
           jobs:
           - job: Windows
-            timeoutInMinutes: 90
+            # Allow build, test, publishing, etc to take 90 minutes, plus allow for 2 hours of manual validation before publishing to test-tools feed.
```
nohwnd: The comment (to me) reads like we will automatically publish to test-tools if we don't do anything validation, but in reality it will reject the validation and the build.

azure-pipelines-official.yml
```diff
@@ -184,20 +185,28 @@ extends:
                   ArtifactName: TestResults_Windows_Attempt$(System.JobAttempt)
                 condition: always()
 
+            # Manual validation is only needed for RTM.
+            # This allows us to keep re-building the same RTM version if the manual validation fails, because in that case we wouldn't have published it to test-tools already.
+            # We will only publish if manual validation is successful.
```
nohwnd:
```suggestion
            # This allows us to keep re-building the same RTM version until it passes manual validation, then we publish it to test-tools, where we have only one chance at publishing the final (RTM) package.
```

azure-pipelines-official.yml
Youssef1313: Updated the comment. Is it clearer now?

PR #7219, opened 2026-01-14
Report discovered telemetry events correctly for MTP

Youssef1313:

### DIFF
--- a/src/Platform/Microsoft.Testing.Platform/Telemetry/TelemetryProperties.cs
+++ b/src/Platform/Microsoft.Testing.Platform/Telemetry/TelemetryProperties.cs
@@ -10,7 +10,7 @@ internal static class TelemetryProperties
     public const string ReporterIdPropertyName = "reporter id";
     public const string IsCIPropertyName = "is ci";
 
-    public const string VersionValue = "19";
+    public const string VersionValue = "20";
 
     public const string True = "true";
     public const string False = "false";


src/Platform/Microsoft.Testing.Platform/Telemetry/TelemetryProperties.cs
```diff
@@ -10,7 +10,7 @@ internal static class TelemetryProperties
     public const string ReporterIdPropertyName = "reporter id";
     public const string IsCIPropertyName = "is ci";
 
-    public const string VersionValue = "19";
+    public const string VersionValue = "20";
```
Youssef1313, line 13: @nohwnd Is this something we should do?

Evangelink: Bumping the version is usually helpful when you are making a change to the collected telemetry so that you can do mappings of info based on version.

I would say that in this case, this shouldn't be needed.

nohwnd: This does make change in the reported properties, because they will now be reported on a different event, so to get correct data you need to be aware of the break and correctly grab properties from both events, otherwise you will just get new data (or old data), or will inspect new data unnecessarily. So bumping the version here is very useful, and follows the intent of the version - to make big breaking changes to the telemetry easier to query.

PR #7223, opened 2026-01-14
Fix TypeCache to throw on duplicate attributes when necessary

### LINKED ISSUE #4333: `GetFirstDerivedAttributeOrDefault` implementation isn't consistent with doc comment
Doc says:

> /// <exception cref="InvalidOperationException">Throws when multiple attributes are found (the attribute must allow multiple).</exception>

Implementation:

```csharp
        Attribute[] cachedAttributes = GetCustomAttributesCached(attributeProvider, inherit);

        foreach (Attribute cachedAttribute in cachedAttributes)
        {
            if (AttributeComparer.IsDerived<TAttribute>(cachedAttribute))
            {
                return (TAttribute)cachedAttribute;
            }
        }

        return null;
```

My suggestion:

1. Remove the doc comment about throwing an exception and leave the method as-is
2. Introduce a new method that throws an exception and name it `GetSingleDerivedAttributeOrDefault`. Note that throwing exception may still be relevant even if AllowMultiple is not true (ONLY IF the attribute is not sealed). See https://github.com/microsoft/testfx/issues/4331 for details.
3. Revise all usages, and switch from `GetFirstDerivedAttributeOrDefault` to `GetSingleDerivedAttributeOrDefault` if needed.
4. My expectation is that after we do these three steps, we will end up having `GetFirstDerivedAttributeOrDefault` called with **sealed** attributes only. If this happened, we can switch all these to `GetFirstNonDerivedAttributeOrDefault` and delete `GetFirstDerivedAttributeOrDefault`

Youssef1313:
Fixes #4333

### DIFF
--- a/test/UnitTests/MSTestAdapter.PlatformServices.UnitTests/Execution/UnitTestRunnerTests.cs
+++ b/test/UnitTests/MSTestAdapter.PlatformServices.UnitTests/Execution/UnitTestRunnerTests.cs
@@ -260,7 +260,7 @@ public sealed class UnitTestRunnerTests : TestContainer
 
         results.Should().NotBeNull();
         results.Length.Should().Be(1);
-        results[0].Outcome.Should().Be(UTF.UnitTestOutcome.Failed);
+        results[0].Outcome.Should().Be(UTF.UnitTestOutcome.Error);
         results[0].IgnoreReason.Should().Be(expectedMessage);
     }
 


src/Adapter/MSTestAdapter.PlatformServices/Helpers/ReflectHelper.cs
```diff
@@ -87,6 +92,37 @@ internal class ReflectHelper : MarshalByRefObject
...
+        Attribute[] cachedAttributes = GetCustomAttributesCached(attributeProvider);
+
+        TAttribute? foundAttribute = null;
+        foreach (Attribute cachedAttribute in cachedAttributes)
+        {
+            if (cachedAttribute is TAttribute cachedAttributeAsTAttribute)
+            {
+                if (foundAttribute is not null)
+                {
+                    throw new InvalidOperationException($"Multiple attributes of type '{typeof(TAttribute)}' found.");
```
Youssef1313, line 116: TODO: localize.

fhnaseer (approved):

PR #7232, opened 2026-01-15
Localize error message for duplicate attributes

Youssef1313:

### DIFF
--- a/test/IntegrationTests/MSTest.Acceptance.IntegrationTests/CustomAttributesTests.cs
+++ b/test/IntegrationTests/MSTest.Acceptance.IntegrationTests/CustomAttributesTests.cs
@@ -31,7 +31,6 @@ public sealed class CustomAttributesTests : AcceptanceTestBase<CustomAttributesT
         TestHostResult testHostResult = await testHost.ExecuteAsync(cancellationToken: TestContext.CancellationToken);
 
         testHostResult.AssertExitCodeIsNot(ExitCodes.Success);
-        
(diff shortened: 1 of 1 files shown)


src/Adapter/MSTestAdapter.PlatformServices/Resources/Resource.resx
```diff
@@ -455,4 +455,7 @@ but received {4} argument(s), with types '{5}'.</value>
   <data name="UTA_OutRefParametersNotSupported" xml:space="preserve">
     <value>Warning: Test method '{0}.{1}' has 'out' or 'ref' parameters which are not supported and may be disallowed in a future version of MSTest.</value>
   </data>
+  <data name="DuplicateAttributeError" xml:space="preserve">
+    <value>Multiple attributes of type '{0}' were found. Only one is allowed.</value>
```
Evangelink: It's probably fine as-is, I would suggest `Only one attribute of type '{0}' is allowed, but multiple were found.`

test/IntegrationTests/MSTest.Acceptance.IntegrationTests/CustomAttributesTests.cs
```diff
@@ -31,7 +31,8 @@ public async Task DuplicateTestClassAttribute_ShouldFail(string tfm)
         TestHostResult testHostResult = await testHost.ExecuteAsync(cancellationToken: TestContext.CancellationToken);
 
         testHostResult.AssertExitCodeIsNot(ExitCodes.Success);
-        testHostResult.AssertStandardErrorContains("Multiple attributes of type 'Microsoft.VisualStudio.TestTools.UnitTesting.TestClassAttribute' found.");
+
```
Evangelink:
```suggestion

```

PR #7244, opened 2026-01-16
Create GitHub workflow for publishing MSTest/MTP

Youssef1313:


.github/workflows/publish-packages-to-nugetorg.yml
```diff
@@ -0,0 +1,43 @@
...
+          fetch-depth: 0
+
+      - name: Setup .NET SDK
+        uses: actions/setup-dotnet@67a3573c9a986a3f9c594539f4ab511d57bb3ce9 # v4.3.1
+        with:
+          global-json-file: global.json
+
+      - name: Download packages from test-tools feed
+        run: |
+          dotnet run .github/workflows/PackageDownloader.cs -- ${{ github.event.inputs.mstest_version }} ${{ github.event.inputs.mstest_engine_version }} ${{ github.event.inputs.mtp_version }}
```
Youssef1313, line 34: The basic idea here is:

1. Packages are built via our internal pipeline, as usual.
2. Packages are published to test-tools feed from our internal pipeline, as usual.
3. This step will then download the packages from test-tools.
4. The next step will authenticate to nuget. We will set Trusted Publishing for it to work. Unfortunately, trusted publishing isn't yet supported with AzDO Pipelines AFAIK, that's why I'm doing this with GitHub Actions.
5. We push to nuget with the short-living key obtained in the previous step.

This simplifies our current approach to release to nuget.org which is very manual.

@blowdart Does that sound good? Any concerns?

blowdart: It'd work. Do I think it's a good idea? No. Just publish from AzDO like everyone else, because now you not only have to carefully validate actions and azdo flows, which you're used to, but a whole new thing, a single .cs file

Issue #7056, opened 2025-12-05
MTPv2: clean output in GitHub Actions?


bart-vmware:
When using MTPv2 in GitHub Actions, the output contains a lot of garbage that makes it difficult to read the log. For example, a single assembly occurs multiple times:

<img width="2096" height="678" alt="Image" src="https://github.com/user-attachments/assets/4c1865a5-286e-4a56-a07c-bc146517f263" />

Compare that with VSTest, which just logs a clear single line:

<img width="918" height="25" alt="Image" src="https://github.com/user-attachments/assets/f17aadc2-051c-4896-8f20-025331fe1baf" />

Is there a way to produce a clean output in the log on GitHub Actions? Such as per assembly/framework:
- Number of tests passed/failed/skipped
- Each failed test with its name and stack trace

nohwnd:
The progress is reported every 3 seconds for every dll that is currently running. When the UI is not ansi capable this is the only way to report. I am not opposed to disabling the output in non-interactive scenarios. But then you also don't get any sense of what was progress.

bart-vmware:
Using `--no-progress` is more readable, but it doesn't show the number of passed/failed/skipped tests per assembly.

PR #7248, opened 2026-01-17
Disable progress on non-capable terminals

Youssef1313:
On every repo I migrated to MTP, I have always added `--no-progress` in CI. This PR makes that a default behavior. When not using a terminal that is capable of writing the progress in place, we don't write progress altogether. IMO, it's never useful to keep writing progress every 3 seconds and always results in a messy output. We will need to also make a similar change to dotnet/sdk.

Related to https://github.com/microsoft/testfx/issues/7056 and other similar issues (not closing any until work is done for dotnet test).

### DIFF
--- a/src/Platform/Microsoft.Testing.Platform/OutputDevice/Terminal/TestProgressStateAwareTerminal.cs
+++ b/src/Platform/Microsoft.Testing.Platform/OutputDevice/Terminal/TestProgressStateAwareTerminal.cs
@@ -21,8 +21,6 @@ internal sealed partial class TestProgressStateAwareTerminal : IDisposable
 
     private readonly ITerminal _terminal;
     private readonly Func<bool?> _showProgress;
-    private readonly bool _writeProgressImmediatelyAfterOutput;
-    private readonly int _updateEvery;
     private TestProgressState?[] _progressItems = [];
     private bool? _showProgressCached;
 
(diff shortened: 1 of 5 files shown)


src/Platform/Microsoft.Testing.Platform/OutputDevice/Terminal/TestProgressStateAwareTerminal.cs
```diff
@@ -43,7 +41,10 @@ private void ThreadProc()
     {
         try
         {
-            while (!_cts.Token.WaitHandle.WaitOne(_updateEvery))
+            // When writing to ANSI, we update the progress in place and it should look responsive so we
+            // update every half second, because we only show seconds on the screen, so it is good enough.
+            // When writing to non-ANSI, we never show progress as the output can get long and messy.
+            while (!_cts.Token.WaitHandle.WaitOne(500))
```
Evangelink, line 47: Not: use a constant for the timeout and add a comment explaining why this value.

Youssef1313: I'll follow-up.

Youssef1313: @Evangelink https://github.com/microsoft/testfx/pull/7252

PR #7258, opened 2026-01-19
Rework MSTEST0016 message and description

### LINKED ISSUE #5427: Clarify wording and documentation of MSTEST0016
See https://developercommunity.visualstudio.com/t/MSTEST0016-warns-about-non-issue/10881654#T-ND10887700

Evangelink:
Attempts to fix #5427

### DIFF
--- a/src/Analyzers/MSTest.Analyzers/TestClassShouldHaveTestMethodAnalyzer.cs
+++ b/src/Analyzers/MSTest.Analyzers/TestClassShouldHaveTestMethodAnalyzer.cs
@@ -33,12 +33,12 @@ public sealed class TestClassShouldHaveTestMethodAnalyzer : DiagnosticAnalyzer
         isEnabledByDefault: true);
 
     /// <inheritdoc cref="Resources.TestClassShouldHaveTestMethodMessageFormat_BaseClassHasAssemblyAttributes" />
-    public static readonly DiagnosticDescriptor TestClassShouldHaveTestMethodRule_BaseClassHasAssemblyAttributes = TestClassShouldHaveTestMethodRule
+    public static readonly DiagnosticDescriptor TestClassShouldHaveTestMethodRuleBaseClassHasAssemblyAttributes = TestClassShouldHaveTestMethodRule
         .WithMessage(new LocalizableResourceString(nameof(Resources.TestClassShouldHaveTestMethodMessageFormat_BaseClassHasAssemblyAttributes), Resources.ResourceManager, typeof(Resources)));
 
     /// <inheritdoc />
     public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics { get; }
-        = ImmutableArray.Create(TestClassShouldHaveTestMethodRule, TestClassShouldHaveTestMethodRule_BaseClassHasAssemblyAttributes);
+        = ImmutableArray.Create(TestClassShouldHaveTestMethodRule, TestClassShouldHaveTestMethodRuleBaseClassHasAssemblyAttributes);
 
     /// <inheritdoc />
     public override void Initialize(AnalysisContext context)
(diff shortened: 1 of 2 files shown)


src/Analyzers/MSTest.Analyzers/Resources.resx
```diff
@@ -323,10 +323,19 @@ The type declaring these methods should also respect the following rules:
...
+Exception: A 'static' test class is valid without test methods if it contains at least one method marked with '[AssemblyInitialize]', '[AssemblyCleanup]', '[GlobalTestInitialize]', or '[GlobalTestCleanup]'.
+
+Common scenarios where this rule triggers:
+- The class was intended to be a base class containing shared test setup but was incorrectly marked with '[TestClass]'. Consider removing the '[TestClass]' attribute from base classes.
+- Test methods were accidentally removed or commented out.
+- The class contains only helper methods and should not be marked as a test class.</value>
   </data>
   <data name="TestClassShouldHaveTestMethodMessageFormat" xml:space="preserve">
-    <value>Test class '{0}' should have at least one test method or be 'static' with method(s) marked by '[AssemblyInitialize]' and/or '[AssemblyCleanup]'</value>
+    <value>Test class '{0}' does not contain any test method (method marked with '[TestMethod]'). Add a test method, or remove '[TestClass]' if this is a base class or not intended to be a test class.</value>
```
Youssef1313, line 338: Should this message also suggest marking the class as static if it contains asm init/cleanup or global init/cleanup?

Evangelink: I was wondering about it but thought it was better to keep the message short and have all cases in the description. Honestly I don't know if people use description or not.

Youssef1313: Description shows in IDE-only IIRC. So it's less observed by users.

Evangelink: Shall I go for a long message with the static special case then?

Youssef1313: Hmm, what about having different messages based on the case we are reporting?

If we were able to detect a fixture, we produce a message that is more direct about marking the class as static. Otherwise, we produce a message about missing test methods.

That way, we can still have short messages, but be precise per the scenario we are reporting.

src/Analyzers/MSTest.Analyzers/TestClassShouldHaveTestMethodAnalyzer.cs
```diff
@@ -32,9 +32,13 @@ public sealed class TestClassShouldHaveTestMethodAnalyzer : DiagnosticAnalyzer
         DiagnosticSeverity.Info,
         isEnabledByDefault: true);
 
+    /// <inheritdoc cref="Resources.TestClassShouldHaveTestMethodMessageFormat_BaseClassHasAssemblyAttributes" />
+    public static readonly DiagnosticDescriptor TestClassShouldHaveTestMethodRule_BaseClassHasAssemblyAttributes = TestClassShouldHaveTestMethodRule
```
Evangelink: @copilot please don't use underscores in the field name

src/Analyzers/MSTest.Analyzers/TestClassShouldHaveTestMethodAnalyzer.cs
```diff
@@ -98,18 +105,44 @@ private static void AnalyzeSymbol(SymbolAnalysisContext context, INamedTypeSymbo
...
+        }
+
+        // Non-static class that inherits assembly attributes from base class - suggest making it static
+        if (!classSymbol.IsStatic && hasAssemblyAttributeInBaseClass && baseClassWithAssemblyAttribute is not null)
         {
-            context.ReportDiagnostic(classSymbol.CreateDiagnostic(TestClassShouldHaveTestMethodRule, classSymbol.Name));
+            context.ReportDiagnostic(classSymbol.CreateDiagnostic(TestClassShouldHaveTestMethodRuleBaseClassHasAssemblyAttributes, classSymbol.Name, baseClassWithAssemblyAttribute.Name));
+
+            return;
         }
```
Youssef1313, line 143: I'm not sure if this is really addressing the original reported complaint.

Per the description in the linked DC ticket, I think a repro would be something like this:

```csharp
[TestClass]
public abstract class TestClass
{
    [AssemblyInitialize]
    public static void M(TestContext _) { }
}
```

So, asm init is in the same class being analyzed. Given that the class doesn't have any non-static members, the correct action user needs to take is to mark the class `static` instead of `abstract`, and they don't need to inherit from it anywhere.

A different scenario to consider in this case is when the abstract class have other non-static members. In this case, we might have two options:

1. Not report a diagnostic at all and consider it okay.
2. Report a diagnostic and suggest that the fixtures get moved to a different class that is marked as static.

PR #7266, opened 2026-01-20
Use abstraction over EqtTrace on platform services

### LINKED ISSUE #2140: Use logger interface and not EqtTrace logger
## Summary

MSTest is currently using eqt trace logger everywhere for technical logs. This is preventing any logs to be produced when working in runner mode. There is also many limitations to how EqtTrace works that can cause locks and slowness.

Instead we should flow some interface and change the logger based on the actual runner.

Evangelink:
Fixes #2140


Evangelink:
Locally the test `TestDataSourceFromAppConfig` consistently fails, I can repro doing it inside the playground. There is a crash at the end of the lambda of `TestMethodRunner`. I haven't been able to diagnose why it's failing but it seems the error doesn't appear in CI.

PR #7271, opened 2026-01-21
Reduce platform services dependency on VSTest

Youssef1313:
The end goal is that we are using platform-agnostic abstraction for any core logic.

So, platform services should have minimal dependency on VSTest or MTP. This PR moves EqtTrace usages from platform services to MSTest.TestAdapter instead.

This also unifies our logging strategies, and ensure it works correctly in appdomains.


src/Adapter/MSTest.TestAdapter/TestingPlatformAdapter/MSTestBridgedTestFramework.cs
```diff
@@ -37,6 +38,8 @@ protected override Task SynchronizedDiscoverTestsAsync(VSTestDiscoverTestExecuti
         }
 
         PlatformServiceProvider.Instance.AdapterTraceLogger = new BridgedTraceLogger(_loggerFactory.CreateLogger("mstest-trace"));
+        // TODO: Merge AdapterTraceLogger above with TraceLoggerHelper.Instance here, and have a sensible implementation other than NopTraceLogger.
```
Youssef1313: Note to self:

Validate the behavior of the existing AdapterTraceLogger and the new TraceLoggerHelper with appdomain enabled in the following cases:

1. VSTest, logging from both discovery and execution AppDomain.
2. MTP, logging from both discovery and execution AppDomain.

It's probably best to merge the two in this PR.

PR #7272, opened 2026-01-21
Move adapter utilities to adapter project

Evangelink:


Evangelink:
Closing as we will go with changes from Youssef in #7271 and #7277

PR #7315, opened 2026-01-30
[Experimental] Add TestContext.Current

### LINKED ISSUE #6137: Consider exposing TestContext.Current static property
We have it internal already here:

https://github.com/microsoft/testfx/blob/cd0fd625fd72bad3483d2665ae04f18fc0c65bed/src/Adapter/MSTestAdapter.PlatformServices/Services/TestContextImplementation.cs#L113

I think it's much easier to use it directly this way.

Benefits:

- No need to inject it via constructor or property
- No need to have it as a parameter on fixture methods
- Its availability via static async local will hopefully prevent users from incorrect usages of TestContext and passing it around. At any place the user needs TestContext, they can simply access the correct one (if available).
- We will keep all the old ways supported for backward compatibility, but we start documenting that it's not the recommended way.

Evangelink:
Fix #6137


src/Adapter/MSTestAdapter.PlatformServices/Services/TestContextImplementation.cs
```diff
@@ -327,10 +327,16 @@ public void Dispose()
     internal readonly struct ScopedTestContextSetter : IDisposable
     {
         internal ScopedTestContextSetter(TestContextImplementation? testContext)
-            => CurrentTestContextAsyncLocal.Value = testContext;
+        {
+            CurrentTestContextAsyncLocal.Value = testContext;
+            TestContext.SetCurrentContext(testContext);
+        }
```
Youssef1313: It's best to unify this so we have a single value and a single source of truth.

src/TestFramework/TestFramework.Extensions/TestContext.cs
```diff
@@ -15,6 +15,26 @@ namespace Microsoft.VisualStudio.TestTools.UnitTesting;
...
+    /// it returns <see langword="null"/>.
+    /// </remarks>
+    [Experimental("MSTESTEXP", UrlFormat = "https://aka.ms/mstest/diagnostics#{0}")]
+    public static TestContext? Current => CurrentContext.Value;
+
+    /// <summary>
+    /// Sets the current <see cref="TestContext"/> for the async context.
+    /// </summary>
+    /// <param name="testContext">The test context to set, or <see langword="null"/> to clear.</param>
+    [EditorBrowsable(EditorBrowsableState.Never)]
```
Youssef1313: Feels redundant.

src/Adapter/MSTestAdapter.PlatformServices/Services/TestContextImplementation.cs
```diff
@@ -326,11 +322,16 @@ public void Dispose()
 
     internal readonly struct ScopedTestContextSetter : IDisposable
     {
+        private readonly TestContext? _previousContext;
+
         internal ScopedTestContextSetter(TestContextImplementation? testContext)
-            => CurrentTestContextAsyncLocal.Value = testContext;
+        {
+            _previousContext = TestContext.Current;
```
Youssef1313: Is this ever non-null? And if so, do we have a realistic use case?

PR #7317, opened 2026-01-30
Fix stack trace assertions for Release mode on .NET Framework

### LINKED ISSUE #5109: Investigate why stacktrace are empty in release mode
Some tests are checking the `StackTrace` and it's either null or empty when in release mode. We have some helper doing cleanup of them so I am wondering if the logic is not working properly with MTP as the processing is different (more "normal" app like).

AB#2404445

Evangelink:
Use [MethodImpl(NoInlining)] on .NET Framework to prevent JIT from inlining small methods, which would remove stack frames. The assertions now check for the DummyTestClass method frames instead of lambda frames.

Fixes #5109

### DIFF
--- a/test/UnitTests/MSTestAdapter.PlatformServices.UnitTests/Execution/TestMethodInfoTests.cs
+++ b/test/UnitTests/MSTestAdapter.PlatformServices.UnitTests/Execution/TestMethodInfoTests.cs
@@ -1485,53 +1485,33 @@ public class TestMethodInfoTests : TestContainer
             }
         }
 
-        // On .NET Framework, the JIT aggressively inlines small methods in Release mode,
-        // which removes stack frames. NoInlining ensures the method appears in stack traces.
-#if NETFRAMEWORK
         [MethodImpl(MethodImplOptions.NoInlining | MethodImplOptions.NoOptimization)]
-#endif
         public void DummyTestInitializeMethod() => TestInitializeMethodBody(this);
 
-#if NETFRAMEWORK
         [MethodImpl(MethodImplOptions.NoInlining | MethodImplOptions.NoOptimization)]
-#endif
         public async Task DummyTestInitializeMethodAsync() => await TestInitializeMethodBodyAsync(this);
 
-#if NETFRAMEWORK
         [MethodImpl(MethodImplOptions.NoInlining | MethodImplOptions.NoOptimization)]
-#endif
         public void DummyTestCleanupMethod() => TestCleanupMethodBody(this);
 
-#if NETFRAMEWORK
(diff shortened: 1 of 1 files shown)


test/UnitTests/MSTestAdapter.PlatformServices.UnitTests/Execution/TestMethodInfoTests.cs
```diff
@@ -1470,33 +1455,53 @@ public TestContext TestContext
             }
         }
 
+        // On .NET Framework, the JIT aggressively inlines small methods in Release mode,
+        // which removes stack frames. NoInlining ensures the method appears in stack traces.
+#if NETFRAMEWORK
         [MethodImpl(MethodImplOptions.NoInlining | MethodImplOptions.NoOptimization)]
+#endif
```
Youssef1313: Here we used to have the NoInlining unconditionally. Any reason to change this?

test/UnitTests/MSTestAdapter.PlatformServices.UnitTests/Execution/TestMethodInfoTests.cs
```diff
@@ -937,10 +924,8 @@ public async Task TestMethodInfoInvokeShouldAppendStackTraceInformationIfBothTes
 
         result.Outcome.Should().Be(UTF.UnitTestOutcome.Failed);
         exception.Should().NotBeNull();
-#if DEBUG
         ((TestFailedException)exception.InnerExceptions[0]).StackTraceInformation!.ErrorStackTrace.Contains("Microsoft.VisualStudio.TestPlatform.MSTestAdapter.UnitTests.Execution.TestMethodInfoTests.DummyTestClass.DummyTestMethod()").Should().BeTrue();
```
Youssef1313, line 957: I'm quite lost because what I recall is that we have had failures due to missing line numbers in stack traces. But looking at the assertions, we are not even validating the existence of file path or line numbers.

Youssef1313 (approved):
I'm concerned about not fully knowing what have changed over time for these tests. But given this is not any product change, I'm approving.

test/UnitTests/MSTestAdapter.PlatformServices.UnitTests/Execution/TestMethodInfoTests.cs
Evangelink: I don't think so, I added the conditional as a cleanup before submitting the PR so I most likely edited too much.

Evangelink: I don't think so, I added the conditional as a cleanup before submitting the PR so I most likely edited too much.

PR #7318, opened 2026-01-30
Add settings to launch debugger on test failure

### LINKED ISSUE #2033: Option to launch debugger on test failure
## Summary

It would be useful to be able to automatically trigger a debugger launch call on failed test/assertion. This is especially interesting when the issue is quite randomic and we are not having enough info to try to replicate the issue.

This should not be the default configuration but more an option that can be enabled. We could consider having some specific filter flag too to avoid launching the debugger on any failed test but rather on a list of tests.

cc @MarcoRossignoli @nohwnd

Evangelink:
Fixes #2033.

I have thought about having this as a platform feature but there is really few code to add. Besides, it would need to be introduced through a new mechanism as the message bus we are using could result in the debugger launch being triggered pretty far from the time of error.

### DIFF
--- a/src/Adapter/MSTestAdapter.PlatformServices/MSTestSettings.cs
+++ b/src/Adapter/MSTestAdapter.PlatformServices/MSTestSettings.cs
@@ -666,12 +666,7 @@ internal sealed class MSTestSettings
                     case "LAUNCHDEBUGGERONASSERTIONFAILURE":
                         {
                             string value = reader.ReadInnerXml();
-                            // Support backward compatibility with boolean values
-                            if (bool.TryParse(value, out result))
-                            {
-                                settings.LaunchDebuggerOnAssertionFailure = result ? DebuggerLaunchMode.Enabled : DebuggerLaunchMode.Disabled;
-                            }
-                            else if (TryParseEnum(value, out DebuggerLaunchMode mode))
+                            if (TryParseEnum(value, out DebuggerLaunchMode mode))
                             {
                                 settings.LaunchDebuggerOnAssertionFailure = mode;
                             }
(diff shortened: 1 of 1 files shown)


src/TestFramework/TestFramework/Assertions/Assert.cs
```diff
@@ -37,8 +37,22 @@ private Assert()
     [DoesNotReturn]
     [StackTraceHidden]
     internal static void ThrowAssertFailed(string assertionName, string? message)
-        => throw new AssertFailedException(
+    {
+        if (AssertionFailureSettings.LaunchDebuggerOnFailure)
```
Evangelink: This is acting only on assertion failures, we could also consider the same call in case of an exception being thrown or a timeout occuring in the test.

aha8med7788-svg (approved):
@copilot

src/Adapter/MSTestAdapter.PlatformServices/MSTestSettings.cs
```diff
@@ -656,6 +663,26 @@ private static MSTestSettings ToSettings(XmlReader reader, IMessageLogger? logge
                             break;
                         }
 
+                    case "LAUNCHDEBUGGERONASSERTIONFAILURE":
+                        {
+                            string value = reader.ReadInnerXml();
+                            // Support backward compatibility with boolean values
```
Youssef1313: We don't have backcompat concerns here as it's a new feature. Do we still need to support both the bool and the enum?

src/Adapter/MSTestAdapter.PlatformServices/MSTestSettings.cs
```diff
@@ -783,6 +810,28 @@ private static void ParseBooleanSetting(IConfiguration configuration, string key
...
 
+    private static void ParseDebuggerLaunchModeSetting(IConfiguration configuration, string key, IMessageLogger? logger, Action<DebuggerLaunchMode> setSetting)
+    {
+        if (configuration[$"mstest:{key}"] is not string value)
+        {
+            return;
+        }
+
+        // Support backward compatibility with boolean values
+        if (bool.TryParse(value, out bool boolResult))
```
Youssef1313: Same here.

src/Adapter/MSTestAdapter.PlatformServices/MSTestSettings.cs
Evangelink: My bad, I haven't reviewed the actual update from copilot here.

PR #7355, opened 2026-02-06
Add `Assert.Scope()` for soft assertions

### LINKED ISSUE #571: Is there any way to soft assert in MSTEST just like verify assert in TESTNG

Evangelink:
Fixes #571

### DIFF
--- a/test/IntegrationTests/MSTest.Acceptance.IntegrationTests/SoftAssertionTests.cs
+++ b/test/IntegrationTests/MSTest.Acceptance.IntegrationTests/SoftAssertionTests.cs
@@ -41,7 +41,7 @@ public sealed class SoftAssertionTests : AcceptanceTestBase<SoftAssertionTests.T
         // Validate the output includes the aggregate message and that inner exception stack traces
         // point to the test method (assertion call site).
         testHostResult.AssertOutputMatchesRegex(
-            """failed ScopeWithMultipleFailures \(\d+ms\)[\s\S]+2 assertion\(s\) failed within the assert scope\.[\s\S]+Assert\.AreEqual failed\. Expected:<1>\. Actual:<2>\.[\s\S]+at UnitTest1\.ScopeWithMultipleFailures\(\)[\s\S]+Assert\.IsTrue failed\.[\s\S]+at UnitTest1\.ScopeWithMultipleFailures\(\)""");
+            """failed ScopeWithMultipleFailures \(\d+ms\)[\s\S]+2 assertion\(s\) failed within the assert scope\.[\s\S]+at Microsoft\.VisualStudio\.TestTools\.UnitTesting\.AssertScope\.Dispose\(\)[\s\S]+at UnitTest1\.ScopeWithMultipleFailures\(\)""");
     }
 
     [TestMethod]


src/TestFramework/TestFramework/Assertions/Assert.cs
```diff
@@ -37,8 +37,19 @@ private Assert()
     [DoesNotReturn]
     [StackTraceHidden]
     internal static void ThrowAssertFailed(string assertionName, string? message)
```
Evangelink, line 39: Given this PR and #2033, I think it'd be good to expose a "ReportFailure" method that could be used by devs extending MSTest assertions so they would benefits from the various features (launch debugger, soft assertions....). I am not doing it here because we can decide not to go with it for now.

src/TestFramework/TestFramework/Assertions/Assert.cs
```diff
@@ -37,8 +37,19 @@ private Assert()
...
-        => throw new AssertFailedException(
-            string.Format(CultureInfo.CurrentCulture, FrameworkMessages.AssertionFailed, assertionName, message));
+    {
+        var assertionFailedException = new AssertFailedException(string.Format(CultureInfo.CurrentCulture, FrameworkMessages.AssertionFailed, assertionName, message));
+        AssertScope? scope = AssertScope.Current;
+        if (scope is not null)
+        {
+            scope.AddError(assertionFailedException);
+#pragma warning disable CS8763 // A method marked [DoesNotReturn] should not return.
```
Evangelink: @Youssef1313 I don't know what we prefer but it feels better for me to say we continue to help compiler with the default behavior and we know we may have some FPs when code is used under assertion scope. It would otherwise be a breaking change for many people if we had to update all assertion APIs to no longer respect some of these DoesNotReturn compilation indication.

src/TestFramework/TestFramework/Assertions/Assert.Scope.cs
```diff
@@ -0,0 +1,29 @@
...
+    /// <code>
+    /// using (Assert.Scope())
+    /// {
+    ///     Assert.AreEqual(1, 2);  // collected, not thrown
+    ///     Assert.IsTrue(false);   // collected, not thrown
+    /// }
+    /// // AssertFailedException is thrown here with all collected failures.
+    /// </code>
+    /// </example>
+    public static IDisposable Scope() => new AssertScope();
```
Evangelink, line 29: Public API is quite limited, do we want to make it experimental still?

src/TestFramework/TestFramework/Assertions/AssertScope.cs
```diff
@@ -0,0 +1,53 @@
...
+{
+    private static readonly AsyncLocal<AssertScope?> CurrentScope = new();
+
+    private readonly List<AssertFailedException> _errors = [];
+    private readonly AssertScope? _previousScope;
+    private bool _disposed;
+
+    internal AssertScope()
+    {
+        _previousScope = CurrentScope.Value;
```
Youssef1313: Do we want to allow nested scopes? I don't personally see a good use case.

src/TestFramework/TestFramework/Assertions/AssertScope.cs
```diff
@@ -0,0 +1,53 @@
...
+    /// <summary>
+    /// Gets the current active <see cref="AssertScope"/>, or <see langword="null"/> if no scope is active.
+    /// </summary>
+    internal static AssertScope? Current => CurrentScope.Value;
+
+    /// <summary>
+    /// Adds an assertion failure message to the current scope.
+    /// </summary>
+    /// <param name="error">The assertion failure message.</param>
+    internal void AddError(AssertFailedException error) => _errors.Add(error);
```
Youssef1313: Consider throwing ObjectDisposedException here if `_disposed` is true.

src/TestFramework/TestFramework/Assertions/AssertScope.cs
```diff
@@ -0,0 +1,53 @@
...
+    {
+        if (_disposed)
+        {
+            return;
+        }
+
+        _disposed = true;
+        CurrentScope.Value = _previousScope;
+
+        if (_errors.Count > 0)
```
Youssef1313: Can we special case when `_errors.Count == 1` so we don't create `AggregateException`?

src/TestFramework/TestFramework/Assertions/AssertScope.cs
```diff
@@ -0,0 +1,53 @@
...
+
+/// <summary>
+/// Represents a scope in which assertion failures are collected instead of thrown immediately.
+/// When the scope is disposed, all collected failures are thrown as a single <see cref="AssertFailedException"/>.
+/// </summary>
+internal sealed class AssertScope : IDisposable
+{
+    private static readonly AsyncLocal<AssertScope?> CurrentScope = new();
+
+    private readonly List<AssertFailedException> _errors = [];
```
Youssef1313: Should we use a thread-safe collection here?

For example, in theory user can create a scope, then create X threads and starts them, then each thread does some piece of work and asserts some result. Then waits for all those threads to finish before disposing the scope.

src/TestFramework/TestFramework/Assertions/Assert.Scope.cs
Youssef1313: Fine for me (and probably best) to be experimental, but curious what exactly do you think is limited.

Evangelink: I mean we are effectively only introducing the API but no extra return type nor special logic, meaning we can easily make changes in case of a problem. I'll still go wth the experimental.

src/TestFramework/TestFramework/Assertions/AssertScope.cs
```diff
@@ -0,0 +1,67 @@
...
+        {
+            return;
+        }
+
+        _disposed = true;
+        CurrentScope.Value = null;
+
+        if (_errors.Count == 1 && _errors.TryDequeue(out AssertFailedException? singleError))
+        {
+            throw singleError;
```
Evangelink: @Youssef1313 am I not sure if it's better to simply throw the exception or still wrap it under an exception to say it's assert scope which failed. What's your preference?

src/TestFramework/TestFramework/Assertions/Assert.Scope.cs
Youssef1313: Got it. I think experimental is still useful in case we wanted to roll-back the feature completely.

src/TestFramework/TestFramework/Assertions/AssertScope.cs
Youssef1313: I would say throw as-is for the simple case of single failure.

test/UnitTests/TestFramework.UnitTests/Assertions/AssertTests.ScopeTests.cs
```diff
@@ -0,0 +1,127 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using AwesomeAssertions;
+
+namespace Microsoft.VisualStudio.TestPlatform.TestFramework.UnitTests;
+
+public partial class AssertTests
```
Youssef1313, line 8: Can we add some integration tests, with more scenarios like:

```csharp
using _ = Assert.Scope();
string x = null;
Assert.IsNotNull(x);
Assert.AreEqual(1, x.Length); // throws NullReferenceException
```

How are the exceptions going to aggregate in this case?

src/TestFramework/TestFramework/Assertions/Assert.cs
Youssef1313: I think it might be better to attempt removing those `DoesNotReturn` attributes so we have more accurate representation for both scenarios.

Potentially problematic APIs:

- Assert.Fail: this is one API where we publicly flow the `DoesNotReturn`. I cannot see any use case for `Assert.Fail` to be called under assertion scope. Maybe `Assert.Fail` should always throw immediately even when used when a scope is set?
- Assert.ContainsSingle: what happens for its return value when the collection is empty or having multiple elements?
- The internal `CheckParameterNotNull` helper. Maybe this one should be throwing ArgumentNullException instead?
- `IsExactInstanceOfType`/`IsInstanceOfType`/`IsNotNull` can no longer meet the NotNull constraints correctly.

I'm not sure what's the best to do here 😕

docs/RFCs/011-Soft-Assertions-Nullability-Design.md
```diff
@@ -0,0 +1,263 @@
...
+- `IsNotNull` — annotated `[NotNull]` on the value parameter
+- `IsInstanceOfType` — annotated `[NotNull]` on the value parameter
+- `IsExactInstanceOfType` — annotated `[NotNull]` on the value parameter
+- `Fail` — semantically means "unconditional failure"; annotated `[DoesNotReturn]` on public API
+- `ContainsSingle` — returns the matched element; returning `default` in soft mode would give callers a bogus `null`/`default(T)` causing downstream errors
+
+**Tier 2 — Soft, but annotations removed:** Assertions that had conditional `[DoesNotReturnIf]` annotations. The annotation is removed so the compiler no longer assumes the condition is guaranteed. The assertions become soft (collected within a scope).
+
+- `IsTrue` — `[DoesNotReturnIf(false)]` removed
+- `IsFalse` — `[DoesNotReturnIf(true)]` removed
```
Youssef1313, line 90: Is this going to disrupt users who rely on these annotations?

e.g, users who use `Assert.IsTrue(x is not null)` will be impacted in a similar way exactly as those using `Assert.IsNotNull(x)`.

docs/RFCs/011-Soft-Assertions-Nullability-Design.md
```diff
@@ -0,0 +1,263 @@
...
+using (Assert.Scope())
+{
+    Assert.AreEqual(1, actual.X);  // failure collected, execution continues
+    Assert.AreEqual(2, actual.Y);  // failure collected, execution continues
+    Assert.IsTrue(actual.IsValid); // failure collected, execution continues
+}
+// Dispose() throws AggregateException-like AssertFailedException with all 3 failures
+```
+
+### The nullability problem
```
Youssef1313: Pinging @RikkiGibson here as well, in case there are any ideas we haven't considered.

docs/RFCs/011-Soft-Assertions-Nullability-Design.md
Evangelink: Yes that's the case. These 2 APIs can be used to do many things so I wasn't sure about what to do. I am fine to get started with them being hard assertions and wait for feedback. This is what would have the lowest impact.

docs/RFCs/011-Soft-Assertions-Nullability-Design.md
```diff
@@ -0,0 +1,263 @@
...
+
+**Tier 2 — Soft, but annotations removed:** Assertions that had conditional `[DoesNotReturnIf]` annotations. The annotation is removed so the compiler no longer assumes the condition is guaranteed. The assertions become soft (collected within a scope).
+
+- `IsTrue` — `[DoesNotReturnIf(false)]` removed
+- `IsFalse` — `[DoesNotReturnIf(true)]` removed
+
+**Tier 3 — Soft, no annotation impact:** All other assertions that don't carry type-narrowing annotations. These become fully soft within a scope.
+
+- `AreEqual`, `AreNotEqual`, `AreSame`, `AreNotSame`
+- `Inconclusive`
```
Evangelink: For this one, I also wonder if there is any case where it's interesting/useful to continue. Maybe this can be a hard assertion too.

src/TestFramework/TestFramework/Assertions/Assert.cs
```diff
@@ -36,7 +39,52 @@ private Assert()
...
+    /// <param name="assertionName">
+    /// name of the assertion throwing an exception.
+    /// </param>
+    /// <param name="message">
+    /// The assertion failure message.
+    /// </param>
+    [StackTraceHidden]
+    internal static void ReportSoftAssertFailure(string assertionName, string? message)
+    {
+        var assertionFailedException = new AssertFailedException(string.Format(CultureInfo.CurrentCulture, FrameworkMessages.AssertionFailed, assertionName, message));
```
Evangelink:
```suggestion
        LaunchDebuggerIfNeeded();
        var assertionFailedException = new AssertFailedException(string.Format(CultureInfo.CurrentCulture, FrameworkMessages.AssertionFailed, assertionName, message));
```

docs/RFCs/011-Soft-Assertions-Nullability-Design.md
RikkiGibson: It feels like this is an issue with postconditions in general... `Assert.Scoped()` means assertions are no longer enforcing postconditions. If you had a test helper doing something like:
- `Assert.AreEqual("blah", item.Prop);`
- `MyTestHelper(item.Prop);` which explodes in some unhelpful way when we didn't first verify Prop had expected form

well, then using `Assert.Scoped()` will make `MyTestHelper()` behave badly.

I would honestly start by just trying leaving all the nullable attributes on, but not actually ensuring any of the postconditions, when in `Assert.Scoped()` context. Since that is essentially what you are doing already with all postconditions unrelated to nullability. See how that works in practice, and, if the usability feels bad, you could consider introducing certain assertions that throw regardless of whether you're in scoped context or not.

Indeed maybe the user would end up wanting to update MyTestHelper to straight up `Assert.Fail()` in some way that always throws so it can do the most helpful thing whether it is used in scoped or non-scoped context. That would just be part of the adoption/onboarding cost of using the feature.

RikkiGibson: A more likely place this might occur is in the following:

```cs
Assert.Equal(2, items.Length);
Assert.Equal("blah", items[0].Value); // possibly out of bounds since Length postcondition is not checked
```

Evangelink: Thank you @RikkiGibson for your feedback and view here!

src/TestFramework/TestFramework/Assertions/Assert.IsNull.cs
```diff
@@ -199,5 +199,5 @@ public static void IsNotNull([NotNull] object? value, string? message = "", [Cal
 
     [DoesNotReturn]
     private static void ThrowAssertIsNotNullFailed(string? message)
```
Youssef1313: Should this helper and all similar helpers be renamed to `Report` instead of `Throw` for consistency?

Also can you please revise the RFC for naming when discussing this topic?

src/TestFramework/TestFramework/Assertions/Assert.Fail.cs
```diff
@@ -22,5 +22,5 @@ public sealed partial class Assert
     /// </exception>
     [DoesNotReturn]
     public static void Fail(string message = "")
-        => ThrowAssertFailed("Assert.Fail", BuildUserMessage(message));
+        => ReportAssertFailed("Assert.Fail", BuildUserMessage(message), forceThrow: true);
```
Youssef1313: Should we call `ThrowAssertFailed` right away instead of `ReportAssertFailed`?

src/TestFramework/TestFramework/Assertions/Assert.Inconclusive.cs
Youssef1313, line 1: Do we want this to also always-throw immediately?

src/TestFramework/TestFramework/Assertions/Assert.cs
```diff
@@ -26,17 +26,48 @@ private Assert()
...
+        if (!forceThrow)
+        {
+            AssertScope? scope = AssertScope.Current;
+            if (scope is not null)
+            {
+                scope.AddError(assertionFailedException);
+                return;
+            }
+        }
```
Youssef1313:
```suggestion
        if (!forceThrow && AssertScope.Current is { } scope)
        {
            scope.AddError(assertionFailedException);
            return;
        }
```

src/TestFramework/TestFramework/Assertions/Assert.cs
```diff
@@ -26,17 +26,48 @@ private Assert()
...
     [DoesNotReturn]
     [StackTraceHidden]
-    internal static void ThrowAssertFailed(string assertionName, string? message)
+    internal static void ThrowAssertFailed(AssertFailedException exception)
+    {
+        LaunchDebuggerIfNeeded();
+        throw exception;
+    }
+
+    [StackTraceHidden]
```
Youssef1313: I don't think `LaunchDebuggerIfNeeded` needs `StackTraceHidden`

src/TestFramework/TestFramework/Assertions/Assert.cs
```diff
@@ -26,17 +26,48 @@ private Assert()
...
+    internal static void ReportAssertFailed(string assertionName, string? message, bool forceThrow = false)
+    {
+        var assertionFailedException = new AssertFailedException(string.Format(CultureInfo.CurrentCulture, FrameworkMessages.AssertionFailed, assertionName, message));
+        if (!forceThrow)
+        {
+            AssertScope? scope = AssertScope.Current;
+            if (scope is not null)
+            {
+                scope.AddError(assertionFailedException);
```
Youssef1313: Should we also launch debugger here is needed as well?

src/TestFramework/TestFramework/Assertions/Assert.Inconclusive.cs
Youssef1313: Oh it already does. That's good :)

src/TestFramework/TestFramework/Assertions/Assert.That.cs
Youssef1313, line 11: `Assert.That` seems to always throw?

src/TestFramework/TestFramework/Assertions/AssertScope.cs
```diff
@@ -0,0 +1,71 @@
...
+        {
+            return;
+        }
+
+        _disposed = true;
+        CurrentScope.Value = null;
+
+        if (_errors.Count == 1 && _errors.TryDequeue(out AssertFailedException? singleError))
+        {
+            Assert.ThrowAssertFailed(singleError);
```
Youssef1313: Can we just `throw singleError;` here?
We should have already launched debugger when the error was about to be added to the scope, and I don't think it's worth attempting to launch debugger here again as well.

src/TestFramework/TestFramework/Assertions/AssertScope.cs
```diff
@@ -0,0 +1,71 @@
...
+        }
+
+        if (!_errors.IsEmpty)
+        {
+            Assert.ThrowAssertFailed(new AssertFailedException(
```
Youssef1313: Same comment here about throwing directly to skip the launch debugger call.

test/UnitTests/TestFramework.UnitTests/Assertions/AssertTests.ScopeTests.cs
```diff
@@ -0,0 +1,216 @@
...
+    {
+        Action action = () =>
+        {
+            using (Assert.Scope())
+            {
+                Assert.AreEqual(1, 2);
+            }
+        };
+
+        action.Should().Throw<AssertFailedException>()
```
Youssef1313, line 30: Should we assert that it is the Dispose specifically that throws?
like:

```
var scope = Assert.Scope();
Assert.AreEqual(1, 2);
Action action = () => scope.Dispose();

action.Should()...
```

Youssef1313 (approved):
Approved. But I would love to see a full E2E test

src/TestFramework/TestFramework/Assertions/AssertScope.cs
```diff
@@ -59,16 +61,22 @@ public void Dispose()
...
+        // Snapshot the ConcurrentQueue into an array to avoid multiple O(n) enumerations
+        // (ConcurrentQueue<T>.Count is O(n)) and to get a consistent view for branching,
+        // message formatting, and building the AggregateException.
+        AssertFailedException[] errorsSnapshot = _errors.ToArray();
+        if (errorsSnapshot.Length == 1)
         {
-            throw singleError;
+            // Use ExceptionDispatchInfo to preserve the original stack trace captured at the
+            // assertion call site, rather than resetting it to point at Dispose.
+            ExceptionDispatchInfo.Capture(errorsSnapshot[0]).Throw();
```
Youssef1313: Is this the right place to capture? Or should it be in `AddError` instead?

Youssef1313: And how do we handle that for the case of multiple exceptions?

src/TestFramework/TestFramework/Assertions/Assert.cs
```diff
@@ -37,6 +37,39 @@ private Assert()
...
+    /// </param>
+#pragma warning disable CS8763 // A method marked [DoesNotReturn] should not return - Deliberately keeping [DoesNotReturn] annotation while using soft assertions. Within an AssertScope, the postcondition is not enforced (same as all other assertion postconditions in scoped mode).
+    [DoesNotReturn]
+    [StackTraceHidden]
+    internal static void ReportAssertFailed(string assertionName, string? message)
+    {
+        LaunchDebuggerIfNeeded();
+        AssertFailedException assertionFailedException = CreateAssertFailedException(assertionName, message);
+        if (AssertScope.Current is { } scope)
+        {
```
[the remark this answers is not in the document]
Evangelink: Addressed. The throw/catch pattern in `ReportAssertFailed` captures the stack trace at the assertion call site, and in `Dispose`, `ExceptionDispatchInfo` preserves it during re-throw. Additionally, moved `ExceptionDispatchInfo.Capture` to `AddError` per @Youssef1313's suggestion (see thread below) — this captures the dispatch info at the earliest point where the error is stored, making the design cleaner.

src/TestFramework/TestFramework/Assertions/AssertScope.cs
Evangelink: Moved `ExceptionDispatchInfo.Capture` into `AddError` — it now stores `ExceptionDispatchInfo` in the queue instead of raw exceptions. This captures the dispatch info at the earliest point, right where the error is received.

For multiple exceptions: in `Dispose`, we extract `SourceException` from each `ExceptionDispatchInfo` using `Array.ConvertAll(errorsSnapshot, static edi => edi.SourceException)` and pass those to the `AggregateException`. Each inner exception retains its original stack trace (populated via the throw/catch in `ReportAssertFailed`).

For single exception: `errorsSnapshot[0].Throw()` re-throws preserving the original call-site stack trace.

The queue type is now `ConcurrentQueue<ExceptionDispatchInfo>` and `AddError` does:
```csharp
_errors.Enqueue(ExceptionDispatchInfo.Capture(error));
```

PR #7380, opened 2026-02-10
MTP extensions cleanup for cancellation

### LINKED ISSUE #5063: Revise our usage of `return;` after `IsCancellationRequested` check

Youssef1313:
Closes #5063

This is more aligned with the usual cancellation behavior of most BCL APIs, and in many cases we already pass the token to BCL APIs, so we at least now guarantee having a unified behavior (if there are any noticeable behavior differences).

### DIFF
--- a/src/Platform/Microsoft.Testing.Extensions.HangDump/HangDumpActivityIndicator.cs
+++ b/src/Platform/Microsoft.Testing.Extensions.HangDump/HangDumpActivityIndicator.cs
@@ -83,11 +83,7 @@ internal sealed class HangDumpActivityIndicator : IDataConsumer, ITestSessionLif
     {
         CancellationToken cancellationToken = testSessionContext.CancellationToken;
         ApplicationStateGuard.Ensure(_namedPipeClient is not null);
-
-        if (!await IsEnabledAsync().ConfigureAwait(false) || cancellationToken.IsCancellationRequested)
-        {
-            return;
-        }
+        cancellationToken.ThrowIfCancellationRequested();
 
         try
         {
@@ -139,8 +135,8 @@ internal sealed class HangDumpActivityIndicator : IDataConsumer, ITestSessionLif
 
     public async Task ConsumeAsync(IDataProducer dataProducer, IData value, CancellationToken cancellationToken)
     {
-        if (cancellationToken.IsCancellationRequested
-            || value is not TestNodeUpdateMessage nodeChangedMessage)
+        cancellationToken.ThrowIfCancellationRequested();
+        if (value is not TestNodeUpdateMessage nodeChangedMessage)
         {
             return;
         }
--- a/src/Platform/Microsoft.Testing.Extensions.TrxReport/TrxDataConsumer.cs
+++ b/src/Platform/Microsoft.Testing.Extensions.TrxReport/TrxDataConsumer.cs
@@ -104,10 +104,7 @@ internal sealed class TrxReportGenerator :
 
     public Task ConsumeAsync(IDataProducer dataProducer, IData value, CancellationToken cancellationToken)
     {
-        if (!_isEnabled || cancellationToken.IsCancellationRequested)
-        {
-            return Task.CompletedTask;
-        }
+        cancellationToken.ThrowIfCancellationRequested();
 
         try
         {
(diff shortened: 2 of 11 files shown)


src/Platform/Microsoft.Testing.Extensions.HangDump/HangDumpActivityIndicator.cs
```diff
@@ -83,11 +83,7 @@ public async Task OnTestSessionStartingAsync(ITestSessionContext testSessionCont
     {
         CancellationToken cancellationToken = testSessionContext.CancellationToken;
         ApplicationStateGuard.Ensure(_namedPipeClient is not null);
-
-        if (!await IsEnabledAsync().ConfigureAwait(false) || cancellationToken.IsCancellationRequested)
```
Evangelink, line 87: You have dropped the check for IsEnabledAsync, is this expected?

src/Platform/Microsoft.Testing.Extensions.TrxReport/TrxDataConsumer.cs
```diff
@@ -104,10 +104,7 @@ public TrxReportGenerator(
 
     public Task ConsumeAsync(IDataProducer dataProducer, IData value, CancellationToken cancellationToken)
     {
-        if (!_isEnabled || cancellationToken.IsCancellationRequested)
-        {
-            return Task.CompletedTask;
-        }
+        cancellationToken.ThrowIfCancellationRequested();
```
Evangelink, line 107: Check of is enabled was dropped, is this expected?

src/Platform/Microsoft.Testing.Extensions.TrxReport/TrxDataConsumer.cs
```diff
@@ -168,10 +165,7 @@ public Task ConsumeAsync(IDataProducer dataProducer, IData value, CancellationTo
     public async Task OnTestSessionStartingAsync(ITestSessionContext testSessionContext)
     {
         CancellationToken cancellationToken = testSessionContext.CancellationToken;
-        if (!_isEnabled || cancellationToken.IsCancellationRequested)
-        {
-            return;
-        }
+        cancellationToken.ThrowIfCancellationRequested();
```
Evangelink, line 168: Check of is enabled was dropped, is this expected?

src/Platform/Microsoft.Testing.Extensions.TrxReport/TrxDataConsumer.cs
```diff
@@ -210,10 +204,7 @@ await _trxTestApplicationLifecycleCallbacks.NamedPipeClient.RequestReplyAsync<Te
     public async Task OnTestSessionFinishingAsync(ITestSessionContext testSessionContext)
     {
         CancellationToken cancellationToken = testSessionContext.CancellationToken;
-        if (!_isEnabled || cancellationToken.IsCancellationRequested)
```
Evangelink, line 213: Check of is enabled was dropped, is this expected?

src/Platform/Microsoft.Testing.Extensions.HangDump/HangDumpActivityIndicator.cs
Youssef1313: Yes, it's always true. The extensions are not called in the first place if they are not enabled.

src/Platform/Microsoft.Testing.Extensions.TrxReport/TrxDataConsumer.cs
Youssef1313: Same as above

src/Platform/Microsoft.Testing.Extensions.TrxReport/TrxDataConsumer.cs
Youssef1313: Same as above

src/Platform/Microsoft.Testing.Extensions.TrxReport/TrxDataConsumer.cs
Youssef1313: Same as above

Evangelink (approved):

PR #7400, opened 2026-02-13
Fix test flakiness

### LINKED ISSUE #7398: Investigate `ServerCanBeStartedAndAborted_TcpIp` test flakiness
This test is often flaky:

src/Platform/Microsoft.Testing.Platform/ServerMode/JsonRpc/MessageHandlerFactory.cs#L54

src/Platform/Microsoft.Testing.Platform/ServerMode/JsonRpc/MessageHandlerFactory.cs(54,1): error : [ServerCanBeStartedAndAborted_TcpIp] Test method Microsoft.Testing.Platform.UnitTests.ServerTests.ServerCanBeStartedAndAborted_TcpIp threw exception:
System.InvalidOperationException: The operation is not allowed on non-connected sockets.

Evangelink:
Fix #7398

### DIFF
--- a/src/Platform/Microsoft.Testing.Platform/Hosts/ServerTestHost.cs
+++ b/src/Platform/Microsoft.Testing.Platform/Hosts/ServerTestHost.cs
@@ -150,10 +150,13 @@ internal sealed partial class ServerTestHost : CommonHost, IServerTestHost, IDis
             && (ex is OperationCanceledException
                 // the TcpClient/stream was disposed while an async operation was in flight.
                 or ObjectDisposedException
-                // ConnectAsync completed at the OS level at the same instant the cancellation callback closed the socket,
-                // leaving it in a non-connected state when GetStream() is called.
+                // TOCTOU race in the runtime: ConnectAsync completed successfully at the OS level,
+                // but the cancellation callback (registered via CancellationToken.UnsafeRegister
+                // in SocketAsyncEventArgs.ProcessIOCPResult) called CancelIoEx, tearing down
+                // the socket's connected state before GetStream() could read it.
                 or InvalidOperationException
-                // the socket was closed by the cancellation callback.
+                // the pending overlapped I/O was cancelled by CancelIoEx (Windows) from the
+                // cancellation callback, completing with SocketError.OperationAborted.
                 or SocketException { SocketErrorCode: SocketError.OperationAborted }))
         {
         }
--- a/src/Platform/Microsoft.Testing.Platform/ServerMode/JsonRpc/MessageHandlerFactory.cs
+++ b/src/Platform/Microsoft.Testing.Platform/ServerMode/JsonRpc/MessageHandlerFactory.cs
@@ -54,11 +54,20 @@ internal sealed partial class ServerModeManager
 #else
                 await client.ConnectAsync(host: _host, port: _port).WithCancellationAsync(cancellationToken, observeException: true).ConfigureAwait(false);
 #endif
-                // On NETCOREAPP, the ConnectAsync overload that takes a CancellationToken
-                // registers a callback that closes the socket. In that case, if connect
-                // completes at the OS level at the same instant the token fires,
-                // ConnectAsync can return successfully while the socket is already closed,
-                // causing GetStream() to throw InvalidOperationException ("non-connected sockets").
+                // On NETCOREAPP, the ConnectAsync overload accepting a CancellationToken
+                // delegates to SocketAsyncEventArgs, whose ProcessIOCPResult method
+                // (SocketAsyncEventArgs.Windows.cs) registers a cancellation callback via
+                // CancellationToken.UnsafeRegister. On Windows, the callback calls
+                // Interop.Kernel32.CancelIoEx to cancel the pending overlapped I/O;
+                // the completed I/O then surfaces as SocketError.OperationAborted.
+                //
+                // Because the registration happens *after* the OS call returns, there is
+                // a TOCTOU race: the connect can complete successfully at the OS level at
+                // the same instant the token fires. CancelIoEx finds no pending I/O to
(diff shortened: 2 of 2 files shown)


src/Platform/Microsoft.Testing.Platform/Hosts/ServerTestHost.cs
Youssef1313: Should this line move outside the try?

src/Platform/Microsoft.Testing.Platform/Hosts/ServerTestHost.cs
```diff
@@ -140,20 +140,26 @@ protected override async Task<int> InternalRunAsync(CancellationToken cancellati
...
-        catch (ObjectDisposedException) when (cancellationToken.IsCancellationRequested)
+        catch (Exception ex) when
+            // When the cancellation token fires during TCP connect or message handling, several
+            // exception types can surface depending on the exact timing:
+            (cancellationToken.IsCancellationRequested
+            // the standard cancellation path.
+            && (ex is OperationCanceledException
+                // the TcpClient/stream was disposed while an async operation was in flight.
+                or ObjectDisposedException
+                // ConnectAsync completed at the OS level at the same instant the cancellation callback closed the socket,
```
Youssef1313: I don't quite get how can ConnectAsync completes and after that the socket gets closed?

That sounds very strange.

src/Platform/Microsoft.Testing.Platform/ServerMode/JsonRpc/MessageHandlerFactory.cs
```diff
@@ -45,14 +45,33 @@ public async Task<IMessageHandler> CreateMessageHandlerAsync(CancellationToken c
...
-            await client.ConnectAsync(host: _host, port: _port).WithCancellationAsync(cancellationToken, observeException: true).ConfigureAwait(false);
+                await client.ConnectAsync(host: _host, port: _port).WithCancellationAsync(cancellationToken, observeException: true).ConfigureAwait(false);
 #endif
-            NetworkStream stream = client.GetStream();
-            return new TcpMessageHandler(client, clientToServerStream: stream, serverToClientStream: stream, FormatterUtilities.CreateFormatter());
+                // On NETCOREAPP, the ConnectAsync overload that takes a CancellationToken
+                // registers a callback that closes the socket. In that case, if connect
+                // completes at the OS level at the same instant the token fires,
+                // ConnectAsync can return successfully while the socket is already closed,
+                // causing GetStream() to throw InvalidOperationException ("non-connected sockets").
```
Youssef1313: Looking at the internal implementation in runtime, this doesn't sound correct. The usages of cancellation token are mostly `ThrowIfCancellationRequested` which means we get a typical OperationCanceledException.

I could be missing something, so I would like to have better idea here.

src/Platform/Microsoft.Testing.Platform/Hosts/ServerTestHost.cs
Evangelink: Good catch! The logging call doesn't need to be inside the cancellation catch filter. I'll move it before.

src/Platform/Microsoft.Testing.Platform/Hosts/ServerTestHost.cs
Evangelink: I investigated the .NET runtime implementation. The race is real and was observed in CI (#7398).

Here is how it happens:

1. `ConnectAsync(host, port, cancellationToken)` is initiated. The runtime registers a cancellation callback on the token.
2. The OS-level connect completes successfully. The completion callback runs on a thread pool thread and calls `SetToConnected()` which sets `_isConnected = true`.
3. The cancellation callback fires **concurrently on another thread** and calls `Socket.Dispose()` which calls `SetToDisconnected()` and sets `_isConnected = false`.
4. The awaiter resumes and `ConnectAsync` returns to the caller with no exception (because the SAEA reported `SocketError.Success`).
5. `GetStream()` checks `TcpClient.Connected` which reads `_isConnected` (now `false`) and throws `InvalidOperationException`.

This is a TOCTOU race between the connect completion and the cancellation callback running on different threads. The runtime itself acknowledges this kind of race in `Socket.Tasks.cs` (`ThrowException` method has a comment: "There is a race here anyway").

The `cancellationToken.ThrowIfCancellationRequested()` before `GetStream()` narrows the window but does not fully eliminate it, which is why the `InvalidOperationException` catch in `ServerTestHost` serves as the safety net.

src/Platform/Microsoft.Testing.Platform/ServerMode/JsonRpc/MessageHandlerFactory.cs
Evangelink: You are right that in the standard cancellation path, `ConnectAsync` throws `OperationCanceledException` directly (the runtime converts `SocketError.OperationAborted` via `ThrowIfCancellationRequested()`). That is the common case.

However, the race described in the comment does occur: issue #7398 shows exactly this `InvalidOperationException` from `GetStream()` in CI. The mechanism is that the cancellation callback calls `Socket.Dispose()` -> `SetToDisconnected()` concurrently with `ConnectAsync` completing successfully, causing `TcpClient.Connected` to return `false` when `GetStream()` checks it.

The `ThrowIfCancellationRequested()` here is a belt-and-suspenders measure that catches the majority of cases before hitting `GetStream()`. The remaining edge case is handled by the `InvalidOperationException` catch in `ServerTestHost`.

src/Platform/Microsoft.Testing.Platform/Hosts/ServerTestHost.cs
Youssef1313: > The runtime registers a cancellation callback on the token.

Where does that happen? I was looking into the implementation and I'm unable to find this part.

PR #7405, opened 2026-02-14
Fix FP with proper assert analyzer and IComparable

### LINKED ISSUE #7343: Possible false positive for proper assert method analyzer
When detecting `Assert.IsTrue(expr > expr2)` and suggest using `Assert.IsGreaterThan` and similar APIs, we should probably verify that the types implement `IComparable` before suggesting the fix. We also need to add test for nullable value types (e.g, `TimeSpan?`)

Evangelink:
Fix #7343

### DIFF
--- a/src/Analyzers/MSTest.Analyzers/UseProperAssertMethodsAnalyzer.cs
+++ b/src/Analyzers/MSTest.Analyzers/UseProperAssertMethodsAnalyzer.cs
@@ -260,12 +260,13 @@ public sealed class UseProperAssertMethodsAnalyzer : DiagnosticAnalyzer
 
             INamedTypeSymbol objectTypeSymbol = context.Compilation.GetSpecialType(SpecialType.System_Object);
             context.Compilation.TryGetOrCreateTypeByMetadataName(WellKnownTypeNames.SystemLinqEnumerable, out INamedTypeSymbol? enumerableTypeSymbol);
+            INamedTypeSymbol? iComparableOfTSymbol = context.Compilation.GetTypeByMetadataName("System.IComparable`1");
 
-            context.RegisterOperationAction(context => AnalyzeInvocationOperation(context, assertTypeSymbol, objectTypeSymbol, enumerableTypeSymbol), OperationKind.Invocation);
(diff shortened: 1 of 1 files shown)


src/Analyzers/MSTest.Analyzers/UseProperAssertMethodsAnalyzer.cs
```diff
@@ -568,6 +574,28 @@ private static ComparisonCheckStatus RecognizeComparisonCheck(
...
+        => ImplementsIComparableOfSelf(compilation, candidateType)
+            && compilation.ClassifyCommonConversion(leftType, candidateType) is { Exists: true, IsImplicit: true }
+            && compilation.ClassifyCommonConversion(rightType, candidateType) is { Exists: true, IsImplicit: true };
+
+    private static bool ImplementsIComparableOfSelf(Compilation compilation, ITypeSymbol type)
+    {
+        if (compilation.GetTypeByMetadataName("System.IComparable`1") is not INamedTypeSymbol iComparableOfT)
+        {
+            return false;
+        }
```
Youssef1313: This should be retrieved in compilation start and then get passed down.

src/Analyzers/MSTest.Analyzers/UseProperAssertMethodsAnalyzer.cs
```diff
@@ -553,6 +554,11 @@ private static ComparisonCheckStatus RecognizeComparisonCheck(
                 return ComparisonCheckStatus.Unknown;
             }
 
+            if (!CanUseTypesWithComparableAsserts(compilation, binaryOperation.LeftOperand.Type, binaryOperation.RightOperand.Type))
```
Youssef1313: We probably need to check that the type of left operand is the same as the type of right operand, and that the type is `IComparable<T>` where `T` is the type in question. We don't need  `ClassifyCommonConversion`, most likely.

src/Analyzers/MSTest.Analyzers/UseProperAssertMethodsAnalyzer.cs
```diff
@@ -553,6 +554,11 @@ private static ComparisonCheckStatus RecognizeComparisonCheck(
...
+            {
+                return ComparisonCheckStatus.Unknown;
```
Youssef1313: Also the logic above that special cases Enum might no longer be needed as the new code path will likely handle enums correctly.

PR #7407, opened 2026-02-14
Make tracing env variable case insensitive

### LINKED ISSUE #7385: make tracing env variable case insensitive
## Describe the bug

'TESTINGPLATFORM_DIAGNOSTIC_VERBOSITY', was expecting 'Trace', 'Debug', 'Information', 'Warning', 'Error', or 'Critical' but got 'trace'.

## Version used

1.8.x might be fixed already.

Evangelink:
Fix #7385

### DIFF
--- a/test/UnitTests/Microsoft.Testing.Platform.UnitTests/TestApplicationDiagnosticVerbosityTests.cs
+++ b/test/UnitTests/Microsoft.Testing.Platform.UnitTests/TestApplicationDiagnosticVerbosityTests.cs
@@ -6,7 +6,6 @@ using Microsoft.Testing.Platform.Logging;
 namespace Microsoft.Testing.Platform.UnitTests;
 
 [TestClass]
-[UnsupportedOSPlatform("browser")]
 public sealed class TestApplicationDiagnosticVerbosityTests
 {
     [TestMethod]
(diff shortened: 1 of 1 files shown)


test/UnitTests/Microsoft.Testing.Platform.UnitTests/TestApplicationDiagnosticVerbosityTests.cs
```diff
@@ -0,0 +1,29 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using Microsoft.Testing.Platform.Logging;
+
+namespace Microsoft.Testing.Platform.UnitTests;
+
+[TestClass]
+[UnsupportedOSPlatform("browser")]
```
Youssef1313: The whole test project doesn't support browser anyways. Do we somehow get a warning if we don't add this attribute here??

test/UnitTests/Microsoft.Testing.Platform.UnitTests/TestApplicationDiagnosticVerbosityTests.cs
```diff
@@ -0,0 +1,29 @@
...
+public sealed class TestApplicationDiagnosticVerbosityTests
+{
+    [TestMethod]
+    public void TryParseDiagnosticVerbosity_WhenLowercaseValue_ReturnsParsedValue()
+    {
+        bool hasValue = TestApplication.TryParseDiagnosticVerbosity("trace", out LogLevel parsedLogLevel);
```
Youssef1313, line 14: Consider adding a test for an invalid value that would throw NotSupportedException.

Issue #3491, opened 2024-08-07
Progress remains on screen when output is written from test


nohwnd:
## Describe the bug

When the initial implementation of https://github.com/microsoft/testfx/pull/3292 is merged, when there is test output, it can cause the progress bar remain on the screen, because we did not delete it when writing output. 

This can also happen when test controller writes to console, because we need to write "delete progress" before any line is output.

In process we can "block" user console writes to synchronize this. But with 2 processes we could only move to the "parent" process, which is not following the current architecture.

In this output the progress should not remain on screen, but we are seeing output from 2 different processes, the testhost writing the progress, and the testhost controller writing the Slowest 10 tests.

![image](https://github.com/user-attachments/assets/e006c5f9-b8f6-44a0-aaa3-cf38205d3313)

When the Slowest 10 tests text reaches the screen, we do not issue ansi codes for "clear progress" (go up one line and delete line), because we cannot be aware of it (it comes from a different process). Console writes it, and then we emit clear progress, because the process still thinks the progress bar is the last line, which deletes 1 line of the testhost controller output (there are only 9 lines for slowest tests visible).

Unless we block all the output, or fully sync the output from the 2 processes we cannot avoid this. This is a problem common to other implementations as well, e.g. here the modern progress bar in Powershell, where we can see the same problem by writing to the console from a different thread while progress is reported: 

![image](https://github.com/user-attachments/assets/b712a188-b172-40d1-8d67-c648a0323d00)

```powershell
 start-threadjob { 1..100 | % { start-sleep -Milliseconds 30 ; [console]::writeLine("aaargh $_")}}; ForEach ($Item in $Collection) {
    Write-Progress -PercentComplete ($Item/100*100) -Status "Processing Items" -Activity "Item $item of 100"
    Start-Sleep -Milliseconds 50
}
```

We can mitigate this problem in process by wrapping the standard Console.Out with a TextWriter that calls the clear progress callback before user outputs to screen. 

## Steps To Reproduce

Have a test or test controller that writes to screen while there is progress bar.

nohwnd:
I tried few approaches to remedy this, without locking output: 

- Move cursor to start of line when I am done rendering the progress, this allows user to overwrite the progress, and so we don't delete their progress, but it leaves a mess on the screen, especially the timestamp that is rendered all the way to the right, because emitting new-line will not automatically clear the rest of the line.

[...]

PR #7408, opened 2026-02-14
Add tests for terminal concurrency issue

Evangelink:
Relates to #3491

### DIFF
--- a/test/IntegrationTests/Microsoft.Testing.Platform.Acceptance.IntegrationTests/ConsoleTests.cs
+++ b/test/IntegrationTests/Microsoft.Testing.Platform.Acceptance.IntegrationTests/ConsoleTests.cs
@@ -67,7 +67,7 @@ public class ConsoleTests : AcceptanceTestBase<ConsoleTests.TestAssetFixture>
             };
         }
 
-        TestHostResult testHostResult = await testHost.ExecuteAsync("--ignore-exit-code 8", environmentVariables);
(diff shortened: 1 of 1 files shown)


test/IntegrationTests/Microsoft.Testing.Platform.Acceptance.IntegrationTests/ConsoleTests.cs
```diff
@@ -31,6 +31,30 @@ This currently doesn't pass because SystemConsole is writing to StreamWriter dir
     public async Task StartUpdateShouldNotBeInterruptedConsoleIsNotReplaced(string tfm)
         => await ConsoleTestsCoreAsync(tfm, null);
 
+    [DynamicData(nameof(TargetFrameworks.AllForDynamicData), typeof(TargetFrameworks))]
+    [TestMethod]
+    [Ignore("""
```
Youssef1313, line 36: We are currently having two skipped tests in this class related to synchronization. I haven't looked into the new test yet, but just wanting to make sure it's not testing what is already skipped.

Evangelink: I have checked: this new ignored test documents cross-process output interleaving (issue #3491/#4425), while the two existing ignored tests are SystemConsole lock/synchronization cases around StartUpdate/StopUpdate. They cover different failure modes; the new unit test asserts the in-process erase/write/render ordering

PR #7410, opened 2026-02-14
MTP: ensure post-session display runs after cancellation

### LINKED ISSUE #6099: Suspicious cancellation logic in MTP
https://github.com/microsoft/testfx/blob/0962dc3b939033d6a12243ccb796dd44a03710c1/src/Platform/Microsoft.Testing.Platform/Hosts/CommonTestHost.cs#L133-L154

We always call `DisplayAfterSessionEndRunAsync` at the end, regardless of whether or not we are canceled. But then we pass `testSessionCancellationToken`. That feels suspicious to me.

The comment says:

```
        // We keep the display after session out of the OperationCanceledException catch because we want to notify the IPlatformOutputDevice
        // also in case of cancellation. Most likely it needs to notify users that the session was canceled.
```

But in that case shouldn't we pass `CancellationToken.None` to explicitly state that we want the work to be done even in case of cancellation?

Evangelink:
**Summary**
Fixes an inconsistency in cancellation flow for MTP test sessions.

[CommonHost.ExecuteRequestAsync](vscode-file://vscode-app/c:/Users/amauryleve/AppData/Local/Programs/Microsoft%20VS%20Code%20Insiders/16e49a8b88/resources/app/out/vs/code/electron-browser/workbench/workbench.html) intentionally calls post-session display outside the cancellation catch, but [DisplayAfterSessionEndRunAsync](vscode-file://vscode-app/c:/Users/amauryleve/AppData/Local/Programs/Microsoft%20VS%20Code%20Insiders/16e49a8b88/resources/app/out/vs/code/electron-browser/workbench/workbench.html) was still receiving the canceled session token.

This change passes [CancellationToken.None](vscode-file://vscode-app/c:/Users/amauryleve/AppData/Local/Programs/Microsoft%20VS%20Code%20Insiders/16e49a8b88/resources/app/out/vs/code/electron-browser/workbench/workbench.html) to ensure final output/session-finishing notifications are not skipped when cancellation is requested.

**Changes**
Updated [CommonTestHost.cs:195](vscode-file://vscode-app/c:/Users/amauryleve/AppData/Local/Programs/Microsoft%20VS%20Code%20Insiders/16e49a8b88/resources/app/out/vs/code/electron-browser/workbench/workbench.html) to call post-session display with a non-cancelable token.

Added regression test [CommonHostTests.cs](vscode-file://vscode-app/c:/Users/amauryleve/AppData/Local/Programs/Microsoft%20VS%20Code%20Insiders/16e49a8b88/resources/app/out/vs/code/electron-browser/workbench/workbench.html) verifying canceled sessions still invoke post-session display with a non-cancelable token.

**Validation**
Ran: dotnet test [Microsoft.Testing.Platform.UnitTests.csproj](http://_vscodecontentref_/5) --filter "FullyQualifiedName~CommonHostTests.ExecuteRequestAsync_WhenSessionIsCancelled_UsesCancellationTokenNoneForDisplayAfterSessionEndRun"

Result: pass.

**Issue**
Closes #6099

### DIFF
--- a/src/Platform/Microsoft.Testing.Platform/Hosts/CommonTestHost.cs
+++ b/src/Platform/Microsoft.Testing.Platform/Hosts/CommonTestHost.cs
@@ -194,7 +194,8 @@ internal abstract class CommonHost(ServiceProvider serviceProvider) : IHost
 
     private static async Task DisplayAfterSessionEndRunAsync(ProxyOutputDevice outputDevice, ITestSessionContext sessionInfo)
     {
-        // Display after session end
+        // Display after session end even when the session cancellation token is canceled.
+        // We intentionally pass a non-cancelable token so final output/cleanup notifications are not skipped.
         await outputDevice.DisplayAfterSessionEndRunAsync(CancellationToken.None).ConfigureAwait(false);
 
         // We want to ensure that the output service is the last one to run


src/Platform/Microsoft.Testing.Platform/Hosts/CommonTestHost.cs
```diff
@@ -195,7 +195,7 @@ private static async Task DisplayBeforeSessionStartAsync(ProxyOutputDevice outpu
     private static async Task DisplayAfterSessionEndRunAsync(ProxyOutputDevice outputDevice, ITestSessionContext sessionInfo)
     {
         // Display after session end
-        await outputDevice.DisplayAfterSessionEndRunAsync(sessionInfo.CancellationToken).ConfigureAwait(false);
+        await outputDevice.DisplayAfterSessionEndRunAsync(CancellationToken.None).ConfigureAwait(false);
```
Youssef1313, line 199: Let's add a comment explaining why we pass None.

Youssef1313 (approved):
LGTM 🎉

I only reviewed the product change. Please ensure that the added test would have been failing on current main.

PR #7412, opened 2026-02-14
[MTP AzDO] Display TFM and test name

### LINKED ISSUE #5547: Add additional (optional?) info to azdo report plugin
## Summary

Seeing the failure like this is useful, but in Checks view, we don't know what method it is coming from. And in checks and PR view, we don't know what tfm it is coming from, e.g. if there is .net framework specific problem, it would be useful to know what are the affected tfms.

Image

Evangelink:
Fixes #5547

### DIFF
--- a/test/UnitTests/Microsoft.Testing.Extensions.UnitTests/AzureDevOpsTests.cs
+++ b/test/UnitTests/Microsoft.Testing.Extensions.UnitTests/AzureDevOpsTests.cs
@@ -45,7 +45,7 @@ public sealed class AzureDevOpsTests
         // Trim ##. If we keep it, then when the test fails, the assertion failure will get printed to screen and picked up incorrectly by AzDO, because it scans all output for the ##vso... pattern
         var logger = new TextLogger();
         string? text = AzureDevOpsReporter.GetErrorText("MyTestDisplayName", "Some custom reason\nwith\rnewline", error, "severity", new SystemFileSystem(), logger, "net9.0")?.TrimStart('#');
-        Assert.AreEqual("vso[task.logissue type=severity;sourcepath=test/UnitTests/Microsoft.Testing.Extensions.UnitTests/AzureDevOpsTests.cs;linenumber=19;columnnumber=1][MyTestDisplayName] [net9.0] Some custom reason%0Awith%0Dnewline", text, $"\nLogs:\n{string.Join("\n", logger.Logs)}");
+        Assert.AreEqual("vso[task.logissue type=severity;sourcepath=test/UnitTests/Microsoft.Testing.Extensions.UnitTests/AzureDevOpsTests.cs;linenumber=38;columnnumber=1][MyTestDisplayName] [net9.0] Some custom reason%0Awith%0Dnewline", text, $"\nLogs:\n{string.Join("\n", logger.Logs)}");
     }
 
     private class TextLogger : ILogger


test/UnitTests/Microsoft.Testing.Extensions.UnitTests/AzureDevOpsTests.cs
```diff
@@ -64,6 +64,32 @@ public void ReportsWithoutDisplayNameWhenNull()
 #endif
     }
 
+    [TestMethod]
+    public void ReportsWithTargetFrameworkWhenDisplayNameIsNull()
+    {
+#if NETFRAMEWORK && !NET472_OR_GREATER
```
Evangelink: Alternatively, we can change the settings of the test project.

src/Platform/Microsoft.Testing.Extensions.AzureDevOpsReport/AzureDevOpsReporter.cs
```diff
@@ -308,6 +308,43 @@ private async Task WriteExceptionAsync(string testDisplayName, string? explanati
...
+    private static string FormatMessage(string? testDisplayName, string? targetFrameworkMoniker, string message)
+        => (RoslynString.IsNullOrEmpty(targetFrameworkMoniker), RoslynString.IsNullOrEmpty(testDisplayName)) switch
+        {
+            (true, true) => message,
+            (true, false) => $"[{testDisplayName}] {message}",
+            (false, true) => $"[{targetFrameworkMoniker}] {message}",
+            _ => $"[{targetFrameworkMoniker}] [{testDisplayName}] {message}",
+        };
+
+    private static string? TryGetTargetFrameworkMoniker()
```
nohwnd: There is more nuance to this, would be nice to re-use this code for consistency: https://github.com/microsoft/testfx/blob/main/src/Platform/Microsoft.Testing.Platform/OutputDevice/TargetFrameworkParser.cs

src/Platform/Microsoft.Testing.Extensions.AzureDevOpsReport/AzureDevOpsReporter.cs
```diff
@@ -308,6 +308,43 @@ private async Task WriteExceptionAsync(string testDisplayName, string? explanati
         return null;
     }
 
+    private static string FormatMessage(string? testDisplayName, string? targetFrameworkMoniker, string message)
+        => (RoslynString.IsNullOrEmpty(targetFrameworkMoniker), RoslynString.IsNullOrEmpty(testDisplayName)) switch
+        {
+            (true, true) => message,
+            (true, false) => $"[{testDisplayName}] {message}",
+            (false, true) => $"[{targetFrameworkMoniker}] {message}",
```
nohwnd: Personally I would position the test name first and the additional info like tfm second, maybe later we will also want to include architecture, and then it gets harder to see what is going on. 

the first one seems less scary, and more organized:

```
aaaa net9.0
aaaa net10.0
aaaa net11.0
bbb net9.0
bbb net10.0
```
```
net9.0 aaaa
net10.0 aaaa
net11.0 aaaa
net9.0 bbb
net10.0 bbb
```

src/Platform/Microsoft.Testing.Extensions.AzureDevOpsReport/AzureDevOpsReporter.cs
```diff
@@ -308,6 +308,19 @@ private async Task WriteExceptionAsync(string testDisplayName, string? explanati
...
+        => (RoslynString.IsNullOrEmpty(testDisplayName), RoslynString.IsNullOrEmpty(targetFrameworkMoniker)) switch
+        {
+            (true, true) => message,
+            (true, false) => $"[{targetFrameworkMoniker}] {message}",
+            (false, true) => $"[{testDisplayName}] {message}",
+            _ => $"[{testDisplayName}] [{targetFrameworkMoniker}] {message}",
+        };
+
+    private static string? TryGetTargetFrameworkMoniker()
+        => TargetFrameworkParser.GetShortTargetFramework(Assembly.GetEntryAssembly()?.GetCustomAttribute<TargetFrameworkAttribute>()?.FrameworkDisplayName)
```
Youssef1313, line 311: Should we mark `TargetFrameworkParser` with `[Embedded]` and source-link it to the needed project(s) to avoid future issues with IVT? This can be a follow-up PR, doesn't need to happen now.

src/Platform/Microsoft.Testing.Extensions.AzureDevOpsReport/AzureDevOpsReporter.cs
```diff
@@ -308,6 +308,19 @@ private async Task WriteExceptionAsync(string testDisplayName, string? explanati
...
+    private static string FormatMessage(string? testDisplayName, string? targetFrameworkMoniker, string message)
+        => (RoslynString.IsNullOrEmpty(testDisplayName), RoslynString.IsNullOrEmpty(targetFrameworkMoniker)) switch
+        {
+            (true, true) => message,
+            (true, false) => $"[{targetFrameworkMoniker}] {message}",
+            (false, true) => $"[{testDisplayName}] {message}",
+            _ => $"[{testDisplayName}] [{targetFrameworkMoniker}] {message}",
+        };
+
+    private static string? TryGetTargetFrameworkMoniker()
```
Youssef1313: I think this cannot be null.

We are likely missing a `[return: NotNullIfNotNull(nameof(frameworkDescription))]` attribute on `GetShortTargetFramework`

src/Platform/Microsoft.Testing.Extensions.AzureDevOpsReport/AzureDevOpsReporter.cs
```diff
@@ -163,7 +165,7 @@ private async Task WriteExceptionAsync(string testDisplayName, string? explanati
         await _outputDisplay.DisplayAsync(this, new FormattedTextOutputDeviceData(line), cancellationToken).ConfigureAwait(false);
     }
 
-    internal static /* for testing */ string? GetErrorText(string? testDisplayName, string? explanation, Exception? exception, string severity, IFileSystem fileSystem, ILogger logger)
+    internal static /* for testing */ string? GetErrorText(string? testDisplayName, string? explanation, Exception? exception, string severity, IFileSystem fileSystem, ILogger logger, string? targetFrameworkMoniker = null)
```
Youssef1313: If `targetFrameworkMoniker` cannot be null (see my other comment), I think we should make this parameter mandatory as well and always pass something in tests.

Youssef1313: Can `testDisplayName` be null?

src/Platform/Microsoft.Testing.Extensions.AzureDevOpsReport/AzureDevOpsReporter.cs
```diff
@@ -308,6 +308,19 @@ private async Task WriteExceptionAsync(string testDisplayName, string? explanati
         return null;
     }
 
+    private static string FormatMessage(string? testDisplayName, string? targetFrameworkMoniker, string message)
+        => (RoslynString.IsNullOrEmpty(testDisplayName), RoslynString.IsNullOrEmpty(targetFrameworkMoniker)) switch
+        {
+            (true, true) => message,
```
Youssef1313: Should we show `<unknown-test>` or something similar? I imagine this won't happen in practice anyways.

src/Platform/Microsoft.Testing.Extensions.AzureDevOpsReport/AzureDevOpsReporter.cs
```diff
@@ -308,6 +308,19 @@ private async Task WriteExceptionAsync(string testDisplayName, string? explanati
...
+            (true, false) => $"[{targetFrameworkMoniker}] {message}",
```
Youssef1313: Same here about showing some sort of "unknown test" marker.

src/Platform/Microsoft.Testing.Platform/Helpers/TargetFrameworkParser.cs
```diff
@@ -1,7 +1,7 @@
 ﻿// Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
-namespace Microsoft.Testing.Platform.OutputDevice;
+namespace Microsoft.Testing.Platform.Helpers;
```
Youssef1313: This breaks binary compatibility with Microsoft.Testing.Platform.MSBuild

Youssef1313: Let's fix this one separately. First by making this `Embedded`, then after few releases we rename to reduce the likelihood of this breaking someone.

In most cases where I see version mismatch between MTP and its extensions, it's usually only one minor version difference because other deps still didn't follow.

src/Platform/Microsoft.Testing.Extensions.AzureDevOpsReport/AzureDevOpsReporter.cs
Evangelink: Will do a global follow-up for IVT apis

src/Platform/Microsoft.Testing.Extensions.AzureDevOpsReport/AzureDevOpsReporter.cs
```diff
@@ -163,7 +165,7 @@ private async Task WriteExceptionAsync(string testDisplayName, string? explanati
         await _outputDisplay.DisplayAsync(this, new FormattedTextOutputDeviceData(line), cancellationToken).ConfigureAwait(false);
     }
 
-    internal static /* for testing */ string? GetErrorText(string? testDisplayName, string? explanation, Exception? exception, string severity, IFileSystem fileSystem, ILogger logger)
+    internal static /* for testing */ string? GetErrorText(string? testDisplayName, string? explanation, Exception? exception, string severity, IFileSystem fileSystem, ILogger logger, string targetFrameworkMoniker)
```
Youssef1313: Why `testDisplayName` is nullable?

PR #7413, opened 2026-02-14
Allow parallelize attribute on class

### LINKED ISSUE #5555: Opt-in for parallel-execution on a class level, instead of current assembly level
### Discussed in https://github.com/microsoft/testfx/discussions/5553

<div type='discussions-op-text'>

<sup>Originally posted by **thomasstormark3shape** May  2, 2025</sup>
Both Xunit and Nunit has the option to opt-in for parallel test execution on a class-level or method level.

With MSTest its only possible to set [assembly: Parallelize(…)], and then opting out for each class or method by [DoNotParallelize].

When you have an old large test project with houndreds of test classes, and you find out that a lot does NOT parallelize - it would be good to gradually opt-in during migration period, and finally go for assembly level default parallelize.

Any thoughts on this feature?</div>

Evangelink:
I need a little more reviews and manual tests

Fixes #5555


Evangelink:
I agree overall but let's chat about it next week. Putting aside the refactoring on the execution manager, this is mainly just adding some extra properties to the test method object.

Evangelink:
Closing the PR, it was good to make investigations and see limitations and potential conficts.

PR #7423, opened 2026-02-17
Cleanup MTP async data consumer

Youssef1313:
This refactors the logic to avoid the arbitrary Task.Delay calls in DrainDataAsync, and cleans up the logic.


Evangelink:
Pushed an updated commit that merges `main` and addresses both review findings:

1. **Race condition between channel completion and recreation**
   The processor no longer completes and recreates its channel during drain. Instead, `DrainDataAsync` enqueues a sentinel `AsyncConsumerDataProcessorMessage` (a drain marker carrying a `TaskCompletionSource<bool>`) and awaits it. Because the consumer is single-reader and the channel is FIFO, the marker fires only after every item that was already in the channel has been processed. The channel itself is never closed during a drain, so concurrent `PublishAsync` calls cannot hit `ChannelClosedException`/`InvalidOperationException`.

2. **Lost publisher/consumer loop detection (and the now-failing `DrainDataAsync_Loop_ShouldFail` test)**
   Loop detection moved up to `AsynchronousMessageBus.DrainDataAsync`. Each processor now returns a `bool` from `DrainDataAsync` indicating whether at least one payload was processed (tracked via a `_processedCount` counter snapshotted around the marker). The bus drains the **distinct** processors (`_consumerProcessor.Values` rather than `_dataTypeConsumers.Values`, so consumers subscribed to multiple data types are drained only once per round) in a loop, bailing out as soon as a round processes nothing. If we still keep processing data after `MaxDrainAttempts = 5` rounds, the bus throws `InvalidOperationException` with the existing `"Publisher/Consumer loop detected during the drain after ..."` message, which is what `DrainDataAsync_Loop_ShouldFail` asserts on. No more arbitrary `Task.Delay`s.

Side effects:
- `IAsyncConsumerDataProcessor.DrainDataAsync()` now returns `Task<bool>`.
- The two `AsyncConsumerDataProcessor` implementations no longer mutate their channel/consume-task fields (both are `readonly` again).
- Merge conflict from PR #8201 (TestHostBuilder split) resolved by applying the `IEnvironment` parameter removal in `TestHostBuilder.Framework.cs`.

All 2223 `Microsoft.Testing.Platform.UnitTests` pass locally on net8.0, net9.0, and net48.

Evangelink:
Pushed `25faf891` to address the second pass of Copilot review findings:

- **Counter incrementing before `ConsumeAsync` completes (net.cs:63, netstandard.cs:53)**
  Replaced the per-processor `_processedCount` snapshotted around the marker with a `ReceivedCount` counter incremented in `PublishAsync`. The bus now compares each processor's `ReceivedCount` between rounds, so any payload enqueued by an in-flight consumer (including ones that race with the drain marker) triggers another round — matching the loop-detection behavior of the old polling implementation.

- **OperationCanceledException on drain marker write (net.cs:99)**
  `DrainDataAsync` now catches `OperationCanceledException` (with the same `CancellationToken` guard as the consume loop) and returns gracefully, matching the previous "bail out on cancellation" semantic so shutdown/cleanup paths don't see unexpected aborts on NETCOREAPP.

- **Hard-coded drain limit (AsynchronousMessageBus.cs:16)**
  Restored the `TESTINGPLATFORM_MESSAGEBUS_DRAINDATA_ATTEMPTS` environment variable as an escape hatch. The constructor takes `IEnvironment` again; the default stays at 5 but can be raised for longer (still finite) producer/consumer chains. The threading of `IEnvironment` is restored in `TestHostBuilder.Framework.cs`, `TestHostControllersTestHost.cs`, and the test constructor calls.

Other notes:
- `IAsyncConsumerDataProcessor.DrainDataAsync()` returns `Task` again now that the bus drives loop detection.
- I also had to regenerate the `FrameworkMessages.*.xlf` files for `AreEquivalent` strings dropped by an earlier OneLocBuild check-in on `main`; without that fix every PR's CI build fails before our changes can compile (that fix is the prior commit, `26585e4b`).

Local validation:
- `dotnet test` for `AsynchronousMessageBusTests` is green on net8.0, net9.0, and net48 in both Debug and Release.
- The `DrainDataAsync_Loop_ShouldFail` test passed 5 consecutive Release runs.

Evangelink:
CI is now in good shape after the latest push:

| Job | Result |
| --- | --- |
| Build Linux Debug | ⚠️ Failure on pre-existing flaky `VSTest_DiscoverTests_Succeeds` / `VSTest_RunTests_Succeeds` integration tests (also fail on PR #8288, unrelated to this PR's MTP message-bus changes) |
| Build Linux Release | ✅ |
| Build MacOS Debug | ✅ |
| Build MacOS Release | ✅ |
| Build Windows Debug | ✅ |
| Build Windows Release | ✅ |
| Build WindowsSamples Debug | ✅ |
| Build WindowsSamples Release | ✅ |

The PR is `MERGEABLE`, all three (then later four) Copilot review findings have been addressed, and the dedicated `AsynchronousMessageBusTests` (including `DrainDataAsync_Loop_ShouldFail`) pass on net8.0/net9.0/net48 in both Debug and Release locally and in CI.

PR #7426, opened 2026-02-18
Improve Assert.AreSame message for nulls

### LINKED ISSUE #7421: Improve AreSame error messages in case of null values
## Summary
Currently the AreSame overload methods produce an AssertFailedException when either expected or actual are null and the other one isn't, but these messages do not mention anything about null values.

## Example
```c#
string? expected = "test";
string? actual = null;
Assert.AreSame(expected, actual);
```
### Actual result:
`Assert.AreSame failed. 'expected' expression: 'expected', 'actual' expression: 'actual'.`

### Proposed result:
`Assert.AreSame failed. 'expected' expression: 'expected', 'actual' expression: 'actual'. 'actual' was null.`

Evangelink:
Fix #7421

### DIFF
--- a/src/TestFramework/TestFramework/Assertions/Assert.AreSame.cs
+++ b/src/TestFramework/TestFramework/Assertions/Assert.AreSame.cs
@@ -80,9 +80,13 @@ public sealed partial class Assert
     public readonly struct AssertAreNotSameInterpolatedStringHandler<TArgument>
     {
         private readonly StringBuilder? _builder;
+        private readonly TArgument? _notExpected;
+        private readonly TArgument? _actual;
 
         public AssertAreNotSameInterpolatedStringHandler(int literalLength, int formattedCount, TArgument? notExpected, TArgument? actual, out bool shouldAppend)
         {
+            _notExpected = notExpected;
+            _actual = actual;
             shouldAppend = IsAreNotSameFailing(notExpected, actual);
             if (shouldAppend)
             {
@@ -95,7 +99,7 @@ public sealed partial class Assert
             if (_builder is not null)
             {
                 _builder.Insert(0, string.Format(CultureInfo.CurrentCulture, FrameworkMessages.CallerArgumentExpressionTwoParametersMessage, "notExpected", notExpectedExpression, "actual", actualExpression) + " ");
-                ThrowAssertAreNotSameFailed(_builder.ToString());
+                ThrowAssertAreNotSameFailed(_notExpected, _actual, _builder.ToString());
             }
         }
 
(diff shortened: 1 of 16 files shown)


test/UnitTests/TestFramework.UnitTests/Assertions/AssertTests.AreSame.cs
```diff
@@ -68,6 +68,48 @@ public void AreSame_InterpolatedString_BothAreValueTypes_ShouldFailWithSpecializ
         action.Should().Throw<Exception>().WithMessage("Assert.AreSame failed. Do not pass value types to AreSame(). Values converted to Object will never be the same. Consider using AreEqual(). 'expected' expression: '1', 'actual' expression: '1'. User-provided message System.Object");
     }
 
+    public void AreSame_ExpectedNull_ShouldFailWithNullMessage()
+    {
+        object? expected = null;
+        Action action = () => Assert.AreSame(expected, new object());
+        action.Should().Throw<Exception>().WithMessage("Assert.AreSame failed. Expected is <null>. 'expected' expression: 'expected', 'actual' expression: 'new object()'.");
```
Evangelink, line 75: We could also consider mixing the null part with the expressions, like:

`Assert.AreSame failed. 'expected' expression: 'expected' is <null>, 'actual' expression: 'new object()'."`

src/TestFramework/TestFramework/Assertions/Assert.AreSame.cs
```diff
@@ -187,14 +191,22 @@ private static bool IsAreSameFailing<T>(T? expected, T? actual)
     [DoesNotReturn]
     private static void ThrowAssertAreSameFailed<T>(T? expected, T? actual, string userMessage)
     {
-        string finalMessage = userMessage;
-        if (expected is ValueType && actual is ValueType)
-        {
-            finalMessage = string.Format(
+        string finalMessage = expected is null
```
Youssef1313, line 194: nit: Nesting ternary operators like this is way less readable IMO. I know this is to satisfy an analyzer, but I don't necessarily agree with the styling in case of nesting.

src/TestFramework/TestFramework/Assertions/Assert.AreSame.cs
```diff
@@ -240,14 +252,23 @@ public static void AreNotSame<T>(T? notExpected, T? actual, string? message = ""
...
         }
     }
 
     private static bool IsAreNotSameFailing<T>(T? notExpected, T? actual)
         => object.ReferenceEquals(notExpected, actual);
 
     [DoesNotReturn]
-    private static void ThrowAssertAreNotSameFailed(string userMessage)
-        => ThrowAssertFailed("Assert.AreNotSame", userMessage);
+    private static void ThrowAssertAreNotSameFailed<T>(T? notExpected, T? actual, string userMessage)
```
Youssef1313, line 263: I don't get why the message customization is partially done here and partially on the caller? Can we put all the logic together in one place?

Youssef1313: I actually misread the change. Resolving.

src/TestFramework/TestFramework/Assertions/Assert.AreSame.cs
Evangelink: Let's disable the analyzer in a follow-up PR

PR #7437, opened 2026-02-20
Add default settings for vscode

nohwnd:
Disable using workspace generated solution in VSCode, and prefer our own existing solution, to shorten restore, and avoid including sample projects that are not included.

Uses testfx slnx by default to avoid always waiting and having to choose one.

Workaround for https://github.com/microsoft/vscode-dotnettools/issues/2805

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,0 +1,5 @@
+{
+    "dotnet.automaticallyCreateSolutionInWorkspace": false,
+    "dotnet.enableWorkspaceBasedDevelopment": false,
+    "dotnet.solution.autoOpen": "TestFx.slnx",
+}


Evangelink (approved):

.vscode/settings.json
```diff
@@ -0,0 +1,5 @@
+{
+    "dotnet.automaticallyCreateSolutionInWorkspace": false,
+    "dotnet.enableWorkspaceBasedDevelopment": false,
+    "dotnet.solution.autoOpen": "TestFx.slnx",
```
nohwnd, line 4: in theory just this option is enough, but if you remove it in the future you will again get the workspace based development and behavior that mostly works, but not always.

PR #7443, opened 2026-02-20
Reduce IVT by using Embedded attribute and source link

Evangelink:


src/Platform/Microsoft.Testing.Platform.MSBuild/Microsoft.Testing.Platform.MSBuild.csproj
```diff
@@ -1,15 +1,28 @@
...
+    <Compile Include="$(RepoRoot)src\Platform\Microsoft.Testing.Platform\Helpers\EnvironmentVariableConstants.cs" Link="Helpers\EnvironmentVariableConstants.cs" />
+    <Compile Include="$(RepoRoot)src\Platform\Microsoft.Testing.Platform\Helpers\ExitCodes.cs" Link="Helpers\ExitCodes.cs" />
+    <Compile Include="$(RepoRoot)src\Platform\Microsoft.Testing.Platform\Helpers\TimeoutHelper.cs" Link="Helpers\TimeoutHelper.cs" />
+    <Compile Include="$(RepoRoot)src\Platform\Microsoft.Testing.Platform\Helpers\TimeSpanParser.cs" Link="Helpers\TimeSpanParser.cs" />
+    <Compile Include="$(RepoRoot)src\Platform\Microsoft.Testing.Platform\Helpers\System\IEnvironment.cs" Link="Helpers\IEnvironment.cs" />
+    <Compile Include="$(RepoRoot)src\Platform\Microsoft.Testing.Platform\Helpers\System\SystemEnvironment.cs" Link="Helpers\SystemEnvironment.cs" />
+    <Compile Include="$(RepoRoot)src\Platform\Microsoft.Testing.Platform\Helpers\System\ITask.cs" Link="Helpers\ITask.cs" />
+    <Compile Include="$(RepoRoot)src\Platform\Microsoft.Testing.Platform\Helpers\System\SystemTask.cs" Link="Helpers\SystemTask.cs" />
```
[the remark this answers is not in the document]
Evangelink: The `IEnvironment`, `SystemEnvironment`, `ITask`, and `SystemTask` types intentionally do **not** have the `[Embedded]` attribute because they cross the DI/service provider boundary. Extension projects (HangDump, TrxReport, Retry, Telemetry, etc.) obtain instances of these types from `Microsoft.Testing.Platform`'s service provider via `serviceProvider.GetEnvironment()` and `serviceProvider.GetTask()`. If these types were source-embedded locally, the locally-compiled type and the Platform DLL's type would be distinct CLR types, causing `CS1503` ("cannot convert") errors at all service provider call sites.

Only types that are used purely as static utilities within each assembly (like `ExitCodes`, `ApplicationStateGuard`, `RoslynString`, etc.) can be marked `[Embedded]` — they never need to be assignment-compatible across assembly boundaries.

For `Microsoft.Testing.Platform.MSBuild` and `Microsoft.Testing.Extensions.MSBuild`, the System files ARE source-embedded because these projects use the IPC types internally and create their own instances (e.g., `new SystemEnvironment()`), without bridging types from the Platform's service provider. `CS0436` is suppressed in these projects to handle the expected type shadowing.

src/Platform/Microsoft.Testing.Platform/Helpers/ApplicationStateGuard.cs
```diff
@@ -23,7 +26,7 @@ public static void Ensure([DoesNotReturnIf(false)] bool condition, [CallerFilePa
         {
             throw new InvalidOperationException(string.Format(
                 CultureInfo.InvariantCulture,
-#if PLATFORM_MSBUILD
+#if TESTING_PLATFORM_SOURCE_EMBEDDED
                 "Unexpected state in file '{0}' at line '{1}'",
```
Youssef1313: I see we have been doing that for the MSBuild extension, but it's not ideal to lose localization.

src/Platform/Microsoft.Testing.Extensions.MSBuild/Microsoft.Testing.Extensions.MSBuild.csproj
```diff
@@ -4,13 +4,27 @@
     <TargetFrameworks>$(SupportedNetFrameworks);netstandard2.0</TargetFrameworks>
     <!-- Packaged as part of Microsoft.Testing.Platform.MSBuild -->
     <IsPackable>false</IsPackable>
+    <DefineConstants>$(DefineConstants);TESTING_PLATFORM_SOURCE_EMBEDDED</DefineConstants>
+    <NoWarn>$(NoWarn);CS0436</NoWarn>
```
Youssef1313: Where is that warning happening exactly?

src/Platform/Microsoft.Testing.Platform.MSBuild/Microsoft.Testing.Platform.MSBuild.csproj
```diff
@@ -1,15 +1,28 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
   <PropertyGroup>
     <TargetFrameworks>$(SupportedNetFrameworks);netstandard2.0</TargetFrameworks>
-    <DefineConstants>$(DefineConstants);PLATFORM_MSBUILD</DefineConstants>
-    <NoWarn>$(NoWarn);NU5100</NoWarn>
+    <DefineConstants>$(DefineConstants);TESTING_PLATFORM_SOURCE_EMBEDDED</DefineConstants>
+    <NoWarn>$(NoWarn);NU5100;CS0436</NoWarn>
```
Youssef1313: Same question here

src/Platform/Microsoft.Testing.Platform/Helpers/ExitCodes.cs
```diff
@@ -9,8 +11,7 @@ namespace Microsoft.Testing.Platform.Helpers;
 /// </summary>
 // TODO: Consider changing this to an enum, and rename to 'ExitCode' to follow enum naming convention.
 // Being an enum makes it easier to do 'Enum.IsDefined' checks to validate if an exit code is a known MTP exit code.
-// Note: Changing this to enum is binary breaking for extensions that consume this via IVT :/
```
Youssef1313, line 12: The comment still partially applies when using MTP <= 2.1 with MTP extension >= 2.2

src/Platform/Microsoft.Testing.Platform/Helpers/ApplicationStateGuard.cs
Evangelink: This comment is on the outdated diff. The previous code (`main`) had `#if PLATFORM_MSBUILD` blocks that used **hardcoded English strings** — this PR actually **restores localization** by using `EmbeddedPlatformResources` (via `#if TESTING_PLATFORM_SOURCE_EMBEDDED`) instead.

The `EmbeddedPlatformResources.resx` is now embedded into both MSBuild projects, so `ApplicationStateGuard` uses the localized resource strings everywhere instead of the old hardcoded fallback.

src/Platform/Microsoft.Testing.Extensions.MSBuild/Microsoft.Testing.Extensions.MSBuild.csproj
Evangelink: CS0436 ("The type 'X' in 'file.cs' conflicts with the imported type 'X' in 'Assembly.dll'") fires because these two projects both:
1. Have a `<ProjectReference>` to `Microsoft.Testing.Platform`
2. Source-embed types from it (e.g., `ExitCodes`, `ApplicationStateGuard`, `RoslynString`, `IEnvironment`, `SystemEnvironment`, etc.)

So the compiler sees the same types defined locally (via `<Compile Include="...">`) AND in the referenced `Microsoft.Testing.Platform.dll`. The `[Embedded]` attribute signals to tooling that this is intentional source embedding, but the C# compiler still emits CS0436 for the type shadowing conflict.

Specifically, the files that trigger it are all the ones listed under `<!-- Embedded helpers from Microsoft.Testing.Platform -->` in the csproj — any type that exists both as a source-compiled file and in the referenced assembly.

The suppression was already present in `Microsoft.Testing.Platform.MSBuild.csproj` on `main` (for the IPC types that were already source-embedded), this PR just extends it to `Microsoft.Testing.Extensions.MSBuild.csproj` as well because the same pattern is now applied there too.

src/Platform/Microsoft.Testing.Platform.MSBuild/Microsoft.Testing.Platform.MSBuild.csproj
Evangelink: Same as above — see the [reply on Microsoft.Testing.Extensions.MSBuild.csproj](https://github.com/microsoft/testfx/pull/7443#discussion_r2895232152) for the full explanation. The CS0436 suppression was already on `main` in this file (for the IPC source-embedded types); this PR extends the embedded file set, so the suppression covers the newly-added helpers too.

src/Platform/Microsoft.Testing.Platform/Helpers/ExitCodes.cs
Evangelink: Good point — I've updated the comment to clarify that the backward compat concern is specifically about extensions that reference the type directly from the assembly (via IVT) rather than via source embedding:

```csharp
// Note: Changing this to enum would be binary breaking for extensions built against MTP <= 2.1 that still consume this via IVT
// (those extensions reference the class directly from the MTP assembly, not via source embedding).
```

src/Platform/Microsoft.Testing.Platform.MSBuild/Microsoft.Testing.Platform.MSBuild.csproj
Youssef1313: There are no suppressions on main that I can see. Something is off here, we shouldn't suppress CS0436

Evangelink: Generated reply reviewed by @Evangelink:

You're right there's no CS0436 suppression on `main`. The suppression is newly needed because this PR now source-embeds additional helper types (ExitCodes, ApplicationStateGuard, RoslynString, EnvironmentVariableConstants, TimeoutHelper, TimeSpanParser, IEnvironment, SystemEnvironment, ITask, SystemTask) alongside the IPC types that were already source-embedded.

These projects still need IVT to `Microsoft.Testing.Platform` because their source-embedded IPC code (NamedPipeClient, BaseSerializer, etc.) references **non-embedded** internal types like `PlatformResources` and `VoidResponseFieldsId`. For `Microsoft.Testing.Extensions.MSBuild` specifically, it also uses non-embedded internal types `AppVersion` and `TestRequestExecutionTimeInfo`.

So the IVT is required → internal types are visible from the referenced assembly → the source-embedded types conflict with the assembly's copies → CS0436 fires. The suppression is the standard approach when intentionally source-embedding types from a referenced assembly while maintaining IVT for other internal types.

Added inline csproj comments documenting exactly why CS0436 suppression is needed in both MSBuild projects.

src/Platform/Microsoft.Testing.Extensions.MSBuild/Microsoft.Testing.Extensions.MSBuild.csproj
Evangelink: Generated reply reviewed by @Evangelink:

Adding to the explanation above — I've added an inline csproj comment documenting why CS0436 is suppressed here:

```xml
<!-- CS0436: Source-embedded types (IPC, helpers) conflict with the same types visible via IVT from Microsoft.Testing.Platform.
     IVT is still required because this project consumes non-embedded internal types (AppVersion, TestRequestExecutionTimeInfo). -->
```

The IVT can't be fully removed from this project because `MSBuildCommandLineProvider`, `MSBuildConsumer`, and other classes directly use `AppVersion.DefaultSemVer` (internal static class) and `TestRequestExecutionTimeInfo` (internal struct) from Platform.

test/IntegrationTests/Microsoft.Testing.Platform.Acceptance.IntegrationTests/Microsoft.Testing.Platform.Acceptance.IntegrationTests.csproj
```diff
@@ -7,10 +7,21 @@
     <OutputType>Exe</OutputType>
     <DefineConstants Condition=" '$(FastAcceptanceTest)' == 'true'">$(DefineConstants);SKIP_INTERMEDIATE_TARGET_FRAMEWORKS</DefineConstants>
     <UseAppHost Condition="$([MSBuild]::IsOSPlatform('OSX'))">false</UseAppHost>
+    <DefineConstants>$(DefineConstants);TESTING_PLATFORM_SOURCE_EMBEDDED</DefineConstants>
   </PropertyGroup>
```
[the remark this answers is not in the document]
Evangelink: Generated reply reviewed by @Evangelink:

Fixed — reordered `DefineConstants` so the unconditional `TESTING_PLATFORM_SOURCE_EMBEDDED` comes first, then the conditional `SKIP_INTERMEDIATE_TARGET_FRAMEWORKS` appends to it. Both symbols are now reliably preserved regardless of `FastAcceptanceTest` value.

test/IntegrationTests/MSTest.Acceptance.IntegrationTests/MSTest.Acceptance.IntegrationTests.csproj
```diff
@@ -6,8 +6,17 @@
     <OutputType>Exe</OutputType>
     <DefineConstants Condition=" '$(FastAcceptanceTest)' == 'true'">$(DefineConstants);SKIP_INTERMEDIATE_TARGET_FRAMEWORKS</DefineConstants>
     <UseAppHost Condition="$([MSBuild]::IsOSPlatform('OSX'))">false</UseAppHost>
+    <DefineConstants>$(DefineConstants);TESTING_PLATFORM_SOURCE_EMBEDDED</DefineConstants>
   </PropertyGroup>
```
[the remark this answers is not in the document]
Evangelink: Generated reply reviewed by @Evangelink:

Fixed — same reordering applied here. The unconditional `TESTING_PLATFORM_SOURCE_EMBEDDED` now comes first, and the conditional `SKIP_INTERMEDIATE_TARGET_FRAMEWORKS` appends after it.

src/Platform/Microsoft.Testing.Platform.MSBuild/Microsoft.Testing.Platform.MSBuild.csproj
Youssef1313: I don't think the suppression should be needed or is correct. It's very concerning to me.

Evangelink: I was able to get rid of it by removing more inclusion of file from the platform (task, environment)

test/Utilities/Microsoft.Testing.TestInfrastructure/ProcessFactory.cs
```diff
@@ -21,6 +23,8 @@ public static IProcessHandle Start(ProcessConfiguration config, bool cleanDefaul
             RedirectStandardOutput = true,
             RedirectStandardError = true,
             RedirectStandardInput = true,
+            StandardOutputEncoding = Encoding.UTF8,
+            StandardErrorEncoding = Encoding.UTF8,
```
Youssef1313, line 27: Why is this needed now and wasn't needed before?

Evangelink: We are now testing localization and French accents were being badly rendered by default encoding.

PR #7444, opened 2026-02-20
Improve assertions error messages with structured format

### LINKED ISSUE #7421: Improve AreSame error messages in case of null values

Evangelink:
Fix #7421

### DIFF
--- a/test/UnitTests/TestFramework.UnitTests/Assertions/AssertTests.ThrowsExceptionTests.cs
+++ b/test/UnitTests/TestFramework.UnitTests/Assertions/AssertTests.ThrowsExceptionTests.cs
@@ -13,10 +13,7 @@ public partial class AssertTests
     {
         Action action = () => Assert.ThrowAssertFailed("name", "{");
         action.Should().Throw<AssertFailedException>()
-            .WithMessage("""
-            name
-            {
-            """);
+            .WithMessage("name" + Environment.NewLine + "{");
     }
     #endregion
 
--- a/test/UnitTests/TestFramework.UnitTests/Assertions/AssertTests.cs
+++ b/test/UnitTests/TestFramework.UnitTests/Assertions/AssertTests.cs
@@ -31,10 +31,9 @@ public partial class AssertTests
         Action act = () => Assert.Equals("test", "test");
 #pragma warning restore CS0618 // Type or member is obsolete
         act.Should().Throw<AssertFailedException>()
-           .WithMessage("""
-            Assert.Fail
-            Assert.Equals should not be used for Assertions. Please use Assert.AreEqual & overloads instead.
-            """);
+           .WithMessage(
+                "Assert.Fail" + Environment.NewLine +
+                "Assert.Equals should not be used for Assertions. Please use Assert.AreEqual & overloads instead.");
     }
 
     public void ObsoleteReferenceEqualsMethodThrowsAssertFailedException()
@@ -44,10 +43,9 @@ public partial class AssertTests
         Action act = () => Assert.ReferenceEquals(obj, obj);
 #pragma warning restore CS0618 // Type or member is obsolete
         act.Should().Throw<AssertFailedException>()
-           .WithMessage("""
-            Assert.Fail
-            Assert.ReferenceEquals should not be used for Assertions. Please use Assert.AreSame & overloads instead.
-            """);
+           .WithMessage(
+                "Assert.Fail" + Environment.NewLine +
+                "Assert.ReferenceEquals should not be used for Assertions. Please use Assert.AreSame & overloads instead.");
     }
 #endif
     #endregion


src/TestFramework/TestFramework/Assertions/Assert.AreEqual.cs
```diff
@@ -544,27 +529,15 @@ private static bool AreEqualFailing(decimal expected, decimal actual, decimal de
...
-
         if (expected is null || actual is null)
         {
-            return string.Format(
-                CultureInfo.CurrentCulture,
-                FrameworkMessages.AreEqualFailMsg,
-                userMessage,
-                ReplaceNulls(expected),
-                ReplaceNulls(actual));
+            string message = string.IsNullOrEmpty(userMessage) ? string.Empty : userMessage;
```
Youssef1313: If `userMessage` can be null, this line can be simplified to:

```suggestion
            string message = userMessage ?? string.Empty;
```

Otherwise, it could be just:


```suggestion
            string message = userMessage;
```

src/TestFramework/TestFramework/Assertions/Assert.AreSame.cs
```diff
@@ -177,26 +179,44 @@ public static void AreSame<T>(T? expected, T? actual, string? message = "", [Cal
...
-        string finalMessage = userMessage;
+        string message = string.IsNullOrEmpty(userMessage) ? string.Empty : userMessage!;
+
+        // When both values have the same string representation, include hash codes
+        // to help the user understand they are different object instances.
+        string expectedFormatted = FormatValue(expected);
+        string actualFormatted = FormatValue(actual);
+        bool sameToString = expected is not null && actual is not null && expectedFormatted == actualFormatted;
+        string expectedValue = sameToString
+            ? expectedFormatted + $" (HashCode={RuntimeHelpers.GetHashCode(expected!)})"
```
Youssef1313: This might be confusing to users as-is. Instead of `HashCode=`, maybe we should be explicit this is `RuntimeHelpers.GetHashCode`. Otherwise, users might think this is equivalent to `x.GetHashCode()` (the difference is only if the type overrides the default `GetHashCode`)

src/TestFramework/TestFramework/Assertions/Assert.cs
```diff
@@ -74,125 +83,337 @@ private static bool ShouldLaunchDebugger()
...
-    /// The parameter.
-    /// </param>
-    /// <param name="assertionName">
-    /// The assertion Name.
-    /// </param>
-    /// <param name="parameterName">
-    /// parameter name.
-    /// </param>
-    internal static void CheckParameterNotNull([NotNull] object? param, string assertionName, string parameterName)
+    internal static string FormatCollectionParameter(string paramName, string expression, IEnumerable collection)
```
Youssef1313: paramName is always expected to be `collection` here right? Do we need this `paramName` parameter?

src/TestFramework/TestFramework/Assertions/Assert.cs
```diff
@@ -74,125 +83,337 @@ private static bool ShouldLaunchDebugger()
...
+        int currentLength = 0;
+        bool truncated = false;
+
+        // Perf: wrap in try-catch so that faulting enumerators (e.g. collection modified during
+        // iteration, or user-defined iterators that throw) don't bubble up from assertion formatting.
+        try
         {
-            string finalMessage = string.Format(CultureInfo.CurrentCulture, FrameworkMessages.NullParameterToAssert, parameterName);
-            ThrowAssertFailed(assertionName, finalMessage);
+            foreach (object? item in collection)
```
Youssef1313, line 285: Multiple enumerations are extremely risky IMO.
The original assert API that failed has likely already started enumerating the collection.

Youssef1313: There is no guarantee that enumerating the given IEnumerable will be leading to the same original result as the first enumeration that happened earlier.

test/UnitTests/TestFramework.UnitTests/Assertions/AssertTests.AreSame.cs
```diff
@@ -16,7 +16,12 @@ public void AreSame_PassSameObject_ShouldPass()
     public void AreSame_PassDifferentObject_ShouldFail()
     {
         Action action = () => Assert.AreSame(new object(), new object());
-        action.Should().Throw<Exception>().WithMessage("Assert.AreSame failed. 'expected' expression: 'new object()', 'actual' expression: 'new object()'.");
+        action.Should().Throw<Exception>().WithMessage("""
+            Assert.AreSame failed.
+            Expected references to be the same.
+              expected (new object()): <System.Object> (HashCode=*)
```
Youssef1313: Should this look like the following instead?

> expected 'new object()': <System.Object> (HashCode=*)

nohwnd: Yeah probably (even though then it looks like string...), personally I always trip over the expressions in the assertion messages lately, it is hard to know what is going on there when your code that produces the value is longer than few characters. 

or when your expression is `actual`, more examples on the general approach would be nice that explore what will happen for common usecases.

test/UnitTests/TestFramework.UnitTests/Assertions/AssertTests.AreEqualTests.cs
```diff
@@ -302,7 +371,12 @@ public void AreEqualTwoObjectsDifferentTypeShouldFail()
     {
         Action action = () => Assert.AreEqual(new object(), 1);
         action.Should().Throw<AssertFailedException>()
-            .And.Message.Should().Contain("Assert.AreEqual failed. Expected:<System.Object (System.Object)>. Actual:<1 (System.Int32)>.");
+            .WithMessage("""
+                Assert.AreEqual failed.
+                Expected values to be equal.
+                  expected (new object()): <System.Object> (System.Object)
```
nohwnd: Looking through the code to see places where I would be confused.

This is imho very hard to interpret. The additional info about expressions and types should aid in understanding what is going on. And I would avoid putting anything before the actual values, instead I would prefer aligning the values on the same level, and add additional information after if needed, something like:

```
Assert.AreEqual failed.
Expected values to be equal.
    expected: System.Object [System.Object]
    actual:    1 [System.Int32]
 at Assert.AreEqual(new object(), 1).
```

This seems okay, and IMHO I have all the info needed. The values, the types, and what was called.

nohwnd: What does the `< >` do in the output? Is that telling me that this was unprintable value or something?

test/UnitTests/TestFramework.UnitTests/Assertions/AssertTests.AreEqualTests.cs
```diff
@@ -373,7 +447,12 @@ public async Task GenericAreEqual_InterpolatedString_DifferentValues_ShouldFail(
         DateTime dateTime = DateTime.Now;
         Func<Task> action = async () => Assert.AreEqual(0, 1, $"User-provided message. {o}, {o,35}, {await GetHelloStringAsync()}, {new DummyIFormattable()}, {dateTime:tt}, {dateTime,5:tt}");
         (await action.Should().ThrowAsync<Exception>())
-            .WithMessage($"Assert.AreEqual failed. Expected:<0>. Actual:<1>. 'expected' expression: '0', 'actual' expression: '1'. User-provided message. DummyClassTrackingToStringCalls,     DummyClassTrackingToStringCalls, Hello, DummyIFormattable.ToString(), {string.Format(null, "{0:tt}", dateTime)}, {string.Format(null, "{0,5:tt}", dateTime)}");
+            .WithMessage("""
+                Assert.AreEqual failed. User-provided message. DummyClassTrackingToStringCalls,     DummyClassTrackingToStringCalls, Hello, DummyIFormattable.ToString()*
```
nohwnd: the calls to dummyclass are something that I care very less about in most cases, yet it clutters the output a lot. I prefer to see the values that were compared first, and then maybe see what was exactly called, first stack frame is usually good enough for that?

test/UnitTests/TestFramework.UnitTests/Assertions/AssertTests.AreEqualTests.cs
```diff
@@ -458,7 +557,12 @@ public async Task DecimalAreNotEqual_InterpolatedString_SameValues_ShouldFail()
         DateTime dateTime = DateTime.Now;
         Func<Task> action = async () => Assert.AreNotEqual(1.0m, 1.1m, 0.2m, $"User-provided message. {o}, {o,35}, {await GetHelloStringAsync()}, {new DummyIFormattable()}, {dateTime:tt}, {dateTime,5:tt}");
         (await action.Should().ThrowAsync<Exception>())
-            .WithMessage($"Assert.AreNotEqual failed. Expected a difference greater than <0.2> between expected value <1.0> and actual value <1.1>. 'notExpected' expression: '1.0m', 'actual' expression: '1.1m'. User-provided message. DummyClassTrackingToStringCalls,     DummyClassTrackingToStringCalls, Hello, DummyIFormattable.ToString(), {string.Format(null, "{0:tt}", dateTime)}, {string.Format(null, "{0,5:tt}", dateTime)}");
+            .WithMessage("""
+                Assert.AreNotEqual failed. User-provided message. DummyClassTrackingToStringCalls,     DummyClassTrackingToStringCalls, Hello, DummyIFormattable.ToString()*
+                Expected a difference greater than <0.2>.
+                  notExpected: 1.0m
```
nohwnd: This is existing code, but I am not 100% on board with this naming, makes me do additional mental math on what you meant here. The message does great job at telling me what is the expectation, and it already uses language appropriate for the Not*Assertion.

negative assertion:
Expected a difference **greater** than <0.2>.

positive assertion: 
Expected a difference **no greater** than <0>.

nohwnd: the `<>` also seem extra, do they serve some purpose somewhere?

test/UnitTests/TestFramework.UnitTests/Assertions/AssertTests.AreEqualTests.cs
```diff
@@ -1436,10 +1632,11 @@ public void AreEqualStringWithSpecialCharactersShouldEscape()
         Action action = () => Assert.AreEqual("aa\ta", "aa a");
         action.Should().Throw<AssertFailedException>()
             .WithMessage("""
-            Assert.AreEqual failed. String lengths are both 4 but differ at index 2. 'expected' expression: '"aa\ta"', 'actual' expression: '"aa a"'.
-            Expected: "aa␉a"
-            But was:  "aa a"
-            -------------^
+            Assert.AreEqual failed.
+            String lengths are both 4 but differ at index 2.
+              expected ("aa\ta"): "aa?a"
```
nohwnd: One more example of the expression not really helping to explain what is going on

test/UnitTests/TestFramework.UnitTests/Assertions/AssertTests.AreEqualTests.cs
```diff
@@ -1436,10 +1632,11 @@ public void AreEqualStringWithSpecialCharactersShouldEscape()
...
+              actual:             "aa a"
+            -------------------------^
```
nohwnd: This arrow should start  at the same level as the actual word, not 2 chars before.

test/UnitTests/TestFramework.UnitTests/Assertions/AssertTests.AreEqualTests.cs
```diff
@@ -1451,10 +1648,11 @@ public void AreEqualLongStringsShouldTruncateAndShowContext()
...
         action.Should().Throw<AssertFailedException>()
             .WithMessage("""
-            Assert.AreEqual failed. String lengths are both 201 but differ at index 100. 'expected' expression: 'expected', 'actual' expression: 'actual'.
-            Expected: "...aaaaaaaaaaaaaaaaaabcccccccccccccccc..."
-            But was:  "...aaaaaaaaaaaaaaaaaadcccccccccccccccc..."
-            --------------------------------^
+            Assert.AreEqual failed.
+            String lengths are both 201 but differ at index 100.
+              expected: "...aaaaaaaaaaaaaaaaaabcccccccccccccccc..."
+              actual:   "...aaaaaaaaaaaaaaaaaadcccccccccccccccc..."
```
nohwnd, line 1701: values are aligned, this is how it should be everywhere.

test/UnitTests/TestFramework.UnitTests/Assertions/AssertTests.AreEqualTests.cs
```diff
@@ -1722,4 +1925,94 @@ private void StringPreviewsAreEqual(string expected, string actual)
...
+    {
+        var expected = new ObjectWithLongToString();
+        var actual = new ObjectWithLongToString();
+
+        Action action = () => Assert.AreEqual(expected, actual);
+        action.Should().Throw<AssertFailedException>()
+            .WithMessage($"""
+                Assert.AreEqual failed.
+                Expected values to be equal.
+                  expected: {new string('L', 256)}... (300 chars)
```
nohwnd: what is this telling me, that there are 300 chars in total (then how do I know at how many I am looking at), or that there are additional 300 chars?

test/UnitTests/TestFramework.UnitTests/Assertions/AssertTests.AreSame.cs
nohwnd: Commented in few other places, where this seems problematic.

test/UnitTests/TestFramework.UnitTests/Assertions/AssertTests.cs
```diff
@@ -65,4 +65,230 @@ private sealed class DummyIFormattable : IFormattable
...
+            collection.Add(new string((char)('a' + (i % 26)), 30));
+        }
+
+        Action action = () => Assert.Contains("not-there", collection);
+        action.Should().Throw<AssertFailedException>()
+            .WithMessage("""
+                Assert.Contains failed.
+                Expected collection to contain the specified item.
+                  expected: "not-there"
+                  collection*[*...*] (20 elements)
```
nohwnd: maybe use `item`, rather than element? shorter and probably easier to grasp

test/UnitTests/TestFramework.UnitTests/Assertions/AssertTests.cs
```diff
@@ -65,4 +65,230 @@ private sealed class DummyIFormattable : IFormattable
...
+        var inner1 = new List<int> { 1, 2 };
+        var inner2 = new List<int> { 3, 4 };
+        var outer = new List<List<int>> { inner1, inner2 };
+
+        Action action = () => Assert.IsNull(outer);
+        action.Should().Throw<AssertFailedException>()
+            .WithMessage("""
+                Assert.IsNull failed.
+                Expected value to be null.
+                  value (outer): [[1, 2] (2 elements), [3, 4] (2 elements)] (2 elements)
```
nohwnd: This will go wild very fast, I would avoid adding the count when you are able to render the whole thing. how I do it in pester is that I set a limit and add `...10 more` when the array is too big.

There is also width limit to the output, which when I cross it I format values on new lines, in interactive session that is width of the rendering space, in CI it is arbitrary value something like 120.

test/UnitTests/TestFramework.UnitTests/Assertions/AssertTests.cs
```diff
@@ -65,4 +65,230 @@ private sealed class DummyIFormattable : IFormattable
...
+        }
+
+        var outer = new List<List<int>> { inner };
+
+        Action action = () => Assert.IsNull(outer);
+        action.Should().Throw<AssertFailedException>()
+            .WithMessage("""
+                Assert.IsNull failed.
+                Expected value to be null.
+                  value (outer): [[0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, ...] (50 elements)] (1 element)
```
nohwnd: This kind of output forces me to count how many were rendered. 

value (outer): [[0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, ...35 more]]

Is imho much easier to read, we can see that the array is doubled, and it will be easier to see when there is more than 1 element.

test/UnitTests/TestFramework.UnitTests/Assertions/AssertTests.cs
```diff
@@ -65,4 +65,230 @@ private sealed class DummyIFormattable : IFormattable
...
+    {
+        var collection = new List<int> { 42 };
+
+        Action action = () => Assert.Contains(99, collection);
+        action.Should().Throw<AssertFailedException>()
+            .WithMessage("""
+                Assert.Contains failed.
+                Expected collection to contain the specified item.
+                  expected: 99
+                  collection*[42] (1 element)
```
nohwnd: No idea how to interpret this, why it is not actual:

test/UnitTests/TestFramework.UnitTests/Assertions/AssertTests.cs
Evangelink: There is a limit in number of char for the length of the expression and the length of the value (for all assertion results). There are some examples.

test/UnitTests/TestFramework.UnitTests/Assertions/AssertTests.AreEqualTests.cs
Evangelink: That's a complex user message, to ensure we correctly format all the interpolated parts (we had some bugs) so it would not be something in the stack trace.

test/UnitTests/TestFramework.UnitTests/Assertions/AssertTests.AreEqualTests.cs
Evangelink: I don't get the first comment, what's the problem and what's your suggestion?

nohwnd: Expected a difference greater than <0.2>. notExpected: 1.0m actual: 1.0m
 Expected a difference no greater than <0>. expected: 1.0m actual: 0.9m
 
 1.0m is the value I provided, that drives the decision if the assertion will fail or not, but it says "notExpected/unexpected".

src/TestFramework/TestFramework/Assertions/Assert.AreSame.cs
Evangelink: Maybe just `Hash=XXX`?

test/UnitTests/TestFramework.UnitTests/Assertions/AssertTests.AreSame.cs
```diff
@@ -97,7 +137,96 @@ public async Task AreNotSame_InterpolatedString_PassSameObject_ShouldFail()
...
+    }
+
+    public void AreSame_WithNewlineInToString_ShouldEscapeNewlines()
+    {
+        Action action = () => Assert.AreSame(new ObjectWithNewlineToString(), new ObjectWithNewlineToString());
+        action.Should().Throw<AssertFailedException>()
+            .WithMessage("""
+                Assert.AreSame(new ObjectWithNewlineToString(), new ObjectWithNewlineToString()) failed.
+                Expected references to be the same.
+                  expected: line1\r\nline2\nline3 (Hash=*)
```
Evangelink, line 190: Maybe we can consider ObjectUID instead of Hash

src/TestFramework/TestFramework/Resources/xlf/FrameworkMessages.es.xlf
```diff
@@ -17,46 +17,20 @@
...
+        <note />
       </trans-unit>
-      <trans-unit id="AreEqualStringDiffActualPrefix">
-        <source>But was:  </source>
-        <target state="translated">Pero fue:  </target>
-        <note></note>
+      <trans-unit id="AreEqualFailNew">
+        <source>Expected values to be equal.</source>
+        <target state="new">Expected values to be equal.</target>
+        <note />
```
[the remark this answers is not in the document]
Evangelink: The XLF files were regenerated by building the project (`dotnet build`). The format is now whatever the build tooling produces.

src/TestFramework/TestFramework/Assertions/Assert.IsExactInstanceOfType.cs
```diff
@@ -405,19 +398,21 @@ private static bool IsNotExactInstanceOfTypeFailing(object? value, [NotNullWhen(
...
         {
-            finalMessage = string.Format(
-                CultureInfo.CurrentCulture,
-                FrameworkMessages.IsNotExactInstanceOfFailMsg,
-                userMessage,
-                wrongType.ToString(),
-                value!.GetType().ToString());
+            message += Environment.NewLine + $"  wrongType: {wrongType}"
+                     + Environment.NewLine + $"  actualType: {value!.GetType()}";
         }
```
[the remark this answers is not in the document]
Evangelink: Fixed. When `wrongType` is null, the message now always includes the failure message and shows `wrongType: (null)` as a diagnostic. Also applied the same fix to `ThrowAssertIsExactInstanceOfTypeFailed` (line 323) to handle `expectedType` being null. Updated tests accordingly.

src/TestFramework/TestFramework/Resources/xlf/FrameworkMessages.zh-Hans.xlf
```diff
@@ -17,46 +17,20 @@
...
+        <note />
       </trans-unit>
-      <trans-unit id="AreEqualStringDiffActualPrefix">
-        <source>But was:  </source>
-        <target state="translated">但却是: </target>
-        <note></note>
+      <trans-unit id="AreEqualFailNew">
+        <source>Expected values to be equal.</source>
+        <target state="new">Expected values to be equal.</target>
+        <note />
```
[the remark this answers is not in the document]
Evangelink: The XLF files were regenerated by building the project (`dotnet build`). The format is now whatever the build tooling produces.

src/TestFramework/TestFramework/Assertions/Assert.IsInstanceOfType.cs
```diff
@@ -409,19 +407,21 @@ private static bool IsNotInstanceOfTypeFailing(object? value, [NotNullWhen(false
...
-        string finalMessage = userMessage;
+        string message = string.IsNullOrEmpty(userMessage) ? string.Empty : userMessage!;
+        if (wrongType is not null)
+        {
+            message += Environment.NewLine + FrameworkMessages.IsNotInstanceOfTypeFailNew;
+        }
+
+        message += Environment.NewLine + FormatParameter(nameof(value), valueExpression, value);
         if (wrongType is not null)
         {
```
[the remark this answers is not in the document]
Evangelink: Fixed. When `wrongType` is null, the message now always includes the failure message and shows `wrongType: (null)` as a diagnostic. Also applied the same fix to `ThrowAssertIsInstanceOfTypeFailed` (line 332) to handle `expectedType` being null. Updated tests accordingly.

test/UnitTests/TestFramework.UnitTests/Assertions/AssertTests.AreEqualTests.cs
Evangelink: This is general design feedback on the expression-based messaging approach, and the code referenced here has been reworked since. Resolving as outdated.

src/TestFramework/TestFramework/Assertions/Assert.IsExactInstanceOfType.cs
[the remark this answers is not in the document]
Evangelink: Already fixed. When `wrongType` is null, the message now includes `wrongType: null` / `expectedType: null` as a diagnostic line. Applied to both `ThrowAssertIsNotExactInstanceOfTypeFailed` and `ThrowAssertIsExactInstanceOfTypeFailed`.

src/TestFramework/TestFramework/Assertions/Assert.cs
```diff
@@ -74,125 +77,345 @@ private static bool ShouldLaunchDebugger()
...
-        => BuildUserMessageForTwoExpressions(format, patternExpression, "pattern", valueExpression, "value");
+        string elementList = string.Join(", ", elements);
+        if (truncated)
+        {
+            int remaining = totalCount - displayedCount;
+            string remainingText = knownCount is null
+                ? $"{remaining}+"
+                : $"{remaining}";
+            elementList += $", ... {remainingText} more";
+        }
```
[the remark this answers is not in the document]
Evangelink: Fixed. A separate `failedEnumeration` flag now tracks whether enumeration was cut short by an exception. When `failedEnumeration` is true, the suffix is simply `", ..."` instead of a numeric `"... N more"` suffix. This avoids the misleading `"... 0+ more"` output.

src/TestFramework/TestFramework/Assertions/Assert.That.cs
```diff
@@ -36,23 +36,113 @@ public static void That(Expression<Func<bool>> condition, string? message = null
...
+
+        return binary.NodeType switch
+        {
+            ExpressionType.Equal => $"Expected {leftValue} to equal {rightValue}.",
+            ExpressionType.NotEqual => $"Expected {leftValue} to not equal {rightValue}.",
+            ExpressionType.GreaterThan => $"Expected {leftValue} to be greater than {rightValue}.",
+            ExpressionType.GreaterThanOrEqual => $"Expected {leftValue} to be greater than or equal to {rightValue}.",
+            ExpressionType.LessThan => $"Expected {leftValue} to be less than {rightValue}.",
+            ExpressionType.LessThanOrEqual => $"Expected {leftValue} to be less than or equal to {rightValue}.",
+            _ => FrameworkMessages.IsTrueFailNew,
```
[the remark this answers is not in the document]
Evangelink: All user-facing failure sentences in `Assert.That` have been moved to `FrameworkMessages.resx` resources (e.g., `AssertThatExpectedTrueFailNew`, `AssertThatExpectedFalseFailNew`, `AssertThatEqualFailNew`, etc.). The code now uses `string.Format(CultureInfo.InvariantCulture, FrameworkMessages.AssertThat*FailNew, ...)` consistently.

src/TestFramework/TestFramework/Assertions/Assert.That.cs
```diff
@@ -36,23 +36,113 @@ public static void That(Expression<Func<bool>> condition, string? message = null
...
+    private static string TryEvaluateFormatted(Expression expr)
+    {
+        try
+        {
+            object? value = Expression.Lambda(expr).Compile().DynamicInvoke();
+            return FormatValue(value);
+        }
+        catch (Exception)
+        {
+            return GetCleanMemberName(expr);
```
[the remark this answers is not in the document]
Evangelink: Acknowledged. `TryEvaluateFormatted` only runs on the failure path, and errors are caught and fall back to displaying the expression name. A comment has been added documenting this design decision. The risk is acceptable given the diagnostic-only use case.

src/TestFramework/TestFramework/Assertions/Assert.cs
```diff
@@ -74,125 +77,356 @@ private static bool ShouldLaunchDebugger()
...
+
+        if (value is string s)
+        {
+            return EscapeNewlines(Truncate($"\"{s}\"", maxLength));
+        }
+
+        // For collections, show a preview with element values
+        if (value is IEnumerable enumerable)
+        {
+            return FormatCollectionPreview(enumerable);
```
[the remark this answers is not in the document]
Evangelink: Fixed. `FormatValue` now passes `maxLength` through to `FormatCollectionPreview(enumerable, maxLength)`. The behavior is now consistent regardless of runtime type.

src/TestFramework/TestFramework/Assertions/Assert.Contains.cs
```diff
@@ -771,73 +771,129 @@ public static void IsInRange<T>(T minValue, T maxValue, T value, string? message
...
-            CultureInfo.CurrentCulture,
-            FrameworkMessages.DoesNotContainPredicateFailMsg,
-            userMessage);
-        ThrowAssertFailed("Assert.DoesNotContain", finalMessage);
+        string callSite = FormatCallSite("Assert.IsInRange", new StringPair(nameof(value), valueExpression));
+        string message = string.Format(CultureInfo.CurrentCulture, FrameworkMessages.IsInRangeFailNew, FormatValue(value), FormatValue(minValue), FormatValue(maxValue));
+        message += FormatAlignedParameters(
+            new StringPair("range", $"[{FormatValue(minValue)}, {FormatValue(maxValue)}]"),
+            new StringPair(nameof(value), FormatValue(value)));
+        message = AppendUserMessage(message, userMessage);
```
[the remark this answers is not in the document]
Evangelink: Fixed. The `minValueExpression` and `maxValueExpression` parameters have been removed from `ThrowAssertIsInRangeFailed`. The method now only takes the values needed for the output.

test/IntegrationTests/MSTest.Acceptance.IntegrationTests/TrxReportTests.cs
```diff
@@ -33,7 +33,15 @@ public async Task TrxReport_WhenTestFails_ContainsExceptionInfoInOutput(string t
...
-        Assert.Contains("Assert.AreEqual failed. Expected:&lt;1&gt;. Actual:&lt;2&gt;.", trxContent, trxContent);
+        Assert.Contains(
+            """
+            Assert.AreEqual(1, 2)
+            Expected values to be equal.
+              expected: 1
+              actual:   2
+            """,
+            trxContent,
+            trxContent);
```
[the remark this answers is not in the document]
Evangelink: Already handled — the expected string already uses `.ReplaceLineEndings()` to normalize line endings to `Environment.NewLine`, making the assertion cross-platform safe.

test/UnitTests/TestFramework.UnitTests/Assertions/AssertTests.AreEqualTests.cs
Evangelink: This is general design feedback on the expression-based messaging approach, and the referenced code has been reworked since. Resolving as outdated.

src/TestFramework/TestFramework/Assertions/Assert.IsExactInstanceOfType.cs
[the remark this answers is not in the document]
Evangelink: Already fixed. When `wrongType` is null, the message now includes `wrongType: null` as a diagnostic line. Applied to both `ThrowAssertIsNotExactInstanceOfTypeFailed` and `ThrowAssertIsExactInstanceOfTypeFailed`.

test/UnitTests/TestFramework.UnitTests/Assertions/AssertTests.ThrowsExceptionTests.cs
```diff
@@ -13,7 +13,7 @@ public void ThrowAssertFailedDoesNotThrowIfMessageContainsInvalidStringFormatCom
     {
         Action action = () => Assert.ThrowAssertFailed("name", "{");
         action.Should().Throw<AssertFailedException>()
-            .WithMessage("*name failed. {*");
+            .WithMessage("name\n{");
     }
```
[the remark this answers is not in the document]
Evangelink: Fixed. Replaced raw string literal with explicit `Environment.NewLine` concatenation: `.WithMessage("name" + Environment.NewLine + "{")`.

test/UnitTests/TestFramework.UnitTests/Assertions/AssertTests.cs
```diff
@@ -31,7 +31,7 @@ public void ObsoleteEqualsMethodThrowsAssertFailedException()
         Action act = () => Assert.Equals("test", "test");
 #pragma warning restore CS0618 // Type or member is obsolete
         act.Should().Throw<AssertFailedException>()
-           .WithMessage("*Assert.Equals should not be used for Assertions*");
+           .WithMessage("Assert.Fail\nAssert.Equals should not be used for Assertions. Please use Assert.AreEqual & overloads instead.");
     }
```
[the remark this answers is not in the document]
Evangelink: Fixed. Replaced raw string literals with explicit `Environment.NewLine` concatenation in both `ObsoleteEqualsMethodThrowsAssertFailedException` and `ObsoleteReferenceEqualsMethodThrowsAssertFailedException` tests.

src/TestFramework/TestFramework/Assertions/Assert.cs
Youssef1313: This was marked as resolved, but I think this is still the second enumeration and my comment is still valid.

docs/RFCs/011-Structured-Assertion-Messages.md
```diff
@@ -0,0 +1,337 @@
...
+## Message Format
+
+Every assertion failure message follows this structure:
+
+```text
+<CallSite>
+[<UserMessage>]
+<FrameworkMessage>
+[  <param1>: <value1>]
+[  <param2>: <value2>]
```
nohwnd, line 43: I don't really agree with this  ordering. How the assertion was called is a minor detail that is often uninteresting, and can pollute the message when long variable names are used, or be useless like `Assert.AreEqual(expected, actual)`, in examples below parameters are cut to 2? even in cases where the tolerance is a useful value to see. The callsite info is imho nice to have, so we want to make it less prominent when not needed, and complete when user is looking for it.

This seem like workaround for making info that is not always useful too prominent in the message just because it is "cool".

User message should probably come second, providing additional information about the context of the assertion.

What I would like to see is: 
```
<summary without expanded values, unless the values are really short, with support for multiline>
<user message>
expected: <expected>
actual: <actual>
<any other info specific to the assertion>

<callsite>
<stack trace>
```

e.g.
```
Assert.AreEqual failed. Expected strings to be case sensitively equal. But they have different lengths, and differ at index 10.
  rendition of the string diff

<some user message if needed>
expected: full string
actual: full string
 
<any other info specific to the assertion, like case sensitivity, enconding, whatever is relevant to the assertion at hand>
 
<the call expression>
 
<stack trace.>
```
This way you have idea of what is wrong right away. you don't care about how ong the strings are because you don't end up with:
 
Expected the string <8 million chars> to be equal to <9 million chars.>.
 
Where you need to painstaikingly search for what even went wrong (fluent assertions style).
 
You also don't end up with:
 
 Assert.ContainsAny([ 300 lines of output for array or process ], .... ) failed because it does not contain [40 items].

docs/RFCs/011-Structured-Assertion-Messages.md
```diff
@@ -0,0 +1,337 @@
...
+Assert.That(x > 10)
+CollectionAssert.AreEqual
+StringAssert.Contains
+```
+
+For `Assert.*` methods, expressions are captured via `[CallerArgumentExpression]` and truncated at 50 characters. For `CollectionAssert` and `StringAssert` (legacy APIs without expression capture), the method name alone is shown.
+
+#### Omitted Parameters
+
+When overloads accept additional parameters not captured by `CallerArgumentExpression` (such as `delta`, `ignoreCase`, `culture`), the call site uses a trailing `...` to signal that the displayed signature is abbreviated:
```
nohwnd, line 62: this loses important info, and seems unncessary to do when the caller expression info is given less prominence, so it can be as long as it wants.

docs/RFCs/011-Structured-Assertion-Messages.md
```diff
@@ -0,0 +1,337 @@
...
+
+This avoids mixing runtime values with source expressions in the call site.
+
+#### Lambda Stripping
+
+For `Assert.That`, the `() =>` lambda wrapper is stripped from the call site since it is syntactic noise:
+
+```text
+// Source code: Assert.That(() => x > 10)
+// Call site:   Assert.That(x > 10)
```
nohwnd, line 77: If assert.that was a motivation for putting the callsite first, then it should be part of the failure summary. And the callsite can be repeated on the bottom to fit the same structure.

docs/RFCs/011-Structured-Assertion-Messages.md
```diff
@@ -0,0 +1,337 @@
...
+
+```text
+Assert.AreEqual(result, 42)
+The calculation returned an unexpected value
+Expected values to be equal.
+  expected: 42
+  actual:   37
+```
+
+This was a deliberate choice. Earlier iterations prefixed user messages with `Message:` or embedded them inline with the framework message. Both approaches made the user's intent harder to spot in multi-line output. Placing the user message on its own line — before the framework explanation — gives it the highest visual priority.
```
nohwnd, line 92: if it is no longer prefixed then there is nothing to search for, and harder to tell where the assertion starts and ends, it also does not have the highest visual priority when it is second line after the callsite.

docs/RFCs/011-Structured-Assertion-Messages.md
```diff
@@ -0,0 +1,337 @@
...
+
+### Line 3: Framework Message
+
+The framework's explanation of what was expected. All messages follow the tone **"Expected [subject] to [verb phrase]."**:
```
nohwnd, line 96: this is not met for some of the string assertions. they start with: `String lengths are both`

docs/RFCs/011-Structured-Assertion-Messages.md
```diff
@@ -0,0 +1,337 @@
...
+This tone was chosen after evaluating several alternatives:
+
+| Style | Example | Verdict |
+| ----- | ------- | ------- |
+| Passive: "String does not match..." | `String does not contain the expected substring.` | Rejected — describes outcome, not expectation |
+| Factual: "Wrong exception type was thrown." | `No exception was thrown.` | Rejected — not actionable |
+| Active nominal: "Expected a non-null value." | `Expected a positive value.` | Rejected — inconsistent structure with parameterized variants |
+| **Active verbal: "Expected [X] to [Y]."** | `Expected value to be null.` | **Chosen** — consistent, actionable, parameterizable |
+
+The verbal form scales naturally to parameterized messages like `Expected value {0} to be greater than {1}.` and negative forms like `Expected value to not be null.`
```
nohwnd, line 115: > The verbal form scales naturally to parameterized messages like `Expected value {0} to be greater than {1}.` and negative forms like `Expected value to not be null.`


The form should specify where will a description of the `expected` and `actual` value go. Because they should be consistent, and based on how the sentence is structured and based on the assertion it might be little unnatural for some assertions.

E.g. in Pester we use `Expected <expected value>, because <user message>, but got <actual value>.`

While here the descriptions in many assertions are: 

`Expected <actual value description> to <expected value description>`
`Expected <actual value description> to <expected value description>, but <actual value>`.

Which I guess is fine, as long as it is consistent. 

```
Expected value 3 to be in range [5, 10].
Expected string to start with the specified prefix.
Expected value to be an instance of <System.String>.
Expected collection to have 0 item(s) but found 1.
Expected value to not be null.
```

But also others that don't follow the same pattern and put expectation right after Expected:
```
Expected exactly one item to match the predicate but found 3 item(s).*
Expected the specified exception type to be thrown.
```

docs/RFCs/011-Structured-Assertion-Messages.md
Evangelink: A couple of points:

1. Often time the user message is seen by users as more important than ours (we also had requests about hiding ours) so I think it needs to come first.

1. callsite is already part of the stack trace so I don't see the value of adding it again.

1. We are limiting every parameter length so you won't end up with `Assert.Contains([ 300 lines of output]`

1. In some other comment, you are mentioning ability to search for errors and to visually identify the begining of the error. I don't see how `Assert.X failed` is easier than `Assert.X(...)`. I could understand having a generic prefix (e.g. `Assert failed`) that could always be looked up at.

nohwnd: 1. People don't tend to go praise changes they consider standard. Few people asking for hiding our message is probably not a good indicator that we should do this change for everyone. Assertion suites of other frameworks don't do this, fluent assertions does not do this.

2. the expression is not part of stack trace in the nice way that you can write imho. stack trace shows the names from assembly metadata, it does not show the variables that were used to call the method.

3. limiting the length is only necessary because the info is too prominent. If you move it down, you can keep it as whole, and info won't be lost when actually needed. The problem is not that there is lot of info on the screen, but that the important info is buried in the middle of the long output. This is why I would like short summary without expanded values on the top. Maybe shortening pages of text to just the 300 relevant lines, but otherwise as much info as we can get.

4. This was comment to the user message being first, and assumed that the callsite would move down. So the assertion would start with user message. Which is hard to find, unlike looking for `Assert.` No strong opinion here, it was just an observation.

Evangelink: Fluent assertions messages are the worst ever created, I have always hated them a lot :)

Evangelink: I have asked about 20 people internally (I know it's small number) and it seems that the consensus is that user message should be the most important and displayed before our "summary message".

nohwnd: Imho would be nice to put together 2 complete examples and make a public poll. I don't have strong opinion on whether the user message should go first or second. Other than that it makes it harder to know where the whole assertion starts when you see it in CI. But I also rarely use custom messages if there is enough variety of assertions that allow me to express what I want (e.g. I don't have to implement custom "collection contains string that matches this pattern, by combining assert true and regex).

nohwnd: There will always be different opinions (as we can see here), this form is totally fine for me:

```
<summary without expanded values, unless the values are really short, with support for multiline>
<user message>
expected: <expected>
actual: <actual>
<any other info specific to the assertion>

<callsite>
<stack trace>
```

PR #7454, opened 2026-02-22
Delete TRX reordering logic from TestingPlatformAutoRegisteredExtensions

Youssef1313:
I cannot see a reason why we need this reordering.


Evangelink (commented):
It definitely was done on purpose but I don't recall the use case. Maybe it's now un-needed because of other refactorings.

@MarcoRossignoli do you recall why it was done?

PR #7460, opened 2026-02-24
Use WaitForExitAsync from polyfill

SimonCropp:

### DIFF
--- a/src/Platform/Microsoft.Testing.Extensions.HangDump/Helpers/IProcessExtensions.cs
+++ b/src/Platform/Microsoft.Testing.Extensions.HangDump/Helpers/IProcessExtensions.cs
@@ -119,7 +119,8 @@ internal static class IProcessExtensions
 
         int timeout = 5_000;
         // This will read the output streams till the end.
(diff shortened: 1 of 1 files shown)


src/Platform/Microsoft.Testing.Extensions.HangDump/Helpers/IProcessExtensions.cs
```diff
@@ -118,17 +118,9 @@ internal static async Task<int> GetParentPidMacOsAsync(Process process, ILogger
         ps.BeginErrorReadLine();
 
         int timeout = 5_000;
-#if !NETSTANDARD2_0
```
Youssef1313, line 121: @nohwnd Do we have a bug here for the .NET Standard code path that we don't wait for exit?

nohwnd: discusses offline, overlooked the #else part. 

In both cases we wait for exit.

nohwnd (approved):

PR #7461, opened 2026-02-24
Use XDocument.LoadAsync from polyfill

SimonCropp:

### DIFF
--- a/eng/Version.Details.xml
+++ b/eng/Version.Details.xml
@@ -1,29 +1,29 @@
 <?xml version="1.0" encoding="utf-8"?>
 <Dependencies>
   <ToolsetDependencies>
-    <Dependency Name="Microsoft.DotNet.Arcade.Sdk" Version="11.0.0-beta.26123.4">
+    <Dependency Name="Microsoft.DotNet.Arcade.Sdk" Version="11.0.0-beta.26125.2">
       <Uri>https://github.com/dotnet/arcade</Uri>
-      <Sha>0831f9c135e88dd09993903ed5ac1a950285ac96</Sha>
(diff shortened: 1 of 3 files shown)


src/Platform/Microsoft.Testing.Extensions.VSTestBridge/TestHostControllers/RunSettingsEnvironmentVariableProvider.cs
```diff
@@ -69,12 +69,7 @@ public async Task<bool> IsEnabledAsync()
         if (runSettingsFilePath is not null)
         {
             using IFileStream fileStream = _fileSystem.NewFileStream(runSettingsFilePath, FileMode.Open, FileAccess.Read);
-#if NETCOREAPP
             _runSettings = await XDocument.LoadAsync(fileStream.Stream, LoadOptions.None, CancellationToken.None).ConfigureAwait(false);
```
Youssef1313, line 72: Looks like the polyfill implementation here is fully synchronous?

https://github.com/SimonCropp/Polyfill/blob/41a7eaab9c7b2d66093f99807a5479d9b9ed5d0a/src/Polyfill/XDocumentPolyfill.cs#L34

That sounds different a bit from the current implementation.

Evangelink: @SimonCropp ^^^

SimonCropp:
i pushed a new polyfill

but travelling this weekend so might not get to to looking at this for a few days

feel free to queue up any comments

PR #7483, opened 2026-03-03
Cleanup TrxReportEngine

Youssef1313:


src/Platform/Microsoft.Testing.Extensions.TrxReport/TrxReportEngine.cs
```diff
@@ -391,8 +380,12 @@ private static void AddTestLists(XElement testRun, string uncategorizedTestId)
         testRun.Add(testLists);
     }
 
-    private void AddResults(string testAppModule, XElement testRun, out XElement testDefinitions, out XElement testEntries, out string uncategorizedTestId, out bool hasFailedTests)
+    private (int Passed, int Failed, int Skipped, int Timedout) AddResults(TestNodeUpdateMessage[] testNodeUpdateMessages, string testAppModule, XElement testRun, out XElement testDefinitions, out XElement testEntries, out string uncategorizedTestId, out bool hasFailedTests)
```
nohwnd: Subjective but I would use a class / struct.

PR #7491, opened 2026-03-04
Cleanup TrxReportEngine

Youssef1313:

### DIFF
--- a/test/IntegrationTests/Microsoft.Testing.Platform.Acceptance.IntegrationTests/TrxTests.cs
+++ b/test/IntegrationTests/Microsoft.Testing.Platform.Acceptance.IntegrationTests/TrxTests.cs
@@ -513,6 +513,8 @@ public class DummyTestFramework : ITestFramework, IDataProducer
 
     public async Task ExecuteRequestAsync(ExecuteRequestContext context)
     {
+        var testMethodIdentifier = new TestMethodIdentifierProperty(string.Empty, string.Empty, "DummyClassName", "Test", 0, Array.Empty<string>(), string.Empty);
+
         await context.MessageBus.PublishAsync(this, new TestNodeUpdateMessage(context.Request.Session.SessionUid,
             new TestNode()
             {
@@ -520,6 +522,7 @@ public class DummyTestFramework : ITestFramework, IDataProducer
                 DisplayName = "FailingTest",
                 Properties = new PropertyBag(
                     new FailedTestNodeStateProperty("Expected 1 but got 2"),
+                    testMethodIdentifier,
                     new TrxExceptionProperty("Expected 1 but got 2", "   at DummyTestFramework.ExecuteRequestAsync() in Program.cs:line 50"))
             }));
         context.Complete();


src/Platform/Microsoft.Testing.Extensions.TrxReport/TrxDataConsumer.cs
```diff
@@ -106,38 +106,31 @@ public Task ConsumeAsync(IDataProducer dataProducer, IData value, CancellationTo
...
-                    }
-                    else
-                    {
-                        sessionFileArtifacts.Add(fileArtifact);
-                    }
-
-                    break;
-            }
-        }
-        catch (OperationCanceledException ex) when (ex.CancellationToken == cancellationToken)
```
drognanar, line 138: why do we no longer need to catch the exception? is the caller meant to be handling cancellation

src/Platform/Microsoft.Testing.Extensions.TrxReport/TrxReportEngine.cs
```diff
@@ -277,6 +292,10 @@ private async Task AddResultSummaryAsync(XElement testRun, string resultSummaryO
             new XAttribute("pending", 0));
         resultSummary.Add(counters);
 
+        // TODO: VSTest adds two additional things to RunInfos
```
drognanar, line 295: is there an overarching issue to fix these todos?

src/Platform/Microsoft.Testing.Extensions.TrxReport/TrxReportEngine.cs
```diff
@@ -547,121 +579,45 @@ private SummaryCounts AddResults(TestNodeUpdateMessage[] testNodeUpdateMessages,
             results.Add(unitTestResult);
 
             // TestDefinitions
-            var unitTest = new XElement(
```
drognanar, line 550: why is this no longer needed?

drognanar: or was it extracted to a separate method?

src/Platform/Microsoft.Testing.Extensions.TrxReport/TrxDataConsumer.cs
nohwnd: Looks like line 107 throws the same exception, so the parent code must handle it already or let it bubble up.

Youssef1313: Yup. Caller is supposed to handle it.

src/Platform/Microsoft.Testing.Extensions.TrxReport/TrxReportEngine.cs
Youssef1313: Not yet. I'm planning to follow-up directly with fixes/investigations on those after merging this PR. If I cannot get them done in the near future I'll create an issue.

src/Platform/Microsoft.Testing.Extensions.TrxReport/TrxReportEngine.cs
Youssef1313: Extracted to a separate method.

PR #7496, opened 2026-03-06
Refactor OTel handling out of TestApplicationResult

Youssef1313:


src/Platform/Microsoft.Testing.Platform/Telemetry/OpenTelemetryResultHandler.cs
```diff
@@ -0,0 +1,172 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using Microsoft.Testing.Platform.Extensions.Messages;
+using Microsoft.Testing.Platform.Helpers;
+
+namespace Microsoft.Testing.Platform.Telemetry;
+
+internal sealed class OpenTelemetryResultHandler
```
Evangelink, line 9: This newly extracted class has no unit tests. The existing `TestApplicationResultTests.cs` always passes `null` for the `otelService` parameter, meaning the entire OTel path is untested.

At minimum, tests should cover:
- `NotifyInProgress` → `NotifyPassed`/`NotifyFailed`/`NotifySkipped` lifecycle (activity creation, tag setting, disposal)
- `NotifyInProgress` when `StartActivity` returns `null` (activity not tracked)
- `NotifyDiscovered` / `NotifyUnknown` (counter increments)
- `HandleTestResult` when no matching activity exists (early return path)

src/Platform/Microsoft.Testing.Platform/Telemetry/OpenTelemetryResultHandler.cs
```diff
@@ -0,0 +1,172 @@
...
+    private readonly IEnvironment _environment;
+    private readonly ICounter<int> _totalDiscoveredTests;
+    private readonly ICounter<int> _totalStartedTests;
+    private readonly ICounter<int> _totalCompletedTests;
+    private readonly ICounter<int> _totalPassedTests;
+    private readonly ICounter<int> _totalFailedTests;
+    private readonly ICounter<int> _totalSkippedTests;
+    private readonly ICounter<int> _totalUnknownedTests;
+    private readonly IHistogram<double> _totalDuration;
+    private readonly Dictionary<TestNodeUid, IPlatformActivity> _testActivities = [];
```
Evangelink, line 21: `_testActivities` holds `IPlatformActivity` objects (which extend `IDisposable`). If a test reports `InProgress` but never reaches a terminal state (crash, abort, bug in state reporting), those activities remain undisposed.

I know this is a pre-existing issue (the old inline code had the same problem), but now that `OpenTelemetryResultHandler` clearly owns the `_testActivities` dictionary, it would be a good opportunity to implement `IDisposable` and dispose all remaining entries in cleanup.

src/Platform/Microsoft.Testing.Platform/Telemetry/OpenTelemetryResultHandler.cs
```diff
@@ -0,0 +1,172 @@
...
+{
+    private readonly IPlatformOpenTelemetryService _otelService;
+    private readonly IEnvironment _environment;
+    private readonly ICounter<int> _totalDiscoveredTests;
+    private readonly ICounter<int> _totalStartedTests;
+    private readonly ICounter<int> _totalCompletedTests;
+    private readonly ICounter<int> _totalPassedTests;
+    private readonly ICounter<int> _totalFailedTests;
+    private readonly ICounter<int> _totalSkippedTests;
+    private readonly ICounter<int> _totalUnknownedTests;
```
Evangelink, line 19: Pre-existing typo: "Unknowned" is not a word — should be `_totalUnknownTests`. The metric name `"tests.unknown"` is correct; only the C# field name has this issue. Since this is a brand new file, it's a good time to fix it.

src/Platform/Microsoft.Testing.Platform/Telemetry/OpenTelemetryResultHandler.cs
```diff
@@ -0,0 +1,172 @@
...
+using Microsoft.Testing.Platform.Helpers;
+
+namespace Microsoft.Testing.Platform.Telemetry;
+
+internal sealed class OpenTelemetryResultHandler
+{
+    private readonly IPlatformOpenTelemetryService _otelService;
+    private readonly IEnvironment _environment;
+    private readonly ICounter<int> _totalDiscoveredTests;
+    private readonly ICounter<int> _totalStartedTests;
```
Evangelink, line 14: Nit: this counter is write-only (fire-and-forget telemetry), which is fine by design. But given there are no tests for any of the counters, the risk of silent regression from this refactor is elevated. Reinforces the need for unit tests on this class.

src/Platform/Microsoft.Testing.Platform/Telemetry/OpenTelemetryResultHandler.cs
```diff
@@ -0,0 +1,172 @@
...
+    }
+
+    internal void NotifyDiscovered()
+        => _totalDiscoveredTests.Add(1);
+
+    internal void NotifyPassed(TestNode testNode, TestNodeStateProperty stateProperty)
+    {
+        _totalPassedTests.Add(1);
+        _totalCompletedTests.Add(1);
+        HandleTestResult(testNode, stateProperty);
```
Evangelink, line 44: Nit: `NotifyPassed`, `NotifyFailed`, and `NotifySkipped` all internally call `_totalCompletedTests.Add(1)`. If a new terminal state is added in the future, the developer must remember to include this call. A small shared helper or an explicit comment could reduce this risk.

src/Platform/Microsoft.Testing.Platform/Telemetry/OpenTelemetryResultHandler.cs
Youssef1313: Pre-existing issue. Moving forward with the current refactoring and will fix later.

src/Platform/Microsoft.Testing.Platform/Telemetry/OpenTelemetryResultHandler.cs
Youssef1313: Pre-existing issue. Moving forward with the current refactoring and will fix later.

src/Platform/Microsoft.Testing.Platform/Telemetry/OpenTelemetryResultHandler.cs
Youssef1313: Pre-existing issue. Moving forward with the current refactoring and will fix later.

src/Platform/Microsoft.Testing.Platform/Telemetry/OpenTelemetryResultHandler.cs
Youssef1313: Pre-existing issue. Moving forward with the current refactoring and will fix later.

src/Platform/Microsoft.Testing.Platform/Telemetry/OpenTelemetryResultHandler.cs
Youssef1313: Pre-existing issue. Moving forward with the current refactoring and will fix later.

PR #7506, opened 2026-03-08
Use MoveFile and remove CopyFile/DeleteFile

SimonCropp:
Unify file-moving logic by removing the NETCOREAPP conditional and calling IFileSystem.MoveFile unconditionally in RetryOrchestrator. Remove CopyFile and DeleteFile members from IFileSystem and their implementations in SystemFileSystem, simplifying the file system abstraction.

### DIFF
--- a/src/Platform/Microsoft.Testing.Platform/Helpers/System/IFileSystem.cs
+++ b/src/Platform/Microsoft.Testing.Platform/Helpers/System/IFileSystem.cs
@@ -21,9 +21,5 @@ internal interface IFileSystem
 
     Task<string> ReadAllTextAsync(string path);
 
-    void CopyFile(string sourceFileName, string destFileName, bool overwrite = false);
-
-    void DeleteFile(string path);
-
     string[] GetFiles(string path, string searchPattern, SearchOption searchOption);
 }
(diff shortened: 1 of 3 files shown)


src/Platform/Microsoft.Testing.Platform/Helpers/System/IFileSystem.cs
```diff
@@ -21,9 +21,5 @@ internal interface IFileSystem
 
     Task<string> ReadAllTextAsync(string path);
 
-    void CopyFile(string sourceFileName, string destFileName, bool overwrite = false);
-
-    void DeleteFile(string path);
```
Youssef1313, line 26: This is a binary breaking change. It's fine to simplify the RetryOrchestrator, but we shouldn't remove these members from the interface, at least for few minor versions.

Youssef1313: We should also add a comment explicitly stating that while these are unused, they are needed so that older versions of Retry can remain compatible with newer versions of MTP.

SimonCropp: my understanding is that all nugets are deployed at the same time. given that they must have a min version on the core. so it should be safe to remove internal APIs. unless there are external projects that use internals and are deployed in an out of sync way?

Youssef1313: We deploy at the same time, and we have min version. But given that the version constraint is "min" and not "strictly equal", this is problematic.

Consider:

1. Project uses MSTest 4.1 (bringing MTP 2.1)
2. Project uses Retry 2.1
3. Project uses TRX 2.1

Then after our next release, if the user updated TRX to 2.2. They will hit an issue.
The Retry version being used is 2.1, while the core MTP used is 2.2 (brought via TRX). Retry 2.1 will attempt to access CopyFile/DeleteFile, but these will not be present at runtime in 2.2.

Evangelink:
Moving forward by closing this PR.

PR #7507, opened 2026-03-08
NamedPipeServer callbacks do no async work. so make the signature sync

SimonCropp:


src/Platform/Microsoft.Testing.Extensions.HangDump/HangDumpActivityIndicator.cs
```diff
@@ -112,16 +112,16 @@ await _namedPipeClient.RequestReplyAsync<ConsumerPipeNameRequest, VoidResponse>(
         }
     }
 
-    private async Task<IResponse> CallbackAsync(IRequest request)
+    private IResponse Callback(IRequest request)
     {
         if (request is GetInProgressTestsRequest)
         {
-            await _logger.LogDebugAsync($"Received '{nameof(GetInProgressTestsRequest)}'").ConfigureAwait(false);
+            _logger.LogDebug($"Received '{nameof(GetInProgressTestsRequest)}'");
```
Youssef1313, line 119: This is potentially running on a threadpool thread, and the logger implementation is very likely synchronized via a semaphore. Waiting on the semaphore synchronously might lead to blocking threadpool thread in case there is some contention in logging, and I think this is a risky change to make.

SimonCropp: does that mean the sync log APIs should be removed? and all callers of them made async?

Youssef1313: I wouldn't say it should be removed, but we could revise usages to switch to async when it makes sense.

For example, if we are running something in a custom non-threadpool thread, the sync version makes sense, and (depending on the exact work done on that thread) might even be preferred over the async variant. But in other cases, the async version will make more sense (especially when it's something already running on a threadpool thread)

PR #7508, opened 2026-03-08
remove redundant null checks

SimonCropp:

### DIFF
--- a/src/Adapter/MSTestAdapter.PlatformServices/Services/TestContextImplementation.cs
+++ b/src/Adapter/MSTestAdapter.PlatformServices/Services/TestContextImplementation.cs
@@ -255,12 +255,6 @@ internal sealed class TestContextImplementation : TestContext, ITestContext, IDi
     /// <returns>True if found.</returns>
     public bool TryGetPropertyValue(string propertyName, out object? propertyValue)
     {
-        if (_properties == null)
-        {
-            propertyValue = null;
-            return false;
-        }
-
         return _properties.TryGetValue(propertyName, out propertyValue);
     }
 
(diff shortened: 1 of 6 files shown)


src/Platform/Microsoft.Testing.Platform/Telemetry/TelemetryManager.cs
```diff
@@ -101,12 +101,9 @@ private async Task ShowTelemetryBannerFirstNoticeAsync(ServiceProvider servicePr
 
         ITestApplicationModuleInfo testApplicationModuleInfo = serviceProvider.GetTestApplicationModuleInfo();
 #pragma warning disable RS0030 // Do not use banned APIs - There is no easy way to disable it for all members
-        string? directory = environment.GetFolderPath(Environment.SpecialFolder.LocalApplicationData, Environment.SpecialFolderOption.Create);
```
Youssef1313, line 104: This one I'm not sure about. Some tests might be mocking stuff in a way that breaks this. If we can write the tests in a better way, that could be a good option. Otherwise, we probably keep this one for now.

But other than that, the rest of the changes look very good!

src/Adapter/MSTestAdapter.PlatformServices/Services/TestContextImplementation.cs
```diff
@@ -255,12 +255,6 @@ public void SetDataConnection(object? dbConnection)
     /// <returns>True if found.</returns>
     public bool TryGetPropertyValue(string propertyName, out object? propertyValue)
     {
-        if (_properties == null)
-        {
-            propertyValue = null;
-            return false;
-        }
-
         return _properties.TryGetValue(propertyName, out propertyValue);
```
Youssef1313, line 258: > error IDE0022: (NETCORE_ENGINEERING_TELEMETRY=Build) Use expression body for method

PR #7511, opened 2026-03-09
Simplify retry extension logic to use built-in filters

### LINKED ISSUE #7500: Simplify retry extension
In future, we should be able to simplify this. We don't need a custom filter at all. Instead, we can use `--filter-uid` and pass all the UIDs. We need to be very careful though when the list is long, in which case we might need to create a response file for that. And in that case we should carefully decide where to write such file to not leak any info to other potential users on the same system.

_Originally posted by @Youssef1313 in https://github.com/microsoft/testfx/pull/7409#discussion_r2895833407_

Evangelink:
Fixes #7500


src/Platform/Microsoft.Testing.Extensions.Retry/RetryExtensions.cs
```diff
@@ -38,14 +37,5 @@ CompositeExtensionFactory<RetryDataConsumer> compositeExtensionFactory
 
         builder.TestHostOrchestrator
             .AddTestHostOrchestrator(serviceProvider => new RetryOrchestrator(serviceProvider));
-
-        if (builder.TestHost is not TestHostManager testHostManager)
-        {
-            throw new InvalidOperationException(
-                ExtensionResources.RetryProviderRequiresDefaultTestHostManagerErrorMessage);
```
Youssef1313, line 49: Can we delete this resource?

src/Platform/Microsoft.Testing.Extensions.Retry/RetryOrchestrator.cs
```diff
@@ -170,6 +170,15 @@ public async Task<int> OrchestrateTestHostExecutionAsync(CancellationToken cance
             finalArguments.Add($"--{RetryCommandLineOptionsProvider.RetryFailedTestsPipeNameOptionName}");
             finalArguments.Add(retryFailedTestsPipeServer.PipeName);
 
+            // When retrying, replace any existing test filter with --filter-uid for the failed tests
+            if (lastListOfFailedId is { Length: > 0 })
+            {
+                RemoveOption(finalArguments, TreeNodeFilterCommandLineOptionsProvider.TreenodeFilter);
+                RemoveOption(finalArguments, PlatformCommandLineProvider.FilterUidOptionKey);
```
Youssef1313, line 177: I'm not sure how accurate this is.

In theory, I think we can have `--filter-uid` as a value of some other option. While this is really not a realistic use case, I think it's possible in theory. Do we care?

src/Platform/Microsoft.Testing.Extensions.Retry/RetryOrchestrator.cs
```diff
@@ -170,6 +170,15 @@ public async Task<int> OrchestrateTestHostExecutionAsync(CancellationToken cance
             finalArguments.Add($"--{RetryCommandLineOptionsProvider.RetryFailedTestsPipeNameOptionName}");
             finalArguments.Add(retryFailedTestsPipeServer.PipeName);
 
+            // When retrying, replace any existing test filter with --filter-uid for the failed tests
+            if (lastListOfFailedId is { Length: > 0 })
```
Youssef1313, line 174: If this is large enough, it might exceed the command-line length limit.

src/Platform/Microsoft.Testing.Extensions.Retry/RetryOrchestrator.cs
```diff
@@ -170,6 +170,51 @@ public async Task<int> OrchestrateTestHostExecutionAsync(CancellationToken cance
...
+                    // Write to retryRootFolder (not the per-attempt folder) so it won't be included
+                    // in the final results move.
+                    string responseFilePath = Path.Combine(retryRootFolder, $"retry-filter-uids-{attemptCount}.rsp");
+                    using (IFileStream stream = _fileSystem.NewFileStream(responseFilePath, FileMode.Create, FileAccess.Write))
+                    using (var writer = new StreamWriter(stream.Stream))
+                    {
+                        await writer.WriteAsync($"--{PlatformCommandLineProvider.FilterUidOptionKey}").ConfigureAwait(false);
+                        foreach (string uid in lastListOfFailedId)
+                        {
+                            await writer.WriteAsync($" \"{uid}\"").ConfigureAwait(false);
```
Youssef1313: I think we need to verify what are the escaping rules from our rsp parsing implementation and if we need to escape some stuff if the uid contains them

test/IntegrationTests/Microsoft.Testing.Platform.Acceptance.IntegrationTests/RetryFailedTestsTests.cs
```diff
@@ -306,46 +340,60 @@ public async Task ExecuteRequestAsync(ExecuteRequestContext context)
         string resultDir = Environment.GetEnvironmentVariable("RESULTDIR")!;
         bool crash = Environment.GetEnvironmentVariable("CRASH") == "1";
 
+        var uidFilter = (context.Request as TestExecutionRequest)?.Filter as TestNodeUidListFilter;
```
Youssef1313, line 339: I'm wondering how did this test work before? How was it respecting the retry filter?

src/Platform/Microsoft.Testing.Extensions.Retry/RetryOrchestrator.cs
```diff
@@ -170,6 +170,51 @@ public async Task<int> OrchestrateTestHostExecutionAsync(CancellationToken cance
...
+                        }
+                    }
+
+                    finalArguments.Add($"@{responseFilePath}");
```
Youssef1313, line 214: If this path has a space, we might fail.

We need proper escaping.

Youssef1313: Hmm, actually not, that's handled. `finalArguments` is the unescaped arguments which will either be handled by runtime or by PasteArguments.

It might still be worth to add a test

test/IntegrationTests/Microsoft.Testing.Platform.Acceptance.IntegrationTests/RetryFailedTestsTests.cs
Evangelink: Good question. Before this PR, the dummy test framework's `ExecuteRequestAsync` ran all 3 test methods unconditionally — it never checked `context.Request.Filter`. The old `RetryExecutionFilterFactory` did create a `TestNodeUidListFilter` and attach it to the request, but the dummy framework simply ignored it.

The retry tests still passed because:
1. First attempt: all 3 tests run, test 1 fails (env var controls failure)
2. Retry: `RetryExecutionFilterFactory` sets a UID filter for just UID 1, but the dummy framework ignores it and runs all 3 tests again. Test 1 now passes (state file/env var was set in attempt 1).
3. The test only asserted overall success and attempt count, not that only the failed test was re-run.

So the retry was technically **not filtering correctly** — it re-ran all tests on every retry, which was wasteful but happened to produce the right overall result. This PR fixes that by using `--filter-uid` on the command line (handled by the platform) and having the test asset explicitly check the filter via `IsIncluded`.

src/Platform/Microsoft.Testing.Extensions.Retry/RetryOrchestrator.cs
Evangelink: Confirmed — as you noted in your follow-up, the `@{responseFilePath}` string goes into `finalArguments` which is the unescaped argument list. On .NET 8+, it's passed to `ProcessStartInfo.ArgumentList` (runtime handles quoting). On older TFMs, it goes through `PasteArguments.Paste()` which properly quotes arguments containing spaces. So paths with spaces are correctly handled in both paths.

src/Platform/Microsoft.Testing.Extensions.Retry/RetryOrchestrator.cs
Evangelink: I investigated `ResponseFileHelper.SplitCommandLine` — here are the escaping rules:

- The RSP parser splits each line by whitespace and uses `"` for grouping (like shell quoting).
- `CurrentToken()` calls `.Replace("\"", string.Empty)` — all `"` characters are stripped from the resulting token.
- Lines starting with `#` are treated as comments and skipped.

**What this means for UIDs:**
- UIDs with whitespace: wrapping in `"..."` correctly groups them ✓
- UIDs starting with `#`: if on their own line without quotes, they'd be treated as a comment ✗
- UIDs containing literal `"`: the quote characters get stripped by the parser, so they can't round-trip safely ✗

**Fix applied:** Changed from writing all UIDs on a single space-separated line to writing one UID per line, each wrapped in `"`. This handles whitespace and `#`-prefixed UIDs correctly. Added a comment documenting the `"` limitation (extremely unlikely in practice since UIDs are typically type/method names or GUIDs).

Also fixed a related issue: `StartsWith('-')` uses the `string.StartsWith(char)` overload which isn't available on `netstandard2.0` — changed to `arguments[idx][0] != '-'` pattern.

src/Platform/Microsoft.Testing.Extensions.Retry/RetryOrchestrator.cs
```diff
@@ -204,10 +204,15 @@ public async Task<int> OrchestrateTestHostExecutionAsync(CancellationToken cance
                     using (IFileStream stream = _fileSystem.NewFileStream(responseFilePath, FileMode.Create, FileAccess.Write))
                     using (var writer = new StreamWriter(stream.Stream))
                     {
-                        await writer.WriteAsync($"--{PlatformCommandLineProvider.FilterUidOptionKey}").ConfigureAwait(false);
+                        // Write one UID per line. The RSP parser (ResponseFileHelper.SplitCommandLine) splits
+                        // each line by whitespace and uses '"' for grouping. Wrapping each UID in quotes
+                        // handles UIDs containing whitespace or starting with '#' (comment marker).
+                        // Note: UIDs containing literal '"' characters cannot be safely round-tripped
```
Youssef1313, line 210: If that cannot be preserved in rsp, that's a big problem.

Can we validate the UID for all major test frameworks in cases like a parameterized test that has a string input where the value passed contains double quotes?

This is not an uncommon scenario.

src/Platform/Microsoft.Testing.Extensions.Retry/RetryOrchestrator.cs
```diff
@@ -204,10 +204,15 @@ public async Task<int> OrchestrateTestHostExecutionAsync(CancellationToken cance
                     using (IFileStream stream = _fileSystem.NewFileStream(responseFilePath, FileMode.Create, FileAccess.Write))
                     using (var writer = new StreamWriter(stream.Stream))
                     {
-                        await writer.WriteAsync($"--{PlatformCommandLineProvider.FilterUidOptionKey}").ConfigureAwait(false);
+                        // Write one UID per line. The RSP parser (ResponseFileHelper.SplitCommandLine) splits
```
Youssef1313, line 207: One UID per line sounds very weird requirement. It should be valid to have all in a single line. Reading this comment, it doesn't make it any clear to me.

YuliiaKovalova (approved):
looks good to me.

PR #7520, opened 2026-03-10
Improve ordering of `ITestSessionLifetimeHandler.OnTestSessionFinishingAsync`

Youssef1313:


Youssef1313:
It should be helping more than TRX. For example:

https://github.com/spekt/testlogger/blob/e7ae9beb3365e70b7a43903df88cce3738742d0f/src/TestLogger/TestReporter.cs#L27

This is a consumer lifetime handler, so it's delayed after non-consumers. This ensures that it will consume `SessionFileArtifact` produced by CC (non-consumer) before `OnTestSessionFinishingAsync` is called.

I think in most cases, the ordering problem is solved just by making non-consumers first, without any need for extension authors to have to think about ordering which is likely to be more messy (IMO).

Evangelink (commented):
Alright, let's see :)

PR #7525, opened 2026-03-11
Make MSBuild extension more strict

### LINKED ISSUE #6954: Incompletely configured MTP in `dotnet test` VSTest *silently* doesn't run tests
## Describe the bug

If one omits the single critical step of adding e.g. `<EnableNUnitRunner>true</EnableNUnitRunner>` (for NUnit) in one's test project file, while running in vstest mode of `dotnet test` due to the absence of the `.test.runner = Microsoft.Testing.Platform` global.json config, the output of `dotnet test` is like this, with exit code 0:

```
  Determining projects to restore...
  Restored /Users/patrick/Documents/GitHub/repro2/test.fsproj (in 610 ms).
  test -> /Users/patrick/Documents/GitHub/repro2/bin/Debug/net10.0/test.dll
  Run tests: '/Users/patrick/Documents/GitHub/repro2/bin/Debug/net10.0/test.dll' [net10.0|arm64]
  Tests succeeded: '/Users/patrick/Documents/GitHub/repro2/bin/Debug/net10.0/test.dll' [net10.0|arm64]
```

If you don't *already* know that an actually-running test project will output this additional line:

```
Passed! - Failed: 0, Passed: 1, Skipped: 0, Total: 1, Duration: 125ms - test.dll (net10.0|arm64)
```

then there is nothing to tell you that you have configured your project in such a way that no tests will ever run.
(Also ideally one would not have to read the output of `dotnet test` with your human eyes to know whether your tests ran.)

This is a user error - following the docs correctly will result in a project that works - but it's a pretty hostile failure mode, and the console output "Tests succeeded" is actively a falsehood (no tests in fact succeeded).

## Steps To Reproduce

See https://github.com/Smaug123/MicrosoftTestingPlatformWeirdness/ at commit 548b60721f3e4b452b6277b6ccbeb45c1de260ba for a repro. That repo also omits `<OutputType>Exe</OutputType>` but the result is the same with that included.

```xml

  <PropertyGroup>
    <TargetFramework>net9.0</TargetFramework>
    <IsTestProject>true</IsTestProject>
    <OutputType>Exe</OutputType>
    <TestingPlatformDotnetTestSupport>true</TestingPlatformDotnetTestSupport>
  </PropertyGroup>

  <ItemGroup>
    <Compile Include="Test.fs" />
    <PackageReference Include="Microsoft.NET.Test.Sdk" Version="18.0.1" />
    <PackageReference Include="NUnit" Version="4.3.2" />
    <PackageReference Include="NUnit3TestAdapter" Version="5.2.0" />
  </ItemGroup>
</Project>
```

## Expected behavior

Either the test correctly fails; or the test runner correctly fails and informs me that there were no tests, ideally telling me why; or the build fails, informing me that no tests will ever run.

## Actual behavior

```
  Determining projects to restore...
  Restored /Users/patrick/Documents/GitHub/repro2/test.fsproj (in 610 ms).
  test -> /Users/patrick/Documents/GitHub/repro2/bin/Debug/net10.0/test.dll
  Run tests: '/Users/patrick/Documents/GitHub/repro2/bin/Debug/net10.0/test.dll' [net10.0|arm64]
  Tests succeeded: '/Users/patrick/Documents/GitHub/repro2/bin/Debug/net10.0/test.dll' [net10.0|arm64]
```

## Additional context

The answer might be "we're going to deprecate VSTest mode", which is fine, although the error message even in the `dotnet test`-MTP-mode case is pretty obscure, being this plus a link to a [page of documentation](https://learn.microsoft.com/en-gb/dotnet/core/tools/dotnet-test?tabs=dotnet-test-with-mtp) that *doesn't* contain information about how to configure the project correctly:

```
global.json defines test runner to be Microsoft.Testing.Platform. All projects must use that test runner.
The following test projects are using VSTest test runner:
test.fsproj
```

I don't know how the testing platform works well enough to say what that message ought to be.

Youssef1313:
Fixes #6954

### DIFF
--- a/src/Platform/Microsoft.Testing.Platform.MSBuild/Tasks/InvokeTestingPlatformTask.cs
+++ b/src/Platform/Microsoft.Testing.Platform.MSBuild/Tasks/InvokeTestingPlatformTask.cs
@@ -400,6 +400,7 @@ public class InvokeTestingPlatformTask : Build.Utilities.ToolTask, IDisposable
                 Log.LogError(Resources.MSBuildResources.DidNotReceiveRunSummaryInfo, TargetPath.ItemSpec.Trim());
                 return false;
             }
+
             Log.LogMessage(MessageImportance.High, Resources.MSBuildResources.TestsSucceeded, TargetPath.ItemSpec.Trim(), TargetFramework.ItemSpec, TestArchitecture.ItemSpec);
         }
 


src/Platform/Microsoft.Testing.Platform.MSBuild/buildMultiTargeting/Microsoft.Testing.Platform.MSBuild.VSTest.targets
```diff
@@ -1,6 +1,7 @@
 <Project>
   <!-- Override VSTest target for single tfm -->
   <Target Name="VSTest">
-    <CallTarget Targets="InvokeTestingPlatform" />
+    <Error Text="'TestingPlatformDotnetTestSupport' was set to true, but the project did not opt-in MTP. If this is not a test project" Condition="'$(IsTestingPlatformApplication)' != 'true'" />
```
Evangelink: It could be interesting to point out to our doc, maybe create a new entry in FAQ about this so there is something actionable.

Also the last part seems to be missing ending "If this is not a test project"

src/Platform/Microsoft.Testing.Platform.MSBuild/Tasks/InvokeTestingPlatformTask.cs
```diff
@@ -386,7 +387,17 @@ public override bool Execute()
...
+        if (_moduleInfo is null)
+        {
+            Log.LogError(Resources.MSBuildResources.DidNotReceiveModuleInfo, TargetPath.ItemSpec.Trim());
+            return false;
+        }
+        else if (!_receivedRunSummaryInfoRequest)
+        {
+            Log.LogError(Resources.MSBuildResources.DidNotReceiveRunSummaryInfo, TargetPath.ItemSpec.Trim());
+            return false;
+        }
```
Evangelink: Isn't this adding potentially confusing messages if the return value is false?

Youssef1313: Why do you think it's confusing messages?

src/Platform/Microsoft.Testing.Platform.MSBuild/buildMultiTargeting/Microsoft.Testing.Platform.MSBuild.VSTest.targets
Youssef1313: `https://learn.microsoft.com/dotnet/core/testing/migrating-vstest-microsoft-testing-platform` might be the best page to point at.

Youssef1313:
```suggestion
    <Error Text="'TestingPlatformDotnetTestSupport' was set to true, but the project did not opt-in MTP. If this is not a test project, set 'IsTestingPlatformApplication' to false. For more information, see https://aka.ms/migrate-vstest-mtp" Condition="'$(IsTestingPlatformApplication)' != 'true'" />
```

src/Platform/Microsoft.Testing.Platform.MSBuild/Tasks/InvokeTestingPlatformTask.cs
```diff
@@ -386,7 +387,17 @@ public override bool Execute()
...
+        else if (returnValue)
         {
```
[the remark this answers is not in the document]
Youssef1313: @Evangelink Is Copilot suggestion here how you want it to be? I'm fine with it.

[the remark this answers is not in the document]
Evangelink: Yes! I was wondering if in case of failures (return value is false) we would end up displaying some "error message" not directly related to the current issue.

PR #7535, opened 2026-03-12
Simplify DynamicData reflection lookups using FlattenHierarchy

Evangelink:
Replace the manual `DeclaredOnly` + `BaseType` walk in `DynamicDataOperations` with direct reflection lookups using `BindingFlags.FlattenHierarchy`, and inline the helper methods at their call sites.

**Behavior change:** `FlattenHierarchy` does not return private static members from base types, while the previous loop-based approach did. This is accepted as a reasonable tradeoff — accessing private members of a base class via `[DynamicData]` is an unusual pattern.

The `DynamicDataShouldBeValidAnalyzer` has been updated to match: `TryGetMemberCore` now filters out private members when looking at base types.


Youssef1313 (commented):
- Have we seen real perf issues here that warrants adding a cache?
- I feel like walking the inheritance hierarchy is already not the right approach. Instead, we should avoid `BindingFlags.DeclaredOnly` and let the .NET runtime do its thing. The .NET reflection implementation has some [caching](https://github.com/dotnet/runtime/blob/b36efcf356fad713b23767110792eee9968d2985/src/coreclr/System.Private.CoreLib/src/System/RuntimeType.CoreCLR.cs#L2577) already, so if we are able to avoid walking the hierarchy ourselves and let the caching behavior of reflection do its job, I think that's the best.

src/TestFramework/TestFramework/Attributes/DataSource/DynamicDataOperations.cs
```diff
@@ -5,7 +5,7 @@ namespace Microsoft.VisualStudio.TestTools.UnitTesting;
 
 internal static class DynamicDataOperations
 {
-    private const BindingFlags DeclaredOnlyLookup = BindingFlags.Public | BindingFlags.NonPublic | BindingFlags.Instance | BindingFlags.Static | BindingFlags.DeclaredOnly;
+    private const BindingFlags MemberLookup = BindingFlags.Public | BindingFlags.NonPublic | BindingFlags.Instance | BindingFlags.Static | BindingFlags.FlattenHierarchy;
```
Youssef1313, line 8: I guess the `FlatternHierarchy` flag shouldn't be needed.

src/TestFramework/TestFramework/Attributes/DataSource/DynamicDataOperations.cs
```diff
@@ -167,56 +167,11 @@ private static bool TryGetData(object dataSource, [NotNullWhen(true)] out IEnume
     }
 
     private static FieldInfo? GetFieldConsideringInheritance(Type type, string fieldName)
```
Youssef1313: Maybe we can inline this now as it's only a single line?

Youssef1313 (approved):
LGTM, with two small comments.

src/TestFramework/TestFramework/Attributes/DataSource/DynamicDataOperations.cs
Evangelink: `FlattenHierarchy` is actually needed here. Without it, `GetProperty`/`GetMethod`/`GetField` with `BindingFlags.Static` will **not** return inherited static members — .NET's `RuntimeType.FilterApplyBase` explicitly filters out inherited static members unless `FlattenHierarchy` is set. Since dynamic data sources must be static (validated in `GetDataFromProperty`, `GetDataFromMethod`, `GetDataFromField`), removing this flag would break the case where a test class inherits a static data source from a base class.

One thing worth noting: there's a subtle behavior change vs. the old manual `DeclaredOnly` + loop approach — `FlattenHierarchy` does **not** return **private** static members from base classes, while the old loop did. This seems acceptable since accessing private members of a base class is an unusual pattern, but flagging it in case you think it matters.

src/TestFramework/TestFramework/Attributes/DataSource/DynamicDataOperations.cs
Evangelink: Good call — inlined all three helper methods at their call sites and removed the methods. 👍

src/TestFramework/TestFramework/Attributes/DataSource/DynamicDataOperations.cs
Youssef1313: I think DynamicData isn't interested in static members anyways.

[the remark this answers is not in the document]
Youssef1313: I'm fine with the behavior change. Let's ensure that the analyzer that validates dynamic data handles it correctly though.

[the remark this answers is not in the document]
Youssef1313: In `DynamicDataShouldBeValidAnalyzer`, we have `TryGetMemberCore` local function. We can possibly add `bool disallowPrivate` there, and we pass `false` in the first iteration, and true on subsequent iterations.

src/TestFramework/TestFramework/Attributes/DataSource/DynamicDataOperations.cs
```diff
@@ -166,57 +166,4 @@ private static bool TryGetData(object dataSource, [NotNullWhen(true)] out IEnume
         return false;
     }
```
Youssef1313:
```suggestion
```

Youssef1313 (approved):

src/TestFramework/TestFramework/Attributes/DataSource/DynamicDataOperations.cs
Evangelink: Removed, thanks!

src/TestFramework/TestFramework/Attributes/DataSource/DynamicDataOperations.cs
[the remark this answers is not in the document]
Evangelink: Thanks for flagging the behavior difference. The change is intentional and accepted per @Youssef1313's feedback in the other thread — accessing private static members from a base type via `[DynamicData]` is an uncommon pattern, and `FlattenHierarchy` aligns better with conventional .NET expectations.

Regarding `AmbiguousMatchException`: this is not a new risk since the member name + type are resolved from the attribute, not discovered broadly.

The analyzer has also been updated to match the new runtime behavior (filtering out private members from base types), so diagnostics stay in sync.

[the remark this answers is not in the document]
Evangelink: Done — PR title and description have been updated to reflect the actual approach (simplified `FlattenHierarchy` lookups, no caching).

Per @Youssef1313's suggestion, the analyzer has been updated: `TryGetMemberCore` now takes a `bool disallowPrivate` parameter — `false` for the declaring type, `true` for base types — filtering out `Accessibility.Private` members from base types to match the `FlattenHierarchy` runtime behavior.

A test `WhenPrivateMemberIsFromBase_Diagnostic` has been added to cover this.

PR #7536, opened 2026-03-12
Optimize CollectionAssert: hoist GetTypeInfo() and avoid redundant eumerations

Evangelink:
- AllItemsAreInstancesOfType: Move expectedType.GetTypeInfo() outside the loop since it is loop-invariant, avoiding a redundant call per collection element.

- AreNotEquivalent: Cache Count() results in local variables instead of calling Count() and then Any() separately, eliminating a third full enumeration of the collection.

- AreEquivalent: Replace Any() call with a check against the already-cached count variable.

### DIFF
--- a/src/TestFramework/TestFramework/Assertions/CollectionAssert.cs
+++ b/src/TestFramework/TestFramework/Assertions/CollectionAssert.cs
@@ -504,7 +504,7 @@ public sealed class CollectionAssert
         }
 
         // If both collections are empty, they are equivalent.
-        if (!expected.Any())
+        if (expectedCollectionCount == 0)
         {
             return;
         }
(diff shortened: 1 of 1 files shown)


src/TestFramework/TestFramework/Assertions/CollectionAssert.cs
```diff
@@ -740,11 +742,11 @@ public static void AllItemsAreInstancesOfType(
     {
         Assert.CheckParameterNotNull(collection, "CollectionAssert.AllItemsAreInstancesOfType", "collection");
         Assert.CheckParameterNotNull(expectedType, "CollectionAssert.AllItemsAreInstancesOfType", "expectedType");
+        TypeInfo expectedTypeInfo = expectedType.GetTypeInfo();
```
Youssef1313: The whole `GetTypeInfo` call seems redundant to me. We should just compare `System.Type`

Evangelink: Good call — removed `GetTypeInfo()` entirely. Now using `Type.IsAssignableFrom(Type)` directly, which works on all targeted frameworks.

PR #7541, opened 2026-03-12
Optimize TestMethodRunner: fix double enumeration, reduce allocations

Evangelink:
- Fix double enumeration in TryExecuteFoldedDataDrivenTestsAsync by materializing GetData() into an IReadOnlyList once instead of calling .Any() then re-enumerating
- Remove unnecessary List allocation in UpdateResultsWithParentInfo by mutating in place and returning void
- Cache result[result.Length - 1] into a local in ExecuteAsync catch block
- Skip redundant first-element self-comparison in GetAggregateOutcome
- Use Stopwatch.StartNew() in ExecuteTestFromDataSourceAttributeAsync

### DIFF
--- a/src/Adapter/MSTestAdapter.PlatformServices/Execution/TestMethodRunner.cs
+++ b/src/Adapter/MSTestAdapter.PlatformServices/Execution/TestMethodRunner.cs
@@ -205,8 +205,8 @@ internal sealed class TestMethodRunner
                 continue;
             }
 
-            // This code is to execute tests. To discover the tests code is in AssemblyEnumerator.ProcessTestDataSourceTests.
-            // Any change made here should be reflected in AssemblyEnumerator.ProcessTestDataSourceTests as well.
+            // This code is to execute tests. To discover the tests code is in AssemblyEnumerator.TryUnfoldITestDataSource.
+            // Any change made here should be reflected in AssemblyEnumerator.TryUnfoldITestDataSource as well.
             bool dataSourceHasData = false;
             foreach (object?[] data in testDataSource.GetData(_testMethodInfo.MethodInfo))
             {
--- a/src/Adapter/MSTestAdapter.PlatformServices/Discovery/AssemblyEnumerator.cs
+++ b/src/Adapter/MSTestAdapter.PlatformServices/Discovery/AssemblyEnumerator.cs
@@ -269,34 +269,18 @@ internal class AssemblyEnumerator : MarshalByRefObject
     private static bool TryUnfoldITestDataSource(ITestDataSource dataSource, UnitTestElement test, ReflectionTestMethodInfo methodInfo, List<UnitTestElement> tests, ref int globalTestCaseIndex)
     {
         // Otherwise, unfold the data source and verify it can be serialized.
-        IEnumerable<object?[]>? data;
 
-        // This code is to discover tests. To run the tests code is in TestMethodRunner.ExecuteDataSourceBasedTests.
-        // Any change made here should be reflected in TestMethodRunner.ExecuteDataSourceBasedTests as well.
-        data = dataSource.GetData(methodInfo);
+        // This code is to discover tests. To run the tests code is in TestMethodRunner.TryExecuteFoldedDataDrivenTestsAsync.
+        // Any change made here should be reflected in TestMethodRunner.TryExecuteFoldedDataDrivenTestsAsync as well.
+        IEnumerable<object?[]> dataEnumerable = dataSource.GetData(methodInfo);
         string? testDataSourceIgnoreMessage = (dataSource as ITestDataSourceIgnoreCapability)?.IgnoreMessage;
 
-        if (!data.Any())
-        {
-            if (!MSTestSettings.CurrentSettings.ConsiderEmptyDataSourceAsInconclusive)
-            {
-                throw dataSource.GetExceptionForEmptyDataSource(methodInfo);
-            }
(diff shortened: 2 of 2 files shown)


src/Adapter/MSTestAdapter.PlatformServices/Execution/TestMethodRunner.cs
Youssef1313: In case of exception, I would expect `result` to always be null here. Looks like we have some dead code here and we can simplify to just create a single TestResult right away, unless I'm missing something very obvious. Cleaning up in https://github.com/microsoft/testfx/pull/7546

src/Adapter/MSTestAdapter.PlatformServices/Execution/TestMethodRunner.cs
```diff
@@ -214,13 +215,13 @@ private async Task<bool> TryExecuteFoldedDataDrivenTestsAsync(List<TestResult> r
                 continue;
             }
 
-            IEnumerable<object?[]>? dataSource;
-
             // This code is to execute tests. To discover the tests code is in AssemblyEnumerator.ProcessTestDataSourceTests.
             // Any change made here should be reflected in AssemblyEnumerator.ProcessTestDataSourceTests as well.
-            dataSource = testDataSource.GetData(_testMethodInfo.MethodInfo);
+            IReadOnlyList<object?[]> dataSource = testDataSource.GetData(_testMethodInfo.MethodInfo) is IReadOnlyList<object?[]> dataList
```
Youssef1313: Aside from double calling GetData, I'm not really sure if this is more optimized or if it's worse.

Youssef1313: What I probably would do is call GetData normally, without materializing to a list, but still enumerate once by keeping a bool like `dataSourceHasData = false` before the loop, and set it to true inside the loop. After the loop, if it remained false, we know we didn't have data. That way, we didn't need to materialize to a list, and we didn't double enumerate.

src/Adapter/MSTestAdapter.PlatformServices/Execution/TestMethodRunner.cs
```diff
@@ -449,9 +449,9 @@ private static UnitTestOutcome GetAggregateOutcome(List<TestResult> results)
 
         // Get aggregate outcome.
         UnitTestOutcome aggregateOutcome = results[0].Outcome;
-        foreach (TestResult result in results)
+        for (int i = 1; i < results.Count; i++)
```
Youssef1313, line 439: This should have roughly the same perf as before IMO, but I'm not opposed to the change.

src/Adapter/MSTestAdapter.PlatformServices/Execution/UnitTestRunner.cs
```diff
@@ -273,12 +273,12 @@ private static async Task<TestResult> RunAssemblyInitializeIfNeededAsync(TestMet
 
     private static async Task<TestResult?> RunAssemblyCleanupAsync(ITestContext testContext, TypeCache typeCache, TestResult[] results)
     {
+        var testContextImpl = testContext as TestContextImplementation;
         IEnumerable<TestAssemblyInfo> assemblyInfoCache = typeCache.AssemblyInfoListWithExecutableCleanupMethods;
```
Youssef1313, line 273: `assemblyInfoCache` is always a single-element today AFAIK. This is not saving anything IMO, but I don't mind the change.

src/Adapter/MSTestAdapter.PlatformServices/Execution/UnitTestRunner.cs
```diff
@@ -318,32 +318,25 @@ private static bool IsTestMethodRunnable(
...
 
-        string? ignoreMessage = ignoreMessageOnClass;
-        if (StringEx.IsNullOrEmpty(ignoreMessage) && shouldIgnoreMethod)
-        {
-            ignoreMessage = ignoreMessageOnMethod;
-        }
-
         if (shouldIgnoreClass || shouldIgnoreMethod)
         {
+            string? ignoreMessage = StringEx.IsNullOrEmpty(ignoreMessageOnClass) ? ignoreMessageOnMethod : ignoreMessageOnClass;
```
Youssef1313: There is a slight behavior change here.

If class is ignored with empty message, and method is not ignored.

- Previous behavior: we create ignore result with empty message.
- New behavior: we create ignore result with null message.

I have no idea if this will have any user-visible impact or not, but it's probably safer to keep the old behavior.


```suggestion
            string? ignoreMessage = shouldIgnoreMethod && StringEx.IsNullOrEmpty(ignoreMessageOnClass) ? ignoreMessageOnMethod : ignoreMessageOnClass;
```

src/Adapter/MSTestAdapter.PlatformServices/Execution/TestMethodRunner.cs
```diff
@@ -84,14 +84,15 @@ internal async Task<TestResult[]> ExecuteAsync(string? initializationLogs, strin
             }
 
 #pragma warning disable IDE0056 // Use index operator
+            TestResult lastResult = result[result.Length - 1];
```
Youssef1313: Eliminating this part completely in https://github.com/microsoft/testfx/pull/7546

Youssef1313 (approved):

PR #7547, opened 2026-03-13
Cleanup AssemblyInitialize failure handling

Youssef1313:

### DIFF
--- a/test/UnitTests/MSTestAdapter.PlatformServices.UnitTests/Execution/TestAssemblyInfoTests.cs
+++ b/test/UnitTests/MSTestAdapter.PlatformServices.UnitTests/Execution/TestAssemblyInfoTests.cs
@@ -79,9 +79,10 @@ public class TestAssemblyInfoTests : TestContainer
 
         _testAssemblyInfo.AssemblyInitializeMethod = null;
 
-        await _testAssemblyInfo.RunAssemblyInitializeAsync(null!);
+        TestResult result = await _testAssemblyInfo.RunAssemblyInitializeAsync(null!);
 
         assemblyInitCallCount.Should().Be(0);
+        result.Outcome.Should().Be(UnitTestOutcome.Passed);
     }
 
     public async Task RunAssemblyInitializeShouldNotExecuteAssemblyInitializeIfItHasAlreadyExecuted()
@@ -92,9 +93,10 @@ public class TestAssemblyInfoTests : TestContainer
         _testAssemblyInfo.IsAssemblyInitializeExecuted = true;
         _testAssemblyInfo.AssemblyInitializeMethod = typeof(DummyTestClass).GetMethod("AssemblyInitializeMethod")!;
 
-        await _testAssemblyInfo.RunAssemblyInitializeAsync(_testContext);
+        TestResult result = await _testAssemblyInfo.RunAssemblyInitializeAsync(_testContext);
 
         assemblyInitCallCount.Should().Be(0);
+        result.Outcome.Should().Be(UnitTestOutcome.Passed);
     }
 
     public async Task RunAssemblyInitializeShouldExecuteAssemblyInitialize()
@@ -103,9 +105,10 @@ public class TestAssemblyInfoTests : TestContainer
         DummyTestClass.AssemblyInitializeMethodBody = _ => assemblyInitCallCount++;
         _testAssemblyInfo.AssemblyInitializeMethod = typeof(DummyTestClass).GetMethod("AssemblyInitializeMethod")!;
 
-        await _testAssemblyInfo.RunAssemblyInitializeAsync(_testContext);
+        TestResult result = await _testAssemblyInfo.RunAssemblyInitializeAsync(_testContext);
 
         assemblyInitCallCount.Should().Be(1);
+        result.Outcome.Should().Be(UnitTestOutcome.Passed);
     }
 
     public async Task RunAssemblyInitializeShouldSetAssemblyInitializeExecutedFlag()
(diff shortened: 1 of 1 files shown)


src/Adapter/MSTestAdapter.PlatformServices/Execution/UnitTestRunner.cs
```diff
@@ -244,15 +244,11 @@ internal async Task<TestResult[]> RunSingleTestAsync(UnitTestElement unitTestEle
...
     private static async Task<TestResult> RunAssemblyInitializeIfNeededAsync(TestMethodInfo testMethodInfo, ITestContext testContext)
     {
-        var result = new TestResult { Outcome = UnitTestOutcome.Passed };
+        TestResult? result = null;
 
         try
         {
-            await testMethodInfo.Parent.Parent.RunAssemblyInitializeAsync(testContext.Context).ConfigureAwait(false);
-        }
-        catch (TestFailedException ex)
```
Youssef1313, line 253: Start reviewing form here. The core idea is to avoid letting `RunAssemblyInitializeAsync` from throwing an exception that we immediately catch here. Instead, it returns the exception.

src/Adapter/MSTestAdapter.PlatformServices/Execution/TestAssemblyInfo.cs
```diff
@@ -161,7 +161,7 @@ public async Task RunAssemblyInitializeAsync(TestContext testContext)
                     }
                     catch (Exception ex)
                     {
-                        AssemblyInitializationException = ex;
+                        AssemblyInitializationException = GetTestFailedExceptionFromAssemblyInitializeException(ex, AssemblyInitializeMethod);
```
Youssef1313, line 168: We were setting `AssemblyInitializationException` here to a value other than TestFailedException.

After that, we used to re-assign it to a TestFailedException that we create.

After the refactoring, the logic to create the `TestFailedException` is separated to a new method `GetTestFailedExceptionFromAssemblyInitializeException`, and we simply call it here and assign once to `AssemblyInitializationException`.

In addition, all `throw`s are changed to `return`s.

test/UnitTests/MSTestAdapter.PlatformServices.UnitTests/Execution/TestAssemblyInfoTests.cs
```diff
@@ -139,7 +139,8 @@ public async Task RunAssemblyInitializeShouldThrowTestFailedExceptionOnAssertion
 #pragma warning restore RS0030 // Do not use banned APIs
         _testAssemblyInfo.AssemblyInitializeMethod = typeof(DummyTestClass).GetMethod("AssemblyInitializeMethod")!;
 
-        TestFailedException exception = (await new Func<Task>(() => _testAssemblyInfo.RunAssemblyInitializeAsync(_testContext)).Should().ThrowAsync<TestFailedException>()).Which;
+        var exception = (await _testAssemblyInfo.RunAssemblyInitializeAsync(_testContext)).TestFailureException as TestFailedException;
```
Evangelink, line 146: nit: After the `as TestFailedException` cast, `exception` is `TestFailedException?` — the subsequent `.Outcome`, `.Message`, `.StackTraceInformation!` accesses have no compile-time null safety (`.Should().NotBeNull()` is only a runtime guard).

Consider using `.Should().BeOfType<TestFailedException>().Which` instead, which returns a strongly-typed non-null result and restores compile-time safety:

```csharp
var exception = (await _testAssemblyInfo.RunAssemblyInitializeAsync(_testContext))
    .TestFailureException.Should().BeOfType<TestFailedException>().Which;
```

This applies to all similarly-patterned tests in this file.

test/UnitTests/MSTestAdapter.PlatformServices.UnitTests/Execution/TestAssemblyInfoTests.cs
```diff
@@ -126,9 +126,9 @@ public async Task RunAssemblyInitializeShouldSetAssemblyInitializationExceptionO
 #pragma warning restore RS0030 // Do not use banned APIs
         _testAssemblyInfo.AssemblyInitializeMethod = typeof(DummyTestClass).GetMethod("AssemblyInitializeMethod")!;
 
-        Func<Task> action = () => _testAssemblyInfo.RunAssemblyInitializeAsync(_testContext);
-
-        await action.Should().ThrowAsync<Exception>();
+        TestResult testResult = await _testAssemblyInfo.RunAssemblyInitializeAsync(_testContext);
```
Evangelink, line 133: nit: Several unchanged tests (`RunAssemblyInitializeShouldNotInvokeIfAssemblyInitializeIsNull`, `RunAssemblyInitializeShouldExecuteAssemblyInitialize`, `RunAssemblyInitializeShouldSetAssemblyInitializeExecutedFlag`, etc.) now discard the `TestResult` return value. Since the method signature changed from `Task` to `Task<TestResult>`, it would be nice to assert `result.Outcome.Should().Be(UnitTestOutcome.Passed)` on the success-path tests for completeness.

test/UnitTests/MSTestAdapter.PlatformServices.UnitTests/Execution/TestAssemblyInfoTests.cs
Youssef1313: It's a compile-time null safety, that's why we don't get a warning.

test/UnitTests/MSTestAdapter.PlatformServices.UnitTests/Execution/TestAssemblyInfoTests.cs
Youssef1313: @copilot Implement this please.

PR #7551, opened 2026-03-13
Cleanup integration tests and avoid redundant work

Youssef1313:


test/IntegrationTests/Microsoft.Testing.Platform.Acceptance.IntegrationTests/IgnoreExitCodeTests.cs
```diff
@@ -100,7 +100,6 @@ public async Task If_IgnoreExitCode_Specified_Should_Return_Success_ExitCode(str
 
         string assetPath = generator.TargetAssetPath;
         string globalPackagesPath = AcceptanceFixture.NuGetGlobalPackagesFolder.Path;
-        await DotnetCli.RunAsync($"restore {assetPath} -r {RID}", globalPackagesPath, cancellationToken: TestContext.CancellationToken);
```
Youssef1313, line 103: This `restore` call wasn't passing `-c`. So, for all "Release" test cases, we were restoring twice. Once for debug and once for release. The whole restore call isn't needed. We will let build do both the restore+build in a single call.

test/IntegrationTests/Microsoft.Testing.Platform.Acceptance.IntegrationTests/MSBuildTests.Solution.cs
```diff
@@ -8,17 +8,7 @@ public class MSBuildTests_Solution : AcceptanceTestBase<NopAssetFixture>
...
-        foreach ((string SingleTfmOrMultiTfm, BuildConfiguration BuildConfiguration, bool IsMultiTfm) entry in GetBuildMatrixSingleAndMultiTfmBuildConfiguration())
-        {
-            {
-                yield return new(entry.SingleTfmOrMultiTfm, entry.BuildConfiguration, entry.IsMultiTfm);
-            }
-        }
-    }
-
-    [DynamicData(nameof(GetBuildMatrix))]
+    [DynamicData(nameof(GetBuildMatrixSingleAndMultiTfmBuildConfiguration))]
```
Youssef1313, line 11: This is only code cleanup. `GetBuildMatrix` was only delegating to `GetBuildMatrixSingleAndMultiTfmBuildConfiguration`. Now we use `GetBuildMatrixSingleAndMultiTfmBuildConfiguration` directly.

test/IntegrationTests/Microsoft.Testing.Platform.Acceptance.IntegrationTests/NativeAotTests.cs
```diff
@@ -85,13 +85,8 @@ public async Task NativeAotTests_WillRunWithExitCodeZero(string tfm)
...
             addPublicFeeds: true);
 
-        await DotnetCli.RunAsync(
-            $"restore {generator.TargetAssetPath} -r {RID}",
-            AcceptanceFixture.NuGetGlobalPackagesFolder.Path,
-            retryCount: 0,
-            cancellationToken: TestContext.CancellationToken);
         DotnetMuxerResult compilationResult = await DotnetCli.RunAsync(
-            $"publish {generator.TargetAssetPath} -r {RID} -f {tfm}",
+            $"publish {generator.TargetAssetPath} -r {RID}",
```
Youssef1313, line 89: This is only a cleanup. The project is a single TFM already.

test/IntegrationTests/MSTest.Acceptance.IntegrationTests/AppDomainTests.cs
```diff
@@ -9,93 +9,24 @@ namespace MSTest.Acceptance.IntegrationTests;
...
-}
-""";
-
     [TestMethod]
     [DataRow(true)]
     [DataRow(false)]
     [DataRow(null)]
     public async Task RunTests_With_VSTest(bool? disableAppDomain)
     {
-        using TestAsset testAsset = await TestAsset.GenerateAssetAsync(
```
Youssef1313, line 85: Here we were re-generating and re-building the same asset over and over. This is now done once in class initialize.

Issue #6907, opened 2025-11-07
TaskExtensionsTests are flaky


Youssef1313:
Disabling in https://github.com/microsoft/testfx/pull/6908

Evangelink:
CLosing as complete, I haven't seen the issue in a while.

Youssef1313:
@Evangelink Because it's worked-around in https://github.com/microsoft/testfx/blob/06dc5549fd427e00686766954348cee8a7c74c08/test/UnitTests/Microsoft.Testing.Platform.UnitTests/Helpers/TaskExtensionsTests.cs#L87 and that should be reverted.

Youssef1313:
Actually whole test method (and another test method in class) is marked with `[Ignore]`. This issue is still relevant.

PR #7564, opened 2026-03-16
Avoid plugging in TerminalOutputDevice when running with pipe protocol

### LINKED ISSUE #7161: Consider: Disable MTP's own TerminalTestReporter when running under server mode (pipe specifically maybe?)
When running with dotnet test using the pipe protocol, SDK's TerminalTestReporter is responsible for reporting the test results, progress, etc...

Doing that work in the test app itself might lead to:

1. Work being thrown away.
2. Unnecessary console writes.
3. Duplicate information printed by TerminalTestReporter when we attempt to print the test app process standard output/err.

This is also quite related to https://github.com/dotnet/sdk/issues/51615.

Youssef1313:
Fixes #7161


src/Platform/Microsoft.Testing.Platform/ServerMode/DotnetTest/IPC/Constants.cs
```diff
@@ -37,5 +37,9 @@ internal static class HandshakeMessagePropertyNames
 
 internal static class ProtocolConstants
 {
-    internal const string Version = "1.0.0";
+    // The change between 1.0.0 and 1.0.1 is that TerminalOutputDevice is no longer plugged in.
+    // That's not really a protocol change, but we use the version to signify to SDK that it can avoid output redirection.
+    // So, when SDK declares itself as supporting 1.0.1, and MTP is also using 1.0.1, and we negotiate to that version.
+    // Then SDK can assume that MTP output doesn't interfere with SDK output, and we can safely let live output to work.
+    internal const string SupportedVersions = "1.0.0;1.0.1";
```
Evangelink: Shall we consider this as feature rather than bug fix (so change being 1.1.0)?

Youssef1313: Sure. The versioning we have here is an implementation today anyways and doesn't leak anywhere. But I'm fine using 1.1.0

src/Platform/Microsoft.Testing.Platform/Hosts/TestHostBuilder.cs
```diff
@@ -207,9 +207,12 @@ public async Task<IHost> BuildAsync(
         serviceProvider.AddService(policiesService);
 
         bool hasServerFlag = commandLineHandler.TryGetOptionArgumentList(PlatformCommandLineProvider.ServerOptionKey, out string[]? protocolName);
-        bool isJsonRpcProtocol = protocolName is null || protocolName.Length == 0 || protocolName[0].Equals(PlatformCommandLineProvider.JsonRpcProtocolName, StringComparison.OrdinalIgnoreCase);
+        bool isJsonRpcProtocol = hasServerFlag &&
+            (protocolName is null || protocolName.Length == 0 || protocolName[0].Equals(PlatformCommandLineProvider.JsonRpcProtocolName, StringComparison.OrdinalIgnoreCase));
 
-        ProxyOutputDevice proxyOutputDevice = await _outputDisplay.BuildAsync(serviceProvider, hasServerFlag && isJsonRpcProtocol).ConfigureAwait(false);
+        bool isPipeProtocol = hasServerFlag && protocolName?.Length == 1 && protocolName[1].Equals(PlatformCommandLineProvider.DotnetTestCliProtocolName, StringComparison.Ordinal);
```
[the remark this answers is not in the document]
Evangelink: False positive — the code uses `protocolName[0]`, not `[1]`. That said, the underlying suggestion to reuse `DotnetTestHelper.HasDotnetTestServerOption(...)` was a good one and `TestHostBuilder.CommonServices.cs` now routes the pipe-protocol detection through that helper in d6701cb, so the parsing is no longer duplicated.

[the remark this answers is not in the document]
Evangelink: Addressed in d6701cb by going with the no-op route you suggested below. Pipe protocol now installs `NopPlatformOutputDevice` (`IsEnabledAsync` → `false`), so `ProxyOutputDevice.OriginalOutputDevice` stays non-null and the `IsEnabledAsync` / `TryInitializeAsync` / `TryAddService` calls in `TestHostBuilder.CommonServices.cs` no longer need any null guards.

src/Platform/Microsoft.Testing.Platform/OutputDevice/ProxyOutputDevice.cs
```diff
@@ -11,17 +11,21 @@ internal sealed class ProxyOutputDevice : IOutputDevice
...
 
-    public ProxyOutputDevice(IPlatformOutputDevice originalOutputDevice, ServerModePerCallOutputDevice? serverModeOutputDevice)
+    public ProxyOutputDevice(IPlatformOutputDevice? originalOutputDevice, ServerModePerCallOutputDevice? serverModeOutputDevice)
     {
         OriginalOutputDevice = originalOutputDevice;
         _serverModeOutputDevice = serverModeOutputDevice;
     }
 
-    internal IPlatformOutputDevice OriginalOutputDevice { get; }
+    internal IPlatformOutputDevice? OriginalOutputDevice { get; }
```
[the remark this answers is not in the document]
Evangelink: Addressed in d6701cb by introducing `NopPlatformOutputDevice` and dropping the nullable-property change. `ProxyOutputDevice.OriginalOutputDevice`, `GetPlatformOutputDevice()`, and `TestFrameworkBuilderData.PlatformOutputDisplayService` are all back to non-null, so no downstream consumer has to become null-aware.

src/Platform/Microsoft.Testing.Platform/OutputDevice/OutputDeviceManager.cs
```diff
@@ -43,8 +45,13 @@ internal async Task<ProxyOutputDevice> BuildAsync(ServiceProvider serviceProvide
                 : null);
     }
 
-    public static IPlatformOutputDevice GetDefaultTerminalOutputDevice(ServiceProvider serviceProvider)
+    private static IPlatformOutputDevice? GetDefaultTerminalOutputDevice(ServiceProvider serviceProvider, bool isPipeProtocol)
     {
+        if (isPipeProtocol)
+        {
+            return null;
+        }
```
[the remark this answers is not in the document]
Evangelink: Addressed in d6701cb. `BuildAsync` now short-circuits for pipe protocol and returns `new ProxyOutputDevice(new NopPlatformOutputDevice(), serverModeOutputDevice: null)`, so `GetDefaultTerminalOutputDevice` is never called in that mode and keeps its non-null return type.

src/Platform/Microsoft.Testing.Platform/ServerMode/DotnetTest/IPC/Constants.cs
```diff
@@ -37,5 +37,9 @@ internal static class HandshakeMessagePropertyNames
 
 internal static class ProtocolConstants
 {
-    internal const string Version = "1.0.0";
+    // The change between 1.0.0 and 1.1.0 is that TerminalOutputDevice is no longer plugged in.
+    // That's not really a protocol change, but we use the version to signify to SDK that it can avoid output redirection.
+    // So, when SDK declares itself as supporting 1.1.0, and MTP is also using 1.1.0, and we negotiate to that version.
+    // Then SDK can assume that MTP output doesn't interfere with SDK output, and we can safely let live output to work.
```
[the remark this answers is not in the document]
Evangelink: Already addressed in 52d6ce1 (kept in the rebased commit d6701cb) — the comment now references 1.1.0 in both spots and `SupportedVersions` advertises `1.0.0;1.1.0`.

src/Platform/Microsoft.Testing.Platform/OutputDevice/OutputDeviceManager.cs
```diff
@@ -43,8 +45,13 @@ internal async Task<ProxyOutputDevice> BuildAsync(ServiceProvider serviceProvide
                 : null);
     }
 
-    public static IPlatformOutputDevice GetDefaultTerminalOutputDevice(ServiceProvider serviceProvider)
+    private static IPlatformOutputDevice? GetDefaultTerminalOutputDevice(ServiceProvider serviceProvider, bool isPipeProtocol)
     {
+        if (isPipeProtocol)
+        {
+            return null;
```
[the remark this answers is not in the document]
Evangelink: No longer applicable: in d6701cb the pipe-protocol short-circuit happens in `BuildAsync` (returns a `NopPlatformOutputDevice`-backed `ProxyOutputDevice`), so `GetDefaultTerminalOutputDevice` keeps its existing non-null return contract and the name stays accurate.

Evangelink:
Rebased onto current `main` (resolving the conflict from #8201's TestHostBuilder refactor) and reworked the patch to address the earlier review feedback:

* Introduced a new `NopPlatformOutputDevice` (`IsEnabledAsync` → `false`) that is installed for the pipe protocol. This avoids the nullable-property surface area changes (`ProxyOutputDevice.OriginalOutputDevice`, `GetPlatformOutputDevice()`, `TestFrameworkBuilderData.PlatformOutputDisplayService`) and keeps every downstream consumer null-free, while `RegisterAsServiceOrConsumerOrBothAsync` naturally skips it as a data consumer.
* Pipe-protocol detection now reuses `DotnetTestHelper.HasDotnetTestServerOption` (added to `BuildContext.IsPipeProtocol`).
* Kept Youssef's `1.1.0` protocol bump and matching `SupportedVersions = "1.0.0;1.1.0"` advertisement during handshake.
* Added a small `NopPlatformOutputDeviceTests` covering identity, `IsEnabledAsync`, and no-op behavior of all `Display*` / `HandleProcessRole*` methods.

Local validation on Windows Debug:
* `Microsoft.Testing.Platform` builds clean on net8.0 / net9.0 / netstandard2.0 (0 warnings).
* `Microsoft.Testing.Platform.UnitTests` net8.0: 1164 / 1164 succeeded (2 known #6907 skips).

Youssef1313 (commented):
The force push makes it harder to navigate through the history.

But IIRC, I thought this must require a protocol version bump.

src/Platform/Microsoft.Testing.Platform/ServerMode/DotnetTest/IPC/Constants.cs
```diff
@@ -55,5 +55,11 @@ internal static class HandshakeMessageExecutionModes
...
 internal static class ProtocolConstants
 {
-    internal const string Version = "1.0.0";
+    // The change between 1.0.0 and 1.1.0 is that TerminalOutputDevice is no longer plugged in.
+    // That's not really a protocol change, but we use the version to signal to the SDK that it
+    // can safely keep the test host's standard output/error visible (TerminalTestReporter and
+    // host output will no longer collide).
+    // When both sides advertise 1.1.0 and we negotiate to that version, the SDK can keep its
+    // live output enabled.
+    internal const string SupportedVersions = "1.0.0;1.1.0";
```
Youssef1313, line 92: Oh I overlooked the PR and didn't see this in my first pass. That sounds good

src/Platform/Microsoft.Testing.Platform/OutputDevice/OutputDeviceManager.cs
```diff
@@ -17,8 +17,16 @@ internal sealed class PlatformOutputDeviceManager
     public void SetPlatformOutputDevice(Func<IServiceProvider, IPlatformOutputDevice> platformOutputDeviceFactory)
         => _platformOutputDeviceFactory = platformOutputDeviceFactory ?? throw new ArgumentNullException(nameof(platformOutputDeviceFactory));
 
-    internal async Task<ProxyOutputDevice> BuildAsync(ServiceProvider serviceProvider, bool useServerModeOutputDevice)
+    internal async Task<ProxyOutputDevice> BuildAsync(ServiceProvider serviceProvider, bool useServerModeOutputDevice, bool isPipeProtocol)
     {
+        // Under the dotnet test pipe protocol, the SDK's TerminalTestReporter owns all
+        // user-facing output, so we deliberately install a no-op device here. See PR #7564
+        // and dotnet/sdk#51615 for the broader context.
+        if (isPipeProtocol)
```
Youssef1313, line 25: I lost some context, but are we expected to use the "nop" output device only when using the 1.1.0 protocol?

That might be challenging though as I think we don't handshake that early during building the app.

Evangelink:
### 🧪 Test quality grade — PR #7564



<table>
  <thead>
  </thead>
  <tbody>
    <tr>
      <td>new</td>
      <td><code>NopPlatformOutputDeviceTests.<br>DisplayMethods_<br>AreNoOps_<br>AndDoNotThrow</code></td>
      <td>A</td>
      <td>90–100</td>
      <td>Comment-separated AAA phases; <code>VerifyNoOtherCalls()</code> on both mocks confirms no producer/data interaction.</td>
    </tr>
    <tr>
      <td>new</td>
      <td><code>NopPlatformOutputDeviceTests.<br>Identity_<br>IsStableAndNonEmpty</code></td>
      <td>A</td>
      <td>90–100</td>
      <td>Multiple equality assertions cover all identity-contract properties; non-empty check on Description is appropriately flexible.</td>
    </tr>
    <tr>
      <td>new</td>
      <td><code>NopPlatformOutputDeviceTests.<br>IsEnabledAsync_<br>ReturnsFalse_<br>SoExtensionRegistrationSkipsIt</code></td>
      <td>A</td>
      <td>90–100</td>
      <td>Direct boolean assertion on <code>IsEnabledAsync</code>; focused and complete for a single-return contract.</td>
    </tr>
  </tbody>
</table>

`/grade-tests`.</sub>



> 🤖 **Automated content by GitHub Copilot.** Posted via a maintainer's GitHub token, so it appears under their account — the account owner did **not** write or approve this content personally. Generated by the [Grade Tests on PR (on open / sync)](https://github.com/microsoft/testfx/actions/runs/27691885597/agentic_workflow) workflow. · 130 AIC · ⌖ 13 AIC · [◷]( · [◷](https://github.com/search?q=repo%3Amicrosoft%2Ftestfx+%22gh-aw-workflow-call-id%3A+microsoft%2Ftestfx%2Fgrade-tests-on-pr.agent%22&type=issues))

Evangelink (commented):
> [!NOTE]
> 🤖 **Automated review by GitHub Copilot.** Posted via a maintainer's GitHub token, so it appears under their account — the account owner did **not** write or approve this content personally. Generated by the [Expert Code Review workflow](https://github.com/microsoft/testfx/actions/runs/27691885608). To request a follow-up action, reply by tagging `@copilot` directly.

## Review Summary — PR #7564: Avoid plugging in TerminalOutputDevice when running with pipe protocol

The core approach is correct and well-motivated. Installing `NopPlatformOutputDevice` at startup cleanly resolves the `TerminalOutputDevice` ↔ `TerminalTestReporter` output collision (issue #7161) without requiring a complex deferred swap or runtime switching of the output device. The new class is clean, the unit tests follow project conventions (MSTest assertions, Moq mocks, no `AwesomeAssertions`), and the version-bump mechanism (`"1.0.0;1.1.0"`) is a reasonable way to signal the SDK that it can safely enable its live output.

---

### Dimension verdict table

| # | Dimension | Verdict |
|---|-----------|---------|
| 1 | Algorithmic Correctness | ⚠️ MODERATE — see inline |
| 2 | Threading & Concurrency | ✅ LGTM |
| 3 | Security & IPC Contract Safety | ✅ LGTM |
| 4 | Public API & Binary Compatibility | ✅ LGTM (all new types are `internal sealed`) |
| 5 | Performance & Allocations | ✅ LGTM |
| 6 | Cross-TFM Compatibility | ✅ LGTM |
| 7 | Resource & IDisposable Management | ✅ LGTM |
| 8 | Defensive Coding at Boundaries | ✅ LGTM |
| 9 | Localization & Resources | ✅ LGTM (description on internal class, `IsEnabledAsync = false`) |
| 10 | Test Isolation | ✅ LGTM |
| 11 | Assertion Quality | ✅ LGTM |
| 12 | Flakiness Patterns | ✅ LGTM |
| 13 | Test Completeness & Coverage | ⚠️ NIT — see inline (`TestHostOrchestrator` role missing) |
| 14 | Data-Driven Test Coverage | N/A |
| 15 | Code Structure & Simplification | ✅ LGTM |
| 16 | Naming & Conventions | ✅ LGTM |
| 17 | Documentation Accuracy | ⚠️ NIT — see inline (PR ref instead of issue ref; TODO without issue) |
| 18 | Analyzer & Code Fix Quality | N/A |
| 19 | IPC Wire Compatibility | ✅ LGTM — `SupportedVersions = "1.0.0;1.1.0"` is backward-compatible: the SDK sends its *own* supported version back (not a mirror), so old SDKs respond with `"1.0.0"` and `IsVersionCompatible("1.0.0", "1.0.0;1.1.0")` correctly returns `true`. |
| 20 | Build Infrastructure & Dependencies | ✅ LGTM |
| 21 | Scope & PR Discipline | ✅ LGTM |
| 22 | PowerShell Scripting Hygiene | N/A |

**18/22 dimensions clean. 0 BLOCKING. 0 MAJOR. 1 MODERATE. 3 NITs.**

---

### Notable observations (not flagged inline)

**`context.IsPipeProtocol` in `BuildContext`**: The field is set and then used only in the single `BuildAsync(...)` call. It could be a local variable today, but storing it in the context follows the established pattern of other protocol-state flags (`HasServerFlag`, `IsJsonRpcProtocol`) and leaves it accessible for any future build stage that might need it — acceptable.

**`HasDotnetTestServerOption()` without requiring `--dotnet-test-pipe`**: `NopPlatformOutputDevice` is installed whenever `--server dotnet-test-cli` is present, even if `--dotnet-test-pipe` is absent (in which case no pipe is actually connected). This is an edge case that `dotnet test` never triggers in practice (it always supplies both), so the risk is negligible.

**Missing `OutputDeviceManager.BuildAsync(isPipeProtocol: true)` unit test**: The `NopPlatformOutputDeviceTests` correctly validates the new class in isolation, but there is no test that calls `PlatformOutputDeviceManager.BuildAsync(..., isPipeProtocol: true)` and asserts the returned `ProxyOutputDevice` wraps a `NopPlatformOutputDevice`. Adding one would give regression coverage for the bypass logic itself.

> 🤖 **Automated content by GitHub Copilot.** Posted via a maintainer's GitHub token, so it appears under their account — the account owner did **not** write or approve this content personally. Generated by the [Expert Code Review (on PR ready)](https://github.com/microsoft/testfx/actions/runs/27691885608/agentic_workflow) workflow. · 712 AIC · ⌖ 13.8 AIC · [◷]()

src/Platform/Microsoft.Testing.Platform/OutputDevice/OutputDeviceManager.cs
```diff
@@ -17,8 +17,16 @@ internal sealed class PlatformOutputDeviceManager
     public void SetPlatformOutputDevice(Func<IServiceProvider, IPlatformOutputDevice> platformOutputDeviceFactory)
         => _platformOutputDeviceFactory = platformOutputDeviceFactory ?? throw new ArgumentNullException(nameof(platformOutputDeviceFactory));
 
-    internal async Task<ProxyOutputDevice> BuildAsync(ServiceProvider serviceProvider, bool useServerModeOutputDevice)
+    internal async Task<ProxyOutputDevice> BuildAsync(ServiceProvider serviceProvider, bool useServerModeOutputDevice, bool isPipeProtocol)
     {
+        // Under the dotnet test pipe protocol, the SDK's TerminalTestReporter owns all
+        // user-facing output, so we deliberately install a no-op device here. See PR #7564
```
Evangelink: **[NIT] Documentation Accuracy**

The comment cross-references the current PR (`#7564`) rather than the canonical issue (`#7161`). PR numbers are less stable references — PRs can be superseded, but the issue persists as the authoritative tracking point. Prefer:

```csharp
// ...deliberately install a no-op device here. See #7161 and dotnet/sdk#51615 for the broader context.
```

src/Platform/Microsoft.Testing.Platform/OutputDevice/OutputDeviceManager.cs
```diff
@@ -17,8 +17,16 @@ internal sealed class PlatformOutputDeviceManager
...
+        // and dotnet/sdk#51615 for the broader context.
+        if (isPipeProtocol)
+        {
+            return new ProxyOutputDevice(new NopPlatformOutputDevice(), serverModeOutputDevice: null);
+        }
+
         // TODO: SetPlatformOutputDevice isn't public yet.
```
Evangelink: **[NIT] TODO Comment Policy**

Pre-existing `TODO: SetPlatformOutputDevice isn't public yet.` lacks a GitHub issue reference. The repository policy requires the form `// TODO(#<issue>): ...`. Since this PR already touches the file, this is a good opportunity to either add the issue number or convert it to a plain explanatory comment if no issue exists.

Example: `// TODO(#NNNN): Expose SetPlatformOutputDevice publicly. Before doing so, decide whether to pass useServerModeOutputDevice to it.`

src/Platform/Microsoft.Testing.Platform/ServerMode/DotnetTest/IPC/Constants.cs
Evangelink, line 92: **[MODERATE] Algorithmic Correctness – Unconditional suppression before version negotiation**

`NopPlatformOutputDevice` is installed for **all** pipe-protocol connections at startup, before the protocol handshake in `DotnetTestConnection.IsCompatibleProtocolAsync`. The comment explains the intended semantics:

> When both sides advertise 1.1.0 and we negotiate to that version, the SDK can keep its live output enabled.

However, the NopPlatformOutputDevice is active even when the negotiated version ends up as `1.0.0` (old SDK). In that scenario:
- **MTP side**: `NopPlatformOutputDevice` → silent.
- **SDK side**: receives `1.0.0` negotiation → suppresses its own `TerminalTestReporter` live output (as it did before the collision fix).
- **Result**: both sides produce no live output during the run — a regression from the pre-PR behaviour where MTP's `TerminalOutputDevice` would at least show terminal progress.

This is a deliberate trade-off (no collision, no live output) rather than a bug, but it should be explicitly documented so maintainers are aware of the old-SDK user experience gap. E.g.:

```csharp
// NOTE: This suppression applies regardless of the negotiated protocol version. With an old SDK
// that only supports 1.0.0, both sides will produce no live output. Users must update to an SDK
// that negotiates 1.1.0 to see live output via the SDK's TerminalTestReporter.
```

test/UnitTests/Microsoft.Testing.Platform.UnitTests/OutputDevice/NopPlatformOutputDeviceTests.cs
```diff
@@ -0,0 +1,53 @@
...
+        Mock<IOutputDeviceData> dataMock = new();
+
+        // All Display* and HandleProcessRole methods should be no-ops that complete successfully.
+        await device.DisplayAsync(producerMock.Object, dataMock.Object, CancellationToken.None);
+        await device.DisplayBannerAsync("banner", CancellationToken.None);
+        await device.DisplayBannerAsync(bannerMessage: null, CancellationToken.None);
+        await device.DisplayBeforeSessionStartAsync(CancellationToken.None);
+        await device.DisplayAfterSessionEndRunAsync(CancellationToken.None);
+        await device.HandleProcessRoleAsync(TestProcessRole.TestHost, CancellationToken.None);
+        await device.HandleProcessRoleAsync(TestProcessRole.TestHostController, CancellationToken.None);
```
Evangelink, line 52: **[NIT] Test Completeness**

`TestProcessRole` has three values — `TestHost`, `TestHostController`, and `TestHostOrchestrator` — but only the first two are exercised here. `TestHostOrchestrator` is a valid role passed by `TestHostBuilder.Modes.cs` (`HandleProcessRoleAsync(TestProcessRole.TestHostOrchestrator, ...)`), so a no-op assertion for it is appropriate:

```csharp
await device.HandleProcessRoleAsync(TestProcessRole.TestHostOrchestrator, CancellationToken.None);
```

src/Platform/Microsoft.Testing.Platform/OutputDevice/OutputDeviceManager.cs
Evangelink: Done in 259586324 — the comment now references `#7161` (and `dotnet/sdk#51615`) instead of the PR number.

src/Platform/Microsoft.Testing.Platform/OutputDevice/OutputDeviceManager.cs
Evangelink: Done in 259586324 — there's no tracking issue for exposing `SetPlatformOutputDevice`, so I converted the `TODO` into a plain explanatory comment per the TODO policy.

src/Platform/Microsoft.Testing.Platform/ServerMode/DotnetTest/IPC/Constants.cs
Evangelink: Done in 259586324 — added a `NOTE` on `SupportedVersions` documenting that the no-op device is installed for all pipe-protocol connections regardless of negotiated version, so an old (1.0.0-only) SDK yields no live output until updated to a 1.1.0-negotiating SDK.

test/UnitTests/Microsoft.Testing.Platform.UnitTests/OutputDevice/NopPlatformOutputDeviceTests.cs
Evangelink: Done in 259586324 — added a `HandleProcessRoleAsync(TestProcessRole.TestHostOrchestrator, ...)` no-op assertion alongside the other two roles.

src/Platform/Microsoft.Testing.Platform/Hosts/TestHostBuilder.cs
```diff
@@ -116,6 +116,8 @@ private sealed class BuildContext(
 
         public bool IsJsonRpcProtocol { get; set; }
 
+        public bool IsPipeProtocol { get; set; }
```
Youssef1313: Why is this property needed?

Evangelink: Good point — it didn't need to be on the build context. The pipe-protocol flag is only read here in TestHostBuilder.CommonServices.cs (unlike IsJsonRpcProtocol/the server flag, which TestHostBuilder.Modes.cs also reads), so in bdbae0e38 I dropped the IsPipeProtocol context property and turned it into a local isPipeProtocol variable passed straight into BuildAsync.

src/Platform/Microsoft.Testing.Platform/OutputDevice/OutputDeviceManager.cs
Evangelink: Right — we can't scope it to the negotiated version, because (as you note) the handshake in DotnetTestConnection.IsCompatibleProtocolAsync happens well after the app is built and the output device is plugged in. So the no-op device is installed for *all* pipe-protocol connections, regardless of the version we end up negotiating. The trade-off (an old 1.0.0-only SDK gets no live output until it's updated to negotiate 1.1.0) is documented in the NOTE on ProtocolConstants.SupportedVersions. The 1.1.0 bump is just the signal to the SDK that it's now safe to keep its own \TerminalTestReporter\ live output enabled without colliding with ours.

Evangelink:
Merged latest main into the branch and fixed the broken Debug pipeline. The failure was the acceptance test `DotnetTestPipe_TestAppAdvertisesAndNegotiatesProtocolV100_Today` (added on main after this branch forked) — it was a deliberate baseline asserting the host advertises *only* `1.0.0`, designed to start failing once testfx bumps past 1.0.0. This PR's bump to `1.0.0;1.1.0` is exactly that trigger, so I updated the baseline accordingly in bdbae0e38:

- Renamed it to `DotnetTestPipe_TestAppAdvertises100And110_NegotiatesDownToV100WithOldSdk` — now asserts the host advertises `1.0.0;1.1.0` and that an old (1.0.0-only) SDK still negotiates **down** to `1.0.0`.
- Added `DotnetTestPipe_TestAppNegotiatesV110WithUpToDateSdk` — when the SDK advertises 1.1.0, negotiation selects the highest mutual version (`1.1.0`).

Merge conflict resolutions: folded `main`'s `hasServerFlag` local + `BuildAsync(serviceProvider, IsJsonRpcProtocol, ...)` signature change together with this PR's pipe-protocol path, and kept `ProtocolConstants.SupportedVersions` on top of main's restructured handshake-properties dictionary in `DotnetTestConnection`.

src/Platform/Microsoft.Testing.Platform/ServerMode/DotnetTest/IPC/Constants.cs
```diff
@@ -77,5 +77,17 @@ internal static class HandshakeMessageExecutionModes
...
+    // host output will no longer collide).
+    // When both sides advertise 1.1.0 and we negotiate to that version, the SDK can keep its
+    // live output enabled.
+    //
+    // NOTE: The no-op output device is installed for all pipe-protocol connections, regardless
+    // of the negotiated protocol version. With an old SDK that only supports 1.0.0, both sides
+    // will produce no live output (the SDK suppresses its TerminalTestReporter to avoid colliding
+    // with the host output it expected before this change). Users must update to an SDK that
+    // negotiates 1.1.0 to see live output via the SDK's TerminalTestReporter.
+    internal const string SupportedVersions = "1.0.0;1.1.0";
```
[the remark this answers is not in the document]
Evangelink: Good catch — this was the compile break in the latest CI run. Fixed in 51bf574d6: both ProtocolVersion_IsStable contract tests (ProtocolTests.cs and DotnetTestProtocolContractTests.cs) now reference ProtocolConstants.SupportedVersions and pin the new advertised list \1.0.0;1.1.0\. I went with updating the tests rather than keeping a legacy \Version\ alias, since the value is now a semicolon-separated list and a singular \Version\ name would be misleading.

test/UnitTests/Microsoft.Testing.Platform.UnitTests/OutputDevice/NopPlatformOutputDeviceTests.cs
```diff
@@ -0,0 +1,54 @@
...
+
+    [TestMethod]
+    public void Identity_IsStableAndNonEmpty()
+    {
+        var device = new NopPlatformOutputDevice();
+
+        Assert.AreEqual(nameof(NopPlatformOutputDevice), device.Uid);
+        Assert.AreEqual(nameof(NopPlatformOutputDevice), device.DisplayName);
+        Assert.AreEqual("1.0.0", device.Version);
+        Assert.IsFalse(string.IsNullOrWhiteSpace(device.Description));
```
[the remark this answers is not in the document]
Evangelink: Done in 51bf574d6 — the test now asserts \device.Version == PlatformVersion.Version\ instead of the hard-coded \1.0.0\, in step with the production change.

src/Platform/Microsoft.Testing.Platform/OutputDevice/NopPlatformOutputDevice.cs
```diff
@@ -0,0 +1,40 @@
...
+/// and null-free. <see cref="IsEnabledAsync"/> returns <see langword="false"/> so
+/// <c>RegisterAsServiceOrConsumerOrBothAsync</c> skips it as a data consumer.
+/// </remarks>
+internal sealed class NopPlatformOutputDevice : IPlatformOutputDevice
+{
+    public string Uid => nameof(NopPlatformOutputDevice);
+
+    public string Version => "1.0.0";
+
+    public string DisplayName => nameof(NopPlatformOutputDevice);
```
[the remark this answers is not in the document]
Evangelink: Agreed. Fixed in 51bf574d6 — NopPlatformOutputDevice.Version now returns PlatformVersion.Version, matching TerminalOutputDevice and the other IExtension implementations.

test/IntegrationTests/Microsoft.Testing.Platform.Acceptance.IntegrationTests/DotnetTestPipe/DotnetTestPipeBaselineTests.cs
```diff
@@ -23,12 +23,19 @@ public class DotnetTestPipeBaselineTests : AcceptanceTestBase<DotnetTestPipeBase
 
     public TestContext TestContext { get; set; } = null!;
 
+    // The test host (testfx) now advertises both protocol 1.0.0 and 1.1.0. The 1.1.0 bump signals
+    // that TerminalOutputDevice is no longer plugged in under the pipe protocol (see
+    // microsoft/testfx#7161 and dotnet/sdk#51615). This mirrors ProtocolConstants.SupportedVersions
+    // on the host side.
+    private const string HostAdvertisedProtocolVersions = "1.0.0;1.1.0";
```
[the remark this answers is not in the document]
Evangelink: I intentionally kept this as a literal rather than referencing \ProtocolConstants.SupportedVersions\. This is an acceptance test that pins the actual wire output the host sends — referencing the same constant the host code uses would make the first assertion tautological (it would pass no matter what the host advertises). Hard-coding the expected list means an unintended change to the advertised versions is caught here, and a deliberate change forces a conscious update — same contract-pin rationale as \ProtocolVersion_IsStable\ in the unit tests, which also pins the literal instead of the constant. Happy to switch if you'd prefer the DRY version, but I'd lean on keeping the explicit pin for the wire contract.

test/UnitTests/Microsoft.Testing.Platform.UnitTests/OutputDevice/NopPlatformOutputDeviceTests.cs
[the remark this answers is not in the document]
Evangelink: Correction: asserting device.Version == PlatformVersion.Version actually broke CI, because PlatformVersion is an \[Embedded]\ per-assembly constant — the test assembly's value (\4.3.0-ci\) differs from the Platform assembly's (\2.3.0-ci\) that the device returns. In 70f2a9706 the test now asserts the version is non-empty (\Assert.IsFalse(string.IsNullOrWhiteSpace(device.Version))\), which matches the \Identity_IsStableAndNonEmpty\ intent and the adjacent Description check, and avoids the brittle hard-coded \1.0.0\. The production change (returning \PlatformVersion.Version\ instead of a literal) is kept. Full Platform.UnitTests suite is green locally (1196 passed, 0 failed).

test/IntegrationTests/Microsoft.Testing.Platform.Acceptance.IntegrationTests/DotnetTestPipe/DotnetTestPipeBaselineTests.cs
```diff
@@ -7,7 +7,7 @@ namespace Microsoft.Testing.Platform.Acceptance.IntegrationTests.DotnetTestPipe;
 
 /// <summary>
 /// Baseline tests for the <c>--server dotnettestcli</c> pipe protocol that lock down today's
-/// behavior so subsequent phases (quiet output device under pipe mode, protocol 1.0.1, live log
+/// behavior so subsequent phases (quiet output device under pipe mode, protocol 1.1.0, live log
 /// and stream forwarding) can be reviewed as targeted diffs against these assertions.
```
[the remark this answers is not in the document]
Evangelink: Good point — updated the class summary in eb13dc19e. It no longer frames the no-op output device / 1.1.0 bump as a 'subsequent phase'; it now states this PR installs the no-op device under pipe mode and bumps the protocol to 1.1.0, and that these tests lock that down (only live log / stream forwarding remain as future phases).

0101 (approved):

Issue #7557, opened 2026-03-13
`dotnet test` reports error even when exit code 8 is ignored


Lexy2:
## Version used
Microsoft.Testing.Platform 2.1.0
net10.0
NUnit 4
NUnit3TestAdapter 6.1.0

## Steps To Reproduce
Add test packages to a project but don't add any tests to it.
Run `dotnet test --ignore-exit-code 8`

Repro attached.

[mtp-bug.zip](https://github.com/user-attachments/files/25987374/mtp-bug.zip)

-->

## Expected behavior
Command succeeds, exit code is 0.
Exit code 8 is respected at the `dotnet test` level.

## Actual behavior
Command fails, exit code is 1.
Exit code 8 is respected only if you run the MTP executable itself.

## Additional context
Workaround - `dotnet test --ignore-exit-code 1;8`

Youssef1313:
This is where we detect that there is a handshake failure.

https://github.com/dotnet/sdk/blob/57a4a8a7967da5450bb07173aa5472a9896e7e28/src/Cli/dotnet/Commands/Test/MTP/TestApplicationHandler.cs#L280

We print "zero tests" message in this case here:

https://github.com/dotnet/sdk/blob/57a4a8a7967da5450bb07173aa5472a9896e7e28/src/Cli/dotnet/Commands/Test/MTP/Terminal/TerminalTestReporter.cs#L775

The message should clearly indicate that it's handshake failure. There are also two callers of handshake failures:

- When negotiation of protocol version fails (i.e, no common supported protocol versions between test app and SDK)
- When handshake isn't received at all.

These two cases should have two different messages.

PR #7566, opened 2026-03-17
Avoid setting IsTestingPlatformApplication in MTP.targets

Youssef1313:
See discussion in #7557

In short, we should leave this up to test frameworks.

- This shouldn't be a big change for MSTest/NUnit in most common/normal cases.
- It's also not a big change for TUnit as it explicitly sets IsTestingPlatformApplication to true anyways.
- It might be a problem for xunit.v3, but I don't know if we should still take this change as it makes more sense IMO. **IMPORTANT TO DISCUSS BEFORE MERGE**

### DIFF
--- a/test/UnitTests/MSTest.Analyzers.UnitTests/MSTest.Analyzers.UnitTests.csproj
+++ b/test/UnitTests/MSTest.Analyzers.UnitTests/MSTest.Analyzers.UnitTests.csproj
@@ -12,6 +12,11 @@
     <!-- This will exclude compile and runtime assets from NuGet so that they flow via ProjectReference -->
     <UseMSTestFromSource>true</UseMSTestFromSource>
     <EnableMSTestRunner>true</EnableMSTestRunner>
+
+    <!-- In real-world projects, this is not needed. -->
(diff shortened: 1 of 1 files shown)


Youssef1313:
Ping @bradwilson. This is technically a breaking change that's expected to have only an impact for xunit.v3 which doesn't have a "gated" property for choosing between VSTest/MTP. If this is merged, what will happen if you want to preserve the current behavior is:

- Next version of xunit.v3 will need to explicitly set `<IsTestingPlatformApplication>true</IsTestingPlatformApplication>` in package props/targets.
- Existing xunit.v3 versions will not preserve the original behavior whenever users update to latest MTP, and users will then need to also update xunit.v3 along with MTP.

If nothing is done, then the most likely visible impact will be for .NET 10 SDK users using the MTP-based test command, which will now not recognize xunit.v3 test projects as MTP.

Youssef1313:
And how big deal the breaking change will be for you?

Maybe to reduce the impact, you can start adding `<IsTestingPlatformApplication>true</IsTestingPlatformApplication>` explicitly in xunit props/targets in the next version, and we postpone the breaking change after majority of your users are already on a version that won't notice the break here?

PR #7573, opened 2026-03-19
Simplfy NuGet cache folder handling in integration tests

Youssef1313:


test/Utilities/Microsoft.Testing.TestInfrastructure/TestAssetFixtureBase.cs
```diff
@@ -38,15 +34,15 @@ public async Task InitializeAsync(CancellationToken cancellationToken) =>
...
             testAsset.DotnetResult = result;
             _testAssets.TryAdd(asset.ID, testAsset);
         });
 #else
         await Task.WhenAll(GetAssetsToGenerate().Select(async asset =>
         {
             TestAsset testAsset = await TestAsset.GenerateAssetAsync(asset.Name, asset.Code, _tempDirectory);
-            DotnetMuxerResult result = await DotnetCli.RunAsync($"build {testAsset.TargetAssetPath} -c Release", _nugetGlobalPackagesDirectory.Path, callerMemberName: asset.Name, cancellationToken: cancellationToken);
+            DotnetMuxerResult result = await DotnetCli.RunAsync($"build {testAsset.TargetAssetPath} -c Release", callerMemberName: asset.Name, cancellationToken: cancellationToken);
             testAsset.DotnetResult = result;
```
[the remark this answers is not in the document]
Youssef1313: Existing issue. Will handle separately.

test/IntegrationTests/Microsoft.Testing.Platform.Acceptance.IntegrationTests/Helpers/AcceptanceFixture.cs
```diff
@@ -6,16 +6,36 @@ namespace Microsoft.Testing.Platform.Acceptance.IntegrationTests;
...
+    {
+        if (s_nugetCache is not null)
+        {
+            try
+            {
+                Directory.Delete(s_nugetCache, recursive: true);
+            }
+            catch (IOException)
+            {
+            }
```
Evangelink, line 40: **Cleanup robustness**: `Directory.Delete(path, recursive: true)` can also throw `UnauthorizedAccessException` (e.g., read-only files left by NuGet) and potentially other `SystemException` variants. Consider catching a broader set of exceptions:

```csharp
catch (Exception) when (e is IOException or UnauthorizedAccessException)
{
}
```

Or use a simple `catch (Exception)` with a comment explaining best-effort cleanup, to avoid failing the test run on cleanup issues.

test/IntegrationTests/Microsoft.Testing.Platform.Acceptance.IntegrationTests/Helpers/AcceptanceFixture.cs
```diff
@@ -6,16 +6,36 @@ namespace Microsoft.Testing.Platform.Acceptance.IntegrationTests;
...
-    public static void AssemblyCleanup()
-        => NuGetGlobalPackagesFolder.Dispose();
+    public static void AssemblyCleanup(TestContext context)
+    {
+        if (s_nugetCache is not null)
+        {
+            try
+            {
+                Directory.Delete(s_nugetCache, recursive: true);
+            }
```
Evangelink, line 37: **Orphaned parent directory**: `s_nugetCache` is `TempDirectory.TestSuiteDirectory/<RandomId>/.packages`. Cleanup deletes `.packages` but leaves the `<RandomId>/` directory behind. Over multiple runs this accumulates empty directories. Consider deleting `Path.GetDirectoryName(s_nugetCache)` instead, or track the parent path separately.

test/IntegrationTests/Microsoft.Testing.Platform.Acceptance.IntegrationTests/Helpers/AcceptanceFixture.cs
```diff
@@ -6,16 +6,36 @@ namespace Microsoft.Testing.Platform.Acceptance.IntegrationTests;
...
+        // Ensure all integration tests restore packages in a centralized place other than the NuGet cache.
+        // The centralized place also changes between runs (RandomId.Next()) so that re-packaging locally works as expected.
+        // So, when running Build.cmd -pack, running test, running Build.cmd -pack again, and running test again, the latest
+        // packages should be picked.
+        // If we restore to the same place (whether or not it is the machine-wide cache), NuGet will consider restore up-to-date and
+        // will use stale packages.
+        s_nugetCache = Path.Combine(TempDirectory.TestSuiteDirectory, RandomId.Next(), ".packages");
+        Directory.CreateDirectory(s_nugetCache);
+        Environment.SetEnvironmentVariable("NUGET_PACKAGES", s_nugetCache);
     }
```
Evangelink, line 27: **Scope concern**: This `NUGET_PACKAGES` setup only covers the `Microsoft.Testing.Platform.Acceptance.IntegrationTests` assembly. `TestAssetFixtureBase` is shared infrastructure also used by `MSTest.Acceptance.IntegrationTests` and the performance runner. Does `MSTest.Acceptance.IntegrationTests` have an equivalent `AcceptanceFixture` that sets `NUGET_PACKAGES`? If not, those tests will fall back to the machine-wide NuGet cache and may pick up stale packages after re-packing — which is exactly the problem this PR aims to fix.

Consider either:
1. Ensuring all test assemblies using `TestAssetFixtureBase` have their own `[AssemblyInitialize]` with the same setup, or
2. Moving this logic into `TestAssetFixtureBase.InitializeAsync` itself (since it's the shared entry point).

test/Utilities/Automation.CLI/VSTestConsoleLocator.cs
```diff
@@ -51,14 +48,17 @@ InvalidOperationException GetExceptionForVSTestConsoleNotFound(string expectedPa
...
         {
             Assert.IsTrue(Directory.Exists(nugetPackagesFolderPath), $"Found environment variable 'NUGET_PACKAGES' and NuGet package folder '{nugetPackagesFolderPath}' should exist");
-
-            return nugetPackagesFolderPath;
+            string fullPackagePath = Path.Combine(nugetPackagesFolderPath, packageName, packageVersion);
+            if (Directory.Exists(fullPackagePath))
+            {
+                return fullPackagePath;
+            }
         }
```
Evangelink, line 62: **Silent fallback when `NUGET_PACKAGES` is set**: When `NUGET_PACKAGES` is explicitly set (as done by `AcceptanceFixture`) but the specific package/version folder doesn't exist there, this method silently falls back to `%USERPROFILE%\.nuget\packages`. This can mask a real problem — if the package *should* be in the explicit cache but isn't, silently picking it up from the user profile cache defeats the purpose of setting `NUGET_PACKAGES` in the first place.

Consider failing with a clear error when `NUGET_PACKAGES` is set but the package isn't found there, rather than falling back.

test/IntegrationTests/MSTest.Acceptance.IntegrationTests/AppDomainTests.cs
```diff
@@ -125,6 +125,10 @@ we end up with a -dev or -ci version which will lose resolution over -preview de
     <UseVSTest>true</UseVSTest>
   </PropertyGroup>
 
+  <ItemGroup>
+    <PackageDownload Include="Microsoft.TestPlatform" Version="[$MicrosoftNETTestSdkVersion$]" />
+  </ItemGroup>
+
```
Evangelink, line 131: **Consistency nit**: `AppDomainTests` adds `<PackageDownload Include="Microsoft.TestPlatform">` to ensure the VSTest platform package is pre-downloaded into the NuGet cache (needed since `VSTestConsoleLocator` looks it up at runtime). Are there other test assets targeting `.NET Framework` + `UseVSTest` that also need this `PackageDownload` added? Without it, those assets may fail at runtime if the test platform package isn't already in the cache.

test/Performance/MSTest.Performance.Runner/Steps/DotnetTrace.cs
```diff
@@ -24,8 +24,7 @@ public DotnetTrace(string arguments, string reportFileName, CompressionLevel com
 
     public async Task<Files> ExecuteAsync(BuildArtifact payload, IContext context)
     {
-        string nugetRestoreFolder = Path.Combine(payload.TestAsset.TargetAssetPath, ".packages");
-        await DotnetCli.RunAsync($"tool install --tool-path \"{payload.TestAsset.TargetAssetPath}\" dotnet-trace", nugetRestoreFolder);
+        await DotnetCli.RunAsync($"tool install --tool-path \"{payload.TestAsset.TargetAssetPath}\" dotnet-trace");
```
Evangelink, line 27: The performance runner previously used per-asset `.packages` folders. With this change, `dotnet tool install` will use whatever `NUGET_PACKAGES` is set to (or the machine-wide cache). If the perf runner doesn't have an assembly-level fixture setting `NUGET_PACKAGES` like the acceptance tests do, this may reduce hermeticity and lead to stale tool versions being picked up across local runs.

test/IntegrationTests/MSTest.Acceptance.IntegrationTests/OutputTests.cs
```diff
@@ -26,7 +26,7 @@ TestContext message
             """);
     }
 
-    public sealed class TestAssetFixture() : TestAssetFixtureBase(AcceptanceFixture.NuGetGlobalPackagesFolder)
+    public sealed class TestAssetFixture : TestAssetFixtureBase
```
Evangelink, line 29: **Inconsistent primary constructor syntax**: Some fixtures in this PR were changed to `TestAssetFixture : TestAssetFixtureBase` (no parens, e.g., `OutputTests.cs`, `ShowOutputOptionTests.cs`, `TimeoutTests.cs`), while others keep `TestAssetFixture() : TestAssetFixtureBase()`. Since the base class no longer has a parameterized constructor, the empty parens are unnecessary. Consider picking one style consistently across all files in this PR.

test/IntegrationTests/Microsoft.Testing.Platform.Acceptance.IntegrationTests/Helpers/AcceptanceFixture.cs
Youssef1313: I don't think it's worth handling additional exceptions unless they really show up in practice.

test/IntegrationTests/Microsoft.Testing.Platform.Acceptance.IntegrationTests/Helpers/AcceptanceFixture.cs
Youssef1313: It's usually not a big problem, but fixed.

test/IntegrationTests/Microsoft.Testing.Platform.Acceptance.IntegrationTests/Helpers/AcceptanceFixture.cs
Youssef1313: The class is shared with MSTest.Acceptance.IntegrationTests.

test/Utilities/Automation.CLI/VSTestConsoleLocator.cs
Youssef1313: It's fine for the purpose of locating vstest.console.

test/Performance/MSTest.Performance.Runner/Steps/DotnetTrace.cs
Youssef1313: I don't think it's a problem.

test/IntegrationTests/MSTest.Acceptance.IntegrationTests/OutputTests.cs
Youssef1313: This is very minimal stylistic thing. We can address later.

test/IntegrationTests/MSTest.Acceptance.IntegrationTests/AppDomainTests.cs
Youssef1313: There is none.

Evangelink (approved):

PR #7590, opened 2026-03-23
MTP: Allow test results to have separate display name while keeping the original info for TestNode.DisplayName

Youssef1313:
TODOs:

- [ ] Add integration tests for TRX for a folded MSTest parameterized test, and assert proper test definition.
- [ ] Revise TRX implementation for any places with DisplayName and confirm if it's the display name of the test result (TestNodeUpdateMessage) or if it's the display name of the test case.


Youssef1313:
After discussion with @Evangelink, we will address it differently and changes will be specific to TRX. I'll prepare another PR.

PR #7592, opened 2026-03-23
Fix MSTEST0049 false positives inside expression trees (#7585)

### LINKED ISSUE #7585: MSTEST0049 False positive with expressions of Moq
## Describe the bug

When mocking an API, if an overload with the cancellation token exists, the rule will trigger

## Version used

Latest

## Steps To Reproduce

See https://github.com/microsoft/vstest/pull/15539/changes#diff-b72c57420168ad8dde36e79c1710bf4211a778e9c356582e8ee287bbee45041dR4

Evangelink:
Skip reporting MSTEST0049 diagnostics when the invocation is inside an expression tree (e.g. Moq Setup lambdas), where the code fix cannot be meaningfully applied.

- Add IsInsideExpressionTree helper that walks up the operation tree to detect lambdas converted to Expression<T>
- Add SystemLinqExpressionsLambdaExpression to WellKnownTypeNames
- Add 3 tests covering expression tree suppression and regular lambda non-suppression

Fixes #7585

### DIFF
--- a/src/Analyzers/MSTest.Analyzers/FlowTestContextCancellationTokenAnalyzer.cs
+++ b/src/Analyzers/MSTest.Analyzers/FlowTestContextCancellationTokenAnalyzer.cs
@@ -57,10 +57,10 @@ public sealed class FlowTestContextCancellationTokenAnalyzer : DiagnosticAnalyze
                 return;
             }
 
-            context.Compilation.TryGetOrCreateTypeByMetadataName(WellKnownTypeNames.SystemLinqExpressionsLambdaExpression, out INamedTypeSymbol? lambdaExpressionSymbol);
+            context.Compilation.TryGetOrCreateTypeByMetadataName(WellKnownTypeNames.SystemLinqExpressionsExpression1, out INamedTypeSymbol? linqExpressionType);
 
             context.RegisterOperationAction(
-                context => AnalyzeInvocation(context, cancellationTokenSymbol, testContextSymbol, classCleanupAttributeSymbol, assemblyCleanupAttributeSymbol, testMethodAttributeSymbol, lambdaExpressionSymbol),
+                context => AnalyzeInvocation(context, cancellationTokenSymbol, testContextSymbol, classCleanupAttributeSymbol, assemblyCleanupAttributeSymbol, testMethodAttributeSymbol, linqExpressionType),
                 OperationKind.Invocation);
         });
     }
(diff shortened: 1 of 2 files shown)


src/Analyzers/MSTest.Analyzers/FlowTestContextCancellationTokenAnalyzer.cs
```diff
@@ -138,6 +153,33 @@ private static void AnalyzeInvocation(
...
+    {
+        if (lambdaExpressionSymbol is null)
+        {
+            return false;
+        }
+
+        IOperation? current = operation.Parent;
+        while (current is not null)
+        {
+            if (current is IAnonymousFunctionOperation)
```
Youssef1313: https://github.com/dotnet/sdk/blob/f25f3107b3554e141cbd09113cccf6c321f45143/src/Microsoft.CodeAnalysis.NetAnalyzers/src/Utilities/Compiler/Extensions/IOperationExtensions.cs#L552-L565

Looks like the implementation in SDK (moved from roslyn-analyzers) walks up to either LocalFunction or AnonymousFunction?

src/Analyzers/MSTest.Analyzers/FlowTestContextCancellationTokenAnalyzer.cs
```diff
@@ -138,6 +153,33 @@ private static void AnalyzeInvocation(
...
+            {
+                // Check if the parent converts this lambda to an Expression<T>.
+                if (current.Parent is IConversionOperation conversion &&
```
Youssef1313: Implementation referenced in my earlier comment doesn't check IConversionOperation, it just accesses `Type` right away.

src/Analyzers/MSTest.Analyzers/FlowTestContextCancellationTokenAnalyzer.cs
```diff
@@ -138,6 +153,33 @@ private static void AnalyzeInvocation(
...
+                    conversion.Type is INamedTypeSymbol namedType &&
+                    namedType.Inherits(lambdaExpressionSymbol))
```
Youssef1313: No personal preference for me. But implementation from sdk uses `Expression<T>` instead of `LambdaExpression`, and checks for exact type instead of inheritance. I think today, `Expression<T>` is the only known type that inherits `LambdaExpression`.

src/Analyzers/MSTest.Analyzers/FlowTestContextCancellationTokenAnalyzer.cs
```diff
@@ -138,6 +153,33 @@ private static void AnalyzeInvocation(
...
+                {
+                    return true;
+                }
```
Youssef1313: We should probably stop walking up once we find anonymous function.

src/Analyzers/MSTest.Analyzers/FlowTestContextCancellationTokenAnalyzer.cs
Evangelink: Thanks 👍

PR #7619, opened 2026-03-27
Try MSBuild's `/Isolate`

Youssef1313:

### DIFF
--- a/Directory.Build.rsp
+++ b/Directory.Build.rsp
@@ -1,0 +1,1 @@
+/Isolate


ViktorHofer:
Yeah Arcade seems to be in-compatible with /graph or /isolate. You could try to change that though. Additionally, you could also always experiment with just building the repo with `dotnet build` directly instead of going through eng/common/build.cmd/sh.

Youssef1313:
Closing for now. Not planned to be investigated soon-ish.

PR #7637, opened 2026-03-31
Revert "Avoid printing stdout/stderr properties in console (#7600)" and add --show-stdout/--show-stderr options

### LINKED ISSUE #7612: Revise recent output change before shipping 4.2
VSTest might be showing the stdout/stderr properties for passed tests when running with `--logger "console;verbosity=detailed"`, and always show them for failing.

In that case we should probably do the same.

We should revisit the change in #7600 before we ship it.

Youssef1313:
Fixes #7612

### DIFF
--- a/src/Platform/Microsoft.Testing.Platform/OutputDevice/Terminal/TerminalTestReporter.cs
+++ b/src/Platform/Microsoft.Testing.Platform/OutputDevice/Terminal/TerminalTestReporter.cs
@@ -334,7 +334,7 @@ internal sealed partial class TerminalTestReporter : IDisposable
        string? expected,
        string? actual,
        string? standardOutput,
-       string? standardError)
+       string? errorOutput)
     {
         FlatException[] flatExceptions = ExceptionFlattener.Flatten(errorMessage, exception);
         TestCompleted(
@@ -347,7 +347,7 @@ internal sealed partial class TerminalTestReporter : IDisposable
             expected,
             actual,
             standardOutput,
-            standardError);
+            errorOutput);
     }
 
     private void TestCompleted(
@@ -360,7 +360,7 @@ internal sealed partial class TerminalTestReporter : IDisposable
         string? expected,
         string? actual,
         string? standardOutput,
-        string? standardError)
+        string? errorOutput)
     {
         if (_testProgressState is null)
         {
@@ -406,7 +406,7 @@ internal sealed partial class TerminalTestReporter : IDisposable
                 expected,
                 actual,
                 standardOutput,
-                standardError));
+                errorOutput));
         }
     }
 
@@ -426,7 +426,7 @@ internal sealed partial class TerminalTestReporter : IDisposable
         string? expected,
         string? actual,
         string? standardOutput,
-        string? standardError)
+        string? errorOutput)
     {
         if (outcome == TestOutcome.Passed && !GetShowPassedTests())
         {
@@ -479,8 +479,8 @@ internal sealed partial class TerminalTestReporter : IDisposable
         };
         string? stderrToShow = _options.ShowStderr switch
         {
-            OutputShowMode.All => standardError,
-            OutputShowMode.Failed => isFailed ? standardError : null,
+            OutputShowMode.All => errorOutput,
+            OutputShowMode.Failed => isFailed ? errorOutput : null,
             OutputShowMode.None => null,
             _ => throw ApplicationStateGuard.Unreachable(),
         };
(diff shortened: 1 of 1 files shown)


src/Platform/Microsoft.Testing.Platform/OutputDevice/Terminal/TerminalTestReporter.cs
```diff
@@ -332,7 +332,9 @@ internal void TestCompleted(
        string? errorMessage,
        Exception? exception,
        string? expected,
-       string? actual)
+       string? actual,
+       string? standardOutput,
+       string? errorOutput)
```
Evangelink: Nit: The parameter is called `errorOutput` here but `standardError` everywhere else in the chain — in `TerminalOutputDevice.ConsumeAsync` (the local variable), in `FormatStandardAndErrorOutput` (the parameter), and in the underlying `StandardErrorProperty.StandardError`. It's also easy to confuse with `errorMessage` (the test failure explanation).

Consider renaming to `standardError` across the three method signatures (`TestCompleted` internal, `TestCompleted` private, `RenderTestCompleted`) and in the switch expression body in `RenderTestCompleted`.

src/Platform/Microsoft.Testing.Platform/OutputDevice/Terminal/TerminalTestReporter.cs
```diff
@@ -458,6 +468,23 @@ private void RenderTestCompleted(
...
+            _ => standardOutput,
+        };
+        string? stderrToShow = _options.ShowStderr switch
+        {
+            OutputShowMode.All => errorOutput,
+            OutputShowMode.Failed => isFailed ? errorOutput : null,
+            OutputShowMode.None => null,
+            _ => errorOutput,
+        };
+        FormatStandardAndErrorOutput(terminal, stdoutToShow, stderrToShow);
```
Evangelink, line 487: `OutputShowMode` has exactly 3 values (`All`, `Failed`, `None`), all handled above. The `_ =>` default arm is unreachable dead code. Elsewhere in this file (e.g. the `TestOutcome` switch), unreachable paths use `throw ApplicationStateGuard.Unreachable()`. Same treatment should be applied here for both switch expressions:

```suggestion
            _ => throw ApplicationStateGuard.Unreachable(),
```

src/Platform/Microsoft.Testing.Platform/OutputDevice/Terminal/TerminalTestReporterOptions.cs
```diff
@@ -32,6 +32,34 @@ internal sealed class TerminalTestReporterOptions
...
     /// </summary>
     public AnsiMode AnsiMode { get; init; }
+
+    /// <summary>
+    /// Gets a value indicating when to show standard output.
+    /// </summary>
+    public OutputShowMode ShowStdout { get; init; } = OutputShowMode.All;
+
+    /// <summary>
+    /// Gets a value indicating when to show standard error output.
```
Evangelink, line 42: We discussed whether `Failed` would be a better default since it directly addresses the original complaint in #7612 (polluted output). We agreed `Failed` is the preferred default long-term, but changing it now would be disruptive for users who rely on seeing output for all tests. Keeping `All` as the default for now and planning to switch to `Failed` in the next major version.

test/IntegrationTests/MSTest.Acceptance.IntegrationTests/ShowOutputOptionTests.cs
```diff
@@ -0,0 +1,120 @@
...
+        public override (string ID, string Name, string Code) GetAssetsToGenerate() => (ProjectName, ProjectName,
+                SourceCode
+                .PatchTargetFrameworks(TargetFrameworks.All)
+                .PatchCodeWithReplace("$MSTestVersion$", MSTestVersion));
+
+        private const string SourceCode = """
+#file ShowOutputOptionTest.csproj
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
```
Evangelink, line 160: The acceptance tests cover `none` and `failed` modes but are missing coverage for:

1. **`all` (explicit)** — verify `--show-stdout all` shows output for both passing and failing tests
2. **Default (unspecified)** — verify that omitting the option behaves the same as `all`
3. **Invalid argument** — verify `--show-stdout invalid` produces the expected `TerminalShowOutputOptionInvalidArgument` error message

These would strengthen confidence in the feature, especially the invalid argument case which exercises the validation path.

Youssef1313: @copilot Add tests for this.

src/Platform/Microsoft.Testing.Platform/OutputDevice/Terminal/TerminalTestReporter.cs
Youssef1313: Will rename later. This is pre-existing naming.

Evangelink (approved):

Issue #7646, opened 2026-04-01
IndexOutOfRangeException in the command line parsor when run from GitHub Actions.


AraHaan:
## Describe the bug

Currently in one of my repositories I tried to get the MTP to work with ``xunit.v3`` when running with .NET SDK 10.0.200 in GitHub Actions, I saw someone was able to use it with 10.0.201 so I tried that on my repository. However this happens on my repository as soon as my action goes to run ``dotnet build -c Release --no-build`` on the tests which runs the tests and results in a failure to run the identify & run my tests.

Locally when I use my .NET 11 daily build SDK I am unable to reproduce this issue and it is annoying me greatly.

The failed build: https://github.com/Elskom/IDisposableGenerator/actions/runs/23832789717/job/69469814486?pr=51
the PR: https://github.com/Elskom/IDisposableGenerator/pull/51

## Version used

The latest version that ``xunit.v3`` references and the one that ``.NET 10 SDK v10.0.201`` uses.

## Steps To Reproduce

1. add reference to ``xunit.v3``.
2. add a ``global.json`` which only enables MTP.
3. Reference ``10.0.201`` .NET SDK with my GitHub Action:
```yaml
      - name: Install latest .NET 10 SDK
        uses: Elskom/setup-latest-dotnet@main
        with:
          SDK_VERSION: '10.0.201'
          RUNTIME_VERSIONS: ''
```

## Expected behavior
For the tests to run and pass.

## Actual behavior
no tests are run at all.

## Additional context

The GitHub Action I made basically uses the dotnet-install scripts to install the .NET SDK I want into GitHub actions which is great for when I sometimes want to include the daily preview .NET SDK builds in some of my CI workflows.

PR #7648, opened 2026-04-01
Avoid IndexOutOfRangeException in command-line parsing

Youssef1313:
Fixes https://github.com/microsoft/testfx/issues/7646

### DIFF
--- a/test/UnitTests/Microsoft.Testing.Platform.UnitTests/CommandLine/ArgumentArityTests.cs
+++ b/test/UnitTests/Microsoft.Testing.Platform.UnitTests/CommandLine/ArgumentArityTests.cs
@@ -24,7 +24,7 @@ public sealed class ArgumentArityTests
     ];
 
     [TestMethod]
-    public async Task ParseAndValidate_EmptyOption_ShouldNotThrowException()
+    public async Task ParseAndValidate_EmptyArgument_ShouldNotThrowException()
     {
         // Arrange
         string[] args = [string.Empty];


src/Platform/Microsoft.Testing.Platform/CommandLine/Parser.cs
```diff
@@ -54,7 +54,7 @@ private static CommandLineParseResult Parse(List<string> args, IEnvironment envi
 
             // If it's the first argument and it doesn't start with - then it's the tool name
             // TODO: This won't work correctly if the first argument provided is a response file that contains the tool name.
-            if (isFirstRealArgument && currentArg[0] != '-')
+            if (isFirstRealArgument && currentArg.Length > 0 && currentArg[0] != '-')
```
Evangelink, line 57: The production fix is correct and minimal — this was the only unguarded indexing on `currentArg` in the method. The `StartsWith('@')` above is safe on empty strings, and the length checks on lines 68-69 (`Length > 1`, `Length > 2`) already had proper guards. 👍

test/UnitTests/Microsoft.Testing.Platform.UnitTests/CommandLine/ArgumentArityTests.cs
```diff
@@ -23,6 +23,28 @@ public sealed class ArgumentArityTests
...
+            _extensionCommandLineOptionsProviders, new Mock<ICommandLineOptions>().Object);
+
+        // Assert
+        Assert.IsFalse(result.IsValid);
+#pragma warning disable SA1027 // Use tabs correctly
+        Assert.AreEqual(
+            """
+            Invalid command line arguments:
+            	- Unexpected argument 
+            """, result.ErrorMessage, StringComparer.Ordinal);
```
Evangelink, line 44: Per repo conventions in `.github/copilot-instructions.md`:

> "All assertions must be written using FluentAssertions style of assertion."

The new test uses `Assert.IsFalse` and `Assert.AreEqual`. Consider switching to:
```csharp
result.IsValid.Should().BeFalse();
result.ErrorMessage.Should().Contain("Unexpected argument");
```

Using `Should().Contain(...)` instead of exact multi-line match would also avoid the `#pragma disable SA1027` and make the test more resilient to harmless formatting changes in the error message (indentation, newlines, trailing space).

(The existing tests in the same file also use `Assert.*` — that's a pre-existing inconsistency, but a new test is a good opportunity to follow the stated convention.)

test/UnitTests/Microsoft.Testing.Platform.UnitTests/CommandLine/ArgumentArityTests.cs
```diff
@@ -23,6 +23,28 @@ public sealed class ArgumentArityTests
...
+        string[] args = [string.Empty];
+        CommandLineParseResult parseResult = CommandLineParser.Parse(args, new SystemEnvironment());
+
+        // Act
+        ValidationResult result = await CommandLineOptionsValidator.ValidateAsync(parseResult, _systemCommandLineOptionsProviders,
+            _extensionCommandLineOptionsProviders, new Mock<ICommandLineOptions>().Object);
+
+        // Assert
+        Assert.IsFalse(result.IsValid);
+#pragma warning disable SA1027 // Use tabs correctly
```
Evangelink, line 39: Disabling SA1027 to embed a literal tab in the raw string is a bit of code smell in a test. If you do keep the exact-match assertion, you can avoid the pragma by constructing the expected string with `"\t"`:

```csharp
var expected = string.Join(
    Environment.NewLine,
    "Invalid command line arguments:",
    "\t- Unexpected argument ",
    string.Empty);
result.ErrorMessage.Should().Be(expected);
```

Though as noted above, a `Should().Contain("Unexpected argument")` assertion would sidestep this entirely and still validate the regression.

PR #7649, opened 2026-04-01
Avoid ANSI and progress output when running in LLM environment

### LINKED ISSUE #7647: Add "no progress" and "no ANSI" environment variables
## Summary

Allow disabling progress and ANSI using environment variables.

## Background and Motivation

When Copilot runs tests, displaying progress and ANSI are just a waste of tokens. It's possible to configure Copilot's terminal using `chat.tools.terminal.terminalProfile.<os>` and define environment variables. For example:

```json
{
  "chat.tools.terminal.terminalProfile.windows": {
    "icon": "copilot",
    "path": "pwsh",
    "env": {
      "MSBUILDTERMINALLOGGER": "off"
    }
  }
}
```

## Proposed Feature

Add `TESTINGPLATFORM_NOPROGRESS` and `TESTINGPLATFORM_NOANSI` as defaults (overridable by CLI args).

Youssef1313:
Fixes #7647

### DIFF
--- a/src/Platform/Microsoft.Testing.Platform/OutputDevice/TerminalOutputDevice.cs
+++ b/src/Platform/Microsoft.Testing.Platform/OutputDevice/TerminalOutputDevice.cs
@@ -126,7 +126,7 @@ internal sealed partial class TerminalOutputDevice : IHotReloadPlatformOutputDev
 
         AnsiMode ansiMode = AnsiMode.AnsiIfPossible;
         // In LLM environments, prefer simple text output so that LLM can parse it easily.
-        // Note that NoAnsi also implies also no progress.
+        // Note that NoAnsi also implies no progress.
         if (noAnsi || LLMEnvironmentDetector.IsLLMEnvironment())
         {
             // User explicitly specified --no-ansi.
@@ -152,7 +152,7 @@ internal sealed partial class TerminalOutputDevice : IHotReloadPlatformOutputDev
             showPassed = () => true;
         }
 
-        Func<bool?> shouldShowProgress = noProgress || ansiMode is AnsiMode.NoAnsi or AnsiMode.SimpleAnsi || LLMEnvironmentDetector.IsLLMEnvironment()
+        Func<bool?> shouldShowProgress = noProgress || ansiMode is AnsiMode.NoAnsi or AnsiMode.SimpleAnsi
             // User preference is to not show progress.
             // Or, we are in terminal that's not capable of changing cursor and we can't update progress in-place.
             // In that case, we force disable progress as well.


src/Platform/Microsoft.Testing.Platform/Helpers/LLMEnvironmentDetector.cs
```diff
@@ -0,0 +1,193 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+namespace Microsoft.Testing.Platform.Helpers;
+
+// Copy from https://github.com/dotnet/sdk/tree/1e5d8e39d3026edb222cdf4f8d8240f1eb99f24b/src/Cli/Microsoft.DotNet.Cli.Definitions/Telemetry
+internal static class LLMEnvironmentDetector
+{
```
Evangelink, line 8: This class calls `Environment.GetEnvironmentVariable()` directly (suppressing RS0030 each time), but the project's [`BannedSymbols.txt`](https://github.com/microsoft/testfx/blob/main/src/Platform/Microsoft.Testing.Platform/BannedSymbols.txt) bans the **entire `System.Environment` type** with the directive _"Use 'IEnvironment' instead"_.

`TerminalOutputDevice` (the caller of this class) already properly uses the `IEnvironment` abstraction via constructor injection. `CIEnvironmentDetector` in TestFramework follows the same DI pattern.

Additionally, the static design (`static class` with a static auto-property initialized at class load) makes this completely untestable — you can't mock environment variables without actually setting them in the process, and the result is frozen for the process lifetime.

**Suggestion:** Follow the `CIEnvironmentDetector` pattern — make this a non-static class with constructor-injected `IEnvironment`. Provide a singleton default via a static `Instance` property if convenient static access is still desired. This aligns with the architectural conventions and enables unit testing.

src/Platform/Microsoft.Testing.Platform/Helpers/LLMEnvironmentDetector.cs
```diff
@@ -0,0 +1,193 @@
...
+        new EnvironmentDetectionRuleWithResult<string>("windsurf", new AnyPresentEnvironmentRule("WINDSURF_SESSION")),
+        // (proposed) generic flag for Agentic usage
+        new EnvironmentDetectionRuleWithResult<string>("generic_agent", new BooleanEnvironmentRule("AGENT_CLI")),
+    ];
+
+    private static string? LLMEnvironment { get; } = GetLLMEnvironment();
+
+    private static string? GetLLMEnvironment()
+    {
+        string?[] results = DetectionRules.Select(r => r.GetResult()).Where(r => r != null).ToArray();
```
Evangelink, line 57: `GetLLMEnvironment()` builds a comma-separated string of all matched LLM environment names, but `IsLLMEnvironment()` only checks `!RoslynString.IsNullOrEmpty(...)` — the actual environment names are never exposed, logged, or returned to any caller. The generic `EnvironmentDetectionRuleWithResult<T>` adds complexity for a feature (returning _which_ LLM was detected) that isn't currently used.

If the string is intended for future telemetry/logging, consider adding a comment documenting that intent. Otherwise, this could be simplified to a `bool` check that short-circuits on the first match (avoiding unnecessary allocations from `Select` + `Where` + `ToArray` + `string.Join`).

src/Platform/Microsoft.Testing.Platform/Helpers/LLMEnvironmentDetector.cs
```diff
@@ -0,0 +1,193 @@
...
+    }
+
+    /// <summary>
+    /// Rule that matches when any of the specified environment variables is set to "true".
+    /// </summary>
+    private sealed class BooleanEnvironmentRule : EnvironmentDetectionRule
+    {
+        private readonly string[] _variables;
+
+        public BooleanEnvironmentRule(params string[] variables)
```
Evangelink, line 82: The constructors use `?? throw new ArgumentNullException(nameof(...))` but the project's `BannedSymbols.txt` bans `ArgumentNullException.#ctor` in favor of `ArgumentGuard`.

Since this is a copy from dotnet/sdk, the deviation may be intentional to keep the code synchronized — but worth calling out for awareness. If strict alignment with project conventions is preferred, these should use `ArgumentGuard`.

src/Platform/Microsoft.Testing.Platform/OutputDevice/TerminalOutputDevice.cs
```diff
@@ -125,7 +125,9 @@ await _policiesService.RegisterOnAbortCallbackAsync(
         bool inCI = string.Equals(_environment.GetEnvironmentVariable("TF_BUILD"), "true", StringComparison.OrdinalIgnoreCase) || string.Equals(_environment.GetEnvironmentVariable("GITHUB_ACTIONS"), "true", StringComparison.OrdinalIgnoreCase);
 
         AnsiMode ansiMode = AnsiMode.AnsiIfPossible;
-        if (noAnsi)
+        // In LLM environments, prefer simple text output so that LLM can parse it easily.
+        // Note that NoAnsi also implies no progress.
+        if (noAnsi || LLMEnvironmentDetector.IsLLMEnvironment())
         {
```
Evangelink, line 131: The comment inside this `if` block still reads _"User explicitly specified --no-ansi. We should respect that."_ but the condition now also covers auto-detected LLM environments. Consider updating to reflect both paths, e.g.:

```csharp
// User explicitly specified --no-ansi, or an LLM environment was auto-detected.
// Use plain text output for simpler parsing.
```

(This was flagged as a suppressed low-confidence comment by the automated reviewer — I agree it should be addressed.)

src/Platform/Microsoft.Testing.Platform/OutputDevice/TerminalOutputDevice.cs
Evangelink, line 1: The linked issue (#7647) specifically requests adding `TESTINGPLATFORM_NOPROGRESS` and `TESTINGPLATFORM_NOANSI` environment variables as defaults (overridable by CLI args). This PR takes a different approach (auto-detection via heuristics). Both approaches are complementary — auto-detection is great for zero-config UX, but explicit env vars give users deterministic control regardless of whether their tool is in the detection list.

Consider whether the env var support from #7647 should also be included here, or if that's deferred to a follow-up. The naming convention would fit naturally alongside existing vars in `EnvironmentVariableConstants.cs` (`TESTINGPLATFORM_NOBANNER`, `TESTINGPLATFORM_DIAGNOSTIC`, etc.).

PR #7700, opened 2026-04-09
Cleanup and simplify hang dump implementation

Youssef1313:


src/Platform/Microsoft.Testing.Extensions.HangDump/HangDumpExtensions.cs
```diff
@@ -34,11 +33,7 @@ public static void AddHangDumpProvider(this ITestApplicationBuilder builder)
             throw new PlatformNotSupportedException("Hang dump extension is not available on ios nor tvos");
         }
 
-        var environment = new SystemEnvironment();
-        CurrentTestApplicationModuleInfo testApplicationModuleInfo = new(environment, new SystemProcessHandler());
-        string namedPipeSuffix = Guid.NewGuid().ToString("N");
-        PipeNameDescription pipeNameDescription = NamedPipeServer.GetPipeName(Guid.NewGuid().ToString("N"), environment);
-        HangDumpConfiguration hangDumpConfiguration = new(testApplicationModuleInfo, pipeNameDescription, namedPipeSuffix);
+        PipeNameDescription pipeNameDescription = NamedPipeServer.GetPipeName(Guid.NewGuid().ToString("N"));
```
Evangelink, line 36: Nit: `HangDumpActivityIndicator` still calls the 2-parameter compatibility overload `NamedPipeServer.GetPipeName(Guid.NewGuid().ToString("N"), _environment)` on line 94 (not changed in this PR), while this call now uses the 1-parameter overload. The 2-parameter overload is a compat shim that discards `IEnvironment`. Consider updating the other call site too for consistency (can be a follow-up).

src/Platform/Microsoft.Testing.Extensions.HangDump/HangDumpActivityIndicator.cs
```diff
@@ -117,13 +111,8 @@ private async Task<IResponse> CallbackAsync(IRequest request)
         if (request is GetInProgressTestsRequest)
         {
             await _logger.LogDebugAsync($"Received '{nameof(GetInProgressTestsRequest)}'").ConfigureAwait(false);
-            return new GetInProgressTestsResponse([.. _testsCurrentExecutionState.Select(x => (x.Value.Name, (int)_clock.UtcNow.Subtract(x.Value.StartTime).TotalSeconds))]);
-        }
-        else if (request is ExitSignalActivityIndicatorTaskRequest)
-        {
-            await _logger.LogDebugAsync($"Received '{nameof(ExitSignalActivityIndicatorTaskRequest)}'").ConfigureAwait(false);
             _exitSignalActivityIndicatorAsync = true;
```
Evangelink, line 114: The `GetInProgressTestsRequest` handler now carries the side-effect of setting `_exitSignalActivityIndicatorAsync = true` (stopping the heartbeat). This is fine — the request is only sent when the controller detects a hang timeout and is about to dump+kill, so coupling the "stop heartbeat" signal with the "get in-progress tests" query is reasonable and removes a network round-trip. Just noting this for future reference: if the protocol changes and `GetInProgressTestsRequest` is ever sent for informational purposes (without a hang), this side-effect would need to be revisited.

src/Platform/Microsoft.Testing.Extensions.HangDump/HangDumpEnvironmentVariableProvider.cs
```diff
@@ -33,40 +35,24 @@ public HangDumpEnvironmentVariableProvider(ICommandLineOptions commandLineOption
     public Task UpdateAsync(IEnvironmentVariables environmentVariables)
     {
         environmentVariables.SetVariable(
-            new(_hangDumpConfiguration.PipeNameKey, _hangDumpConfiguration.PipeNameValue, false, true));
-        environmentVariables.SetVariable(
-            new(HangDumpConfiguration.NamedPipeNameSuffixEnvironmentVariable, _hangDumpConfiguration.NamedPipeSuffix, false, true));
+            new EnvironmentVariable(PipeNameEnvironmentVariableName, _pipeName, isSecret: false, isLocked: true));
```
Evangelink, line 38: Nice improvement: using named parameters (`isSecret: false, isLocked: true`) is much clearer than the old positional `new(..., false, true)`. And collapsing 2 env vars into 1 makes both set and validation far simpler.

src/Platform/Microsoft.Testing.Extensions.HangDump/DESIGN.md
Evangelink, line 1: Excellent addition. This makes the dual-pipe handshake understandable at a glance. Step 3 in the visualization nicely documents the side-effect merge (GetInProgressTestsRequest now also stops the heartbeat).

src/Platform/Microsoft.Testing.Extensions.HangDump/Serializers/ExitSignalActivityIndicatorTaskRequest.cs
```diff
@@ -1,20 +0,0 @@
...
-using Microsoft.Testing.Platform.IPC;
-using Microsoft.Testing.Platform.IPC.Serializers;
-
-namespace Microsoft.Testing.Extensions.HangDump.Serializers;
-
-internal sealed class ExitSignalActivityIndicatorTaskRequest : IRequest;
-
-internal sealed class ExitSignalActivityIndicatorTaskRequestSerializer : BaseSerializer, INamedPipeSerializer
-{
-    public int Id => 6;
```
Evangelink, line 13: The removed `ExitSignalActivityIndicatorTaskRequestSerializer` had `Id => 6`. The remaining serializer IDs are 3, 4, 5, 7 — leaving a gap at 6. This is harmless since IDs don't need to be contiguous, but ID 6 should be considered "retired" if new messages are added later to avoid accidental reuse.

Evangelink (approved):
Clean, well-scoped simplification. The old hash-based env var naming (`TESTINGPLATFORM_HANGDUMP_PIPENAME_<hash>_<suffix>` + a separate suffix env var) was unnecessary complexity — the pipe name itself is already a unique GUID. Collapsing to a single `TESTINGPLATFORM_HANGDUMP_PIPENAME` env var, removing `HangDumpConfiguration`, `FNV_1aHashHelper`, and `ExitSignalActivityIndicatorTaskRequest` removes 3 files and ~73 net lines while making the code easier to follow. The `DESIGN.md` is a valuable addition.

One minor nit about inconsistent `GetPipeName` overload usage (see inline). No blocking issues — all changed types are `internal`, so no public API surface concerns.

src/Platform/Microsoft.Testing.Extensions.HangDump/HangDumpExtensions.cs
Youssef1313: Can clean this up later.

src/Platform/Microsoft.Testing.Extensions.HangDump/Serializers/ExitSignalActivityIndicatorTaskRequest.cs
Youssef1313: The gap is fine. And reusing the ID 6 is also fine if we wanted to reuse.

PR #7720, opened 2026-04-13
Use source-only filter package in VSTestBridge

Youssef1313:


Evangelink (commented):
Is the code fully equivalent? I recall we had to do some tricks for MTP and I think (but not 100% sure) that we also fixed something on our end because it was not done on VSTest side.

Youssef1313:
I don't have any history around that and not aware of any differences between the implementation of MTP and VSTest.

Evangelink (approved):

PR #7731, opened 2026-04-16
Use PipeOptions.Asynchronous for our NamedPipeClientStream

Youssef1313:


src/Platform/Microsoft.Testing.Platform/IPC/NamedPipeClient.cs
```diff
@@ -16,7 +16,7 @@ namespace Microsoft.Testing.Platform.IPC;
 [UnsupportedOSPlatform("browser")]
 internal sealed class NamedPipeClient : NamedPipeBase, IClient
 {
-    private const PipeOptions CurrentUserPipeOptions = PipeOptions.None
+    private const PipeOptions CurrentUserPipeOptions = PipeOptions.Asynchronous
```
Evangelink: Nit (non-blocking): Now that this includes `PipeOptions.Asynchronous`, the constant name `CurrentUserPipeOptions` is slightly misleading — the server-side counterpart is already named `AsyncCurrentUserPipeOptions`. Consider renaming to match for consistency. Fine to do in a follow-up.

Evangelink (approved):
**LGTM** — correct and important fix.

Without `PipeOptions.Asynchronous`, the `NamedPipeClientStream` opens the OS handle in synchronous mode, so every `ReadAsync`/`WriteAsync`/`ConnectAsync` call silently blocks a thread pool thread instead of using IOCP (Windows) or native async I/O (Unix). The server side (`NamedPipeServer`) already uses `PipeOptions.Asynchronous` — the client was the only side missing it.

Low risk: `NamedPipeClient` is `internal sealed`, all async call sites already use `ConfigureAwait(false)`, and CI is green across all platforms. The synchronous `WaitForPipeDrain()` call works fine on overlapped handles (same pattern as the server). Only nit is the constant naming inconsistency with the server side (`CurrentUserPipeOptions` vs `AsyncCurrentUserPipeOptions`).

PR #7732, opened 2026-04-16
Cleanup TimeoutHelper and increase the timeout to 5 mins

Youssef1313:
Especially for Code Coverage, connecting to the pipe may take long. I think it's safe to increase the timeout, or even remove it completely (which I personally would vote for, but I decided to make the more minimal change in this PR).

### DIFF
--- a/src/Platform/Microsoft.Testing.Platform/Helpers/TimeSpanParser.cs
+++ b/src/Platform/Microsoft.Testing.Platform/Helpers/TimeSpanParser.cs
@@ -24,7 +24,7 @@ internal static partial class TimeSpanParser
         if (RoslynString.IsNullOrWhiteSpace(time))
         {
             result = TimeSpan.Zero;
-            return true;
+            return false;
         }
 
         Match match = Pattern.Match(time);
--- a/src/Platform/Microsoft.Testing.Platform/Helpers/TimeoutHelper.cs
+++ b/src/Platform/Microsoft.Testing.Platform/Helpers/TimeoutHelper.cs
@@ -19,10 +19,10 @@ internal static class TimeoutHelper
             ? customHangTimeout
             : TimeSpan.FromMinutes(5);
 
-        DefaultHangTimeoutSeconds = (int)DefaultHangTimeSpanTimeout.TotalSeconds;
+        DefaultHangTimeoutSeconds = DefaultHangTimeSpanTimeout.TotalSeconds;
     }
 
-    public static int DefaultHangTimeoutSeconds { get; private set; }
+    public static double DefaultHangTimeoutSeconds { get; }
 
     public static TimeSpan DefaultHangTimeSpanTimeout { get; }
 }
--- a/src/Platform/Microsoft.Testing.Platform/Helpers/TimeoutHelper.cs
+++ b/src/Platform/Microsoft.Testing.Platform/Helpers/TimeoutHelper.cs
@@ -14,26 +14,13 @@ internal static class TimeoutHelper
         string? customAntiHangTimeout = Environment.GetEnvironmentVariable(EnvironmentVariableConstants.TESTINGPLATFORM_DEFAULT_HANG_TIMEOUT);
 #pragma warning restore RS0030 // Do not use banned APIs
 
-        // Took from Tpv2 experience, it's the timeout for the testhost.exe start.
-        DefaultHangTimeSpanTimeout = TimeSpan.FromSeconds(90);
-
-        if (customAntiHangTimeout is not null)
-        {
-            if (TimeSpanParser.TryParse(customAntiHangTimeout, out TimeSpan customHangTimeout))
-            {
-                DefaultHangTimeSpanTimeout = customHangTimeout;
-            }
-        }
+        DefaultHangTimeSpanTimeout =
+            TimeSpanParser.TryParse(customAntiHangTimeout, out TimeSpan customHangTimeout)
+            ? customHangTimeout
+            : TimeSpan.FromMinutes(5);
 
         DefaultHangTimeoutSeconds = (int)DefaultHangTimeSpanTimeout.TotalSeconds;
-        DefaultHangTimeoutMilliseconds = (int)DefaultHangTimeSpanTimeout.TotalMilliseconds;
     }
-
-    /// <summary>
-    /// Gets defaultAntiHangTimeout* values are used as timeout for every wait operation in the test platform.
-    /// Are not intended for any timeout logic, but only to avoid infinite waits in case of test platform hangs.
-    /// </summary>
-    public static int DefaultHangTimeoutMilliseconds { get; private set; }
 
     public static int DefaultHangTimeoutSeconds { get; private set; }
 
--- a/src/Platform/Microsoft.Testing.Platform/Hosts/TestHostBuilder.cs
+++ b/src/Platform/Microsoft.Testing.Platform/Hosts/TestHostBuilder.cs
@@ -619,7 +619,7 @@ internal sealed class TestHostBuilder(IFileSystem fileSystem, IRuntimeFeature ru
         string? seconds = configuration[PlatformConfigurationConstants.PlatformTestHostControllersManagerNamedPipeClientConnectTimeoutSeconds];
 
         // Default timeout is 30 seconds
-        int timeoutSeconds = seconds is null ? TimeoutHelper.DefaultHangTimeoutSeconds : int.Parse(seconds, CultureInfo.InvariantCulture);
+        double timeoutSeconds = seconds is null ? TimeoutHelper.DefaultHangTimeoutSeconds : double.Parse(seconds, CultureInfo.InvariantCulture);
         await logger.LogDebugAsync($"Setting PlatformTestHostControllersManagerNamedPipeClientConnectTimeoutSeconds '{timeoutSeconds}'").ConfigureAwait(false);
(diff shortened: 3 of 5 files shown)


src/Platform/Microsoft.Testing.Platform/Hosts/TestHostBuilder.cs
```diff
@@ -619,7 +619,7 @@ await LogTestHostCreatedAsync(
         string? seconds = configuration[PlatformConfigurationConstants.PlatformTestHostControllersManagerNamedPipeClientConnectTimeoutSeconds];
 
         // Default timeout is 30 seconds
```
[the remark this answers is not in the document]
Youssef1313:
```suggestion
```

src/Platform/Microsoft.Testing.Platform/Helpers/TimeSpanParser.cs
```diff
@@ -24,7 +24,7 @@ public static bool TryParse(string? time, out TimeSpan result)
         if (RoslynString.IsNullOrWhiteSpace(time))
         {
             result = TimeSpan.Zero;
-            return true;
+            return false;
```
Evangelink, line 27: The `TryParse` bug fix is correct — returning `true` for null/whitespace violates the .NET `TryParse` convention and previously caused `TimeSpan.Zero` to be treated as "successfully parsed."

However, this is a behavior change that affects other callers (e.g., `HangDumpCommandLineProvider.ValidateOptionArgumentsAsync`). Consider adding unit tests covering the edge cases: `TryParse(null)`, `TryParse("")`, `TryParse("   ")` should all return `false`, and `TryParse("90s")`, `TryParse("5m")` should return `true` with correct values. This would prevent regressions given the method is used in CLI validation paths.

Also worth noting: previously, setting `TESTINGPLATFORM_DEFAULT_HANG_TIMEOUT` to an empty string would result in `TryParse("")` → `true` with `TimeSpan.Zero`, effectively creating a **zero timeout** that would fail every wait operation. The new behavior (falling back to the default) is strictly better.

src/Platform/Microsoft.Testing.Platform/Helpers/TimeoutHelper.cs
```diff
@@ -14,28 +14,15 @@ static TimeoutHelper()
         string? customAntiHangTimeout = Environment.GetEnvironmentVariable(EnvironmentVariableConstants.TESTINGPLATFORM_DEFAULT_HANG_TIMEOUT);
 #pragma warning restore RS0030 // Do not use banned APIs
 
-        // Took from Tpv2 experience, it's the timeout for the testhost.exe start.
-        DefaultHangTimeSpanTimeout = TimeSpan.FromSeconds(90);
+        DefaultHangTimeSpanTimeout =
+            TimeSpanParser.TryParse(customAntiHangTimeout, out TimeSpan customHangTimeout)
+            ? customHangTimeout
+            : TimeSpan.FromMinutes(5);
```
Evangelink, line 20: `DefaultHangTimeSpanTimeout` is the **global** anti-hang timeout used broadly across the platform — not just for named-pipe connections. Callers include:

| Caller | Usage |
|---|---|
| `FileLogger.InternalSyncLog` | `_semaphore.Wait(DefaultHangTimeSpanTimeout)` |
| `FileLogger.InternalAsyncLogAsync` | `_semaphore.WaitAsync(DefaultHangTimeSpanTimeout)` |
| `FileLogger.Dispose` | `_logLoop.Wait(DefaultHangTimeSpanTimeout)` |
| `FileLogger.DisposeAsync` | `_logLoop.TimeoutAfterAsync(DefaultHangTimeSpanTimeout)` |
| `HangDumpProcessLifetimeHandler.Dispose` | `_activityIndicatorTask.Wait(DefaultHangTimeSpanTimeout)` |
| `TestHostBuilder` | Named pipe client connect timeout |
| `TestHostControllersTestHost` | Named pipe server wait timeout |

For the `FileLogger` semaphore waits and log flush, 5 minutes is excessive. If a semaphore or log flush is hung, waiting 5 minutes before detecting it degrades the user experience significantly compared to the original 90 seconds.

Would it make sense to keep a shorter default for the general anti-hang timeout (or at most ~2 minutes) and introduce a dedicated, longer timeout specifically for named-pipe connection scenarios that need it (e.g., Code Coverage)?

test/UnitTests/Microsoft.Testing.Platform.UnitTests/Messages/AsynchronousMessageBusTests.cs
```diff
@@ -30,7 +30,7 @@ public async Task UnexpectedTypePublished_ShouldFail()
 
         // Fire consume with a good message
         await proxy.PublishAsync(new DummyProducer("DummyProducer", typeof(InvalidTypePublished.ValidDataToProduce)), new InvalidTypePublished.ValidDataToProduce());
-        consumer.Published.WaitOne(TimeoutHelper.DefaultHangTimeoutMilliseconds);
+        consumer.Published.WaitOne();
```
Evangelink, line 33: `WaitOne()` without a timeout blocks indefinitely. If the event is never signaled (e.g., regression in message bus publishing, race condition), this test will **hang forever** in CI instead of failing with a clear message — burning pipeline time until the global job timeout kills it.

Consider replacing with a bounded wait that asserts success:

```suggestion
        Assert.IsTrue(
            consumer.Published.WaitOne(TimeoutHelper.DefaultHangTimeSpanTimeout),
            "Expected Published event to be signaled within the timeout");
```

This uses the `TimeSpan` overload (compatible with the removal of the `int` milliseconds property) and surfaces failures immediately with a meaningful message.

src/Platform/Microsoft.Testing.Platform/Hosts/TestHostBuilder.cs
```diff
@@ -618,8 +618,7 @@ await LogTestHostCreatedAsync(
         await logger.LogDebugAsync($"Connecting to named pipe '{pipeName}'").ConfigureAwait(false);
         string? seconds = configuration[PlatformConfigurationConstants.PlatformTestHostControllersManagerNamedPipeClientConnectTimeoutSeconds];
 
-        // Default timeout is 30 seconds
-        int timeoutSeconds = seconds is null ? TimeoutHelper.DefaultHangTimeoutSeconds : int.Parse(seconds, CultureInfo.InvariantCulture);
+        double timeoutSeconds = seconds is null ? TimeoutHelper.DefaultHangTimeoutSeconds : double.Parse(seconds, CultureInfo.InvariantCulture);
```
Evangelink, line 621: Pre-existing issue, but: `double.Parse` (previously `int.Parse`) will throw `FormatException` on invalid configuration input. Since this value comes from a configuration source (`PlatformConfigurationConstants.PlatformTestHostControllersManagerNamedPipeClientConnectTimeoutSeconds`), a `double.TryParse` with fallback to the default would be more resilient. Same applies to the identical pattern in `TestHostControllersTestHost.cs`.

Not blocking, just noting the opportunity.

Evangelink (changes requested):
Nice cleanup overall — the `TimeSpanParser.TryParse` bug fix and `TimeoutHelper` simplification are solid improvements.

**Key concerns:**

1. **`WaitOne()` without timeout** in `AsynchronousMessageBusTests` — this will hang indefinitely on failure instead of producing a clear test error. Should use a bounded wait with assertion.

2. **5-minute default blast radius** — `DefaultHangTimeSpanTimeout` is used broadly (FileLogger semaphores, log flushes, HangDump dispose, not just pipe connections). 5 minutes is generous for scenarios like semaphore acquisition and log flushing. Consider a dedicated timeout for the pipe connection scenario or a lower general default.

3. **Missing test coverage** for the `TryParse` behavior change (null/empty/whitespace now returns `false` instead of `true`).

The code structure changes (`int` → `double`, property simplification, conditional cleanup) are all correct.

src/Platform/Microsoft.Testing.Platform/Helpers/TimeSpanParser.cs
Youssef1313: Other callers mostly don't come across the null code path. It's a fine change for the case of empty or whitespace as it's invalid anyways.

src/Platform/Microsoft.Testing.Platform/Helpers/TimeoutHelper.cs
Youssef1313: All these are fine to wait for longer. The purpose is hang detection. If nothing is hanging, we are good.

I even would prefer to remove the timeout from **all** callsites. But as said in PR description, I did the more minimal change instead of going too far with removing completely.

test/UnitTests/Microsoft.Testing.Platform.UnitTests/Messages/AsynchronousMessageBusTests.cs
Youssef1313: It's fine. We already run with hangdump. If this test hanged we will get a dump and understand why.

src/Platform/Microsoft.Testing.Platform/Hosts/TestHostBuilder.cs
Youssef1313: I don't think it's a good idea to use TryParse instead of Parse. If user set an invalid string in configuration, it's better that we throw an exception.

Evangelink (approved):

PR #7734, opened 2026-04-16
Delete dead code

Youssef1313:
We should have never introduced this as a public non-experimental API. It was introduced in 2.1.0 and wasn't used and also was never implemented fully with dotnet test. While it's binary breaking change, our users of such API will be test framework authors which is a very limited audience and no one is using it today. So it's a safe breaking change.

### DIFF
--- a/src/Platform/Microsoft.Testing.Platform/CommandLine/CommandLineOptionsValidator.cs
+++ b/src/Platform/Microsoft.Testing.Platform/CommandLine/CommandLineOptionsValidator.cs
@@ -3,9 +3,7 @@
 
 using Microsoft.Testing.Platform.Extensions;
 using Microsoft.Testing.Platform.Extensions.CommandLine;
-using Microsoft.Testing.Platform.Extensions.OutputDevice;
 using Microsoft.Testing.Platform.Helpers;
-using Microsoft.Testing.Platform.OutputDevice;
 using Microsoft.Testing.Platform.Resources;
 
 namespace Microsoft.Testing.Platform.CommandLine;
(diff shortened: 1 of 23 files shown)


src/Platform/Microsoft.Testing.Platform/CommandLine/CommandLineOptionsValidator.cs
```diff
@@ -70,30 +68,6 @@ public static async Task<ValidationResult> ValidateAsync(
         return await ValidateConfigurationAsync(extensionOptionsByProvider.Keys, systemOptionsByProvider.Keys, commandLineOptions).ConfigureAwait(false);
```
Evangelink, line 68: Nit: `CommandLineOptionObsoleteWarning` resource string in `PlatformResources.resx` (and its 13 `.xlf` translation files) is now orphaned — its only consumer was `CheckForObsoleteOptionsAsync` which this PR deletes. Worth cleaning up in this PR to avoid leaving dead resources behind.

src/Platform/Microsoft.Testing.Platform/ServerMode/DotnetTest/IPC/ObjectFieldIds.cs
```diff
@@ -36,7 +36,7 @@ internal static class CommandLineOptionMessageFieldsId
     public const ushort Description = 2;
     public const ushort IsHidden = 3;
     public const ushort IsBuiltIn = 4;
-    public const ushort ObsolescenceMessage = 5;
+    // public const ushort ObsolescenceMessage = 5;
```
Evangelink: Style nit: commented-out code in a protocol field-ID file is a bit noisy. Consider either deleting the line entirely (the feature was never used in practice, so no deployed traffic uses field ID 5) or keeping a named constant to formally reserve the ID:

```csharp
// Reserved: field ID 5 was ObsolescenceMessage, removed as unused.
```

Either way is fine — just not the half-alive `// public const ...` form.

Evangelink (approved):
Clean removal of unused API surface. The deletion is thorough across all layers (public API, help display, validator, IPC serialization) and IPC backward compat is safe thanks to the field-skip `default` branch in the deserializer.

Two minor nits left as inline comments:
1. Orphaned `CommandLineOptionObsoleteWarning` resource string + 13 xlf files should be cleaned up.
2. Commented-out field ID constant — prefer either deleting or a proper reservation comment.

Neither is blocking. LGTM.

PR #7735, opened 2026-04-16
Cleanup logic around discover internals

Youssef1313:

### DIFF
--- a/src/Adapter/MSTestAdapter.PlatformServices/Execution/TestAssemblyInfo.cs
+++ b/src/Adapter/MSTestAdapter.PlatformServices/Execution/TestAssemblyInfo.cs
@@ -24,7 +24,12 @@ internal sealed class TestAssemblyInfo
     /// </summary>
     /// <param name="assembly">Sets the <see cref="Assembly"/> this class is representing. </param>
     internal TestAssemblyInfo(Assembly assembly)
-        => Assembly = assembly;
+    {
+        Assembly = assembly;
+        DiscoversInternals = assembly.GetCustomAttribute<DiscoverInternalsAttribute>() is not null;
+    }
+
+    internal bool DiscoversInternals { get; }
 
     internal List<(MethodInfo Method, TimeoutInfo? TimeoutInfo)> GlobalTestInitializations { get; } = [];
 
--- a/src/Adapter/MSTestAdapter.PlatformServices/Execution/TypeCache.cs
+++ b/src/Adapter/MSTestAdapter.PlatformServices/Execution/TypeCache.cs
@@ -32,8 +32,6 @@ internal sealed class TypeCache : MarshalByRefObject
     /// </summary>
     private readonly ConcurrentDictionary<string, TestClassInfo?> _classInfoCache = new();
 
-    private readonly ConcurrentDictionary<string, bool> _discoverInternalsCache = new();
-
     /// <summary>
     /// Initializes a new instance of the <see cref="TypeCache"/> class.
     /// </summary>
@@ -135,9 +133,9 @@ internal sealed class TypeCache : MarshalByRefObject
 
         string typeName = testMethod.FullClassName;
 
+#if NETCOREAPP
         // Using GetOrAdd to ensure we calculate only once when this is called by different threads in parallel.
         // Using a static lambda to ensure we don't capture.
-#if NETCOREAPP
         return _classInfoCache.GetOrAdd(typeName, CreateTestClassInfo, (this, testMethod));
 #else
         // On .NET Framework, we don't have the GetOrAdd overload that prevents capturing lambdas.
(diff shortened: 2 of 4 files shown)


src/Adapter/MSTestAdapter.PlatformServices/Execution/TestAssemblyInfo.cs
```diff
@@ -24,7 +24,12 @@ internal sealed class TestAssemblyInfo
     /// </summary>
     /// <param name="assembly">Sets the <see cref="Assembly"/> this class is representing. </param>
     internal TestAssemblyInfo(Assembly assembly)
-        => Assembly = assembly;
+    {
+        Assembly = assembly;
+        DiscoversInternals = assembly.GetCustomAttribute<DiscoverInternalsAttribute>() is not null;
```
Evangelink, line 29: Consider using `ReflectHelper.HasDiscoverInternalsAttribute(assembly)` here instead of `assembly.GetCustomAttribute<DiscoverInternalsAttribute>()`.

On .NET Framework, `ReflectionOperations.GetCustomAttributes` (used by `HasDiscoverInternalsAttribute`) delegates to `ReflectionOperationsNetFrameworkAttributeHelpers` which handles **reflection-only load** scenarios, while the direct `Assembly.GetCustomAttribute<T>()` call does not.

This is pre-existing (the old `TypeCache` code also used the direct call), but since you're moving this into `TestAssemblyInfo` as the single source of truth, it's a good opportunity to consolidate on the same abstraction used by the discovery path:

```csharp
internal TestAssemblyInfo(Assembly assembly)
{
    Assembly = assembly;
    DiscoversInternals = ReflectHelper.HasDiscoverInternalsAttribute(assembly);
}
```

Temperature: 🟢 low — this is a suggestion, not a blocker.

src/Adapter/MSTestAdapter.PlatformServices/Execution/TypeCache.cs
```diff
@@ -655,7 +653,7 @@ private void UpdateInfoIfTestInitializeOrCleanupMethod(
     /// <returns>
     /// The TestMethodInfo for the given test method. Null if the test method could not be found.
     /// </returns>
-    private TestMethodInfo ResolveTestMethodInfo(TestMethod testMethod, TestClassInfo testClassInfo)
+    private static TestMethodInfo ResolveTestMethodInfo(TestMethod testMethod, TestClassInfo testClassInfo)
```
Evangelink, line 656: Nice cleanup — these methods are correctly promoted to `static` now that `_discoverInternalsCache` is gone. No instance state is accessed.

src/Adapter/MSTestAdapter.PlatformServices/Helpers/ReflectHelper.cs
```diff
@@ -161,13 +161,11 @@ internal string[] GetTestCategories(MemberInfo categoryAttributeProvider, Type o
...
-    /// Gets discover internals assembly level attribute.
+    /// Returns whether the assembly has discover internals attribute.
     /// </summary>
     /// <param name="assembly"> The test assembly. </param>
-    internal static DiscoverInternalsAttribute? GetDiscoverInternalsAttribute(Assembly assembly)
-        => PlatformServiceProvider.Instance.ReflectionOperations.GetCustomAttributes(assembly, typeof(DiscoverInternalsAttribute))
-            .OfType<DiscoverInternalsAttribute>()
-            .FirstOrDefault();
+    internal static bool HasDiscoverInternalsAttribute(Assembly assembly)
+        => PlatformServiceProvider.Instance.ReflectionOperations.GetCustomAttributes(assembly, typeof(DiscoverInternalsAttribute)).Length > 0;
```
Evangelink, line 168: 👍 This is more efficient than the old `.OfType<DiscoverInternalsAttribute>().FirstOrDefault()` — avoids the LINQ iterator allocation and type checks.

Evangelink (commented):
Nice cleanup — this correctly resolves the two TODOs and simplifies the caching story.

**What this PR does well:**
- Eliminates the redundant `_discoverInternalsCache` (keyed by assembly name string) in favor of a property on `TestAssemblyInfo` (already cached by `Assembly` instance in `_testAssemblyInfoCache`).
- Property access is cheaper than `ConcurrentDictionary.TryGetValue` per test method resolution.
- `HasDiscoverInternalsAttribute` using `.Length > 0` is more efficient than the old LINQ `.OfType<T>().FirstOrDefault()`.
- The `static` method promotions are correct — no instance state is accessed after the cache removal.
- Thread safety is preserved: `DiscoversInternals` is set once in the constructor and is read-only.
- No public API surface change (property is `internal`).

**One suggestion:** The `TestAssemblyInfo` constructor uses `Assembly.GetCustomAttribute<T>()` directly while `HasDiscoverInternalsAttribute` goes through `PlatformServiceProvider.Instance.ReflectionOperations` (which handles reflection-only load on .NET Framework). Consider using `ReflectHelper.HasDiscoverInternalsAttribute(assembly)` in the constructor for consistency. See inline comment.

Evangelink (approved):

src/Adapter/MSTestAdapter.PlatformServices/Execution/TestAssemblyInfo.cs
Youssef1313: I kept the existing behavior that directly did `assembly.GetCustomAttribute<...>`.

PR #7737, opened 2026-04-16
Cleanup dead IPC code

Youssef1313:

### DIFF
--- a/src/Platform/Microsoft.Testing.Platform/IPC/IClient.cs
+++ b/src/Platform/Microsoft.Testing.Platform/IPC/IClient.cs
@@ -6,11 +6,7 @@ using Microsoft.CodeAnalysis;
 namespace Microsoft.Testing.Platform.IPC;
 
 [Embedded]
-internal interface IClient :
-#if NETCOREAPP
-    IAsyncDisposable,
-#endif
-    IDisposable
+internal interface IClient : IDisposable
 {
     bool IsConnected { get; }
 
--- a/src/Platform/Microsoft.Testing.Platform/IPC/NamedPipeClient.cs
+++ b/src/Platform/Microsoft.Testing.Platform/IPC/NamedPipeClient.cs
@@ -251,25 +251,4 @@ internal sealed class NamedPipeClient : NamedPipeBase, IClient
         _namedPipeClientStream.Dispose();
         _disposed = true;
     }
-
-#if NETCOREAPP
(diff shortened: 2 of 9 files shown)


src/Platform/Microsoft.Testing.Platform/IPC/IClient.cs
```diff
@@ -6,11 +6,7 @@
 namespace Microsoft.Testing.Platform.IPC;
 
 [Embedded]
-internal interface IClient :
-#if NETCOREAPP
-    IAsyncDisposable,
-#endif
-    IDisposable
+internal interface IClient : IDisposable
 {
```
Evangelink, line 10: nit: `IClient` lost `IAsyncDisposable` and `NamedPipeClient.DisposeAsync()` was removed, but `IServer` still conditionally extends `IAsyncDisposable` and `NamedPipeServer.DisposeAsync()` is kept. If async dispose on the IPC types is dead code, it'd be nice to clean up the server side too for consistency. If `NamedPipeServer.DisposeAsync()` is intentionally kept because it has meaningful async teardown logic (awaiting the loop task), might be worth calling that out in the PR description.

src/Platform/Microsoft.Testing.Platform/IPC/NamedPipeClient.cs
```diff
@@ -251,25 +251,4 @@ public void Dispose()
         _namedPipeClientStream.Dispose();
         _disposed = true;
     }
```
Evangelink, line 253: nit: The removed code had explicit comments warning about `MissingMethodException` in mixed-version scenarios (old extension + new MTP core). Could you add a brief note to the PR description on why it's now safe to remove these compat shims? e.g., all extensions ship in lockstep, or enough versions have shipped since the `[Obsolete]` markers were added.

Evangelink (approved):
Cleanup looks correct — all in-repo call sites are updated consistently and the diff is minimal. Two minor notes inline, neither blocking.

src/Platform/Microsoft.Testing.Platform/IPC/IClient.cs
Youssef1313: The server still needs DisposeAsync

src/Platform/Microsoft.Testing.Platform/IPC/NamedPipeClient.cs
Youssef1313: Enough version shipped + we already dropped Polyfill which already breaks compatibility with older versions.

PR #7745, opened 2026-04-17
Avoid BlockingCollection usage in MTP

Youssef1313:
This better ensures we don't end up in long blocks while running in the thread pool, and is more performant.

Our custom channel implementation has been used by MTP MessageBus for a while already, which is a very core component in MTP, and no issues were noticed.

### DIFF
--- a/src/Platform/Microsoft.Testing.Extensions.Telemetry/AppInsightsProvider.cs
+++ b/src/Platform/Microsoft.Testing.Extensions.Telemetry/AppInsightsProvider.cs
@@ -147,6 +147,11 @@ internal sealed partial class AppInsightsProvider :
             {
                 while (_payloads.TryRead(out (string EventName, IDictionary<string, object> ParamsMap) payload))
                 {
+                    if (_flushTimeoutOrStop.Token.IsCancellationRequested)
+                    {
+                        return;
+                    }
+
                     (string eventName, IDictionary<string, object> paramsMap) = payload;
 #endif
 
@@ -280,6 +285,11 @@ internal sealed partial class AppInsightsProvider :
 #if NETCOREAPP
         await _payloads.Writer.WriteAsync((eventName, paramsMap), cancellationToken).ConfigureAwait(false);
 #else
+        if (cancellationToken.IsCancellationRequested)
+        {
+            return Task.FromCanceled(cancellationToken);
+        }
+
         _payloads.Write((eventName, paramsMap));
         return Task.CompletedTask;
 #endif
--- a/src/Platform/Microsoft.Testing.Extensions.Telemetry/AppInsightsProvider.cs
+++ b/src/Platform/Microsoft.Testing.Extensions.Telemetry/AppInsightsProvider.cs
@@ -9,6 +9,9 @@ using Microsoft.Testing.Platform;
 using Microsoft.Testing.Platform.Configurations;
 using Microsoft.Testing.Platform.Helpers;
 using Microsoft.Testing.Platform.Logging;
+#if !NETCOREAPP
+using Microsoft.Testing.Platform.Messages;
+#endif
 using Microsoft.Testing.Platform.Services;
 using Microsoft.Testing.Platform.Telemetry;
 
@@ -45,7 +48,7 @@ internal sealed partial class AppInsightsProvider :
 #if NETCOREAPP
     private readonly Channel<(string EventName, IDictionary<string, object> ParamsMap)> _payloads;
 #else
-    private readonly BlockingCollection<(string EventName, IDictionary<string, object> ParamsMap)> _payloads;
+    private readonly SingleConsumerUnboundedChannel<(string EventName, IDictionary<string, object> ParamsMap)> _payloads;
 #endif
 #if DEBUG
     // Telemetry properties that are allowed to contain unhashed information.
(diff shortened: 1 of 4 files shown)


src/Platform/Microsoft.Testing.Extensions.Telemetry/AppInsightsProvider.cs
```diff
@@ -274,7 +280,7 @@ Task LogEventAsync(string eventName, IDictionary<string, object> paramsMap, Canc
 #if NETCOREAPP
         await _payloads.Writer.WriteAsync((eventName, paramsMap), cancellationToken).ConfigureAwait(false);
 #else
-        _payloads.Add((eventName, paramsMap), cancellationToken);
+        _payloads.Write((eventName, paramsMap));
```
Evangelink, line 293: The old code passed `cancellationToken` to `BlockingCollection.Add(item, cancellationToken)`. The new code calls `_payloads.Write(...)` ignoring the token entirely, while the NETCOREAPP path still observes it via `WriteAsync(..., cancellationToken)`.

For an unbounded channel, `Write` never blocks so there's nothing to *cancel*, but callers may still expect `OperationCanceledException` if the token is already signaled. Consider adding a guard:

```suggestion
#else
        cancellationToken.ThrowIfCancellationRequested();
        _payloads.Write((eventName, paramsMap));
```

src/Platform/Microsoft.Testing.Extensions.Telemetry/AppInsightsProvider.cs
```diff
@@ -101,14 +104,11 @@ public AppInsightsProvider(
...
-        _telemetryTask = task.Run(IngestLoopAsync, _testApplicationCancellationTokenSource.CancellationToken);
 #else
-        // Keep the custom thread to avoid to waste one from thread pool.
-        // We have some await but we should stay on the custom thread if not for special cases like trace log or exception.
-        _payloads = [];
-        _telemetryTask = _task.RunLongRunning(IngestLoopAsync, "Telemetry AppInsightsProvider", _testApplicationCancellationTokenSource.CancellationToken);
+        _payloads = new SingleConsumerUnboundedChannel<(string EventName, IDictionary<string, object> ParamsMap)>();
 #endif
 
+        _telemetryTask = task.Run(IngestLoopAsync, _testApplicationCancellationTokenSource.CancellationToken);
```
Evangelink, line 111: The old code intentionally used `_task.RunLongRunning(IngestLoopAsync, "Telemetry AppInsightsProvider", ...)` which spawns a dedicated named thread — visible as "Telemetry AppInsightsProvider" in crash dumps and debugger. Switching to `task.Run(...)` loses that thread name for diagnostics.

Since `SingleConsumerUnboundedChannel.WaitToReadAsync` is truly async (TCS-based), moving to the ThreadPool is functionally fine — no thread will be blocked. But the loss of the named thread is a minor diagnostics regression.

Was this intentional? If so, a brief comment explaining the reasoning would be helpful.

src/Platform/Microsoft.Testing.Extensions.Telemetry/AppInsightsProvider.cs
```diff
@@ -134,90 +134,96 @@ private async Task IngestLoopAsync()
...
         _testApplicationCancellationTokenSource.CancellationToken.Register(_flushTimeoutOrStop.Cancel);
+
         try
         {
 #if NETCOREAPP
             while (await _payloads.Reader.WaitToReadAsync(_flushTimeoutOrStop.Token).ConfigureAwait(false))
             {
-                (string eventName, IDictionary<string, object> paramsMap) = await _payloads.Reader.ReadAsync().ConfigureAwait(false);
+                {
+                    (string eventName, IDictionary<string, object> paramsMap) = await _payloads.Reader.ReadAsync().ConfigureAwait(false);
```
Evangelink, line 144: Nit: the extra `{` here creates a redundant scope block to align the shared body (post-`#endif`) at the same indentation as the non-NETCOREAPP inner `while` loop. It compiles fine, but makes the preprocessor structure harder to follow at a glance. Not a blocker.

Evangelink (approved):
Clean, well-motivated improvement — replacing `BlockingCollection` with the proven `SingleConsumerUnboundedChannel` avoids thread-pool blocking on non-NETCOREAPP and aligns both TFMs around the same async pattern.

**What I like:**
- The channel implementation is already battle-tested in `AsyncConsumerDataProcessor` (MessageBus).
- `FileLogger` gets a nice simplification: the `_channel` field name is now shared across TFMs, eliminating several `#if`/`#else` blocks (`EnsureAsyncLogObjectsAreNotNull`, `MemberNotNull`, null guards).
- `[Embedded]` attribute addition is the correct pattern for source-linked files.

**Minor items (see inline):**
1. `LogEventAsync` (non-NETCOREAPP) now silently drops the `cancellationToken` — consider adding `cancellationToken.ThrowIfCancellationRequested()` before `Write`.
2. The switch from `RunLongRunning` to `task.Run` loses the named thread for diagnostics — was this intentional?
3. Redundant nested braces in the NETCOREAPP ingest loop path (cosmetic).

None of these are blockers. LGTM overall.

src/Platform/Microsoft.Testing.Extensions.Telemetry/AppInsightsProvider.cs
Youssef1313: This is needed to align the `#if` and the `#else` paths.

src/Platform/Microsoft.Testing.Extensions.Telemetry/AppInsightsProvider.cs
Youssef1313: This is following the existing .NET Core path which is correct.

src/Platform/Microsoft.Testing.Extensions.Telemetry/AppInsightsProvider.cs
Youssef1313: Fixed (by returning a canceled task instead of directly throwing though)

src/Platform/Microsoft.Testing.Extensions.Telemetry/AppInsightsProvider.cs
```diff
@@ -134,90 +134,101 @@ private async Task IngestLoopAsync()
...
-                        await _logger.LogErrorAsync("Error during telemetry report.", ex).ConfigureAwait(false);
-                        lastLoggedError = _clock.UtcNow;
+                        // If we have a lot of issues with the network we could have a lot of logs here.
+                        // We log one error every 3 seconds.
+                        // We could do better back-pressure.
+                        if (_logger.IsEnabled(LogLevel.Error) && (!lastLoggedError.HasValue || (lastLoggedError.Value - _clock.UtcNow).TotalSeconds > 3))
+                        {
+                            await _logger.LogErrorAsync("Error during telemetry report.", ex).ConfigureAwait(false);
+                            lastLoggedError = _clock.UtcNow;
+                        }
```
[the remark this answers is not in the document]
Youssef1313: Pre-existing bug. This needs to be fixed in a follow-up PR.

Evangelink (approved):

Issue #12, opened 2016-09-18
ExpectedException Support


AbhitejJohn:
## Description

Add back support for Expected Exception which was made obsolete. This would helps in two ways:
1. Transition from V1 would be made much simpler.
2. The Create IntelliTest Wizard which currently only generates method level attributes would continue to work until fixed later on.
## Steps to reproduce
1. Create a Unit Test project.
2. Add the V2 test fx and adapter to it.
3. Add a test method which throws an exception and add an ExpectedException attribute targeting the same exception.
## Expected behavior

The test should pass.
## Actual behavior

The test fails and the build throws a warning that this is an obsolete attribute.
## Environment

> Please share additional details about the test environment.
> Operating system, Build version of vstest.console, Package version of MSTest
> framework and adapter

Issue #11, opened 2016-09-16
Error installing MSTest.TestFramework on .net core


codito:
## Description

MSTest.TestFramework should support netcoreapp1.0 target.
## Steps to reproduce
- VS from WPT, 25715.03
- Create a new console application
- Install MSTest.TestFramework
## Expected behavior

Test framework installs correctly.
## Actual behavior

Test framework install fails with following error:

```
Package MSTest.TestFramework 1.0.1-preview is not compatible with netcoreapp1.0 (.NETCoreApp,Version=v1.0). Package MSTest.TestFramework 1.0.1-preview supports:
  - dotnet (.NETPlatform,Version=v5.0)
  - net451 (.NETFramework,Version=v4.5.1)
One or more packages are incompatible with .NETCoreApp,Version=v1.0.
```
## Environment

MSTest.TestFramework 1.0.1-preview.

Issue #10, opened 2016-09-14
Packaging mstest for .net core


codito:
## Description

In .net core, the dependencies are resolved from deps.json. The adapter must be referenced from a test project such that the `MSTest.TestAdapter.dll` is located by the test platform.

Currently, the nuget package ships with `props` files which copy the test adapter to output directories. It is not supported in .net core (see #6). What is the way forward?
- Ship different nuget package for each runtime
- Ship single nuget package, let the adapter be resolved from the corresponding runtime directory e.g. `netcoreapp1.0` for .net core.

Thoughts?

PR #7750, opened 2026-04-17
RFC 012 - Structured Assertion Messages

Evangelink:
## Summary

Add RFC 012 proposing a standardized, structured multi-line format for assertion failure messages across all MSTest `Assert.*` methods. `CollectionAssert` and `StringAssert` are considered legacy and are out of scope — their messages will be updated opportunistically but are not specified by this RFC.

## Motivation

The current single-line concatenated format is hard to scan in CI output, buries user messages, has no visual hierarchy, and breaks readability with long values. This RFC proposes a consistent multi-line structure that puts the most important information first.

## What's in the RFC

- **Proposed structure**: universal `Assertion failed.` prefix, summary line, user message, labeled evidence block, call-site expression, stack trace
- **Complete assertion message catalog**: every MSTest assertion mapped to its structured failure message, including sub-cases (null values, wrong types, etc.)
- **Diff diagnostics**: rules for string and collection comparison summaries
- **Value rendering rules**: conventions for null, strings, collections, types, etc.
- **Value and collection truncation**: configurable limits with context preservation
- **Call-site expression handling**: multiline argument collapsing, unavailable expressions
- **Open questions**: user message placement, truncation limits, custom comparer display, and more

## Requesting review

Looking for team feedback on the overall structure and the open questions listed at the end of the RFC.


docs/RFCs/011-Structured-Assertion-Messages.md
```diff
@@ -0,0 +1,1273 @@
...
+| --------- | ------ |
+| `AreEqual` (int) | `Assertion failed. Expected values to be equal.` |
+| `AreEqual` (string) | `Assertion failed. Expected strings to be equal (case-sensitive).` |
+| `IsTrue` | `Assertion failed. Expected condition to be true.` |
+| `IsNull` | `Assertion failed. Expected value to be null.` |
+| `IsInstanceOfType` | `Assertion failed. Expected value to be an instance of String.` |
+| `ThrowsExactly` | `Assertion failed. Expected exception of type ArgumentException but no exception was thrown.` |
+| `Contains` (collection) | `Assertion failed. Expected collection to contain the specified element.` |
+
+The summary follows the verbal form **"Expected [subject] to [verb phrase]."** for consistency. Short values like type names or small counts can be inlined when they improve clarity without bloating the line.
```
nohwnd, line 82: As an additional example that might clarify it for others like me, 

Assert <actual description>, to <expected summary>, [optional] but <additional information about actual>.

E.g. Assert.Contains('f', ['a', 'b', 'c']), - Expected collection to contain the specified element. (we want f to be present in the actual collection)


This already breaks in subtle way with the AreEqual, and I think it is fine.

docs/RFCs/011-Structured-Assertion-Messages.md
```diff
@@ -0,0 +1,1273 @@
...
+The blank line creates a clear visual boundary between the narrative (summary + user message) and the evidence (values + details). Values are flush-left — no indentation — so they can be copy-pasted directly (e.g. in a TDD workflow where you update an expected value from the failure output).
+
+Rules:
+
+- `expected:` and `actual:` labels are left-aligned.
+- For short values, the value follows the label on the same line after a single space: `expected: 42`.
+- For visual alignment of same-line values, shorter labels are padded with trailing spaces to match the longest label in the block (e.g. `actual:` is padded to align with `expected:`).
+- When an assertion does not have a natural `expected` / `actual` pair (e.g. `Assert.Fail`, `Assert.Inconclusive`), these lines are omitted entirely.
+
+Multi-line values (e.g. JSON objects, large collections) are displayed starting on the next line, flush-left with no indentation:
```
nohwnd, line 116: agree, at least 1 indentation would be nice, tu requires to reflow the text, which breaks it when resizing the window locally, or not knowing the window width, so probably nothing we can do.

docs/RFCs/011-Structured-Assertion-Messages.md
```diff
@@ -0,0 +1,1273 @@
...
+
+This keeps copy-paste zero-friction — you can select the value lines and paste them directly into source code without stripping leading whitespace.
+
+Note: When a multi-line value is a string, it is still quoted according to the Value Rendering Rules. The opening quote appears on the line after the label, and the closing quote appears at the end of the value:
+
+```text
+expected:
+"line one\nline two\nline three"
+actual:
+"line one\nLINE TWO\nline three"
```
nohwnd, line 133: for terminal output I replaced the values with their printable values (they are single char width, and almost everything can render them), but kept newlines, it made formatted text way easier to compare visually.

docs/RFCs/011-Structured-Assertion-Messages.md
```diff
@@ -0,0 +1,1273 @@
...
+```text
+expected:   42
+actual:     37
+ignoreCase: true
+culture:    tr-TR
+```
+
+In this example, `expected:` and `actual:` are the core value labels, while `ignoreCase:` and `culture:` are assertion-specific details. All labels share a single alignment column within the block.
+
+Note: Alignment is applied **within each evidence block as a whole**. All labels in the block (value labels like `expected:` / `actual:` and detail labels like `ignoreCase:` / `culture:`) are padded to match the longest label in the block. This means values and details share a single alignment column.
```
nohwnd: how will the assertions pass this one ? will we split on the first `:`?

nohwnd: or pass arrays of labels and arrays of values?

docs/RFCs/011-Structured-Assertion-Messages.md
```diff
@@ -0,0 +1,1273 @@
...
+Assert.AreEqual(""" { "name": "Alice", "age": 30 } """, actualJson)
+```
+
+**Approach 2 — Omit the multiline argument and show `...`:** Detect that the expression contains newlines and replace it with `...`:
+
+```text
+Assert.AreEqual(..., actualJson)
+```
+
+**Recommendation:** Use Approach 2 (omit with `...`) when the captured expression contains newlines. The full value is already displayed in the `expected:` / `actual:` lines, so repeating it in the call-site is redundant. The call-site should identify *where* the call happened, not *what* the values were.
```
nohwnd: Seems reasonable to not repeat the literal value, needs better way to communicate this than `...` imho. e..g Assert.AreEqual(<expected>, actualJson);

docs/RFCs/011-Structured-Assertion-Messages.md
```diff
@@ -0,0 +1,1273 @@
...
+   at MyTests.ValidationTests.InvalidInput_ShouldThrow() in ValidationTests.cs:line 18
+```
+
+### Assert.ThrowsExactly (wrong exception type)
+
+```text
+Assertion failed. Expected exception of exactly type ArgumentException but caught InvalidOperationException.
+
+expected type: System.ArgumentException
+actual type:   System.InvalidOperationException
```
nohwnd, line 301: this should print the actual value as well, not just the type. This shoudl be applied everywhere, test the quality of the value, but also show what you compared.

docs/RFCs/011-Structured-Assertion-Messages.md
```diff
@@ -0,0 +1,1273 @@
...
+```
+
+### Assert.AreEqual (large strings)
+
+```text
+Assertion failed. Expected strings to be equal (case-sensitive).
+Strings differ at 1 location(s). First difference at index 1042.
+
+expected:
+"...configuration that spans many lines and contains the production database
```
nohwnd, line 314: what is the limit here, I assume this is example and limit is 1k+ characters?

nohwnd: here we don't have any indicator of where the strings differs. it would be nice to reserve some character for highlighting the difference, and specify it in this spec. would allow us to show diffs inline as well.

docs/RFCs/011-Structured-Assertion-Messages.md
```diff
@@ -0,0 +1,1273 @@
...
+   at MyTests.ConfigTests.ShouldLoadProductionConfig() in ConfigTests.cs:line 88
+```
+
+### Assert.Contains (collection)
+
+```text
+Assertion failed. Expected collection to contain the specified element.
+
+expected to contain: "banana"
+collection:          ["apple", "cherry", "date"]
```
nohwnd: other assertions don't use spaces in the descriptions, I thought they are based on parameter names.
on the "collection" line I would like to see it mention "actual" to make it obvious what is provided expected value, and what is the received actual value.

docs/RFCs/011-Structured-Assertion-Messages.md
```diff
@@ -0,0 +1,1273 @@
...
+    """
+    {
+      "name": "Alice",
+      "age": 30
+    }
+    """,
+    actualJson);
+```
+
+Output — the multiline expression is replaced with `...` in the call-site:
```
nohwnd: same comment as above, ... is not great imho

docs/RFCs/011-Structured-Assertion-Messages.md
```diff
@@ -0,0 +1,1273 @@
...
+
+expected type: System.String
+actual type:   System.Int32
+actual value:  42
+```
+
+#### Assert.IsInstanceOfType (value is null)
+
+```text
+Assertion failed. Expected value to be an instance of String.
```
nohwnd: nit: the wording here differs from IsExactInstanceOfType. there it says it should be value exactly of type, and here that value should be instance of string.

docs/RFCs/011-Structured-Assertion-Messages.md
```diff
@@ -0,0 +1,1273 @@
...
+collection:  ["apple", "cherry", "date"]
+```
+
+#### Assert.HasCount
+
+```text
+Assertion failed. Expected collection to have 5 element(s) but found 3.
+
+expected count: 5
+actual count:   3
```
nohwnd, line 848: print collection as well, similarly for other assertions below

docs/RFCs/011-Structured-Assertion-Messages.md
```diff
@@ -0,0 +1,1273 @@
...
+
+actual count: 0
+```
+
+### Assert — Comparison (IComparable)
+
+#### Assert.IsInRange
+
+```text
+Assertion failed. Expected value to be in range [5, 10].
```
nohwnd: imho ranges are now represented as 5..10

docs/RFCs/011-Structured-Assertion-Messages.md
```diff
@@ -0,0 +1,1273 @@
...
+
+value:    3
+minValue: 5
+maxValue: 10
```
nohwnd: this looks backwards? expectation should go first. would like to see the "value" mention "actual "  in the name

docs/RFCs/011-Structured-Assertion-Messages.md
```diff
@@ -0,0 +1,1273 @@
...
+```
+
+#### Assert.IsGreaterThan
+
+```text
+Assertion failed. Expected value to be greater than the lower bound.
+
+lowerBound: 10
+actual:     7
```
nohwnd: here spaces are not used again, unlike in some of the items above, seeing it side by side spacing in camelCase is nice.

docs/RFCs/011-Structured-Assertion-Messages.md
```diff
@@ -0,0 +1,1273 @@
...
+Note: Today, `Assert.Inconclusive` uses the same `"{0} failed. {1}"` format as other assertions (producing `Assert.Inconclusive failed. <message>`). This RFC proposes dropping the word "failed" for inconclusive results, since an inconclusive outcome is not a failure — it signals that the test could not be run to completion. This is a **breaking change** to the `AssertInconclusiveException.Message` format and should be listed alongside the other backward-compatibility notes. `Assert.Inconclusive` throws `AssertInconclusiveException` (not `AssertFailedException`) and is intentionally excluded from the universal `Assertion failed.` prefix.
+
+#### Assert.That
+
+```text
+Assertion failed. Expected condition to be true.
+
+condition: order.Total > 0
+values:
+  order.Total = -5
```
nohwnd: other assertion say to not use margin

docs/RFCs/011-Structured-Assertion-Messages.md
```diff
@@ -0,0 +1,1273 @@
...
+
+```text
+Assertion failed. Expected collection to not contain the specified element.
+```
+
+#### CollectionAssert.AllItemsAreNotNull
+
+```text
+Assertion failed. Expected all items in the collection to be non-null.
+Found null element at index 2.
```
nohwnd, line 998: show the collection

docs/RFCs/011-Structured-Assertion-Messages.md
```diff
@@ -0,0 +1,1273 @@
...
+```
+
+#### CollectionAssert.AllItemsAreUnique
+
+```text
+Assertion failed. Expected all items in the collection to be unique.
+Duplicate element found: "apple".
```
nohwnd, line 1007: this description uses first capital letter and spaces, a 3rd variation of the output. Imho spaces + lower case letters is nicer

docs/RFCs/011-Structured-Assertion-Messages.md
```diff
@@ -0,0 +1,1273 @@
...
+Element at index 2 is of type Int32.
+
+expected type: System.String
+actual type:   System.Int32 (at index 2)
+```
+
+#### CollectionAssert.IsSubsetOf
+
+```text
+Assertion failed. Expected collection to be a subset of the specified collection.
```
nohwnd, line 1026: outputs are missing

docs/RFCs/011-Structured-Assertion-Messages.md
```diff
@@ -0,0 +1,1273 @@
...
+```
+
+#### CollectionAssert.AllItemsAreInstancesOfType
+
+```text
+Assertion failed. Expected all items in the collection to be instances of String.
+Element at index 2 is of type Int32.
+
+expected type: System.String
+actual type:   System.Int32 (at index 2)
```
nohwnd, line 1019: show the collection

nohwnd: and ideally a marker for the wrong value, or the value, followed by collection

docs/RFCs/011-Structured-Assertion-Messages.md
```diff
@@ -0,0 +1,1273 @@
...
+Assertion failed. Expected collections to not contain the same elements.
+```
+
+#### CollectionAssert.AreEqual
+
+```text
+Assertion failed. Expected collections to be equal (same elements in same order).
+Collections have 5 element(s). 2 element(s) differ. First difference at index 2.
+
+expected[2]: "cherry"
```
nohwnd, line 1067: 4th variation here, this uses indexes, are we going to re-print them all?

docs/RFCs/011-Structured-Assertion-Messages.md
```diff
@@ -0,0 +1,1273 @@
...
+
+**Arguments against:**
+
+- Breaks the grep anchor — `Assertion failed.` is no longer always on line 1.
+- Makes it harder to visually scan where assertion boundaries are in a multi-failure log.
+- When no user message is provided, line 1 changes identity, which is inconsistent.
+
+### Recommendation
+
+This RFC proposes **Option A** as the default. The user message appears after the summary but before the values, giving it high prominence while preserving the grep-friendly prefix on line 1. However, this is explicitly an open question and feedback is welcome.
```
nohwnd, line 1203: agree.

docs/RFCs/011-Structured-Assertion-Messages.md
```diff
@@ -0,0 +1,1273 @@
...
+
+To ensure consistency across all assertions, values displayed in the evidence block follow these rendering conventions:
+
+| Value | Rendering | Notes |
+| ----- | --------- | ----- |
+| `null` | `(null)` | Parenthesized to distinguish from the literal string `"null"`. |
+| Empty string `""` | `""` | Shown as quoted empty string. |
+| Whitespace-only string | `"   "` | Quoted, so whitespace is visible. |
+| Strings | `"value"` | Always quoted with double quotes to delimit boundaries. |
+| Strings with embedded quotes | `"she said \"hello\""` | Internal double quotes are backslash-escaped. |
```
nohwnd, line 1248: this feels unnecessary, any additional escaping we add to the string potentially misalignes it with the expected string, so you look and think " this is 4 chars off" but in reality is it just 2 quote characters off
```
"a"
"\"b\""
```
very easy to see here, not super easy to see in larger output.

nohwnd: escaping chars with their printable versions worked well for me, especially you need to escape the escape, to avoid rendering ansi. I would not replace \n to keep user text formatting.

docs/RFCs/011-Structured-Assertion-Messages.md
```diff
@@ -0,0 +1,1273 @@
...
+| Strings with control characters | `"line1\nline2\ttab"` | Control characters rendered as C#-style escape sequences (`\n`, `\r`, `\t`, `\0`). Other non-printable characters rendered as `\uXXXX`. |
+| Numeric types | `42`, `3.14`, `-7` | Default `ToString()` formatting. No quotes. |
+| Boolean | `true`, `false` | Lowercase, no quotes. |
+| Types | `System.String` | Fully qualified type name in evidence blocks. Short name (`String`) in summary lines for readability. No quotes. |
+| Collections | `["a", "b", "c"]` | JSON-style array notation. Elements follow the same rendering rules recursively. |
```
nohwnd, line 1253: consider checking longest element size and rendering on newline with 4 margin if any element is 30+ chars? Or total lenght is 200+ chars?

Should render both actual and expected the same

docs/RFCs/011-Structured-Assertion-Messages.md
```diff
@@ -0,0 +1,1273 @@
...
+
+All newlines within the message (between the assertion prefix and the stack trace) use `Environment.NewLine` to ensure consistent rendering across platforms. The test runner is responsible for the stack trace formatting.
+
+## Backward Compatibility
+
+This is a **breaking change** for anyone who parses assertion messages as structured data (e.g. regex-based log parsers). The `AssertFailedException.Message` property will contain the new multi-line format.
+
+Mitigation:
+
+- The change ships in a new major version (MSTest v4).
```
nohwnd: there already is v4, and we did already change wording of some of the assertions without bumping version. we will also evolve them, so while a breaking change, we should probably handle it as output change, not as change requiring new major version every time.

docs/RFCs/011-Structured-Assertion-Messages.md
```diff
@@ -0,0 +1,1273 @@
...
+Note: `Assert.Fail` has no summary line beyond the prefix — the user message (if any) is the entire content. Because there is no evidence block, the blank-line separator between narrative and evidence is omitted. When no user message is provided, the output is simply `Assertion failed.`
+
+#### Assert.Inconclusive
+
+```text
+Assert.Inconclusive. Database server not available for integration tests.
+```
+
+Note: `Assert.Inconclusive` throws `AssertInconclusiveException` (not `AssertFailedException`) and uses a distinct prefix. It is intentionally excluded from the universal `Assertion failed.` prefix because an inconclusive result is not a failure — it signals that the test could not be run to completion.
+
```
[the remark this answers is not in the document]
Evangelink: Done — the `Assert.Inconclusive` format change (`"failed"` → dropped) is now explicitly called out in the Backward Compatibility section as a specific breaking change, alongside the general message format change.

docs/RFCs/011-Structured-Assertion-Messages.md
Evangelink: Done — adopted your `<param>` suggestion. All `...` placeholders in call-site expressions now use named parameter placeholders like `<delta>`, `<expected>`, etc. This makes it clear which argument occupies that position.

docs/RFCs/011-Structured-Assertion-Messages.md
Evangelink: Good question — this is now captured as **unresolved question #11** (Evidence block internal API). The options listed are: (a) arrays of label/value pairs, (b) split on the first `:` per line, or (c) a structured `EvidenceBlock` type. This needs to be decided before implementation since it affects extensibility and third-party assertion authors. I think (a) is the right approach.

docs/RFCs/011-Structured-Assertion-Messages.md
Evangelink: Yes, the example shows truncated output — the actual limit is configurable (default TBD, current proposal is 1024 characters). Added your diff highlighting idea to **unresolved question #3** — it now mentions reserving a character or escape for inline diff highlighting, which would enable showing diffs inline as well. This would be valuable but we need to define the exact mechanism (ANSI escape? marker chars? `^` caret line?).

docs/RFCs/011-Structured-Assertion-Messages.md
Evangelink: Done — two changes applied:
1. All labels standardized to spaced lowercase (e.g. `lower bound:` instead of `lowerBound:`, `ignore case:` instead of `ignoreCase:`)
2. `collection:` renamed to `actual:` throughout all collection assertion evidence blocks to make the expected/actual distinction clear

docs/RFCs/011-Structured-Assertion-Messages.md
Evangelink: Great points — this is now captured as **unresolved question #10** (Control character rendering strategy). The three options being considered:

1. C#-style escape sequences (`\n`, `\t`) — current proposal
2. Unicode printable replacements (`␊`, `␉`) — preserves single-char-width alignment as you mentioned
3. Keep actual newlines (to preserve user text formatting) but escape other non-printable chars with printable equivalents

Your concern about backslash-escaped quotes misaligning the visual diff is well taken — `\"` turns 1 char into 2, which shifts all subsequent positions. Printable replacement characters avoid that problem entirely. This needs more thought before we commit to an approach.

docs/RFCs/011-Structured-Assertion-Messages.md
Evangelink: Good idea — this is now **unresolved question #12** (Collection multi-line rendering). The proposed heuristic: render one-element-per-line when any element is 30+ chars or total rendered length exceeds 200+ chars. Both actual and expected would use the same rendering style for visual comparison, as you suggested.

docs/RFCs/011-Structured-Assertion-Messages.md
Evangelink: The indexed labels (`expected[2]:` / `actual[2]:`) are specific to `CollectionAssert.AreEqual` because the index is essential context — you need to know *where* in the collection the mismatch is. Only the first differing element pair is shown (not the full collections), with the total diff count in the summary line signaling if more fixes are needed. This is intentionally different from unordered comparisons (`AreEquivalent`) which show `missing:` / `unexpected:` lists instead.

docs/RFCs/011-Structured-Assertion-Messages.md
Evangelink: Agreed — updated the Backward Compatibility section. It now says these changes are part of MSTest v4 (which already includes other breaking changes) and that message format changes are treated as output changes within the major version, not requiring a new major version for each evolution.

docs/RFCs/011-Structured-Assertion-Messages.md
Evangelink: Done — wording is now parallel across both:
- `IsInstanceOfType`: "Expected value to be of type String (or derived)."
- `IsExactInstanceOfType`: "Expected value to be exactly of type ArgumentException."

Both follow the same "Expected value to be [exactly] of type X" pattern.

docs/RFCs/011-Structured-Assertion-Messages.md
Evangelink: Done — added `actual value:` to `IsExactInstanceOfType`, `IsNotExactInstanceOfType`, `IsNotInstanceOfType`, and `IsInstanceOfType` evidence blocks. Also added collection values to all CollectionAssert sections that were missing them (Contains, DoesNotContain, AllItemsAreNotNull, AllItemsAreUnique, IsSubsetOf, etc.).

docs/RFCs/012-Structured-Assertion-Messages.md
```diff
@@ -0,0 +1,1273 @@
...
+
+This keeps copy-paste zero-friction — you can select the value lines and paste them directly into source code without stripping leading whitespace.
+
+Note: When a multi-line value is a string, it is still quoted according to the Value Rendering Rules. The opening quote appears on the line after the label, and the closing quote appears at the end of the value:
+
+```text
+expected:
+"line one\nline two\nline three"
+actual:
+"line one\nLINE TWO\nline three"
```
Evangelink: Good point — this is tracked as **unresolved question #10** (Control character rendering strategy). The three options under consideration are: (1) C#-style escape sequences, (2) Unicode printable replacements like `␊`/`␉` which preserve single-char-width alignment, and (3) keeping actual newlines while escaping other non-printable chars with printable equivalents. Your experience with option 2+3 (printable replacements but keeping newlines) is valuable input for that decision.

docs/RFCs/012-Structured-Assertion-Messages.md
```diff
@@ -0,0 +1,1273 @@
...
+| --------- | ------ |
+| `AreEqual` (int) | `Assertion failed. Expected values to be equal.` |
+| `AreEqual` (string) | `Assertion failed. Expected strings to be equal (case-sensitive).` |
+| `IsTrue` | `Assertion failed. Expected condition to be true.` |
+| `IsNull` | `Assertion failed. Expected value to be null.` |
+| `IsInstanceOfType` | `Assertion failed. Expected value to be an instance of String.` |
+| `ThrowsExactly` | `Assertion failed. Expected exception of type ArgumentException but no exception was thrown.` |
+| `Contains` (collection) | `Assertion failed. Expected collection to contain the specified element.` |
+
+The summary follows the verbal form **"Expected [subject] to [verb phrase]."** for consistency. Short values like type names or small counts can be inlined when they improve clarity without bloating the line.
```
Evangelink: Thanks for the clarification — this is a helpful way to think about the verbal form pattern. Agreed that `AreEqual` doesn't map cleanly to "Assert \<actual\> to \<expected\>" but that's fine.

docs/RFCs/012-Structured-Assertion-Messages.md
```diff
@@ -0,0 +1,1273 @@
...
+collection:  ["apple", "cherry", "date"]
+```
+
+#### Assert.HasCount
+
+```text
+Assertion failed. Expected collection to have 5 element(s) but found 3.
+
+expected count: 5
+actual count:   3
```
Evangelink: The `CollectionAssert` legacy section this comment originally targeted has been removed from scope (see "Scope: Legacy Assertion Classes" note). The remaining `Assert.Contains`, `Assert.DoesNotContain`, `Assert.ContainsSingle`, `Assert.HasCount`, and `Assert.IsEmpty` assertions all already show the collection in their evidence blocks.

Evangelink:
The RFC has matured significantly and I believe we've reached a solid foundation. I'd like to propose we move toward a final review pass.

If there are remaining points of divergence, let's try to resolve them pragmatically — we don't need perfect consensus on every detail to ship a good v1. The structured format is designed to be additive: we can refine rendering rules, diff heuristics, and configuration options in future iterations based on real-world feedback without breaking the overall structure.

Let's aim to get this approved and into implementation. Feedback from actual usage will be far more valuable than further theoretical debate at this point.

cc @nohwnd

docs/RFCs/011-Structured-Assertion-Messages.md
```diff
@@ -0,0 +1,1207 @@
...
+actual:   System.Object (hash: 0x4D5E6F)
+```
+
+Note: Hash codes are non-deterministic across runs in modern .NET. They are useful for same-run debugging but should not be relied upon for snapshot-based test verification of assertion messages.
+
+#### Assert.AreSame (expected is null)
+
+```text
+Assertion failed. Expected both values to refer to the same object.
+Expected is null.
```
nohwnd: expected and actual should be printed here, even if one of them is null. otherwise we are losing info.

unless that is implied. I don't see it in all the AreSame examples.

docs/RFCs/011-Structured-Assertion-Messages.md
```diff
@@ -0,0 +1,1207 @@
...
+expected type: System.String (or derived)
+actual:        null
+```
+
+#### Assert.IsNotInstanceOfType
+
+```text
+Assertion failed. Expected value to not be of type String (or derived).
+
+wrong type:   System.String (or derived)
```
nohwnd: not expected type: ?  (or this is already part of the public api? )

docs/RFCs/011-Structured-Assertion-Messages.md
```diff
@@ -0,0 +1,1207 @@
...
+Assertion failed. Expected exception of type ArgumentException (or derived) but no exception was thrown.
+```
+
+#### Assert.Throws (wrong exception type)
+
+```text
+Assertion failed. Expected exception of type ArgumentException (or derived) but caught InvalidOperationException.
+
+expected type: System.ArgumentException (or derived)
+actual type:   System.InvalidOperationException
```
nohwnd: should show the exception as well, when we have it. otherwise we mask the failure user might want to see (e.g. they did not initialize their code properly)

docs/RFCs/011-Structured-Assertion-Messages.md
```diff
@@ -0,0 +1,1193 @@
+# RFC 011 - Structured Assertion Messages
+
+- [ ] Approved in principle
+- [x] Under discussion
+- [ ] Implementation
+- [ ] Shipped
```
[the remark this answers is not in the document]
Evangelink: @copilot please change number to be 12

nohwnd (approved):

Evangelink:
Thank you @nohwnd

drieseng:
I know I'm late to the party, but wouldn't it have been better to put the user message after the failure detail (expected value, ...)?

The only reason why I use a user message is to provide more details/context in case of an assert failure. For me, the failure itself should be upfront.

This is no big deal though. I do not use user messages that frequently.

Keep up the great work!

Evangelink:
@drieseng this one is tricky because I also use it for extra details but we have quite a few users using it at the main message. It can also push it to far away in case of big actual or assert (e.g. long string or long collection).

We might end up changing that or making it configurable. 

I have seen some framework (I don't recall which language) where users can configure the shape of the assertion message and it felt interesting.

Keep up with reporting suggestions and commenting, it's really important for me to have user feedback.

PR #7772, opened 2026-04-22
Handle empty Assembly.Location on Android CoreCLR

### LINKED ISSUE #7769: MSTest throws NotSupportedException when Assembly.Location returns empty string (Android CoreCLR)
## Description
 
 MSTest throws `NotSupportedException` when `Assembly.Location` returns an empty string. This happens on .NET for Android when using the CoreCLR runtime, where assemblies are 
memory-mapped and don't have a file-backed location.

We discovered this issue when switching the default runtime from MonoVM to CoreCLR in Debug builds in https://github.com/dotnet/android/pull/11169 

 ## Steps To Reproduce
 
 1. Create an Android instrumented test project using MSTest with `EnableMSTestRunner=true`
 2. Use the Microsoft Testing Platform API to run tests:
    ```csharp
    builder.AddMSTest(() => [GetType().Assembly]);
    using ITestApplication app = await builder.BuildAsync();
    await app.RunAsync();

 3. Run the app on an Android device/emulator using the CoreCLR runtime (not MonoVM)

### Expected behavior

Tests are discovered and executed normally, regardless of what Assembly.Location returns.

### Actual behavior

MSTest throws:
```
 System.NotSupportedException: Running tests in any of the provided sources is not supported for the selected platform
```
## Root Cause

The failure chain is:

 1. SynchronizedSingleSessionVSTestBridgedTestFramework.ExecuteRequestAsync calls: _getTestAssemblies().Select(x => x.Location).ToArray()
 2. MSTestDiscovererHelpers.AreValidSources() validates that each source path ends with .dll or .exe (via TestSourceHost.ValidSourceExtensions)
 3. On CoreCLR Android, Assembly.Location returns "" for assemblies loaded from memory-mapped data (via the custom assembly load hook clr_external_assembly_probe)
 4. The empty string fails the .dll/.exe extension check → NotSupportedException

This works with MonoVM because Mono resolves Assembly.Location to a file path even for memory-mapped assemblies. CoreCLR does not.

## Suggested Fix

`AreValidSources` (or the code that calls it) should handle the case where `Assembly.Location` returns an empty string. When assemblies are provided directly via `builder.AddMSTest(() =>
[assembly])`, the framework already has the Assembly reference and should not need a valid file path to discover and run tests.

## Workaround

In dotnet/android app, we can set `<EmbedAssembliesIntoApk>true</EmbedAssembliesIntoApk>` in the project file. This embeds assemblies inside the APK so they are extracted to disk at runtime, making `Assembly.Location`
return a valid path. This disables fast-deploy, so incremental deployment is significantly slower (unusable for day-to-day dev experience).

## Environment

 - MSTest 4.1.0
 - .NET for Android (net11.0-android)
 - CoreCLR runtime on Android (arm64-v8a)
 - Android API 36

nohwnd:
`Assembly.Location` is empty on Android CoreCLR (assemblies are memory-mapped), which makes `AreValidSources` throw `NotSupportedException` because empty string is neither `.dll` nor `.exe`.

Fall back to `<AssemblyName>.dll` when `Location` is empty. Works on .NET Core because `FileOperations.DoesFileExist` always returns `true` there, and `FileOperations.LoadAssembly` loads by name, not by path — so the already-loaded assembly is found.

Added tests for empty / null `Location`, null name, a real dynamic in-memory assembly, and a file-backed one.

Fixes #7769

### DIFF
--- a/src/Platform/Microsoft.Testing.Extensions.VSTestBridge/SynchronizedSingleSessionVSTestAndTestAnywhereAdapter.cs
+++ b/src/Platform/Microsoft.Testing.Extensions.VSTestBridge/SynchronizedSingleSessionVSTestAndTestAnywhereAdapter.cs
@@ -150,7 +150,7 @@ public abstract class SynchronizedSingleSessionVSTestBridgedTestFramework : VSTe
         => ExecuteRequestWithRequestCountGuardAsync(async () =>
         {
 #pragma warning disable IL3000 // Avoid accessing Assembly file path when publishing as a single file
-            string[] testAssemblyPaths = [.. _getTestAssemblies().Select(x => GetAssemblyPath(x))];
+            string[] testAssemblyPaths = [.. _getTestAssemblies().Select(GetAssemblyPath)];
 #pragma warning restore IL3000 // Avoid accessing Assembly file path when publishing as a single file
             switch (request)
             {
@@ -187,7 +187,9 @@ public abstract class SynchronizedSingleSessionVSTestBridgedTestFramework : VSTe
 #pragma warning disable IL3000 // Avoid accessing Assembly file path when publishing as a single file
     internal static string GetAssemblyPath(Assembly assembly)
     {
+#pragma warning disable IL3000 // Avoid accessing Assembly file path when publishing as a single file
         string location = assembly.Location;
+#pragma warning restore IL3000 // Avoid accessing Assembly file path when publishing as a single file
         if (!string.IsNullOrEmpty(location))
         {
             return location;
--- a/src/Platform/Microsoft.Testing.Extensions.VSTestBridge/SynchronizedSingleSessionVSTestAndTestAnywhereAdapter.cs
+++ b/src/Platform/Microsoft.Testing.Extensions.VSTestBridge/SynchronizedSingleSessionVSTestAndTestAnywhereAdapter.cs
@@ -150,7 +150,7 @@ public abstract class SynchronizedSingleSessionVSTestBridgedTestFramework : VSTe
         => ExecuteRequestWithRequestCountGuardAsync(async () =>
         {
 #pragma warning disable IL3000 // Avoid accessing Assembly file path when publishing as a single file
-            string[] testAssemblyPaths = [.. _getTestAssemblies().Select(x => x.Location)];
+            string[] testAssemblyPaths = [.. _getTestAssemblies().Select(x => GetAssemblyPath(x))];
 #pragma warning restore IL3000 // Avoid accessing Assembly file path when publishing as a single file
             switch (request)
             {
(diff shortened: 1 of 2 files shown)


simonrozsival (approved):
This change will unblock us in dotnet/android

Evangelink (approved):
Looks good! The fix correctly solves the reported issue with minimal risk. The synthetic path approach works because on .NET Core, `FileOperations.DoesFileExist` always returns `true` and `FileOperations.LoadAssembly` resolves by assembly name (not file path), so the synthetic `name.dll` path is just a token to pass `AreValidSources` validation.

A few minor suggestions below — none are blocking.

**Summary of suggestions:**
1. **Remove the now-redundant `#pragma warning disable IL3000` at the call site** — the pragma has moved into `GetAssemblyPath`, so the outer pragma is no longer needed.
2. **Use method group** `Select(GetAssemblyPath)` instead of `Select(x => GetAssemblyPath(x))`.
3. **Narrow the pragma scope** inside `GetAssemblyPath` to just the `assembly.Location` line.
4. The hardcoded `.dll` extension is fine for the Android CoreCLR scenario (test assemblies are always DLLs there), but a brief comment documenting this assumption would be helpful for future readers.

src/Platform/Microsoft.Testing.Extensions.VSTestBridge/SynchronizedSingleSessionVSTestAndTestAnywhereAdapter.cs
```diff
@@ -150,7 +150,7 @@ protected sealed override Task ExecuteRequestAsync(TestExecutionRequest request,
         => ExecuteRequestWithRequestCountGuardAsync(async () =>
         {
 #pragma warning disable IL3000 // Avoid accessing Assembly file path when publishing as a single file
-            string[] testAssemblyPaths = [.. _getTestAssemblies().Select(x => x.Location)];
+            string[] testAssemblyPaths = [.. _getTestAssemblies().Select(x => GetAssemblyPath(x))];
```
Evangelink: Nit: Now that `Assembly.Location` is accessed inside `GetAssemblyPath` (which has its own `#pragma`), the outer `#pragma warning disable/restore IL3000` on lines 152 and 154 is redundant and can be removed.

Also, this can use a method group instead of a lambda:
```suggestion
            string[] testAssemblyPaths = [.. _getTestAssemblies().Select(GetAssemblyPath)];
```

src/Platform/Microsoft.Testing.Extensions.VSTestBridge/SynchronizedSingleSessionVSTestAndTestAnywhereAdapter.cs
```diff
@@ -179,6 +179,31 @@ public void Dispose()
...
 
+    /// <summary>
+    /// Gets the path of an assembly, falling back to the assembly name when
+    /// <see cref="Assembly.Location"/> returns an empty string (e.g. on Android CoreCLR
+    /// where assemblies are memory-mapped).
+    /// </summary>
+#pragma warning disable IL3000 // Avoid accessing Assembly file path when publishing as a single file
+    internal static string GetAssemblyPath(Assembly assembly)
+    {
+        string location = assembly.Location;
```
Evangelink, line 191: Nit: Consider narrowing the `#pragma` scope to just this line (the only one that accesses `Assembly.Location`), rather than wrapping the entire method. Keeps the suppression minimal for trimming/AOT analysis:
```suggestion
#pragma warning disable IL3000 // Avoid accessing Assembly file path when publishing as a single file
        string location = assembly.Location;
#pragma warning restore IL3000 // Avoid accessing Assembly file path when publishing as a single file
```

Evangelink (approved):

PR #7787, opened 2026-04-23
Fix TestRunDirectories crash on Android with relative/empty paths

jonathanpeppers:
Continuation of #7772.

On .NET for Android, `Assembly.Location` returns a relative path like `"AndroidTest1.dll"` (MonoVM) or an empty string (CoreCLR). `Path.GetDirectoryName` returns `""` for these inputs, and then `Directory.CreateDirectory("")` throws `ArgumentException`.

This would fail in Release mode on Android, or if the "fast deployment" feature is disabled.

Guard against empty/relative paths by checking the result of `Path.GetDirectoryName` and falling back to the standard `RootDeploymentDirectory/Out` path.

Full stack trace from AZDO build
(https://dev.azure.com/dnceng-public/public/_build/results?buildId=1392743):

    INSTRUMENTATION_RESULT: error=System.ArgumentException: The value cannot be an empty string. (Parameter 'path')
     at System.ArgumentException.ThrowNullOrEmptyException(String argument, String paramName)
     at System.ArgumentException.ThrowIfNullOrEmpty(String argument, String paramName)
     at System.IO.Directory.CreateDirectory(String path)
     at Microsoft.VisualStudio.TestPlatform.MSTestAdapter.PlatformServices.Utilities.FileUtility.CreateDirectoryIfNotExists(String directory)
     at Microsoft.VisualStudio.TestPlatform.MSTestAdapter.PlatformServices.Utilities.DeploymentUtilityBase.CreateDeploymentDirectories(IRunContext runContext, String firstTestSource)
     at Microsoft.VisualStudio.TestPlatform.MSTestAdapter.PlatformServices.TestDeployment.Deploy(IEnumerable`1 testCases, IRunContext runContext, IFrameworkHandle frameworkHandle, ITestSourceHandler testSourceHandler, TestRunCancellationToken cancellationToken)
     at Microsoft.VisualStudio.TestPlatform.MSTest.TestAdapter.MSTestExecutor.RunTestsAsync(...)
     at Microsoft.VisualStudio.TestTools.UnitTesting.MSTestBridgedTestFramework.SynchronizedRunTestsAsync(...)
     at Microsoft.Testing.Extensions.VSTestBridge.SynchronizedSingleSessionVSTestBridgedTestFramework.ExecuteRequestAsync(...)
     at Microsoft.Testing.Platform.Builder.TestApplication.RunAsync()
     at DotNetNewAndroidTestMonoVM.TestInstrumentation.<OnStart>b__2_0()
    INSTRUMENTATION_CODE: 0

Related: dotnet/android#11195 (random PR I saw this), #7769

/cc @nohwnd

### DIFF
--- a/src/Adapter/MSTestAdapter.PlatformServices/Deployment/TestRunDirectories.cs
+++ b/src/Adapter/MSTestAdapter.PlatformServices/Deployment/TestRunDirectories.cs
@@ -45,8 +45,12 @@ internal sealed class TestRunDirectories
         RootDeploymentDirectory = rootDirectory;
         InDirectory = Path.Combine(RootDeploymentDirectory, DeploymentInDirectorySuffix);
 
-        OutDirectory = isAppDomainCreationDisabled && firstTestSource is not null
-            ? Path.GetDirectoryName(firstTestSource)!
+        string? testSourceDirectory = isAppDomainCreationDisabled && !StringEx.IsNullOrEmpty(firstTestSource)
+            ? Path.GetDirectoryName(firstTestSource)
+            : null;
+
+        OutDirectory = !StringEx.IsNullOrEmpty(testSourceDirectory)
+            ? testSourceDirectory!
             : Path.Combine(RootDeploymentDirectory, DeploymentOutDirectorySuffix);
 
         InMachineNameDirectory = Path.Combine(InDirectory, Environment.MachineName);
--- a/src/Adapter/MSTestAdapter.PlatformServices/Deployment/TestRunDirectories.cs
+++ b/src/Adapter/MSTestAdapter.PlatformServices/Deployment/TestRunDirectories.cs
@@ -49,9 +49,9 @@ internal sealed class TestRunDirectories
             ? Path.GetDirectoryName(firstTestSource)
             : null;
 
-        OutDirectory = !StringEx.IsNullOrEmpty(testSourceDirectory)
-            ? testSourceDirectory!
-            : Path.Combine(RootDeploymentDirectory, DeploymentOutDirectorySuffix);
+        OutDirectory = StringEx.IsNullOrEmpty(testSourceDirectory)
+            ? Path.Combine(RootDeploymentDirectory, DeploymentOutDirectorySuffix)
+            : testSourceDirectory;
 
         InMachineNameDirectory = Path.Combine(InDirectory, Environment.MachineName);
     }
--- a/test/UnitTests/MSTestAdapter.PlatformServices.UnitTests/Deployment/TestRunDirectoriesTests.cs
+++ b/test/UnitTests/MSTestAdapter.PlatformServices.UnitTests/Deployment/TestRunDirectoriesTests.cs
@@ -21,4 +21,19 @@ public class TestRunDirectoriesTests : TestContainer
 
     public void InMachineNameDirectoryShouldReturnMachineSpecificDeploymentDirectory()
         => _testRunDirectories.InMachineNameDirectory.Should().Be(Path.Combine(@"C:\temp\In", Environment.MachineName));
+
+    public void OutDirectoryShouldFallBackWhenFirstTestSourceIsEmpty()
+    {
+        // Simulates Android CoreCLR where Assembly.Location returns "" (before synthetic path).
+        var directories = new TestRunDirectories(@"C:\temp", firstTestSource: string.Empty, isAppDomainCreationDisabled: true);
+        directories.OutDirectory.Should().Be(@"C:\temp\Out");
+    }
+
+    public void OutDirectoryShouldFallBackWhenFirstTestSourceIsRelativePath()
+    {
+        // Simulates Android MonoVM / CoreCLR after PR #7772 where Assembly.Location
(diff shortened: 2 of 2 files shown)


src/Adapter/MSTestAdapter.PlatformServices/Deployment/TestRunDirectories.cs
```diff
@@ -45,9 +45,13 @@ public TestRunDirectories(string rootDirectory, string? firstTestSource, bool is
         RootDeploymentDirectory = rootDirectory;
         InDirectory = Path.Combine(RootDeploymentDirectory, DeploymentInDirectorySuffix);
 
-        OutDirectory = isAppDomainCreationDisabled && firstTestSource is not null
-            ? Path.GetDirectoryName(firstTestSource)!
-            : Path.Combine(RootDeploymentDirectory, DeploymentOutDirectorySuffix);
+        string? testSourceDirectory = isAppDomainCreationDisabled && !StringEx.IsNullOrEmpty(firstTestSource)
```
Youssef1313, line 48: I think this code path should normally be not reachable unless you use `DeploymentItemAttribute`. Are you using this attribute? I would personally strongly suggest trying to get rid of it.

Youssef1313: Oh, this is caused by https://github.com/microsoft/testfx/blob/fd42aa03bac038cffd2b2f79b10852f1ee146208/src/Adapter/MSTestAdapter.PlatformServices/Services/TestDeployment.cs#L133-L135

@Evangelink fwiw, I would consider avoid doing this completely, at least under .NET Core.

jonathanpeppers: Somehow `DeploymentEnabled` is on with the configuration:
* https://github.com/dotnet/android/blob/d19a499e5808950c507b7412dc6747fd5d0c5070/src/Microsoft.Android.Templates/androidtest/TestInstrumentation.cs#L35-L41

It's writing a trx to disk and a later step will `adb pull` the file locally.

Copilot seems to think that a .testsettings file like this would turn it off:
```xml
<RunSettings>
  <MSTest>
    <DeploymentEnabled>false</DeploymentEnabled>
  </MSTest>
</RunSettings>
```

Youssef1313: Copilot suggestion is quite correct. You can name the file as `.runsettings` and pass it via `--settings`.

Or alternatively that can be done via `testconfig.json` and be passed via `--config-file`. In  json it can be similar to https://github.com/microsoft/testfx/blob/ddbf1fa776ecdcdcce970e5e585fd88b8c972222/src/Adapter/MSTestAdapter.PlatformServices/Services/MSTestAdapterSettings.cs#L155-L160

But I do believe that in this case, regardless of DeploymentEnabled, MSTest shouldn't really create deployment directories. This can be a good workaround for you though.

Evangelink (commented):
@jonathanpeppers would you mind helping me setup some E2E test for android so we can have something to catch issues/regressions? I am all good with having it as a follow-up.

src/Adapter/MSTestAdapter.PlatformServices/Deployment/TestRunDirectories.cs
Evangelink: > @Evangelink fwiw, I would consider avoid doing this completely, at least under .NET Core.

As I told you, it's been "forced" on us by some internal teams so there is nearly 0 chances to be able to get rid of it.

Youssef1313: @Evangelink This is not about deprecating deployment item.

It's about avoiding to create empty folders under .NET Core when deployment item is not used altogether, which I feel is unlikely to cause issues for anyone.

Evangelink (approved):
I'm moving forward with a merge, please keep raising issues/PR if you face other problems. I renew my request for some E2E test :)

Thank you!

jonathanpeppers:
> would you mind helping me setup some E2E test for android so we can have something to catch issues/regressions? I am all good with having it as a follow-up.

@Evangelink I wouldn't recommend doing it here, just because booting Android emulators are pretty flaky. We do have an end-to-end test in the Android repo where I found this.

Are the packages here pushed to [Maestro/darc](https://maestro.dot.net/)? The dotnet/android repo could consume builds weekly (maybe not daily), and that would be pretty easy to setup.

Evangelink:
Yes, look at the `test-tools` feed (see https://github.com/microsoft/testfx/blob/main/docs/preview.md)

jonathanpeppers:
Ok, I see MSTest 4.3.0-preview.26224.2 on the ".NET Core Tooling Dev" darc channel, "test-tools" NuGet feed. I'll give it a try.

jonathanpeppers:
This will let me dogfood the fix here as well, thanks:
* https://github.com/dotnet/android/pull/11209

PR #7810, opened 2026-04-24
Add MSTest.Windows.AppTesting thin-layer for Windows desktop E2E testing

Evangelink:
## Summary

Introduces a new `MSTest.Windows.AppTesting` NuGet package that provides base classes for end-to-end testing of Windows desktop applications (WinForms, WPF, Win32). Follows the same thin-layer pattern as `Microsoft.Playwright.MSTest` — a lifecycle adapter over the built-in `System.Windows.Automation` API with zero external dependencies.

## Base class hierarchy

```
AutomationTest          ← PlaywrightTest equivalent (hierarchy root, extensible)
  └─ ApplicationTest    ← BrowserTest equivalent (process launch/teardown)
       └─ WindowTest    ← PageTest equivalent (MainWindow as AutomationElement)
```

- **`AutomationTest`** — root of the hierarchy, extension point for future configuration
- **`ApplicationTest`** — launches the app via `Process.Start`, waits for main window handle, manages teardown with `CloseMainWindow`/`Kill`. Uses `[STATestClass]` for COM/STA threading. Configurable via virtual properties (`ApplicationPath`, `ApplicationArguments`, `ApplicationStartTimeout`)
- **`WindowTest`** — exposes `MainWindow` as `System.Windows.Automation.AutomationElement`. Users can layer FlaUI or other libraries on top for richer element interaction

## SDK integration

Mirrors the Playwright/Aspire SDK pattern exactly:

- `<EnableWindowsAppTesting>true</EnableWindowsAppTesting>` MSBuild property in MSTest.Sdk
- `WindowsAppTesting.targets` feature file wired into ClassicEngine, VSTest, and NativeAOT targets
- Package version templated through `Sdk.props.template`
- Works as both an SDK feature flag and a standalone `<PackageReference>`
- Includes a build-time error when targeting non-Windows TFMs

## User experience

```xml
<Project Sdk="MSTest.Sdk/4.3.0">
  <PropertyGroup>
    <TargetFramework>net8.0-windows</TargetFramework>
    <EnableWindowsAppTesting>true</EnableWindowsAppTesting>
  </PropertyGroup>
</Project>
```

```csharp
[TestClass]
public class MyAppTests : WindowTest
{
    public override string ApplicationPath => @"C:\MyApp\MyApp.exe";

    [TestMethod]
    public void MainWindow_IsVisible()
    {
        Assert.IsNotNull(MainWindow);
    }
}
```

## What's included

- **New project**: `src/TestFramework/TestFramework.Windows.AppTesting/` — the package source
- **SDK wiring**: Feature targets, props template, NativeAOT guard
- **Sample**: `samples/public/DemoMSTestSdk/ProjectUsingWindowsAppTesting/` — Calculator E2E test
- **Acceptance tests**: `WindowsAppTestingSdkTests.cs` — tests both MTP runner and VSTest modes using notepad.exe

## Design decisions

- **`System.Windows.Automation` over FlaUI**: Zero external dependencies. Uses the managed UIA wrapper from the Windows Desktop runtime. Users who want a richer API can layer FlaUI on top.
- **`[STATestClass]`**: Required for COM interop with Windows UI Automation
- **Separate package + SDK flag**: Ships from the same repo, usable both ways
- **`EnableWindowsTargeting`**: Always sets Windows TFMs with `EnableWindowsTargeting` for non-Windows CI compatibility
- **Race-safe teardown**: `ApplicationTearDown` uses try/finally with `InvalidOperationException` handling for process exit races


test/IntegrationTests/MSTest.Acceptance.IntegrationTests/WindowsAppTestingSdkTests.cs
```diff
@@ -0,0 +1,105 @@
...
+            .PatchCodeWithReplace("$MSTestVersion$", MSTestVersion)
+            .PatchCodeWithReplace("$TargetFramework$", DesktopTargetFramework)
+            .PatchCodeWithReplace("$ExtraProperties$", string.Empty));
+
+        DotnetMuxerResult compilationResult = await DotnetCli.RunAsync(
+            $"test --no-progress --no-ansi -c Release {testAsset.TargetAssetPath}",
+            workingDirectory: testAsset.TargetAssetPath,
+            cancellationToken: TestContext.CancellationToken);
+
+        compilationResult.AssertExitCodeIs(ExitCodes.Success);
```
Evangelink: **[Coverage]** The MTP runner test only asserts the exit code — it does not verify how many tests actually ran.

**Impact**: If the generated test project were misconfigured and 0 tests ran (e.g., `[OSCondition]` silently skipped everything, or the test class wasn't discovered), the exit code would still be `0` and this test would pass silently.

The comparable `AspireSdkTests` and `PlaywrightSdkTests` tests use `AssertOutputContainsSummary` for the MTP runner case:
```csharp
testHostResult.AssertOutputContainsSummary(0, 2, 0);
```

**Suggestion**: Add a count assertion similar to the VSTest test at line 102:
```csharp
compilationResult.AssertExitCodeIs(ExitCodes.Success);
compilationResult.AssertOutputContainsSummary(failed: 0, passed: 2, skipped: 0);
// or the string form:
compilationResult.AssertOutputContains("passed: 2");
```

test/IntegrationTests/MSTest.Acceptance.IntegrationTests/WindowsAppTestingSdkTests.cs
```diff
@@ -0,0 +1,105 @@
...
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using Microsoft.Testing.Platform.Acceptance.IntegrationTests;
+using Microsoft.Testing.Platform.Acceptance.IntegrationTests.Helpers;
+using Microsoft.Testing.Platform.Helpers;
+
+namespace MSTest.Acceptance.IntegrationTests;
+
+[TestClass]
+public sealed class WindowsAppTestingSdkTests : AcceptanceTestBase<NopAssetFixture>
```
Evangelink: **[Structure]** This class uses `NopAssetFixture` and calls `TestAsset.GenerateAssetAsync` inside each test method, which means the asset is compiled from scratch on every test invocation. The established pattern for SDK feature tests in this repo (`AspireSdkTests`, `PlaywrightSdkTests`) uses a dedicated inner `TestAssetFixture : TestAssetFixtureBase` that generates the asset once in `ClassInitialize`.

**Impact**: Each of the two tests will independently restore NuGet packages, compile the test project, and run it — doubling the build cost with no benefit.

**Suggestion**: Follow the `AspireSdkTests` pattern:
```csharp
public sealed class WindowsAppTestingSdkTests : AcceptanceTestBase<WindowsAppTestingSdkTests.TestAssetFixture>
{
    public sealed class TestAssetFixture() : TestAssetFixtureBase()
    {
        public const string ProjectName = "WindowsAppTestingSdk";
        public string ProjectPath => GetAssetPath(ProjectName);
        public override (string ID, string Name, string Code) GetAssetsToGenerate() => ...;
    }
}
```
Then both tests can call `TestHost.LocateFrom(AssetFixture.ProjectPath, ...)` to run against the already-built asset.

samples/public/DemoMSTestSdk/ProjectUsingWindowsAppTesting/CalculatorTests.cs
```diff
@@ -0,0 +1,50 @@
...
+        // Use non-localized AutomationId values to locate UI elements.
+        AutomationElement? button1 = MainWindow.FindFirst(
+            TreeScope.Descendants,
+            new PropertyCondition(AutomationElement.AutomationIdProperty, "num1Button"));
+
+        AutomationElement? button2 = MainWindow.FindFirst(
+            TreeScope.Descendants,
+            new PropertyCondition(AutomationElement.AutomationIdProperty, "num2Button"));
+
+        Assert.IsNotNull(button1, "Could not find the button with AutomationId 'num1Button'.");
```
Evangelink, line 59: **[Flakiness]** On Windows 10/11, `calc.exe` launches the Store/UWP Calculator, not the classic Win32 Calculator. The UWP Calculator's `AutomationElement` tree uses different automation IDs (`num1Button` / `num2Button` were from the classic Calculator). `MainWindow.FindFirst(...)` will return `null` for both, causing this test to always fail on modern Windows.

**Impact**: This sample test is broken on any modern Windows installation, which will confuse users trying to follow it as a reference.

**Suggestion**: Either replace `calc.exe` with a Win32 app whose automation structure is stable (e.g., `notepad.exe`, consistent with the acceptance test), or remove the `Calculator_NumberButtons_AreVisible` test and add a comment explaining that the button lookup is illustrative and may need to be adapted for the specific app version. For example:
```csharp
// Modern Windows Calculator (UWP) uses different AutomationIds.
// Replace "num1Button" with the actual ID for your target app.
```

Evangelink (commented):
### Summary

**Date**: 2026-05-11

### Key Findings

1. **[Coverage] MTP runner test has no test-count assertion** (`WindowsAppTestingSdkTests.cs:80`) — Only `AssertExitCodeIs` is called. If 0 tests ran, this silently passes. The VSTest counterpart correctly asserts `"Passed: 2"`. `AspireSdkTests` and `PlaywrightSdkTests` both use `AssertOutputContainsSummary` for the MTP case.

2. **[Structure] `NopAssetFixture` + per-test `GenerateAssetAsync` diverges from repo pattern** (`WindowsAppTestingSdkTests.cs:11`) — Every test independently compiles the test asset from source. The repo's `AspireSdkTests` / `PlaywrightSdkTests` use an inner `TestAssetFixture : TestAssetFixtureBase` that generates the asset once in `ClassInitialize`, then both tests locate and run against the pre-built asset.

3. **[Flakiness] `CalculatorTests.cs` automation IDs are broken on modern Windows** (`CalculatorTests.cs:47`) — `calc.exe` on Windows 10/11 is a UWP app whose element tree does not contain `num1Button` / `num2Button`. The sample test will always fail as written on any modern Windows installation, which undermines its value as reference documentation.

### Recommendations

- Add `AssertOutputContainsSummary(0, 2, 0)` (or equivalent) after the `AssertExitCodeIs` in the MTP runner test.
- Refactor `WindowsAppTestingSdkTests` to use the `TestAssetFixture` pattern to match Aspire/Playwright tests and avoid redundant per-test compilation.
- Replace the broken `Calculator_NumberButtons_AreVisible` test in the sample with either a stable Win32 app (e.g., `notepad.exe`) or add a clear comment that the automation IDs are app-version-specific and must be adapted.

---

*Generated by [Test Expert Reviewer 🧪](https://github.com/microsoft/testfx/actions/runs/25696190605)*

> 🧪 *Test quality reviewed by [Test Expert Reviewer 🧪](https://github.com/microsoft/testfx/actions/runs/25696190605)*

Evangelink (commented):
### Summary

**Date**: 2026-05-11

### Key Findings

#### 🔴 Correctness — Critical

**`[TestClass]` on `WindowTest` silently drops STA threading for the entire hierarchy** (`WindowTest.cs` line 42)

`STATestClassAttribute` is declared with `[AttributeUsage(Inherited = false)]`. Because of this:
- `GetCustomAttributes(inherit: true)` on `WindowTest` does **not** return `[STATestClass]` from `ApplicationTest`
- MSTest resolves `ClassAttribute` as a plain `TestClassAttribute`, so `ClassAttribute is STATestClassAttribute` is `false`
- `GetTestMethodAttribute` returns a regular `TestMethodAttribute` — tests run on MTA, not STA

The same logic applies to every user class that writes `[TestClass] class MyTests : WindowTest`. All of them will run on MTA, causing Windows UI Automation COM interop to fail with `InvalidOperationException` or `COMException (0x80010106)` at runtime.

**Fix needed**: Change `[TestClass]` → `[STATestClass]` on `WindowTest` (same for `ApplicationTest` is already correct). Document clearly that user test classes must also use `[STATestClass]`. Longer term, consider changing `STATestClassAttribute` to `Inherited = true` so the constraint propagates automatically.

---

#### 🟡 Defensive — Important

**`AutomationElement.FromHandle` can return `null`** (`WindowTest.cs` line 58)

If the window handle becomes invalid between `ApplicationSetup` and `WindowSetup`, `FromHandle` returns `null`, which is silently stored in the non-nullable `MainWindow` property. Tests then crash with an uninformative `NullReferenceException`. A null check with a descriptive `InvalidOperationException` is needed.

---

#### 🟡 Resources — Minor

**`Process.Kill()` without `entireProcessTree: true`** (`ApplicationTest.cs` line 107)

Desktop applications often spawn child processes. Killing only the top-level process leaves children alive (handles, file locks), which can cause subsequent tests to fail. Use `Kill(entireProcessTree: true)` — available on both `net8.0-windows` and `net9.0-windows`.

---

#### 🟡 Correctness — Minor

**Early process exit not detected in setup polling loop** (`ApplicationTest.cs` line 69)

If the app crashes at startup, the loop spins for the full `ApplicationStartTimeout` and then throws a misleading `TimeoutException`. Adding an `AppProcess.HasExited` check inside the loop and surfacing the exit code gives developers actionable diagnostics.

---

### Positive Observations

- The thin-layer pattern (mirroring Playwright/Aspire) is a clean and familiar design for MSTest users.
- The teardown `try/finally` with `InvalidOperationException` handling correctly handles the race between `HasExited` checks and process operations.
- SDK integration (ClassicEngine, VSTest, NativeAOT guard, `Sdk.props.template`) is done consistently with the Playwright/Aspire pattern.
- `EnableWindowsTargeting` for non-Windows CI builds is a thoughtful touch.
- `PublicAPI.Unshipped.txt` is properly maintained.

---

*Generated by [Expert Code Reviewer](https://github.com/microsoft/testfx/actions/runs/25696190614)*

> 🧠 *Reviewed by [Expert Code Reviewer 🧠](https://github.com/microsoft/testfx/actions/runs/25696190614)*

src/TestFramework/TestFramework.Windows.AppTesting/WindowTest.cs
```diff
@@ -0,0 +1,59 @@
...
+///         Assert.IsNotNull(result);
+///     }
+/// }
+/// </code>
+/// <para>
+/// For a richer element interaction API, add a library like FlaUI on top and use
+/// <see cref="MainWindow"/> handle to bridge into it.
+/// </para>
+/// </remarks>
+[TestClass]
```
Evangelink: **[Correctness — Critical]** `[TestClass]` here silently drops STA threading for `WindowTest` and every class that derives from it.

`STATestClassAttribute` is declared with `[AttributeUsage(Inherited = false)]` (see `STATestClassAttribute.cs`). This means `GetCustomAttributes(inherit: true)` on `WindowTest` — or on any user class like `[TestClass] class MyTests : WindowTest` — does **not** return the `[STATestClass]` that sits on `ApplicationTest`. MSTest's `TypeCache` calls `GetSingleAttributeOrDefault<TestClassAttribute>(classType)` on the concrete class; the result is a plain `TestClassAttribute`, so `ClassAttribute is STATestClassAttribute` evaluates to `false`, and `GetTestMethodAttribute` returns a regular `TestMethodAttribute` instead of `STATestMethodAttribute`. All test methods end up running on an MTA thread.

Windows UI Automation is COM-based and **requires an STA thread** for `AutomationElement.FromHandle`, `FindFirst`, and any pattern invocation. On MTA you will see `InvalidOperationException: The calling thread must be an STA thread` or `COMException (0x80010106)`.

**Impact**: Every test that inherits from `WindowTest` (including the samples and acceptance tests in this PR) will fail at runtime with a COM threading exception.

**Suggestion**: Change the attribute on `WindowTest` to `[STATestClass]` (instead of `[TestClass]`). Also update the documentation/samples to show users that their own test class must be annotated `[STATestClass]` (not `[TestClass]`), because the `Inherited = false` constraint means the attribute still cannot flow down from `WindowTest` to user classes automatically. Alternatively, consider changing `STATestClassAttribute.AttributeUsage` to `Inherited = true` in a companion change so the whole hierarchy inherits STA by default.

src/TestFramework/TestFramework.Windows.AppTesting/WindowTest.cs
```diff
@@ -0,0 +1,59 @@
...
+    public AutomationElement MainWindow { get; private set; } = null!;
+
+    /// <summary>
+    /// Obtains the application's main window <see cref="AutomationElement"/> before each test method.
+    /// </summary>
+    [TestInitialize]
+    public void WindowSetup()
+    {
+        MainWindow = AutomationElement.FromHandle(AppProcess.MainWindowHandle);
+    }
```
Evangelink: **[Defensive — Important]** `AutomationElement.FromHandle` can return `null`.

The Windows UIA documentation states that `FromHandle` returns `null` when the element is not accessible or the handle is stale. The window that was confirmed in `ApplicationSetup` can become inaccessible (e.g., if the process is killed or crashes between the two `[TestInitialize]` calls). `MainWindow` is typed as `AutomationElement` (non-nullable), so a `null` from `FromHandle` is silently stored as a `null!` value. Every subsequent call in a test that does `MainWindow.Current.Name` will then throw an uninformative `NullReferenceException` with no hint that the problem was the automation element, not the test code itself.

**Suggestion**:
```csharp
MainWindow = AutomationElement.FromHandle(AppProcess.MainWindowHandle)
    ?? throw new InvalidOperationException(
        $"Could not obtain an AutomationElement for the main window of '{ApplicationPath}' (handle 0x{AppProcess.MainWindowHandle:X}). The process may have exited.");
```

src/TestFramework/TestFramework.Windows.AppTesting/ApplicationTest.cs
```diff
@@ -0,0 +1,121 @@
...
+            if (appProcess is not null && !appProcess.HasExited)
+            {
+                try
+                {
+                    _ = appProcess.CloseMainWindow();
+                    if (!appProcess.WaitForExit(5000))
+                    {
+                        appProcess.Kill();
+                        _ = appProcess.WaitForExit(5000);
+                    }
```
Evangelink, line 120: **[Resources — Minor]** `Process.Kill()` without `entireProcessTree: true` orphans child processes.

Desktop applications (especially WPF/WinForms apps and Win32 launchers) routinely spawn helper or worker child processes (COM servers, crash handlers, update agents, etc.). When `CloseMainWindow` fails and the top-level process is force-killed without `entireProcessTree: true`, those child processes remain alive, consuming handles and potentially locking files, which can cause subsequent test runs to fail or accumulate zombie processes over a long test session.

`entireProcessTree` is available on .NET 5+ (which covers both `net8.0-windows` and `net9.0-windows` targets of this library).

**Suggestion**:
```csharp
appProcess.Kill(entireProcessTree: true);
```

src/TestFramework/TestFramework.Windows.AppTesting/ApplicationTest.cs
```diff
@@ -0,0 +1,121 @@
...
+            startInfo.Arguments = ApplicationArguments;
+        }
+
+        AppProcess = Process.Start(startInfo)
+            ?? throw new InvalidOperationException($"Failed to start process: {ApplicationPath}");
+
+        // Wait for the main window to be created
+        Stopwatch sw = Stopwatch.StartNew();
+        while (AppProcess.MainWindowHandle == IntPtr.Zero && sw.Elapsed < ApplicationStartTimeout)
+        {
```
Evangelink, line 76: **[Correctness — Minor]** The polling loop does not detect early process exit, giving a misleading `TimeoutException`.

If the application crashes at startup (e.g., missing runtime, missing DLL, bad config), `AppProcess.HasExited` becomes `true` almost immediately. The current loop will continue spinning for the full `ApplicationStartTimeout` (default 10 s), then throw:

> `TimeoutException: Application 'foo.exe' did not create a main window within 00:00:10.`

This message implies the app is still running but slow, not that it crashed. The actual exit code that would explain the failure is discarded.

**Suggestion** — add an early-exit check inside the loop:
```csharp
while (AppProcess.MainWindowHandle == IntPtr.Zero && sw.Elapsed < ApplicationStartTimeout)
{
    AppProcess.Refresh();
    if (AppProcess.HasExited)
    {
        throw new InvalidOperationException(
            $"Application '{ApplicationPath}' exited with code {AppProcess.ExitCode} before a main window was created.");
    }
    // ... sleep
}
```

samples/public/DemoMSTestSdk/ProjectUsingWindowsAppTesting/CalculatorTests.cs
Evangelink: Fixed. Added a `<remarks>` doc comment explaining that on modern Windows 10/11, `calc.exe` launches the Store/UWP Calculator with different AutomationIds than the classic Win32 Calculator. Also added an inline comment on the `Calculator_NumberButtons_AreVisible` test noting that the AutomationIds are illustrative and need to be replaced for the actual target app version.

test/IntegrationTests/MSTest.Acceptance.IntegrationTests/WindowsAppTestingSdkTests.cs
Evangelink: Fixed. Refactored to follow the `AspireSdkTests`/`PlaywrightSdkTests` pattern with a proper inner `TestAssetFixture : TestAssetFixtureBase` class. The asset is now generated and built once in `ClassInitialize`, and both tests reuse it via `TestHost.LocateFrom`.

test/IntegrationTests/MSTest.Acceptance.IntegrationTests/WindowsAppTestingSdkTests.cs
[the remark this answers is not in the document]
Evangelink: Fixed. Refactored to use `TestHost.LocateFrom` + `ExecuteAsync` for the MTP runner test, and added `testHostResult.AssertOutputContainsSummary(failed: 0, passed: 2, skipped: 0)` to verify test counts.

src/TestFramework/TestFramework.Windows.AppTesting/ApplicationTest.cs
```diff
@@ -0,0 +1,121 @@
...
+        try
+        {
+            if (appProcess is not null && !appProcess.HasExited)
+            {
+                try
+                {
+                    _ = appProcess.CloseMainWindow();
+                    if (!appProcess.WaitForExit(5000))
+                    {
+                        appProcess.Kill();
```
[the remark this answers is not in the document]
Evangelink: Fixed. Changed to `appProcess.Kill(entireProcessTree: true)` to ensure child processes are terminated during teardown.

src/TestFramework/TestFramework.Windows.AppTesting/WindowTest.cs
Evangelink: Fixed. Changed `[TestClass]` to `[STATestClass]` on `WindowTest`. Also updated:
- The XML doc `<code>` example to show `[STATestClass]` instead of `[TestClass]`
- The `NotepadTests` in the acceptance test source to use `[STATestClass]`
- The `CalculatorTests` sample to use `[STATestClass]`

src/TestFramework/TestFramework.Windows.AppTesting/ApplicationTest.cs
Evangelink: Fixed. Changed to `appProcess.Kill(entireProcessTree: true)` to avoid orphaning child processes during teardown.

test/IntegrationTests/MSTest.Acceptance.IntegrationTests/WindowsAppTestingSdkTests.cs
Evangelink: Fixed. Refactored to use `TestHost.LocateFrom` + `ExecuteAsync` for the MTP runner test, and added `testHostResult.AssertOutputContainsSummary(failed: 0, passed: 2, skipped: 0)` — matching the pattern from `AspireSdkTests` and `PlaywrightSdkTests`.

src/TestFramework/TestFramework.Windows.AppTesting/ApplicationTest.cs
Evangelink: Fixed. Added an early-exit check inside the polling loop that detects `AppProcess.HasExited` and throws an `InvalidOperationException` with the exit code, instead of spinning for the full timeout and giving a misleading `TimeoutException`.

src/TestFramework/TestFramework.Windows.AppTesting/WindowTest.cs
Evangelink: Fixed. Added null-coalescing throw with a descriptive message:
```csharp
MainWindow = AutomationElement.FromHandle(AppProcess.MainWindowHandle)
    ?? throw new InvalidOperationException(
        $"Could not obtain an AutomationElement for the main window of '{ApplicationPath}' (handle 0x{AppProcess.MainWindowHandle:X}). The process may have exited.");
```

Evangelink (commented):
### Summary

**Workflow**: Expert Code Reviewer 🧠
**Date**: 2026-05-12

### Key Findings

All four issues flagged in the previous expert review (run [25696190614](https://github.com/microsoft/testfx/actions/runs/25696190614)) have been fully addressed in the latest commits:

| Previous Finding | Status |
|---|---|
| `[TestClass]` on `WindowTest` silently drops STA threading for the entire hierarchy | ✅ Fixed — `[STATestClass]` applied; docs and samples updated |
| `AutomationElement.FromHandle` null return not handled | ✅ Fixed — null-coalescing throw with descriptive message added |
| `Process.Kill()` without `entireProcessTree: true` | ✅ Fixed — child processes now terminated on force-kill |
| Early process exit not detected in polling loop | ✅ Fixed — `HasExited` check with exit code in error message |

Additionally, both Test Expert findings have been addressed: `AssertOutputContainsSummary` is now called in the MTP runner test, and the `TestAssetFixture` pattern has been adopted to match `AspireSdkTests`/`PlaywrightSdkTests`.

### Positive Observations

- **Correctness**: `ApplicationSetup`/`WindowSetup` `[TestInitialize]` ordering is correct — MSTest calls base-class initializers first, so the process is guaranteed to be running before the `AutomationElement` is obtained.
- **Teardown**: `try/finally` with `InvalidOperationException` guard correctly handles the race between `HasExited` and process operations; `WaitForExit(5000)` after `Kill` avoids zombie handles.
- **SDK integration**: `WindowsAppTesting.targets` correctly validates the Windows TFM at build time and is guarded from NativeAOT; the `Sdk.props.template` versioning aligns with the Playwright/Aspire pattern.
- **Acceptance test**: `NotepadTests` uses a stable Win32 app and validates both a non-null `MainWindow` and a non-empty title — robust against Notepad UI changes.

### No New Issues

No correctness, threading, performance, API compatibility, resource management, or security issues were found in the updated code.

---

*Generated by [Expert Code Reviewer 🧠](https://github.com/microsoft/testfx/actions/runs/25723095444)*

> 🧠 *Reviewed by [Expert Code Reviewer 🧠](https://github.com/microsoft/testfx/actions/runs/25723095444)*

> 🧠 *Reviewed by [Expert Code Reviewer 🧠](https://github.com/microsoft/testfx/actions/runs/25723095444)*

samples/public/DemoMSTestSdk/ProjectUsingWindowsAppTesting/CalculatorTests.cs
```diff
@@ -0,0 +1,60 @@
...
+        Assert.IsNotNull(button2, "Could not find the button with AutomationId 'num2Button'.");
+    }
```
Evangelink, line 61: **[Flakiness]** `Calculator_NumberButtons_AreVisible` will **always fail** on any modern Windows machine. The UWP Calculator shipped with Windows 10/11 does not expose `num1Button` / `num2Button` in its accessibility tree — `FindFirst` will return `null` for both, and both assertions will fail. The class-level `<remarks>` block warns about this, but comments alone do not prevent the test from failing.

**Impact**: The primary usage sample ships as a test that fails immediately when run by any user who follows the documentation. A broken "hello world" sample is worse than no sample.

**Suggestions**:
- Use a stable, universally available Win32 app for the sample, e.g. `notepad.exe` (already used in the acceptance tests). The acceptance tests show a solid two-test pattern against Notepad that works on all Windows versions.
- If keeping Calculator as the subject is important for illustration, mark the test `[Ignore("Replace AutomationIds with your app's actual values before running")]` so it fails loudly at intent rather than silently producing confusing automation errors.
- Alternatively, keep the file as-is but do **not** include `Calculator_NumberButtons_AreVisible` in the compiled sample project — only include the `Calculator_MainWindow_HasExpectedTitle` test, which can pass on UWP Calculator since the main window handle is valid.

test/IntegrationTests/MSTest.Acceptance.IntegrationTests/WindowsAppTestingSdkTests.cs
```diff
@@ -0,0 +1,107 @@
...
+
+[STATestClass]
+public class NotepadTests : WindowTest
+{
+    public override string ApplicationPath => "notepad.exe";
+
+    [TestMethod]
+    public void Notepad_MainWindow_IsNotNull()
+    {
+        Assert.IsNotNull(MainWindow);
```
Evangelink: **[Assertion]** `Assert.IsNotNull(MainWindow)` is a vacuous assertion here. `WindowSetup()` (the `[TestInitialize]` method in `WindowTest`) already throws `InvalidOperationException` if `AutomationElement.FromHandle` returns null — the assertion is only reached when `MainWindow` is already guaranteed non-null. If the window setup fails, the test will fail with an exception, not with this assertion.

**Impact**: False confidence — the assertion looks like it is verifying behaviour, but it can never be the actual source of a failure. The failing case is silently absorbed by the `[TestInitialize]` throw path.

**Suggestion**: Assert something the `MainWindow` element actually exposes, which would survive a future change to `WindowSetup`'s null-guard:
```csharp
public void Notepad_MainWindow_IsNotNull()
{
    Assert.AreEqual(ControlType.Window, MainWindow.Current.ControlType,
        "Expected the main window element to be of control type Window.");
}
```
Or drop this test entirely and fold the non-null guarantee into the `Notepad_MainWindow_HasExpectedTitle` test (already sufficient on its own).

test/IntegrationTests/MSTest.Acceptance.IntegrationTests/WindowsAppTestingSdkTests.cs
```diff
@@ -0,0 +1,107 @@
...
+
+    [TestMethod]
+    [OSCondition(OperatingSystems.Windows, IgnoreMessage = "Windows app testing is Windows-only")]
+    public async Task EnableWindowsAppTesting_WhenUsingMSTestRunner_RunsDesktopTests()
+    {
+        var testHost = TestHost.LocateFrom(AssetFixture.ProjectPath, TestAssetFixture.ProjectName, DesktopTargetFramework);
+        TestHostResult testHostResult = await testHost.ExecuteAsync(cancellationToken: TestContext.CancellationToken);
+
+        testHostResult.AssertExitCodeIs(ExitCode.Success);
+        testHostResult.AssertOutputContainsSummary(failed: 0, passed: 2, skipped: 0);
```
Evangelink, line 30: **[Coverage]** Both the MTP runner test and the VSTest test share the same pre-built test asset, and that asset's `global.json` specifies `runner: VSTest`. This means both acceptance tests verify a single configuration (VSTest-mode build). The MTP runner test runs the binary directly via `TestHost.LocateFrom` / `ExecuteAsync`, which bypasses the `dotnet test` layer — so it works correctly here. However, there is no acceptance test covering the case where `global.json` omits the `runner` key (the default MTP-native mode). If a future MSTest.Sdk change alters how `EnableWindowsAppTesting` interacts with the default runner selection, neither test would catch the regression.

**Suggestion**: Consider a second `TestAssetFixture` variant (or a parameterised asset) that omits the `global.json` / uses `runner: MTPRunner`, so the MTP path is tested end-to-end with its own build artifact, mirroring how other SDK feature tests distinguish runner configurations.

Evangelink (commented):
### Summary

**Date**: 2026-05-12

### Key Findings

The PR has been significantly improved since the previous review (the `AssertOutputContainsSummary` gap and `TestAssetFixture` pattern have both been addressed ✅). Three issues remain:

1. **[Assertion] `Notepad_MainWindow_IsNotNull` is a vacuous assertion** (`WindowsAppTestingSdkTests.cs` embedded NotepadTests, line 79) — `WindowSetup` already guarantees `MainWindow` is non-null (throws if not). The assertion can never be the actual cause of a test failure; it gives false confidence of coverage.

2. **[Flakiness] `Calculator_NumberButtons_AreVisible` always fails on modern Windows** (`CalculatorTests.cs` lines 44-59) — The UWP Calculator (`calc.exe` on Windows 10/11) does not expose `num1Button`/`num2Button` in its automation tree. The class-level XML doc notes this, but a broken sample test is more confusing than helpful. Using `notepad.exe` (already stable in the acceptance tests) or adding `[Ignore]` would prevent users from running into an immediate, unexplained failure.

3. **[Coverage] Single build artifact used for both MTP and VSTest acceptance tests** (`WindowsAppTestingSdkTests.cs` line 25) — Both tests share an asset configured with `runner: VSTest` in `global.json`. The MTP runner path (`TestHost.ExecuteAsync`) works because it bypasses `dotnet test`, but no acceptance test covers a default/MTP-native build, so an MSTest.Sdk regression in that mode would go undetected.

### Recommendations

- Replace the vacuous `Assert.IsNotNull(MainWindow)` with a meaningful assertion (e.g., `Assert.AreEqual(ControlType.Window, MainWindow.Current.ControlType)`), or remove the redundant test and rely on the title test alone.
- Replace `calc.exe` in the sample with `notepad.exe` (or apply `[Ignore]` to the broken test) so the shipped sample is immediately runnable.
- Add a second test asset variant (or parameterise the existing one) that builds without `runner: VSTest` to cover the default MTP runner path end-to-end.

---

> 🧪 *Test quality reviewed by [Test Expert Reviewer 🧪](https://github.com/microsoft/testfx/actions/runs/25723095420)*

> 🧪 *Test quality reviewed by [Test Expert Reviewer 🧪](https://github.com/microsoft/testfx/actions/runs/25723095420)*

samples/public/DemoMSTestSdk/ProjectUsingWindowsAppTesting/CalculatorTests.cs
Evangelink: Fixed. Added `[Ignore("AutomationIds are app-version-specific. Replace 'num1Button'/'num2Button' with actual IDs for your target application.")]` to `Calculator_NumberButtons_AreVisible` so it fails at intent rather than with confusing automation errors on modern Windows.

src/TestFramework/TestFramework.Windows.AppTesting/ApplicationTest.cs
```diff
@@ -0,0 +1,126 @@
...
+
+    /// <summary>
+    /// Gets the path to the application executable to launch.
+    /// Override this property to specify the application under test.
+    /// Defaults to the <c>DESKTOP_TEST_APP_PATH</c> environment variable.
+    /// </summary>
+    public virtual string ApplicationPath
+        => Environment.GetEnvironmentVariable(AppPathEnvVar)
+           ?? throw new InvalidOperationException(
+               $"Override {nameof(ApplicationPath)} or set the '{AppPathEnvVar}' environment variable to the path of the application under test.");
```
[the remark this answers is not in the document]
Evangelink: Fixed. Changed `ApplicationPath` to use `string.IsNullOrWhiteSpace` instead of the null-coalescing operator, so empty/whitespace values from the environment variable are treated as unset and throw the same descriptive `InvalidOperationException`.

test/IntegrationTests/MSTest.Acceptance.IntegrationTests/WindowsAppTestingSdkTests.cs
Evangelink: Good point. The current tests do exercise both runner paths (MTP via `TestHost.ExecuteAsync` and VSTest via `dotnet test`), but both use the same VSTest-mode build artifact. Adding a second asset variant without `global.json` (MTP-native mode) would improve coverage. I'll track this as a follow-up improvement — the current setup already catches most regressions since the `EnableWindowsAppTesting` SDK wiring is the same regardless of runner, and the MTP test executes the binary directly.

test/IntegrationTests/MSTest.Acceptance.IntegrationTests/WindowsAppTestingSdkTests.cs
Evangelink: Fixed. Replaced the vacuous `Assert.IsNotNull(MainWindow)` with a meaningful assertion:
```csharp
Assert.AreEqual(ControlType.Window, MainWindow.Current.ControlType,
    "Expected the main window element to be of control type Window.");
```
Also renamed the test to `Notepad_MainWindow_IsVisible` to better reflect what it now verifies.

src/Package/MSTest.Sdk/Sdk/Features/WindowsAppTesting.targets
```diff
@@ -0,0 +1,31 @@
...
+<Project xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+
+  <PropertyGroup>
+    <_IsWindowsTfm>false</_IsWindowsTfm>
+    <_IsWindowsTfm Condition="$(TargetFramework.Contains('-windows'))">true</_IsWindowsTfm>
+  </PropertyGroup>
+
+  <Target Name="_ValidateWindowsAppTestingPlatform" BeforeTargets="Restore;Build"
+          Condition=" '$(_IsWindowsTfm)' != 'true' ">
+    <Error Text="MSTest.Windows.AppTesting requires a Windows target framework (e.g. net8.0-windows). Current TargetFramework: '$(TargetFramework)'." />
```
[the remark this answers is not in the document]
Evangelink: Fixed. Added `'$(TargetFramework)' != ''` guard to the validation target condition, so the error only fires during inner builds where `TargetFramework` is set. Cross-targeting outer builds (where `TargetFramework` is empty) are now skipped.

src/TestFramework/TestFramework.Windows.AppTesting/WindowTest.cs
```diff
@@ -0,0 +1,61 @@
...
+///
+///         var result = MainWindow.FindFirst(TreeScope.Descendants,
+///             new PropertyCondition(AutomationElement.AutomationIdProperty, "ResultText"));
+///         Assert.IsNotNull(result);
+///     }
+/// }
+/// </code>
+/// <para>
+/// For a richer element interaction API, add a library like FlaUI on top and use
+/// <see cref="MainWindow"/> handle to bridge into it.
```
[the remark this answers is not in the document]
Evangelink: Fixed. Updated the remarks to say "the `MainWindow` `AutomationElement`" instead of "handle" — it's an `AutomationElement`, not an `IntPtr`.

test/IntegrationTests/MSTest.Acceptance.IntegrationTests/WindowsAppTestingSdkTests.cs
```diff
@@ -0,0 +1,108 @@
...
+        DotnetMuxerResult dotnetTestResult = await DotnetCli.RunAsync(
+            $"test {testHost.FullName}",
+            workingDirectory: AssetFixture.ProjectPath,
+            failIfReturnValueIsNotZero: false,
+            warnAsError: false,
+            suppressPreviewDotNetMessage: false,
+            cancellationToken: TestContext.CancellationToken);
+
+        dotnetTestResult.AssertExitCodeIs(0);
+        dotnetTestResult.AssertOutputContains("Passed!  - Failed:     0, Passed:     2, Skipped:     0, Total:     2");
```
Evangelink, line 49: **[Assertion]** The VSTest test doesn't verify that VSTest was actually invoked. Both `AspireSdkTests` and `PlaywrightSdkTests` assert `dotnetTestResult.AssertOutputContains("VSTest version")` to confirm the run went through VSTest rather than falling back to the MTP runner.

**Impact**: False confidence — if `dotnet test` routes through the MTP runner due to a misconfiguration, this test still passes, making the "WhenUsingVSTest" claim hollow.

**Suggestion**: Add `dotnetTestResult.AssertOutputContains("VSTest version");` before or after line 41:
```csharp
dotnetTestResult.AssertExitCodeIs(0);
dotnetTestResult.AssertOutputContains("VSTest version");
dotnetTestResult.AssertOutputContains("Passed!  - Failed:     0, Passed:     2, Skipped:     0, Total:     2");
```

samples/public/DemoMSTestSdk/ProjectUsingWindowsAppTesting/CalculatorTests.cs
```diff
@@ -0,0 +1,61 @@
...
+    /// Override this to point to your own application executable.
+    /// </summary>
+    public override string ApplicationPath => "calc.exe";
+
+    [TestMethod]
+    public void Calculator_MainWindow_HasExpectedTitle()
+    {
+        // The MainWindow property is automatically populated by WindowTest
+        // after launching the application specified by ApplicationPath.
+        Assert.IsNotNull(MainWindow);
```
Evangelink: **[Assertion]** `Assert.IsNotNull(MainWindow)` is vacuous here. `WindowSetup()` in the base class already throws `InvalidOperationException` if `AutomationElement.FromHandle` returns null, so by the time this test method runs, `MainWindow` is guaranteed non-null by the framework lifecycle. This assertion can never catch a real bug.

**Impact**: Misleads users copying this sample into thinking they need/can test for a null `MainWindow`.

**Suggestion**: Remove the vacuous assertion and rely on the non-empty title check below it, or replace it with a meaningful structural check — e.g., mirroring the acceptance test's embedded `NotepadTests`:
```csharp
Assert.AreEqual(ControlType.Window, MainWindow.Current.ControlType,
    "Expected the main window element to be of control type Window.");
```

Evangelink (commented):
### Summary

**Date**: 2026-05-12

### Key Findings

The PR has addressed all issues from the two previous test reviews ✅:
- `AssertOutputContainsSummary(0, 2, 0)` is now present in the MTP runner test
- Vacuous `Assert.IsNotNull(MainWindow)` was replaced with the meaningful `ControlType.Window` assertion in `NotepadTests`
- `Calculator_NumberButtons_AreVisible` now carries `[Ignore]` with an explanatory message

Two minor issues remain:

1. **[Assertion] `EnableWindowsAppTesting_WhenUsingVSTest_RunsDesktopTests` does not assert `"VSTest version"`** (`WindowsAppTestingSdkTests.cs` line 42) — Both `AspireSdkTests` and `PlaywrightSdkTests` include `dotnetTestResult.AssertOutputContains("VSTest version")` to confirm VSTest was actually invoked. Without it, the test passes even if the run silently fell back to the MTP runner, making the `WhenUsingVSTest` claim unverifiable.

2. **[Assertion] `Assert.IsNotNull(MainWindow)` in the sample `CalculatorTests.cs` is vacuous** (line 35) — `WindowSetup()` already throws before the test body if `MainWindow` cannot be obtained, so this assertion can never catch a real failure. The same fix applied to the embedded `NotepadTests` (using `Assert.AreEqual(ControlType.Window, ...)`) should be applied here for consistency.

### Recommendations

- Add `dotnetTestResult.AssertOutputContains("VSTest version");` in `EnableWindowsAppTesting_WhenUsingVSTest_RunsDesktopTests`.
- Replace `Assert.IsNotNull(MainWindow)` in `CalculatorTests.Calculator_MainWindow_HasExpectedTitle` with `Assert.AreEqual(ControlType.Window, MainWindow.Current.ControlType, ...)`, consistent with the acceptance test.

---

> 🧪 *Test quality reviewed by [Test Expert Reviewer 🧪](https://github.com/microsoft/testfx/actions/runs/25726445947)*

> 🧪 *Test quality reviewed by [Test Expert Reviewer 🧪](https://github.com/microsoft/testfx/actions/runs/25726445947)*

src/TestFramework/TestFramework.Windows.AppTesting/ApplicationTest.cs
```diff
@@ -0,0 +1,133 @@
...
+            TimeSpan remainingTime = ApplicationStartTimeout - sw.Elapsed;
+            if (remainingTime > TimeSpan.Zero)
+            {
+                Thread.Sleep(remainingTime < TimeSpan.FromMilliseconds(50)
+                    ? remainingTime
+                    : TimeSpan.FromMilliseconds(50));
+            }
+        }
+
+        if (AppProcess.MainWindowHandle == IntPtr.Zero)
```
Evangelink, line 94: **[Correctness — Minor]** The `MainWindowHandle` value checked here is the cached value from the last `AppProcess.Refresh()` call, which happened at the **top** of the loop body—before the final `Thread.Sleep`. If the application's main window becomes available during that last ≤50 ms sleep (e.g., slow machine near the timeout boundary), the stale cache will still show `IntPtr.Zero` and a `TimeoutException` will be thrown even though the window is ready.

**Suggestion:** Add one final `Refresh()` before this post-loop check:
```csharp
AppProcess.Refresh(); // ensure latest state after final sleep interval
if (AppProcess.MainWindowHandle == IntPtr.Zero)
{
    throw new TimeoutException(...);
}
```

Evangelink (commented):
### Summary

**Workflow**: Expert Code Reviewer 🧠
**Date**: 2026-05-12

### Key Findings

The four issues from the first expert review (run [25696190614](https://github.com/microsoft/testfx/actions/runs/25696190614)) are all correctly addressed in the current HEAD. Three new minor observations remain:

#### 🟡 Minor — Correctness
**`AppProcess.Refresh()` missing before final timeout check** (`ApplicationTest.cs:93`)
The cached `MainWindowHandle` value is up to one poll interval (≤50 ms) stale when the while loop exits due to timeout. A final `Refresh()` before the post-loop check prevents a false `TimeoutException` at the boundary.

#### 🟡 Minor — Resource management
**`Win32Exception` not caught in `ApplicationTearDown`** (`ApplicationTest.cs:121`)
`CloseMainWindow()` and `Kill()` are both documented to throw `Win32Exception` (e.g., different desktop session, insufficient permissions). The current `catch (InvalidOperationException)` misses these, causing the process to run orphaned when teardown is triggered against an elevated or session-isolated process.

#### 🔵 Informational — Cross-TFM dependency
**`UseWPF=true` forces WPF as transitive dependency** (`TestFramework.Windows.AppTesting.csproj:6`)
NuGet propagates framework references, so all consumers (Win32 or WinForms test projects) silently acquire WPF. The package only needs UIAutomation assemblies, which are available without WPF via `<FrameworkReference Include="Microsoft.WindowsDesktop.App" />`.

### Positive Observations

- **Process teardown** is robust: `[TestCleanup]` runs even when `[TestInitialize]` throws, the `finally` block always disposes the .NET handle, and `Kill(entireProcessTree: true)` correctly handles child processes.
- **`[STATestClass]` placement** is correct: `ApplicationTest` and `WindowTest` both carry it; `AutomationTest` intentionally uses `[TestClass]` as the extension-point root. STA is inherited by derived user classes.
- **Startup polling loop** correctly calls `Refresh()` before each `HasExited` and `MainWindowHandle` check; the `Thread.Sleep(min(remaining, 50ms))` avoids overshooting the timeout significantly.
- **Null safety** in `WindowSetup` — the null-coalescing throw on `AutomationElement.FromHandle` gives an actionable error message.
- **SDK wiring** mirrors the Playwright/Aspire patterns precisely: feature targets, props template, NativeAOT guard, and the build-time TFM error are all consistent with the existing infrastructure.

---

*Generated by [Expert Code Reviewer 🧠](https://github.com/microsoft/testfx/actions/runs/25726445937)*

> 🧠 *Reviewed by [Expert Code Reviewer 🧠](https://github.com/microsoft/testfx/actions/runs/25726445937)*

src/TestFramework/TestFramework.Windows.AppTesting/ApplicationTest.cs
```diff
@@ -0,0 +1,133 @@
...
+                try
+                {
+                    _ = appProcess.CloseMainWindow();
+                    if (!appProcess.WaitForExit(5000))
+                    {
+                        appProcess.Kill(entireProcessTree: true);
+                        _ = appProcess.WaitForExit(5000);
+                    }
+                }
+                catch (InvalidOperationException)
```
Evangelink: **[Resources — Minor]** The inner `catch (InvalidOperationException)` does not cover `Win32Exception`, which both `CloseMainWindow()` and `Kill()` are documented to throw — for example, when `PostMessage` fails (process in a different desktop/session) or when the test runner lacks permission to kill an elevated process.

If `CloseMainWindow()` throws a `Win32Exception`, execution leaves the inner try without ever calling `Kill()`. The `finally` block then disposes the .NET `Process` handle, but the OS process keeps running — silently orphaned across tests.

**Suggestion:**
```csharp
catch (Exception ex) when (ex is InvalidOperationException or Win32Exception)
{
    // Process exited or became inaccessible between state checks.
}
```
Or alternatively, move `Kill()` into the `finally` block with a best-effort guard so it always runs when cleanup is warranted.

src/TestFramework/TestFramework.Windows.AppTesting/TestFramework.Windows.AppTesting.csproj
```diff
@@ -0,0 +1,38 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFrameworks>net8.0-windows;net9.0-windows</TargetFrameworks>
+    <EnableWindowsTargeting Condition=" '$(OS)' != 'Windows_NT' ">true</EnableWindowsTargeting>
+    <UseWPF>true</UseWPF>
```
Evangelink: **[Cross-TFM — Minor]** `UseWPF=true` brings in the full WPF framework reference (`Microsoft.WindowsDesktop.App.WPF`). Because NuGet propagates framework references transitively, consumers of this package — including Win32 or WinForms-only test projects — will silently acquire WPF as a compile-time and runtime dependency. This can force WPF's static initialisation, increase binary size, and occasionally conflict with `UseWindowsForms=true` project settings.

The package only needs `UIAutomationClient.dll` and `UIAutomationTypes.dll` — both are available in the Windows Desktop runtime without WPF. Consider replacing `UseWPF` with explicit framework references:

```xml
<ItemGroup>
  <FrameworkReference Include="Microsoft.WindowsDesktop.App" />
</ItemGroup>
```
or directly reference just the UIAutomation assemblies to keep the WPF dependency opt-in.

src/TestFramework/TestFramework.Windows.AppTesting/ApplicationTest.cs
Evangelink: Fixed. Added `AppProcess.Refresh()` before the post-loop `MainWindowHandle` check so the latest state is used after the final sleep interval.

src/TestFramework/TestFramework.Windows.AppTesting/ApplicationTest.cs
Evangelink: Fixed. Widened the catch to also cover `Win32Exception`:
```csharp
catch (Exception ex) when (ex is InvalidOperationException or Win32Exception)
{
    // The process exited or became inaccessible between state checks and shutdown operations.
}
```

samples/public/DemoMSTestSdk/ProjectUsingWindowsAppTesting/CalculatorTests.cs
Evangelink: Fixed. Replaced the vacuous `Assert.IsNotNull(MainWindow)` with a meaningful structural check:
```csharp
Assert.AreEqual(ControlType.Window, MainWindow.Current.ControlType,
    "Expected the main window element to be of control type Window.");
```

src/TestFramework/TestFramework.Windows.AppTesting/TestFramework.Windows.AppTesting.csproj
Evangelink: Fixed. Replaced `<UseWPF>true</UseWPF>` with an explicit `<FrameworkReference Include="Microsoft.WindowsDesktop.App" />` so consumers don't transitively acquire the full WPF framework reference. The UIAutomation assemblies are available through the base Windows Desktop runtime.

test/IntegrationTests/MSTest.Acceptance.IntegrationTests/WindowsAppTestingSdkTests.cs
Evangelink: Fixed. Added `dotnetTestResult.AssertOutputContains("VSTest version");` to confirm the run went through VSTest rather than the MTP runner.

test/IntegrationTests/MSTest.Acceptance.IntegrationTests/WindowsAppTestingSdkTests.cs
```diff
@@ -0,0 +1,109 @@
...
+using Microsoft.Testing.Platform.Acceptance.IntegrationTests;
+using Microsoft.Testing.Platform.Acceptance.IntegrationTests.Helpers;
+using Microsoft.Testing.Platform.Helpers;
+
+namespace MSTest.Acceptance.IntegrationTests;
+
+[TestClass]
+public sealed class WindowsAppTestingSdkTests : AcceptanceTestBase<WindowsAppTestingSdkTests.TestAssetFixture>
+{
+    private static readonly string DesktopTargetFramework = $"{TargetFrameworks.NetCurrent}-windows";
```
Evangelink: **[Coverage]** Only `NetCurrent-windows` (`net10.0-windows`) is tested. `PlaywrightSdkTests` — the closest equivalent in this repo — uses `[DynamicData(nameof(TargetFrameworks.AllForDynamicData), typeof(TargetFrameworks))]` and `.PatchTargetFrameworks(TargetFrameworks.All)` to cover every supported TFM. A regression on `net8.0-windows` would not be caught here.

**Impact**: If the library ships support for `net8.0-windows` and a future change breaks it, the acceptance test suite passes silently.

**Suggestion**: Define Windows TFMs by mapping `TargetFrameworks.Net` and drive both tests with `[DynamicData]`:
```csharp
// In the test class
private static readonly string[] s_desktopTargetFrameworks =
    TargetFrameworks.Net.Select(tfm => $"{tfm}-windows").ToArray();

public static IEnumerable<object[]> DesktopTargetFrameworksForDynamicData =>
    s_desktopTargetFrameworks.Select(tfm => new object[] { tfm });

[TestMethod]
[DynamicData(nameof(DesktopTargetFrameworksForDynamicData))]
[OSCondition(OperatingSystems.Windows, IgnoreMessage = "Windows app testing is Windows-only")]
public async Task EnableWindowsAppTesting_WhenUsingMSTestRunner_RunsDesktopTests(string tfm)
{
    var testHost = TestHost.LocateFrom(AssetFixture.ProjectPath, TestAssetFixture.ProjectName, tfm);
    ...
}
```
And update `GetAssetsToGenerate` to call `.PatchTargetFrameworks(TargetFrameworks.Net.Select(t => $"{t}-windows").ToArray())` (or a similar helper).

Evangelink (commented):
### Summary

**Date**: 2026-05-12

### Key Findings

1. **✅ [Coverage] MTP runner test count assertion — FIXED** (`WindowsAppTestingSdkTests.cs:24`) — The previously missing `AssertOutputContainsSummary(failed: 0, passed: 2, skipped: 0)` call is now present. Both the MTP and VSTest tests correctly validate test counts, consistent with `AspireSdkTests` and `PlaywrightSdkTests`.

2. **[Coverage] Only `NetCurrent-windows` TFM tested** (`WindowsAppTestingSdkTests.cs:13`) — `PlaywrightSdkTests` (the closest equivalent) uses `[DynamicData(nameof(TargetFrameworks.AllForDynamicData))]` across all TFMs. The Windows AppTesting tests only exercise `net10.0-windows`. If the package supports `net8.0-windows`, regressions on that TFM are invisible to CI. See inline comment for a concrete fix using `TargetFrameworks.Net` mapped to Windows TFMs.

### Observations

- All 25 review threads from the previous cycle are resolved or outdated — no open items carried over.
- Test isolation is clean: `ApplicationSetup`/`ApplicationTearDown` bracketing is correct. The `Thread.Sleep` in the polling loop is bounded to ≤50 ms per iteration and protected by `ApplicationStartTimeout` — acceptable for a UI automation framework.
- The embedded `NotepadTests.cs` assertions (`ControlType.Window`, non-empty title) are adequate for an acceptance test verifying the library plumbing. No false-positive risk.
- `[OSCondition(OperatingSystems.Windows)]` correctly guards both test methods, so non-Windows CI will report skips rather than failures.

### Recommendations

- Extend TFM coverage to all `TargetFrameworks.Net` entries suffixed with `-windows` (see inline comment on line 13).

---

*Generated by [Test Expert Reviewer 🧪](https://github.com/microsoft/testfx/actions/runs/25728022265)*

> 🧪 *Test quality reviewed by [Test Expert Reviewer 🧪](https://github.com/microsoft/testfx/actions/runs/25728022265)*

src/TestFramework/TestFramework.Windows.AppTesting/TestFramework.Windows.AppTesting.csproj
```diff
@@ -0,0 +1,41 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFrameworks>net8.0-windows;net9.0-windows</TargetFrameworks>
+    <EnableWindowsTargeting Condition=" '$(OS)' != 'Windows_NT' ">true</EnableWindowsTargeting>
+  </PropertyGroup>
+
+  <ItemGroup>
```
Evangelink, line 8: **[API Compat / Dependencies]** `Microsoft.WindowsDesktop.App` is overly broad — it includes both WPF **and** WinForms assemblies, but this package only uses `System.Windows.Automation` which lives in the WPF subsystem.

**Impact**: Consumers of `MSTest.Windows.AppTesting` will transitively pull in WinForms assemblies they don't need. A user testing a pure Win32 or WPF app ends up with both WPF and WinForms in their test binary. This is a regression vs. the previous `UseWPF=true` approach, which brought in only WPF.

**Suggestion**: Use the more precise framework reference:
```xml
<FrameworkReference Include="Microsoft.WindowsDesktop.App.WPF" />
```
`UIAutomationClient.dll` and `UIAutomationTypes.dll` (which back `System.Windows.Automation`) are part of the WPF shared framework, so this is the minimal correct reference.

Evangelink (commented):
### Summary

**Workflow**: Expert Code Reviewer 🧠
**Date**: 2026-05-12

### Key Findings

The third round of fixes (commit `c9ac13f`) resolved all previously open correctness issues:
- ✅ `AppProcess.Refresh()` now called before the post-timeout `MainWindowHandle` check
- ✅ `Win32Exception` now caught in `ApplicationTearDown` alongside `InvalidOperationException`
- ✅ `UseWPF=true` replaced with an explicit `FrameworkReference`

#### 🟡 One remaining concern

**`FrameworkReference Include="Microsoft.WindowsDesktop.App"`** (csproj line 8) — the switch from `UseWPF=true` replaced one WPF-only reference with a broader reference that includes both WPF and WinForms. Since `System.Windows.Automation` resides in the WPF subsystem, `Microsoft.WindowsDesktop.App.WPF` is the correct minimal reference. See inline comment for details.

#### ✅ What was verified

- **Lifecycle ordering**: `ApplicationSetup` → `WindowSetup` (base-first init), single `ApplicationTearDown` cleanup — correct MSTest ordering
- **Process teardown**: `CloseMainWindow` → `WaitForExit(5s)` → `Kill(entireProcessTree)` → `Dispose` in try/finally — no leaks in any error path
- **`Win32Exception` handling**: correctly catches post-HasExited races in teardown
- **Post-loop `Refresh()` guard**: main window handle is freshly validated before `WindowSetup` consumes it
- **SDK integration**: NativeAOT guard, feature-flag pattern, nuspec wiring, version templating — all correct
- **Public API surface**: `PublicAPI.Unshipped.txt` is complete and accurate

### Recommendations

Change `<FrameworkReference Include="Microsoft.WindowsDesktop.App" />` to `<FrameworkReference Include="Microsoft.WindowsDesktop.App.WPF" />` to avoid pulling WinForms assemblies into consumers that don't need them.

---

*Generated by [Expert Code Reviewer 🧠](https://github.com/microsoft/testfx/actions/runs/25728022272)*

> 🧠 *Reviewed by [Expert Code Reviewer 🧠](https://github.com/microsoft/testfx/actions/runs/25728022272)*

src/TestFramework/TestFramework.Windows.AppTesting/TestFramework.Windows.AppTesting.csproj
Evangelink: Good point — narrowed the framework reference to `Microsoft.WindowsDesktop.App.WPF` since `UIAutomationClient.dll` and `UIAutomationTypes.dll` are part of the WPF shared framework. This avoids pulling in WinForms assemblies unnecessarily.

samples/public/DemoMSTestSdk/ProjectUsingWindowsAppTesting/ProjectUsingWindowsAppTesting.csproj
```diff
@@ -0,0 +1,43 @@
+<Project Sdk="MSTest.Sdk">
+
+  <PropertyGroup>
+    <TargetFramework>net8.0-windows</TargetFramework>
+    <ImplicitUsings>enable</ImplicitUsings>
+    <Nullable>enable</Nullable>
+    <!-- Enables the MSTest Windows App Testing feature -->
+    <EnableWindowsAppTesting>true</EnableWindowsAppTesting>
+  </PropertyGroup>
+
```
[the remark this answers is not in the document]
Evangelink: The `samples/public/global.json` SDK version is bumped as part of the release process (same pattern as the Playwright sample). The sample will build once 4.3.0 is published to NuGet. No change needed here.

test/IntegrationTests/MSTest.Acceptance.IntegrationTests/WindowsAppTestingSdkTests.cs
Evangelink: Good catch! Updated both test methods to use `[DynamicData]` driven by `TargetFrameworks.Net` mapped to `-windows` suffixes, and switched `GetAssetsToGenerate` to use `.PatchTargetFrameworks(DesktopTargetFrameworks)`. This now covers all supported .NET TFMs (net8.0-windows, net10.0-windows).

Evangelink (commented):
### Summary

**Workflow**: Expert Code Reviewer 🧠
**Date**: 2026-05-12

### Key Findings

All open findings from previous expert reviews are resolved in the current HEAD (`76e6f0b`):

| Previous finding | Status |
|---|---|
| `AppProcess.Refresh()` missing before final timeout check | ✅ Fixed — `Refresh()` now called unconditionally before the `TimeoutException` check |
| `Win32Exception` not caught in `ApplicationTearDown` | ✅ Fixed — `catch (Exception ex) when (ex is InvalidOperationException or Win32Exception)` |
| `FrameworkReference` overly broad (`Microsoft.WindowsDesktop.App`) | ✅ Fixed — now scoped to `Microsoft.WindowsDesktop.App.WPF`, which is the minimum required for `System.Windows.Automation` |

No new correctness, threading, performance, API-compat, cross-TFM, resource, or security issues were found in this pass.

### Positive Observations

- **Startup polling loop** is correct: each iteration calls `Refresh()` and `HasExited` before sleeping, avoiding a busy-spin; the trailing `Refresh()` + handle check after the loop exits is clean.
- **Teardown** is robust: `try/finally` guarantees `Dispose()` even if `CloseMainWindow()`/`Kill()` throw; `Kill(entireProcessTree: true)` handles child processes; `WaitForExit(5000)` post-kill avoids zombie handles.
- **`[STATestClass]`** placement is correct: `ApplicationTest` and `WindowTest` carry it, and since `Inherited = true` (the default for `AttributeUsageAttribute`), all user-derived classes automatically run on the STA thread without needing to redeclare the attribute — while `AutomationTest` deliberately stays `[TestClass]` as the extension-point root.
- **`WindowSetup` ordering**: `[TestInitialize]` methods run base-to-derived, so `ApplicationSetup` (which sets `AppProcess`) always precedes `WindowSetup` (which reads `AppProcess.MainWindowHandle`). If `ApplicationSetup` throws, `WindowSetup` is skipped by the MSTest runner, so no NRE path exists.
- **PublicAPI.Unshipped.txt** correctly declares the full public surface for the new package.
- **SDK integration** mirrors the Playwright/Aspire pattern exactly: feature targets, `NativeAOT.targets` guard, `Sdk.props.template` version injection, Central Package Management support.

### Overall Assessment

The PR is **ready to merge**. The implementation is clean, safe, and consistent with the existing thin-layer pattern used by Playwright and Aspire.

---

*Generated by [Expert Code Reviewer 🧠](https://github.com/microsoft/testfx/actions/runs/25730117723)*

> 🧠 *Reviewed by [Expert Code Reviewer 🧠](https://github.com/microsoft/testfx/actions/runs/25730117723)*

Evangelink (commented):
### Summary

**Date**: 2026-05-12

### Key Findings

All issues from previous test reviews have been addressed ✅, and the latest commit (`76e6f0b`) adds clean multi-TFM coverage:

- `AssertOutputContainsSummary(0, 2, 0)` is present in the MTP runner test ✓
- `AssertOutputContains("VSTest version")` confirms VSTest invocation in the VSTest test ✓
- `AssertOutputContains("Passed! - Failed: 0, Passed: 2, ...")` verifies test counts in VSTest mode ✓
- `[DynamicData(nameof(DesktopTargetFrameworksForDynamicData))]` now runs both tests across all supported Windows TFMs (consistent with `PlaywrightSdkTests` approach for platform-specific TFMs) ✓
- `PatchTargetFrameworks(DesktopTargetFrameworks)` correctly builds the asset for all Windows TFMs in a single pass ✓
- `TestAssetFixture` pattern matches the repo's Aspire/Playwright convention — asset built once, reused by all test cases ✓
- Assertions in the embedded `NotepadTests.cs` and `CalculatorTests.cs` sample are meaningful (`ControlType.Window` check, not vacuous `IsNotNull`) ✓

### Recommendations

No further changes needed. Tests are well-isolated, structurally consistent with existing acceptance tests, and cover the expected behavior across all Windows TFMs.

---

*Generated by [Test Expert Reviewer 🧪](https://github.com/microsoft/testfx/actions/runs/25730117718)*

> 🧪 *Test quality reviewed by [Test Expert Reviewer 🧪](https://github.com/microsoft/testfx/actions/runs/25730117718)*

> 🧪 *Test quality reviewed by [Test Expert Reviewer 🧪](https://github.com/microsoft/testfx/actions/runs/25730117718)*

src/TestFramework/TestFramework.Windows.AppTesting/WindowTest.cs
```diff
@@ -0,0 +1,61 @@
...
+/// }
+/// </code>
+/// <para>
+/// For a richer element interaction API, add a library like FlaUI on top and use
+/// the <see cref="MainWindow"/> <see cref="AutomationElement"/> to bridge into it.
+/// </para>
+/// </remarks>
+[STATestClass]
+public class WindowTest : ApplicationTest
+{
```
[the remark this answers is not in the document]
Evangelink: `WindowTest` already carries `[STATestClass]` on the class itself (line 42), so derived classes inherit STA behavior automatically — users don't need to add `[STATestClass]` themselves. The XML doc code sample and the integration test asset both show `[STATestClass]` as the recommended pattern for clarity, but it's not strictly required when inheriting from `WindowTest`.

src/TestFramework/TestFramework.Windows.AppTesting/TestFramework.Windows.AppTesting.csproj
```diff
@@ -0,0 +1,41 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFrameworks>net8.0-windows;net9.0-windows</TargetFrameworks>
+    <EnableWindowsTargeting Condition=" '$(OS)' != 'Windows_NT' ">true</EnableWindowsTargeting>
+  </PropertyGroup>
```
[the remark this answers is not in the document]
Evangelink: Added the project to `TestFx.slnx` so it's included in the main Windows build/pack.

Evangelink (commented):
### Summary

**Date**: 2026-05-12

### Key Findings

Both issues flagged in the previous test review (2026-05-12T09:55) have been resolved ✅:

1. **[Assertion] VSTest version assertion** — `dotnetTestResult.AssertOutputContains("VSTest version")` is now present in `EnableWindowsAppTesting_WhenUsingVSTest_RunsDesktopTests` (line 44), confirming VSTest was actually invoked.

2. **[Assertion] Vacuous `Assert.IsNotNull(MainWindow)` in sample** — `CalculatorTests.Calculator_MainWindow_HasExpectedTitle` now uses `Assert.AreEqual(ControlType.Window, MainWindow.Current.ControlType, ...)`, matching the meaningful assertion style established in the embedded `NotepadTests`.

The test suite is well-structured:
- Both MTP runner and VSTest invocation paths are covered
- Each test asserts exit code **and** explicit pass/fail/skip counts
- `OSCondition` correctly guards Windows-only tests for non-Windows CI
- `[STATestClass]` threading requirement is respected throughout
- `[Ignore]` with an explanatory message covers the AutomationId-specific sample test

### Recommendations

None — all previously flagged issues have been addressed.

---

> 🧪 *Test quality reviewed by [Test Expert Reviewer 🧪](https://github.com/microsoft/testfx/actions/runs/25732804904)*

> 🧪 *Test quality reviewed by [Test Expert Reviewer 🧪](https://github.com/microsoft/testfx/actions/runs/25732804904)*

Evangelink (commented):
### Summary

**Workflow**: Expert Code Reviewer 🧠
**Date**: 2026-05-12

### Key Findings

All three observations from the previous expert review (run [25732804835](https://github.com/microsoft/testfx/actions/runs/25732804835)) are addressed in commit `76e6f0b`:

| Previous Finding | Status |
|---|---|
| `AppProcess.Refresh()` missing before final timeout check (`ApplicationTest.cs:93`) | ✅ `Refresh()` call is present at line 94 |
| `Win32Exception` not caught in `ApplicationTearDown` | ✅ `catch (Exception ex) when (ex is InvalidOperationException or Win32Exception)` on line 123 |
| `UseWPF=true` forces WPF as transitive dependency | ✅ Narrowed to `<FrameworkReference Include="Microsoft.WindowsDesktop.App.WPF" />` — correct minimal reference since `UIAutomationClient.dll` lives in the WPF framework subset |

### Fresh Analysis (No New Issues Found)

After reviewing all changed files in the current HEAD (`66d35e0`):

- **Correctness**: Lifecycle ordering is correct — MSTest calls base-class `[TestInitialize]` first, so `AppProcess` is always set before `WindowSetup` accesses it; `[TestCleanup]` on `ApplicationTest` runs even when `[TestInitialize]` throws, so process handles are always cleaned up.
- **Threading**: `[STATestClass]` is correctly applied to both `ApplicationTest` and `WindowTest`; the sample and acceptance test both redeclare `[STATestClass]` on the user test class (required because `STATestClassAttribute` has `Inherited = false`).
- **Resources**: `try/finally` in `ApplicationTearDown` guarantees disposal; `Kill(entireProcessTree: true)` handles child processes; `WaitForExit(5000)` after kill prevents zombie handles.
- **API surface**: `PublicAPI.Unshipped.txt` is complete and correct; all public types have XML docs; no `init` accessors in new public API.
- **Cross-TFM**: `_ValidateWindowsAppTestingPlatform` target errors at Restore/Build time for non-Windows TFMs; `EnableWindowsTargeting` guard in the `.csproj` handles non-Windows CI builds.
- **Acceptance tests**: Both MTP runner and VSTest paths now assert `AssertOutputContainsSummary(failed: 0, passed: 2, skipped: 0)` / `"Passed:     2"`, consistent with `AspireSdkTests` and `PlaywrightSdkTests`; `[DynamicData]` loops over all desktop TFMs.

### Overall Assessment

The implementation is clean, correct, and complete. The thin-layer pattern faithfully mirrors Playwright/Aspire, the teardown logic is race-safe, and all previously identified issues have been resolved. No correctness, threading, performance, API compatibility, resource management, or security concerns remain.

---

*Generated by [Expert Code Reviewer 🧠](https://github.com/microsoft/testfx/actions/runs/25732804835)*

> 🧠 *Reviewed by [Expert Code Reviewer 🧠](https://github.com/microsoft/testfx/actions/runs/25732804835)*

PR #7818, opened 2026-04-24
[Perf Improver] perf: reduce allocations in test discovery and data-driven test execution

Evangelink:
…tion

- ValidSourceExtensions in TestSourceHandler now returns a cached static readonly collection instead of allocating a new List<string> on every property access. This eliminates unnecessary heap allocations during test source validation.

- ExecuteTestWithDataSourceAsync in TestMethodRunner no longer creates two ReflectionTestMethodInfo instances when testDataSource is not null and GetDisplayName returns null (requiring fallback to ComputeDefaultDisplayName). The instance is now created once and reused, halving the allocations for this code path in data-driven tests.


nohwnd:
lgmt, measurements would be nice. :)

nohwnd (approved):

PR #7862, opened 2026-04-27
Throw PlatformNotSupportedException in STATestMethodAttribute on non-Windows

Evangelink:
## Summary

`STATestMethodAttribute.ExecuteAsync` had an ambiguous `// TODO: Throw?` comment on the non-Windows code path (line 79). The test would silently run without STA context on non-Windows, with no indication to the user.

## Change

Replace the `// TODO: Throw?` with a `PlatformNotSupportedException`. STA (Single Threaded Apartment) is a Windows-only COM threading concept — silently running without it is misleading. Failing explicitly makes the contract clear.

The error message guides users to use `[OSCondition(OperatingSystems.Windows)]` to skip the test on non-Windows platforms.

## Changes

- **`STATestMethodAttribute.cs`**: Throw `PlatformNotSupportedException` on non-Windows instead of silently executing without STA.
- **`FrameworkMessages.resx`**: Added `STATestMethodNonWindowsNotSupported` resource string.
- **`xlf/` files**: Auto-generated by build.

### DIFF
--- a/src/TestFramework/TestFramework/Attributes/TestMethod/STATestMethodAttribute.cs
+++ b/src/TestFramework/TestFramework/Attributes/TestMethod/STATestMethodAttribute.cs
@@ -73,14 +73,12 @@ public class STATestMethodAttribute : TestMethodAttribute
             return await ExecuteCoreAsync(testMethod).ConfigureAwait(false);
         }
 
-#if NETFRAMEWORK
-        if (Environment.OSVersion.Platform != PlatformID.Win32NT)
-#else
+#if !NETFRAMEWORK
         if (!RuntimeInformation.IsOSPlatform(OSPlatform.Windows))
-#endif
         {
             throw new PlatformNotSupportedException(FrameworkMessages.STATestMethodNonWindowsNotSupported);
         }
+#endif
 
         TestResult[]? results = null;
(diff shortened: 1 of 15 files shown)


Evangelink:
Sorry I asked to create PR in the wrong chat... will do the changes I wanted here

src/TestFramework/TestFramework/Attributes/TestMethod/STATestMethodAttribute.cs
```diff
@@ -73,13 +73,14 @@ public override async Task<TestResult[]> ExecuteAsync(ITestMethod testMethod)
             return await ExecuteCoreAsync(testMethod).ConfigureAwait(false);
         }
 
-#if !NETFRAMEWORK
+#if NETFRAMEWORK
+        if (Environment.OSVersion.Platform != PlatformID.Win32NT)
```
Youssef1313: I'm not sure about the correctness of this check. The enum, for example, has `Win32Windows` (which is Windows 95 or Windows 98 - yes extremely old and unlikely anyone using that today 😄)

We shouldn't really care about Mono, and I would go simple and just introduce the check under `#if !NETFRAMEWORK`. Let's not bother with anything Mono.

Evangelink:
@copilot please revert last commit

Evangelink:
@Youssef1313 PR is ready

nohwnd (approved):

PR #7865, opened 2026-04-27
Fix command line parser splitting quoted paths with spaces

### LINKED ISSUE #2891: Debugging tests fail when path contains spaces using MTP
## Describe the Issue
Running tests from VS Code with C# Dev Kit fails to start when diagnostic logging is enabled and the diagnostic output directory contains spaces.

## Steps to Reproduce
1. Create a new MSTest project in a folder whose path contains spaces: `/home/<user>/src/Example Solution`
2. Ensure the test project uses MTP.
3. Open the folder in a fresh VS Code window and wait for C# Dev Kit to discover tests.
4. In Test Explorer, choose Debug on any test.

## Expected Behaviour
The test process starts and logs diagnostics to the correct directory.

## Actual Behaviour
The test host exits immediately. Output shows the diagnostic path split at spaces.

```console
Diagnostic file (level 'Information' with async flush): /home/<user>/src/Example Solution/artifacts/build/bin/Example.Tests/debug/"/home/<user>/src/Example/log_260317001604517.diag
Option '--diagnostic-output-directory' from provider 'Platform command line provider' (UID: PlatformCommandLineProvider) expects at most 1 arguments
```

## Evidence
With VS Code C# debug log tracing enabled, debug console shows the path quoted but escaped, so the directory is not treated as a single token.

```console
[Start Debugging] CreateProcess((null), "/home/<user>/src/Example Solution/artifacts/build/bin/Example.Tests/debug/Example.Tests"  --server --diagnostic --diagnostic-verbosity Information --diagnostic-output-directory \"/home/<user>/src/Example Solution/artifacts/build/bin/Example.Tests/debug/Log\" --client-port 43653 , inheritHandles:0, launchFlags:0x4, currentDirectory:/home/<user>/src/Example Solution/artifacts/build/bin/Example.Tests...
```

Extension version: 2.10.3
VS Code version: Code 1.111.0 (ce099c1ed25d9eb3076c11e4a280f3eb52b4fbeb, 2026-03-06T23:06:10Z)
OS version: Windows_NT x64 10.0.26200
Modes:
Remote OS version: Linux x64 6.6.87.2-microsoft-standard-WSL2

<details>
<summary>System Info</summary>

|Item|Value|
|---|---|
|CPUs|Intel(R) Core(TM) Ultra 5 125U (14 x 2688)|
|GPU Status|2d_canvas: enabled<br>GPU0: VENDOR= 0x8086, DEVICE=0x7d45 [Intel(R) Graphics], DRIVER_VENDOR=Intel, DRIVER_VERSION=32.0.101.7026 *ACTIVE*<br>GPU1: VENDOR= 0x1414, DEVICE=0x008c [Microsoft Basic Render Driver], DRIVER_VERSION=10.0.26100.7309<br>Machine model name: <br>Machine model version: <br>direct_rendering_display_compositor: disabled_off_ok<br>gpu_compositing: enabled<br>multiple_raster_threads: enabled_on<br>opengl: enabled_on<br>rasterization: enabled<br>raw_draw: disabled_off_ok<br>skia_graphite: disabled_off<br>trees_in_viz: disabled_off<br>video_decode: enabled<br>video_encode: enabled<br>webgl: enabled<br>webgl2: enabled<br>webgpu: enabled<br>webnn: disabled_off|
|Load (avg)|undefined|
|Memory (System)|31.51GB (12.98GB free)|
|Process Argv|--folder-uri=vscode-remote://wsl+Ubuntu-24.04/home/lhook/src/Cloud%20Application%20Guidelines/Canary%20API --remote=wsl+Ubuntu-24.04 --crash-reporter-id 3dadc642-0059-4492-a57b-4855c93d5ec0|
|Screen Reader|no|
|VM|0%|

|Item|Value|
|---|---|
|Remote|WSL: Ubuntu-24.04|
|OS|Linux x64 6.6.87.2-microsoft-standard-WSL2|
|CPUs|Intel(R) Core(TM) Ultra 5 125U (14 x 0)|
|Memory (System)|15.37GB (11.87GB free)|
|VM|0%|
</details><details>
<summary>A/B Experiments</summary>

```
vsliv368:30146709
binariesv615:30325510
nativeloc1:31344060
dwcopilot:31170013
471b6256:31263136
dwoutputs:31242946
copilot_t_ci:31333650
e5gg6876:31282496
pythonrdcb7:31342333
6518g693:31463988
aj953862:31281341
63221493:31336333
envsdeactivate2:31464701
cloudbuttont:31379625
aihoversummaries_f:31469309
3efgi100_wstrepl:31403338
55ia6109:31457054
use-responses-api:31390855
je187915:31401257
linter-cursor-jump:31475312
ec5jj548:31422691
cp_cls_t_966_ss:31454198
4je02754:31466945
ge8j1254_inline_auto_hint_haiku:31431912
nes-autoexp-10:31446583
00h15499_gpt_53_codex:31464543
7a04d226_do_not_restore_last_panel_session:31438103
cp_cls_c_1081:31454833
ia-use-proxy-models-svc:31452481
a43f0574a:31442826
e9c30283:31461165
test_treatment2:31471001
nes-conv-2-3:31474331
subagent_3ccgc321:31467997
idci7584:31464702
edit_mode_hidden:31461530
864ei723_large_tool_results_to_disk:31460878
showingstats:31471631
55364912:31471672
nes-extended-on:31455476
chat:31457767
8hig5102:31473806
jdddd261:31472041
nes-rcnt-ctrl:31471618
9g0ce865_5_3_codex_prompt_disabled:31475327
cpptoolson-v2:31475363
4dgh1208:31471592
editor1:31474144
thinking_c:31471065
c3257622:31473910
commenticon:31473926
nes-notsn-1:31475122

```

</details>

nohwnd:
When a path with spaces is passed to `--diagnostic-output-directory` (or any option expecting a single argument), the OS can split it across multiple `args[]` entries. The parser then sees each fragment as a separate argument, fails arity validation, and the test host exits immediately.

This happens in practice when VS Code C# Dev Kit launches MTP on Linux/WSL with paths like `/home/user/repro folder/bin/Debug/net10.0/Log`.

The fix teaches the parser to reassemble quoted argument fragments: when an arg starts with an unmatched `\"`, `'`, or `\\\"` it merges subsequent args until the closing quote. Also extends `TryUnescape` to strip escaped double-quotes (`\\"...\\"`).

Fixes https://github.com/microsoft/vscode-dotnettools/issues/2891

### DIFF
--- a/src/Platform/Microsoft.Testing.Platform/CommandLine/Parser.cs
+++ b/src/Platform/Microsoft.Testing.Platform/CommandLine/Parser.cs
@@ -78,6 +78,10 @@ internal static class CommandLineParser
 
             if (currentArg is not null)
             {
+                // When a quoted value is split across multiple args (e.g. a path with spaces),
+                // reassemble the fragments into a single argument.
+                currentArg = TryMergeQuotedArguments(args, ref i, currentArg);
+
                 if (currentOption is null)
                 {
                     errors.Add(string.Format(CultureInfo.InvariantCulture, PlatformResources.CommandLineParserUnexpectedArgument, args[i]));
@@ -114,6 +118,71 @@ internal static class CommandLineParser
             currentOption = currentOption.TrimStart('-');
         }
 
+        // When a quoted argument value is split across multiple args[] entries (e.g. a path
+        // containing spaces like "/path/my dir/log"), reassemble the fragments into a single
+        // string. Handles double-quote, single-quote, and escaped-quote (\") wrapping.
+        static string TryMergeQuotedArguments(List<string> args, ref int i, string currentArg)
+        {
+            string trimmed = currentArg.Trim();
+            if (trimmed.Length == 0)
+            {
+                return currentArg;
+            }
+
+            // Determine the opening quote style and whether it's already closed.
+            string? closingSuffix;
+            if (trimmed.StartsWith("\\\"", StringComparison.Ordinal))
+            {
+                // Escaped double-quote: \"...\"
+                if (trimmed.Length > 4 && trimmed.EndsWith("\\\"", StringComparison.Ordinal))
+                {
+                    return currentArg;
+                }
+
+                closingSuffix = "\\\"";
+            }
+            else if (trimmed[0] == '"')
+            {
+                if (trimmed.Length > 1 && trimmed[trimmed.Length - 1] == '"')
+                {
+                    return currentArg;
+                }
+
+                closingSuffix = "\"";
+            }
+            else if (trimmed[0] == '\'')
+            {
+                if (trimmed.Length > 1 && trimmed[trimmed.Length - 1] == '\'')
+                {
+                    return currentArg;
+                }
+
+                closingSuffix = "'";
+            }
+            else
+            {
+                return currentArg;
+            }
+
+            // Merge subsequent args until the closing quote is found.
+            StringBuilder merged = new(currentArg);
+            while (i + 1 < args.Count)
+            {
+                string nextArg = args[i + 1];
+                i++;
+                merged.Append(' ').Append(nextArg);
+
+                if (nextArg.TrimEnd().EndsWith(closingSuffix, StringComparison.Ordinal))
+                {
+                    return merged.ToString();
+                }
+            }
+
+            // Closing quote was never found — return what we assembled.
+            // TryUnescape / arity validation will report appropriate errors downstream.
+            return merged.ToString();
+        }
+
         static bool TryUnescape(string input, string? option, IEnvironment environment, [NotNullWhen(true)] out string? unescapedArg, [NotNullWhen(false)] out string? error)
         {
             unescapedArg = input;
@@ -132,6 +201,15 @@ internal static class CommandLineParser
                 }
 
                 unescapedArg = input[1..^1];
+                return true;
+            }
+
+            // Handle escaped double-quotes: \"...\"
+            // Some process launchers (e.g. VS Code on Linux) emit escaped quotes that arrive
+            // as literal backslash-quote characters in the args array.
+            if (input.StartsWith("\\\"", StringComparison.Ordinal) && input.EndsWith("\\\"", StringComparison.Ordinal) && input.Length > 4)
+            {
+                unescapedArg = input[2..^2];
                 return true;
             }
 
--- a/test/UnitTests/Microsoft.Testing.Platform.UnitTests/CommandLine/CommandLineTests.cs
+++ b/test/UnitTests/Microsoft.Testing.Platform.UnitTests/CommandLine/CommandLineTests.cs
@@ -198,6 +198,31 @@ public sealed class CommandLineTests
             new("option3", ["c"]),
             new("option4", ["d"]),
         }.ToArray(), []));
+
+        // Quoted path with spaces split across multiple args (double-quotes)
+        yield return (31, ["--option1", "\"path/with", "spaces/log\""], null, new CommandLineParseResultWrapper(null, new List<CommandLineParseOption> { new("option1", ["path/with spaces/log"]) }.ToArray(), []));
+
+        // Quoted path with spaces split across three args (double-quotes)
+        yield return (32, ["--option1", "\"/home/user/Example", "Solution/artifacts/build", "bin/Log\""], null, new CommandLineParseResultWrapper(null, new List<CommandLineParseOption> { new("option1", ["/home/user/Example Solution/artifacts/build bin/Log"]) }.ToArray(), []));
+
+        // Escaped double-quote path with spaces split across multiple args
+        yield return (33, ["--option1", "\\\"/path/with", "spaces/log\\\""], null, new CommandLineParseResultWrapper(null, new List<CommandLineParseOption> { new("option1", ["/path/with spaces/log"]) }.ToArray(), []));
+
+        // Single-quoted path with spaces split across multiple args
+        yield return (34, ["--option1", "'path/with", "spaces/log'"], null, new CommandLineParseResultWrapper(null, new List<CommandLineParseOption> { new("option1", ["path/with spaces/log"]) }.ToArray(), []));
+
+        // Quoted path with spaces using = delimiter
+        yield return (35, ["--option1=\"path/with", "spaces/log\""], null, new CommandLineParseResultWrapper(null, new List<CommandLineParseOption> { new("option1", ["path/with spaces/log"]) }.ToArray(), []));
+
+        // Quoted path with spaces using : delimiter
+        yield return (36, ["--option1:\"path/with", "spaces/log\""], null, new CommandLineParseResultWrapper(null, new List<CommandLineParseOption> { new("option1", ["path/with spaces/log"]) }.ToArray(), []));
+
+        // Quoted path with spaces followed by another option
+        yield return (37, ["--option1", "\"path/with", "spaces\"", "--option2", "value"], null, new CommandLineParseResultWrapper(null, new List<CommandLineParseOption>
+        {
+            new("option1", ["path/with spaces"]),
+            new("option2", ["value"]),
+        }.ToArray(), []));
     }
 
     [TestMethod]


src/Platform/Microsoft.Testing.Platform/CommandLine/Parser.cs
```diff
@@ -78,6 +78,10 @@ private static CommandLineParseResult Parse(List<string> args, IEnvironment envi
 
             if (currentArg is not null)
             {
+                // When a quoted value is split across multiple args (e.g. a path with spaces),
+                // reassemble the fragments into a single argument.
+                currentArg = TryMergeQuotedArguments(args, ref i, currentArg);
```
Evangelink, line 83: Bug: After `TryMergeQuotedArguments` advances `i`, `args[i]` now points to the **last merged fragment** rather than the original argument. The error message will report something like `"Unexpected argument spaces/log\""` instead of the opening fragment.

Consider capturing the original value before the merge:
```csharp
if (currentArg is not null)
{
    string originalArg = args[i];
    currentArg = TryMergeQuotedArguments(args, ref i, currentArg);

    if (currentOption is null)
    {
        errors.Add(string.Format(CultureInfo.InvariantCulture, PlatformResources.CommandLineParserUnexpectedArgument, originalArg));
    }
```

Or alternatively, use `currentArg` (the merged result) so the error shows the full reconstructed value — that might actually be more helpful for debugging.

src/Platform/Microsoft.Testing.Platform/CommandLine/Parser.cs
```diff
@@ -114,6 +118,71 @@ static void ParseOptionAndSeparators(string arg, out string? currentOption, out
...
+
+            // Merge subsequent args until the closing quote is found.
+            StringBuilder merged = new(currentArg);
+            while (i + 1 < args.Count)
+            {
+                string nextArg = args[i + 1];
+                i++;
+                merged.Append(' ').Append(nextArg);
+
+                if (nextArg.TrimEnd().EndsWith(closingSuffix, StringComparison.Ordinal))
```
Evangelink, line 175: If the closing quote is never found, this loop consumes **all remaining args** — including subsequent `--options`. This can silently swallow intended options and produce confusing downstream errors.

Consider adding a guard to stop merging when encountering an arg that looks like an option (`--` prefix):
```csharp
while (i + 1 < args.Count)
{
    string nextArg = args[i + 1];

    // Don't consume what looks like a new option — the opening quote was probably not part of a split path.
    if (nextArg.StartsWith("--", StringComparison.Ordinal) || nextArg.StartsWith("-", StringComparison.Ordinal))
    {
        break;
    }

    i++;
    merged.Append(' ').Append(nextArg);
    ...
}
```

This would limit the blast radius of a missing closing quote. The tradeoff is that a path genuinely containing `--` would break, but that's extremely unlikely and the current behaviour (eating everything) is worse.

Alternatively, you could limit the merge to a small lookahead (e.g. 10 args max) as a safety net.

src/Platform/Microsoft.Testing.Platform/CommandLine/Parser.cs
```diff
@@ -114,6 +118,71 @@ static void ParseOptionAndSeparators(string arg, out string? currentOption, out
...
+                closingSuffix = "'";
+            }
+            else
+            {
+                return currentArg;
+            }
+
+            // Merge subsequent args until the closing quote is found.
+            StringBuilder merged = new(currentArg);
+            while (i + 1 < args.Count)
```
Evangelink, line 169: Nit: The `StringBuilder` is initialized with `currentArg` (untrimmed), but the quote detection works on `trimmed`. If `currentArg` has leading whitespace (e.g. `  "path/with`), the merged result starts with that whitespace. This works because the caller does `.Trim()` before passing to `TryUnescape`, but it's a subtle coupling.

Consider initializing with `trimmed` instead for consistency, or add a brief comment explaining that the caller trims.

src/Platform/Microsoft.Testing.Platform/CommandLine/Parser.cs
```diff
@@ -114,6 +118,71 @@ static void ParseOptionAndSeparators(string arg, out string? currentOption, out
...
+            string trimmed = currentArg.Trim();
+            if (trimmed.Length == 0)
+            {
+                return currentArg;
+            }
+
+            // Determine the opening quote style and whether it's already closed.
+            string? closingSuffix;
+            if (trimmed.StartsWith("\\\"", StringComparison.Ordinal))
+            {
```
Evangelink, line 135: The `Length > 4` threshold for escaped quotes (`\"...\"`): `\"\"` is 4 chars and would NOT be treated as an already-closed pair here. This is consistent with the same check in `TryUnescape`, so empty-content escaped quotes (`\"\"`) fall through unhandled. Is this intentional? If so, a brief comment would help (`// Length > 4 ensures non-empty content between the escaped quotes`).

src/Platform/Microsoft.Testing.Platform/CommandLine/Parser.cs
```diff
@@ -114,6 +118,71 @@ static void ParseOptionAndSeparators(string arg, out string? currentOption, out
...
+            StringBuilder merged = new(currentArg);
+            while (i + 1 < args.Count)
+            {
+                string nextArg = args[i + 1];
+                i++;
+                merged.Append(' ').Append(nextArg);
+
+                if (nextArg.TrimEnd().EndsWith(closingSuffix, StringComparison.Ordinal))
+                {
+                    return merged.ToString();
```
Evangelink, line 177: The `EndsWith` check on the closing suffix could be ambiguous for escaped-quote (`\"`). Consider the fragment `some-path/log\"extra\"` — it ends with `\"` and would be accepted as the closing match even though the `\"` in the middle was not the matching close.

In practice this is unlikely for file paths, but worth noting.

test/UnitTests/Microsoft.Testing.Platform.UnitTests/CommandLine/CommandLineTests.cs
```diff
@@ -198,6 +198,31 @@ public static string ParserTestDataFormat(MethodInfo methodInfo, object?[]? data
...
+
+        // Quoted path with spaces followed by another option
+        yield return (37, ["--option1", "\"path/with", "spaces\"", "--option2", "value"], null, new CommandLineParseResultWrapper(null, new List<CommandLineParseOption>
+        {
+            new("option1", ["path/with spaces"]),
+            new("option2", ["value"]),
+        }.ToArray(), []));
     }
 
     [TestMethod]
```
Evangelink, line 228: Good test coverage for the happy paths. A few additional test cases would strengthen the suite:

1. **Unclosed quote** (no matching close — verifies graceful degradation):
   ```csharp
   yield return (38, ["--option1", "\"path/with", "no-close"], null, ...expected error or merged result...);
   ```

2. **Non-split escaped quotes** (already-closed `\"value\"` in a single arg — verifies TryUnescape handles it):
   ```csharp
   yield return (39, ["--option1", "\\\"already-complete\\\""], null, 
       new CommandLineParseResultWrapper(null, new List<CommandLineParseOption> { new("option1", ["already-complete"]) }.ToArray(), []));
   ```

3. **Unclosed quote followed by another option** (verifies the merge doesn't swallow `--option2`):
   ```csharp
   yield return (40, ["--option1", "\"unclosed", "--option2", "value"], null, ...);
   ```
   This would expose the greedy-merge behavior I flagged in the `TryMergeQuotedArguments` comment.

Evangelink (changes requested):
## Overall Assessment

The fix addresses a real, user-facing bug (VS Code C# Dev Kit on Linux splitting quoted paths) and the approach is sound. The code follows existing conventions well (static local functions, ordinal comparisons, clear comments). Good test coverage for the primary scenarios.

### Summary of findings

**Bug (1 issue)**
- **Error message corruption after merge**: When `currentOption is null` and the merge advances `i`, the error message references `args[i]` which now points to the last merged fragment instead of the original argument.

**Design concerns (1 issue, medium severity)**
- **Greedy merge with no safety bound**: If a closing quote is never found, `TryMergeQuotedArguments` consumes ALL remaining args, including subsequent `--options`. This can produce very confusing downstream errors. Consider stopping when encountering an arg that looks like an option.

**Minor / Nit (2 issues)**
- `StringBuilder` initialization uses untrimmed `currentArg` while detection uses `trimmed` — subtle coupling with the caller's `.Trim()`.
- The `Length > 4` threshold for empty escaped quotes could use a clarifying comment.

**Missing test coverage (3 suggestions)**
- Unclosed quote (no matching close)
- Non-split escaped quotes (`\"value\"` in a single arg)
- Unclosed quote followed by a real option (exposes greedy merge)

The core logic is correct for all tested scenarios. The greedy merge and error-message issues are the items I'd want addressed before merging.

test/UnitTests/Microsoft.Testing.Platform.UnitTests/CommandLine/CommandLineTests.cs
```diff
@@ -198,6 +198,31 @@ public static string ParserTestDataFormat(MethodInfo methodInfo, object?[]? data
...
+        yield return (31, ["--option1", "\"path/with", "spaces/log\""], null, new CommandLineParseResultWrapper(null, new List<CommandLineParseOption> { new("option1", ["path/with spaces/log"]) }.ToArray(), []));
+
+        // Quoted path with spaces split across three args (double-quotes)
+        yield return (32, ["--option1", "\"/home/user/Example", "Solution/artifacts/build", "bin/Log\""], null, new CommandLineParseResultWrapper(null, new List<CommandLineParseOption> { new("option1", ["/home/user/Example Solution/artifacts/build bin/Log"]) }.ToArray(), []));
+
+        // Escaped double-quote path with spaces split across multiple args
+        yield return (33, ["--option1", "\\\"/path/with", "spaces/log\\\""], null, new CommandLineParseResultWrapper(null, new List<CommandLineParseOption> { new("option1", ["/path/with spaces/log"]) }.ToArray(), []));
+
+        // Single-quoted path with spaces split across multiple args
+        yield return (34, ["--option1", "'path/with", "spaces/log'"], null, new CommandLineParseResultWrapper(null, new List<CommandLineParseOption> { new("option1", ["path/with spaces/log"]) }.ToArray(), []));
```
Youssef1313, line 212: This is not making sense to me.

A path with space should be received via entrypoint args as a single argument. Generally, if you do `TestApp.exe --option "Value With Space"` on a command-line, what should happen is that `args` in entrypoint has `--option` followed by `Value With Space` as the second element, with the double quotes already removed for you.

PR #7876, opened 2026-04-27
Convert ExitCodes from static class to enum

### LINKED ISSUE #7807: Convert `ExitCodes` constants class to an `enum`
**Priority**: Low
**Estimated Effort**: Small

`src/Platform/Microsoft.Testing.Platform/Helpers/ExitCodes.cs:12` has a TODO to convert to an `enum` and rename to `ExitCode` (singular). This improves type safety, discoverability, and is consistent with .NET conventions. This would be a breaking change to the public API surface and should be planned for the next major version.

Evangelink:
## Summary

Converts the internal `ExitCodes` static class with `public const int` fields to an `internal enum`. This improves type safety, discoverability, and enables `Enum.IsDefined` checks for validating known MTP exit codes.

## Breaking change analysis

**No breaking change**, because:

1. **`const int` values are inlined at compile time** — extensions compiled against the old `static class ExitCodes` have the literal integer values baked into their IL. The consumer assembly has zero runtime reference to the `ExitCodes` type. Verified by disassembling a Consumer.dll — the string "ExitCodes" does not appear.

2. **Runtime verified** — built an old consumer against `static class ExitCodes`, then swapped the library to `enum ExitCodes` without recompiling the consumer. It runs correctly.

3. **Out-of-repo IVT consumers don't use ExitCodes** — checked `MSTest.Engine`, `OpenTelemetry`, `AzureFoundry`, `Platform.AI`, `HotReload`, `CrashDump`, `Telemetry`, `VSTestBridge` — none reference `ExitCodes`.

4. **Source-embedded consumers always recompile** — extensions that DO use `ExitCodes` (Retry, TrxReport, HangDump, MSBuild) embed `ExitCodes.cs` via `<Compile Include>` and compile their own internal copy.

5. **Not in any `PublicAPI.txt`** — `ExitCodes` is `internal`, so no public API surface change.

## Changes

- Converted `ExitCodes` from `internal static class` to `internal enum` (same numeric values)
- Added `(int)` casts at all usage sites in source code
- Added `ExitCodes` enum overloads to `AcceptanceAssert.AssertExitCodeIs`/`AssertExitCodeIsNot` for both `TestHostResult` and `DotnetMuxerResult`
- Added explicit `using Microsoft.Testing.Platform.Helpers` to `AcceptanceAssert.cs` (shared across projects with different global usings)

Fixes #7807

### DIFF
--- a/src/Platform/Microsoft.Testing.Platform/Helpers/ExitCodes.cs
+++ b/src/Platform/Microsoft.Testing.Platform/Helpers/ExitCodes.cs
@@ -10,7 +10,7 @@ namespace Microsoft.Testing.Platform.Helpers;
 /// On POSIX systems the standard exit code is 0 for success and any number from 1 to 255 for anything else.
 /// </summary>
 [Embedded]
-internal enum ExitCodes
+internal enum ExitCode
 {
     Success = 0,
     GenericFailure = 1,
--- a/src/Platform/Microsoft.Testing.Platform/Helpers/ExitCodes.cs
+++ b/src/Platform/Microsoft.Testing.Platform/Helpers/ExitCodes.cs
@@ -9,25 +9,21 @@ namespace Microsoft.Testing.Platform.Helpers;
 /// We use positive exit codes for failure because POSIX/BASH exit codes are unsigned 8-bit integers.
 /// On POSIX systems the standard exit code is 0 for success and any number from 1 to 255 for anything else.
 /// </summary>
-// TODO: Consider changing this to an enum, and rename to 'ExitCode' to follow enum naming convention.
-// Being an enum makes it easier to do 'Enum.IsDefined' checks to validate if an exit code is a known MTP exit code.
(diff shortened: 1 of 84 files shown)


src/Platform/Microsoft.Testing.Platform/Helpers/ExitCodes.cs
```diff
@@ -9,25 +9,21 @@ namespace Microsoft.Testing.Platform.Helpers;
 /// We use positive exit codes for failure because POSIX/BASH exit codes are unsigned 8-bit integers.
 /// On POSIX systems the standard exit code is 0 for success and any number from 1 to 255 for anything else.
 /// </summary>
-// TODO: Consider changing this to an enum, and rename to 'ExitCode' to follow enum naming convention.
```
Youssef1313, line 12: Renaming wasn't addressed here.

PR #7925, opened 2026-04-29
Fix log accumulation across DynamicData test invocations

### LINKED ISSUE #7846: Log collection is broken, >20 GB RAM usage for simple project
## Describe the bug

We have issues with a project, where the test project takes up huge amounts of RAM and eventually crashes.
We have "traced" this back to some broken logging.
It seems that (for `DynamicData` tests), logs are collected wrongly. 
Each subsequent test from the DynamicData contains ALL logs from ALL previous invocations in the TEST SUMMARY where all logs are collected and presented at once.

Evidence:
Proper log when you look at a single test:

Image

However when you open the "summary" log, the same test shows ALL OTHER logs first.
We would not expect CONSOLE_0_0 to see in the TestMethod1 (100) logs. It should also start with CONSOLE_100_0 like the individual test's log above.

Image

The test was 1000 DynamicData tests, each test has 1000 iterations logging 2 lines, so we'd expect around 2'000'000 lines of logs give or take, however the log file has over 12'000'000 lines (and seems to be truncated after around test 110)

Something is VERY wrong and it causes us a lot of test weirdness.

## Version used

We are using MTP/MSTest via MSTest.SDK 4.2.1

## Steps To Reproduce

A repro project is attached. Just try to open the test results in VS (April 2026 version if that makes a difference) and you'll see :)

[mstestRAM.zip](https://github.com/user-attachments/files/27095336/mstestRAM.zip)

## Expected behavior

Not so much RAM in use, and logs collected correctly

## Actual behavior

Too much RAM in use, and summary logs collected wrongly

Image

Evangelink:
## Summary

Fixes #7846

When running `DynamicData` tests with multiple invocations, the `TestContextImplementation` output buffers (`_stdOutStringBuilder`, `_stdErrStringBuilder`, `_traceStringBuilder`) were never cleared between invocations. Since DynamicData tests reuse the same `TestContext` instance across all invocations, logs accumulated exponentially:

- Test 1 result: logs from test 1
- Test 2 result: logs from test 1 + test 2
- Test 3 result: logs from test 1 + test 2 + test 3
- ...

This caused >20 GB RAM usage for simple projects and truncated/broken log output in test summaries.

## Root Cause

`GetOut()`, `GetErr()`, and `GetTrace()` on `TestContextImplementation` returned the `StringBuilder` contents via `ToString()` but never cleared them. In contrast, `GetAndClearDiagnosticMessages()` and `GetResultFiles()` already followed a get-then-clear pattern.

## Fix

Replace `GetOut()`, `GetErr()`, `GetTrace()` with `GetAndClearOutput()`, `GetAndClearError()`, `GetAndClearTrace()` that clear the `StringBuilder` after extracting the value — matching the existing `GetAndClearDiagnosticMessages()` pattern.

## Changes

- **`TestContextImplementation.cs`** — Replace non-clearing getters with `GetAndClearOutput/Error/Trace` methods
- **`TestMethodInfo.cs`** — Update call site in `InvokeAsync` finally block
- **`UnitTestRunner.cs`** — Update call sites in assembly init/cleanup
- **`TestClassInfo.cs`** — Update call sites in class init/cleanup
- **`TestMethodInfoTests.cs`** — Add 3 regression tests verifying buffers don't accumulate across invocations
- **`TestContextImplementationTests.cs`** — Update existing thread-safety test to use renamed methods

### DIFF
--- a/src/Adapter/MSTestAdapter.PlatformServices/Services/TestContextImplementation.cs
+++ b/src/Adapter/MSTestAdapter.PlatformServices/Services/TestContextImplementation.cs
@@ -380,12 +380,27 @@ internal sealed class TestContextImplementation : TestContext, ITestContext, IDi
         return _testContextMessageStringBuilder;
     }
 
-    internal string? GetOut()
-        => _stdOutStringBuilder?.ToString();
+    internal string? GetAndClearOutput()
+    {
+        string? output = _stdOutStringBuilder?.ToString();
+        _stdOutStringBuilder?.Clear();
 
-    internal string? GetErr()
-        => _stdErrStringBuilder?.ToString();
+        return output;
+    }
 
-    internal string? GetTrace()
-        => _traceStringBuilder?.ToString();
+    internal string? GetAndClearError()
+    {
+        string? error = _stdErrStringBuilder?.ToString();
+        _stdErrStringBuilder?.Clear();
+
+        return error;
+    }
+
+    internal string? GetAndClearTrace()
+    {
+        string? trace = _traceStringBuilder?.ToString();
+        _traceStringBuilder?.Clear();
+
+        return trace;
+    }
 }
--- a/src/Adapter/MSTestAdapter.PlatformServices/Services/TestContextImplementation.cs
+++ b/src/Adapter/MSTestAdapter.PlatformServices/Services/TestContextImplementation.cs
@@ -47,6 +47,15 @@ internal sealed class TestContextImplementation : TestContext, ITestContext, IDi
             => _builder.Clear();
 
         [MethodImpl(MethodImplOptions.Synchronized)]
+        internal string GetAndClear()
+        {
+            string result = _builder.ToString();
+            _builder.Clear();
+
+            return result;
+        }
+
+        [MethodImpl(MethodImplOptions.Synchronized)]
         public override string ToString()
             => _builder.ToString();
     }
@@ -381,26 +390,11 @@ internal sealed class TestContextImplementation : TestContext, ITestContext, IDi
     }
 
     internal string? GetAndClearOutput()
-    {
-        string? output = _stdOutStringBuilder?.ToString();
-        _stdOutStringBuilder?.Clear();
-
-        return output;
-    }
+        => _stdOutStringBuilder?.GetAndClear();
 
     internal string? GetAndClearError()
-    {
-        string? error = _stdErrStringBuilder?.ToString();
-        _stdErrStringBuilder?.Clear();
-
-        return error;
-    }
+        => _stdErrStringBuilder?.GetAndClear();
 
     internal string? GetAndClearTrace()
-    {
-        string? trace = _traceStringBuilder?.ToString();
-        _traceStringBuilder?.Clear();
-
-        return trace;
-    }
+        => _traceStringBuilder?.GetAndClear();
 }
--- a/test/UnitTests/MSTestAdapter.PlatformServices.UnitTests/Execution/TestMethodInfoTests.cs
+++ b/test/UnitTests/MSTestAdapter.PlatformServices.UnitTests/Execution/TestMethodInfoTests.cs
@@ -1501,11 +1501,14 @@ public class TestMethodInfoTests : TestContainer
         ((string[])expectedArguments[1]).SequenceEqual((string[])resolvedArguments[1]!).Should().BeTrue();
     }
 
+    // Regression tests for https://github.com/microsoft/testfx/issues/7846
+    // Verify that log output buffers are cleared between invocations to prevent
+    // exponential memory growth with DynamicData tests.
+    // NOTE: The TestClassInfo (class init/cleanup) and UnitTestRunner (assembly init/cleanup)
+    // call sites use the same GetAndClear* methods tested in isolation in
+    // TestContextImplementationTests.GetAndClear{Output,Error,Trace}_ShouldReturnContentThenClearBuffer.
     public async Task InvokeAsync_ShouldNotAccumulateLogOutputAcrossMultipleInvocations()
     {
-        // Regression test for https://github.com/microsoft/testfx/issues/7846
-        // Verify that log output buffers are cleared between invocations to prevent
-        // exponential memory growth with DynamicData tests.
         DummyTestClass.TestMethodBody = _ => _testContextImplementation.WriteConsoleOut("invocation_output");
 
         TestResult result1 = await _testMethodInfo.InvokeAsync(null);
@@ -1517,7 +1520,6 @@ public class TestMethodInfoTests : TestContainer
 
     public async Task InvokeAsync_ShouldNotAccumulateLogErrorAcrossMultipleInvocations()
     {
-        // Regression test for https://github.com/microsoft/testfx/issues/7846
         DummyTestClass.TestMethodBody = _ => _testContextImplementation.WriteConsoleErr("error_output");
 
         TestResult result1 = await _testMethodInfo.InvokeAsync(null);
@@ -1529,7 +1531,6 @@ public class TestMethodInfoTests : TestContainer
 
     public async Task InvokeAsync_ShouldNotAccumulateDebugTraceAcrossMultipleInvocations()
     {
-        // Regression test for https://github.com/microsoft/testfx/issues/7846
         DummyTestClass.TestMethodBody = _ => _testContextImplementation.WriteTrace("trace_output");
 
         TestResult result1 = await _testMethodInfo.InvokeAsync(null);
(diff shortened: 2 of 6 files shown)


test/UnitTests/MSTestAdapter.PlatformServices.UnitTests/Services/TestContextImplementationTests.cs
```diff
@@ -360,8 +360,8 @@ public void WritesFromBackgroundThreadShouldNotThrow()
         });
 
         t.Start();
-        _ = testContextImplementation.GetOut();
-        _ = testContextImplementation.GetErr();
+        _ = testContextImplementation.GetAndClearOutput();
+        _ = testContextImplementation.GetAndClearError();
         t.Join();
     }
 }
```
Evangelink, line 404: **[Coverage]** `GetAndClearTrace()` was also renamed (from `GetTrace()`) and also calls `Clear()` under concurrency, but `WritesFromBackgroundThreadShouldNotThrow` was only updated for `Output` and `Error`. Adding a concurrent `GetAndClearTrace()` call here would complete the thread-safety coverage for all three new methods.

**Impact**: A race condition in the `Trace` buffer's clear path would go undetected.

**Suggestion**: Add `_ = testContextImplementation.GetAndClearTrace();` alongside the other two calls in the background-thread test.

Evangelink, line 404: **[Coverage]** `TestContextImplementationTests.cs` has no test that directly verifies the *clearing* part of `GetAndClearOutput/Error/Trace`. The renamed thread-safety test only checks that no exception is thrown; it doesn't assert the buffer is empty after the call.

**Impact**: The core new contract — that calling `GetAndClear*()` resets the buffer so a subsequent call returns `null`/empty — is untested in isolation. A regression that breaks clearing (but not the read) would not be caught here.

**Suggestion**: Add a focused test, e.g.:
```csharp
public void GetAndClearOutput_ShouldReturnContentThenClearBuffer()
{
    _testContextImplementation = CreateTestContextImplementation();
    _testContextImplementation.WriteConsoleOut("hello");

    string? first  = _testContextImplementation.GetAndClearOutput();
    string? second = _testContextImplementation.GetAndClearOutput();

    first.Should().Be("hello");
    second.Should().BeNullOrEmpty();
}
```
A similar test for `GetAndClearError` and `GetAndClearTrace` would give full coverage of the fix.

test/UnitTests/MSTestAdapter.PlatformServices.UnitTests/Execution/TestMethodInfoTests.cs
```diff
@@ -1501,6 +1501,44 @@ public void ResolveArgumentsShouldReturnPopulatedParamsWithAllProvided()
...
+    {
+        // Regression test for https://github.com/microsoft/testfx/issues/7846
+        DummyTestClass.TestMethodBody = _ => _testContextImplementation.WriteTrace("trace_output");
+
+        TestResult result1 = await _testMethodInfo.InvokeAsync(null);
+        TestResult result2 = await _testMethodInfo.InvokeAsync(null);
+
+        result1.DebugTrace.Should().Be("trace_output");
+        result2.DebugTrace.Should().Be("trace_output");
+    }
```
Evangelink, line 1541: **[Coverage]** The regression tests for log accumulation only cover the `TestMethodInfo.InvokeAsync` path. The PR also changes `TestClassInfo` (class init/cleanup, 3 call sites) and `UnitTestRunner` (assembly init/cleanup, 3 call sites), but there are no corresponding accumulation tests for those paths.

**Impact**: If the clear-on-read pattern is ever accidentally reverted in one of those call sites, the tests here won't catch it.

**Suggestion**: Add parallel regression tests that simulate two class-init or assembly-init runs (or at minimum a comment noting these paths are covered by acceptance tests, if that is the case).

Evangelink (commented):
### Summary

**Date**: 2026-04-29

### Key Findings

1. **[Coverage — Minor]** `WritesFromBackgroundThreadShouldNotThrow` in `TestContextImplementationTests.cs` was updated for `Output` and `Error` but not for `Trace`. `GetAndClearTrace()` also calls `Clear()` under concurrency and should be included in the thread-safety test.

2. **[Coverage — Important]** No test in `TestContextImplementationTests.cs` directly verifies the *clearing* contract of the new `GetAndClear*` methods. The only test for these methods is the thread-safety check (which only asserts no exception). A dedicated read-then-clear test would document and protect the core invariant introduced by this fix.

3. **[Coverage — Minor]** Regression tests for log accumulation are scoped to `TestMethodInfo.InvokeAsync`. The `TestClassInfo` (class init/cleanup) and `UnitTestRunner` (assembly init/cleanup) call sites that were also changed have no equivalent accumulation tests.

### Recommendations

- Add a `GetAndClearOutput_ShouldReturnContentThenClearBuffer`-style test (and equivalents for Error/Trace) in `TestContextImplementationTests.cs` to directly validate the clearing behavior.
- Extend `WritesFromBackgroundThreadShouldNotThrow` to also call `GetAndClearTrace()` for completeness.
- Consider adding accumulation regression tests for the class/assembly init-cleanup paths, or add a comment noting they're exercised by acceptance tests.

The existing accumulation tests in `TestMethodInfoTests.cs` are well-structured and correctly detect the regression: `result2.LogOutput.Should().Be("invocation_output")` would fail with the old non-clearing code (it would produce `"invocation_outputinvocation_output"`), so the core fix is properly verified.

---

*Generated by [Test Expert Reviewer 🧪](https://github.com/microsoft/testfx/actions/runs/25092478385)*

> 🧪 *Test quality reviewed by [Test Expert Reviewer 🧪](https://github.com/microsoft/testfx/actions/runs/25092478385)*

Evangelink (commented):
### Summary

**Date**: 2026-04-29

### Key Findings

**[Threading — Minor/Accepted]** `GetAndClearOutput/Error/Trace` each perform two individually-synchronized calls (`ToString()` + `Clear()`). Content written by a background thread in the TOCTOU window is silently discarded. This matches the pre-existing `GetAndClearDiagnosticMessages()` pattern and is an accepted design trade-off in this codebase. See inline comment.

### Positive Observations

- **Root cause addressed correctly**: The `TestContextImplementation` output buffers now clear after each read, consistent with the established `GetAndClearDiagnosticMessages` pattern. The exponential memory growth scenario is fully closed.
- **Bonus fix in multi-assembly cleanup**: In `UnitTestRunner.RunAssemblyCleanupAsync`, the `foreach` over multiple `TestAssemblyInfo` entries previously accumulated logs across iterations (each `GetOut()` returned all prior logs); the new `GetAndClearOutput()` correctly scopes each iteration to only its own cleanup logs.
- **Regression tests are valid**: The three new `public async Task` tests in `TestMethodInfoTests.cs` are properly discoverable by `TestFramework.ForTestingMSTest` (which discovers all public parameterless methods returning `void` or `Task` in `TestContainer` subclasses — no `[TestMethod]` attribute required).
- **Exclusive control flow**: In both `TestClassInfo.DoRunAsync` and `RunAssemblyCleanupAsync`, the failure path (creates a new `TestResult` and returns) and the success path (`lastResult +=`) are mutually exclusive, so there is no double-clear or double-append risk.
- **No public API surface changed**: All three renamed methods remain `internal`.

### Recommendations

- Consider adding a brief comment to the three new methods noting the intentional TOCTOU (consistent with `GetAndClearDiagnosticMessages`), to signal future maintainers that this is deliberate rather than an oversight.

---

*Generated by [Expert Code Reviewer](https://github.com/microsoft/testfx/actions/runs/25092478391)*

> 🧠 *Reviewed by [Expert Code Reviewer 🧠](https://github.com/microsoft/testfx/actions/runs/25092478391)*

src/Adapter/MSTestAdapter.PlatformServices/Services/TestContextImplementation.cs
```diff
@@ -380,12 +380,27 @@ private SynchronizedStringBuilder GetTestContextMessagesStringBuilder()
...
-    internal string? GetOut()
-        => _stdOutStringBuilder?.ToString();
+    internal string? GetAndClearOutput()
+    {
+        string? output = _stdOutStringBuilder?.ToString();
+        _stdOutStringBuilder?.Clear();
 
-    internal string? GetErr()
-        => _stdErrStringBuilder?.ToString();
+        return output;
```
Evangelink: **[Threading]** The `ToString()` → `Clear()` sequence is not atomic: any content written by a background thread between these two synchronized calls will be silently discarded.

**Mechanism**: `SynchronizedStringBuilder` serializes each individual call via `[MethodImpl(MethodImplOptions.Synchronized)]`, but the pair of calls has a TOCTOU window. A background thread that writes between step 1 and step 2 has its data captured in neither the returned string nor the post-clear buffer.

**Impact**: Potential silent data loss for log output generated from background threads during test teardown. This affects all three new methods (`GetAndClearOutput`, `GetAndClearError`, `GetAndClearTrace`).

**Context**: This is consistent with the pre-existing `GetAndClearDiagnosticMessages()` extension which also uses a non-atomic `GetDiagnosticMessages()` + `ClearDiagnosticMessages()` pair, so it is an accepted trade-off. Worth noting in a code comment to prevent future refactoring that accidentally "fixes" this by removing the synchronization entirely.

PR #6762, opened 2025-10-17
Add support for net10

Evangelink:

PR #7976, opened 2026-05-01
Add net10.0 to SupportedNetFrameworks, drop net9.0 (EOL May 12 2026)

Rolling2405:
## Summary

Updates SupportedNetFrameworks in Directory.Build.props from 
et8.0;net9.0 to 
et8.0;net10.0.

.NET 9.0 STS reaches end-of-life on **May 12, 2026**. The repository has already been moving in this direction:
- TargetFrameworks.cs in test infrastructure lists 
et10.0 as primary with 
et8.0 as secondary (no net9.0)
- Previous commit removed net9.0 from test utilities entirely

This single-property change propagates 
et10.0 automatically to all ~30 source and test projects that reference \.

## Changes

- Directory.Build.props: 
et8.0;net9.0 → 
et8.0;net10.0

## Note on closed PR #6762

That PR was opened in October 2025 when net10.0 was still pre-release. Now that net10.0 is GA and net9.0 is at EOL, the timing is right.

### DIFF
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -29,7 +29,7 @@
     <ModernUwpMinimum>net9.0-windows10.0.17763.0</ModernUwpMinimum>
     <WinUiMinimum>net8.0-windows10.0.18362.0</WinUiMinimum>
 
-    <SupportedNetFrameworks>net8.0;net9.0</SupportedNetFrameworks>
+    <SupportedNetFrameworks>net8.0;net10.0</SupportedNetFrameworks>
   </PropertyGroup>
 
   <!-- Build config -->
--- a/eng/Version.Details.xml
+++ b/eng/Version.Details.xml
@@ -1,29 +1,29 @@
 <?xml version="1.0" encoding="utf-8"?>
 <Dependencies>
   <ToolsetDependencies>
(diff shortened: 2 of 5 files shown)


Directory.Build.props
```diff
@@ -29,7 +29,7 @@
     <ModernUwpMinimum>net9.0-windows10.0.17763.0</ModernUwpMinimum>
     <WinUiMinimum>net8.0-windows10.0.18362.0</WinUiMinimum>
 
-    <SupportedNetFrameworks>net8.0;net9.0</SupportedNetFrameworks>
+    <SupportedNetFrameworks>net8.0;net10.0</SupportedNetFrameworks>
```
Youssef1313, line 32: I don't think we should do that. It's completely fine to still keep net9.0, and there is no obvious need to add net10.0.

Users today can already target net10.0 and will be consuming our net9.0 artifacts without any problem.

Evangelink:
Thanks for the contribution, but we won't be taking this in.

Dropping `net9.0` is a breaking change, so it needs to be planned for a breaking change release — and we're not doing that right now.

As for adding `net10.0`, as @Youssef1313 pointed out, it doesn't add much value for end users at the moment, so there's no strong reason to do it either.

Closing for now.

PR #7977, opened 2026-05-01
perf: use Lock, FrozenDictionary, Span, and throw helpers on NET8+/NET9+

Rolling2405:
## Summary

This PR stacks on top of #FIXME (the NET10 TFM PR from branch \eat/net10-tfm\) and adds a set of performance and correctness improvements, all guarded by appropriate \#if\ TFM conditions so existing targets (netstandard2.0, net462, net8.0, etc.) compile unchanged.

> **Note:** We cannot build testfx locally because it requires the .NET 11 preview SDK pinned in \global.json\. All changes are mechanically safe: no algorithmic changes, pure API substitutions with equivalent semantics, all guarded by the correct TFM conditions.

---

### Finding 1 — \RandomId.cs\: Fix CA2002 correctness bug + \Lock\ upgrade (NET9+)

**This is a correctness fix, not just a perf improvement.** The previous code did \lock (Pool)\ where \Pool\ is a \const string\. String constants are interned, meaning *any other code in the process* that happens to intern the same string literal can deadlock or interfere with this lock. This is the CA2002 violation ('Do not lock on objects with weak identity').

Fix: introduce a dedicated \s_lock\ field (\System.Threading.Lock\ on NET9+, \object\ on older targets) and lock on that instead.

### Finding 2 — \ClassCleanupManager.cs\: Dedicated lock + \CollectionsMarshal\ (NET8+)

- The class already manually serialised all access to \_remainingTestCountsByClass\ via \lock\. The \ConcurrentDictionary\ was unnecessary overhead; replaced with plain \Dictionary\.
- Added a dedicated \_lock\ field (\System.Threading.Lock\ on NET9+, \object\ otherwise) instead of locking on the collection itself.
- \ShouldRunEndOfAssemblyCleanup\ now reads inside the lock (was previously safe only because \ConcurrentDictionary.IsEmpty\ is atomic; plain \Dictionary.Count\ is not thread-safe without a lock).
- \MarkTestComplete\ on NET8+ uses \CollectionsMarshal.GetValueRefOrNullRef\ to get a direct reference to the value, avoiding a second dictionary lookup for the write-back — a single lookup replaces the old TryGetValue + indexer set pattern.
- \MarkClassComplete\ uses \TryGetValue\ + \Remove\ instead of \ConcurrentDictionary.TryRemove\ (compatible with all TFMs including netstandard2.0 and net462 since all access is already behind the lock).

### Finding 4 — \TypeNameHelper.cs\: \FrozenDictionary\ for \BuiltInTypeNames\ (NET8+)

\BuiltInTypeNames\ is a read-only lookup table initialised once at startup. On NET8+ it is now built as a \FrozenDictionary\, which optimises its internal hash structure for read-heavy workloads and produces faster \TryGetValue\ calls.

### Finding 5 — \SerializerUtilities.cs\: \FrozenDictionary\ for \Serializers\/\Deserializers\ (NET8+)

Same rationale as above. Both dictionaries are populated once in the static constructor and then only read. On NET8+ they are frozen after population via \.ToFrozenDictionary()\.

### Finding 7 — \Assert.That.cs\: \AsSpan\ + \SequenceEqual\ for substring comparisons (NET8+)

Five \input.Substring(i, N).Equals(literal)\ / \input.Substring(i, N) == literal\ patterns replaced with \input.AsSpan(i, N).SequenceEqual(literal)\ on NET8+. \AsSpan\ avoids allocating a temporary string heap object; \SequenceEqual\ on \ReadOnlySpan\ is as fast as the string comparison but allocation-free.

### Finding 11 — \MSTestDiscoverer.cs\: \ArgumentNullException.ThrowIfNull\ (NET6+)

Replaces verbose \if (x is null) throw new ArgumentNullException(nameof(x));\ blocks with the single-call helper added in .NET 6.

### Finding 12 — \MSTestExecutor.cs\: \ArgumentNullException.ThrowIfNull\ (NET6+)

Same pattern as Finding 11, applied to the two \RunTestsAsync\ overloads.

### Finding 14 — \AssertScope.cs\: \ObjectDisposedException.ThrowIf\ (NET7+)

Resolves the existing \#pragma warning disable CA1513\ comment by conditionally using \ObjectDisposedException.ThrowIf\ on NET7+ (where the helper was introduced), falling back to the manual check on older targets.

### Finding 10 — \Assert.AreEqual.String.cs\: Range operator instead of \Substring\ (all TFMs)

\	ext.Substring(0, text.Length - 3)\ → \	ext[..^3]\ and \	ext.Substring(3)\ → \	ext[3..]\. No \#if\ guard needed — C# range syntax on \string\ is supported on all project TFMs (it compiles to the same IL as the \Substring\ calls).

---

## Testing

All changes are conservative substitutions:
- \Lock\ / \object\ — same semantics, just a better-typed lock primitive
- \FrozenDictionary\ — same \IDictionary\ interface, immutable after construction
- \AsSpan\/\SequenceEqual\ — same ordinal byte-by-byte comparison, no allocation
- \ArgumentNullException.ThrowIfNull\ / \ObjectDisposedException.ThrowIf\ — same exception type and message, same conditions
- Range operators — exact same \Substring\ semantics

### DIFF
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -29,7 +29,7 @@
     <ModernUwpMinimum>net9.0-windows10.0.17763.0</ModernUwpMinimum>
     <WinUiMinimum>net8.0-windows10.0.18362.0</WinUiMinimum>
 
-    <SupportedNetFrameworks>net8.0;net9.0</SupportedNetFrameworks>
+    <SupportedNetFrameworks>net8.0;net10.0</SupportedNetFrameworks>
   </PropertyGroup>
 
   <!-- Build config -->
--- a/eng/Version.Details.xml
+++ b/eng/Version.Details.xml
@@ -1,29 +1,29 @@
 <?xml version="1.0" encoding="utf-8"?>
 <Dependencies>
   <ToolsetDependencies>
-    <Dependency Name="Microsoft.DotNet.Arcade.Sdk" Version="11.0.0-beta.26229.1">
+    <Dependency Name="Microsoft.DotNet.Arcade.Sdk" Version="11.0.0-beta.26230.2">
       <Uri>https://github.com/dotnet/arcade</Uri>
-      <Sha>3ed6c23fee67b840de0fb5473ea6a95fc667b0a9</Sha>
+      <Sha>8a7b2d3d39078db20f33067c7929b88fdff4ee63</Sha>
     </Dependency>
-    <Dependency Name="Microsoft.DotNet.Build.Tasks.Templating" Version="11.0.0-beta.26229.1">
+    <Dependency Name="Microsoft.DotNet.Build.Tasks.Templating" Version="11.0.0-beta.26230.2">
       <Uri>https://github.com/dotnet/arcade</Uri>
-      <Sha>3ed6c23fee67b840de0fb5473ea6a95fc667b0a9</Sha>
+      <Sha>8a7b2d3d39078db20f33067c7929b88fdff4ee63</Sha>
     </Dependency>
-    <Dependency Name="Microsoft.DotNet.XliffTasks" Version="11.0.0-beta.26229.1">
+    <Dependency Name="Microsoft.DotNet.XliffTasks" Version="11.0.0-beta.26230.2">
       <Uri>https://github.com/dotnet/arcade</Uri>
-      <Sha>3ed6c23fee67b840de0fb5473ea6a95fc667b0a9</Sha>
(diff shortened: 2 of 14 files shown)


Youssef1313 (commented):
There doesn't seem to be any proven performance improvements backed up with any real-world numbers to justify the complexity added here.

Evangelink:
Closing this since it depends on #7976, which was closed.

PR #8095, opened 2026-05-11
[Perf Improver] perf: eliminate LINQ allocations in IsIgnored hot path

Evangelink:
- **docs: add AwesomeAssertions to glossary**
- **docs: update glossary - weekly full scan 2026-05-04**
- **perf: eliminate yield-iterator allocations in test execution hot path**
- **feat(fv): Task 1+2 — research two new targets; informal spec for UnitTestOutcomeHelper.ToTestOutcome**
- **perf: eliminate LINQ allocations in IsIgnored hot path**
- **test: add unit tests for `AttributeExtensions.IsIgnored` grouping semantics**


src/Adapter/MSTestAdapter.PlatformServices/Execution/TestMethodRunner.cs
```diff
@@ -178,14 +178,15 @@ internal async Task<TestResult[]> RunTestMethodAsync()
 
     private async Task<bool> TryExecuteDataSourceBasedTestsAsync(List<TestResult> results)
     {
-        DataSourceAttribute[] dataSourceAttribute = _testMethodInfo.GetAttributes<DataSourceAttribute>();
-        if (dataSourceAttribute is { Length: 1 })
+        // IsAttributeDefined avoids the array allocation of GetAttributes<DataSourceAttribute>().
+        // DataSourceAttribute is sealed and does not allow multiple, so presence == exactly one.
```
Evangelink: **Nitpick**: The comment merges two separate guarantees without explaining how each contributes to correctness.

`sealed` is relevant because `IsAttributeDefined<DataSourceAttribute>` checks `attribute is TAttribute`, which matches derived types too. Without `sealed`, someone could apply both a `DataSourceAttribute` and a subclass instance — `IsAttributeDefined` would return `true` while there are effectively two DataSource-family attributes, silently changing behavior from the old `{ Length: 1 }` guard.

`AllowMultiple = false` (the default when not set) is what enforces "at most one instance of `DataSourceAttribute` itself".

Consider making the connection explicit, e.g.:

```csharp
// IsAttributeDefined avoids the array allocation of GetAttributes<DataSourceAttribute>().
// DataSourceAttribute is sealed (no subclass can coexist alongside it) and
// has AllowMultiple = false (default), so IsAttributeDefined is equivalent to Length == 1.
```

**Why it matters**: The current wording may lead a future reader to believe `sealed` alone prevents multiples, which could cause an incorrect refactoring if applied to a non-sealed attribute.

Evangelink (commented):
### Summary

**Date**: 2026-05-11

### Key Findings

The PR cleanly eliminates LINQ allocations in the `IsIgnored` hot path and two related call sites. The logic is preserved correctly. Two minor nitpicks:

1. **Important** — `TestMethodRunner.cs` L182: The comment conflates the roles of `sealed` and `AllowMultiple = false`. `sealed` prevents a subclass co-existence scenario that would break the `IsAttributeDefined`-based equivalence; `AllowMultiple = false` prevents duplicate base-type instances. Both are necessary and deserve separate explanations.

2. **Minor** — `AttributeHelpers.cs` L50: When marking a group satisfied, `groupState.FirstMessage` is preserved in the new tuple even though it can never be read in that state. `null` would be cleaner.

### Positive Highlights

- ✅ The two-pass fast/slow-path split is a well-considered design: the first pass avoids a Dictionary allocation in the common case (no `ConditionBaseAttribute` present).
- ✅ The `GetRetryAttribute` refactoring in `TestMethodInfo.cs` is clean with no behavioral changes.
- ✅ Good inline comments throughout explaining *why* the allocation-avoiding approach is used.

---

*🔍 Meticulously inspected by [PR Nitpick Reviewer](https://github.com/microsoft/testfx/actions/runs/25660866636)*

> 🔍 *Meticulously inspected by [PR Nitpick Reviewer 🔍](https://github.com/microsoft/testfx/actions/runs/25660866636)*

src/Adapter/MSTestAdapter.PlatformServices/Helpers/AttributeHelpers.cs
```diff
@@ -9,27 +9,58 @@ internal static class AttributeExtensions
...
+                groups[conditionAttr.GroupName] = (shouldRun, shouldRun ? null : conditionAttr.IgnoreMessage);
+            }
+            else if (!groupState.Satisfied)
             {
-                bool shouldRun = attribute.Mode == ConditionMode.Include ? attribute.IsConditionMet : !attribute.IsConditionMet;
                 if (shouldRun)
                 {
-                    atLeastOneInGroupIsSatisfied = true;
-                    break;
+                    groups[conditionAttr.GroupName] = (true, groupState.FirstMessage);
```
Evangelink: **Nitpick**: When a group transitions from unsatisfied → satisfied, `groupState.FirstMessage` is preserved but can never be returned — a satisfied group is not reported as ignored. Storing `null` would be more semantically honest and slightly cheaper (no field copy):

```csharp
groups[conditionAttr.GroupName] = (true, null);
```

**Why it matters**: Keeping dead data in the tuple makes readers wonder whether `FirstMessage` could still be used after the group becomes satisfied.

src/Adapter/MSTestAdapter.PlatformServices/Helpers/AttributeHelpers.cs
```diff
@@ -9,27 +9,58 @@ internal static class AttributeExtensions
...
+        // Slow path: manual grouping instead of LINQ GroupBy to avoid iterator/Lookup allocations
+        Dictionary<string, (bool Satisfied, string? FirstMessage)> groups = [];
+        foreach (Attribute attr in allAttributes)
+        {
+            if (attr is not ConditionBaseAttribute conditionAttr)
+            {
+                continue;
+            }
+
+            bool shouldRun = conditionAttr.Mode == ConditionMode.Include ? conditionAttr.IsConditionMet : !conditionAttr.IsConditionMet;
```
Evangelink, line 44: **[Coverage]** The manual Dictionary-based grouping logic here replaces the LINQ `GroupBy` and is the most semantically complex part of this refactoring. There are currently no unit tests for `AttributeExtensions.IsIgnored` anywhere in the test suite — the only coverage comes from integration/acceptance tests, which don't directly exercise individual grouping combinations.

**Impact**: If a subtle bug exists in the new grouping logic (e.g., multi-attribute same-group scenarios, the `FirstMessage` tracking across null/non-null `IgnoreMessage` values, or multi-group "all-must-be-satisfied" semantics), it would be missed until an integration test hit it — with a much harder debugging experience.

**Suggestion**: Add unit tests in `test/UnitTests/MSTestAdapter.PlatformServices.UnitTests/Helpers/AttributeHelpersTests.cs` covering at minimum:
- No `ConditionBaseAttribute` present → `(false, null)`
- Single attribute, satisfied / unsatisfied
- Two attributes in the **same** group: first unsatisfied then satisfied (→ not ignored)
- Two attributes in the same group: both unsatisfied (→ ignored, first non-null message returned)
- Two attributes in **different** groups, one unsatisfied (→ ignored)
- `IgnoreMessage` precedence: first non-null message is returned when multiple unsatisfied entries have different messages

Evangelink (commented):
### Summary

**Date**: 2026-05-11

### Key Findings

1. **[Coverage — `IsIgnored` grouping logic]** The most impactful change in this PR is the replacement of LINQ `GroupBy` with a manual `Dictionary<string, (bool Satisfied, string? FirstMessage)>` grouping in `AttributeExtensions.IsIgnored`. The new logic is non-trivial (handling first-message tracking across null/non-null `IgnoreMessage` values, multi-group cross-cutting semantics, and early-exit once a group is satisfied). There are **no unit tests** for `AttributeExtensions.IsIgnored` in the entire test suite — only acceptance/integration tests provide indirect coverage. The PR is a pure performance refactoring with no test additions to validate the preserved semantics.

### Recommendations

- Add a `AttributeHelpersTests.cs` unit test file covering the grouping semantics of `IsIgnored`: single group (all unsatisfied, at least one satisfied), multiple groups, `IgnoreMessage` null/non-null first-message tracking. This protects against any future regression of the refactored path.

---

*Generated by [Test Expert Reviewer](https://github.com/microsoft/testfx/actions/runs/25660867264)*

> 🧪 *Test quality reviewed by [Test Expert Reviewer 🧪](https://github.com/microsoft/testfx/actions/runs/25660867264)*

Evangelink (commented):
### Summary

**Date**: 2026-05-11

### Analysis

Reviewed three code-level changes in this PR across `AttributeHelpers.cs`, `TestMethodInfo.cs`, and `TestMethodRunner.cs`.

#### `AttributeHelpers.IsIgnored` (main change)

The refactor replaces `LINQ GroupBy` + yield-iterator with a two-pass manual scan over the already-cached `Attribute[]`:

1. **Fast path** — a single scan to detect any `ConditionBaseAttribute`. If none is found, returns immediately with zero allocations. Correct: matches the old `GetAttributes<ConditionBaseAttribute>()` returning empty.
2. **Slow path** — a `Dictionary<string,(bool,string?)>` replaces the `Lookup<string,ConditionBaseAttribute>` that `GroupBy` produced.

I traced the grouping semantics in detail against the original code, including edge cases:

- **First attribute in group is satisfied, rest are not**: both paths mark the group satisfied ✓
- **All attributes unsatisfied, first has null message**: new code correctly picks up the first non-null message via `groupState.FirstMessage is null` guard ✓
- **Attribute interleaving across multiple groups**: Dictionary insertion-order is respected (same as LINQ GroupBy source order), so the returned `ignoreMessage` is identical ✓

#### `TestMethodInfo.GetRetryAttribute`

Direct iteration over `GetCustomAttributesCached()` replaces the yield-iterator path. Duplicate-detection (`ThrowMultipleAttributesException` when `found is not null`) is structurally equivalent to the old `enumerator.MoveNext()` second-check. The `[DoesNotReturn]` method makes it impossible to accidentally reach `found = retryAttr` after a throw. ✓

#### `TestMethodRunner.TryExecuteDataSourceBasedTestsAsync`

The old code checked `dataSourceAttribute is { Length: 1 }`, implying exactly-one semantics. The new code uses `IsAttributeDefined<DataSourceAttribute>`, skipping the array allocation. This is safe because `DataSourceAttribute` is `sealed` and declared with `[AttributeUsage(AllowMultiple = false, Inherited = false)]`, making the "at most one" invariant enforced at compile time and unavoidable at runtime. ✓

### Positive Observations

- The fast-path scan is a textbook hot-path optimization: zero allocations in the overwhelmingly common case where no condition attributes are present.
- `GetCustomAttributesCached` is used consistently across all three change sites, keeping reflection cost amortised.
- `IsAttributeDefined` is already a well-established helper in `ReflectHelper`; repurposing it here is idiomatic.

### Conclusion

No correctness, threading, performance regression, API compatibility, or resource management issues were found. The changes deliver the stated allocation reduction with semantically equivalent behavior.

---

*Generated by [Expert Code Reviewer](https://github.com/microsoft/testfx/actions/runs/25660867324)*

> 🧠 *Reviewed by [Expert Code Reviewer 🧠](https://github.com/microsoft/testfx/actions/runs/25660867324)*

Evangelink:
@copilot address review comments

Evangelink:

PR #8128, opened 2026-05-11
[Test Improver] Add unit tests for ExtensionValidationHelper.ValidateUniqueExtension

Evangelink:
&lt;empty&gt;

### DIFF
--- a/test/UnitTests/Microsoft.Testing.Platform.UnitTests/Helpers/ExtensionValidationHelperTests.cs
+++ b/test/UnitTests/Microsoft.Testing.Platform.UnitTests/Helpers/ExtensionValidationHelperTests.cs
@@ -1,0 +1,229 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using Microsoft.Testing.Platform.Helpers;
+
+namespace Microsoft.Testing.Platform.UnitTests;
+
+[TestClass]
+public sealed class ExtensionValidationHelperTests
+{
+    // Generic overload: ValidateUniqueExtension<T>(IEnumerable<T>, IExtension, Func<T, IExtension>)
+    [TestMethod]
+    public void ValidateUniqueExtension_WhenExistingExtensionsIsNull_ThrowsArgumentNullException()
+    {
+        IEnumerable<IExtension> existingExtensions = null!;
+        IExtension newExtension = CreateExtension("uid1");
+
+        Assert.ThrowsExactly<ArgumentNullException>(
+            () => existingExtensions.ValidateUniqueExtension(newExtension, x => x));
+    }
+
+    [TestMethod]
+    public void ValidateUniqueExtension_WhenNewExtensionIsNull_ThrowsArgumentNullException()
+    {
+        IExtension[] existingExtensions = [];
+        IExtension newExtension = null!;
+
+        Assert.ThrowsExactly<ArgumentNullException>(
+            () => existingExtensions.ValidateUniqueExtension(newExtension, x => x));
+    }
+
+    [TestMethod]
+    public void ValidateUniqueExtension_WhenExtensionSelectorIsNull_ThrowsArgumentNullException()
+    {
+        IExtension[] existingExtensions = [];
+        IExtension newExtension = CreateExtension("uid1");
+
+        Assert.ThrowsExactly<ArgumentNullException>(
+            () => existingExtensions.ValidateUniqueExtension<IExtension>(newExtension, null!));
+    }
+
+    [TestMethod]
+    public void ValidateUniqueExtension_WhenCollectionIsEmpty_DoesNotThrow()
+    {
+        IExtension[] existingExtensions = [];
+        IExtension newExtension = CreateExtension("uid1");
+
+        existingExtensions.ValidateUniqueExtension(newExtension, x => x);
+    }
+
+    [TestMethod]
+    public void ValidateUniqueExtension_WhenNoDuplicateUid_DoesNotThrow()
+    {
+        IExtension[] existingExtensions = [CreateExtension("uid-A"), CreateExtension("uid-B")];
+        IExtension newExtension = CreateExtension("uid-C");
+
+        existingExtensions.ValidateUniqueExtension(newExtension, x => x);
+    }
+
+    [TestMethod]
+    public void ValidateUniqueExtension_WhenDuplicateUidExists_ThrowsInvalidOperationException()
+    {
+        IExtension[] existingExtensions = [CreateExtension("uid-dup")];
+        IExtension newExtension = CreateExtension("uid-dup");
+
+        Assert.ThrowsExactly<InvalidOperationException>(
+            () => existingExtensions.ValidateUniqueExtension(newExtension, x => x));
+    }
+
+    [TestMethod]
+    public void ValidateUniqueExtension_WhenDuplicateUidExists_ErrorMessageContainsUid()
+    {
+        const string duplicateUid = "my-duplicate-uid";
+        IExtension[] existingExtensions = [CreateExtension(duplicateUid)];
+        IExtension newExtension = CreateExtension(duplicateUid);
+
+        InvalidOperationException ex = Assert.ThrowsExactly<InvalidOperationException>(
+            () => existingExtensions.ValidateUniqueExtension(newExtension, x => x));
+
+        Assert.Contains(duplicateUid, ex.Message);
+    }
+
+    [TestMethod]
+    public void ValidateUniqueExtension_WhenDuplicateUidExists_ErrorMessageContainsTypeNames()
+    {
+        const string duplicateUid = "my-duplicate-uid";
+        IExtension existing = CreateExtension(duplicateUid);
+        IExtension newExtension = CreateExtension(duplicateUid);
+
+        InvalidOperationException ex = Assert.ThrowsExactly<InvalidOperationException>(
+            () => new[] { existing }.ValidateUniqueExtension(newExtension, x => x));
+
+        Assert.Contains(existing.GetType().ToString(), ex.Message);
+    }
+
+    [TestMethod]
+    public void ValidateUniqueExtension_WhenMultipleDuplicatesExist_ErrorMessageContainsAllTypes()
+    {
+        const string duplicateUid = "shared-uid";
+        FakeExtensionA existing1 = new(duplicateUid);
+        FakeExtensionB existing2 = new(duplicateUid);
+        FakeExtensionC newExtension = new(duplicateUid);
+
+        InvalidOperationException ex = Assert.ThrowsExactly<InvalidOperationException>(
+            () => new IExtension[] { existing1, existing2 }.ValidateUniqueExtension(newExtension, x => x));
+
+        Assert.Contains(typeof(FakeExtensionA).ToString(), ex.Message);
+        Assert.Contains(typeof(FakeExtensionB).ToString(), ex.Message);
+        Assert.Contains(typeof(FakeExtensionC).ToString(), ex.Message);
+    }
+
+    [TestMethod]
+    public void ValidateUniqueExtension_WithWrapperType_SelectsExtensionViaSelector()
+    {
+        const string duplicateUid = "wrapper-uid";
+        ExtensionWrapper existing = new(CreateExtension(duplicateUid));
+        IExtension newExtension = CreateExtension(duplicateUid);
+
+        Assert.ThrowsExactly<InvalidOperationException>(
+            () => new[] { existing }.ValidateUniqueExtension(newExtension, w => w.Extension));
+    }
+
+    [TestMethod]
+    public void ValidateUniqueExtension_WithWrapperType_WhenNoDuplicate_DoesNotThrow()
+    {
+        ExtensionWrapper existing = new(CreateExtension("uid-X"));
+        IExtension newExtension = CreateExtension("uid-Y");
+
+        new[] { existing }.ValidateUniqueExtension(newExtension, w => w.Extension);
+    }
+
+    // Simple overload: ValidateUniqueExtension(IEnumerable<IExtension>, IExtension)
+    [TestMethod]
+    public void ValidateUniqueExtension_SimpleOverload_WhenExistingExtensionsIsNull_ThrowsArgumentNullException()
+    {
+        IEnumerable<IExtension> existingExtensions = null!;
+        IExtension newExtension = CreateExtension("uid1");
+
+        Assert.ThrowsExactly<ArgumentNullException>(
+            () => existingExtensions.ValidateUniqueExtension(newExtension));
+    }
+
+    [TestMethod]
+    public void ValidateUniqueExtension_SimpleOverload_WhenNewExtensionIsNull_ThrowsArgumentNullException()
+    {
+        IExtension[] existingExtensions = [];
+
+        Assert.ThrowsExactly<ArgumentNullException>(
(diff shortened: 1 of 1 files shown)


test/UnitTests/Microsoft.Testing.Platform.UnitTests/Helpers/ExtensionValidationHelperTests.cs
```diff
@@ -0,0 +1,229 @@
...
+        public string Uid => uid;
+
+        public string Version => "1.0";
+
+        public string DisplayName => uid;
+
+        public string Description => uid;
+
+        public Task<bool> IsEnabledAsync() => Task.FromResult(true);
+    }
```
Evangelink: **Nitpick (Important)**: `FakeExtension`, `FakeExtensionA`, `FakeExtensionB`, and `FakeExtensionC` (lines 176–226) have identical implementations — only their class names differ. The duplication could be reduced by extracting a shared base class or abstract helper:

```csharp
private abstract class FakeExtensionBase(string uid) : IExtension
{
    public string Uid => uid;
    public string Version => "1.0";
    public string DisplayName => uid;
    public string Description => uid;
}

private sealed class FakeExtension(string uid)  : FakeExtensionBase(uid);
private sealed class FakeExtensionA(string uid) : FakeExtensionBase(uid);
private sealed class FakeExtensionB(string uid) : FakeExtensionBase(uid);
private sealed class FakeExtensionC(string uid) : FakeExtensionBase(uid);
```

**Why it matters**: DRY maintenance — any future change to the `IExtension` contract (e.g. a new member) currently requires editing four identical classes.

test/UnitTests/Microsoft.Testing.Platform.UnitTests/Helpers/ExtensionValidationHelperTests.cs
```diff
@@ -0,0 +1,229 @@
...
+    {
+        const string duplicateUid = "my-duplicate-uid";
+        IExtension existing = CreateExtension(duplicateUid);
+        IExtension newExtension = CreateExtension(duplicateUid);
+
+        InvalidOperationException ex = Assert.ThrowsExactly<InvalidOperationException>(
+            () => new[] { existing }.ValidateUniqueExtension(newExtension, x => x));
+
+        Assert.Contains(existing.GetType().ToString(), ex.Message);
+    }
```
Evangelink, line 85: **Nitpick (Important)**: Both `existing` and `newExtension` are created via `CreateExtension`, so they are both `FakeExtension`. The single `Assert.Contains(existing.GetType().ToString(), ...)` assertion therefore passes regardless of whether the new extension's type is included in the error message — because both share the same type name.

The source code does include `newExtension` in the `allDuplicates` list, so the intent is to verify both. Consider using distinct types for `existing` vs `newExtension` (e.g. `FakeExtensionA` / `FakeExtensionB`) and asserting both are present, similar to `ValidateUniqueExtension_WhenMultipleDuplicatesExist_ErrorMessageContainsAllTypes`.

**Why it matters**: The test currently doesn't actually exercise the "new extension's type is in the message" path.

test/UnitTests/Microsoft.Testing.Platform.UnitTests/Helpers/ExtensionValidationHelperTests.cs
```diff
@@ -0,0 +1,229 @@
...
+
+    [TestMethod]
+    public void ValidateUniqueExtension_WhenDuplicateUidExists_ErrorMessageContainsTypeNames()
+    {
+        const string duplicateUid = "my-duplicate-uid";
+        IExtension existing = CreateExtension(duplicateUid);
+        IExtension newExtension = CreateExtension(duplicateUid);
+
+        InvalidOperationException ex = Assert.ThrowsExactly<InvalidOperationException>(
+            () => new[] { existing }.ValidateUniqueExtension(newExtension, x => x));
```
Evangelink: **Nitpick (Minor)**: Several places use `new[] { ... }` to build arrays (lines 91, 105, 120, 129) while most other tests in this file use C# collection expressions (`[]`, `[a, b]`). For consistency, prefer collection expressions:

```csharp
// line 91
() => [existing].ValidateUniqueExtension(newExtension, x => x));

// line 105
() => (new IExtension[] { existing1, existing2 }).ValidateUniqueExtension(newExtension, x => x));
// → [(IExtension)existing1, existing2].ValidateUniqueExtension(...)
```

**Why it matters**: Keeps the file's initialization style consistent and reads more naturally alongside the collection-expression patterns already used in the same file.

test/UnitTests/Microsoft.Testing.Platform.UnitTests/Helpers/ExtensionValidationHelperTests.cs
```diff
@@ -0,0 +1,229 @@
...
+
+        Assert.Contains(existing.GetType().ToString(), ex.Message);
```
Evangelink: **Nitpick (Minor)**: `existing.GetType().ToString()` (runtime expression) is inconsistent with `typeof(FakeExtensionA).ToString()` used in the adjacent test (lines 107-109). Since the concrete type is known statically, prefer the static form for symmetry:

```csharp
Assert.Contains(typeof(FakeExtension).ToString(), ex.Message);
```

**Why it matters**: Consistent use of `typeof(T)` makes the intent explicit and avoids accidental reliance on the runtime type of a polymorphic variable.

test/UnitTests/Microsoft.Testing.Platform.UnitTests/Helpers/ExtensionValidationHelperTests.cs
```diff
@@ -0,0 +1,229 @@
...
+        IExtension[] existingExtensions = [CreateExtension("uid-A"), CreateExtension("uid-B")];
+        IExtension newExtension = CreateExtension("uid-C");
+
+        existingExtensions.ValidateUniqueExtension(newExtension);
+    }
+
+    [TestMethod]
+    public void ValidateUniqueExtension_SimpleOverload_WhenDuplicateUidExists_ThrowsInvalidOperationException()
+    {
+        const string duplicateUid = "duplicated";
```
Evangelink: **Nitpick (Minor)**: The duplicate-UID string for the simple-overload test is `"duplicated"` (line 164) while the equivalent generic-overload tests use `"my-duplicate-uid"` (lines 73, 86). The actual value doesn't matter functionally, but keeping a consistent sentinel (e.g. always `"uid-dup"` or always `"my-duplicate-uid"`) makes the tests easier to scan at a glance.

**Why it matters**: Small naming inconsistencies accumulate and make test suites harder to skim when looking for related cases.

Evangelink (commented):
### Summary

**Date**: 2026-05-11

### Key Findings

1. **[Important] Duplicated fake-extension class bodies** — `FakeExtension`, `FakeExtensionA/B/C` have identical `IExtension` implementations; only the class name differs. A shared base class would eliminate the duplication and make future `IExtension` contract changes a single-site edit.

2. **[Important] Weak type-name assertion in `...ErrorMessageContainsTypeNames`** — both `existing` and `newExtension` are `FakeExtension`, so the single `Assert.Contains(existing.GetType().ToString(), ...)` passes even if the new extension's type is absent from the error message. Using distinct types for the two variables would actually exercise the full `allDuplicates` list the source builds.

3. **[Minor] Mixed collection-creation styles** — `new[] { ... }` / `new IExtension[] { ... }` sit alongside collection expressions (`[]`, `[a, b]`). Prefer collection expressions throughout for consistency.

4. **[Minor] `existing.GetType().ToString()` vs `typeof(T).ToString()`** — adjacent tests use the static `typeof` form; this one uses a runtime expression on a variable whose static type is known.

5. **[Minor] Inconsistent duplicate-UID sentinel strings** — `"my-duplicate-uid"` (generic overload tests) vs `"duplicated"` (simple overload). No functional difference, but a consistent string makes related test cases easier to scan together.

### Positive Highlights

- Clear, well-named test methods following `MethodUnderTest_Scenario_ExpectedResult` convention throughout.
- Good separation of generic and simple-overload test groups with section comments.
- Complete coverage of null-guard paths for all three parameters of the generic overload.
- Nice use of the `ExtensionWrapper` record to exercise the selector overload — a non-obvious but important test scenario.

---

*Generated by [PR Nitpick Reviewer](https://github.com/microsoft/testfx/actions/runs/25677266600)*

> 🔍 *Meticulously inspected by [PR Nitpick Reviewer 🔍](https://github.com/microsoft/testfx/actions/runs/25677266600)*

test/UnitTests/Microsoft.Testing.Platform.UnitTests/Helpers/ExtensionValidationHelperTests.cs
```diff
@@ -0,0 +1,229 @@
...
+        IExtension newExtension = CreateExtension("uid1");
+
+        existingExtensions.ValidateUniqueExtension(newExtension, x => x);
+    }
+
+    [TestMethod]
+    public void ValidateUniqueExtension_WhenNoDuplicateUid_DoesNotThrow()
+    {
+        IExtension[] existingExtensions = [CreateExtension("uid-A"), CreateExtension("uid-B")];
+        IExtension newExtension = CreateExtension("uid-C");
```
Evangelink, line 55: **[Coverage]** The production code uses `extensionSelector(x).Uid == newExtension.Uid` for comparison, which is ordinal/case-sensitive for strings. There is no test documenting whether `"uid-a"` and `"uid-A"` are treated as duplicates or not.

**Impact**: Undefined/untested contract — if the comparison semantics ever change (e.g., to `OrdinalIgnoreCase`), no test would catch the regression, and callers have no documented behaviour to rely on.

**Suggestion**: Add a test such as:
```csharp
[TestMethod]
public void ValidateUniqueExtension_WhenUidsDifferOnlyByCase_DoesNotThrow()
{
    IExtension[] existingExtensions = [CreateExtension("uid-A")];
    IExtension newExtension = CreateExtension("uid-a");

    // UID comparison is case-sensitive; "uid-a" != "uid-A"
    existingExtensions.ValidateUniqueExtension(newExtension, x => x);
}
```

test/UnitTests/Microsoft.Testing.Platform.UnitTests/Helpers/ExtensionValidationHelperTests.cs
```diff
@@ -0,0 +1,229 @@
...
+
+        existingExtensions.ValidateUniqueExtension(newExtension, x => x);
+    }
+
+    [TestMethod]
+    public void ValidateUniqueExtension_WhenDuplicateUidExists_ThrowsInvalidOperationException()
```
Evangelink: **[Structure]** `ValidateUniqueExtension_WhenDuplicateUidExists_ThrowsInvalidOperationException` is a strict subset of `ValidateUniqueExtension_WhenDuplicateUidExists_ErrorMessageContainsUid` (lines 70–81), which already asserts `ThrowsExactly<InvalidOperationException>` and additionally verifies the message content.

**Impact**: Dead test weight — adds maintenance overhead with no additional coverage.

**Suggestion**: Remove this test; the exception-type assertion is already covered by the `_ErrorMessageContainsUid` test below.

test/UnitTests/Microsoft.Testing.Platform.UnitTests/Helpers/ExtensionValidationHelperTests.cs
Evangelink: **[Assertion]** Both `existing` and `newExtension` are created with `CreateExtension(...)`, so they are both `FakeExtension` instances. The assertion `Assert.Contains(existing.GetType().ToString(), ex.Message)` would pass even if the production code omitted the new extension's type from the error message, since the same type name covers both.

**Impact**: False positive — the test won't catch a regression where the new extension's type is dropped from the error message.

**Suggestion**: Use two different concrete types so the assertion can independently verify each type appears. For example, use `FakeExtensionA` for `existing` and `FakeExtensionB` (or `FakeExtension`) for `newExtension`, then assert both type names are present. The `_WhenMultipleDuplicatesExist_ErrorMessageContainsAllTypes` test already does this pattern correctly — this test should follow the same approach.

Evangelink (commented):
### Summary

**Date**: 2026-05-11

### Key Findings

The tests are well-structured, fully isolated, and cover the happy paths and null-guard contracts for both overloads of `ValidateUniqueExtension`. No flakiness risks or shared-state issues were found.

Three improvements worth considering:

1. **[Assertion — Weak]** `_ErrorMessageContainsTypeNames` creates both `existing` and `newExtension` as the same type (`FakeExtension`), so the single `Assert.Contains(existing.GetType().ToString(), ...)` assertion doesn't independently verify the new extension's type in the message. A regression that drops the new type would go undetected.

2. **[Structure — Redundant]** `_WhenDuplicateUidExists_ThrowsInvalidOperationException` is a strict subset of `_WhenDuplicateUidExists_ErrorMessageContainsUid`, which already asserts the exception type. The first test can be removed.

3. **[Coverage — Missing edge case]** No test documents whether UID comparison is case-sensitive. The production code uses `==` (ordinal, case-sensitive), but this is unverified and undocumented in tests.

### Recommendations

- Strengthen `_ErrorMessageContainsTypeNames` by using distinct types for existing vs. new extension (similar to `_WhenMultipleDuplicatesExist_ErrorMessageContainsAllTypes`).
- Remove the redundant `_ThrowsInvalidOperationException` test.
- Add a case-sensitivity test for UID comparison to document the contract.

---

*Generated by [Test Expert Reviewer](https://github.com/microsoft/testfx/actions/runs/25677268912)*

> 🧪 *Test quality reviewed by [Test Expert Reviewer 🧪](https://github.com/microsoft/testfx/actions/runs/25677268912)*

Evangelink:
@copilot address review comments

Evangelink (commented):
### Summary

**Workflow**: Expert Code Reviewer 🧠
**Date**: 2026-05-11

### Key Findings

No issues found across all review categories:

- **Correctness**: All test assertions are logically correct and match the expected behavior of `ExtensionValidationHelper.ValidateUniqueExtension`. The `Assert.Contains(substring, container)` argument order follows the project convention.
- **Threading**: Tests are synchronous and self-contained; no shared mutable state.
- **Performance**: Test-only code; no hot-path concerns.
- **API Compat**: No public API surface changes.
- **Cross-TFM**: No TFM-specific APIs used.
- **Resources**: No disposable resources involved.
- **Security / IPC**: Not applicable to test helpers.
- **Defensive coding**: Null guard tests correctly exercise the three `ArgumentNullException` paths in the implementation.

### Positive Observations

- Tests cover both overloads (generic selector and simple `IEnumerable<IExtension>`).
- Error-message content is verified for UID and all involved type names, including the multi-duplicate case.
- The `ExtensionWrapper` test correctly exercises the selector indirection path.
- Assertion style is consistent with the rest of `Microsoft.Testing.Platform.UnitTests` (MSTest `Assert.ThrowsExactly` / `Assert.Contains`; no AwesomeAssertions).

### Recommendations

None required.

---

*Generated by [Expert Code Reviewer](https://github.com/microsoft/testfx/actions/runs/25677269024)*

> 🧠 *Reviewed by [Expert Code Reviewer 🧠](https://github.com/microsoft/testfx/actions/runs/25677269024)*

test/UnitTests/Microsoft.Testing.Platform.UnitTests/Helpers/ExtensionValidationHelperTests.cs
Evangelink: @copilot address review comment

PR #8151, opened 2026-05-12
Update agentic workflows

Evangelink:

### DIFF
--- a/.github/aw/actions-lock.json
+++ b/.github/aw/actions-lock.json
@@ -20,15 +20,15 @@
       "version": "v8.0.1",
       "sha": "3e5f45b2cfb9172054b4087a40e8e0b5a5461e7c"
     },
-    "actions/github-script@v8": {
-      "repo": "actions/github-script",
-      "version": "v8",
-      "sha": "ed597411d8f924073f98dfc5c65a23a2325f34cd"
-    },
     "actions/github-script@v9": {
       "repo": "actions/github-script",
       "version": "v9",
       "sha": "373c709c69115d41ff229c7e5df9f8788daa9553"
+    },
+    "actions/github-script@v9.0.0": {
+      "repo": "actions/github-script",
+      "version": "v9.0.0",
+      "sha": "d746ffe35508b1917358783b479e04febd2b8f71"
     },
     "actions/setup-python@v6.2.0": {
       "repo": "actions/setup-python",
@@ -40,10 +40,10 @@
       "version": "v7.0.1",
       "sha": "043fb46d1a93c77aae656e7c1c64a875d1fc6a0a"
     },
-    "github/codeql-action/upload-sarif@v4.35.1": {
+    "github/codeql-action/upload-sarif@v4.35.4": {
       "repo": "github/codeql-action/upload-sarif",
-      "version": "v4.35.1",
-      "sha": "0e9f55954318745b37b7933c693bc093f7336125"
+      "version": "v4.35.4",
+      "sha": "3ce22a6e336a7fcc318bc58ae1986395bdc83ba7"
     },
     "github/gh-aw-actions/setup-cli@v0.68.3": {
       "repo": "github/gh-aw-actions/setup-cli",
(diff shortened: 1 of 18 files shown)


.github/aw/actions-lock.json
```diff
@@ -20,16 +20,16 @@
...
-    },
     "actions/github-script@v9": {
       "repo": "actions/github-script",
       "version": "v9",
       "sha": "373c709c69115d41ff229c7e5df9f8788daa9553"
     },
+    "actions/github-script@v9.0.0": {
+      "repo": "actions/github-script",
+      "version": "v9.0.0",
+      "sha": "d746ffe35508b1917358783b479e04febd2b8f71"
```
Evangelink, line 31: **Nitpick** (Minor): Two separate lock entries now coexist for `actions/github-script` at the same major version — `@v9` (SHA `373c709c...`) and `@v9.0.0` (SHA `d746ffe3...`) — pointing to different commits.

Workflows using `@v9` will be pinned to a different commit than those using `@v9.0.0`. If both entries are intentional (e.g., some workflows reference the mutable major tag and others reference the immutable patch tag), this is fine, but it could cause subtle inconsistencies when maintainers expect both to run the same code. It may be worth auditing which workflows use each form and consolidating to `@v9.0.0` (immutable) across the board.

**Why it matters**: Diverging SHAs for what appears to be "the same" action can lead to hard-to-debug behavioral differences between workflows.

.github/aw/actions-lock.json
```diff
@@ -55,10 +55,42 @@
...
+      "image": "ghcr.io/github/gh-aw-mcpg:v0.2.19",
+      "digest": "sha256:44d4d8de7e6c37aaea484eba489940c52df6a0b54078ddcbc9327592d5b3c3dd",
+      "pinned_image": "ghcr.io/github/gh-aw-mcpg:v0.2.19@sha256:44d4d8de7e6c37aaea484eba489940c52df6a0b54078ddcbc9327592d5b3c3dd"
+    },
+    "ghcr.io/github/github-mcp-server:v0.32.0": {
+      "image": "ghcr.io/github/github-mcp-server:v0.32.0",
+      "digest": "sha256:2763823c63bcca718ce53850a1d7fcf2f501ec84028394f1b63ce7e9f4f9be28",
+      "pinned_image": "ghcr.io/github/github-mcp-server:v0.32.0@sha256:2763823c63bcca718ce53850a1d7fcf2f501ec84028394f1b63ce7e9f4f9be28"
+    },
+    "node:lts-alpine": {
```
Evangelink, line 90: **Nitpick** (Minor): `node:lts-alpine` uses a mutable floating alias, unlike every other container entry in this file which uses explicit version tags (e.g., `gh-aw-firewall/agent:0.25.20`, `gh-aw-mcpg:v0.2.19`).

SHA pinning (`@sha256:d1b3b4da...`) does protect the current build from tag drift, but when it comes time to update the digest, reviewers may not notice that `lts-alpine` has shifted to a new major Node.js LTS version (e.g., from Node 22 to Node 24). Consider pinning to an explicit version tag like `node:22-alpine` to make version intent visible at a glance.

**Why it matters**: The tag is the human-readable signal of intent; a floating alias makes it harder to spot an unintentional major-version upgrade when digests are refreshed.

Evangelink (commented):
### Summary

**Date**: 2026-05-12

### Overview

This PR updates the agentic workflow lock files across 18 files. The key change is a **security improvement**: container images referenced in all `*.lock.yml` workflows are now pinned with SHA-256 digests, preventing silent tag mutation attacks. Additionally, `actions/github-script` is upgraded from `v8` to `v9`, `super-linter` moves from `v8.5.0` → `v8.6.0`, and `github/codeql-action` is bumped to `v4.35.4`.

✅ The SHA pinning of all container images is a meaningful security hardening — great change.

### Key Findings

1. **Minor** — Dual `actions/github-script` entries with different SHAs (`@v9` vs `@v9.0.0`) coexist in `actions-lock.json`. This is not an error, but may cause subtle behavioral differences depending on which tag syntax is used by individual workflows.

2. **Minor** — `node:lts-alpine` uses a mutable floating alias, unlike all other container images which use explicit version tags. SHA pinning protects the current build, but makes it easy to miss a major Node.js version bump on future digest refreshes.

### Recommendations

- Consider auditing which workflows reference `actions/github-script@v9` vs `@v9.0.0` and standardize on the immutable patch-version form.
- Consider replacing `node:lts-alpine` with an explicit tag like `node:22-alpine` to make the intended Node.js major version visible without inspecting the SHA.

---

*Generated by [PR Nitpick Reviewer 🔍](https://github.com/microsoft/testfx/actions/runs/25721367123)*

🔍 *Meticulously inspected by the nitpick reviewer workflow*

> 🔍 *Meticulously inspected by [PR Nitpick Reviewer 🔍](https://github.com/microsoft/testfx/actions/runs/25721367123)*

Issue #6156, opened 2025-07-24
Add explicit trigger in azure-pipelines.yml


Youssef1313:
See https://github.com/dotnet/dnceng/issues/6014

PR #8154, opened 2026-05-12
Use `trigger: none` instead of explicit branch triggers in public pipeline

Evangelink:
Relates to #6156

The public pipeline (`azure-pipelines.yml`) is intended to run only on PRs, not on branch pushes. PR #6159 added explicit branch triggers for `main` and `rel/*` to work around the dnceng requirement (dotnet/dnceng#6014) of having an explicit trigger section. However, this causes every non-fork PR to trigger the pipeline twice (once for the push, once for the PR).

Replace the explicit branch trigger with `trigger: none` which:
- Explicitly disables CI (push) triggers — satisfying the dnceng requirement
- Prevents duplicate builds for non-fork PRs
- Keeps PR triggers working as defined by the `pr:` section

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -1,12 +1,4 @@
-trigger:
-  branches:
-    include:
-    - main
-    - rel/*
-  paths:
-    exclude:
-      - .github/**
-      - .gitignore
+trigger: none
 
 # Branch(es) that trigger(s) build(s) on PR
 pr:


Youssef1313:
> However, this causes every non-fork PR to trigger the pipeline twice (once for the push, once for the PR).

I never seen that before. Can you elaborate with some example?

I also prefer this to be looked at by @chcosta.

Evangelink:
Copilot is usually pretty good with pipeline - I also don't see the build run twice for now (maybe this needs to be merged first)

PR #8174, opened 2026-05-13
Guard MarshalByRefObject, [Serializable], and InitializeLifetimeService behind #if NETFRAMEWORK

Evangelink:
## Summary

Audit and clean up `MarshalByRefObject` and `[Serializable]` usage across `src/` to ensure these .NET Framework remoting artifacts are properly scoped to `net462` only.

### `MarshalByRefObject` inheritance (10 classes)

Made all `MarshalByRefObject` base class inheritance conditional on `#if NETFRAMEWORK`:

**Classes with only MarshalByRefObject as base (6):**
- `AssemblyEnumerator`, `ReflectHelper`, `RecursiveDirectoryPath`, `UnitTestRunner`, `TypeCache`, `TestAssemblySettingsProvider`

**Classes implementing MarshalByRefObject + an interface (4):**
- `RemotingMessageLogger`, `NopTraceLogger`, `EqtTraceLogger`, `MTPTraceLogger`

### `InitializeLifetimeService` overrides (6 methods)

Wrapped all `InitializeLifetimeService()` overrides (including `[SecurityCritical]` and the previous `[Obsolete]` attributes) entirely in `#if NETFRAMEWORK`. On `net462`, the override returns `null` for infinite lifetime. On modern .NET, the override doesn't exist since the base class is no longer `MarshalByRefObject`.

Also conditionally included `using System.Security;` to avoid IDE0005 warnings.

### `[Serializable]` attribute (11 non-exception types)

Guarded `[Serializable]` with `#if NETFRAMEWORK` on data types that only needed it for `BinaryFormatter` marshaling across AppDomain boundaries:

- `RecursiveDirectoryPath`, `UnitTestElement`, `TestMethod`, `TestAssemblySettings`, `StackTraceInformation`, `MSTestSettings`, `AssemblyEnumerationResult`, `TestRunDirectories`, `DeploymentItem`, `DeploymentItemOriginType`, `TestResult`

Exception types (`AssertFailedException`, `UnitTestAssertException`, etc.) intentionally retain `[Serializable]` unconditionally — this is a standard .NET convention for exceptions.

### Already correctly guarded (no changes needed)

These were already behind `#if NETFRAMEWORK`:
- `AssemblyLoadWorker`, `StaticStateHelper` (in `AppDomainUtilities.cs`), `AssemblyResolver` (conditional base class)
- All `AppDomain.CreateDomain`/`Unload` calls in `AppDomainUtilities.cs`, `AppDomainWrapper.cs`, `IAppDomain.cs`, `TestSourceHost.cs`

## Result

On modern .NET (`net8.0`/`net9.0`), these classes are now plain types with no remoting or `BinaryFormatter` baggage. On `net462`, the full AppDomain marshaling support is preserved.

## Validation

- ✅ `dotnet build` net9.0 — 0 warnings, 0 errors
- ✅ `dotnet build` net462 — 0 warnings, 0 errors
- ✅ `dotnet build` MSTest.TestAdapter net9.0 — 0 warnings, 0 errors
- ✅ `dotnet build` TestFramework net9.0 — 0 warnings, 0 errors


src/TestFramework/TestFramework/Attributes/TestMethod/TestResult.cs
```diff
@@ -8,7 +8,9 @@ namespace Microsoft.VisualStudio.TestTools.UnitTesting;
 /// <summary>
 /// TestResult object to be returned to adapter.
 /// </summary>
+#if NETFRAMEWORK
 [Serializable]
+#endif
```
Youssef1313, line 13: Consider reverting and keep the changes only in Adapter, or at least have this under `#if NETFRAMEWORK || NETSTANDARD` to be safe. I would be worried if TestFramework netstandard assemblies end up getting loaded in a .NET Framework process where Serializable will be needed.

Under adapter, we no longer target .NET Standard so we are pretty safe there.

Youssef1313: Actually better to simply just revert in this case.

This is a public class and we might not know what users are doing with it.

Evangelink:

src/Adapter/MSTestAdapter.PlatformServices/MSTestSettings.cs
```diff
@@ -27,12 +29,14 @@ internal sealed partial class MSTestSettings
     /// <summary>
     /// Member variable for Adapter settings.
     /// </summary>
+#pragma warning disable IDE0032 // Use auto property — property uses lazy '??=' initialization, which cannot be expressed as an auto-property.
```
Youssef1313: This suppression is not correct. You can use `field` keyword.

src/TestFramework/TestFramework/Attributes/TestMethod/TestResult.cs
Evangelink: I'll move forward like this and see if someone complains - this would give us indication of some unknown scenario so we can improve our tests.

PR #8187, opened 2026-05-13
Apply structured assertion messages to IsTrue, IsFalse, IsNull, IsNotNull

Evangelink:
## Summary

Applies the structured assertion message infrastructure (RFC 012) to the boolean and null assertion methods: `IsTrue`, `IsFalse`, `IsNull`, and `IsNotNull`.

## Changes

- **Assert.IsTrue / Assert.IsFalse**: Use `StructuredAssertionMessage` with a summary stating the expected boolean (e.g. `Expected condition to be true.`), an evidence block showing the `actual:` value, and a call-site expression. Expected/actual metadata is also exposed via `WithExpectedAndActual` for runner consumption.
- **Assert.IsNull / Assert.IsNotNull**: Use `StructuredAssertionMessage` with a summary, an evidence block showing the `actual:` value, and a call-site expression. Expected/actual metadata is also exposed via `WithExpectedAndActual`.
- **Assert.cs**: Added `FormatCallSiteExpression` helper methods for formatting call-site expressions in structured messages.
- **Tests**: Updated existing tests to verify the new structured message format.

## Context

This builds on the structured assertion message infrastructure merged in the parent PR, extending it to the first set of assertion methods (boolean and null checks).


src/TestFramework/TestFramework/Assertions/Assert.IsNull.cs
```diff
@@ -152,14 +152,26 @@ public static void IsNull(object? value, string? message = "", [CallerArgumentEx
...
 
-    private static void ReportAssertIsNullFailed(string? message)
-        => ReportAssertFailed("Assert.IsNull", message);
+    private static void ReportAssertIsNullFailed(object? value, string? message, string valueExpression)
+    {
+        string actualValue = AssertionValueRenderer.RenderValue(value);
+        EvidenceBlock evidence = EvidenceBlock.Create()
+            .AddLine("actual:", actualValue);
+
+        StructuredAssertionMessage structured = new("Expected value to be null.");
```
Evangelink: THis text should be localized

src/TestFramework/TestFramework/Assertions/Assert.IsNull.cs
```diff
@@ -191,13 +203,20 @@ public static void IsNotNull([NotNull] object? value, string? message = "", [Cal
...
 
     private static bool IsNotNullFailing([NotNullWhen(false)] object? value) => value is null;
 
     [DoesNotReturn]
-    private static void ReportAssertIsNotNullFailed(string? message)
-        => ReportAssertFailed("Assert.IsNotNull", message);
+    private static void ReportAssertIsNotNullFailed(string? message, string valueExpression)
+    {
+        // RFC: IsNotNull omits the evidence block since actual is always null
+        StructuredAssertionMessage structured = new("Expected value to not be null.");
```
Evangelink: This should be localized

src/TestFramework/TestFramework/Assertions/Assert.IsTrue.cs
```diff
@@ -150,15 +152,27 @@ public static void IsTrue([DoesNotReturnIf(false)] bool? condition, string? mess
...
 
-    private static void ReportAssertIsTrueFailed(string? message)
-        => ReportAssertFailed("Assert.IsTrue", message);
+    private static void ReportAssertIsTrueFailed(bool? condition, string? message, string conditionExpression)
+    {
+        string actualValue = AssertionValueRenderer.RenderValue(condition);
+        EvidenceBlock evidence = EvidenceBlock.Create()
+            .AddLine("actual:", actualValue);
+
+        StructuredAssertionMessage structured = new("Expected condition to be true.");
```
Evangelink: This should be localized

src/TestFramework/TestFramework/Assertions/Assert.IsTrue.cs
```diff
@@ -188,14 +202,26 @@ public static void IsFalse([DoesNotReturnIf(true)] bool? condition, string? mess
...
     [DoesNotReturn]
-    private static void ReportAssertIsFalseFailed(string userMessage)
-        => ReportAssertFailed("Assert.IsFalse", userMessage);
+    private static void ReportAssertIsFalseFailed(bool? condition, string? message, string conditionExpression)
+    {
+        string actualValue = AssertionValueRenderer.RenderValue(condition);
+        EvidenceBlock evidence = EvidenceBlock.Create()
+            .AddLine("actual:", actualValue);
+
+        StructuredAssertionMessage structured = new("Expected condition to be false.");
```
Evangelink: This should be localized

Evangelink:
## Code Review: PR #8187 — Design-Level Concerns

### 🔴 Backward Compatibility (BLOCKING)

The assertion failure message format changes **completely and unconditionally**:

| | Before | After |
|---|---|---|
| Prefix | `Assert.IsNull failed.` | `Assertion failed.` |
| Structure | Single line | Multi-line with `\n\n` |
| Content | Optional user message appended | Structured blocks: summary / user-msg / evidence / call-site |

**Concrete breakages for existing consumers upgrading to this version:**
1. Any code doing `ex.Message.Contains("Assert.IsNull failed")` → now fails silently (returns `false`)
2. Any CI pipeline, test result parser, or log aggregator matching on the `"Assert.Xxx failed."` prefix → misses assertion failures
3. Snapshot / approval tests capturing `AssertFailedException.Message` exactly → all fail
4. Any `Regex` written against the old single-line format → broken by embedded `\n\n`

There is **no opt-in/opt-out mechanism** and RFC 012 (docs/RFCs/012-Structured-Assertion-Messages.md) is marked `- [ ] Approved in principle` (unchecked) — i.e., the RFC has not been formally approved yet.

**Recommendation:** Before merging, either:
- Gate the new format behind an opt-in (e.g., `.runsettings` key or `AssertionOptions.UseStructuredMessages = true`), defaulting to the legacy format, **or**
- Explicitly document this as a breaking change in the release notes and ensure RFC 012 is formally approved first

---

### i️ Localization Note (Known Tradeoff)

Strings like `"Expected value to be null."`, `"Expected condition to be true."`, `"actual:"` are hardcoded English literals, diverging from the `.resx` / `.xlf` pipeline used by all existing assertion messages. This appears to be an intentional design decision per RFC 012 (structured messages use hardcoded English). If localization of these messages is in scope for a future RFC, it should be tracked explicitly.



> Generated by [Expert Code Review (on open)](https://github.com/microsoft/testfx/actions/runs/25811954178/agentic_workflow) for issue #8187 · ● 100M · [◷](https://github.com/search?q=repo%3Amicrosoft%2Ftestfx+%22gh-aw-workflow-call-id%3A+microsoft%2Ftestfx%2Freview-on-open.agent%22&type=issues)

Evangelink:
**[MAJOR] Localization regression — structured message strings are hardcoded English**

The PR hardcodes four user-facing summary strings directly in C# source:

| String | File |
|--------|------|
| `"Expected value to be null."` | `Assert.IsNull.cs` |
| `"Expected value to not be null."` | `Assert.IsNull.cs` |
| `"Expected condition to be true."` | `Assert.IsTrue.cs` |
| `"Expected condition to be false."` | `Assert.IsTrue.cs` |

Plus the evidence label `"actual:"` (also hardcoded in both files).

The existing assertion messages (`"Assert.IsNull failed."` etc.) were routed through `FrameworkMessages.resx` and localized via satellite assemblies into every supported language. After this PR, a user running MSTest under `es-ES` or `fr-FR` will receive the English strings above instead of their previously-localized messages.

The repo guidelines explicitly require: _"Add a corresponding entry in the resource file (.resx). NEVER manually modify `*.xlf` files. Instead, build the project to automatically generate the corresponding `*.xlf` files."_

**Recommendation:** Either:
1. Add the five strings to `FrameworkMessages.resx` and reference them as `FrameworkMessages.*`, or
2. Explicitly document in RFC 012 that structured messages are intentionally English-only (for machine-parsability) and call this out as a known localization regression in the PR description — so translators can be notified and the decision is deliberate rather than accidental.



> Generated by [Expert Code Review (on open)](https://github.com/microsoft/testfx/actions/runs/25811954178/agentic_workflow) for issue #8187 · ● 100M · [◷](https://github.com/search?q=repo%3Amicrosoft%2Ftestfx+%22gh-aw-workflow-call-id%3A+microsoft%2Ftestfx%2Freview-on-open.agent%22&type=issues)

src/TestFramework/TestFramework/Assertions/Assert.IsTrue.cs
```diff
@@ -150,15 +152,27 @@ public static void IsTrue([DoesNotReturnIf(false)] bool? condition, string? mess
...
+    {
+        string actualValue = AssertionValueRenderer.RenderValue(condition);
+        EvidenceBlock evidence = EvidenceBlock.Create()
+            .AddLine("actual:", actualValue);
+
+        StructuredAssertionMessage structured = new("Expected condition to be true.");
+        structured.WithUserMessage(message);
+        structured.WithEvidence(evidence);
+        structured.WithExpectedAndActual(null, actualValue);
+        structured.WithCallSiteExpression(FormatCallSiteExpression("Assert.IsTrue", conditionExpression));
```
Evangelink: **[MODERATE] Algorithmic Correctness — Wrong `ExpectedText` for boolean assertions**

`WithExpectedAndActual(null, actualValue)` sets `AssertFailedException.ExpectedText = null`, which means IDE test explorers and third-party tooling reading that property will see "no expected value" — but `IsTrue` and `IsFalse` *do* have natural expected values (`true` and `false` respectively).

Per the `AssertFailedException.ExpectedText` docs, `null` is for assertions with *no natural expected value* (e.g. `Assert.Fail`). Boolean assertions are not in that category.

**Recommendation:** Replace `null` with the rendered expected value:
```csharp
// IsTrue
structured.WithExpectedAndActual(AssertionValueRenderer.RenderValue(true), actualValue);

// IsFalse  
structured.WithExpectedAndActual(AssertionValueRenderer.RenderValue(false), actualValue);
```

For `Assert.IsNull`, passing `null` as expected is defensible (the expected *is* null), but the same pattern should be reconsidered there too — consider `"null"` as a string.

src/TestFramework/TestFramework/Assertions/Assert.cs
```diff
@@ -176,6 +176,40 @@ internal static void ThrowAssertFailed(StructuredAssertionMessage structuredMess
...
+    /// </summary>
+    internal static string? FormatCallSiteExpression(string assertionMethodName, string expression)
+    {
+        if (string.IsNullOrWhiteSpace(expression))
+        {
+            return null;
+        }
+
+        // If expression contains newlines (multiline constant), replace with placeholder per RFC
+        return expression.Contains('\n') || expression.Contains('\r')
```
Evangelink: **[BLOCKING] Cross-TFM Compatibility — `string.Contains(char)` missing on `netstandard2.0` / `net462`**

`string.Contains(char)` was introduced in .NET Standard 2.1. The `TestFramework.csproj` targets `netstandard2.0` and `$(NetFrameworkMinimum)` (net462), both of which only have `string.Contains(string)`. The compiler will not implicitly widen `char` to `string`, so this produces **CS1061** on the older TFMs:

```csharp
// FAILS to compile on netstandard2.0 and net462
return expression.Contains('\n') || expression.Contains('\r')
    ? null
    : $"{assertionMethodName}({expression})";
```

No polyfill for this overload exists in `src/Polyfills/`.

**Fix:** Replace every `Contains('\n')` / `Contains('\r')` with the TFM-safe alternatives:
```csharp
return expression.IndexOf('\n') >= 0 || expression.IndexOf('\r') >= 0
    ? null
    : $"{assertionMethodName}({expression})";
```
`string.IndexOf(char)` is available on all targeted TFMs. The same fix must be applied to the two-parameter overload (lines ~207–208).

src/TestFramework/TestFramework/Assertions/Assert.IsTrue.cs
Evangelink: **[MAJOR] Algorithmic Correctness — `WithExpectedAndActual(null, ...)` misrepresents IsTrue/IsFalse semantics**

`Assert.IsTrue` expects the condition to be `true`; `Assert.IsFalse` expects it to be `false`. Passing `null` as the expected argument here is semantically wrong:

```csharp
// IsTrue — expected should be "true", not null
structured.WithExpectedAndActual(null, actualValue);

// IsFalse (same pattern) — expected should be "false", not null
structured.WithExpectedAndActual(null, actualValue);
```

`WithExpectedAndActual` sets `StructuredAssertionMessage.ExpectedText`, which flows directly into:
- `AssertFailedException.ExpectedText` (public API, used by IDE diff viewers)
- `exception.Data["assert.expected"]` (used by test adapters for structured failure reporting)

When `null` is passed, neither is populated, so IDE diff views show no expected value even though "true" / "false" is unambiguous.

**Fix:**
```csharp
// In ReportAssertIsTrueFailed:
structured.WithExpectedAndActual(AssertionValueRenderer.RenderValue(true), actualValue);

// In ReportAssertIsFalseFailed:
structured.WithExpectedAndActual(AssertionValueRenderer.RenderValue(false), actualValue);
```

`IsNull` correctly passes `null` as expected (the value really should be null), so that call site is fine.

src/TestFramework/TestFramework/Assertions/Assert.IsTrue.cs
```diff
@@ -188,14 +202,26 @@ public static void IsFalse([DoesNotReturnIf(true)] bool? condition, string? mess
...
 
     [DoesNotReturn]
-    private static void ReportAssertIsFalseFailed(string userMessage)
-        => ReportAssertFailed("Assert.IsFalse", userMessage);
+    private static void ReportAssertIsFalseFailed(bool? condition, string? message, string conditionExpression)
+    {
+        string actualValue = AssertionValueRenderer.RenderValue(condition);
+        EvidenceBlock evidence = EvidenceBlock.Create()
+            .AddLine("actual:", actualValue);
+
```
Evangelink, line 219: **[MAJOR] Algorithmic Correctness — same `WithExpectedAndActual(null, ...)` issue for `IsFalse`**

See the comment on `Assert.IsTrue.cs` line ~172 for full context. The same bug applies here: `Assert.IsFalse` expects the condition to be `false`, but `null` is passed as the expected value, leaving `AssertFailedException.ExpectedText` and `exception.Data["assert.expected"]` both unpopulated.

**Fix:**
```csharp
structured.WithExpectedAndActual(AssertionValueRenderer.RenderValue(false), actualValue);
```

src/TestFramework/TestFramework/Assertions/Assert.cs
```diff
@@ -176,6 +176,40 @@ internal static void ThrowAssertFailed(StructuredAssertionMessage structuredMess
...
+            ? null
+            : $"{assertionMethodName}({expression})";
+    }
+
+    /// <summary>
+    /// Formats a call-site expression for display at the bottom of a structured assertion message,
+    /// using two captured expressions.
+    /// </summary>
+    internal static string? FormatCallSiteExpression(string assertionMethodName, string expression1, string expression2)
```
Evangelink: **[MODERATE] Dead code — two-parameter `FormatCallSiteExpression` overload has zero callers**

This overload was introduced in this PR but is never called anywhere in the codebase:

```csharp
internal static string? FormatCallSiteExpression(string assertionMethodName, string expression1, string expression2)
```

No assertion method in this PR (or elsewhere in the current code) passes two expressions. Adding an `internal` method with no callers increases maintenance surface and will confuse future contributors.

**Recommendation:** Remove this overload. Re-add it in the PR that migrates `Assert.AreEqual` or another two-argument assertion to structured messages.

src/TestFramework/TestFramework/Assertions/Assert.cs
```diff
@@ -176,6 +176,40 @@ internal static void ThrowAssertFailed(StructuredAssertionMessage structuredMess
...
+    {
+        if (string.IsNullOrWhiteSpace(expression))
+        {
+            return null;
+        }
+
+        // If expression contains newlines (multiline constant), replace with placeholder per RFC
+        return expression.Contains('\n') || expression.Contains('\r')
+            ? null
+            : $"{assertionMethodName}({expression})";
```
Evangelink: **[BLOCKING] Cross-TFM Compatibility**

`string.Contains(char)` was introduced in .NET Core 2.1 / .NET Standard 2.1. It does **not** exist in `netstandard2.0` or `net462`, both of which this project targets (confirmed in `TestFramework.csproj`). This will cause a **compilation failure** on those TFMs.

The same issue appears on lines 209 and 210 in the two-parameter overload below.

**Recommendation:** Replace `expression.Contains('\n')` with `expression.IndexOf('\n') >= 0` (and same for `'\r'`). The `string.IndexOf(char)` overload is available on all target frameworks.

src/TestFramework/TestFramework/Assertions/Assert.IsTrue.cs
Evangelink: **[MAJOR] Algorithmic Correctness — Wrong `ExpectedText` for `IsTrue`**

`WithExpectedAndActual(null, actualValue)` sets `AssertFailedException.ExpectedText = null`. For `Assert.IsTrue`, the natural expected value is `true`. Passing `null` means IDE diff views (e.g. Visual Studio's Test Explorer) and tooling that reads `exception.ExpectedText` / `exception.Data["assert.expected"]` will see no expected value, making the failure message less actionable than it should be.

**Recommendation:** Pass the expected value explicitly:
```csharp
structured.WithExpectedAndActual(AssertionValueRenderer.RenderValue(true), actualValue);
```

src/TestFramework/TestFramework/Assertions/Assert.IsTrue.cs
```diff
@@ -188,14 +202,26 @@ public static void IsFalse([DoesNotReturnIf(true)] bool? condition, string? mess
...
+        StructuredAssertionMessage structured = new("Expected condition to be false.");
+        structured.WithUserMessage(message);
+        structured.WithEvidence(evidence);
+        structured.WithExpectedAndActual(null, actualValue);
+        structured.WithCallSiteExpression(FormatCallSiteExpression("Assert.IsFalse", conditionExpression));
```
Evangelink: **[MAJOR] Algorithmic Correctness — Wrong `ExpectedText` for `IsFalse`**

Same issue as in `ReportAssertIsTrueFailed`: `WithExpectedAndActual(null, actualValue)` leaves `AssertFailedException.ExpectedText` as `null`. For `Assert.IsFalse`, the expected value is `false`.

**Recommendation:**
```csharp
structured.WithExpectedAndActual(AssertionValueRenderer.RenderValue(false), actualValue);
```

src/TestFramework/TestFramework/Assertions/Assert.cs
Evangelink: **[BLOCKING] Cross-TFM Compatibility — `string.Contains(char)` unavailable on `netstandard2.0` / `net462`**

`expression.Contains('\n')` and `expression.Contains('\r')` use the `string.Contains(char)` overload, which was introduced in .NET Standard 2.1 / .NET Core 2.1. This project targets `netstandard2.0` and `$(NetFrameworkMinimum)` (net462), as confirmed in `TestFramework.csproj`. This will cause a **compile-time failure** on those TFMs.

The same issue appears in the two-argument overload of `FormatCallSiteExpression` on the lines below.

**Recommendation:** Replace with `IndexOf`, which is available on all target frameworks:
```csharp
return expression.IndexOf('\n') >= 0 || expression.IndexOf('\r') >= 0
    ? null
    : $"{assertionMethodName}({expression})";
```

Evangelink (commented):
## Review Summary

| # | Dimension | Verdict |
|---|-----------|---------|
| 1 | Algorithmic Correctness | 🟡 1 MODERATE |
| 2 | Threading & Concurrency | ✅ LGTM |
| 3 | Security & IPC | ✅ LGTM |
| 4 | Public API & Binary Compatibility | ✅ LGTM (struct field additions are safe in .NET) |
| 5 | Performance & Allocations | ✅ LGTM (all new alloc on failure path only) |
| 6 | Cross-TFM Compatibility | ✅ LGTM |
| 7 | Resource & IDisposable | ✅ LGTM |
| 8 | Defensive Coding | ✅ LGTM |
| 9 | Localization | ⚠️ Known tradeoff per RFC 012 (hardcoded English, intentional) |
| 10 | **Backward Compatibility** | 🔴 **BLOCKING** |
| 11 | Assertion Quality | ✅ LGTM |
| 12 | Test Completeness | ✅ LGTM |
| 13 | Data-Driven Test Coverage | ✅ LGTM |
| 14 | Code Structure | ✅ LGTM (minor duplication is acceptable) |
| 15 | Naming & Conventions | ✅ LGTM |
| 16 | Documentation Accuracy | ✅ LGTM |
| 17 | Scope & PR Discipline | ✅ LGTM |

✅ 15/17 dimensions clean.

### Blocking

- [ ] **Backward Compatibility**: The assertion failure message format changes completely and unconditionally — no opt-in, no feature flag, RFC 012 not yet formally approved. Consumers doing message matching (`Contains("Assert.IsNull failed")`), log parsers, or snapshot tests will break silently on upgrade. Needs either an opt-in mechanism or explicit RFC approval + release-notes breaking-change documentation.

### Moderate

- [ ] **`WithExpectedAndActual(null, actualValue)` in IsTrue/IsFalse**: Sets `AssertFailedException.ExpectedText = null`, incorrectly signaling "no expected value." The expected values are `true` and `false`. Fix: pass `AssertionValueRenderer.RenderValue(true/false)` as the first argument.

> Generated by [Expert Code Review (on open)](https://github.com/microsoft/testfx/actions/runs/25811954178/agentic_workflow) for issue #8187 · ● 100M

src/TestFramework/TestFramework/Assertions/Assert.IsTrue.cs
Evangelink: **[MAJOR] Algorithmic Correctness — `WithExpectedAndActual(null, ...)` for `IsTrue`**

`IsTrue` has a well-defined expected value: `true`. Passing `null` as the first argument means `AssertFailedException.ExpectedText` and `exception.Data["assert.expected"]` will both be `null`. IDE test runners (Visual Studio Test Explorer) and tooling that uses those properties to render diffs will show no expected value, making failures harder to diagnose.

**Recommendation:**
```csharp
structured.WithExpectedAndActual(AssertionValueRenderer.RenderValue(true), actualValue);
```

src/TestFramework/TestFramework/Assertions/Assert.IsTrue.cs
Evangelink: **[MAJOR] Algorithmic Correctness — `WithExpectedAndActual(null, ...)` for `IsFalse`**

Same issue as in `ReportAssertIsTrueFailed`: the natural expected value for `IsFalse` is `false`, not `null`. Passing `null` leaves `AssertFailedException.ExpectedText` unset.

**Recommendation:**
```csharp
structured.WithExpectedAndActual(AssertionValueRenderer.RenderValue(false), actualValue);
```

src/TestFramework/TestFramework/Assertions/Assert.cs
```diff
@@ -176,6 +176,40 @@ internal static void ThrowAssertFailed(StructuredAssertionMessage structuredMess
...
+    internal static string? FormatCallSiteExpression(string assertionMethodName, string expression)
+    {
+        if (string.IsNullOrWhiteSpace(expression))
+        {
+            return null;
+        }
+
+        // If expression contains newlines (multiline constant), replace with placeholder per RFC
+        return expression.Contains('\n') || expression.Contains('\r')
+            ? null
```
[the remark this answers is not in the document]
Evangelink: Fixed. `FormatCallSiteExpression` now accepts a `paramName` parameter and returns `Assert.X(<paramName>)` for multiline expressions instead of returning `null`.

src/TestFramework/TestFramework/Assertions/Assert.IsNull.cs
Evangelink: Fixed. Moved to `FrameworkMessages.IsNullFailedSummary`.

src/TestFramework/TestFramework/Assertions/Assert.IsNull.cs
Evangelink: Fixed. Moved to `FrameworkMessages.IsNotNullFailedSummary`.

src/TestFramework/TestFramework/Assertions/Assert.IsTrue.cs
Evangelink: Fixed. Moved to `FrameworkMessages.IsTrueFailedSummary`.

src/TestFramework/TestFramework/Assertions/Assert.IsTrue.cs
Evangelink: Fixed. Moved to `FrameworkMessages.IsFalseFailedSummary`.

src/TestFramework/TestFramework/Assertions/Assert.IsTrue.cs
Evangelink: Fixed. Now uses `AssertionValueRenderer.RenderValue(true)` / `RenderValue(false)` for expected values in IsTrue/IsFalse. For IsNull, now uses `AssertionValueRenderer.RenderValue(null)` so `ExpectedText` is properly populated in all cases.

src/TestFramework/TestFramework/Assertions/Assert.cs
Evangelink: Fixed. Replaced all `string.Contains(char)` calls with `IndexOf(char) >= 0` for cross-TFM compatibility. The two-parameter overload was also removed (per separate review comment), so only the single-parameter overload remains with this fix applied.

src/TestFramework/TestFramework/Assertions/Assert.IsTrue.cs
Evangelink: Fixed. Now uses `AssertionValueRenderer.RenderValue(true)` for IsTrue and `RenderValue(false)` for IsFalse so that `ExpectedText` and `exception.Data["assert.expected"]` are properly populated.

src/TestFramework/TestFramework/Assertions/Assert.IsTrue.cs
Evangelink: Fixed. Now uses `AssertionValueRenderer.RenderValue(false)` as the expected value.

src/TestFramework/TestFramework/Assertions/Assert.cs
Evangelink: Fixed. Removed the two-parameter overload. Will re-add when migrating `Assert.AreEqual` or other two-argument assertions.

src/TestFramework/TestFramework/Assertions/Assert.cs
Evangelink: Fixed. Replaced with `IndexOf(char) >= 0`.

src/TestFramework/TestFramework/Assertions/Assert.IsTrue.cs
Evangelink: Fixed. Now uses `AssertionValueRenderer.RenderValue(true)` as the expected value.

src/TestFramework/TestFramework/Assertions/Assert.IsTrue.cs
Evangelink: Fixed. Now uses `AssertionValueRenderer.RenderValue(false)` as the expected value.

src/TestFramework/TestFramework/Assertions/Assert.cs
Evangelink: Fixed. Replaced with `IndexOf(char) >= 0`. The two-parameter overload that also had this issue has been removed entirely.

src/TestFramework/TestFramework/Assertions/Assert.IsTrue.cs
Evangelink: Fixed. Now uses `AssertionValueRenderer.RenderValue(true)` as the expected value.

src/TestFramework/TestFramework/Assertions/Assert.IsTrue.cs
Evangelink: Fixed. Now uses `AssertionValueRenderer.RenderValue(false)` as the expected value.

src/TestFramework/TestFramework/Assertions/Assert.cs
```diff
@@ -176,6 +176,44 @@ internal static void ThrowAssertFailed(StructuredAssertionMessage structuredMess
...
+    /// Formats a call-site expression for display at the bottom of a structured assertion message,
+    /// using two captured expressions. When an expression contains newlines, it is replaced with a
+    /// <c>&lt;paramName&gt;</c> placeholder.
+    /// </summary>
+    internal static string? FormatCallSiteExpression(string assertionMethodName, string expression1, string paramName1, string expression2, string paramName2)
+    {
+        if (string.IsNullOrWhiteSpace(expression1) || string.IsNullOrWhiteSpace(expression2))
+        {
+            return null;
+        }
```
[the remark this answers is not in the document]
Evangelink: Moot — the two-parameter overload has been removed entirely per Evangelink's review comment (dead code with zero callers).

src/TestFramework/TestFramework/Assertions/Assert.IsNull.cs
```diff
@@ -191,13 +203,20 @@ public static void IsNotNull([NotNull] object? value, string? message = "", [Cal
...
     [DoesNotReturn]
-    private static void ReportAssertIsNotNullFailed(string? message)
-        => ReportAssertFailed("Assert.IsNotNull", message);
+    private static void ReportAssertIsNotNullFailed(string? message, string valueExpression)
+    {
+        // RFC: IsNotNull omits the evidence block since actual is always null
+        StructuredAssertionMessage structured = new(FrameworkMessages.IsNotNullFailedSummary);
+        structured.WithUserMessage(message);
+        structured.WithCallSiteExpression(FormatCallSiteExpression("Assert.IsNotNull", valueExpression, "value"));
+
```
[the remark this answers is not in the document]
Evangelink: Acknowledged. `value` is the parameter name of the public `IsNotNull` method, but it's not in scope in the private `ReportAssertIsNotNullFailed` method, so `nameof(value)` won't compile. Added a `#pragma warning disable CA1507` with a comment explaining this. Passing it from the public method would add parameter-passing complexity for a constant string that can't drift from the public API parameter name.

src/TestFramework/TestFramework/Assertions/Assert.IsNull.cs
```diff
@@ -191,13 +203,22 @@ public static void IsNotNull([NotNull] object? value, string? message = "", [Cal
...
-        => ReportAssertFailed("Assert.IsNotNull", message);
+    private static void ReportAssertIsNotNullFailed(string? message, string valueExpression)
+    {
+        // RFC: IsNotNull omits the evidence block since actual is always null
+        StructuredAssertionMessage structured = new(FrameworkMessages.IsNotNullFailedSummary);
+        structured.WithUserMessage(message);
+#pragma warning disable CA1507 // Use nameof - 'value' is the parameter name of the public IsNotNull method, not available in this scope
+        structured.WithCallSiteExpression(FormatCallSiteExpression("Assert.IsNotNull", valueExpression, "value"));
+#pragma warning restore CA1507
+
```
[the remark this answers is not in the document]
Evangelink: Fixed. `ReportAssertIsNotNullFailed` now calls `WithExpectedAndActual("not null", actualValue)` and adds an `actual: null` evidence block, so `ExpectedText`/`ActualText` and the structured `assert.expected`/`assert.actual` data are populated.

src/TestFramework/TestFramework/Assertions/Assert.IsNull.cs
```diff
@@ -191,13 +203,22 @@ public static void IsNotNull([NotNull] object? value, string? message = "", [Cal
...
-    private static void ReportAssertIsNotNullFailed(string? message)
-        => ReportAssertFailed("Assert.IsNotNull", message);
+    private static void ReportAssertIsNotNullFailed(string? message, string valueExpression)
+    {
+        // RFC: IsNotNull omits the evidence block since actual is always null
+        StructuredAssertionMessage structured = new(FrameworkMessages.IsNotNullFailedSummary);
+        structured.WithUserMessage(message);
+#pragma warning disable CA1507 // Use nameof - 'value' is the parameter name of the public IsNotNull method, not available in this scope
+        structured.WithCallSiteExpression(FormatCallSiteExpression("Assert.IsNotNull", valueExpression, "value"));
+#pragma warning restore CA1507
```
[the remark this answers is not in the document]
Evangelink: Fixed. Added a `paramName` parameter to `ReportAssertIsNotNullFailed`; the public `IsNotNull` overloads now pass `nameof(value)`. The `CA1507` pragma was removed.

src/TestFramework/TestFramework/Assertions/Assert.IsNull.cs
```diff
@@ -191,13 +203,22 @@ public static void IsNotNull([NotNull] object? value, string? message = "", [Cal
...
 
     private static bool IsNotNullFailing([NotNullWhen(false)] object? value) => value is null;
 
     [DoesNotReturn]
-    private static void ReportAssertIsNotNullFailed(string? message)
-        => ReportAssertFailed("Assert.IsNotNull", message);
+    private static void ReportAssertIsNotNullFailed(string? message, string valueExpression)
+    {
+        // RFC: IsNotNull omits the evidence block since actual is always null
+        StructuredAssertionMessage structured = new(FrameworkMessages.IsNotNullFailedSummary);
```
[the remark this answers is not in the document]
Evangelink: Fixed. `ReportAssertIsNotNullFailed` now adds an `actual: null` evidence block via `WithEvidence(...)` so the failure output matches the PR description.

src/TestFramework/TestFramework/Assertions/Assert.IsNull.cs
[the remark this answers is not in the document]
Evangelink: Fixed. Added a `paramName` parameter to `ReportAssertIsNotNullFailed`; the public `IsNotNull` overloads now pass `nameof(value)`. The `CA1507` pragma was removed.

src/TestFramework/TestFramework/Assertions/Assert.IsNull.cs
```diff
@@ -152,14 +152,27 @@ public static void IsNull(object? value, string? message = "", [CallerArgumentEx
...
     }
 
     private static bool IsNullFailing(object? value) => value is not null;
 
-    private static void ReportAssertIsNullFailed(string? message)
-        => ReportAssertFailed("Assert.IsNull", message);
+    [DoesNotReturn]
+    private static void ReportAssertIsNullFailed(object? value, string? message, string valueExpression)
+    {
+        string actualValue = AssertionValueRenderer.RenderValue(value);
```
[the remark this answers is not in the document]
Evangelink: Acknowledged. `RenderValue` lives in the shared `AssertionValueRenderer` infrastructure and is used by all assertions that have already migrated to structured messages (e.g. `AreEqual`, `IsInstanceOfType`), not just `IsNull`. Adding a defensive cap for `IEnumerable` is a worthwhile improvement, but it belongs in `AssertionValueRenderer` itself so all assertions benefit and behavior stays consistent. Tracking as a separate follow-up rather than scoping it into this PR.

src/TestFramework/TestFramework/Assertions/Assert.IsNull.cs
```diff
@@ -191,13 +204,25 @@ public static void IsNotNull([NotNull] object? value, string? message = "", [Cal
...
+    private static void ReportAssertIsNotNullFailed(string? message, string valueExpression, string paramName)
+    {
+        string actualValue = AssertionValueRenderer.RenderValue(null);
+        EvidenceBlock evidence = EvidenceBlock.Create()
+            .AddLine("actual:", actualValue);
+
+        StructuredAssertionMessage structured = new(FrameworkMessages.IsNotNullFailedSummary);
+        structured.WithUserMessage(message);
+        structured.WithEvidence(evidence);
+        structured.WithExpectedAndActual("<not-null>", actualValue);
```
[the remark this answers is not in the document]
Evangelink: Fixed in 2c0b8ffbe. Reverted `ExpectedText` from `<not-null>` back to the plain string `"not null"` to avoid the markup confusion / TRX escaping issue you raised.

src/TestFramework/TestFramework/Assertions/Assert.IsTrue.cs
```diff
@@ -150,15 +152,28 @@ public static void IsTrue([DoesNotReturnIf(false)] bool? condition, string? mess
...
+    {
+        string actualValue = AssertionValueRenderer.RenderValue(condition);
+        EvidenceBlock evidence = EvidenceBlock.Create()
+            .AddLine("actual:", actualValue);
+
+        StructuredAssertionMessage structured = new(FrameworkMessages.IsTrueFailedSummary);
+        structured.WithUserMessage(message);
+        structured.WithEvidence(evidence);
+        structured.WithExpectedAndActual(AssertionValueRenderer.RenderValue(true), actualValue);
+        structured.WithCallSiteExpression(FormatCallSiteExpression("Assert.IsTrue", conditionExpression, nameof(condition)));
```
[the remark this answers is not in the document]
Evangelink: Updated the PR description in c50f7e7 (PR body edit) to accurately describe the implementation: only `actual:` appears in the evidence block; the expected value is conveyed by the summary line ("Expected condition to be true."/"...false.") and is also stored via `WithExpectedAndActual` for runner consumption (`AssertFailedException.ExpectedText` and `exception.Data["assert.expected"]`). Adding an `expected: true/false` line to the evidence block would be redundant with the summary, so we kept it minimal.

PR #8218, opened 2026-05-14
Revert "Guard server telemetry against duplicate result updates per test UID (#8192)"

Evangelink:
## Summary

Reverts #8192.

## Why

After the merge, @Youssef1313 pointed out (https://github.com/microsoft/testfx/pull/8192#discussion_r3240664374) that the deduplication guard introduces an asymmetric retry-counting semantic. Under #8192's guard, the retry count became:

| 1st result | 2nd result | retries counted |
|---|---|---|
| pass | pass | 0 |
| pass | fail | 1 |
| fail | fail | 0 |
| fail | pass | 1 |

The "retry" count then depends on outcome shape rather than on how many results were actually reported — particularly noticeable for "folded" tests where multiple sub-cases share one `TestNode.Uid`. The pre-#8192 behavior (every subsequent update for the same UID counts as a retry) is at least symmetric and consistent.

This revert restores that prior behavior. If retry/dedup semantics need to change, that should be a separate, design-led PR rather than the post-hoc guard from #8192.

## Validation

- `ServerDataConsumerServiceTests` runs green (8/8).
- No public API changes, no localization changes.

### DIFF
--- a/test/UnitTests/Microsoft.Testing.Platform.UnitTests/ServerMode/ServerDataConsumerServiceTests.cs
+++ b/test/UnitTests/Microsoft.Testing.Platform.UnitTests/ServerMode/ServerDataConsumerServiceTests.cs
@@ -133,12 +133,12 @@ public sealed class ServerDataConsumerServiceTests : IDisposable
         Assert.AreEqual(0, statistics.TotalDiscoveredTests);
         Assert.AreEqual(1, statistics.TotalPassedTests);
         Assert.AreEqual(0, statistics.TotalFailedTests);
-        Assert.AreEqual(0, statistics.TotalPassedRetries);
+        Assert.AreEqual(1, statistics.TotalPassedRetries);
         Assert.AreEqual(0, statistics.TotalFailedRetries);
     }
 
     [TestMethod]
-    public void PopulateTestNodeStatistics_WithOutcomeChangeForSameUid_CountsAsRetry()
+    public void PopulateTestNodeStatistics_WithEventsForSameUid()
     {
         PropertyBag properties = new(
             new FailedTestNodeStateProperty("failed"));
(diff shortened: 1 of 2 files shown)


test/UnitTests/Microsoft.Testing.Platform.UnitTests/ServerMode/ServerDataConsumerServiceTests.cs
```diff
@@ -133,12 +133,12 @@ public void PopulateTestNodeStatistics_WithDuplicatePassedEvents()
         Assert.AreEqual(0, statistics.TotalDiscoveredTests);
         Assert.AreEqual(1, statistics.TotalPassedTests);
         Assert.AreEqual(0, statistics.TotalFailedTests);
-        Assert.AreEqual(0, statistics.TotalPassedRetries);
+        Assert.AreEqual(1, statistics.TotalPassedRetries);
         Assert.AreEqual(0, statistics.TotalFailedRetries);
     }
```
Evangelink, line 138: The `PopulateTestNodeStatistics_WithDuplicatePassedEvents` test correctly reflects the restored symmetric behavior (duplicate pass → `TotalPassedRetries=1`). However, the deleted `PopulateTestNodeStatistics_WithDuplicateFailedEvents_DoesNotCountAsRetry` test had an expectation that no longer holds (it expected `TotalFailedRetries=0`), so its removal makes sense. But there is now **no test** that documents the symmetric duplicate-failed case: `fail + fail → TotalFailedRetries=1`. Consider adding one to make the symmetry explicit and guard against future regressions, e.g.:

```csharp
[TestMethod]
public void PopulateTestNodeStatistics_WithDuplicateFailedEvents()
{
    PropertyBag properties = new(new FailedTestNodeStateProperty("failed"));
    TestNodeUpdateMessage testNode = new(new SessionUid("1"), new TestNode { Uid = new TestNodeUid("test()"), DisplayName = string.Empty, Properties = properties });

    _service.PopulateTestNodeStatistics(testNode);
    _service.PopulateTestNodeStatistics(testNode);

    TestNodeStatistics statistics = _service.GetTestNodeStatistics();
    Assert.AreEqual(0, statistics.TotalPassedTests);
    Assert.AreEqual(1, statistics.TotalFailedTests);
    Assert.AreEqual(0, statistics.TotalPassedRetries);
    Assert.AreEqual(1, statistics.TotalFailedRetries);
}
```

Evangelink (commented):
## Review Summary

The revert is clean and correct. The root cause of the revert is valid: the guard `if (existingStatistics.HasPassed == hasPassed) return;` was asymmetric in the sense that `pass+pass` counted 0 retries while `fail+pass+pass` (after the outcome flip) would count only 1 retry — effectively swallowing the second duplicate event in the latter case but not having any equivalent deduplication for the failure case prior to the flip. Restoring symmetric behaviour (every subsequent event increments the corresponding retry counter) is the right call.

**Production code** — no issues. The removal of the 5-line guard is minimal and correct.

**Test coverage** — one minor gap: the symmetric counterpart to `WithDuplicatePassedEvents` (i.e., `fail+fail → TotalFailedRetries=1`) is now untested. The old test for that case was rightfully removed (it expected the now-wrong value of 0), but no replacement documents the new expected behaviour. See inline comment for a suggested test.

> Generated by [Expert Code Review (on open)](https://github.com/microsoft/testfx/actions/runs/25857097819/agentic_workflow) for issue #8218 · ● 3.8M

Youssef1313:
I agree with the revert here 👍 (I don't think the PR description is correct though - there is no *retry* involved here IMO)

YuliiaKovalova (approved):

PR #8233, opened 2026-05-14
Add MSTestParallelizeScope/Workers MSBuild properties (#4116)

### LINKED ISSUE #4116: Migrate "[assembly: Parallelize(Scope = ExecutionScope.MethodLevel)]" to SDK project file
# Current Situation

Currently, when a new MS Test SDK project is being created, an additional C# file is being generated ("`MSTestSettings.cs`") that's bloating the file list and may accidentally be deleted when using wildcards to delete a number of tests.

# Desired Situation

Please provide an SDK project property to have this setting available there and remove the additional C# file from the project template.

Evangelink:
Fixes #4116.

## Motivation

Today, opting into assembly-level parallelization with MSTest requires authoring a hand-written `MSTestSettings.cs` file with `[assembly: Parallelize(...)]` (or `[assembly: DoNotParallelize]`). It would be much nicer to express the intent as a single MSBuild property.

## Changes

Two new MSBuild properties are introduced for any project that references `MSTest.TestAdapter` (directly, via the `MSTest` meta-package, or transitively via `MSTest.Sdk`):

| Property | Allowed values | Effect |
|---|---|---|
| `MSTestParallelizeScope` | `None`, `ClassLevel`, `MethodLevel` | Emits `[assembly: Parallelize(Scope = ...)]` (`ClassLevel`/`MethodLevel`) or `[assembly: DoNotParallelize]` (`None`) |
| `MSTestParallelizeWorkers` | non-negative integer | Sets the `Workers` named argument on `[assembly: Parallelize(...)]` |

Both can be combined; either can be set on its own. When only `MSTestParallelizeWorkers` is provided, `Parallelize` defaults its `Scope` to `ClassLevel` (the attribute's own default).

Example with `MSTest.Sdk`:

```xml
<Project Sdk="MSTest.Sdk/4.x.y">
  <PropertyGroup>
    <MSTestParallelizeScope>MethodLevel</MSTestParallelizeScope>
    <MSTestParallelizeWorkers>4</MSTestParallelizeWorkers>
  </PropertyGroup>
</Project>
```

Example with plain MSTest (no `MSTest.Sdk`):

```xml
  <PropertyGroup>
    <OutputType>Exe</OutputType>
    <MSTestParallelizeScope>MethodLevel</MSTestParallelizeScope>
    <MSTestParallelizeWorkers>4</MSTestParallelizeWorkers>
  </PropertyGroup>
  <ItemGroup>
    <PackageReference Include="MSTest" Version="..." />
  </ItemGroup>
</Project>
```

## Implementation

The MSBuild logic lives in `MSTest.TestAdapter`'s `buildTransitive` targets (both the `common` and `uwp` variants), so it applies to any project that references the adapter — directly, via the `MSTest` meta-package, or transitively via `MSTest.Sdk`.

The properties translate into `<AssemblyAttribute>` items, which the standard `WriteCodeFragment` task in `Microsoft.NET.Sdk` emits into the generated `AssemblyInfo.cs` (when `GenerateAssemblyInfo` is `true`, the default). Named arguments are produced via the `_ParameterN_IsLiteral=true` metadata.

Validation (runs `BeforeTargets="GenerateAssemblyInfo;CoreCompile"` so errors surface before code generation):
- Invalid `MSTestParallelizeScope` value → build error.
- `MSTestParallelizeWorkers` set together with `MSTestParallelizeScope=None` → build error.
- Non-integer `MSTestParallelizeWorkers` → build error.
- Properties set while `GenerateAssemblyInfo=false` → build warning (since the attributes won't be emitted).

## Tests

8 new acceptance tests in `test/IntegrationTests/MSTest.Acceptance.IntegrationTests/SdkTests.cs` (all passing locally) covering:

1. `MSTestParallelizeScope=ClassLevel` (no workers)
2. `MSTestParallelizeScope=MethodLevel` + `MSTestParallelizeWorkers=3`
3. `MSTestParallelizeWorkers=2` only (defaults `Scope` to `ClassLevel`)
4. `MSTestParallelizeScope=None` (emits `DoNotParallelize`, parallelization disabled)
5. Invalid scope value → build fails with the validation error
6. Non-integer workers value → build fails with the validation error
7. `MSTestParallelizeScope=None` together with `MSTestParallelizeWorkers` → build fails with the validation error
8. **Plain MSTest (no `MSTest.Sdk`)** with `MSTestParallelizeScope=MethodLevel` + `MSTestParallelizeWorkers=3`

Happy-path tests verify the runner's own `Test Parallelization enabled ... (Workers: N, Scope: X)` diagnostic line, which proves the attribute was emitted with the expected values.

## Backwards compatibility

- No public API surface changes (no entries needed in `PublicAPI.Unshipped.txt`).
- Existing projects with their own `[assembly: Parallelize(...)]` attribute are unaffected as long as they don't also set the new properties (otherwise they would get a duplicate-attribute compile error, as with any other duplicated assembly attribute).


src/Package/MSTest.Sdk/Sdk/Parallelize.targets
```diff
@@ -0,0 +1,52 @@
...
+      <_Parameter1_IsLiteral>true</_Parameter1_IsLiteral>
+    </AssemblyAttribute>
+  </ItemGroup>
+
+  <!-- Both set => [assembly: Parallelize(Scope = ..., Workers = ...)] -->
+  <ItemGroup Condition=" ('$(MSTestParallelizeScope)' == 'ClassLevel' or '$(MSTestParallelizeScope)' == 'MethodLevel') and '$(MSTestParallelizeWorkers)' != '' ">
+    <AssemblyAttribute Include="Microsoft.VisualStudio.TestTools.UnitTesting.Parallelize">
+      <_Parameter1>Scope = Microsoft.VisualStudio.TestTools.UnitTesting.ExecutionScope.$(MSTestParallelizeScope)</_Parameter1>
+      <_Parameter1_IsLiteral>true</_Parameter1_IsLiteral>
+      <_Parameter2>Workers = $(MSTestParallelizeWorkers)</_Parameter2>
```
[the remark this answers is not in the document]
Evangelink: This concern doesn't actually apply: in C# attribute usage, `[Foo(Property = value)]` invokes the parameterless constructor and sets the named property (analogous to object initializers), so the emitted `[Parallelize(Scope = ExecutionScope.MethodLevel, Workers = 4)]` compiles fine — `ParallelizeAttribute` exposes `Scope` and `Workers` as settable properties and has a `public ParallelizeAttribute()` constructor. The acceptance tests verify this end-to-end (the build succeeds and the runtime emits the configured parallelization). Using named properties is also the cleanest path for the `Workers`-only case, since there isn't a positional `(int workers)` constructor.

Note: the file has been moved to `MSTest.TestAdapter`'s buildTransitive targets in a follow-up commit, so this thread is on the now-deleted file.

src/Package/MSTest.Sdk/Sdk/Parallelize.targets
```diff
@@ -0,0 +1,52 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<Project xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+
+  <!--
+    MSTestParallelizeScope and MSTestParallelizeWorkers let users opt-in to assembly-level
+    parallelization without authoring an MSTestSettings.cs file, by emitting the corresponding
+    [assembly: Parallelize(...)] / [assembly: DoNotParallelize] attribute via the standard
+    AssemblyAttribute MSBuild item (consumed by WriteCodeFragment when GenerateAssemblyInfo is true).
+  -->
+  <Target Name="_MSTestSDKValidateParallelizeProperties" BeforeTargets="GenerateAssemblyInfo;CoreCompile">
```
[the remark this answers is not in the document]
Evangelink: The conditions are actually disjoint, so the scenario doesn't manifest in practice:
- The `None + Workers` check requires `Scope == 'None'` exactly, so an invalid scope like `Foo` doesn't trigger it.
- The `Workers` integer check is independent of `Scope` — when both are wrong, reporting both errors in a single build is desirable.

I'd rather keep the rules independent so the user gets the full diagnostic picture in one build pass.

Note: the file has moved to `MSTest.TestAdapter`'s buildTransitive targets in a follow-up commit; the same logic and conditions apply there.

test/IntegrationTests/MSTest.Acceptance.IntegrationTests/SdkTests.cs
```diff
@@ -362,4 +362,176 @@ public async Task SettingIsTestApplicationToFalseReducesAddedExtensionsAndMakesP
         // It's not an executable
         Assert.DoesNotContain(p => p.Value == "Exe", binLog.FindChildrenRecursive<SL.Property>(p => p.Name == "OutputType"));
     }
+
+    private const string ParallelizeAssertSourceCode = """
```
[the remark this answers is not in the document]
Evangelink: Good catch, fixed in 6e7eb8fd8 — renamed to `ParallelizeAssetSourceCode`.

src/Adapter/MSTest.TestAdapter/buildTransitive/Parallelize.targets
```diff
@@ -0,0 +1,52 @@
...
+  -->
+  <Target Name="_MSTestSDKValidateParallelizeProperties" BeforeTargets="GenerateAssemblyInfo;CoreCompile">
+    <Error Condition=" '$(MSTestParallelizeScope)' != '' and '$(MSTestParallelizeScope)' != 'None' and '$(MSTestParallelizeScope)' != 'ClassLevel' and '$(MSTestParallelizeScope)' != 'MethodLevel' "
+           Text="Invalid value '$(MSTestParallelizeScope)' for property MSTestParallelizeScope. Valid values are 'None', 'ClassLevel' and 'MethodLevel'." />
+    <Error Condition=" '$(MSTestParallelizeScope)' == 'None' and '$(MSTestParallelizeWorkers)' != '' "
+           Text="Property MSTestParallelizeWorkers cannot be set when MSTestParallelizeScope is 'None'." />
+    <Error Condition=" '$(MSTestParallelizeWorkers)' != '' and !$([System.Text.RegularExpressions.Regex]::IsMatch('$(MSTestParallelizeWorkers)', '^\d+$')) "
+           Text="Invalid value '$(MSTestParallelizeWorkers)' for property MSTestParallelizeWorkers. The value must be a non-negative integer." />
+    <Warning Condition=" ('$(MSTestParallelizeScope)' != '' or '$(MSTestParallelizeWorkers)' != '') and '$(GenerateAssemblyInfo)' == 'false' "
```
Evangelink, line 17: **[MODERATE] Test Completeness — `GenerateAssemblyInfo=false` warning path untested**

The warning at line 17–18 is emitted when either property is set but `GenerateAssemblyInfo=false`. This is an important guard (properties silently do nothing without it), but no integration test verifies:
- That the warning is actually emitted with the expected text
- That the build still succeeds (exit code 0) despite the warning

**Recommendation:** Add a test that sets `<GenerateAssemblyInfo>false</GenerateAssemblyInfo>` alongside `<MSTestParallelizeScope>ClassLevel</MSTestParallelizeScope>`, asserts exit code 0, and asserts the warning message appears in build output.

src/Adapter/MSTest.TestAdapter/buildTransitive/Parallelize.targets
```diff
@@ -0,0 +1,52 @@
...
+    parallelization without authoring an MSTestSettings.cs file, by emitting the corresponding
+    [assembly: Parallelize(...)] / [assembly: DoNotParallelize] attribute via the standard
+    AssemblyAttribute MSBuild item (consumed by WriteCodeFragment when GenerateAssemblyInfo is true).
+  -->
+  <Target Name="_MSTestSDKValidateParallelizeProperties" BeforeTargets="GenerateAssemblyInfo;CoreCompile">
+    <Error Condition=" '$(MSTestParallelizeScope)' != '' and '$(MSTestParallelizeScope)' != 'None' and '$(MSTestParallelizeScope)' != 'ClassLevel' and '$(MSTestParallelizeScope)' != 'MethodLevel' "
+           Text="Invalid value '$(MSTestParallelizeScope)' for property MSTestParallelizeScope. Valid values are 'None', 'ClassLevel' and 'MethodLevel'." />
+    <Error Condition=" '$(MSTestParallelizeScope)' == 'None' and '$(MSTestParallelizeWorkers)' != '' "
+           Text="Property MSTestParallelizeWorkers cannot be set when MSTestParallelizeScope is 'None'." />
+    <Error Condition=" '$(MSTestParallelizeWorkers)' != '' and !$([System.Text.RegularExpressions.Regex]::IsMatch('$(MSTestParallelizeWorkers)', '^\d+$')) "
```
Evangelink: **[MODERATE] Test Completeness — `Workers=0` boundary case untested**

The regex `^\d+$` accepts `"0"` as a valid value (zero is a non-negative integer). When `[assembly: Parallelize(Workers=0)]` is emitted, MSTest's `ParallelizeAttribute` interprets `Workers=0` differently from omitting it entirely (it typically means "use all available processors" or is treated as default). There is no test verifying:
1. That `Workers=0` passes validation without error
2. What runtime behavior it produces

**Recommendation:** Add a `[DataRow]` or dedicated test with `MSTestParallelizeWorkers=0` to document and lock in the intended boundary behavior.

test/IntegrationTests/MSTest.Acceptance.IntegrationTests/SdkTests.cs
```diff
@@ -362,4 +362,176 @@ public async Task SettingIsTestApplicationToFalseReducesAddedExtensionsAndMakesP
         // It's not an executable
         Assert.DoesNotContain(p => p.Value == "Exe", binLog.FindChildrenRecursive<SL.Property>(p => p.Name == "OutputType"));
```
Evangelink, line 363: **[NIT] Naming — `ParallelizeAssertSourceCode` constant name is misleading**

The constant is named `ParallelizeAssertSourceCode` but it defines _test project source code_, not assertion source code. The name suggests it tests the `Assert` class. Compare with `SingleTestSourceCode` which follows the clear `*SourceCode` convention.

**Recommendation:** Rename to `ParallelizeTestSourceCode` (or `ParallelizeProjectSourceCode`) to match the existing convention and avoid confusion.

test/IntegrationTests/MSTest.Acceptance.IntegrationTests/SdkTests.cs
```diff
@@ -362,4 +362,176 @@ public async Task SettingIsTestApplicationToFalseReducesAddedExtensionsAndMakesP
...
     }
+
+    private const string ParallelizeAssertSourceCode = """
+#file MSTestSdk.csproj
+<Project Sdk="MSTest.Sdk/$MSTestVersion$" >
+
+  <PropertyGroup>
+    <EnableMicrosoftTestingPlatform>true</EnableMicrosoftTestingPlatform>
```
Evangelink: **[MODERATE] Test Completeness — VSTest runner path untested**

All four happy-path tests hardcode `<EnableMicrosoftTestingPlatform>true</EnableMicrosoftTestingPlatform>` and never exercise the VSTest runner path (`<UseVSTest>true</UseVSTest>`). The PR description says _"applies to both runners"_ and `Sdk.targets` imports `Parallelize.targets` unconditionally, but only MTP execution is verified.

The `_Parameter1_IsLiteral` / `AssemblyAttribute` pattern emits C# code via `WriteCodeFragment`, so the assembly attribute itself is runner-agnostic. However, the _observation_ of the parallelization output (`"Test Parallelization enabled"`) is MTP-specific. At minimum, a build-only test (compile + verify no error) with `UseVSTest=true` should be added to confirm the targets don't interfere with VSTest projects.

**Recommendation:** Add one test with `<UseVSTest>true</UseVSTest>` that at least builds successfully and optionally runs the test to verify the attribute is effective.

Evangelink (commented):
## Review: Add MSTestParallelizeScope/Workers MSBuild SDK properties (#4116)

| # | Dimension | Verdict |
|---|-----------|---------|
| 13 | Test Completeness & Coverage | 🟠 3 MODERATE |
| 16 | Naming & Conventions | 🔵 1 NIT |

✅ 19/21 dimensions clean.

- [ ] Test Completeness — VSTest runner path untested (all 4 happy-path tests use MTP only)
- [ ] Test Completeness — `Workers=0` boundary case not exercised
- [ ] Test Completeness — `GenerateAssemblyInfo=false` warning path has no test
- [ ] Naming — `ParallelizeAssertSourceCode` constant is misnamed

**Overall:** The implementation is solid — the MSBuild targets are correctly structured, the three ItemGroup conditions (scope only / workers only / both) are mutually exclusive and exhaustive, validation fires at the right target entry points, and the nuspec/verify-nupkgs counts are in sync. The findings above are all in the test layer and don't affect correctness of the feature itself.

> Generated by [Expert Code Review (on open)](https://github.com/microsoft/testfx/actions/runs/25878148007/agentic_workflow) for issue #8233 · ● 7M

test/IntegrationTests/MSTest.Acceptance.IntegrationTests/SdkTests.cs
```diff
@@ -362,4 +362,235 @@ public async Task SettingIsTestApplicationToFalseReducesAddedExtensionsAndMakesP
...
+
+    [TestMethod]
+    public async Task MSTestParallelizeScope_InvalidValue_FailsBuild()
+    {
+        using TestAsset testAsset = await TestAsset.GenerateAssetAsync(
+            AssetName,
+            SingleTestSourceCode
+            .PatchCodeWithReplace("$MSTestVersion$", MSTestVersion)
+            .PatchCodeWithReplace("$TargetFramework$", TargetFrameworks.NetCurrent)
+            .PatchCodeWithReplace("$ExtraProperties$", "<MSTestParallelizeScope>NotAValidValue</MSTestParallelizeScope>"));
```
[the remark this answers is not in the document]
Evangelink: `SingleTestSourceCode` does contain the `$ExtraProperties$` placeholder (see line 31), so the `FailsBuild` tests work as intended. To remove the inconsistency you noticed, I switched the happy-path tests over to the same `SingleTestSourceCode` constant in 6eacaa6af and deleted the near-identical `ParallelizeAssetSourceCode` duplicate.

src/Package/MSTest.Sdk/MSTest.Sdk.nuspec
```diff
@@ -9,6 +9,7 @@
 
     <file src="$MSBuildProjectDirectory$/Sdk/Sdk.props" target="Sdk" />
     <file src="$MSBuildProjectDirectory$/Sdk/Sdk.targets" target="Sdk" />
+
     <file src="$MSBuildProjectDirectory$/Sdk/Features/Aspire.targets" target="Sdk/Features" />
```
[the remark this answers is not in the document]
Evangelink: Reverted in 6eacaa6af — that blank line was unintended.

test/IntegrationTests/MSTest.Acceptance.IntegrationTests/SdkTests.cs
```diff
@@ -1,4 +1,4 @@
-﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Copyright (c) Microsoft Corporation. All rights reserved.
```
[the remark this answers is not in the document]
Evangelink: Good catch — the BOM was stripped accidentally. Restored in 6eacaa6af.

src/Adapter/MSTest.TestAdapter/buildTransitive/Parallelize.targets
Evangelink: Added `MSTestParallelizeScope_GenerateAssemblyInfoFalse_EmitsWarning` in 6eacaa6af — sets `<GenerateAssemblyInfo>false</GenerateAssemblyInfo>` together with `<MSTestParallelizeScope>ClassLevel</MSTestParallelizeScope>`, asserts exit code 0 (with `warnAsError: false` on the test runner so warnings don't escalate), and asserts the warning text appears in the build output.

test/IntegrationTests/MSTest.Acceptance.IntegrationTests/SdkTests.cs
Evangelink: Added `MSTestParallelizeScope_VSTestRunner_BuildSucceeds` in 6eacaa6af — uses `SingleTestSourceCodeVSTest` with `<UseVSTest>true</UseVSTest>` plus both `MSTestParallelizeScope=MethodLevel` and `MSTestParallelizeWorkers=2`, and asserts the build succeeds. As you noted, the runtime "Test Parallelization enabled" diagnostic is MTP-specific so this stays a build-only check; the underlying `WriteCodeFragment`/`AssemblyAttribute` mechanism is runner-agnostic.

test/IntegrationTests/MSTest.Acceptance.IntegrationTests/SdkTests.cs
Evangelink: Addressed in 6eacaa6af — removed the duplicate `ParallelizeAssetSourceCode` constant entirely and switched the happy-path tests to reuse the existing `SingleTestSourceCode`.

src/Adapter/MSTest.TestAdapter/buildTransitive/uwp/MSTest.TestAdapter.targets
```diff
@@ -51,4 +51,52 @@
     </ItemGroup>
   </Target>
 
+  <!--
+    MSTestParallelizeScope and MSTestParallelizeWorkers let users opt-in to assembly-level
+    parallelization without authoring an MSTestSettings.cs file, by emitting the corresponding
+    [assembly: Parallelize(...)] / [assembly: DoNotParallelize] attribute via the standard
+    AssemblyAttribute MSBuild item (consumed by WriteCodeFragment when GenerateAssemblyInfo is true).
+  -->
+  <Target Name="_MSTestValidateParallelizeProperties" BeforeTargets="GenerateAssemblyInfo;CoreCompile">
```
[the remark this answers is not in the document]
Evangelink: Done in 6eacaa6af — extracted the Parallelize logic into `src/Adapter/MSTest.TestAdapter/buildTransitive/Parallelize.targets`. Both `common/MSTest.TestAdapter.targets` and `uwp/MSTest.TestAdapter.targets` now just `<Import>` it. The shared file is packaged into each TFM-specific `buildTransitive/<tfm>/` folder so the import resolves at consumer time.

src/Adapter/MSTest.TestAdapter/buildTransitive/Parallelize.targets
Evangelink: Added `MSTestParallelizeWorkers_Zero_IsAcceptedAndEmitsParallelizeAttribute` in 6eacaa6af. It confirms `Workers=0` passes validation, the build succeeds, and the runner reports `(Workers: <N>, Scope: ClassLevel)` where `<N>` is whatever `ParallelizeAttribute` resolves 0 to at runtime (it falls back to the host's processor count).

src/Adapter/MSTest.TestAdapter/buildTransitive/common/MSTest.TestAdapter.targets
```diff
@@ -74,4 +74,52 @@
...
+  <!-- Both set => [assembly: Parallelize(Scope = ..., Workers = ...)] -->
+  <ItemGroup Condition=" ('$(MSTestParallelizeScope)' == 'ClassLevel' or '$(MSTestParallelizeScope)' == 'MethodLevel') and '$(MSTestParallelizeWorkers)' != '' ">
+    <AssemblyAttribute Include="Microsoft.VisualStudio.TestTools.UnitTesting.Parallelize">
+      <_Parameter1>Scope = Microsoft.VisualStudio.TestTools.UnitTesting.ExecutionScope.$(MSTestParallelizeScope)</_Parameter1>
+      <_Parameter1_IsLiteral>true</_Parameter1_IsLiteral>
+      <_Parameter2>Workers = $(MSTestParallelizeWorkers)</_Parameter2>
+      <_Parameter2_IsLiteral>true</_Parameter2_IsLiteral>
+    </AssemblyAttribute>
+  </ItemGroup>
+
```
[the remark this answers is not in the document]
Evangelink: I'd rather keep the three branches as they are. Trying to collapse them into one `ItemGroup` with conditional `_Parameter1`/`_Parameter2` metadata gets messy because the parameter ordering depends on which property is set (when only `Workers` is provided it has to land in `_Parameter1`, not `_Parameter2`), and MSBuild metadata on items doesn't support `Condition` on individual metadata in a way that preserves the right ordering cleanly. The current shape — one `ItemGroup` per case, each with a self-explanatory comment — is easier to read and harder to get wrong, and the file is also no longer duplicated across `common` and `uwp` (see the extraction in 6eacaa6af).

src/Adapter/MSTest.TestAdapter/buildTransitive/Parallelize.targets
```diff
@@ -0,0 +1,52 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<Project>
+
+  <!--
+    MSTestParallelizeScope and MSTestParallelizeWorkers let users opt-in to assembly-level
+    parallelization without authoring an MSTestSettings.cs file, by emitting the corresponding
+    [assembly: Parallelize(...)] / [assembly: DoNotParallelize] attribute via the standard
+    AssemblyAttribute MSBuild item (consumed by WriteCodeFragment when GenerateAssemblyInfo is true).
+  -->
+  <Target Name="_MSTestValidateParallelizeProperties" BeforeTargets="GenerateAssemblyInfo;CoreCompile">
```
Youssef1313, line 14: I'm not sure this whole target is needed. If a wrong value resulted in an error, it will be a very clear build error.

src/Adapter/MSTest.TestAdapter/buildTransitive/common/MSTest.TestAdapter.targets
```diff
@@ -74,4 +74,6 @@
       </Content>
     </ItemGroup>
   </Target>
+
+  <Import Project="$(MSBuildThisFileDirectory)Parallelize.targets" />
```
Youssef1313, line 78: I'm not very sure if this needs to happen both under `build` and `buildTransitive` or not.

src/Adapter/MSTest.TestAdapter/MSTest.TestAdapter.csproj
```diff
@@ -84,6 +84,9 @@
     <None Update="buildTransitive\uwp\*" Condition=" '$(TargetFramework)' == '$(UwpMinimum)' ">
       <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
     </None>
+    <None Update="buildTransitive\Parallelize.targets">
+      <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
```
Youssef1313, line 88: Not for this PR, but you should always prefer `IfDifferent` and not `PreserveNewest` (except in props/targets that flows to user because `IfDifferent` isn't supported in all SDK versions - unless you're willing to do a version check and conditionally decide)

Youssef1313 (commented):
You should also consider following-up with template updates in SDK once next stable is available.

src/Adapter/MSTest.TestAdapter/buildTransitive/common/MSTest.TestAdapter.targets
JanKrivanek: `buildTransitive` should be enough - https://github.com/NuGet/Home/wiki/Allow-package--authors-to-define-build-assets-transitive-behavior

JanKrivanek (approved):

src/Adapter/MSTest.TestAdapter/buildTransitive/Parallelize.targets
Evangelink: Good catch — pushed [cc3403b15](https://github.com/microsoft/testfx/commit/cc3403b15) to drop the redundant validation. The invalid-`Scope` and non-integer-`Workers` errors are gone; the `ItemGroup` conditions now emit the assembly attribute for any non-empty, non-`None` `Scope`, so a bad value (e.g. `NotAValidValue`, or `Workers=abc`) flows through `WriteCodeFragment` and is rejected by the C# compiler with a clear `CS0117` / `CS0103`. The two acceptance tests now assert those compiler errors.

I kept the two checks that catch *silent* failure modes (no compile error, no behavior):
- `Scope=None` together with `Workers=N` — the `DoNotParallelize` branch wins and the `Workers` value would be silently dropped.
- Any of the properties set with `GenerateAssemblyInfo=false` — `WriteCodeFragment` doesn't run, so the properties silently do nothing.

Happy to drop those too if you'd rather they fail silently, but I think the diagnostic is genuinely useful in those cases.

src/Adapter/MSTest.TestAdapter/buildTransitive/common/MSTest.TestAdapter.targets
Evangelink: Good observation, but the `build/<tfm>/MSTest.TestAdapter.targets` duplication is pre-existing — it's how the package has shipped for a long time and isn't introduced by this PR. My new `Parallelize.targets` is only shipped under `buildTransitive/`, never under `build/`.

I'd prefer not to change the broader package layout in this PR (it has migration / customer-impact considerations of its own). Happy to do it as a separate PR — could you open an issue to track it?

src/Adapter/MSTest.TestAdapter/MSTest.TestAdapter.csproj
Evangelink: Acknowledged — thanks for the tip. Leaving as `PreserveNewest` here per your "not for this PR" note; happy to follow up separately if you'd like.

src/Adapter/MSTest.TestAdapter/buildTransitive/common/MSTest.TestAdapter.targets
Youssef1313: @JanKrivanek - I mean, it's not compatible with packages.config, but it's a question of "do we care?"

YuliiaKovalova (approved):

src/Adapter/MSTest.TestAdapter/buildTransitive/common/MSTest.TestAdapter.targets
Evangelink: Let's try as-is.

Issue #8230, opened 2026-05-14
[Parent] TRX report improvements and reliability


Evangelink:
This is a tracking issue for improvements and bug fixes related to TRX report generation in MTP. The following issues are related sub-tasks:

- #7345 — Using TrxReport with --test-modules should include test assembly name and target framework in trx file name
- #6778 — When HangDump extension triggers, test run does not produce a .trx file
- #6648 — `Microsoft.Testing.Extensions.TrxReport` should allow customizing `trx` file name via dedicated `MSBuild` property
- #6590 — MTP-generated .trx file is missing
- #5774 — TrxReport keeps all test nodes in memory
- #5634 — TRX file name output should support path

These issues collectively track improvements to TRX file generation, naming, reliability, and memory efficiency in MTP.

> Generated by [Issue Arborist](https://github.com/microsoft/testfx/actions/runs/25870610117/agentic_workflow) · ● 14.2M · [◷](https://github.com/search?q=repo%3Amicrosoft%2Ftestfx+is%3Aissue+%22gh-aw-workflow-call-id%3A+microsoft%2Ftestfx%2Fissue-arborist%22&type=issues)
> - [x] expires  on May 16, 2026, 4:22 PM UTC

Issue #7345, opened 2026-02-04
Using TrxReport with --test-modules should include test assembly name and target framework in trx file name


jbcutting:
## Summary

When running `dotnet test --test-modules <glob> --report-trx`, the .trx files that are created have names of `<username>_<computername>_<UTC timestamp>.trx`. With multiple test modules, the files are indistinguishable from one another. The `--report-trx-filename` parameter doesn't help - with one .trx file, the results seem to overwrite each other since the end result does not contain all the tests that were run.

## Background and Motivation

We have tests broken up into a few dozen test projects by product and code areas. With VSTest, we pass all of the test assemblies and the result goes to a single .trx file, which works well.

With `dotnet test` with MTP, I end up with dozens of .trx files and can't tell which is which without writing something custom to parse them.

We also multi-target (net48 and net10.0), so I'm using `bin/Release/net*/*Test.exe` as a glob pattern. This works great, all of the tests run nicely in parallel with the target framework included in the output. However, it's also not possible to tell the target framework from the resulting .trx files.

## Proposed Feature

What seems simplest to me is to include the test module name and TFM in the file name. This would simplify things drastically.

Additional nice to haves would be:
* The option to output all results to a single .trx.
* Option to create a .trx file by some value such as TFM, category, etc. For example, all net48 tests and all net10.0 tests. TFM is probably easier than something on the tests like category or namespace, but those are something to consider.

Issue #6778, opened 2025-10-21
When HangDump extension triggers, test run does not produce a .trx file


vikukush:
## Describe the bug

When one of tests hangs and Microsoft.Testing.Extensions.HangDump triggers, the test run does not produce a .trx file with results.

## Steps To Reproduce

1. Make one of your tests hang
2. Add Microsoft.Testing.Extensions.HangDump to your project
3. Run the tests with --report-trx flag

## Expected behavior
Hanged test is dumped, .trx file with test results is produced:
1. Tests that ran to completion before the hang have their results (pass or fail) properly recorded
2. Hanged test(s) are marked as such in some form
3. Tests that didn't get to run due to the hang are marked as Unknown or NotStarted or whatever's most appropriate

## Actual behavior
```
Running tests using MSTest Runner [D:\Test\TestAssemblies\MS.Dynamics.Platform.Integration.Office.ScenarioTests.DocumentManagement.exe]
MSTest v3.10.0 (UTC 7/29/2025) [x64 - .NET Framework 4.8.4795.0]
Diagnostic file (level 'Trace' with async flush): D:\data\CHT\df0c8ed9\UserLogs\a5dbfee2\TestResults\log_251021041813979.diag
Diagnostic file (level 'Trace' with async flush): D:\data\CHT\df0c8ed9\UserLogs\a5dbfee2\TestResults\log_251021041818744.diag
Runsettings datacollectors are not supported by Microsoft.Testing.Platform and will be ignored
Runsettings attribute 'TargetPlatform' is not supported by Microsoft.Testing.Platform and will be ignored
Hang dump timeout of '00:20:00' expired
The following tests were still running when dump was taken (format: [<time-elapsed-since-start>] <name>):
[00:19:54] TestDocumentHandlingCreatesDocumentForNewTypeInNonDefaultCompany
Creating dump file 'D:\data\CHT\df0c8ed9\UserLogs\a5dbfee2\TestResults\MS.Dynamics.Platform.Integration.Office.ScenarioTests.DocumentManagement_8668_hang.dmp'
Test application process didn't exit gracefully, exit code is '7'
```

## Additional context

Without an output .trx file it's impossible to understand in automation (CloudTest/QTest) what exactly happened. We'll be forced to try parsing tool's stdout manually and pray that the output text format won't change.

Evangelink:
This should be addressed by PR #8263 (`Stream TRX test results to disk and recover partial TRX on test host crash`), merged in 2026-05-16.

[...]

Issue #6648, opened 2025-10-01
`Microsoft.Testing.Extensions.TrxReport` should allow customizing `trx` file name via dedicated `MSBuild` property


julealgon:
## Summary

The only way to currently customize the `.trx` filename when using MTP with the `Microsoft.Testing.Extensions.TrxReport` extension is by using the much more general purpose `TestingPlatformCommandLineArguments` `MSBuild` property.

I believe forcing the use of this single property makes for a more convoluted configuration experience and creates maintenance/flexibility issues for consumers.

A dedicated, more specific property should be provided that _only_ concerns itself with the format of the `trx` file and nothing else.

## Background and Motivation

We integrated `.trx`-based test reports in our GitHub workflows this week and it was a pain to get things to how we wanted. The current defaults for the file name are, in my opinion, awful, so we wanted to change them to something that would be more readable to the outside.

We used the [dorny/test-reporter GitHub action](https://github.com/dorny/test-reporter) to parse the `.trx` files from each of our projects and generate the test report. The tool relies on the filename to produce the various sections in its report (which by itself makes sense).

We made it work by configuring the `TestingPlatformCommandLineArguments` property like this:

```xml
<TestingPlatformCommandLineArguments>--report-trx --report-trx-filename $(MSBuildProjectName).trx</TestingPlatformCommandLineArguments>
```

Which _worked_, but with some caveats. 

The main issue is that this basically _forces_ us to now _always_ output a `trx` report, even when we don't want to. Because the way these properties work (as cmdline arguments) it is not possible to only specify `--report-trx-filename`, otherwise `dotnet test` will crash with an error saying that `--report-trx-filename` must be used in conjunction with `--report-trx`.

From a CLI tool's perspective, this is very reasonable, but because the CLI argument list is the _only_ way to customize those values, it now becomes impossible to _just_ change the format of the `trx` filenames without having to introduce the extra parameters and without forcing the entire command to _always_ run.

Rather, what we really wanted, was to be able to customize _just_ the format of the filename without touching anything else, and we wanted it to apply _only when actually requesting a trx report_, which we were trying to setup only in our GitHub workflows, not locally.

Yes, it is _possible_ to add a `MSBuild` `Condition` to that property with something like:

```xml
<TestingPlatformCommandLineArguments Condition="'$(GITHUB_ACTIONS)' == 'true'">--report-trx --report-trx-filename $(MSBuildProjectName).trx</TestingPlatformCommandLineArguments>
```

[...]

PR #8263, opened 2026-05-15
Stream TRX test results to disk and recover partial TRX on test host crash

### LINKED ISSUE #5774: TrxReport keeps all test nodes in memory
We keep a list of TestNodeUpdateMessage in memory for TrxReport to generate the report at the end of the test run.

I wonder if we can write to file stream as we get the info and whether it improves Trx perf.

Evangelink:
## Summary

Replaces the in-memory `TestNodeUpdateMessage[]` accumulation in the TRX report generator with a sidecar binary stream written by a background writer task, and wires out-of-process crash recovery so a partial TRX is produced even when the test host crashes mid-run.

## Related issues

- Fixes #5774 — _TrxReport keeps all test nodes in memory_. The streaming store replaces the in-memory list with a slim DTO + sidecar binary file written in batches, eliminating the steady-state memory growth.
- Helps #6778 — _When HangDump extension triggers, test run does not produce a .trx file_. HangDump kills the test host the same way a crash would. With OOP crash recovery in place, the TestHostController now reads back the sidecar and emits a partial TRX containing every test that completed before the hang dump.
- Helps #5145 — _`--report-trx` with `--timeout` does not write trx in case of timeout_. Timeout-induced cancellation is structurally similar to a crash; partial TRX recovery applies in OOP mode.
- Part of #8230 — _[Parent] TRX report improvements and reliability_.

Out of scope of this PR (separate work):

- #6648 / #5634 / #7345 (TRX file naming).
- #6590 (`RunInfos` warnings missing in MTP-generated TRX).

## Motivation

1. **Memory pressure**: long test runs accumulated every `TestNodeUpdateMessage` (with full property bag) until session end. Streaming projected, slim DTOs to disk in batches eliminates the steady-state growth.
2. **No partial TRX on crash**: when the test host crashed, the TestHostController emitted an empty TRX with only the crash banner. Now it reads back the durably-written records from the sidecar.

## What changed

- New `TrxResultStreamingStore` — single background writer draining a `BlockingCollection<TrxTestResult>`, batching by size (`64`) or 500ms time window. `BinaryWriter` over a length-prefixed binary format (`TrxTestResultSerializer`). Writes are retried on transient I/O, and a mid-record failure rewinds + truncates to keep the file aligned.
- New `TrxTestResult` slim DTO + `TrxTestResultExtractor` projecting only the fields the TRX renderer consumes. Captured text fields (stdout / stderr / exception message / stack trace) are individually capped at 1 MiB to bound the per-record size.
- `TrxReportEngine.GenerateReportAsync` now takes `IReadOnlyList<TrxTestResult>` instead of `TestNodeUpdateMessage[]`.
- New `TrxStreamLocationRequest` IPC message (Id = 3) sent from test host to TestHostController as soon as the streaming sidecar is provisioned. The controller stores the path; on crash it reads the sidecar and recovers as many records as possible (`isTestHostCrashed: true` so the crash banner still appears).
- Sidecar lifecycle: deleted on the happy path; kept on `CompleteAsync` timeout; the TestHostController sweeps it on the next clean shutdown.
- New `IFileSystem.NewFileStream(..., FileShare)` overload for the recovery read.

## User-visible behaviour changes

- Captured text fields (stdout / stderr / exception message / stack trace) larger than 1 MiB are truncated with a `[truncated by TRX streaming store]` suffix. A session-end warning lists how many results had truncation.
- After a crash recovery, the controller writes how many results were salvaged on the output device.
- If the IPC notification to the controller fails, the user gets a warning at session end stating that crash recovery is unavailable for this run.
- TRX output for tests without a `TrxTestDefinitionName` property is unchanged from the legacy path (regression protection).

## Tests

- `TrxStreamingSerializerTests` — full round-trip + truncated tail, implausible / negative length prefix, multiple records in order, surrogate pairs / control characters, `+05:45` `DateTimeOffset`.
- `TrxTestResultExtractorTests` — large field truncation (exception message, stdout) + truncation at a UTF-16 surrogate boundary backs off cleanly.
- All existing `TrxTests` were updated to the new DTO shape and continue to pass.

## Validation

- `Microsoft.Testing.Extensions.UnitTests`: 398 / 398 pass across `net8.0` / `net9.0` / `net462` / `net472`.
- `Microsoft.Testing.Platform.UnitTests`: 716 (714 passed, 2 skipped) — same as baseline.
- `Microsoft.Testing.Platform.slnf` builds clean (0 warnings, 0 errors) on `netstandard2.0` / `net462` / `net472` / `net8.0` / `net9.0`.

The implementation passed three rounds of expert review with all flagged issues addressed (silent data loss paths, behaviour regression on `TrxTestDefinitionName`, IPC race window, surrogate-pair truncation boundary, sidecar orphan accumulation).


Evangelink:

src/Platform/Microsoft.Testing.Extensions.TrxReport/Streaming/TrxResultStreamingStore.cs
```diff
@@ -0,0 +1,465 @@
...
+    }
+
+    private async Task WriteLoopAsync()
+    {
+        var batch = new List<TrxTestResult>(_batchSize);
+        try
+        {
+            while (!_queue.IsCompleted)
+            {
+                if (!_queue.TryTake(out TrxTestResult? first, _flushIntervalMs))
```
Youssef1313, line 168: This blocks a threadpool thread which is extremely bad.

N-Olbert:
Is there a preview NuGet available with this? Since 4.2.3 does not include this change.

Youssef1313:
@N-Olbert See https://aka.ms/mstest/preview

PR #8223, opened 2026-05-14
Support placeholders in --report-trx-filename

Evangelink:
Resolves #5364.

The `--report-trx-filename` argument now supports placeholders provided by the existing artifact naming service (`ArtifactNamingHelper`):

| Placeholder | Description |
|-------------|-------------|
| `{pname}` | Test application name (file name of the current test app, without extension) |
| `{pid}`   | Process ID |
| `{asm}`   | Entry assembly name |
| `{tfm}`   | Target framework moniker (e.g. `net9.0`) |
| `{time}`  | Timestamp (`yyyy-MM-dd_HH-mm-ss.fffffff`) |

Example:

```
dotnet test <name>.sln -- --report-trx --report-trx-filename "test_results_{tfm}_{pid}.trx"
```

This fixes the long-standing issue where running the same `--report-trx-filename` value for every project in a solution produced a single TRX overwritten by every test project (the last one wins).

### Implementation notes

- `ArtifactNamingHelper` and `TargetFrameworkParser` are linked into the TrxReport project (mirroring the HangDump extension).
- `{pname}` uses `Path.GetFileNameWithoutExtension(_testApplicationModuleInfo.GetCurrentTestApplicationFullPath())` instead of `Process.ProcessName` so users get the test app name rather than `dotnet`.
- `{pid}` uses `IEnvironment.ProcessId` to avoid both the `Process.GetCurrentProcess()` banned-API rule and the CA1416 browser-platform warning.
- Placeholder resolution happens **before** the existing `ReplaceInvalidFileNameChars` so any unsafe characters in resolved values (e.g. parentheses, spaces) are still sanitized.
- Backward compatible: filenames without `{...}` placeholders are passed through unchanged.
- Existing CLI validation (`.trx` suffix, no path separators) is unchanged — placeholders never contain `/` or `\`, and the user still provides the literal `.trx` suffix.
- The `--report-trx-filename` description in `ExtensionResources.resx` documents the supported placeholders. XLF files were regenerated via `msbuild /t:UpdateXlf`.

### Tests

5 new unit tests added in `TrxTests.cs` covering:

- `{pname}` and `{pid}` resolution
- Unknown placeholders preserved as-is
- Plain filename without placeholders unchanged (backward compat)
- `{time}` resolved from the injected `IClock`
- Invalid characters in resolved values are sanitized

All 144 Trx-related unit tests pass on net462/net472/net8.0/net9.0.

PR #8280, opened 2026-05-16
Add --list-tests json for machine-readable test discovery output

### LINKED ISSUE #3221: Provide extra options to `--list-tests`
## Summary

The current `--list-tests` option is too static and only provide really few informations about the tests. It's also incompatible with extensions while some could have some sense.

Here are some examples of improvements that could be done:

- Display full test info instead of only display name
- Group displayed tests by [grouping] (e.g. namespaces, classes, subclasses...)
- Provide TRX with all tests as not executed

Evangelink:
Adds an optional `json` argument to the existing `--list-tests` flag so MTP can emit a single JSON document on stdout describing every discovered test. The text output and zero-arg behavior are unchanged.

Fixes #3221.
Related: dotnet/sdk#49754.

## CLI surface

| Command | Behavior |
|---|---|
| `--list-tests` | Unchanged (human-readable text) |
| `--list-tests text` | Explicit alias for the default |
| `--list-tests json` | JSON document on stdout (banner / progress / summary / per-test text suppressed; errors routed to stderr) |
| `--list-tests <anything-else>` | Fails with `ExitCode.InvalidCommandLine` and a descriptive validation error |

## Schema v1

Flat array; absent fields are omitted (no `null`), so frameworks that don't populate `TestMethodIdentifierProperty` aren't broken.

```jsonc
{
  ""schemaVersion"": 1,
  ""tests"": [
    {
      ""uid"": ""..."",
      ""displayName"": ""..."",
      ""type"": {
        ""assemblyFullName"": ""..."",
        ""namespace"": ""..."",                  // omitted for global namespace
        ""typeName"": ""MyClass+Nested`1"",      // metadata format with arity & nesting
        ""methodName"": ""MyMethod"",
        ""methodArity"": 0,
        ""returnTypeFullName"": ""..."",
        ""parameterTypeFullNames"": [""..."", ...]
      },
      ""location"": { ""file"": ""..."", ""lineStart"": 12, ""lineEnd"": 18 },
      ""traits"":     [ { ""key"": ""Category"", ""value"": ""Integration"" }, ... ],
      ""properties"": [ { ""key"": ""ExecutionId"", ""value"": ""..."" }, ... ]
    }
  ]
}
```

`traits` and `properties` are both arrays of `{ key, value }` (not JSON objects) so duplicate keys survive serialization and order is stable across runs.

## Implementation notes

- New `DiscoveredTestsJsonSerializer` + tiny `JsonStringWriter` — no dependency on `System.Text.Json` (not available across all targets) or `Jsonite` (`#if !NETCOREAPP` only), so the same code path runs on **netstandard2.0, net462, net8.0, net9.0**.
- Always-LF line endings for stable CI golden output, `U+2028`/`U+2029` escaped, lone surrogates replaced with `U+FFFD` (mirrors `Utf8JsonWriter` defaults).
- `TerminalOutputDevice` detects JSON mode, suppresses banner / progress / per-test text / free-form output on stdout, buffers `DiscoveredTestNodeStateProperty` events from `ConsumeAsync`, and emits the JSON document once at session end from the test-host process only (controller does not double-emit).
- Ctrl+C in JSON mode writes the cancellation message to stderr instead of corrupting the JSON document.
- `PlatformCommandLineProvider` flips `--list-tests` arity from `Zero` to `ZeroOrOne`, validates the argument against `json`/`text` (case-insensitive), and exposes `IsListTestsJsonOutput(...)`.

## Tests

| Suite | Result |
|---|---|
| `Microsoft.Testing.Platform.UnitTests` (full, net462 / net8 / net9) | **2,277 / 2,277** (159 in new/touched test classes) |
| `MSTest.Acceptance.IntegrationTests.TestDiscoveryTests` (incl. new JSON tests × 2 TFMs) | 12 / 12 |
| `MSTest.Acceptance.IntegrationTests.HelpInfoTests` | 6 / 6 |
| `Microsoft.Testing.Platform.Acceptance.IntegrationTests.HelpInfoTests*` | 30 / 30 |
| `Microsoft.Testing.Platform.Acceptance.IntegrationTests.ExecutionTests` | 27 / 27 |
| `Microsoft.Testing.Platform.Acceptance.IntegrationTests.TrxTests` | 28 / 28 |
| `.\build.cmd -pack -c Debug` (warnings-as-errors) | succeeded, 0 warnings, 0 errors |

## Notes

- No public API surface added (option key is internal CLI only).
- `PlatformResources.resx` updated; all 13 XLFs regenerated via `dotnet msbuild … /t:UpdateXlf`.
- No changelog entry — `docs/Changelog-Platform.md` is curated at release time, and v2.2.3 was just cut without an `Unreleased` section.

The change went through two rounds of internal expert review before opening. Highlights addressed in review:
- Errors no longer silently swallowed in JSON mode (routed to stderr).
- `Ctrl+C` no longer corrupts the JSON document (cancellation message goes to stderr).
- `PropertyBag`'s reverse-insertion-order traversal is reversed back to insertion order.
- `properties` is an array (not object) so duplicate keys are preserved.
- `type.namespace` omitted for the global namespace.
- `--list-tests text` accepted as explicit alias of the default for forward compatibility.


Evangelink (commented):
## Review Summary — PR #8280: `--list-tests json`

Overall this is a well-structured, well-tested feature. The custom `JsonStringWriter`, cross-TFM compatibility, lone-surrogate handling, banner/progress suppression, and CLI validation are all cleanly implemented. Tests cover the happy path, special character escaping, and invalid-argument rejection.

### Findings

| # | Dimension | Severity | File | Finding |
|---|-----------|----------|------|---------|
| 1 | Threading/Hot-reload | **MODERATE** | `TerminalOutputDevice.cs:379` | `DisplayAfterHotReloadSessionEndAsync` calls `DisplayAfterSessionEndRunInternalAsync` unconditionally; in a `dotnet watch --list-tests json` session the JSON document is emitted after every cycle and `_discoveredTestsForJson` is never cleared — multiple growing JSON blobs on stdout. |
| 2 | Testability/Architecture | **NIT** | `TerminalOutputDevice.cs:420` | `Console.Error.WriteLineAsync` is used directly (also in the abort callback in `InitializeAsync`) bypassing the injected `_console`. This makes error/stderr output in JSON mode un-redirectable in tests. The comment acknowledges the limitation; worth tracking. |

### Clean dimensions
- **Algorithmic Correctness** — JSON writer logic, escaping, surrogate pair handling all correct.  
- **Threading** — `_discoveredTestsForJson` write path is guarded by MTP's single-consumer-per-pump invariant; the emission path uses `_asyncMonitor`. Correct.  
- **Security** — No path traversal, no serialization of untrusted schemas, no sensitive leakage.  
- **Public API** — No new public API surface; `IsListTestsJsonOutput` is `internal static`. `PublicAPI.Unshipped.txt` not dirtied.  
- **Cross-TFM** — Custom `JsonStringWriter` avoids both `System.Text.Json` and `Jsonite`; no `#if`-guarded APIs needed.  
- **Performance** — `StringBuilder`-based writer; no hot-path LINQ; `PropertyBag.OfType<T>` returns an array, `Array.Reverse` is in-place. Reasonable.  
- **CLI validation** — Argument validated in `ValidateOptionArgumentsAsync`; `ZeroOrOne` arity correct; case-insensitive comparison correct; resource string properly parametrised.  
- **Localization** — New strings in `.resx`; XLFs regenerated via MSBuild, not hand-edited.  
- **Test quality** — Unit tests cover empty, minimal, full, special-char, surrogate, and validation cases. Integration tests cover JSON output schema, invalid arg, and unchanged text mode.  
- **Assertion style** — Uses MSTest conventions consistently throughout.  
- **Scope discipline** — PR is tightly scoped to the new `--list-tests json` flag; no unrelated refactoring.

> Generated by [Expert Code Review (on open)](https://github.com/microsoft/testfx/actions/runs/25958095769/agentic_workflow) for issue #8280 · ● 19.8M

src/Platform/Microsoft.Testing.Platform/OutputDevice/TerminalOutputDevice.cs
```diff
@@ -339,6 +370,19 @@ private async Task DisplayAfterSessionEndRunInternalAsync()
     {
         RoslynDebug.Assert(_terminalTestReporter is not null);
 
+        if (_isListTestsJson)
+        {
+            using (await _asyncMonitor.LockAsync(TimeoutHelper.DefaultHangTimeSpanTimeout).ConfigureAwait(false))
+            {
+                if (_processRole == TestProcessRole.TestHost)
+                {
+                    _console.WriteLine(DiscoveredTestsJsonSerializer.Serialize(_discoveredTestsForJson));
```
Evangelink, line 396: **Hot-reload double-emission (MODERATE)**

`DisplayAfterHotReloadSessionEndAsync` delegates unconditionally to `DisplayAfterSessionEndRunInternalAsync()`, which emits the JSON document. In a `dotnet watch --list-tests json` session this method is called after *every* hot-reload cycle, and `_discoveredTestsForJson` is never cleared between cycles. The result is multiple JSON documents on stdout (one per cycle), each progressively larger due to accumulated test nodes — broken for any consumer that pipes the output.

Suggested fix: clear the buffer after emission, or add an early-return in `DisplayAfterHotReloadSessionEndAsync` to prevent re-emission:

```csharp
public async Task DisplayAfterHotReloadSessionEndAsync(CancellationToken cancellationToken)
{
    if (_isListTestsJson)
    {
        return; // JSON discovery is a point-in-time snapshot; not applicable to hot-reload cycles.
    }
    await DisplayAfterSessionEndRunInternalAsync().ConfigureAwait(false);
}
```

src/Platform/Microsoft.Testing.Platform/OutputDevice/TerminalOutputDevice.cs
```diff
@@ -363,6 +407,28 @@ public async Task DisplayAsync(IOutputDeviceDataProducer producer, IOutputDevice
...
+        {
+            // Machine-readable mode: keep stdout reserved for the JSON document so consumers can
+            // pipe it directly. Errors and exceptions still need surfacing somewhere, so route
+            // them to stderr (bypassing IConsole, which does not abstract stderr today). Warnings
+            // and informational text are dropped to keep stdout strictly JSON.
+            switch (data)
+            {
+                case ErrorMessageOutputDeviceData errorData:
+                    await LogDebugAsync(errorData.Message).ConfigureAwait(false);
+                    await Console.Error.WriteLineAsync(errorData.Message).ConfigureAwait(false);
```
Evangelink: **`Console.Error` bypasses the `IConsole` abstraction (NIT → testability)**

`Console.Error.WriteLineAsync(...)` is used here (and in the abort callback in `InitializeAsync`) rather than the injected `_console`. The comment on line ~513 acknowledges this limitation. While acceptable as an interim measure, it means error/exception output in JSON mode cannot be redirected or verified in unit tests. If `IConsole` ever gains stderr support, this will need updating. Consider at least extracting the `Console.Error` access into a helper so the surface area of the bypass is minimal and easy to find later.

src/Platform/Microsoft.Testing.Platform/Resources/PlatformResources.resx
```diff
@@ -385,7 +385,11 @@ If not specified the file will be generated inside the default 'TestResults' dir
 The available values are 'Trace', 'Debug', 'Information', 'Warning', 'Error', and 'Critical'.</value>
   </data>
   <data name="PlatformCommandLineDiscoverTestsOptionDescription" xml:space="preserve">
-    <value>List available tests.</value>
+    <value>List available tests.
+Optionally accepts 'json' to print the discovered tests as a JSON document on standard output.</value>
```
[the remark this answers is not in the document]
Evangelink: Good catch — fixed in 884c759. The help text now lists both `text` (the default) and `json` so `--help` describes the full set of accepted values.

src/Platform/Microsoft.Testing.Platform/OutputDevice/TerminalOutputDevice.cs
Evangelink: Good catch — fixed in 884c759. `DisplayAfterHotReloadSessionEndAsync` now early-returns in JSON mode with a comment explaining that JSON discovery is a point-in-time snapshot and not applicable to hot-reload cycles.

src/Platform/Microsoft.Testing.Platform/OutputDevice/TerminalOutputDevice.cs
Evangelink: Done in 884c759 — both `Console.Error` access points now route through a single private `WriteToStandardErrorAsync` helper, so the bypass surface is one well-commented method we can swap out when `IConsole` gains stderr support.

Evangelink:
Pushed [884c759](https://github.com/microsoft/testfx/pull/8280/commits/884c759ea) to address review comments and unblock CI:

- **Hot-reload double-emission (Evangelink, MODERATE)** — `DisplayAfterHotReloadSessionEndAsync` now early-returns in JSON mode so `dotnet watch --list-tests json` doesn't emit multiple ever-growing JSON documents.
- **`Console.Error` bypass (Evangelink, NIT)** — both call sites now route through a single `WriteToStandardErrorAsync` helper with a comment explaining it's the sole place that bypasses `IConsole` until `IConsole` gains stderr support.
- **Help text (Copilot)** — the description now mentions both `text` and `json`.
- **CI XLF failure** — `ExtensionResources.pl.xlf` (TRX) was reverted to its pre-#8223 source during my earlier rebase (I discarded it thinking it was unrelated drift; it was actually a translation update that had to stay). Restored from `origin/main`.

Help/info acceptance tests updated for the new description; XLFs regenerated; `.\build.cmd -pack -c Debug` passes with 0 warnings, 0 errors; unit tests 159/159 in the touched classes.

Lorilatschki:
@Evangelink, we are using the `--list-tests` through the dotnet 10 currently. Where do we see in which upcoming SDK 10 update the JSON feature of `--list-tests` is inlcuded?

Evangelink:
> @Evangelink, we are using the `--list-tests` through the dotnet 10 currently. Where do we see in which upcoming SDK 10 update the JSON feature of `--list-tests` is inlcuded?

It's currently only on MTP (not yet released), I will dogfood it a little then add support to SDK side.

PR #8294, opened 2026-05-16
Fix VSTest_RunTests_Succeeds flake by serializing TelemetryTests

Evangelink:
Fixes the `VSTest_RunTests_Succeeds` (and `VSTest_DiscoverTests_Succeeds`) flake that is breaking `main` on Windows `net462` and Linux `net8.0`:

`
CSC : error CS2012: Cannot open '...\TelemetryTests\vstest\obj\Release\net462\TelemetryVSTestProject.dll' for writing
  -- The process cannot access the file '...' because it is being used by another process.
`

## Root cause

`MSTest.Acceptance.IntegrationTests` runs with assembly-level method parallelization (`[assembly: Parallelize(Scope = ExecutionScope.MethodLevel, Workers = 0)]` in `Program.cs`).

`TelemetryTests` has two methods — `VSTest_RunTests_Succeeds` and `VSTest_DiscoverTests_Succeeds` — that both invoke `dotnet test -c Release` against the **same** `AssetFixture.VSTestProjectPath`. When the two run in parallel for the same TFM, both `dotnet test` processes try to (re)build into `obj/Release/<tfm>/TelemetryVSTestProject.dll` and one of them loses the file-lock race, yielding `CS2012`.

## Fix

Add `[DoNotParallelize]` at the `TelemetryTests` class level so the methods that share the asset run serially. The class only has 5 tests, so the perf impact is negligible.

This matches the documented pattern for acceptance test classes that share a single generated mutable asset across multiple methods.

### DIFF
--- a/eng/Version.Details.xml
+++ b/eng/Version.Details.xml
@@ -1,17 +1,17 @@
 <?xml version="1.0" encoding="utf-8"?>
 <Dependencies>
   <ToolsetDependencies>
-    <Dependency Name="Microsoft.DotNet.Arcade.Sdk" Version="11.0.0-beta.26264.3">
+    <Dependency Name="Microsoft.DotNet.Arcade.Sdk" Version="11.0.0-beta.26265.4">
       <Uri>https://github.com/dotnet/arcade</Uri>
-      <Sha>83c4f08dc6c94b6eb035b966e97549e9acbb6231</Sha>
+      <Sha>60e594d8a0e742ae93f96384508dcd690b2fbf01</Sha>
     </Dependency>
-    <Dependency Name="Microsoft.DotNet.Build.Tasks.Templating" Version="11.0.0-beta.26264.3">
+    <Dependency Name="Microsoft.DotNet.Build.Tasks.Templating" Version="11.0.0-beta.26265.4">
       <Uri>https://github.com/dotnet/arcade</Uri>
-      <Sha>83c4f08dc6c94b6eb035b966e97549e9acbb6231</Sha>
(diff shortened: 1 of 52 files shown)


Evangelink (commented):

**Summary:** The fix correctly identifies the root cause (two `dotnet test -c Release` invocations competing for `obj/Release/<tfm>/TelemetryVSTestProject.dll` under assembly-level method parallelization) and applies the minimal, established pattern (`[DoNotParallelize]`) to serialize the class. The comment on the attribute clearly documents *why* serialization is needed, which matches the codebase convention for similar acceptance test classes that share a single mutable generated asset.

Dimensions assessed and found clean:
- **Flakiness Patterns (D12):** The file-system race condition is directly eliminated — no shared mutable path accessed concurrently after this fix.
- **Test Isolation (D10):** `[DoNotParallelize]` at the class level is the correct granularity; only the two VSTest methods truly conflict, but serializing the whole 5-method class is harmless and simpler.
- **Algorithmic Correctness (D1):** The root cause is the parallel build race; `[DoNotParallelize]` is a correct, sufficient fix.
- **Scope & PR Discipline (D21):** Single concern, no unrelated changes, clear issue reference in the PR description.
- All other dimensions (API, threading, security, IPC, etc.) are N/A for a one-line test attribute addition.

> Generated by [Expert Code Review (on open)](https://github.com/microsoft/testfx/actions/runs/25964617933/agentic_workflow) for issue #8294 · ● 3.4M

Evangelink:

PR #8311, opened 2026-05-18
Use comparer type short name in Assert.AreAllDistinct failure message

Evangelink:
## Summary

Aligns `Assert.AreAllDistinct` failure messages with the convention used by `Assert.Contains` (which uses `GetType().Name`) so that the printed `comparer:` line shows just the short type name (e.g. `CaseInsensitiveStringComparer`) instead of the fully-qualified name (e.g. `Microsoft.VisualStudio.TestPlatform.TestFramework.UnitTests.AssertTests+CaseInsensitiveStringComparer`).

This also fixes the corresponding unit test expectations in `AssertTests.AreAll.cs` which already assumed the short name.

## Changes

- `src/TestFramework/TestFramework/Assertions/Assert.AreAllDistinct.cs`: both the generic and non-generic `IEqualityComparer` overloads now pass `comparer.GetType().Name` instead of `comparer.GetType().ToString()`.

## Validation

Built `TestFramework.UnitTests` and ran the 23 `AreAllDistinct` tests on net9.0 and net48 — all pass.

### DIFF
--- a/src/TestFramework/TestFramework/Assertions/Assert.AreAllDistinct.cs
+++ b/src/TestFramework/TestFramework/Assertions/Assert.AreAllDistinct.cs
@@ -67,7 +67,7 @@ public sealed partial class Assert
     {
         CheckParameterNotNull(collection, "Assert.AreAllDistinct", "collection");
         CheckParameterNotNull(comparer, "Assert.AreAllDistinct", "comparer");
-        AreAllDistinctImpl(collection, comparer, comparerTypeName: comparer.GetType().ToString(), message, collectionExpression);
+        AreAllDistinctImpl(collection, comparer, comparerTypeName: comparer.GetType().Name, message, collectionExpression);
     }
 
     /// <summary>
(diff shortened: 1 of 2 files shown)


Evangelink (commented):
## Code Review: PR #8311 — Use `Type.Name` instead of `Type.ToString()` for comparer type name

This is a focused, two-line change that switches the comparer type name reported in assertion failure messages from `comparer.GetType().ToString()` to `comparer.GetType().Name`.

---

### ✅ Dimensions assessed as clean

| Dimension | Status |
|---|---|
| Threading & Concurrency | N/A |
| Security / IPC Contract | N/A |
| Public API Surface | N/A — no API surface change |
| Localization | N/A — no resource string change |
| Cross-TFM Correctness | ✅ Both `.ToString()` and `.Name` work on all TFMs |
| Algorithmic Correctness | ✅ |
| Performance | ✅ `.Name` avoids allocating the full assembly-qualified string |
| Style / Conventions | ✅ |

---

### ⚠️ Noteworthy: Information Loss for Non-Simple Types

**Severity: MINOR**

`Type.Name` returns only the simple name (e.g. `MyComparer`), stripping the namespace. `Type.ToString()` returns the fully qualified name including namespace (e.g. `MyProject.MyComparer`).

For error diagnostics this is a tradeoff:
- `Name` is shorter and cleaner for simple types.
- For **generic comparers**, `Name` returns the raw name with a backtick (e.g. `EqualityComparer\`1`) while `ToString()` returns the full instantiation (e.g. `System.Collections.Generic.EqualityComparer\`1[System.String]`), which is actually more readable.
- Namespace context is lost, which could be ambiguous if two comparers from different namespaces share the same simple name.

A potentially better alternative would be `Type.FullName ?? Type.Name` — this preserves namespace info (non-null for most concrete types) while gracefully falling back for edge cases. However, if the intent is simply to make the message less verbose for common cases, the current change is acceptable and the scope is consistent (only affects the failure message `comparer:` line).

**No blocking issues.** The change is safe, the concern is diagnostic quality only.

> Generated by [Expert Code Review (on open)](https://github.com/microsoft/testfx/actions/runs/26019525170/agentic_workflow) for issue #8311 · ● 2.9M

PR #8292, opened 2026-05-16
Fix TelemetryTests flake from parallel dotnet test races

Evangelink:
## Problem

`TelemetryTests.VSTest_RunTests_Succeeds` and `VSTest_DiscoverTests_Succeeds` are parameterised by target framework via `[DynamicData(TargetFrameworks.AllForDynamicData)]` and both invoke:

`dotnet test -c Release {AssetFixture.VSTestProjectPath} --framework {tfm}`

against the **same** shared `vstest/TelemetryVSTestProject.csproj`. With MSTest parallelism enabled, multiple TFM instances run concurrently and race on MSBuild outputs such as `bin/Release/<tfm>/TelemetryVSTestProject.runtimeconfig.json` and shared `obj/` files.

Observed CI failure:

```
error MSB4018: The ""GenerateRuntimeConfigurationFiles"" task failed unexpectedly.
System.IO.IOException: The process cannot access the file
'...\TelemetryTests\vstest\bin\Release\net10.0\TelemetryVSTestProject.runtimeconfig.json'
because it is being used by another process.
```

## Fix

Apply `[DoNotParallelize]` at the `TelemetryTests` class level to serialize all tests in the class. This is a one-line change, preserves test coverage, and only has a small impact on suite duration.

## Blocked PRs

This flake has been intermittently blocking: #8284 #8259 #8234 #8268 #8270 #8256 #8255 #8189 #8262 #8237 #8231

PR #8312, opened 2026-05-18
Pre-build VSTest project and use --no-build for TelemetryTests

Evangelink:
Follow-up to #8292 which only added `[DoNotParallelize]` to `TelemetryTests`.

The `VSTest_RunTests_Succeeds` and `VSTest_DiscoverTests_Succeeds` cases were still flaking on SDK 11 preview (`11.0.100-preview.5.26227.104`) with:

```
error MSB4018: The "GenerateRuntimeConfigurationFiles" task failed unexpectedly.
System.IO.IOException: The process cannot access the file
'...bin/Release/<tfm>/TelemetryVSTestProject.runtimeconfig.json' because it is being used by another process.
```

Even though `[DoNotParallelize]` serializes the test methods inside the class, every `dotnet test` invocation still re-runs the build (and `GenerateRuntimeConfigurationFiles`), and the shared `bin/<tfm>/` outputs were racing with leftover MSBuild / testhost handles across serialized invocations.

### Fix

- Add `EnsureVSTestProjectBuiltAsync(tfm, ct)` on the fixture, gated by a `SemaphoreSlim` + `HashSet<string>` so the VSTest project is built exactly once per TFM. The TFM is only cached after a successful build to avoid poisoning on cancellation/failure.
- Change `VSTest_*` tests to call the pre-build then run `dotnet test ... --no-build --no-restore` so the build/restore targets that write the locked file never run during the actual tests.
- Keep `[DoNotParallelize]` as defense in depth.

### Verification

- Local `dotnet build` of `MSTest.Acceptance.IntegrationTests.csproj` succeeds.
- Functional verification has to happen in CI since this is an intermittent SDK 11 preview race.

### DIFF
--- a/test/IntegrationTests/MSTest.Acceptance.IntegrationTests/TelemetryTests.cs
+++ b/test/IntegrationTests/MSTest.Acceptance.IntegrationTests/TelemetryTests.cs
@@ -129,8 +129,15 @@ public sealed class TelemetryTests : AcceptanceTestBase<TelemetryTests.TestAsset
     [DynamicData(nameof(TargetFrameworks.AllForDynamicData), typeof(TargetFrameworks))]
     public async Task VSTest_RunTests_Succeeds(string tfm)
     {
+        // Pre-build the VSTest project for this TFM exactly once, then run `dotnet test`
+        // with `--no-build --no-restore` so the build/restore targets that write to
+        // `bin/<tfm>/TelemetryVSTestProject.runtimeconfig.json` do not run per-test.
+        // This removes the GenerateRuntimeConfigurationFiles race that was still flaking
+        // intermittently even with [DoNotParallelize] on the class.
+        await AssetFixture.EnsureVSTestProjectBuiltAsync(tfm, TestContext.CancellationToken);
+
         DotnetMuxerResult testResult = await DotnetCli.RunAsync(
-            $"test -c Release {AssetFixture.VSTestProjectPath} --framework {tfm}",
+            $"test -c Release {AssetFixture.VSTestProjectPath} --framework {tfm} --no-build --no-restore",
             workingDirectory: AssetFixture.VSTestProjectPath,
             failIfReturnValueIsNotZero: false,
             cancellationToken: TestContext.CancellationToken);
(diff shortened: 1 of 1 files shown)


test/IntegrationTests/MSTest.Acceptance.IntegrationTests/TelemetryTests.cs
```diff
@@ -203,6 +215,49 @@ public override (string ID, string Name, string Code) GetAssetsToGenerate()
...
+                {
+                    return;
+                }
+
+                DotnetMuxerResult buildResult = await DotnetCli.RunAsync(
+                    $"build -c Release {VSTestProjectPath} --framework {targetFramework}",
+                    workingDirectory: VSTestProjectPath,
+                    cancellationToken: cancellationToken);
+
+                if (buildResult.ExitCode != 0)
```
Evangelink: The `ExitCode != 0` check is dead code. `DotnetCli.RunAsync` uses `failIfReturnValueIsNotZero: true` by default (see `DotnetCli.cs` line 51), so it already throws an `InvalidOperationException` before `EnsureVSTestProjectBuiltAsync` can inspect `buildResult.ExitCode`. The entire `if` block will never execute. Either pass `failIfReturnValueIsNotZero: false` and keep the manual check, or remove the manual check and let the existing throw propagate.

Evangelink (commented):
The fix correctly addresses the `GenerateRuntimeConfigurationFiles` race by pre-building once per TFM via a `SemaphoreSlim`+`HashSet` guard and then passing `--no-build --no-restore`. One logic bug: the `ExitCode != 0` guard in `EnsureVSTestProjectBuiltAsync` is unreachable dead code because `DotnetCli.RunAsync` defaults to `failIfReturnValueIsNotZero: true` and throws before the caller can inspect the exit code. See inline comment.

> Generated by [Expert Code Review (on open)](https://github.com/microsoft/testfx/actions/runs/26020402043/agentic_workflow) for issue #8312 · ● 3.9M

test/IntegrationTests/MSTest.Acceptance.IntegrationTests/TelemetryTests.cs
Evangelink: @copilot address comment

PR #8347, opened 2026-05-19
Pin MicroBuild signing plugin to 5.4.0

Evangelink:
Fixes a pipeline warning emitted by the `Install MicroBuild plugin` step:

> ##[warning]Please use latest plugin version, version 5.4.0 used instead.

### Root cause
`eng/common/core-templates/steps/install-microbuild.yml` defaults `microbuildPluginVersion` to `'latest'`, which is then passed as the `version` input to `MicroBuildSigningPlugin@4`. The task warns when `latest` is used and falls back to `5.4.0`.

### Fix
`eng/common` is owned by Arcade and local edits would be overwritten by automation, so the fix overrides `microbuildPluginVersion: '5.4.0'` from the calling pipeline templates instead.

### Changes
- `azure-pipelines.yml` – both `jobs.yml` invocations (main build and `WindowsSamples`)
- `azure-pipelines-official.yml` – the `templates-official/jobs/jobs.yml` invocation

### Follow-up
If we want this fix to benefit every Arcade consumer, the long-term solution is a PR to `dotnet/arcade` to either bump the default or drop the `version: latest` input.

### DIFF
--- a/azure-pipelines-official.yml
+++ b/azure-pipelines-official.yml
@@ -101,6 +101,7 @@ extends:
       - template: /eng/common/templates-official/jobs/jobs.yml@self
         parameters:
           enableMicrobuild: true
+          microbuildPluginVersion: '5.4.0'
           enablePublishTestResults: true
           testResultsFormat: 'vstest'
           publishingVersion: 4
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -47,6 +47,7 @@ stages:
   - template: /eng/common/templates/jobs/jobs.yml
     parameters:
       enableMicrobuild: true
+      microbuildPluginVersion: '5.4.0'
       enablePublishTestResults: true
       testResultsFormat: 'vstest'
       artifacts:
@@ -313,6 +314,7 @@ stages:
   - template: /eng/common/templates/jobs/jobs.yml
     parameters:
       enableMicrobuild: true
+      microbuildPluginVersion: '5.4.0'
       artifacts:
         publish:
           logs:


Evangelink (commented):
## Review: Pin MicroBuild signing plugin to 5.4.0

This is a clean, minimal CI-only change. All 21 review dimensions are N/A for pure pipeline YAML changes, except the following that are worth briefly noting:

### ✅ Dimensions Assessment

| # | Dimension | Status |
|---|-----------|--------|
| 1 | Algorithmic Correctness | N/A |
| 2 | Threading & Concurrency | N/A |
| 3 | Security & IPC Contract Safety | N/A |
| 4 | API Design & Public Surface | N/A |
| 5 | Backward Compatibility | N/A |
| 6 | Error Handling & Resilience | N/A |
| 7 | Performance & Allocations | N/A |
| 8 | Test Quality | N/A |
| 9 | Localization | N/A |
| 10 | Cross-TFM Correctness | N/A |
| 11 | Resource & Disposal Management | N/A |
| 12 | Logging & Diagnostics | N/A |
| 13 | Code Style & Conventions | ✅ Clean |
| 14 | Naming & Readability | ✅ Clean |
| 15 | Documentation & Comments | N/A |
| 16 | Dependencies & Versioning | ✅ See note |
| 17 | Build & Packaging | ✅ Clean |
| 18 | Configuration & Feature Flags | ✅ Clean |
| 19 | CI/CD Pipeline | ✅ Clean |
| 20 | Scope Discipline | ✅ Clean |
| 21 | Follow-up Tracking | ✅ Noted in PR description |

### Notes

**Dependencies & Versioning (Dim 16):** Pinning to `5.4.0` eliminates the pipeline warning and makes the version explicit and auditable. The version matches exactly what the MicroBuildSigningPlugin task was already falling back to (`5.4.0`), so this is a no-op functionally and only removes the warning. The PR description correctly notes the long-term fix belongs in `dotnet/arcade`.

**Consistency:** All three call sites (`azure-pipelines.yml` ×2 and `azure-pipelines-official.yml` ×1) are updated uniformly — no call site was missed.

**Arcade ownership:** The PR correctly avoids touching `eng/common` (which is managed by Arcade automation), and instead overrides at the caller level. This is the right approach.

### Verdict

✅ **All clear.** The change is correct, minimal, and consistent. No action items.

> Generated by [Expert Code Review (on open)](https://github.com/microsoft/testfx/actions/runs/26084582658/agentic_workflow) for issue #8347 · ● 2.8M

PR #8307, opened 2026-05-17
Fix Assert.That to not evaluate expression twice (re-opens #6699)

Evangelink:
Re-opens [PR #6699](https://github.com/microsoft/testfx/pull/6699) (which was closed without merging).

Fixes `Assert.That` to not evaluate the lambda body twice for the assert and for the message. Concretely, `Assert.That(() => box.GetValueWithSideEffect() != 0)` used to invoke `GetValueWithSideEffect()` once to evaluate the condition, then re-evaluate sub-expressions a second time to build the diagnostic message - producing wrong/confusing values for side-effecting operands. See #6690.

## What's in this branch
- New `EvaluateExpression` / `EvaluateAllSubExpressions` pair that walks the expression tree bottom-up, caches each sub-expression value, and lets `ExtractDetails` read from the cache instead of recompiling and re-invoking.
- New `AssertThatFailedToEvaluate` resource (`<Failed to evaluate>`) for sub-expressions that the cache couldn't resolve.
- Companion tests in `AssertTests.That.cs`.

## Merge notes
This branch was rebased on top of latest `main` (commit 6b0bae4ac). Resolved conflicts:
- `Assert.That.cs`: kept the new `EvaluateExpression` / cache logic from this branch; adopted main's more efficient `StringBuilder.Append(char, int)` overload while preserving the `MaxConsecutiveParentheses` named constant from this branch.
- `FrameworkMessages.resx`: kept `AssertThatFailedToEvaluate` from this branch alongside all the new RFC 012 summary entries from main; `xlf` files regenerated via `dotnet msbuild /t:UpdateXlf`.
- Guarded `[RequiresDynamicCode(...)]` attributes with `#if NET7_0_OR_GREATER` (the attribute is not available on `netstandard2.0` / `net462` targets and the branch was previously failing to build on those).

## Validation
- `dotnet build src/TestFramework/TestFramework/TestFramework.csproj -c Debug` -> clean across all 4 target frameworks (netstandard2.0, net462, net8.0, net9.0).
- `dotnet test test/UnitTests/TestFramework.UnitTests/TestFramework.UnitTests.csproj -- --treenode-filter "/*/*/AssertTests/That*"` -> **256/256 passed** across net8.0, net8.0-windows10.0.18362.0, net9.0, net48.

Co-authored-by: Copilot <223556219+Copilot@users.noreply.github.com>

Issue #6691, opened 2025-10-08
Assert.That is inconsistent in reporting the details of what is being called


nohwnd:
## Describe the bug

When calling method in expression, the output differs when I run an instance method, or static method, it also reports different amount of detail when using instance method on the same type (the test class), or on other type.

Making it hard to identify where the static method comes from, and overreporting on local methods on the test class.

## Steps To Reproduce

```csharp
using System.Diagnostics.CodeAnalysis;

namespace mstest280;

[TestClass]
public sealed class Test1
{
    [TestMethod]
    public void SameType_Instance()
    {
        string animal = "";
        Assert.That(() => GetAnimal() == animal);
    }

    [TestMethod]
    public void SameType_Static()
    {
        string animal = "";
        Assert.That(() => GetAnimal2() == animal);
    }

    [TestMethod]
    public void OtherType_Instance()
    {
        string animal = "";
        var zoo = new Zoo();
        Assert.That(() => zoo.GetAnimal() == animal);
    }

    [TestMethod]
    public void OtherType_Static()
    {
        string animal = "";
        Assert.That(() => Zoo.GetAnimal2() == animal);
    }

    public string GetAnimal()
    {
        return "Giraffe";
    }

    public static string GetAnimal2()
    {
        return "Giraffe";
    }
}

public class Zoo
{
    public string GetAnimal()
    {
        return "Giraffe";
    }

    public static string GetAnimal2()
    {
        return "Giraffe";
    }
}
```

## Expected behavior

We see 
```
this.GetAnimal() <- instance on same type
Test1.GetAnimal2() <- static on same type

zoo.GetAnimal() <- instance on other type
Zoo.GetAnimal2() <- static on other type
```
## Actual behavior

```
  
Failed SameType_Static [52 ms]
  Error Message:
   Assert.That(() => GetAnimal2() == animal) failed.
Details:
  GetAnimal2() = "Giraffe" <- type is not seen
  animal = ""

Failed SameType_Instance [52 ms]
  Error Message:
   Assert.That(() => GetAnimal() == animal) failed.
Details:
  animal = ""
  mstest280.Test1.GetAnimal() = "Giraffe" <- project name is seen


Failed OtherType_Instance [52 ms]
  Error Message:
   Assert.That(() => zoo.GetAnimal() == animal) failed.
Details:
  animal = ""
  zoo.GetAnimal() = "Giraffe" <- correct


Failed OtherType_Static [52 ms]
  Error Message:
[...]

PR #8359, opened 2026-05-19
Assert.That follow-ups from #8306 review

Evangelink:
Follow-up to #8306 addressing three open review threads.

### #3264506474 / #3264506501 — assignment / unary-update analysis lost RHS detail

The merged PR skipped assignment-style `BinaryExpression` nodes (`=`, `+=`, …) and unary-update `UnaryExpression` nodes (`++x`, `x--`, …) entirely during analysis to keep the writable LHS / operand out of `CaptureRewriter`'s non-writable `Block` wrapping. That kept hand-built expression trees compiling but lost diagnostic detail for any reads on the RHS (assignment) or beneath the writable target.

Moved the writable-context guard into `CaptureRewriter` (new `VisitBinary` / `VisitUnary` / `VisitWritableTarget`) and let `AnalyzeExpression` process both sides normally. The new `VisitWritableTarget` leaves the top-level writable node (e.g. `obj.Field` LHS) unwrapped while still recursing into its children so reads beneath the writable target are still captured.

This also robustly handles the rubber-duck-identified reused-node scenario: in a hand-built tree like `Expression.Assign(field, Expression.Multiply(field, c))` where the same `MemberExpression` instance appears in both LHS and RHS, the rewriter discriminates per syntactic position (LHS stays writable, RHS gets wrapped) rather than per instance.

### #3264506519 — Func/Action filter comment didn't match the code

The comment in `EvaluateAndCollectDetails` claimed the Func/Action filter used "runtime type as the existing code did", but `AnalyzeMemberExpression` added a static-type filter that silently dropped members typed as Func/Action holding a `null` value — a behavior regression vs. the original implementation.

Removed the analysis-time static-type filter, kept the runtime-type filter at detail-build time, and clarified the comments. Null Func/Action members now render as `null` again, matching the historical behavior.

## Other threads on #8306 reviewed but not changed

| Comment | Status |
|---|---|
| #3258059677 (line 468) | Already addressed in the merge — `IsSafeToReevaluate` recurses into children. Test `That_ShortCircuitedExpression_DoesNotReevaluateMemberExpressionWithSideEffectingReceiver` covers it. |
| #3258059733 (line 410) | Already addressed — `IsCapturedThis` uses `IsAssignableFrom` so inherited methods on `this` render correctly. Test `That_InheritedInstanceMethodOnThis_RendersAsThis` locks it down. |
| #3258059756 (line 330) | Already addressed — `IsArrayGetMethod` requires `Object.Type.IsArray`, so arbitrary `Get(...)` methods don't render as indexers. Test `That_ArbitraryGetMethod_RendersAsMethodCall_NotIndexer` covers it. |
| #3260908985 (line 375) | Already addressed — extension-method receivers go through `IsCapturedThis`. Test `That_ExtensionMethodOnThis_RendersAsThis` covers it. |
| #3264506539 (line 117) | Deferred. Caching by `ConditionalWeakTable<Expression, …>` is interesting but `Expression<Func<bool>>` lambdas are typically rebuilt per `Assert.That` call, so the benefit is speculative without a benchmark. The existing inline comment already documents it as a future mitigation. |

## Tests

Adds three regression tests in `test/UnitTests/TestFramework.UnitTests/Assertions/AssertTests.That.cs`:

- `That_ManuallyConstructedAssignExpression_CapturesValuesFromRhsAndLhs` — proves the new behavior: a hand-built `box.Value = rhsValue + 1` tree surfaces both `rhsValue` and the post-assignment `box.Value` in the failure details.
- `That_ManuallyConstructedAssignExpression_WithReusedLhsInstance_DoesNotThrow` — locks down the reused-node case where the same `MemberExpression` instance is both LHS and RHS sub-expression.
- `That_WithNullFuncDelegate_RendersAsNullInDetails` — locks down that `Func<…>`-typed members holding `null` render as `null` (regression guard for the comment/code inconsistency).

All 884 `AssertTests` pass locally.

### DIFF
--- a/src/TestFramework/TestFramework/Assertions/Assert.That.cs
+++ b/src/TestFramework/TestFramework/Assertions/Assert.That.cs
@@ -131,8 +131,9 @@ public static partial class AssertExtensions
             return true;
         }
 
-        // Build details using first-occurrence-per-name semantics, filtering out Func/Action values
-        // (matching the historical behavior, using runtime type as the existing code did).
+        // Build details using first-occurrence-per-name semantics. Func/Action values are filtered
+        // out by their RUNTIME type so that a member statically typed as Func/Action but holding a
+        // `null` value still renders as `null` (preserving the historical behavior).
         for (int i = 0; i < context.CaptureNames.Count; i++)
         {
             string name = context.CaptureNames[i];
@@ -196,13 +197,11 @@ public static partial class AssertExtensions
                 AnalyzeArrayIndexExpression(binaryExpr, context);
                 break;
 
-            // Assignment-style binary nodes (=, +=, -=, …) require writable left-hand operands.
-            // CaptureRewriter would replace the LHS with a Block, making it non-writable and
-            // causing the rewritten lambda to fail to compile. Skip these entirely.
-            case BinaryExpression binaryExpr when IsAssignmentNodeType(binaryExpr.NodeType):
-                break;
-
             case BinaryExpression binaryExpr:
+                // For assignment-style nodes (=, +=, -=, …), the LHS must remain writable so the
+                // rewritten lambda still compiles; the writable-context guard in CaptureRewriter
+                // handles that. We still analyze both sides so reads on either side contribute
+                // capture details.
                 AnalyzeExpression(binaryExpr.Left, context, suppressIntermediateValues);
                 AnalyzeExpression(binaryExpr.Right, context, suppressIntermediateValues);
                 break;
@@ -224,12 +223,9 @@ public static partial class AssertExtensions
 
                 break;
 
-            // Unary update nodes (++x, x++, --x, x--) require writable operands. CaptureRewriter
-            // would replace the operand with a Block, making it non-writable and causing compilation
-            // of the rewritten lambda to fail. Skip these entirely.
-            case UnaryExpression unaryExpr when IsUnaryUpdateNodeType(unaryExpr.NodeType):
-                break;
-
+            // Unary update nodes (++x, x++, --x, x--) require a writable operand. The
+            // writable-context guard in CaptureRewriter ensures the operand is left
+            // unwrapped; analyze it so any read sub-expressions still contribute details.
             case UnaryExpression unaryExpr:
                 AnalyzeExpression(unaryExpr.Operand, context, suppressIntermediateValues);
                 break;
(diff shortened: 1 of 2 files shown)


Evangelink:
Closing this PR — it has been superseded by #8307, which landed on `main` after this PR was opened and **completely re-architected** `src/TestFramework/TestFramework/Assertions/Assert.That.cs` (+1258 / −805 on that single file). All of the code paths this PR was touching no longer exist:

| #8359 touches | After #8307 |
|---|---|
| `CaptureRewriter` / `VisitWritableTarget` | Gone — no rewriting at all. The new code uses bottom-up `EvaluateAllSubExpressions` + constant substitution. |
| `AnalyzeExpression` / `IsAssignmentNodeType` / `IsUnaryUpdateNodeType` | Gone — assignment/update concerns are absorbed by the try/catch + `FailedToEvaluateSentinel` fallback. |
| `AnalyzeMemberExpression` static-type Func/Action filter | Gone — the new code uses only the runtime-type filter (`IsFuncOrActionType(cachedValue?.GetType())`) at detail-extraction time, which already preserves the `null` Func rendering behavior this PR was trying to restore. |

Re: the three new review comments from Copilot:
- **#3266169598** (value-type writable receivers in `CaptureRewriter`) — obsoleted; the rewriter class is gone.
- **#3266169663** (struct receiver test) — obsoleted for the same reason.
- **#3266169703 / #3266169743 / #3266169767** (suggesting `[Test]` attributes) — not applicable; this project uses `TestFramework.ForTestingMSTest`'s `TestContainer`, where every public parameterless method is a test (see `test/Utilities/TestFramework.ForTestingMSTest/TestContainer.cs`).

Rebasing this PR would mean re-writing it from scratch against a fundamentally different design that already addresses the same concerns differently, so the cleaner action is to close.

---

**Heads-up unrelated to this PR**: while comparing the two architectures I noticed #8307 regressed two display fixes from #8306 (issue #6691):
1. The arbitrary-`Get(...)`-method special case at [`HandleMethodCallExpression` line 831](https://github.com/microsoft/testfx/blob/main/src/TestFramework/TestFramework/Assertions/Assert.That.cs#L831) now matches any instance method named `Get` — including `box.Get(key)` on non-array types — and renders it as `box[key]`. The original fix scoped this to `callExpr.Object.Type.IsArray`.
2. The `this.Method(...)` / `Type.Method(...)` display rendering (`GetMethodCallDisplayName` + `IsCapturedThis`) is no longer present; non-boolean method calls fall back to raw `callExpr.ToString()` cleanup.

Happy to open a separate PR re-porting those if you'd like — let me know.

Issue #13, opened 2016-09-19
Description Attribute needs to be added in.


AbhitejJohn:
## Description

DescriptionAttribute was in V1 and migrating to V2 causes build failures. This needs to go via an extensibility point though.
## Steps to reproduce
1. Create a MSTest V1 project with a DescriptionAttribute
2. Migrate to MSTest V2.
3. Build project
## Expected behavior

Project build works.
## Actual behavior

Build fails with a cannot find a DescriptionAttribute
## Environment

This should repro across.

PR #8363, opened 2026-05-19
Assert.That: fix Expression.Assign double-evaluation and assorted cleanups

Evangelink:
Fix a single-pass guarantee violation for `Expression.Assign` (and compound assignments) plus a few related cleanups in `Assert.That`. Also addresses leftover review feedback from #8308.

## Bug: `Expression.Assign` ran the RHS twice

Empirical repro on `main`:

```csharp
static int Counter_Calls;
static int ComputeValue() { Counter_Calls++; return 42; }

// container.Inner.Value = ComputeValue() < 0   (fails)
Assert.That(lambda);
// Counter_Calls == 2   ← bug: ComputeValue ran twice
```

### Root cause
The bottom-up evaluator walked Left and Right, cached the RHS value, then went through the generic "rebuild and compile" path. That replaces children with constants — but the Left of an `Assign` is a *writable target*. The rebuilt `Assign(Constant, Constant)` was invalid, the outer `try/catch` sentineled it, and the parent expression's later rebuild fell back to invoking the *original* assignment from scratch — running the RHS side effects a second time.

The same hazard affects every compound assignment (`AddAssign`, `MultiplyAssign`, …) and the unary assignment forms (`PreIncrementAssign`, `PostIncrementAssign`, …).

### Fix
Special-case assignment node types in `EvaluateAllSubExpressions`:

1. Walk Left to capture diagnostics and cache any side-effecting sub-children.
2. Walk Right exactly once.
3. Rebuild Left via `ReplaceSubExpressionsWithConstants` so the writable wrapper (`MemberExpression` / `IndexExpression` / `ParameterExpression`) is preserved while its sub-children become constants. This also fixes double-evaluation of side-effecting Left sub-children (e.g., `provider.GetBox().Value = 42`).
4. Substitute Right with its cached constant and invoke the rebuilt assignment exactly once.

`RequiresSinglePassEvaluation` now also recognizes assignment node types, so a compound assignment whose only side effect is the assignment itself can't slip into the fast path and double-execute.

## Cleanups

- Removed dead `IsSimpleExpression` / `IsVariableReference` — `IsSimpleExpression` only returned `true` for `ConstantExpression`, which is handled before it's called. The cached-value lookup branch in `GetIndexArgumentDisplay` was unreachable. Simplified `GetIndexArgumentDisplay` and dropped its unused `evaluationCache` parameter. Removed the orphaned `TranslateFailureSentinel` helper as well.
- Cached the netstandard/net462 `Regex` instances for the display-class cleanup and list-init cleanup pipelines (previously a fresh compiled `Regex` was allocated per failed assertion). On NET, list-init patterns are now `GeneratedRegex` instead of inline string patterns.
- Reduced substring allocations in `RemoveAnonymousTypeWrappers` / `TryMatchListInitPattern` via a new non-allocating `HasSubstringAt(input, start, value)` helper backed by `string.CompareOrdinal` + bounds.
- Simplified `IsFuncOrActionType` to call `GetGenericTypeDefinition()` at most once.

## PR #8308 review feedback

The Copilot AI review left two comments on #8308 that landed after merge:

1. *"This new test method is missing a test attribute"* — Not actionable: the file uses `TestContainer`, which treats any public parameterless method as a test (`test/Utilities/TestFramework.ForTestingMSTest/TestContainer.cs:6-11`). No attribute needed.
2. *"...sub-expressions are not object-typed"* — Already addressed by the rename to `That_TwoCallsReturningSameMutatedReference_DeduplicatesInDetails` in the merged version. The test exercises the actual scenario its name describes (reference-identity dedup after mutation) regardless of static type.

## Tests

Added three regression tests:
- `That_ManuallyConstructedAssign_EvaluatesRhsExactlyOnce` — the main bug.
- `That_CompoundAssignmentOnField_AppliesExactlyOnceOnFailure` — exercises the `RequiresSinglePassEvaluation` fix for compound assignments without other side-effecting nodes.
- `That_AssignToFieldOfSideEffectingReceiver_EvaluatesReceiverExactlyOnce` — exercises the `ReplaceSubExpressionsWithConstants` Left-rebuild fix for side-effecting receivers.

All 1,157 TestFramework.UnitTests pass on net48 / net8.0 / net9.0.

### DIFF
--- a/src/TestFramework/TestFramework/Assertions/Assert.That.cs
+++ b/src/TestFramework/TestFramework/Assertions/Assert.That.cs
@@ -148,9 +148,16 @@ public static partial class AssertExtensions
     private static bool RequiresSinglePassEvaluation(Expression expr)
         => expr switch
         {
+            // Assignments and compound assignments are side-effecting by definition; ensure they
+            // always take the single-pass evaluator so the fast path doesn't run them once before
+            // EvaluateAllSubExpressions runs them again on the failure-diagnostic path.
+            BinaryExpression binaryExpr when IsAssignmentBinaryNodeType(binaryExpr.NodeType) => true,
+
             BinaryExpression binaryExpr => binaryExpr.Method is not null
                 || RequiresSinglePassEvaluation(binaryExpr.Left)
                 || RequiresSinglePassEvaluation(binaryExpr.Right),
+
+            UnaryExpression unaryExpr when IsAssignmentUnaryNodeType(unaryExpr.NodeType) => true,
 
             UnaryExpression unaryExpr => unaryExpr.Method is not null
                 || RequiresSinglePassEvaluation(unaryExpr.Operand),
@@ -217,6 +224,32 @@ public static partial class AssertExtensions
 
                     break;
 
+                // Assignment-style binary expressions (Assign and compound assignments such as
+                // AddAssign, MultiplyAssign, ...) require special handling. The Left operand is a
+                // writable target — replacing it wholesale with a ConstantExpression during the
+                // generic "rebuild and compile" path produces an invalid Assign whose compilation
+                // throws. The outer try/catch would then sentinel the parent, and a higher-level
+                // rebuild would re-execute the original assignment, running the RHS side effects
+                // a second time. Handle these inline:
+                //   1. Walk Left to capture diagnostics and cache any side-effecting sub-children
+                //      (receiver/index arguments). Reading a property getter on Left runs it once
+                //      as part of capturing the pre-assignment value.
+                //   2. Walk Right exactly once.
+                //   3. Rebuild Left via ReplaceSubExpressionsWithConstants so its sub-children
+                //      (receiver, index arguments) become constants but the writable wrapper node
+                //      (MemberExpression/IndexExpression/ParameterExpression) is preserved.
+                //   4. Substitute Right with its cached constant and invoke the rebuilt assignment
+                //      exactly once. The cached binary value is the post-assignment value (for
+                //      compound assignments it is the result of the compound operation).
+                case BinaryExpression binaryExpr when IsAssignmentBinaryNodeType(binaryExpr.NodeType):
+                    EvaluateAllSubExpressions(binaryExpr.Left, cache);
+                    EvaluateAllSubExpressions(binaryExpr.Right, cache);
+                    Expression rebuiltAssignLeft = ReplaceSubExpressionsWithConstants(binaryExpr.Left, cache);
+                    Expression rebuiltAssignRight = ReplaceChildWithConstant(binaryExpr.Right, cache);
+                    BinaryExpression rebuiltAssign = binaryExpr.Update(rebuiltAssignLeft, binaryExpr.Conversion, rebuiltAssignRight);
+                    cache[binaryExpr] = Expression.Lambda(rebuiltAssign).Compile().DynamicInvoke();
+                    return;
+
                 case BinaryExpression binaryExpr:
                     // Evaluate both operands before evaluating the binary operation
                     EvaluateAllSubExpressions(binaryExpr.Left, cache);
@@ -230,6 +263,19 @@ public static partial class AssertExtensions
                 // break Queryable scenarios. Treat it as a leaf instead.
                 case UnaryExpression unaryExpr when unaryExpr.NodeType == ExpressionType.Quote:
                     break;
+
+                // Unary assignment-style nodes (PreIncrementAssign, PostIncrementAssign, etc.).
+                // The Operand is a writable target, so the generic rebuild path that replaces it
+                // with a ConstantExpression would produce an invalid node. Walk the Operand's
+                // sub-children so receiver/index arguments execute exactly once, then rebuild the
+                // unary with the Operand's sub-children substituted by cached constants — the
+                // writable wrapper node is preserved.
+                case UnaryExpression unaryExpr when IsAssignmentUnaryNodeType(unaryExpr.NodeType):
+                    EvaluateAllSubExpressions(unaryExpr.Operand, cache);
+                    Expression rebuiltUnaryOperand = ReplaceSubExpressionsWithConstants(unaryExpr.Operand, cache);
+                    UnaryExpression rebuiltUnary = unaryExpr.Update(rebuiltUnaryOperand);
+                    cache[unaryExpr] = Expression.Lambda(rebuiltUnary).Compile().DynamicInvoke();
+                    return;
 
                 case UnaryExpression unaryExpr:
                     // Evaluate the operand before evaluating the unary operation
@@ -753,7 +799,7 @@ public static partial class AssertExtensions
     private static void HandleArrayIndexExpression(BinaryExpression arrayIndexExpr, Dictionary<string, object?> details, Dictionary<Expression, object?> evaluationCache)
     {
         string arrayName = GetCleanMemberName(arrayIndexExpr.Left);
-        string indexValue = GetIndexArgumentDisplay(arrayIndexExpr.Right, evaluationCache);
+        string indexValue = GetIndexArgumentDisplay(arrayIndexExpr.Right);
         string indexerDisplay = $"{arrayName}[{indexValue}]";
         TryAddExpressionValue(arrayIndexExpr, indexerDisplay, details, evaluationCache);
 
@@ -820,7 +866,7 @@ public static partial class AssertExtensions
         {
             // Display as "listName[indexValue]" for readability
             string objectName = GetCleanMemberName(callExpr.Object);
-            string indexValue = GetIndexArgumentDisplay(callExpr.Arguments[0], evaluationCache);
+            string indexValue = GetIndexArgumentDisplay(callExpr.Arguments[0]);
             string indexerDisplay = $"{objectName}[{indexValue}]";
             TryAddExpressionValue(callExpr, indexerDisplay, details, evaluationCache);
 
@@ -832,7 +878,7 @@ public static partial class AssertExtensions
         {
             // Handle multi-dimensional array indexers (e.g., array.Get(i, j) displayed as array[i, j])
             string objectName = GetCleanMemberName(callExpr.Object);
-            string indexDisplay = string.Join(", ", callExpr.Arguments.Select(arg => GetIndexArgumentDisplay(arg, evaluationCache)));
+            string indexDisplay = string.Join(", ", callExpr.Arguments.Select(GetIndexArgumentDisplay));
             string indexerDisplay = $"{objectName}[{indexDisplay}]";
             TryAddExpressionValue(callExpr, indexerDisplay, details, evaluationCache);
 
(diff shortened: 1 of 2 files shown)


test/UnitTests/TestFramework.UnitTests/Assertions/AssertTests.That.cs
```diff
@@ -1323,6 +1323,108 @@ public void That_ManuallyConstructedAssignExpression_SurfacesRhsAndPreservesSide
...
+        CountingComputeValue.Calls.Should().Be(1);
+        container.Inner.Value.Should().Be(42);
+    }
+
+    // Regression: a compound assignment whose only side effect is the assignment itself (no
+    // method or property reads triggering single-pass) was taking the side-effect-free fast
+    // path. The fast path ran the assignment once, then EvaluateAllSubExpressions ran it
+    // again on the failure-diagnostic path. RequiresSinglePassEvaluation now recognizes
+    // assignment node types so these expressions always take the single-pass evaluator.
+    public void That_CompoundAssignmentOnField_AppliesExactlyOnceOnFailure()
```
Evangelink, line 1357: **[BLOCKING] Test Completeness — Missing `[TestMethod]`**

`That_CompoundAssignmentOnField_AppliesExactlyOnceOnFailure` is also missing the `[TestMethod]` attribute — it will never be discovered or run by MSTest.

test/UnitTests/TestFramework.UnitTests/Assertions/AssertTests.That.cs
```diff
@@ -1323,6 +1323,108 @@ public void That_ManuallyConstructedAssignExpression_SurfacesRhsAndPreservesSide
         container.Inner.Value.Should().Be(42);
     }
 
+    // Regression for Assign double-evaluation: the RHS of Expression.Assign was evaluated
+    // once during the bottom-up pass, but the generic "rebuild and compile" path then
+    // produced an invalid Assign(Constant, Constant) that the catch sentineled, and the
+    // parent LessThan was re-invoked from the original tree — re-running ComputeValue a
+    // second time. Assign is now special-cased so the RHS runs exactly once.
+    public void That_ManuallyConstructedAssign_EvaluatesRhsExactlyOnce()
```
Evangelink, line 1331: **[BLOCKING] Test Completeness — Missing `[TestMethod]`**

This method has no `[TestMethod]` attribute (same applies to `That_CompoundAssignmentOnField_AppliesExactlyOnceOnFailure` and `That_AssignToFieldOfSideEffectingReceiver_EvaluatesReceiverExactlyOnce`).

Without `[TestMethod]`, MSTest will never discover or execute any of the three new regression tests. The regressions they guard against are completely unprotected — reverting the production fix in `Assert.That.cs` would not be caught by CI.

**Recommendation:** Add `[TestMethod]` above each of the three method declarations:
```csharp
[TestMethod]
public void That_ManuallyConstructedAssign_EvaluatesRhsExactlyOnce()
```

test/UnitTests/TestFramework.UnitTests/Assertions/AssertTests.That.cs
```diff
@@ -1323,6 +1323,108 @@ public void That_ManuallyConstructedAssignExpression_SurfacesRhsAndPreservesSide
...
+        // Exactly one += 7 must apply — not two.
+        container.Inner.Value.Should().Be(7);
+    }
+
+    // Regression: assigning to a member whose receiver chain has a side effect (e.g.,
+    // `provider.GetBox().Value = 42`) previously ran the receiver method twice because the
+    // rebuild kept the original (unevaluated) Left in place. The Left's sub-children are
+    // now substituted with cached constants while the writable MemberExpression wrapper is
+    // preserved, so the receiver runs exactly once.
+    public void That_AssignToFieldOfSideEffectingReceiver_EvaluatesReceiverExactlyOnce()
```
Evangelink, line 1381: **[BLOCKING] Test Completeness — Missing `[TestMethod]`**

`That_AssignToFieldOfSideEffectingReceiver_EvaluatesReceiverExactlyOnce` is also missing the `[TestMethod]` attribute — it will never be discovered or run by MSTest.

test/UnitTests/TestFramework.UnitTests/Assertions/AssertTests.That.cs
```diff
@@ -1323,6 +1323,108 @@ public void That_ManuallyConstructedAssignExpression_SurfacesRhsAndPreservesSide
...
+        BinaryExpression body = Expression.LessThan(assign, Expression.Constant(0));
+        var lambda = Expression.Lambda<Func<bool>>(body);
+
+        Action act = () => Assert.That(lambda);
+        act.Should().Throw<AssertFailedException>();
+
+        // Receiver method (GetBox) must run exactly once.
+        provider.Calls.Should().Be(1);
+        provider.Box.Value.Should().Be(42);
+    }
```
Evangelink, line 1399: **[MAJOR] Test Isolation — Static mutable counter**

`CountingComputeValue.Calls` is a static mutable field reset with `Calls = 0` inside `That_ManuallyConstructedAssign_EvaluatesRhsExactlyOnce`. If this test class ever runs with parallelism enabled (e.g., via `[Parallelize]` at assembly level or via a future `.runsettings` change), the reset and the assertion `Calls.Should().Be(1)` are not atomic, and another test's `Get()` call could corrupt the count.

Also, `Calls++` is a non-atomic read-modify-write; concurrent increments (even from a single test if the expression tree were ever evaluated on multiple threads) could silently undercount.

**Recommendation:** Move the counter to an instance field, or use `Interlocked.Increment` and `Interlocked.Exchange` for the increment and reset:
```csharp
// Reset
Interlocked.Exchange(ref CountingComputeValue.Calls, 0);
// Increment
Interlocked.Increment(ref CountingComputeValue.Calls);
```
Or, simpler, make `CountingComputeValue` non-static and instantiate it fresh per test.

Evangelink (commented):
| # | Dimension | Verdict |
|---|-----------|---------|
| 13 | Test Completeness | 🔴 BLOCKING — 3 tests missing `[TestMethod]` |
| 10 | Test Isolation | 🟡 MAJOR — static mutable `CountingComputeValue.Calls` |

✅ 19/21 dimensions clean.

- [ ] **Test Completeness (#13)** — `That_ManuallyConstructedAssign_EvaluatesRhsExactlyOnce`, `That_CompoundAssignmentOnField_AppliesExactlyOnceOnFailure`, and `That_AssignToFieldOfSideEffectingReceiver_EvaluatesReceiverExactlyOnce` all lack `[TestMethod]`. MSTest will never discover them; the three regression scenarios they document are completely unguarded in CI.
- [ ] **Test Isolation (#10)** — `CountingComputeValue.Calls` is a non-atomic static mutable field. Reset (`= 0`) and assertion (`Should().Be(1)`) are not atomic under parallel execution. Prefer an instance-based counter or `Interlocked` operations.

The production-side changes (assignment special-casing in `EvaluateAllSubExpressions`, `RequiresSinglePassEvaluation`, `HasSubstringAt`, regex caching) look correct and well-reasoned.

> Generated by [Expert Code Review (on open)](https://github.com/microsoft/testfx/actions/runs/26107192476/agentic_workflow) for issue #8363 · ● 19.5M

src/TestFramework/TestFramework/Assertions/Assert.That.cs
```diff
@@ -217,6 +224,32 @@ private static void EvaluateAllSubExpressions(Expression expr, Dictionary<Expres
...
+                //      exactly once. The cached binary value is the post-assignment value (for
+                //      compound assignments it is the result of the compound operation).
+                case BinaryExpression binaryExpr when IsAssignmentBinaryNodeType(binaryExpr.NodeType):
+                    EvaluateAllSubExpressions(binaryExpr.Left, cache);
+                    EvaluateAllSubExpressions(binaryExpr.Right, cache);
+                    Expression rebuiltAssignLeft = ReplaceSubExpressionsWithConstants(binaryExpr.Left, cache);
+                    Expression rebuiltAssignRight = ReplaceChildWithConstant(binaryExpr.Right, cache);
+                    BinaryExpression rebuiltAssign = binaryExpr.Update(rebuiltAssignLeft, binaryExpr.Conversion, rebuiltAssignRight);
+                    cache[binaryExpr] = Expression.Lambda(rebuiltAssign).Compile().DynamicInvoke();
+                    return;
```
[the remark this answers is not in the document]
Evangelink: Addressed in bc6378f36. The bottom-up walker no longer evaluates the writable wrapper of an assignment LHS:

- New helper `EvaluateAssignmentTargetSubChildren` walks only the receiver / index arguments for `MemberExpression` and `IndexExpression` LHS, and is a no-op for `ParameterExpression`.
- The writable wrapper itself is left to the rebuilt assignment, so a compound assignment on a property invokes the getter exactly once (inside the rebuilt op), and a plain `Assign` on a property invokes it zero times.
- The new value is cached on `Left` (for binary assignments) and on Pre*Assign `Operand` (where the result equals the post-op operand value) so diagnostic rendering doesn't trigger an extra getter call. Post* variants intentionally don't cache on Operand to avoid a second getter call.

Two new regression tests guard this exact scenario:
- `That_AssignToProperty_DoesNotInvokeGetter` — plain Assign on a property: setter 1, getter 0.
- `That_CompoundAssignToProperty_InvokesGetterExactlyOnce` — AddAssign on a property: getter 1, setter 1.

test/UnitTests/TestFramework.UnitTests/Assertions/AssertTests.That.cs
Evangelink: Not actionable here — `AssertTests` derives from `TestContainer` (see `test/Utilities/TestFramework.ForTestingMSTest/TestContainer.cs:6-11`), which treats every public parameterless instance method as a test. None of the surrounding `That_*` tests in this file carry `[TestMethod]` either, and `dotnet test` discovers and runs all three of the new methods:

```
passed That_ManuallyConstructedAssign_EvaluatesRhsExactlyOnce
passed That_CompoundAssignmentOnField_AppliesExactlyOnceOnFailure
passed That_AssignToFieldOfSideEffectingReceiver_EvaluatesReceiverExactlyOnce
```

Adding `[TestMethod]` here would actually be inconsistent with the rest of the file's convention.

test/UnitTests/TestFramework.UnitTests/Assertions/AssertTests.That.cs
Evangelink: Addressed in bc6378f36. `CountingComputeValue` is now an instance class with an instance `Calls` property and an instance `Get()` method, instantiated fresh per test. The test wires it into the expression tree via `Expression.Call(Expression.Constant(counter), ...)`, so there is no shared mutable state across tests. The `#pragma warning disable SA1401` suppression is gone as well.

test/UnitTests/TestFramework.UnitTests/Assertions/AssertTests.That.cs
Evangelink: Not actionable here — `AssertTests` derives from `TestContainer` (see `test/Utilities/TestFramework.ForTestingMSTest/TestContainer.cs:6-11`), which treats every public parameterless instance method as a test. None of the surrounding `That_*` tests in this file carry `[TestMethod]` either, and `dotnet test` discovers and runs all three of the new methods:

```
passed That_ManuallyConstructedAssign_EvaluatesRhsExactlyOnce
passed That_CompoundAssignmentOnField_AppliesExactlyOnceOnFailure
passed That_AssignToFieldOfSideEffectingReceiver_EvaluatesReceiverExactlyOnce
```

Adding `[TestMethod]` here would actually be inconsistent with the rest of the file's convention.

test/UnitTests/TestFramework.UnitTests/Assertions/AssertTests.That.cs
Evangelink: Not actionable here — `AssertTests` derives from `TestContainer` (see `test/Utilities/TestFramework.ForTestingMSTest/TestContainer.cs:6-11`), which treats every public parameterless instance method as a test. None of the surrounding `That_*` tests in this file carry `[TestMethod]` either, and `dotnet test` discovers and runs all three of the new methods:

```
passed That_ManuallyConstructedAssign_EvaluatesRhsExactlyOnce
passed That_CompoundAssignmentOnField_AppliesExactlyOnceOnFailure
passed That_AssignToFieldOfSideEffectingReceiver_EvaluatesReceiverExactlyOnce
```

Adding `[TestMethod]` here would actually be inconsistent with the rest of the file's convention.

drognanar (approved):

YuliiaKovalova (approved):

PR #8395, opened 2026-05-20
Fix MSTEST0058: cover throw new AssertFailedException/InconclusiveException and exempt Assert.Inconclusive in filtered catches

### LINKED ISSUE #7340: MSTEST0058 false positive
## Describe the bug

When using a `[DoesNotReturn]` Assertion like `Assert.Fail`, MSTEST0058 is still triggered.

## Version used

MSTest.Sdk 4.1

## Steps To Reproduce

```
try
{
    (await Assert.ThrowsExactlyAsync<UnauthorizedAccessException>(() => System.IO.File.ReadAllTextAsync(share.Combine("Test.txt"), ct)))
        .Message.ShouldBe($@"Access to the path '{share.Combine("Test.txt")}' is denied.");
}
catch (AssertFailedException ex) when (ex.Message.Contains("no exception was thrown.", StringComparison.Ordinal))
{
#pragma warning disable MSTEST0058 // Avoid assertions in catch blocks
    Assert.Inconclusive("We should not have access to this path. This suggests the share is still mounted and does not disconnect (until a system reboot).");
#pragma warning restore MSTEST0058
}
```

## Expected behavior
no warning when catch always throws assertion exception.

## Actual behavior
warning MSTEST0058: Do not use asserts in catch blocks because they may not fail the test if no exception is thrown (https://learn.microsoft.com/dotnet/core/testing/mstest-analyzers/mstest0058)

Evangelink:
Resolves #7340.

## Summary

Two improvements to the `AvoidAssertsInCatchBlocks` analyzer (MSTEST0058) that together address the false positive in #7340 without losing real coverage:

1. **Cover the desugared `throw` form.** Currently the analyzer only catches `Assert.*` invocations, so users can bypass it by writing `throw new AssertFailedException(...)` (or `AssertInconclusiveException`) directly. The PR registers an additional `OperationKind.Throw` handler that flags object-creation throws of these types (and derived types). Bare `throw;` rethrows and `throw ex;` rethrows are explicitly not flagged because they just propagate the caught exception and don't introduce a new MSTest outcome at the catch site.

2. **Exempt `Assert.Inconclusive` (and `throw new AssertInconclusiveException`) in filtered catches.** `catch (...) when (...) { Assert.Inconclusive(...); }` is the **only** way to demote a caught failure to an `Inconclusive` outcome — the `when` filter expresses the demotion condition explicitly, so there is no silent fall-through. Unfiltered catches keep the warning. `Assert.Fail` / `throw new AssertFailedException` remain flagged in every shape because MSTest already converts any unhandled test-method exception to `Failed`; there is no unique value `Assert.Fail` can provide from a catch.

## Resulting rule shape

| Pattern in `catch` | Flagged? |
|---|---|
| `Assert.Fail(…)` | ✅ Yes |
| `throw new AssertFailedException(…)` (or derived) | ✅ Yes (new) |
| `Assert.Inconclusive(…)` in `catch (…)` (no filter) | ✅ Yes |
| `throw new AssertInconclusiveException(…)` in `catch (…)` (no filter) | ✅ Yes (new) |
| `Assert.Inconclusive(…)` in `catch (…) when (…)` | ❌ Exempt (new) |
| `throw new AssertInconclusiveException(…)` in `catch (…) when (…)` | ❌ Exempt (new) |
| `throw;` rethrow | ❌ Not flagged |
| `throw ex;` rethrow of caught variable | ❌ Not flagged |
| Any other `Assert.*` / `CollectionAssert.*` / `StringAssert.*` | ✅ Yes (unchanged) |

For the rare "custom runtime-built message" case where users want to wrap a caught exception with extra context, `#pragma warning disable MSTEST0058` is the documented escape hatch — it keeps the intentional bypass visible at the call site.

## Changes

- **Analyzer**: `src/Analyzers/MSTest.Analyzers/AvoidAssertsInCatchBlocksAnalyzer.cs`
  - Add `OperationKind.Throw` registration → `AnalyzeThrow` that unwraps implicit `IConversionOperation` and inspects the constructed `IObjectCreationOperation`.
  - Use `ITypeSymbol.DerivesFrom` so user types deriving from `AssertFailedException` / `AssertInconclusiveException` are also covered.
  - `IsInsideCatchClause` now returns the enclosing `ICatchClauseOperation` (with `[NotNullWhen(true)]`) so the filtered-catch check can be done with one walk.
- **Well-known types**: `src/Analyzers/MSTest.Analyzers/Helpers/WellKnownTypeNames.cs` adds entries for the two exception types.
- **Resource update**: `Resources.resx` `AvoidAssertsInCatchBlocksDescription` extended to document the `throw new …` coverage, the filtered-catch exemption, and the pragma escape hatch. Regenerated all 13 `.xlf` locales via `dotnet msbuild /t:UpdateXlf`.
- **Tests**: `test/UnitTests/MSTest.Analyzers.UnitTests/AvoidAssertsInCatchBlocksAnalyzerTests.cs` gains 9 new test methods covering:
  - `Assert.Inconclusive` in filtered catch → no diagnostic
  - `Assert.Inconclusive` in unfiltered catch → diagnostic
  - `Assert.Fail` in filtered catch → diagnostic (no special treatment)
  - `throw new AssertFailedException` in catch → diagnostic
  - `throw new AssertInconclusiveException` in unfiltered catch → diagnostic
  - `throw new AssertInconclusiveException` in filtered catch → no diagnostic
  - `throw;` rethrow in catch → no diagnostic
  - `throw ex;` (rethrow of caught variable) → no diagnostic
  - `throw new AssertFailedException` outside any catch → no diagnostic
  - `throw new MyAssertFailedException` (derived) in catch → diagnostic

## Validation

- Analyzer project builds clean (0 warnings, 0 errors).
- The 17 pre-existing tests plus the 9 new tests are wired up via the same `VerifyCS.VerifyCodeFixAsync(code, code)` pattern used elsewhere in the file. (My local environment is overloaded today — was hitting 30–50 min per `dotnet build` cycle on the test project, so please re-run `MSTest.Analyzers.UnitTests` in CI / a clean environment to confirm; the analyzer change is small and the markers in the new tests were set against the actual diagnostic spans reported by the test verifier.)


src/Analyzers/MSTest.Analyzers/AvoidAssertsInCatchBlocksAnalyzer.cs
```diff
@@ -48,22 +48,38 @@ public override void Initialize(AnalysisContext context)
...
     }
 
-    private static void AnalyzeOperation(
+    private static void AnalyzeInvocation(
         OperationAnalysisContext context,
         INamedTypeSymbol? assertSymbol,
         INamedTypeSymbol? stringAssertSymbol,
-        INamedTypeSymbol? collectionAssertSymbol)
+        INamedTypeSymbol? collectionAssertSymbol,
+        INamedTypeSymbol? assertInconclusiveExceptionSymbol)
```
Evangelink, line 78: **Analyzer Quality — ISSUE: unused parameter triggers IDE0060/CA1801**

`assertInconclusiveExceptionSymbol` is accepted by `AnalyzeInvocation` but is never read inside the method. The filtered-catch exemption check on line 101 delegates to `IsAssertInconclusiveCall(targetMethod, assertSymbol)`, which only uses the `Assert` *class* symbol — not the exception symbol — so the parameter is dead code.

**Concrete failing scenario:** in a CI build that runs with `-warnaserror` (the repo's `eng/common/tools.ps1` passes `/warnaserror`), the IDE0060 diagnostic on this private static method is not suppressed by `UnusedParameterSuppressor` (that suppressor is scoped exclusively to `TestContext` parameters on lifecycle methods), so the build will fail with:

```
error IDE0060: Remove unused parameter 'assertInconclusiveExceptionSymbol' if it is not part of a shipped public API
```

**Fix:** remove the parameter from `AnalyzeInvocation` and drop it from the lambda capture in `Initialize`:

```csharp
// Initialize lambda — drop assertInconclusiveExceptionSymbol from the capture:
context => AnalyzeInvocation(context, assertSymbol, stringAssertSymbol, collectionAssertSymbol),

// AnalyzeInvocation signature — remove last parameter:
private static void AnalyzeInvocation(
    OperationAnalysisContext context,
    INamedTypeSymbol? assertSymbol,
    INamedTypeSymbol? stringAssertSymbol,
    INamedTypeSymbol? collectionAssertSymbol)
```

src/Analyzers/MSTest.Analyzers/AvoidAssertsInCatchBlocksAnalyzer.cs
```diff
@@ -74,26 +90,90 @@ private static void AnalyzeOperation(
             return;
         }
 
-        // Walk up the operation tree to check if we're inside a catch clause
-        if (IsInsideCatchClause(operation))
+        if (!IsInsideCatchClause(operation, out ICatchClauseOperation? enclosingCatch))
+        {
+            return;
```
Evangelink, line 95: **[SUGGESTION] Unused Parameter**

`assertInconclusiveExceptionSymbol` is captured by the closure in `Initialize` and forwarded into `AnalyzeInvocation`, but the method body never reads it — the filtered-catch exemption calls `IsAssertInconclusiveCall(targetMethod, assertSymbol)`, which only needs the `Assert` class symbol.

**Recommendation:** Remove the parameter from the `AnalyzeInvocation` signature and the lambda capture:
```csharp
// In Initialize:
context => AnalyzeInvocation(context, assertSymbol, stringAssertSymbol, collectionAssertSymbol),

// Signature:
private static void AnalyzeInvocation(
    OperationAnalysisContext context,
    INamedTypeSymbol? assertSymbol,
    INamedTypeSymbol? stringAssertSymbol,
    INamedTypeSymbol? collectionAssertSymbol)
```

test/UnitTests/MSTest.Analyzers.UnitTests/AvoidAssertsInCatchBlocksAnalyzerTests.cs
```diff
@@ -294,4 +294,296 @@ public void TestMethod()
...
+            public class MyTestClass
+            {
+                [TestMethod]
+                public void TestMethod()
+                {
+                    try
+                    {
+                        // code
+                    }
+                    catch (Exception ex) when (ex.Message.Contains("foo"))
```
Evangelink, line 376: **[SUGGESTION] Missing Test Coverage**

The test suite covers `Assert.Fail` in a filtered catch (→ diagnostic) and `throw new AssertFailedException` in an *unfiltered* catch (→ diagnostic), but there is no test for:

```csharp
catch (Exception ex) when (someCondition)
{
    throw new AssertFailedException("should still be flagged");  // ← no test
}
```

This is the one code path in `AnalyzeThrow` that exercises the asymmetry between `AssertFailedException` (always flagged) and `AssertInconclusiveException` (exempt in filtered catches). A regression that accidentally made AFE exempt in filtered catches would go undetected.

**Recommendation:** Add one test `ThrowNewAssertFailedExceptionInFilteredCatch_Diagnostic` covering this shape.

Evangelink (commented):
**Analyzer Quality — 1 ISSUE**

`AnalyzeInvocation` accepts `assertInconclusiveExceptionSymbol` but never uses it. The inline comment points to the exact fix. Everything else (registration guards, `IsInsideCatchClause` parent-walk, `AnalyzeThrow` unwrapping, `DerivesFrom` check, filtered-catch exemption logic) is correct.

> Generated by [Expert Code Review (on open)](https://github.com/microsoft/testfx/actions/runs/26172599389/agentic_workflow) for issue #8395 · ● 26.8M

Issue #7249, opened 2026-01-17
[Proposal]: MTP Command-line aliases


Youssef1313:
## Summary

This is a proposal to allow easier migration from VSTest to MTP, by introducing a systematic way of mapping command-line options. The detailed proposal below is more technical, but the less technical summary is that this feature allows us to get back VSTest options such as `--logger` and `--collect`, but only when we know that logger or datacollector can be handled in MTP. External owners of VSTest loggers and data collectors can decide to maintain the VSTest syntax, but they definitely will need to implement the extension for MTP. The amount of VSTest extension owners is definitely much less than the whole user space of testing, so this can be manageable.

## Background and Motivation

Today, we still have many users on VSTest, and we want to make the migration to MTP as smooth as possible. This proposal will serve the purpose of letting `--logger trx`, `--collect CodeCoverage`, etc to remain working, while still producing meaningful errors for unsupported loggers and data collectors.

## Proposed Feature

In MTP command-line extensibility, we have the concept of `CommandLineOption`. Extensions (including test frameworks) can define their own command-line options, and it's an error to define the same command-line option by multiple extensions as it results in an ambiguity.

The proposal here is to have a different concept, `CommandLineOptionAlias`.

- A command-line option alias, in the initial design, will be forced to be single-arity.
- Multiple extensions can register the same command-line option alias `(e.g, `--logger`).
- A command-line option alias will have a `CanHandle` method. Only one of the registered aliases can return true. Otherwise, it's an error.
- If none of the registered alises can handle the given argument, that is also an error.
- It's an error to have a command-line option with the same name as any registered alias.
- An alias will have a `Handle` method that returns a list of option/value which represents the transformation of that alias.
- Command line aliases are likely needed to be evaluated very early. Hence they won't have access to any services. They will be for the purpose of systematic syntactic transformations.

An example usage for that can be as follows:

```csharp
testApplicationBuilder.CommandLineAliases.AddProvider(new TrxLoggerProvider());

internal sealed class TrxLoggerProvider : ICommandLineAlias
{
    public override string AliasName => "logger";

    public bool CanHandle(string[] arguments)
    {
        return arguments[0] == "trx" || arguments[0].StartsWith("trx;");
    }

    public IEnumerable<CommandLineAliasTransformation> Handle(string[] arguments)
    {
        if (arguments[0] == "trx") return [new CommandLineAliasTransformation("report-trx", Array.Empty<string>())];

        // Handle cases like trx;LogFileName=...
[...]

PR #8501, opened 2026-05-22
RFC 014 - Command-line option mappings

Evangelink:
Adds RFC 014 proposing **command-line option mappings**: a new extensibility point that lets an extension declaratively accept a user-facing option (e.g. `--logger trx`, `--collect ""XPlat Code Coverage""`) and rewrite it, at parse time, into one or more first-class MTP options.

## Why

The primary scenario is easing migration from VSTest to MTP. Today, `dotnet test --logger trx --collect ""XPlat Code Coverage""` produces `Unknown option` against MTP — correct but unhelpful for the very large body of CI pipelines and tribal knowledge built around VSTest syntax.

The existing `ICommandLineOptionsProvider` contract intentionally forbids two extensions from registering the same option name. `--logger` and `--collect` are however intrinsically polyvalent: the *value* selects which extension owns the option. A new concept is therefore needed.

## What's in the RFC

- Naming rationale (`Mapping` over `Alias`/`Transformation`/`Shim`/`Legacy…`), explicitly addressing the alias-vs-rewrite discussion from #7249.
- Public API surface: `ICommandLineOptionMappingProvider`, `CommandLineOptionMapping`, `CommandLineOptionMapper` delegate (`TryHandle` style), `CommandLineOptionMappingResult`.
- Resolution algorithm: mappings run after response-file expansion, before validation. Zero claimants → error. Multiple claimants → error. Exactly one → rewrite.
- Worked examples for TRX (including `trx;LogFileName=…`) and Code Coverage.
- Cross-cutting validation: mappings cannot collide with canonical option names, cannot rewrite to unknown options, cannot rewrite to other mappings (no loops).
- Compatibility analysis, 5 rejected alternatives, 5-phase rollout, 5 unresolved questions.

## Discussion seeds

Specific points where reviewer input would be most valuable:

1. **Naming** — keep `Mapping`, or switch back to `Alias` (familiar) or `Transformation` (most accurate)?
2. **Help layout** — separate ""VSTest-style options"" section, or interleave with canonical options?
3. **Warning channel** — should `TryMap` get an `out List<string> warnings` parameter in v1, or defer?
4. **Telemetry** — log only `used_compat_mapping: bool`, or include allowlisted values?

Refs #7249 — this PR provides the design artefact requested by the issue; implementation will follow in a separate PR once the RFC is approved.

### DIFF
--- a/docs/RFCs/014-Command-Line-Option-Mappings.md
+++ b/docs/RFCs/014-Command-Line-Option-Mappings.md
@@ -1,0 +1,354 @@
+# RFC 014 - Command-line option mappings
+
+- [ ] Approved in principle
+- [ ] Under discussion
+- [ ] Implementation
+- [ ] Shipped
+
+## Summary
+
+Introduce **command-line option mappings**: a new extensibility point in Microsoft.Testing.Platform (MTP) that lets an extension declaratively accept a user-facing option (e.g. `--logger trx`, `--collect "XPlat Code Coverage"`) and rewrite it, at parse time, into one or more first-class MTP options (e.g. `--report-trx`, `--coverage`). Mappings differ from regular `CommandLineOption` instances in two ways: multiple extensions are allowed to register the *same* mapping name, and exactly one of them is expected to claim responsibility for a given argument value. The primary scenario is making the migration from VSTest to MTP feel less abrupt without polluting the canonical MTP option set.
+
+## Motivation
+
+### The migration pain point
+
+VSTest has long-established, user-facing options that fan out to many independent extensions:
+
+```text
+dotnet test --logger trx --logger "console;verbosity=detailed" --collect "XPlat Code Coverage" --collect "blame"
+```
+
+In MTP, each of these capabilities is exposed by a distinct extension that registers its own option(s):
+
+| VSTest | MTP equivalent |
+|---|---|
+| `--logger trx` | `--report-trx` (`Microsoft.Testing.Extensions.TrxReport`) |
+| `--logger "console;verbosity=detailed"` | `--output detailed` (`Microsoft.Testing.Platform`) |
+| `--collect "XPlat Code Coverage"` | `--coverage` (`Microsoft.Testing.Extensions.CodeCoverage`) |
+| `--collect "blame"` | `--crashdump` / `--hangdump` (`Microsoft.Testing.Extensions.CrashDump` / `…HangDump`) |
+
+This is a deliberate design choice: each capability is its own MTP extension with its own option schema, its own validation, and its own help entry. It is also a real source of friction for the very large population of users who already type `--logger` and `--collect` from muscle memory, who have CI pipelines that pass these flags, and who follow blog posts and Stack Overflow answers that were written for VSTest.
+
+Today, MTP responds to these flags with `Unknown option '--logger'`. That answer is correct but unhelpful.
+
+### Why this can't be solved with regular `CommandLineOption`
+
+The existing `ICommandLineOptionsProvider` contract forbids two extensions from registering the same option name (`CommandLineOptionsValidator` treats duplicate names as a fatal error). That rule is essential: it guarantees that for any option, there is exactly one provider responsible for its validation, its argument arity, and the meaning of `ICommandLineOptions.TryGetOptionArgumentList`.
+
+But `--logger` and `--collect` are intrinsically polyvalent: the *value* selects which extension is responsible. There is no extension that owns `--logger` as a whole — there is one extension that owns `--logger trx`, another that would own `--logger console`, etc. The single-owner-per-option rule that protects normal options is exactly what blocks expressing this.
+
+### Why not just add `--logger` to a single extension
+
+A naive workaround — making, say, the TRX extension own `--logger` and dispatch by value — collapses the moment a second VSTest-compatible extension wants to participate. It also leaks an unrelated extension's identity (TRX) into the dispatcher for unrelated values (`console`, `html`). The mapping mechanism described below is the smallest surface that lets multiple extensions cooperate without breaking the single-owner invariant for canonical MTP options.
+
+## Naming
+
+Earlier discussion (issue [#7249](https://github.com/microsoft/testfx/issues/7249)) proposed the term **alias**. Feedback noted that "alias" in CLI conventions usually means a pure rename (`-v` ≡ `--verbose`) with identical arity and semantics, whereas this feature actually rewrites one option into one *or more* options with potentially different arity. Shell aliases (`alias ll='ls -la'`) and Git aliases are precedents for the broader meaning, but the ambiguity is real for API consumers.
+
+This RFC uses **mapping** in the public API surface (`ICommandLineOptionMapping`, `CommandLineOptionMapping`, `CommandLineMappings`). Rationale:
+
+- "Mapping" honestly conveys *one user-facing token mapped to one or more canonical options*.
+- It avoids the rename connotation of "alias".
+- It avoids the compiler-pass connotation of "transformation".
+- It avoids the deprecation connotation of "shim" or "legacy", leaving room for non-VSTest uses.
+
+User-facing documentation may still refer to these informally as "compatibility aliases" or "VSTest-style options" — the public API name does not constrain documentation prose.
+
+## Design principles
+
+1. **Mappings never replace canonical options.** Every capability addressable through a mapping is *also* addressable through its canonical MTP option. Mappings are purely additive sugar.
+2. **Mappings are resolved before any option provider sees the command line.** A mapping turns `--logger trx` into `--report-trx` *before* `ICommandLineOptionsProvider.ValidateOptionArgumentsAsync` is called on any provider. Providers therefore never need to know that mappings exist.
+3. **Exactly one mapping handles each occurrence.** Zero handlers → error. Multiple handlers → error. There is no "first match wins" silent ambiguity.
+4. **Mappings cannot loop.** A mapping rewrites to canonical options only, never to other mapping names. This is a property checked at registration / validation time.
+5. **Mappings have no service dependencies.** They run at command-line parse time, before the DI container is built. The contract is intentionally narrow.
+6. **Mappings are opt-in for extension authors.** Existing extensions are unaffected. A new MTP project that does not register any mapping behaves exactly as today.
+
+## Detailed design
+
+### Public API surface
+
+A new namespace under the existing command-line extensibility surface:
+
+```csharp
+namespace Microsoft.Testing.Platform.Extensions.CommandLine;
+
+/// <summary>
+/// Provides one or more <see cref="CommandLineOptionMapping"/> entries that rewrite a user-facing
+/// option name into one or more canonical MTP command-line options at parse time.
+/// </summary>
+public interface ICommandLineOptionMappingProvider : IExtension
+{
+    IReadOnlyCollection<CommandLineOptionMapping> GetCommandLineOptionMappings();
+}
+
+/// <summary>
+/// A single mapping registration. Multiple providers may register mappings with the same
+/// <see cref="Name"/>; exactly one of them is expected to return true from
+/// <see cref="TryMap"/> for any given argument list.
+/// </summary>
+public sealed class CommandLineOptionMapping
+{
+    public CommandLineOptionMapping(
+        string name,
+        string description,
+        ArgumentArity arity,
+        CommandLineOptionMapper map);
+
+    /// <summary>The user-facing option name (without leading dashes), e.g. "logger" or "collect".</summary>
+    public string Name { get; }
+
+    /// <summary>Help-text description. Multiple providers registering the same name MAY have different descriptions; the help renderer concatenates them.</summary>
+    public string Description { get; }
+
+    /// <summary>Arity of the user-facing option (typically <see cref="ArgumentArity.ExactlyOne"/>).</summary>
+    public ArgumentArity Arity { get; }
+
+    /// <summary>The rewriter delegate. See <see cref="CommandLineOptionMapper"/>.</summary>
+    public CommandLineOptionMapper Map { get; }
+}
+
+/// <summary>
+/// Attempts to rewrite a single occurrence of a mapped option.
+/// </summary>
+/// <param name="arguments">The argument list provided to that occurrence. Never null; length respects <see cref="CommandLineOptionMapping.Arity"/>.</param>
+/// <param name="result">When the method returns true, the canonical options this occurrence expands into; otherwise empty.</param>
+/// <returns>True if this mapping claims the occurrence; false otherwise.</returns>
+public delegate bool CommandLineOptionMapper(
+    ReadOnlySpan<string> arguments,
+    out IReadOnlyList<CommandLineOptionMappingResult> result);
+
+/// <summary>
+/// A single canonical option produced by a mapping.
+/// </summary>
+public sealed class CommandLineOptionMappingResult
+{
+    public CommandLineOptionMappingResult(string optionName, params string[] arguments);
+
+    /// <summary>The canonical MTP option name (without leading dashes).</summary>
+    public string OptionName { get; }
+
+    /// <summary>Arguments for the canonical option. May be empty for switch-style options.</summary>
+    public IReadOnlyList<string> Arguments { get; }
+}
+```
+
+Registration mirrors the existing `ICommandLineOptionsProvider` pattern:
+
+```csharp
+namespace Microsoft.Testing.Platform.Builder;
+
+public interface ICommandLineMappingsManager
+{
+    void AddProvider(Func<IServiceProvider, ICommandLineOptionMappingProvider> providerFactory);
+}
+
+public interface ITestApplicationBuilder
+{
+    // existing members ...
+    ICommandLineMappingsManager CommandLineMappings { get; }
+}
+```
+
+### Example: TRX
+
+```csharp
+testApplicationBuilder.CommandLineMappings.AddProvider(_ => new TrxLoggerMapping());
+
+internal sealed class TrxLoggerMapping : ICommandLineOptionMappingProvider
+{
+    public string Uid => nameof(TrxLoggerMapping);
+    public string Version => "1.0.0";
+    public string DisplayName => "TRX logger compatibility mapping";
+    public string Description => "Accepts the VSTest '--logger trx[;LogFileName=...]' syntax.";
+    public Task<bool> IsEnabledAsync() => Task.FromResult(true);
+
+    public IReadOnlyCollection<CommandLineOptionMapping> GetCommandLineOptionMappings() =>
+    [
+        new(
+            name: "logger",
+            description: "VSTest-compatible logger selector. Supported: trx[;LogFileName=<name>].",
+            arity: ArgumentArity.ExactlyOne,
+            map: TryMap),
+    ];
+
+    private static bool TryMap(ReadOnlySpan<string> arguments, out IReadOnlyList<CommandLineOptionMappingResult> result)
+    {
+        string value = arguments[0];
+
+        if (value == "trx")
+        {
+            result = [new CommandLineOptionMappingResult("report-trx")];
+            return true;
+        }
+
+        if (value.StartsWith("trx;", StringComparison.OrdinalIgnoreCase))
+        {
+            List<CommandLineOptionMappingResult> expanded = [new CommandLineOptionMappingResult("report-trx")];
+            foreach (string segment in value.AsSpan(4).ToString().Split(';'))
+            {
+                if (segment.StartsWith("LogFileName=", StringComparison.OrdinalIgnoreCase))
+                {
+                    expanded.Add(new CommandLineOptionMappingResult("report-trx-filename", segment.Substring("LogFileName=".Length)));
+                }
+                // unknown segments fall through; see "Unknown sub-options" below.
+            }
+
+            result = expanded;
+            return true;
+        }
+
+        result = [];
+        return false;
+    }
+}
+```
+
+### Example: Code coverage
+
+```csharp
+internal sealed class CodeCoverageCollectMapping : ICommandLineOptionMappingProvider
+{
+    // ... metadata omitted ...
+
+    public IReadOnlyCollection<CommandLineOptionMapping> GetCommandLineOptionMappings() =>
+    [
+        new(
+            name: "collect",
+            description: "VSTest-compatible data collector. Supported: 'XPlat Code Coverage', 'Code Coverage'.",
+            arity: ArgumentArity.ExactlyOne,
+            map: static (args, out result) =>
+            {
+                if (string.Equals(args[0], "XPlat Code Coverage", StringComparison.OrdinalIgnoreCase) ||
+                    string.Equals(args[0], "Code Coverage", StringComparison.OrdinalIgnoreCase))
+                {
+                    result = [new CommandLineOptionMappingResult("coverage")];
+                    return true;
+                }
+
+                result = [];
+                return false;
+            }),
+    ];
+}
+```
+
+### Resolution algorithm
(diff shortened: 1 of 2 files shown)


Evangelink (commented):
## RFC 014 Review Summary

I've reviewed RFC 014 - Command-line option mappings against 21 expert review dimensions. The RFC proposes a well-motivated solution to a real migration pain point (VSTest → MTP compatibility), with strong design principles and thorough alternatives analysis. However, **critical issues in the proposed API design require changes before approval**.

### Verdict Table

| # | Dimension | Verdict |
|---|-----------|---------|
| 1 | Public API & Binary Compatibility | 🔴 1 BLOCKING, 2 MODERATE |
| 2 | Defensive Coding at Boundaries | 🔴 2 BLOCKING, 2 MAJOR |
| 3 | Algorithmic Correctness | 🟡 2 MAJOR |
| 4 | Naming & Conventions | ⚪ 2 NIT |
| 5 | Documentation Accuracy | ⚪ 1 NIT |
| 6 | Performance & Allocations | 🟡 1 MODERATE (covered above) |

✅ 15/21 dimensions clean.

---

### Critical Issues (Must Fix Before Approval)

#### 🔴 BLOCKING #1: Delegate signature prevents future evolution and lacks safety
**Location**: Lines 117-119 (`CommandLineOptionMapper` delegate)

**Problems**:
1. **Cannot add warnings later**: RFC acknowledges (line 350) wanting to add warnings, but adding an `out` parameter to a delegate is a breaking change
2. **No exception handling**: Extension throws → platform crash during startup
3. **No result validation**: Extension returns `true` with `result = null` → NullReferenceException
4. **No timeout**: Extension infinite loop → platform hangs indefinitely

**Fix**: Replace `bool` return + `out` parameter with a result struct that can carry warnings in future versions:
```csharp
public delegate CommandLineOptionMapperResult CommandLineOptionMapper(
    ReadOnlySpan<string> arguments);

public readonly struct CommandLineOptionMapperResult
{
    public bool Success { get; init; }
    public IReadOnlyList<CommandLineOptionMappingResult> Results { get; init; }
    // Future: public IReadOnlyList<string> Warnings { get; init; }
}
```
Add defensive wrapping (try/catch, validation, timeout) to the RFC's algorithm section.

#### 🔴 BLOCKING #2: Naming inconsistency in manager pattern
**Location**: Lines 141, 149

**Problem**: `ICommandLineMappingsManager` uses plural "Mappings", breaking the universal MTP pattern: `ICommandLineManager`, `ITestHostManager`, `IConfigurationManager`, `ILoggingManager` (all singular).

**Fix**: Rename to `ICommandLineMappingManager` (singular). Update property name to `CommandLineMapping` to match the `TestHost` / `ITestHostManager` pattern.

---

### Major Issues (Should Fix)

#### 🟡 Algorithmic correctness gaps (2 issues)
1. **Empty result list undefined** (line 246): What if `TryMap` returns `true` with empty results? Does occurrence get deleted?
2. **Runtime validation missing** (line 258): Validation rules claim "enforced at startup" but can't detect invalid `OptionName` until runtime

**Fix**: Specify behavior for empty results (error recommended). Add runtime validation step in algorithm.

#### 🟡 Defensive coding gaps (2 more issues beyond BLOCKING)
1. **Unbounded growth** (line 192): Example code splits on `;` with no limits → 100K segments from malicious input
2. **Malformed sub-options** (line 194): No guidance on handling `trx;LogFileName=` (empty value)

**Fix**: Add defensive limits to examples (max segments, length checks). Document error handling approach for malformed sub-options.

---

### Moderate Issues (Recommended Fixes)

#### Constructor design (line 126)
1. **`params` blocks future evolution**: Can't add optional parameters after `params` array
2. **Unnecessary allocation**: `params` creates array wrapper even for zero-argument cases

**Fix**: Add non-params overload for zero-argument case and explicit `IReadOnlyList<string>` overload.

---

### Summary

**Strong points**:
- Clear problem statement and motivation
- Thorough alternatives analysis (5 rejected options)
- Well-defined design principles
- Comprehensive phasing plan
- Threading/concurrency design is sound (synchronous, pre-DI, single-threaded startup)

**Must address before approval**:
- Delegate signature evolution strategy (warnings, safety)
- Naming consistency with existing patterns
- Algorithm specification completeness (empty results, runtime validation)
- Defensive coding guidance (limits, error handling)

The core concept is sound and valuable. The issues are fixable refinements to the API design, not fundamental flaws. Recommend addressing BLOCKING issues and reconsidering the MAJOR issues before moving to implementation.

---

**Review dimensions not applicable to RFC**: Test Isolation, Assertion Quality, Flakiness Patterns, Data-Driven Test Coverage, Analyzer Quality, IPC Wire Compatibility, Resource Management, Localization, Build Infrastructure, Cross-TFM Compatibility (these apply to implementation, not design docs).

> Generated by [Expert Code Review (on open)](https://github.com/microsoft/testfx/actions/runs/26289052625) for issue #8501 · ● 22.6M

docs/RFCs/015-Command-Line-Option-Mappings.md
```diff
@@ -0,0 +1,354 @@
...
+
+/// <summary>
+/// Attempts to rewrite a single occurrence of a mapped option.
+/// </summary>
+/// <param name="arguments">The argument list provided to that occurrence. Never null; length respects <see cref="CommandLineOptionMapping.Arity"/>.</param>
+/// <param name="result">When the method returns true, the canonical options this occurrence expands into; otherwise empty.</param>
+/// <returns>True if this mapping claims the occurrence; false otherwise.</returns>
+public delegate bool CommandLineOptionMapper(
+    ReadOnlySpan<string> arguments,
+    out IReadOnlyList<CommandLineOptionMappingResult> result);
```
Evangelink, line 119: **[BLOCKING] Public API & Binary Compatibility + Defensive Coding**

**Multiple critical issues with the delegate signature:**

1. **Future enhancement blocked**: The RFC explicitly acknowledges (line 350) wanting to add a warning channel later, but adding an `out warnings` parameter would be a **breaking change** to this delegate. Delegates cannot be versioned or extended without breaking compatibility.

   **Recommendation**: Design for warnings now:
   ```csharp
   public delegate CommandLineOptionMapperResult CommandLineOptionMapper(
       ReadOnlySpan<string> arguments);
   
   public readonly struct CommandLineOptionMapperResult
   {
       public bool Success { get; init; }
       public IReadOnlyList<CommandLineOptionMappingResult> Results { get; init; }
       // Future: public IReadOnlyList<string> Warnings { get; init; }
   }
   ```
   This allows adding warnings in a future version without breaking existing callers.

2. **No exception handling**: Extension throws → platform crashes during startup. Any `OutOfMemoryException`, `NullReferenceException`, or custom exception will crash command-line parsing before tests run.

   **Recommendation**: Wrap each `TryMap` call in try/catch. Treat exceptions as "mapping declined", log error, continue to next mapping. Document that `TryMap` must not throw.

3. **No result validation**: Extension can return `true` but set `result = null` or include items with `null` `OptionName` → `NullReferenceException` during replacement.

   **Recommendation**: Validate when `TryMap` returns true: result not null, no null items, each item has non-empty `OptionName`.

4. **No timeout**: Extension infinite loop → platform hangs indefinitely.

   **Recommendation**: Add timeout (5 seconds per call) or pass `CancellationToken` through the delegate signature.

docs/RFCs/014-Command-Line-Option-Mappings.md
```diff
@@ -0,0 +1,354 @@
...
+    public IReadOnlyList<string> Arguments { get; }
+}
+```
+
+Registration mirrors the existing `ICommandLineOptionsProvider` pattern:
+
+```csharp
+namespace Microsoft.Testing.Platform.Builder;
+
+public interface ICommandLineMappingsManager
```
Evangelink: **[NIT] Naming & Conventions**

**Inconsistent with existing manager naming pattern**: All MTP managers use **singular** feature names: `ICommandLineManager`, `ITestHostManager`, `IConfigurationManager`, `ILoggingManager`. The plural "Mappings" breaks this convention.

**Recommendation**: Rename to `ICommandLineMappingManager` (singular "Mapping").

**Related issue**: Property name on line 149 should then be `CommandLineMapping` (matching the `TestHost` / `ITestHostManager` pattern) or the mapping functionality should be integrated into the existing `ICommandLineManager` for better cohesion.

docs/RFCs/015-Command-Line-Option-Mappings.md
```diff
@@ -0,0 +1,354 @@
...
+        if (value.StartsWith("trx;", StringComparison.OrdinalIgnoreCase))
+        {
+            List<CommandLineOptionMappingResult> expanded = [new CommandLineOptionMappingResult("report-trx")];
+            foreach (string segment in value.AsSpan(4).ToString().Split(';'))
+            {
+                if (segment.StartsWith("LogFileName=", StringComparison.OrdinalIgnoreCase))
+                {
+                    expanded.Add(new CommandLineOptionMappingResult("report-trx-filename", segment.Substring("LogFileName=".Length)));
+                }
+                // unknown segments fall through; see "Unknown sub-options" below.
```
Evangelink, line 214: **[NIT] Documentation Accuracy**

**Missing error handling guidance**: The comment "unknown segments fall through" addresses one case but doesn't cover **malformed known options**. What should happen for:
- `trx;LogFileName=` (empty value after `=`)
- `trx;LogFileName` (missing `=`)
- `trx;LogFileName=../../etc/passwd` (path traversal attempt)

Implementers are unclear whether mappings should validate sub-option structure or rely on downstream validation.

**Recommendation**: Add guidance in this section or in "Sub-option handling" (lines 280-285):
> "Mappings should validate the structure of known sub-options and return `false` if they are malformed (e.g., empty value after `=`), letting the platform report 'No registered mapping can handle value X'. Path validation is the responsibility of the downstream canonical option provider."

docs/RFCs/015-Command-Line-Option-Mappings.md
```diff
@@ -0,0 +1,354 @@
...
+The user's original `string[] args` is preserved alongside the expanded set for diagnostics (`--info`, error messages, telemetry).
+
+### Cross-cutting validation rules
+
+Enforced at platform startup, before any user input is processed:
+
+- A mapping name MUST NOT collide with any registered canonical `CommandLineOption.Name`. Diagnostic: `Mapping '--X' conflicts with the canonical option '--X' provided by <provider>. Rename one.`
+- A `CommandLineOptionMappingResult.OptionName` MUST refer to a registered canonical `CommandLineOption`. Diagnostic: `Mapping '--X' rewrites to unknown option '--Y'.`
+- A `CommandLineOptionMappingResult.OptionName` MUST NOT itself be a mapping name (no mapping-to-mapping rewriting; prevents loops and makes the resolution algorithm finite by construction).
+- Mapping names follow the same character set as canonical option names (letters, digits, `-`, `?`).
```
Evangelink, line 278: **[MAJOR] Algorithmic Correctness**

**Runtime validation not specified**: Cross-cutting validation rules (lines 256-257) claim to be "enforced at platform startup" but **cannot detect runtime violations**.

**Concrete scenario**: A buggy mapping returns `OptionName = "non-existent-option"` at runtime. The `TryMap` delegate is opaque at registration — the platform cannot know what `OptionName` values it will return until a user provides arguments.

**Current behavior**: The invalid option name falls through to aggregation/validation (line 247), producing a confusing "unknown option" error rather than a clear "mapping produced invalid result" error.

**Recommendation**:
1. Clarify in line 253 that rules on lines 256-257 are validated at **resolution time** (step 2 of algorithm), not startup.
2. Add explicit check in algorithm step 2 (between lines 246-247):
   > "For each `OptionName` in the claimant's result, verify it refers to a registered canonical option and is not itself a mapping name. If validation fails, emit error: 'Mapping '--X' attempted to rewrite to invalid/unmapped option '--Y'.'"

docs/RFCs/014-Command-Line-Option-Mappings.md
```diff
@@ -0,0 +1,354 @@
...
+public delegate bool CommandLineOptionMapper(
+    ReadOnlySpan<string> arguments,
+    out IReadOnlyList<CommandLineOptionMappingResult> result);
+
+/// <summary>
+/// A single canonical option produced by a mapping.
+/// </summary>
+public sealed class CommandLineOptionMappingResult
+{
+    public CommandLineOptionMappingResult(string optionName, params string[] arguments);
```
Evangelink: **[MODERATE] Public API & Binary Compatibility + Performance**

**Two issues with the constructor:**

1. **`params` array limits future evolution**: C# prohibits parameters after `params`. If future versions need optional parameters (e.g., metadata, flags), the only option is a new constructor, forcing callers to pass arrays explicitly instead of using `params` convenience.

2. **Unnecessary allocation**: `params string[]` allocates an array wrapper on every call, even for zero-argument cases like `new CommandLineOptionMappingResult("report-trx")` (line 201). For 20 `--logger` occurrences, this creates 20+ avoidable allocations during startup.

**Recommendation**: Add explicit overloads:
```csharp
public CommandLineOptionMappingResult(string optionName) 
    : this(optionName, Array.Empty<string>()) { }

public CommandLineOptionMappingResult(string optionName, IReadOnlyList<string> arguments)
{
    OptionName = optionName;
    Arguments = arguments;
}

public CommandLineOptionMappingResult(string optionName, params string[] arguments)
    : this(optionName, (IReadOnlyList<string>)arguments) { }
```
This eliminates allocations for the common switch case and reserves non-params signature for future enhancement.

docs/RFCs/014-Command-Line-Option-Mappings.md
```diff
@@ -0,0 +1,354 @@
...
+
+`CommandLineParser` is augmented with a post-parse pass that runs **after** tokenization and response-file expansion, **before** option validation:
+
+1. Build a lookup table `MappingsByName : string → List<CommandLineOptionMapping>` from every registered provider.
+2. For each parsed option occurrence `--X v1 v2`:
+   - If `X` is not in `MappingsByName`, keep the occurrence unchanged.
+   - Otherwise, for every mapping registered under `X`, call `TryMap` exactly once.
+     - **Zero claimants** → emit an error: `No registered mapping for '--X' can handle value 'v1'. Run '--help' to see canonical alternatives.`
+     - **Multiple claimants** → emit an error: `Multiple mappings claim '--X v1': <provider Uids>. This is a configuration bug — at most one mapping may handle a given value.`
+     - **Exactly one claimant** → replace the occurrence with the canonical occurrences it returned.
```
Evangelink: **[MAJOR] Algorithmic Correctness**

**Undefined behavior for empty result list**: The algorithm states "replace the occurrence with the canonical occurrences it returned" but doesn't specify what happens when the result list is **empty**.

**Concrete scenario**:
```csharp
TryMap(["trx"]) → (true, [])  // Claims but produces nothing
```

The API signature allows `IReadOnlyList<CommandLineOptionMappingResult>` to be empty, but semantics are undefined:
- Does the occurrence get removed (deleting user input)?
- Is this a no-op that should error?
- Does it violate "mappings never replace canonical options" (line 93)?

Examples never return empty lists when returning `true`, suggesting this may be unintentional.

**Recommendation**: Add explicit case after line 246:
> "If the claimant returned an empty result list, emit error: 'Mapping '--X' claimed value 'v1' but produced no canonical options. This is a configuration bug.'"

OR explicitly document that empty results are valid and specify intended behavior (removal, passthrough, etc.).

docs/RFCs/014-Command-Line-Option-Mappings.md
```diff
@@ -0,0 +1,354 @@
...
+        }
+
+        if (value.StartsWith("trx;", StringComparison.OrdinalIgnoreCase))
+        {
+            List<CommandLineOptionMappingResult> expanded = [new CommandLineOptionMappingResult("report-trx")];
+            foreach (string segment in value.AsSpan(4).ToString().Split(';'))
+            {
+                if (segment.StartsWith("LogFileName=", StringComparison.OrdinalIgnoreCase))
+                {
+                    expanded.Add(new CommandLineOptionMappingResult("report-trx-filename", segment.Substring("LogFileName=".Length)));
```
Evangelink: **[MAJOR] Defensive Coding at Boundaries**

**Unbounded growth from malicious input**: The example does `value.AsSpan(4).ToString().Split(';')` with no limits, then loops through all segments. A malicious input like:
```bash
--logger "trx;a=1;b=2;...;z=999999"  # 100,000 segments
```
will allocate 100,000 strings and loop 100,000 times.

`segment.Substring("LogFileName=".Length)` can extract arbitrary-length strings (potential 100MB string from crafted input).

**Recommendation**: Add defensive limits to RFC guidance:
1. Limit split results: `Split(';', maxCount: 50)`
2. Validate segment length before substring: `if (segment.Length > 1000) continue;`
3. Limit total expanded results: `if (expanded.Count > 20) break;`
4. Document recommended limits in the "Sub-option handling" section (lines 280-284).

docs/RFCs/015-Command-Line-Option-Mappings.md
Evangelink: Added recommended handling for malformed known options (empty value treated as no-op, with the option of surfacing a post-DI warning) in the TRX example and the surrounding prose. The platform stays unopinionated; mappings are free to be stricter.

docs/RFCs/014-Command-Line-Option-Mappings.md
```diff
@@ -0,0 +1,354 @@
...
+
+### Why not just add `--logger` to a single extension
+
+A naive workaround — making, say, the TRX extension own `--logger` and dispatch by value — collapses the moment a second VSTest-compatible extension wants to participate. It also leaks an unrelated extension's identity (TRX) into the dispatcher for unrelated values (`console`, `html`). The mapping mechanism described below is the smallest surface that lets multiple extensions cooperate without breaking the single-owner invariant for canonical MTP options.
+
+## Naming
+
+Earlier discussion (issue [#7249](https://github.com/microsoft/testfx/issues/7249)) proposed the term **alias**. Feedback noted that "alias" in CLI conventions usually means a pure rename (`-v` ≡ `--verbose`) with identical arity and semantics, whereas this feature actually rewrites one option into one *or more* options with potentially different arity. Shell aliases (`alias ll='ls -la'`) and Git aliases are precedents for the broader meaning, but the ambiguity is real for API consumers.
+
+This RFC uses **mapping** in the public API surface (`ICommandLineOptionMapping`, `CommandLineOptionMapping`, `CommandLineMappings`). Rationale:
```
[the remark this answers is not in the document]
Evangelink: Naming section now lists ICommandLineOptionMappingProvider / CommandLineOptionMapping / ICommandLineMappingManager (singular Mapping), matching the proposed API surface and the existing manager-naming pattern.

docs/RFCs/014-Command-Line-Option-Mappings.md
```diff
@@ -0,0 +1,354 @@
...
+        string value = arguments[0];
+
+        if (value == "trx")
+        {
+            result = [new CommandLineOptionMappingResult("report-trx")];
+            return true;
+        }
+
+        if (value.StartsWith("trx;", StringComparison.OrdinalIgnoreCase))
+        {
```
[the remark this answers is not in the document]
Evangelink: Made the exact-match check case-insensitive (string.Equals OrdinalIgnoreCase), matching the prefix check and VSTest-compat expectations.

docs/RFCs/014-Command-Line-Option-Mappings.md
```diff
@@ -0,0 +1,354 @@
...
+/// </summary>
+public interface ICommandLineOptionMappingProvider : IExtension
+{
+    IReadOnlyCollection<CommandLineOptionMapping> GetCommandLineOptionMappings();
+}
+
+/// <summary>
+/// A single mapping registration. Multiple providers may register mappings with the same
+/// <see cref="Name"/>; exactly one of them is expected to return true from
+/// <see cref="TryMap"/> for any given argument list.
```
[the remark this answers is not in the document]
Evangelink: Updated the doc comment to <see cref=""Map""/> pointing at the property, since the delegate type is CommandLineOptionMapper rather than a method named TryMap.

docs/RFCs/015-Command-Line-Option-Mappings.md
Evangelink: Acknowledged. Added an Unresolved Questions item making it explicit that adding an out-parameter to the existing delegate would be a binary-breaking change, so the warning-channel decision has to be made before shipping (or absorbed via a sibling delegate type). The delegate now uses IReadOnlyList<string> for arguments to remove the netstandard2.0 Span concern.

docs/RFCs/014-Command-Line-Option-Mappings.md
```diff
@@ -0,0 +1,354 @@
...
+}
+
+/// <summary>
+/// Attempts to rewrite a single occurrence of a mapped option.
+/// </summary>
+/// <param name="arguments">The argument list provided to that occurrence. Never null; length respects <see cref="CommandLineOptionMapping.Arity"/>.</param>
+/// <param name="result">When the method returns true, the canonical options this occurrence expands into; otherwise empty.</param>
+/// <returns>True if this mapping claims the occurrence; false otherwise.</returns>
+public delegate bool CommandLineOptionMapper(
+    ReadOnlySpan<string> arguments,
```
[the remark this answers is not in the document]
Evangelink: Switched the public delegate to IReadOnlyList<string> so we do not expose Span-shaped types in netstandard2.0 public surface (and avoid the System.Memory dependency for callers).

docs/RFCs/014-Command-Line-Option-Mappings.md
Evangelink: Added a defensive-coding example to the TRX mapping (cap on sub-option segments, explicit IsNullOrEmpty guard for the file name) and called it out in the surrounding prose so authors know they own that responsibility.

docs/RFCs/014-Command-Line-Option-Mappings.md
Evangelink: Resolution algorithm now spells out the empty-result case explicitly: returning true with an empty (or null) result is a mapping contract violation that produces a clear error naming the offending provider. Doc on the delegate matches.

docs/RFCs/014-Command-Line-Option-Mappings.md
Evangelink: Replaced ""params string[] arguments"" with ""IReadOnlyList<string> arguments"" in the CommandLineOptionMappingResult constructor — it removes both the params-future-evolution constraint and the unconditional array allocation for the common zero/one-argument case.

docs/RFCs/015-Command-Line-Option-Mappings.md
Evangelink: Split the cross-cutting validation list into rules that are enforceable at platform startup (option-name collisions, charset) and rules that are inherently runtime because the Map delegate is opaque at registration time (unknown OptionName, no mapping-to-mapping recursion). Each runtime rule documents the diagnostic and exit code.

docs/RFCs/014-Command-Line-Option-Mappings.md
```diff
@@ -0,0 +1,354 @@
+# RFC 014 - Command-line option mappings
```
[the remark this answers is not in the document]
Evangelink: Renamed the file to 015-Command-Line-Option-Mappings.md and updated the title to ""RFC 015 - Command-line option mappings"" to avoid the collision with 014-TestRun-Current-PlannedTests.

docs/RFCs/014-Command-Line-Option-Mappings.md
Evangelink: Renamed the registration interface to ICommandLineMappingManager (singular) and updated the Naming section reference to match the existing singular ICommandLineManager / ITestHostManager / IConfigurationManager / ILoggingManager pattern.

PR #8491, opened 2026-05-22
Reset agentics workflows to upstream and add new ones

Evangelink:
Reset the agentic workflows that originated from [githubnext/agentics](https://github.com/githubnext/agentics) back to the upstream sources and add `source:` tracking so future syncs can be driven by `gh aw update`. Also installs a handful of useful new workflows from upstream.

## Reset to upstream + add `source:`

The following workflows were manually copied and tweaked over time; they are now overwritten with the verbatim upstream content and gain a `source:` frontmatter field (or markdown comment for shared fragments). Filenames are renamed to match upstream where they differed.

| Old file | New file (matches upstream) |
| --- | --- |
| `daily-efficiency-improver.md` | `efficiency-improver.md` |
| `daily-malicious-code-scan.md` | `malicious-code-scan.md` |
| `daily-perf-improver.md` | `perf-improver.md` |
| `daily-qa.md` | `adhoc-qa.md` |
| `daily-test-improver.md` | `test-improver.md` |
| `markdown-linter-report.md` | `markdown-linter.md` |
| `repo-historian.md` | `repo-chronicle.md` |
| `code-simplifier.md` | (same) |
| `glossary-maintainer.md` | (same) |
| `issue-arborist.md` | (same) |
| `repository-quality-improver.md` | (same) |
| `shared/reporting.md` | (same) |

Plus the new shared fragment `shared/formatting.md` (newly required by `code-simplifier`).

## New workflows added from upstream

- `dependabot-pr-bundler`
- `dependabot-issue-bundler`
- `duplicate-code-detector`
- `link-checker`
- `pr-fix`
- `sub-issue-closer` (pairs with `issue-arborist`)
- `weekly-issue-activity`

## Local customizations reverted (intentional)

Strict reset, so these are gone:

- `issue-arborist.md`: the DIFC-proxy `curl` workaround (reverts to `gh issue list`); may need to be reapplied if DIFC proxy blocks `/meta` again.
- `daily-qa.md`: lost the `imports: shared/repo-build-setup.md` and the `.dotnet` network allowlist.
- `name:` overrides and local title-prefix tweaks across several workflows.

## Compilation

All 28 workflows compile cleanly in strict mode:

```
gh aw compile --strict --approve
# -> Compiled 28 workflow(s): 0 error(s), 2 warning(s)
```

The two warnings are pre-existing (centralized slash-command suggestion and an automatic env-var extraction in `issue-arborist` for shell-injection safety).

PR #8507, opened 2026-05-22
fix(aw): Issue Arborist - re-apply curl REST API workaround for DIFC proxy /meta block

Evangelink:
The agentic-workflows reset in #8491 reverted the Issue Arborist `Fetch issues data` step back to `gh issue list`, which immediately fails in CI because the gh CLI hits `/meta` against the corporate DIFC proxy and gets HTTP 403. That breaks the run before the MCP server can start, which is what produces the `ERR_SYSTEM: rpc-messages.jsonl is present but zero bytes — MCP telemetry capture failed` annotation seen in [run 26294701799](https://github.com/microsoft/testfx/actions/runs/26294701799).

Re-applies the same fix as #8185: call the GitHub Search REST API directly with `curl`, using `GH_AW_ORIGINAL_GITHUB_API_URL` to bypass the DIFC proxy. The lock file is regenerated via `gh aw compile issue-arborist --strict`.


Evangelink (commented):

This PR correctly re-applies the DIFC proxy workaround from #8185 that was inadvertently lost during the upstream workflow reset in #8491.

## Technical Review Summary

**Changes:**
1. Replaced `gh issue list` with direct `curl` call to GitHub Search API
2. Added `jq` transformation to normalize REST API response format
3. Implemented proper error handling with `|| echo '[]'` fallback
4. Regenerated lock file (removed DIFC proxy infrastructure since it's no longer requested in source `.md`)

**Verification:**
- ✓ `curl` command properly uses `--data-urlencode` to prevent query injection
- ✓ Bearer token correctly passed in Authorization header
- ✓ All shell variables properly quoted
- ✓ `jq` transformation correctly maps REST API fields to expected format (snake_case → camelCase, lowercase → uppercase state)
- ✓ Fallback to empty array `[]` on any failure prevents workflow abortion
- ✓ Lock file changes are mechanical compilation artifacts (updated heredoc delimiters, removed unused DIFC proxy steps)

**Correctness:**
The workaround is sound: `gh issue list` internally calls `/meta` for server-type detection, which the corporate DIFC proxy blocks. By using `curl` directly against the documented GitHub REST API endpoint, the workflow bypasses the gh CLI entirely and avoids the `/meta` call.

The lock file correctly no longer includes DIFC proxy start/stop steps because the source `.md` file (after the upstream reset) doesn't explicitly request DIFC proxy infrastructure — the compiler generates only what's needed.

**Scope:**
Appropriately narrow — re-applies a proven fix without scope creep.

> Generated by [Expert Code Review (on open)](https://github.com/microsoft/testfx/actions/runs/26296680739) for issue #8507 · ● 7.4M

PR #8526, opened 2026-05-23
Add --crash-sequence option to MTP CrashDump extension (#7262)

### LINKED ISSUE #7262: [MTP Dump] Support sequence files similar to VSTest
## Summary

In VSTest, a test sequence file is written in case of crash. It's not supported in VSTest for hang though. I'm not sure if there is a good reason for that.

## Proposed Feature

Support generating a test sequence file by MTP crash dump extension. Consider doing it also for hang dump if reasonable.

Evangelink:
Closes #7262.

## What

Adds a `--crash-sequence on|off` option (default `on`) to `Microsoft.Testing.Extensions.CrashDump` that generates a sequence file recording the start and end of each test as the run progresses. On a crash, the controller parses the file, prints a friendly *"tests still running at the time of the crash"* banner, and publishes the file as a `FileArtifact`.

This addresses the long-standing pain point from the issue: today, on `--crashdump`/`--crash-report`, users have to open the `.dmp` to know which test was running at crash time. `--hangdump` already produces an equivalent `.log` over IPC, but that path is unavailable for crashes because the testhost is dead.

## How

- `CrashDumpSequenceLogger` (new testhost `IDataConsumer + ITestSessionLifetimeHandler`) appends `STARTED\t<iso>\t<uid>\t<displayName>` / `ENDED\t<iso>\t<uid>\t<state>` lines per `TestNodeUpdateMessage` to a sequence file. Uses `StreamWriter.AutoFlush=true` (no `fsync`) so records reach the OS cache and survive `Environment.FailFast` at sub-microsecond cost per write.
- `CrashDumpEnvironmentVariableProvider` computes a path with a per-controller 8-char GUID token (avoids cross-host collisions for parallel testhost launches into the same results directory) and exposes it via `TESTINGPLATFORM_CRASHDUMP_SEQUENCE_FILE`.
- `CrashDumpProcessLifetimeHandler` on crash: parses the file, computes `started \ ended` set, prints the friendly banner, publishes the file. On graceful exit / hangdump kill: deletes the file (no diagnostic value).
- `--crash-sequence` follows the `--ansi` style (on/off/true/false/enable/disable/1/0), validated by the existing `CommandLineOptionArgumentValidator`.

## Performance

The cost is two small writes per test (~10 µs each, no fsync). For 10 000 tests this is ~200 ms total — well under 1 % on huge suites. The user discussed and confirmed this in the design conversation; the default-on policy is justified by the very low cost.

## Limitations (documented in source)

- Best-effort: a test that starts and immediately `FailFast`s within a few ms may not be recorded because the MTP message bus is asynchronous by design. This matches the same trade-off described in the issue ("mtp does not send the currently started test to testhost controller because of possible overhead"). The added value is for the much more common "test runs for some time, then crashes" pattern.
- Sequence file does not survive an OS-level crash or power loss (no `fsync`).

## Tests

- 9 new unit-test datarow combinations cover argument validation and the rejection of `--crash-sequence` without `--crashdump`/`--crash-report`.
- 5 new acceptance tests: sequence-on-crash content, off-by-flag, graceful-exit cleanup, invalid argument, missing main option. The CrashDump test asset was extended to publish fake in-progress test nodes before `FailFast` (with a small sync delay to give the async consumer time to drain, since this is a deterministic test setup).
- Help/info expectations updated and XLF files regenerated via `UpdateXlf`.

## Out of scope

This PR does not change `--hangdump` behavior: it already generates an equivalent in-progress test list over IPC.

### DIFF
--- a/src/Platform/Microsoft.Testing.Extensions.CrashDump/CrashDumpCommandLineProvider.cs
+++ b/src/Platform/Microsoft.Testing.Extensions.CrashDump/CrashDumpCommandLineProvider.cs
@@ -34,22 +34,20 @@ internal sealed class CrashDumpCommandLineProvider : ICommandLineOptionsProvider
 
     public Task<ValidationResult> ValidateOptionArgumentsAsync(CommandLineOption commandOption, string[] arguments)
     {
-        if (commandOption.Name == CrashDumpCommandLineOptions.CrashDumpTypeOptionName)
+        if (commandOption.Name == CrashDumpCommandLineOptions.CrashDumpTypeOptionName
+            && !DumpTypeOptions.Contains(arguments[0], StringComparer.OrdinalIgnoreCase))
         {
-            if (!DumpTypeOptions.Contains(arguments[0], StringComparer.OrdinalIgnoreCase))
-            {
-                return ValidationResult.InvalidTask(string.Format(CultureInfo.InvariantCulture, CrashDumpResources.CrashDumpTypeOptionInvalidType, arguments[0]));
-            }
+            return ValidationResult.InvalidTask(string.Format(CultureInfo.InvariantCulture, CrashDumpResources.CrashDumpTypeOptionInvalidType, arguments[0]));
         }
-        else if (commandOption.Name == CrashDumpCommandLineOptions.CrashSequenceOptionName)
+        else if (commandOption.Name == CrashDumpCommandLineOptions.CrashSequenceOptionName
+            && !CommandLineOptionArgumentValidator.IsValidBooleanArgument(arguments[0]))
         {
-            if (!CommandLineOptionArgumentValidator.IsValidBooleanArgument(arguments[0]))
-            {
-                return ValidationResult.InvalidTask(CrashDumpResources.CrashSequenceOptionInvalidArgument);
-            }
+            return ValidationResult.InvalidTask(CrashDumpResources.CrashSequenceOptionInvalidArgument);
         }
 
-        // TODO: Validate that the file name ends with '.dmp'?
+        // We intentionally do not enforce a '.dmp' extension on --crashdump-filename: dotnet-dump
+        // and the Windows MiniDumpWriteDump APIs both accept arbitrary file names, and users
+        // sometimes script around custom suffixes (e.g. timestamps appended by an outer wrapper).
         return ValidationResult.ValidTask;
     }
 
--- a/src/Platform/Microsoft.Testing.Extensions.CrashDump/CrashDumpCommandLineProvider.cs
+++ b/src/Platform/Microsoft.Testing.Extensions.CrashDump/CrashDumpCommandLineProvider.cs
@@ -15,6 +15,7 @@ internal sealed class CrashDumpCommandLineProvider : ICommandLineOptionsProvider
     [
         new(CrashDumpCommandLineOptions.CrashDumpOptionName, CrashDumpResources.CrashDumpOptionDescription, ArgumentArity.Zero, false),
         new(CrashDumpCommandLineOptions.CrashReportOptionName, CrashDumpResources.CrashReportOptionDescription, ArgumentArity.Zero, false),
+        new(CrashDumpCommandLineOptions.CrashSequenceOptionName, CrashDumpResources.CrashSequenceOptionDescription, ArgumentArity.ExactlyOne, false),
         new(CrashDumpCommandLineOptions.CrashDumpFileNameOptionName, CrashDumpResources.CrashDumpFileNameOptionDescription, ArgumentArity.ExactlyOne, false),
         new(CrashDumpCommandLineOptions.CrashDumpTypeOptionName, CrashDumpResources.CrashDumpTypeOptionDescription, ArgumentArity.ExactlyOne, false)
     ];
@@ -40,6 +41,13 @@ internal sealed class CrashDumpCommandLineProvider : ICommandLineOptionsProvider
                 return ValidationResult.InvalidTask(string.Format(CultureInfo.InvariantCulture, CrashDumpResources.CrashDumpTypeOptionInvalidType, arguments[0]));
             }
         }
+        else if (commandOption.Name == CrashDumpCommandLineOptions.CrashSequenceOptionName)
+        {
+            if (!CommandLineOptionArgumentValidator.IsValidBooleanArgument(arguments[0]))
+            {
+                return ValidationResult.InvalidTask(CrashDumpResources.CrashSequenceOptionInvalidArgument);
+            }
+        }
 
         // TODO: Validate that the file name ends with '.dmp'?
         return ValidationResult.ValidTask;
@@ -59,7 +67,8 @@ internal sealed class CrashDumpCommandLineProvider : ICommandLineOptionsProvider
     private static bool IsCrashDumpMainOptionMissing(ICommandLineOptions commandLineOptions)
     {
         bool hasCrashDumpSubOption = commandLineOptions.IsOptionSet(CrashDumpCommandLineOptions.CrashDumpFileNameOptionName) ||
-            commandLineOptions.IsOptionSet(CrashDumpCommandLineOptions.CrashDumpTypeOptionName);
+            commandLineOptions.IsOptionSet(CrashDumpCommandLineOptions.CrashDumpTypeOptionName) ||
+            commandLineOptions.IsOptionSet(CrashDumpCommandLineOptions.CrashSequenceOptionName);
         bool hasCrashDumpMainOption = commandLineOptions.IsOptionSet(CrashDumpCommandLineOptions.CrashDumpOptionName) ||
             commandLineOptions.IsOptionSet(CrashDumpCommandLineOptions.CrashReportOptionName);
 
--- a/src/Platform/Microsoft.Testing.Extensions.CrashDump/CrashDumpProcessLifetimeHandler.cs
+++ b/src/Platform/Microsoft.Testing.Extensions.CrashDump/CrashDumpProcessLifetimeHandler.cs
@@ -330,10 +330,12 @@ internal sealed class CrashDumpProcessLifetimeHandler : ITestHostProcessLifetime
                 }
             }
         }
-        catch (IOException ex)
+        catch (Exception ex) when (ex is IOException or UnauthorizedAccessException or System.Security.SecurityException or ArgumentException or NotSupportedException)
         {
-            // Best-effort diagnostic. If we cannot read the sequence file we still publish it so the
-            // user can inspect it manually, but skip the friendly summary.
+            // Best-effort diagnostic. If we cannot read the sequence file for any expected reason
+            // (ACLs, sharing violations, malformed path, etc.) we still publish it so the user
+            // can inspect it manually, but skip the friendly summary. Failing the crash-handling
+            // path because of a diagnostic file would be strictly worse than missing the summary.
             await _outputDisplay.DisplayAsync(this, new ErrorMessageOutputDeviceData(string.Format(CultureInfo.InvariantCulture, CrashDumpResources.CrashDumpSequenceFileReadError, sequenceFilePath, ex.Message)), cancellationToken).ConfigureAwait(false);
             await _messageBus.PublishAsync(this, new FileArtifact(new FileInfo(sequenceFilePath), CrashDumpResources.CrashDumpSequenceArtifactDisplayName, CrashDumpResources.CrashDumpSequenceArtifactDescription)).ConfigureAwait(false);
             return;
--- a/src/Platform/Microsoft.Testing.Extensions.CrashDump/CrashDumpProcessLifetimeHandler.cs
+++ b/src/Platform/Microsoft.Testing.Extensions.CrashDump/CrashDumpProcessLifetimeHandler.cs
@@ -2,6 +2,7 @@
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
 using Microsoft.Testing.Extensions.Diagnostics.Resources;
+using Microsoft.Testing.Platform;
 using Microsoft.Testing.Platform.CommandLine;
 using Microsoft.Testing.Platform.Extensions.Messages;
 using Microsoft.Testing.Platform.Extensions.OutputDevice;
@@ -120,6 +121,11 @@ internal sealed class CrashDumpProcessLifetimeHandler : ITestHostProcessLifetime
         if (testHostProcessInformation.HasExitedGracefully
             || (AppDomain.CurrentDomain.GetData("ProcessKilledByHangDump") is string processKilledByHangDump && processKilledByHangDump == "true"))
         {
+            // No crash → the sequence file (if any) has no diagnostic value. Delete it so the user's
+            // results directory is not polluted with stale "tests still running" logs after a clean
+            // run. Hang-dump kills are handled identically because HangDump already produces its own
+            // .log of in-progress tests via IPC.
+            TryDeleteSequenceFile();
             return;
         }
 
@@ -264,6 +270,117 @@ internal sealed class CrashDumpProcessLifetimeHandler : ITestHostProcessLifetime
                     await _messageBus.PublishAsync(this, new FileArtifact(new FileInfo(crashReportFile), CrashDumpResources.CrashReportArtifactDisplayName, CrashDumpResources.CrashReportArtifactDescription)).ConfigureAwait(false);
                 }
             }
+        }
+
+        await TryPublishSequenceFileAsync(cancellationToken).ConfigureAwait(false);
+    }
+
+    private async Task TryPublishSequenceFileAsync(CancellationToken cancellationToken)
+    {
+        string? sequenceFilePath = _netCoreCrashDumpGeneratorConfiguration.SequenceFileName;
+        if (RoslynString.IsNullOrEmpty(sequenceFilePath) || !File.Exists(sequenceFilePath))
+        {
+            return;
+        }
+
+        // Parse the journal to compute the set of tests that started but never ended. We can tolerate
+        // a partially-written final line because the testhost flushes after each whole record; any
+        // half-written tail line would simply fail to parse and be ignored.
+        var inFlight = new Dictionary<string, (string DisplayName, DateTimeOffset StartedAt)>(StringComparer.Ordinal);
+        DateTimeOffset latestSeen = DateTimeOffset.MinValue;
+        try
+        {
+            foreach (string line in File.ReadLines(sequenceFilePath))
+            {
+                if (line.Length == 0 || line[0] == '#')
+                {
+                    continue;
+                }
+
+                // Format: <event>\t<isoTimestamp>\t<uid>\t<displayName-or-state>
+                string[] parts = line.Split('\t');
+                if (parts.Length < 4)
+                {
+                    continue;
+                }
+
+                if (!DateTimeOffset.TryParse(parts[1], CultureInfo.InvariantCulture, DateTimeStyles.RoundtripKind, out DateTimeOffset timestamp))
+                {
+                    continue;
+                }
+
+                if (timestamp > latestSeen)
+                {
+                    latestSeen = timestamp;
+                }
+
+                string uid = parts[2];
+                // Any field originally containing tabs was sanitized to spaces by the testhost, so
+                // a 4-element split is sufficient. Defensively re-join any extras to tolerate future
+                // schema changes.
+                string lastField = parts.Length == 4 ? parts[3] : string.Join("\t", parts, 3, parts.Length - 3);
+
+                if (parts[0].Equals(CrashDumpSequenceLogger.StartedEvent, StringComparison.Ordinal))
+                {
+                    inFlight[uid] = (lastField, timestamp);
+                }
+                else if (parts[0].Equals(CrashDumpSequenceLogger.EndedEvent, StringComparison.Ordinal))
+                {
+                    inFlight.Remove(uid);
+                }
+            }
+        }
+        catch (IOException ex)
+        {
+            // Best-effort diagnostic. If we cannot read the sequence file we still publish it so the
+            // user can inspect it manually, but skip the friendly summary.
+            await _outputDisplay.DisplayAsync(this, new ErrorMessageOutputDeviceData(string.Format(CultureInfo.InvariantCulture, CrashDumpResources.CrashDumpSequenceFileReadError, sequenceFilePath, ex.Message)), cancellationToken).ConfigureAwait(false);
+            await _messageBus.PublishAsync(this, new FileArtifact(new FileInfo(sequenceFilePath), CrashDumpResources.CrashDumpSequenceArtifactDisplayName, CrashDumpResources.CrashDumpSequenceArtifactDescription)).ConfigureAwait(false);
+            return;
+        }
+
+        if (inFlight.Count > 0)
+        {
+            await _outputDisplay.DisplayAsync(this, new ErrorMessageOutputDeviceData(CrashDumpResources.CrashDumpTestsRunningAtCrash), cancellationToken).ConfigureAwait(false);
+            DateTimeOffset anchor = latestSeen == DateTimeOffset.MinValue ? DateTimeOffset.UtcNow : latestSeen;
+            foreach (KeyValuePair<string, (string DisplayName, DateTimeOffset StartedAt)> entry in inFlight.OrderBy(static x => x.Value.StartedAt))
+            {
+                TimeSpan elapsed = anchor - entry.Value.StartedAt;
+                if (elapsed < TimeSpan.Zero)
+                {
+                    elapsed = TimeSpan.Zero;
+                }
+
+                await _outputDisplay.DisplayAsync(this, new ErrorMessageOutputDeviceData($"[{elapsed}] {entry.Value.DisplayName}"), cancellationToken).ConfigureAwait(false);
+            }
+        }
+
+        await _messageBus.PublishAsync(this, new FileArtifact(new FileInfo(sequenceFilePath), CrashDumpResources.CrashDumpSequenceArtifactDisplayName, CrashDumpResources.CrashDumpSequenceArtifactDescription)).ConfigureAwait(false);
+    }
+
+    private void TryDeleteSequenceFile()
+    {
+        string? sequenceFilePath = _netCoreCrashDumpGeneratorConfiguration.SequenceFileName;
+        if (RoslynString.IsNullOrEmpty(sequenceFilePath))
+        {
+            return;
+        }
+
+        try
+        {
+            if (File.Exists(sequenceFilePath))
+            {
+                File.Delete(sequenceFilePath);
+            }
+        }
+        catch (IOException)
+        {
+            // Best-effort cleanup; a leftover sequence file is harmless beyond cluttering the results
+            // directory. Avoid surfacing this as a user-visible error because the run itself succeeded.
+        }
+        catch (UnauthorizedAccessException)
+        {
+            // Same rationale as IOException.
         }
     }
 
--- a/src/Platform/Microsoft.Testing.Extensions.CrashDump/CrashDumpSequenceLogger.cs
+++ b/src/Platform/Microsoft.Testing.Extensions.CrashDump/CrashDumpSequenceLogger.cs
@@ -39,7 +39,10 @@ internal sealed class CrashDumpSequenceLogger : IDataConsumer, ITestSessionLifet
     private readonly IEnvironment _environment;
     private readonly IClock _clock;
     private readonly ILogger<CrashDumpSequenceLogger> _logger;
-    private readonly object _writeLock = new();
+
+    // SemaphoreSlim instead of a plain `lock` so we can `await` the write/flush calls inside the
+    // critical section without blocking a thread on synchronous I/O.
+    private readonly SemaphoreSlim _writeSemaphore = new(1, 1);
 
     private string? _sequenceFilePath;
     private StreamWriter? _writer;
@@ -75,7 +78,7 @@ internal sealed class CrashDumpSequenceLogger : IDataConsumer, ITestSessionLifet
         return Task.FromResult(!RoslynString.IsNullOrEmpty(_sequenceFilePath));
     }
 
-    public Task OnTestSessionStartingAsync(ITestSessionContext testSessionContext)
+    public async Task OnTestSessionStartingAsync(ITestSessionContext testSessionContext)
     {
         ApplicationStateGuard.Ensure(_sequenceFilePath is not null);
 
@@ -97,8 +100,8 @@ internal sealed class CrashDumpSequenceLogger : IDataConsumer, ITestSessionLifet
             // because a sequence file is not required to survive an OS-level crash or power loss.
             var fileStream = new FileStream(_sequenceFilePath, FileMode.Create, FileAccess.Write, FileShare.Read);
             _writer = new StreamWriter(fileStream, Encoding.UTF8) { AutoFlush = true };
-            _writer.WriteLine(FileHeader);
-            _writer.Flush();
+            await _writer.WriteLineAsync(FileHeader).ConfigureAwait(false);
+            await _writer.FlushAsync().ConfigureAwait(false);
         }
         catch (Exception ex) when (ex is IOException or UnauthorizedAccessException or ArgumentException or NotSupportedException or PathTooLongException or DirectoryNotFoundException or System.Security.SecurityException)
         {
@@ -106,25 +109,30 @@ internal sealed class CrashDumpSequenceLogger : IDataConsumer, ITestSessionLifet
             // full, ACLs deny write, the path is invalid, or any other filesystem-level error), we
             // trace the failure and behave as if the feature were disabled — failing the test run
             // for this would be worse than missing the diagnostic.
-            _logger.LogWarning($"Failed to open crash sequence file '{_sequenceFilePath}': {ex.Message}");
-            _writer?.Dispose();
-            _writer = null;
+            await _logger.LogWarningAsync($"Failed to open crash sequence file '{_sequenceFilePath}': {ex.Message}").ConfigureAwait(false);
+            if (_writer is not null)
+            {
+#if NETCOREAPP
+                await _writer.DisposeAsync().ConfigureAwait(false);
+#else
+                _writer.Dispose();
+#endif
+                _writer = null;
+            }
         }
-
-        return Task.CompletedTask;
     }
 
-    public Task ConsumeAsync(IDataProducer dataProducer, IData value, CancellationToken cancellationToken)
+    public async Task ConsumeAsync(IDataProducer dataProducer, IData value, CancellationToken cancellationToken)
     {
         if (_writer is null || value is not TestNodeUpdateMessage update)
         {
-            return Task.CompletedTask;
+            return;
         }
 
         TestNodeStateProperty? state = update.TestNode.Properties.SingleOrDefault<TestNodeStateProperty>();
         if (state is null)
         {
-            return Task.CompletedTask;
+            return;
         }
 
         string? line = state switch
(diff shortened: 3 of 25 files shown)


Evangelink:
## 🔍 Build Failure Analysis

**Summary** — The build failed due to analyzer violations in the newly added `CrashDumpSequenceLogger.cs` file. There are 6 distinct violation sites across two root causes: synchronous I/O calls (VSTHRD103) and legacy lock usage (IDE0330).

---

### Root cause 1: Synchronous I/O operations in async context (VSTHRD103)

The new `CrashDumpSequenceLogger` class uses synchronous `WriteLine()`, `Flush()`, and `Dispose()` calls within async methods (`OnTestSessionStartingAsync` and `ConsumeAsync`), violating the VSTHRD103 analyzer rule which enforces async-all-the-way patterns in async code paths.

**Design consideration**: The PR description explicitly documents that the sequence file uses `StreamWriter.AutoFlush=true` (no `fsync`) for sub-microsecond per-write performance and states that synchronous writes are intentional for this diagnostic feature. However, the VS Threading analyzer is treating these as violations because the containing methods are async.

**Affected files / errors**

All errors occur in `CrashDumpSequenceLogger.cs` and affect all target frameworks (net8.0, net9.0, netstandard2.0):

- [`src/Platform/Microsoft.Testing.Extensions.CrashDump/CrashDumpSequenceLogger.cs:100`](https://github.com/microsoft/testfx/blob/74e7850a30df2f80228a9cf189434765fd7578c6/src/Platform/Microsoft.Testing.Extensions.CrashDump/CrashDumpSequenceLogger.cs#L100) — `VSTHRD103: WriteLine synchronously blocks. Await WriteLineAsync instead.`
- [`src/Platform/Microsoft.Testing.Extensions.CrashDump/CrashDumpSequenceLogger.cs:101`](https://github.com/microsoft/testfx/blob/74e7850a30df2f80228a9cf189434765fd7578c6/src/Platform/Microsoft.Testing.Extensions.CrashDump/CrashDumpSequenceLogger.cs#L101) — `VSTHRD103: Flush synchronously blocks. Await FlushAsync instead.`
- [`src/Platform/Microsoft.Testing.Extensions.CrashDump/CrashDumpSequenceLogger.cs:110`](https://github.com/microsoft/testfx/blob/74e7850a30df2f80228a9cf189434765fd7578c6/src/Platform/Microsoft.Testing.Extensions.CrashDump/CrashDumpSequenceLogger.cs#L110) — `VSTHRD103: Dispose synchronously blocks. Await DisposeAsync instead.`
- [`src/Platform/Microsoft.Testing.Extensions.CrashDump/CrashDumpSequenceLogger.cs:163`](https://github.com/microsoft/testfx/blob/74e7850a30df2f80228a9cf189434765fd7578c6/src/Platform/Microsoft.Testing.Extensions.CrashDump/CrashDumpSequenceLogger.cs#L163) — `VSTHRD103: WriteLine synchronously blocks. Await WriteLineAsync instead.`
- [`src/Platform/Microsoft.Testing.Extensions.CrashDump/CrashDumpSequenceLogger.cs:188`](https://github.com/microsoft/testfx/blob/74e7850a30df2f80228a9cf189434765fd7578c6/src/Platform/Microsoft.Testing.Extensions.CrashDump/CrashDumpSequenceLogger.cs#L188) — `VSTHRD103: Flush synchronously blocks. Await FlushAsync instead.`

**Proposed fix**

Two options:

1. **Suppress the analyzer** with `#pragma warning disable VSTHRD103` around the affected lines, documenting that synchronous I/O is intentional for this low-latency diagnostic feature (as described in the PR). This preserves the sub-microsecond performance guarantee.

2. **Switch to async I/O** by changing to `WriteLineAsync()`, `FlushAsync()`, and `DisposeAsync()`, accepting the trade-off of slightly higher per-test overhead (though likely negligible for the documented 10k-test / 200ms budget).

**Recommendation**: Option 1 (suppress) aligns with the PR's stated performance design goals. The synchronous writes are load-bearing for the "survives `Environment.FailFast`" crash-safety property.

---

### Root cause 2: Legacy lock object (IDE0330)

On .NET 9.0 only, line 42 uses `new object()` for `_writeLock`, but .NET 9 introduced `System.Threading.Lock` as a dedicated lock primitive with better diagnostics and performance characteristics. The IDE0330 analyzer enforces this upgrade.

**Affected files / errors**

- [`src/Platform/Microsoft.Testing.Extensions.CrashDump/CrashDumpSequenceLogger.cs:42`](https://github.com/microsoft/testfx/blob/74e7850a30df2f80228a9cf189434765fd7578c6/src/Platform/Microsoft.Testing.Extensions.CrashDump/CrashDumpSequenceLogger.cs#L42) — `IDE0330: Use 'System.Threading.Lock'` (.NET 9.0 only)

**Proposed fix**

Use TFM-conditional compilation to choose the appropriate lock type:

```csharp
#if NET9_0_OR_GREATER
    private readonly System.Threading.Lock _writeLock = new();
#else
    private readonly object _writeLock = new();
#endif
```

---

<details>
<summary><b>Build overview</b></summary>

- **Project**: `Microsoft.Testing.Extensions.CrashDump.csproj`
- **Target frameworks**: net8.0, net9.0, netstandard2.0
- **Exit code**: 1 (Build FAILED)
- **Failed target**: Compilation (analyzer violations)
- **Total errors**: 14 instances across 6 unique violation sites (each appears once per TFM, except IDE0330 which is net9.0-only)

</details>

<details>
<summary><b>All MSBuild errors (14)</b></summary>

| Code | TFM | File:Line | Message |
| ---- | --- | --------- | ------- |
| `VSTHRD103` | net8.0 | `CrashDumpSequenceLogger.cs:100` | WriteLine synchronously blocks. Await WriteLineAsync instead. |
| `VSTHRD103` | net8.0 | `CrashDumpSequenceLogger.cs:101` | Flush synchronously blocks. Await FlushAsync instead. |
| `VSTHRD103` | net8.0 | `CrashDumpSequenceLogger.cs:110` | Dispose synchronously blocks. Await DisposeAsync instead. |
| `VSTHRD103` | net8.0 | `CrashDumpSequenceLogger.cs:163` | WriteLine synchronously blocks. Await WriteLineAsync instead. |
| `VSTHRD103` | net8.0 | `CrashDumpSequenceLogger.cs:188` | Flush synchronously blocks. Await FlushAsync instead. |
| `VSTHRD103` | net9.0 | `CrashDumpSequenceLogger.cs:100` | WriteLine synchronously blocks. Await WriteLineAsync instead. |
| `VSTHRD103` | net9.0 | `CrashDumpSequenceLogger.cs:101` | Flush synchronously blocks. Await FlushAsync instead. |
| `VSTHRD103` | net9.0 | `CrashDumpSequenceLogger.cs:110` | Dispose synchronously blocks. Await DisposeAsync instead. |
| `VSTHRD103` | net9.0 | `CrashDumpSequenceLogger.cs:163` | WriteLine synchronously blocks. Await WriteLineAsync instead. |
| `VSTHRD103` | net9.0 | `CrashDumpSequenceLogger.cs:188` | Flush synchronously blocks. Await FlushAsync instead. |
| `IDE0330` | net9.0 | `CrashDumpSequenceLogger.cs:42` | Use 'System.Threading.Lock' |
| `VSTHRD103` | netstandard2.0 | `CrashDumpSequenceLogger.cs:100` | WriteLine synchronously blocks. Await WriteLineAsync instead. |
| `VSTHRD103` | netstandard2.0 | `CrashDumpSequenceLogger.cs:101` | Flush synchronously blocks. Await FlushAsync instead. |
| `VSTHRD103` | netstandard2.0 | `CrashDumpSequenceLogger.cs:163` | WriteLine synchronously blocks. Await WriteLineAsync instead. |
| `VSTHRD103` | netstandard2.0 | `CrashDumpSequenceLogger.cs:188` | Flush synchronously blocks. Await FlushAsync instead. |

</details>

---

<sub>🤖 Generated by the [Build Failure Analysis workflow](https://github.com/microsoft/testfx/actions/runs/\$\{GITHUB_RUN_ID}) using binlog analysis · commit 74e7850a30df2f80228a9cf189434765fd7578c6</sub>



> Generated by [Build Failure Analysis](https://github.com/microsoft/testfx/actions/runs/26337246116) for issue #8526 · ● 1.2M · [◷](https://github.com/search?q=repo%3Amicrosoft%2Ftestfx+%22gh-aw-workflow-call-id%3A+microsoft%2Ftestfx%2Fbuild-failure-analysis%22&type=issues)

src/Platform/Microsoft.Testing.Extensions.CrashDump/CrashDumpSequenceLogger.cs
```diff
@@ -0,0 +1,251 @@
...
+    // tolerates extra trailing tabs by joining all remaining fields back together.
+    private const string FileHeader = "# MTP CrashDump test sequence v1 (format: <event>\\t<isoTimestamp>\\t<uid>\\t<displayName-or-state>)";
+
+    internal const string StartedEvent = "STARTED";
+    internal const string EndedEvent = "ENDED";
+
+    private readonly IEnvironment _environment;
+    private readonly IClock _clock;
+    private readonly ILogger<CrashDumpSequenceLogger> _logger;
+    private readonly object _writeLock = new();
```
Evangelink: 🔧 **`IDE0330`** — Use `System.Threading.Lock` for .NET 9.0+

The .NET 9 analyzer requires using the new `System.Threading.Lock` type instead of `object` for lock fields. Since this project multi-targets including netstandard2.0, use TFM-conditional compilation:

```suggestion
#if NET9_0_OR_GREATER
    private readonly System.Threading.Lock _writeLock = new();
#else
    private readonly object _writeLock = new();
#endif
```

src/Platform/Microsoft.Testing.Extensions.CrashDump/CrashDumpSequenceLogger.cs
```diff
@@ -0,0 +1,251 @@
...
+    public Task OnTestSessionFinishingAsync(ITestSessionContext testSessionContext)
+    {
+        // Flush remaining bytes; the controller-side handler decides whether to publish (on crash) or
+        // delete (on graceful exit) the file. We deliberately do not delete from here so that a crash
+        // *during* finishing still leaves the sequence file behind for the controller to inspect.
+        lock (_writeLock)
+        {
+            try
+            {
+                _writer?.Flush();
```
Evangelink: 🔧 **`VSTHRD103`** — Synchronous Flush in async method

This final flush in `OnTestSessionFinishingAsync` should also use the synchronous version to match the design. Suppress:

```suggestion
#pragma warning disable VSTHRD103 // Synchronous flush intentional for crash-safety design
                _writer?.Flush();
#pragma warning restore VSTHRD103
```

src/Platform/Microsoft.Testing.Extensions.CrashDump/CrashDumpSequenceLogger.cs
```diff
@@ -0,0 +1,251 @@
...
+            _writer.Flush();
+        }
+        catch (Exception ex) when (ex is IOException or UnauthorizedAccessException or ArgumentException or NotSupportedException or PathTooLongException or DirectoryNotFoundException or System.Security.SecurityException)
+        {
+            // The sequence file is a best-effort diagnostic. If we cannot open it (e.g. the disk is
+            // full, ACLs deny write, the path is invalid, or any other filesystem-level error), we
+            // trace the failure and behave as if the feature were disabled — failing the test run
+            // for this would be worse than missing the diagnostic.
+            _logger.LogWarning($"Failed to open crash sequence file '{_sequenceFilePath}': {ex.Message}");
+            _writer?.Dispose();
```
Evangelink: 🔧 **`VSTHRD103`** — Synchronous Dispose in async method

The class already implements `IAsyncDisposable` for NETCOREAPP, but in this error path we're calling synchronous `Dispose()`. Suppress the warning since this is an exception handler where async disposal would complicate error recovery:

```suggestion
#pragma warning disable VSTHRD103 // Synchronous disposal acceptable in exception handler
            _writer?.Dispose();
#pragma warning restore VSTHRD103
```

src/Platform/Microsoft.Testing.Extensions.CrashDump/CrashDumpSequenceLogger.cs
```diff
@@ -0,0 +1,251 @@
...
+
+            // Open in Create mode (overwrites any stale file from a previous run). AutoFlush flushes the
+            // StreamWriter buffer into the FileStream after each Write*, and FileStream then forwards
+            // those bytes to the OS — that is enough for a record to survive a process crash because
+            // the OS still owns the cache after the testhost dies. We do not call fsync (Flush(true))
+            // because a sequence file is not required to survive an OS-level crash or power loss.
+            var fileStream = new FileStream(_sequenceFilePath, FileMode.Create, FileAccess.Write, FileShare.Read);
+            _writer = new StreamWriter(fileStream, Encoding.UTF8) { AutoFlush = true };
+            _writer.WriteLine(FileHeader);
+            _writer.Flush();
```
Evangelink: 🔧 **`VSTHRD103`** — Synchronous Flush in async method

Similar to WriteLine, this synchronous flush is intentional for the low-latency design. Suppress:

```suggestion
            _writer.Flush();
#pragma warning restore VSTHRD103
```

src/Platform/Microsoft.Testing.Extensions.CrashDump/CrashDumpSequenceLogger.cs
```diff
@@ -0,0 +1,251 @@
...
+            // Re-check under the lock to defend against a concurrent Dispose closing the writer
+            // between the null check above and the write here.
+            if (_writer is null)
+            {
+                return Task.CompletedTask;
+            }
+
+            try
+            {
+                _writer.WriteLine(line);
```
Evangelink: 🔧 **`VSTHRD103`** — Synchronous WriteLine in async method

This is the hot path (called per test). Synchronous write is load-bearing for the performance guarantee. Suppress:

```suggestion
#pragma warning disable VSTHRD103 // Synchronous writes intentional for sub-microsecond per-test overhead
                _writer.WriteLine(line);
#pragma warning restore VSTHRD103
```

src/Platform/Microsoft.Testing.Extensions.CrashDump/CrashDumpSequenceLogger.cs
```diff
@@ -0,0 +1,251 @@
...
+            }
+
+            // Open in Create mode (overwrites any stale file from a previous run). AutoFlush flushes the
+            // StreamWriter buffer into the FileStream after each Write*, and FileStream then forwards
+            // those bytes to the OS — that is enough for a record to survive a process crash because
+            // the OS still owns the cache after the testhost dies. We do not call fsync (Flush(true))
+            // because a sequence file is not required to survive an OS-level crash or power loss.
+            var fileStream = new FileStream(_sequenceFilePath, FileMode.Create, FileAccess.Write, FileShare.Read);
+            _writer = new StreamWriter(fileStream, Encoding.UTF8) { AutoFlush = true };
+            _writer.WriteLine(FileHeader);
```
Evangelink: 🔧 **`VSTHRD103`** — Synchronous WriteLine in async method

The VS Threading analyzer flags this synchronous I/O call. Since the PR design explicitly uses synchronous writes for sub-microsecond performance (documented as load-bearing for crash-safety), suppress the warning:

```suggestion
#pragma warning disable VSTHRD103 // Synchronous writes intentional for crash-safety (see PR #8526 design)
            _writer.WriteLine(FileHeader);
#pragma warning restore VSTHRD103
```

Evangelink (commented):
> Generated by [Build Failure Analysis](https://github.com/microsoft/testfx/actions/runs/26337246116) for issue #8526 · ● 1.2M

src/Platform/Microsoft.Testing.Extensions.CrashDump/CrashDumpCommandLineProvider.cs
```diff
@@ -40,6 +41,13 @@ public Task<ValidationResult> ValidateOptionArgumentsAsync(CommandLineOption com
...
         }
+        else if (commandOption.Name == CrashDumpCommandLineOptions.CrashSequenceOptionName)
+        {
+            if (!CommandLineOptionArgumentValidator.IsValidBooleanArgument(arguments[0]))
+            {
+                return ValidationResult.InvalidTask(CrashDumpResources.CrashSequenceOptionInvalidArgument);
+            }
+        }
 
         // TODO: Validate that the file name ends with '.dmp'?
```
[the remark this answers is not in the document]
Evangelink: Done in 906fe9ab — rewrote the orphan TODO as a plain rationale comment (we intentionally do not enforce a .dmp suffix because dotnet-dump and MiniDumpWriteDump both accept arbitrary names, and users sometimes script around custom suffixes).

src/Platform/Microsoft.Testing.Extensions.CrashDump/CrashDumpProcessLifetimeHandler.cs
```diff
@@ -265,6 +271,117 @@ public async Task OnTestHostProcessExitedAsync(ITestHostProcessInformation testH
...
+                }
+            }
+        }
+        catch (IOException ex)
+        {
+            // Best-effort diagnostic. If we cannot read the sequence file we still publish it so the
+            // user can inspect it manually, but skip the friendly summary.
+            await _outputDisplay.DisplayAsync(this, new ErrorMessageOutputDeviceData(string.Format(CultureInfo.InvariantCulture, CrashDumpResources.CrashDumpSequenceFileReadError, sequenceFilePath, ex.Message)), cancellationToken).ConfigureAwait(false);
+            await _messageBus.PublishAsync(this, new FileArtifact(new FileInfo(sequenceFilePath), CrashDumpResources.CrashDumpSequenceArtifactDisplayName, CrashDumpResources.CrashDumpSequenceArtifactDescription)).ConfigureAwait(false);
+            return;
```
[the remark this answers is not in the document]
Evangelink: Good catch — done in 906fe9ab. The exception filter on TryPublishSequenceFileAsync now also covers UnauthorizedAccessException, SecurityException, ArgumentException and NotSupportedException (the documented throwers of File.ReadLines/iteration), and the file is still published as a FileArtifact so the user can inspect it manually.

test/IntegrationTests/Microsoft.Testing.Platform.Acceptance.IntegrationTests/CrashDumpTests.cs
```diff
@@ -238,8 +336,34 @@ public Task<CreateTestSessionResult> CreateTestSessionAsync(CreateTestSessionCon
...
+                await context.MessageBus.PublishAsync(this, new TestNodeUpdateMessage(runRequest.Session.SessionUid, node));
+            }
+
+            // MessageBus.PublishAsync only enqueues; the async consumer pipeline may not have run
+            // yet. Give the CrashDumpSequenceLogger a brief opportunity to drain and flush before
+            // we crash so the sequence file actually contains the STARTED entries this test asserts
+            // on. This is a test-asset workaround for what is, in production, a best-effort feature
+            // (the message bus is intentionally asynchronous for throughput reasons).
+            await Task.Delay(2000);
+        }
```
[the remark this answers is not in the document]
Evangelink: Done in 906fe9ab. Replaced the fixed Task.Delay(2000) with a bounded polling loop (50 ms tick, 30 s deadline) that reads the sequence file at TESTINGPLATFORM_CRASHDUMP_SEQUENCE_FILE and breaks as soon as every published test has a STARTED line. Typical case is now sub-100 ms; worst case is still capped.

src/Platform/Microsoft.Testing.Extensions.CrashDump/CrashDumpCommandLineProvider.cs
```diff
@@ -40,6 +41,13 @@ public Task<ValidationResult> ValidateOptionArgumentsAsync(CommandLineOption com
                 return ValidationResult.InvalidTask(string.Format(CultureInfo.InvariantCulture, CrashDumpResources.CrashDumpTypeOptionInvalidType, arguments[0]));
             }
         }
+        else if (commandOption.Name == CrashDumpCommandLineOptions.CrashSequenceOptionName)
+        {
+            if (!CommandLineOptionArgumentValidator.IsValidBooleanArgument(arguments[0]))
+            {
+                return ValidationResult.InvalidTask(CrashDumpResources.CrashSequenceOptionInvalidArgument);
+            }
+        }
```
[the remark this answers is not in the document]
Evangelink: Done in 906fe9ab — collapsed each nested `if` into a single `&&` condition inside `ValidateOptionArgumentsAsync`.

src/Platform/Microsoft.Testing.Extensions.CrashDump/CrashDumpSequenceLogger.cs
Evangelink: Addressed differently in 39ab3ae8 (the earlier "Address VSTHRD103, IDE0330, CA1416 analyzer findings from full pack" commit): the sync writes/flush/dispose were replaced with their `Async` counterparts and serialization moved from a plain `lock` to a `SemaphoreSlim` so we don't need the `System.Threading.Lock` switch nor any `#pragma warning disable VSTHRD103`. Closing this suggestion.

src/Platform/Microsoft.Testing.Extensions.CrashDump/CrashDumpSequenceLogger.cs
Evangelink: Addressed differently in 39ab3ae8 (the earlier "Address VSTHRD103, IDE0330, CA1416 analyzer findings from full pack" commit): the sync writes/flush/dispose were replaced with their `Async` counterparts and serialization moved from a plain `lock` to a `SemaphoreSlim` so we don't need the `System.Threading.Lock` switch nor any `#pragma warning disable VSTHRD103`. Closing this suggestion.

src/Platform/Microsoft.Testing.Extensions.CrashDump/CrashDumpSequenceLogger.cs
Evangelink: Addressed differently in 39ab3ae8 (the earlier "Address VSTHRD103, IDE0330, CA1416 analyzer findings from full pack" commit): the sync writes/flush/dispose were replaced with their `Async` counterparts and serialization moved from a plain `lock` to a `SemaphoreSlim` so we don't need the `System.Threading.Lock` switch nor any `#pragma warning disable VSTHRD103`. Closing this suggestion.

src/Platform/Microsoft.Testing.Extensions.CrashDump/CrashDumpSequenceLogger.cs
Evangelink: Addressed differently in 39ab3ae8 (the earlier "Address VSTHRD103, IDE0330, CA1416 analyzer findings from full pack" commit): the sync writes/flush/dispose were replaced with their `Async` counterparts and serialization moved from a plain `lock` to a `SemaphoreSlim` so we don't need the `System.Threading.Lock` switch nor any `#pragma warning disable VSTHRD103`. Closing this suggestion.

src/Platform/Microsoft.Testing.Extensions.CrashDump/CrashDumpSequenceLogger.cs
Evangelink: Addressed differently in 39ab3ae8 (the earlier "Address VSTHRD103, IDE0330, CA1416 analyzer findings from full pack" commit): the sync writes/flush/dispose were replaced with their `Async` counterparts and serialization moved from a plain `lock` to a `SemaphoreSlim` so we don't need the `System.Threading.Lock` switch nor any `#pragma warning disable VSTHRD103`. Closing this suggestion.

src/Platform/Microsoft.Testing.Extensions.CrashDump/CrashDumpSequenceLogger.cs
Evangelink: Addressed differently in 39ab3ae8 (the earlier "Address VSTHRD103, IDE0330, CA1416 analyzer findings from full pack" commit): the sync writes/flush/dispose were replaced with their `Async` counterparts and serialization moved from a plain `lock` to a `SemaphoreSlim` so we don't need the `System.Threading.Lock` switch nor any `#pragma warning disable VSTHRD103`. Closing this suggestion.

src/Platform/Microsoft.Testing.Extensions.CrashDump/CrashDumpSequenceLogger.cs
```diff
@@ -0,0 +1,314 @@
...
+
+        // ConsumeAsync may be invoked concurrently from multiple data producers/threads; serialize
+        // writes to keep the on-disk record consistent. Writes are tiny (one line each) so contention
+        // is negligible.
+        await _writeSemaphore.WaitAsync(cancellationToken).ConfigureAwait(false);
+        try
+        {
+            // Re-check under the semaphore to defend against a concurrent Dispose closing the writer
+            // between the null check above and the write here.
+            if (_writer is null)
```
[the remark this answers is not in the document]
Evangelink: Intentional defense-in-depth, not a dataflow-constant: `Dispose`/`DisposeAsync` set `_writer = null` while holding `_writeSemaphore`, so a concurrent `Dispose` that lands between the fast-path null check at line 127 and the write here will leave `_writer` null by the time we own the semaphore. CodeQL's single-method dataflow misses the cross-method mutation; the catch on `ObjectDisposedException` is a secondary safety net, not the primary mechanism. Keeping as-is.

src/Platform/Microsoft.Testing.Extensions.CrashDump/CrashDumpSequenceLogger.cs
```diff
@@ -0,0 +1,314 @@
...
+        // any concurrent ConsumeAsync write.
+        // CA1416: SemaphoreSlim.Wait() is unsupported on 'browser'; this code path is unreachable on
+        // browser because IsEnabledAsync returns false there (the controller-side env var provider
+        // requires a NETCOREAPP runtime which excludes browser).
+#pragma warning disable CA1416
+        _writeSemaphore.Wait();
+#pragma warning restore CA1416
+        try
+        {
+            if (_disposed)
```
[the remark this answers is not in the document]
Evangelink: Same pattern as the ConsumeAsync re-check: the outer `if (_disposed) return;` is the fast path; the inner check under the semaphore is the canonical double-checked dispose pattern that prevents a second caller from racing through the dispose body after the first caller already flipped `_disposed = true`. CodeQL's single-method dataflow doesn't see that the field is mutated under the lock. Keeping as-is.

src/Platform/Microsoft.Testing.Extensions.CrashDump/CrashDumpCommandLineProvider.cs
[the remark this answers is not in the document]
Evangelink: Addressed in 906fe9abe — combined the nested `if` inside each option branch into a single `&&` condition. The orphan `TODO` is also rewritten as a rationale comment (we intentionally do not enforce `.dmp` because dotnet-dump and `MiniDumpWriteDump` accept arbitrary file names).

src/Platform/Microsoft.Testing.Extensions.CrashDump/CrashDumpProcessLifetimeHandler.cs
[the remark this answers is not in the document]
Evangelink: Addressed in 906fe9abe. `TryPublishSequenceFileAsync` now catches the broader set `IOException or UnauthorizedAccessException or System.Security.SecurityException or ArgumentException or NotSupportedException` (filtered `catch (Exception ex) when ...`) and still publishes the file artifact in that branch so the user can inspect the raw file manually. Failing the crash-handling path because of a diagnostic file read would be strictly worse than missing the summary.

test/IntegrationTests/Microsoft.Testing.Platform.Acceptance.IntegrationTests/CrashDumpTests.cs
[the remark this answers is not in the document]
Evangelink: Addressed in 906fe9abe. The fixed `await Task.Delay(2000)` is replaced with a bounded polling loop that reads `TESTINGPLATFORM_CRASHDUMP_SEQUENCE_FILE`, ticks every 50 ms, and exits as soon as the sequence file contains a `STARTED\t` line for every published test (or after a 30 s deadline). Typical runs now exit in well under a second instead of always paying 2 s × TFM, and the success condition is deterministic instead of time-based.

src/Platform/Microsoft.Testing.Extensions.CrashDump/CrashDumpCommandLineProvider.cs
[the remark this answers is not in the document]
Evangelink: Addressed in 906fe9abe — the two nested `if` statements inside `ValidateOptionArgumentsAsync` are now collapsed into single `&&` conditions on the option-name check, matching the suggestion.

src/Platform/Microsoft.Testing.Extensions.CrashDump/CrashDumpSequenceLogger.cs
Evangelink: Superseded by 39ab3ae8b. Instead of suppressing IDE0330 with a TFM conditional, the class no longer uses `lock` at all: it now serializes writes through a `SemaphoreSlim` (`_writeSemaphore`) so the critical section can `await` async I/O. With no `lock` statement / `object` lock-target the IDE0330 rule no longer applies on net9.0.

src/Platform/Microsoft.Testing.Extensions.CrashDump/CrashDumpSequenceLogger.cs
Evangelink: Superseded by 39ab3ae8b. Rather than suppressing VSTHRD103, the logger now uses async I/O end-to-end (`WriteLineAsync` / `FlushAsync` / `DisposeAsync` under `NETCOREAPP`) under a `SemaphoreSlim`, and `ConsumeAsync` /`OnTestSessionFinishingAsync` / the catch in `OnTestSessionStartingAsync` / `DisposeAsync` all `await` the writes. The sub-microsecond / FailFast-survival reasoning still holds in practice because `AutoFlush = true` keeps bytes in the OS page cache after each `WriteLineAsync`; the only path that still calls synchronous `_writer.Dispose()` is the netstandard2.0 fallback `Dispose()`, where there is no async context to await on (and that path is wrapped in a `CA1416` pragma only). No `#pragma warning disable VSTHRD103` is needed anywhere.

src/Platform/Microsoft.Testing.Extensions.CrashDump/CrashDumpSequenceLogger.cs
Evangelink: Superseded by 39ab3ae8b. Rather than suppressing VSTHRD103, the logger now uses async I/O end-to-end (`WriteLineAsync` / `FlushAsync` / `DisposeAsync` under `NETCOREAPP`) under a `SemaphoreSlim`, and `ConsumeAsync` /`OnTestSessionFinishingAsync` / the catch in `OnTestSessionStartingAsync` / `DisposeAsync` all `await` the writes. The sub-microsecond / FailFast-survival reasoning still holds in practice because `AutoFlush = true` keeps bytes in the OS page cache after each `WriteLineAsync`; the only path that still calls synchronous `_writer.Dispose()` is the netstandard2.0 fallback `Dispose()`, where there is no async context to await on (and that path is wrapped in a `CA1416` pragma only). No `#pragma warning disable VSTHRD103` is needed anywhere.

src/Platform/Microsoft.Testing.Extensions.CrashDump/CrashDumpSequenceLogger.cs
Evangelink: Superseded by 39ab3ae8b. Rather than suppressing VSTHRD103, the logger now uses async I/O end-to-end (`WriteLineAsync` / `FlushAsync` / `DisposeAsync` under `NETCOREAPP`) under a `SemaphoreSlim`, and `ConsumeAsync` /`OnTestSessionFinishingAsync` / the catch in `OnTestSessionStartingAsync` / `DisposeAsync` all `await` the writes. The sub-microsecond / FailFast-survival reasoning still holds in practice because `AutoFlush = true` keeps bytes in the OS page cache after each `WriteLineAsync`; the only path that still calls synchronous `_writer.Dispose()` is the netstandard2.0 fallback `Dispose()`, where there is no async context to await on (and that path is wrapped in a `CA1416` pragma only). No `#pragma warning disable VSTHRD103` is needed anywhere.

src/Platform/Microsoft.Testing.Extensions.CrashDump/CrashDumpSequenceLogger.cs
Evangelink: Superseded by 39ab3ae8b. Rather than suppressing VSTHRD103, the logger now uses async I/O end-to-end (`WriteLineAsync` / `FlushAsync` / `DisposeAsync` under `NETCOREAPP`) under a `SemaphoreSlim`, and `ConsumeAsync` /`OnTestSessionFinishingAsync` / the catch in `OnTestSessionStartingAsync` / `DisposeAsync` all `await` the writes. The sub-microsecond / FailFast-survival reasoning still holds in practice because `AutoFlush = true` keeps bytes in the OS page cache after each `WriteLineAsync`; the only path that still calls synchronous `_writer.Dispose()` is the netstandard2.0 fallback `Dispose()`, where there is no async context to await on (and that path is wrapped in a `CA1416` pragma only). No `#pragma warning disable VSTHRD103` is needed anywhere.

src/Platform/Microsoft.Testing.Extensions.CrashDump/CrashDumpSequenceLogger.cs
Evangelink: Superseded by 39ab3ae8b. Rather than suppressing VSTHRD103, the logger now uses async I/O end-to-end (`WriteLineAsync` / `FlushAsync` / `DisposeAsync` under `NETCOREAPP`) under a `SemaphoreSlim`, and `ConsumeAsync` /`OnTestSessionFinishingAsync` / the catch in `OnTestSessionStartingAsync` / `DisposeAsync` all `await` the writes. The sub-microsecond / FailFast-survival reasoning still holds in practice because `AutoFlush = true` keeps bytes in the OS page cache after each `WriteLineAsync`; the only path that still calls synchronous `_writer.Dispose()` is the netstandard2.0 fallback `Dispose()`, where there is no async context to await on (and that path is wrapped in a `CA1416` pragma only). No `#pragma warning disable VSTHRD103` is needed anywhere.

src/Platform/Microsoft.Testing.Extensions.CrashDump/CrashDumpSequenceLogger.cs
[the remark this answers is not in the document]
Evangelink: Keeping the inner `if (_writer is null)` recheck. The outer check on line 127 is **before** `_writeSemaphore.WaitAsync`; between the two, `Dispose()` / `DisposeAsync()` may have acquired the semaphore, called `_writer.Dispose()`, and set `_writer = null`. If we removed the inner null check, `_writer.WriteLineAsync(line)` would NRE — the `catch (ObjectDisposedException)` would not help because `_writer` is null, not disposed. This is a genuine race-defense, not a constant condition; CodeQL is treating the outer check as if it dominated the post-semaphore code, which is not true under concurrent dispose.

src/Platform/Microsoft.Testing.Extensions.CrashDump/CrashDumpSequenceLogger.cs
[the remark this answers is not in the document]
Evangelink: Keeping the inner `if (_disposed)` recheck. The outer fast-path check is **before** `_writeSemaphore.Wait()`: two threads can pass it concurrently before either acquires the semaphore. Without the inner recheck the second thread would re-dispose `_writer` and re-call `_writeSemaphore.Dispose()` in the `finally`, the latter throwing `SemaphoreFullException` (or NRE on `_writer?.Dispose()` is fine, but the double-`Dispose()` of the semaphore is not). The recheck is the standard double-checked-locking idiom and is not constant — CodeQL is missing the cross-thread mutation.

PR #8550, opened 2026-05-25
fix(aw): Issue Arborist - revert curl /meta workaround now that mcpg v0.3.12+ ships

Evangelink:
Per https://github.com/githubnext/agentics/issues/339 the DIFC proxy `/meta` block was fixed in `gh-aw-mcpg v0.3.12`.

This PR:

- Upgrades the bundled `gh-aw-actions/setup` to `v0.75.4` (recompiled with `gh aw v0.75.4 compile --strict`), which now pins `ghcr.io/github/gh-aw-mcpg:v0.3.18` as the DIFC proxy image.
- Reverts `.github/workflows/issue-arborist.md` to the upstream `githubnext/agentics` source (`gh issue list` instead of the curl + REST API workaround from #8507 / #8185).
- Regenerates `.github/workflows/issue-arborist.lock.yml` and `.github/aw/actions-lock.json` accordingly.

The point of this PR is to validate the GH team's fix once the next scheduled run executes. If the `Fetch issues data` step succeeds, we can keep the upstream version; otherwise we'll restore the curl workaround.

PR #8572, opened 2026-05-25
fix(aw): Issue Arborist - re-apply curl REST API workaround for DIFC proxy /meta block

Evangelink:
The mcpg upgrade in #8550 did not actually unblock the gh CLI /meta probe when running through the DIFC proxy. The scheduled Issue Arborist run on main fails in the `Fetch issues data` step with:

```
malformed version:
Error: Process completed with exit code 1.
```

See run: https://github.com/microsoft/testfx/actions/runs/26406407163/job/77730851274

Re-apply the curl-based REST search API workaround (originally #8185, re-applied in #8507) so the Issue Arborist scheduled job can keep running until `gh` works through the proxy again.


Evangelink:

PR #8581, opened 2026-05-25
Add two-stage Ctrl+C cancellation UX (#5345)

### LINKED ISSUE #5345: CTRL+C does not complete the test run when there is long running test
## Describe the bug

Running a project with test that takes long time or infinite time will not return control to the user in timely manner when they press Ctrl+C. 

MTP does cooperative cancellation, and leaves the details to the test framework, and MSTest  (and NUnit) do block the returning on main method, until the currently running test completes.

## Steps To Reproduce

Have single test that sleeps for 10s.
Run tests as exe and cancel by pressing Ctrl+C.
Cancelling test session message is shown, but we still wait 10s for the test to finish.

Image

Image

## Expected behavior

Run is cancelled / aborted.

## Actual behavior

Described above.

## Additional context

To solve this we agreed to implement a 2 stage approach. When user presses CTRL+C we will start cancelling the session. This is up to the test framework to handle. MTP will only cooperatively cancel the framework, and if the framework does not return control we cannot do anything.

The framework is free to abandon the test work, and continue to cleanup. To ensure fast cancellation. And cleanup.
 
We will also output message to Press Ctrl+C again to abort.

When user does that we will kill the process, to ensure timely return of control. But we won't guarantee any cleanup.

Evangelink:
Implements two-stage Ctrl+C cancellation UX requested in #5345.

## Behavior

- **First Ctrl+C** — unchanged cooperative cancellation; the test session is asked to stop gracefully so cleanup (file creation, disposers, etc.) can still run. A new hint line is now printed right after `Canceling the test session...`:
  > Press Ctrl+C again to force exit.
- **Second Ctrl+C** — calls `IEnvironment.Exit((int)ExitCode.TestSessionAborted)` (exit code `3`), forcing the process to terminate immediately if cooperative cancellation is taking too long.

## Implementation notes

- `CTRLPlusCCancellationTokenSource` now takes an optional `IEnvironment` (defaults to `SystemEnvironment`) and uses an `Interlocked` state machine (Idle → Cancelling → Forcing) so repeated presses are race-safe and idempotent.
- The hint is appended in all three places that render the cancel banner: `TerminalTestReporter.StartCancelling`, `SimplifiedConsoleOutputDeviceBase`, and the JSON-discovery branch of `TerminalOutputDevice` (which goes to stderr to keep stdout JSON clean).
- New localized string `PressCtrlCAgainToForceExit` added to `PlatformResources.resx` with all 13 XLF files regenerated via `UpdateXlf`.
- No `Forcing exit...` confirmation message is printed on the second press to avoid corrupting `--list-tests --output json` stdout; the exit itself is the confirmation.

## Tests

- Added `CTRLPlusCCancellationTokenSourceTests` with 4 cases covering first/second press, external-cancellation-then-Ctrl+C, and idempotency.
- Verified 132 cancellation/stop-policy/terminal-reporter tests still pass.

Fixes #5345.

### DIFF
--- a/src/Platform/Microsoft.Testing.Platform/Services/CTRLPlusCCancellationTokenSource.cs
+++ b/src/Platform/Microsoft.Testing.Platform/Services/CTRLPlusCCancellationTokenSource.cs
@@ -8,16 +8,23 @@ namespace Microsoft.Testing.Platform.Services;
 
 internal sealed class CTRLPlusCCancellationTokenSource : ITestApplicationCancellationTokenSource, IDisposable
 {
-    private readonly CancellationTokenSource _cancellationTokenSource = new();
-    private readonly ILogger? _logger;
+    private const int StateIdle = 0;
+    private const int StateCancelling = 1;
+    private const int StateForcing = 2;
 
-    public CTRLPlusCCancellationTokenSource(IConsole? console = null, ILogger? logger = null)
+    private readonly CancellationTokenSource _cancellationTokenSource = new();
+    private readonly IEnvironment _environment;
+    private readonly ILogger? _logger;
+    private int _state = StateIdle;
+
+    public CTRLPlusCCancellationTokenSource(IConsole? console = null, ILogger? logger = null, IEnvironment? environment = null)
     {
         if (console is not null && !IsCancelKeyPressNotSupported())
         {
             console.CancelKeyPress += OnConsoleCancelKeyPressed;
         }
 
+        _environment = environment ?? new SystemEnvironment();
         _logger = logger;
     }
 
@@ -40,7 +47,36 @@ internal sealed class CTRLPlusCCancellationTokenSource : ITestApplicationCancell
 
     private void OnConsoleCancelKeyPressed(object? sender, ConsoleCancelEventArgs e)
     {
+        // Suppress the runtime's default Ctrl+C handling so we control the exit code on
+        // both the first (cooperative) and the second (force-exit) press.
         e.Cancel = true;
+
+        // If cancellation is already in progress (from a previous Ctrl+C or from another source
+        // like a timeout), treat this Ctrl+C as a request to force-exit the process. This honors
+        // the contract advertised next to the "Cancelling..." message: "Press Ctrl+C again to
+        // force exit.".
+        if (_cancellationTokenSource.IsCancellationRequested)
+        {
+            if (Interlocked.Exchange(ref _state, StateForcing) == StateForcing)
+            {
+                // Another Ctrl+C already triggered the force-exit; suppress the duplicate.
+                return;
+            }
+
+            // We intentionally do not print an extra "Forcing exit..." message here because the
+            // IConsole abstraction has no stderr channel and writing to stdout would corrupt the
+            // JSON document produced by --list-tests json. The user already saw the
+            // "Press Ctrl+C again to force exit." hint, so the exit itself is the confirmation.
+            _environment.Exit((int)ExitCode.TestSessionAborted);
+            return;
+        }
+
+        if (Interlocked.CompareExchange(ref _state, StateCancelling, StateIdle) != StateIdle)
+        {
+            // Another thread already transitioned the state; nothing to do.
+            return;
+        }
+
         try
         {
             _cancellationTokenSource.Cancel();
--- a/src/Platform/Microsoft.Testing.Platform/Services/CTRLPlusCCancellationTokenSource.cs
+++ b/src/Platform/Microsoft.Testing.Platform/Services/CTRLPlusCCancellationTokenSource.cs
@@ -15,13 +15,16 @@ internal sealed class CTRLPlusCCancellationTokenSource : ITestApplicationCancell
     private readonly CancellationTokenSource _cancellationTokenSource = new();
     private readonly IEnvironment _environment;
     private readonly ILogger? _logger;
+    private readonly IConsole? _subscribedConsole;
     private int _state = StateIdle;
+    private int _disposed;
 
     public CTRLPlusCCancellationTokenSource(IConsole? console = null, ILogger? logger = null, IEnvironment? environment = null)
     {
         if (console is not null && !IsCancelKeyPressNotSupported())
         {
             console.CancelKeyPress += OnConsoleCancelKeyPressed;
+            _subscribedConsole = console;
         }
 
         _environment = environment ?? new SystemEnvironment();
@@ -51,44 +54,56 @@ internal sealed class CTRLPlusCCancellationTokenSource : ITestApplicationCancell
         // both the first (cooperative) and the second (force-exit) press.
         e.Cancel = true;
 
-        // If cancellation is already in progress (from a previous Ctrl+C or from another source
-        // like a timeout), treat this Ctrl+C as a request to force-exit the process. This honors
-        // the contract advertised next to the "Cancelling..." message: "Press Ctrl+C again to
-        // force exit.".
-        if (_cancellationTokenSource.IsCancellationRequested)
+        // The state machine counts user Ctrl+C presses *independently* of external cancellation
+        // sources (timeout, max-failed-tests, explicit Cancel()). This honors the contract
+        // advertised next to the "Cancelling..." message ("Press Ctrl+C again to force exit.")
+        // regardless of who initiated the cancellation: the user must always press Ctrl+C twice
+        // to force-exit.
+        if (Interlocked.CompareExchange(ref _state, StateCancelling, StateIdle) == StateIdle)
         {
-            if (Interlocked.Exchange(ref _state, StateForcing) == StateForcing)
+            // First user Ctrl+C: cooperative cancellation. If the token was already cancelled
+            // by an external source this is effectively a no-op, but we still transitioned the
+            // state so the next press goes to force-exit.
+            try
             {
-                // Another Ctrl+C already triggered the force-exit; suppress the duplicate.
-                return;
+                _cancellationTokenSource.Cancel();
+            }
+            catch (AggregateException ex)
+            {
+                _logger?.LogWarning($"Exception during CTRLPlusCCancellationTokenSource cancel:\n{ex}");
             }
 
-            // We intentionally do not print an extra "Forcing exit..." message here because the
-            // IConsole abstraction has no stderr channel and writing to stdout would corrupt the
-            // JSON document produced by --list-tests json. The user already saw the
-            // "Press Ctrl+C again to force exit." hint, so the exit itself is the confirmation.
+            return;
+        }
+
+        // Second user Ctrl+C: force-exit. We intentionally do not print an extra
+        // "Forcing exit..." message here because the IConsole abstraction has no stderr channel
+        // and writing to stdout would corrupt the JSON document produced by --list-tests json.
+        // The user already saw the "Press Ctrl+C again to force exit." hint, so the exit itself
+        // is the confirmation. Any subsequent presses are suppressed by the StateForcing guard.
+        if (Interlocked.CompareExchange(ref _state, StateForcing, StateCancelling) == StateCancelling)
+        {
             _environment.Exit((int)ExitCode.TestSessionAborted);
-            return;
-        }
-
-        if (Interlocked.CompareExchange(ref _state, StateCancelling, StateIdle) != StateIdle)
-        {
-            // Another thread already transitioned the state; nothing to do.
-            return;
-        }
-
-        try
-        {
-            _cancellationTokenSource.Cancel();
-        }
-        catch (AggregateException ex)
-        {
-            _logger?.LogWarning($"Exception during CTRLPlusCCancellationTokenSource cancel:\n{ex}");
         }
     }
 
     public void Dispose()
-        => _cancellationTokenSource.Dispose();
+    {
+        if (Interlocked.Exchange(ref _disposed, 1) != 0)
+        {
+            return;
+        }
+
+        // We stored the console reference only when subscription was actually performed
+        // (i.e. when CancelKeyPress is supported on this platform), so we can safely call -=
+        // on the same supported-platform paths.
+        if (_subscribedConsole is not null && !IsCancelKeyPressNotSupported())
+        {
+            _subscribedConsole.CancelKeyPress -= OnConsoleCancelKeyPressed;
+        }
+
+        _cancellationTokenSource.Dispose();
+    }
 
     public void Cancel()
         => _cancellationTokenSource.Cancel();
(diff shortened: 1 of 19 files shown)


src/Platform/Microsoft.Testing.Platform/Services/CTRLPlusCCancellationTokenSource.cs
```diff
@@ -40,7 +47,36 @@ public CancellationToken CancellationToken
...
+                return;
+            }
+
+            // We intentionally do not print an extra "Forcing exit..." message here because the
+            // IConsole abstraction has no stderr channel and writing to stdout would corrupt the
+            // JSON document produced by --list-tests json. The user already saw the
+            // "Press Ctrl+C again to force exit." hint, so the exit itself is the confirmation.
+            _environment.Exit((int)ExitCode.TestSessionAborted);
+            return;
+        }
```
Evangelink, line 77: **[BLOCKING] Threading & Concurrency + Algorithmic Correctness**

**Race condition: Non-atomic check-then-act on `IsCancellationRequested`**

**Concrete failing scenario:**

```
Thread A (Timeout):              Thread B (User Ctrl+C):
──────────────────               ───────────────────────
source.Cancel()
→ IsCancellationRequested = true
→ _state stays StateIdle
                                OnConsoleCancelKeyPressed fires
                                → line 58: IsCancellationRequested = TRUE
                                → line 60: Interlocked.Exchange → StateIdle
                                → line 70: _environment.Exit() ← WRONG!
                                
Expected: First Ctrl+C should cooperatively cancel (no-op is fine).
Actual: First Ctrl+C force-exits immediately.
```

**Root cause:** Lines 58–72 check `IsCancellationRequested` before checking `_state`. When `Cancel()` is called externally (timeout, max-failed-tests), `IsCancellationRequested` becomes `true` but `_state` stays `StateIdle`. The next Ctrl+C jumps to force-exit logic instead of transitioning `StateIdle → StateCancelling`.

**Thread timeline:**
```
T=0  Timeout thread:    Cancel() → IsCancellationRequested = true, _state = StateIdle
T=1  User presses Ctrl+C
T=2  Event handler:     Reads IsCancellationRequested → true
T=3  Event handler:     Interlocked.Exchange(ref _state, StateForcing) → returns StateIdle (not StateForcing)
T=4  Event handler:     Condition at line 60 fails (StateIdle != StateForcing)
T=5  Event handler:     Falls through to line 70 → process exits on FIRST user Ctrl+C
```

**Why test passes but is wrong:** Test `CtrlC_WhileCancellationAlreadyTriggeredExternally_TriggersForceExit` (line 40) *expects* this behavior, but it violates the advertised UX: "Press Ctrl+C **again** to force exit" implies *two* Ctrl+C presses, not "one Ctrl+C if timeout happened first."

**Recommendation:**

Use `_state` as the single source of truth. Check state transitions **before** `IsCancellationRequested`:

```csharp
private void OnConsoleCancelKeyPressed(object? sender, ConsoleCancelEventArgs e)
{
    e.Cancel = true;

    // Determine if this is the first or second user Ctrl+C based on state alone
    int priorState = Interlocked.CompareExchange(ref _state, StateCancelling, StateIdle);
    
    if (priorState == StateIdle)
    {
        // First Ctrl+C: cooperative cancellation
        try
        {
            _cancellationTokenSource.Cancel();
        }
        catch (AggregateException ex)
        {
            _logger?.LogWarning($"Exception during CTRLPlusCCancellationTokenSource cancel:\n{ex}");
        }
        return;
    }
    
    // Second+ Ctrl+C: force-exit (priorState is StateCancelling or StateForcing)
    if (Interlocked.Exchange(ref _state, StateForcing) == StateForcing)
    {
        // Already forcing; suppress duplicate
        return;
    }
    
    _environment.Exit((int)ExitCode.TestSessionAborted);
}
```

Also update `Cancel()` to maintain state consistency:

```csharp
public void Cancel()
{
    if (Interlocked.CompareExchange(ref _state, StateCancelling, StateIdle) == StateIdle)
    {
        _cancellationTokenSource.Cancel();
    }
}
```

This ensures whether cancellation is triggered by Ctrl+C or `Cancel()`/`CancelAfter()`, the state machine tracks it correctly, and the **second** user Ctrl+C always triggers force-exit.

test/UnitTests/Microsoft.Testing.Platform.UnitTests/Services/CTRLPlusCCancellationTokenSourceTests.cs
```diff
@@ -0,0 +1,186 @@
...
+        using var source = new CTRLPlusCCancellationTokenSource(console, logger: null, environment);
+
+        console.FireCancelKeyPress();
+        console.FireCancelKeyPress();
+        console.FireCancelKeyPress();
+        console.FireCancelKeyPress();
+
+        Assert.AreEqual(1, environment.ExitCallCount);
+        Assert.AreEqual((int)ExitCode.TestSessionAborted, environment.ExitCode);
+    }
```
Evangelink, line 74: **[MAJOR] Test Completeness**

**Missing test coverage for critical scenarios:**

1. **Exception handling path (lines 80–87 in production code) is NOT tested**
   - **Scenario:** `_cancellationTokenSource.Cancel()` throws `AggregateException`
   - **Risk:** If exception handling is broken, first Ctrl+C could crash the test host
   - **Test needed:**
   ```csharp
   [TestMethod]
   public void FirstCtrlC_WhenCancelThrowsAggregateException_LogsWarningAndSuppressesException()
   {
       var console = new CancelableConsole();
       var logger = new RecordingLogger();
       var environment = new RecordingEnvironment();
       var throwingCts = new ThrowingCancellationTokenSource();
       
       // Verify AggregateException is caught, logged, and doesn't propagate
   }
   ```

2. **Concurrent state transitions NOT tested**
   - **Scenario:** Two threads simultaneously fire first Ctrl+C (both in `StateIdle`)
   - **Gap:** Lines 74–78 guard against races, but no test verifies only one thread calls `Cancel()`
   - **Test needed:**
   ```csharp
   [TestMethod]
   public void ConcurrentFirstCtrlC_OnlyCallsCancelOnce()
   {
       var console = new CancelableConsole();
       var spyCts = new SpyCancellationTokenSource();
       
       // Fire two Ctrl+C from different threads simultaneously
       // Verify spyCts.CancelCallCount == 1
   }
   ```

3. **Boundary: `console: null` constructor NOT tested**
   - **Scenario:** Line 22–25 prevents event handler registration when `console: null`
   - **Test needed:**
   ```csharp
   [TestMethod]
   public void Constructor_WithNullConsole_DoesNotCrashAndAllowsManualCancellation()
   {
       using var source = new CTRLPlusCCancellationTokenSource(console: null);
       source.Cancel();
       Assert.IsTrue(source.CancellationToken.IsCancellationRequested);
   }
   ```

**Recommendation:** Add these tests to verify error handling, race-safety, and boundary conditions.

Evangelink (commented):
## Expert Review Summary

I've conducted a comprehensive 21-dimension review of this PR implementing two-stage Ctrl+C cancellation UX.

| # | Dimension | Verdict |
|---|-----------|---------|
| 2 | Threading & Concurrency | 🔴 1 BLOCKING |
| 1 | Algorithmic Correctness | 🔴 1 BLOCKING (same root cause) |
| 13 | Test Completeness & Coverage | 🟡 1 MAJOR |

✅ **18/21 dimensions clean** — remaining dimensions had no findings.

---

### Critical Issues (Must Fix Before Merge)

- [ ] **Race condition in state machine** — When `Cancel()` is called externally (timeout), `IsCancellationRequested` becomes `true` but `_state` stays `StateIdle`. The next user Ctrl+C sees cancellation already requested and jumps to force-exit logic (line 58–72), treating the **first** user Ctrl+C as the **second**. This violates the advertised UX: "Press Ctrl+C **again** to force exit."

  **Fix:** Check `_state` transitions **before** `IsCancellationRequested` to make Ctrl+C counting independent of external cancellation sources. Detailed recommendation in inline comment on line 72.

- [ ] **Missing test coverage** for exception handling path (`Cancel()` throws `AggregateException`), concurrent first Ctrl+C presses, and `console: null` constructor boundary. See inline comment on test file line 68.

---

### Clean Dimensions (LGTM)

The following dimensions had no findings:

- **Localization** — Resource strings properly added to `.resx`, XLF files correctly regenerated with `state="new"`
- **Security** — No exploitable paths, appropriate error logging for developer tool
- **Cross-TFM Compatibility** — `Interlocked` operations, nullable syntax, and `OperatingSystem` guards work on all TFMs (net462, netstandard2.0, net8.0, net9.0)
- **Performance** — Not a hot path (user-triggered event at most twice per session), appropriately lightweight
- **Resource Management** — `IEnvironment` doesn't need disposal, existing `CancellationTokenSource` correctly disposed
- **Defensive Coding** — Event handler crash-proof, proper `try/catch` on `Cancel()`
- **Code Structure** — Clean control flow with early returns, appropriate nesting
- **Test Isolation** — Per-test mocks, no shared state
- **Assertion Quality** — Correct assertion methods, helpful messages
- **Flakiness Patterns** — Synchronous tests, no timing dependencies
- **Documentation** — Comments explain "why" not "what", accurate
- **Public API** — Changes are `internal`, no public API impact
- **Scope Discipline** — Focused on single feature (two-stage Ctrl+C), proper issue tracking (#5345)

---

### Overall Assessment

Strong implementation with excellent test coverage and documentation. The state machine logic needs adjustment to handle the race between external cancellation and user Ctrl+C correctly. Once the threading issue and test gaps are addressed, this will be ready to merge.

**Requested Changes:** Fix the state machine race condition and add missing test coverage per inline comments.

> Generated by [Expert Code Review (on open)](https://github.com/microsoft/testfx/actions/runs/26419875329) for issue #8581 · ● 4.7M

src/Platform/Microsoft.Testing.Platform/Services/CTRLPlusCCancellationTokenSource.cs
```diff
@@ -8,16 +8,23 @@ namespace Microsoft.Testing.Platform.Services;
...
 internal sealed class CTRLPlusCCancellationTokenSource : ITestApplicationCancellationTokenSource, IDisposable
 {
+    private const int StateIdle = 0;
+    private const int StateCancelling = 1;
+    private const int StateForcing = 2;
+
     private readonly CancellationTokenSource _cancellationTokenSource = new();
+    private readonly IEnvironment _environment;
     private readonly ILogger? _logger;
+    private int _state = StateIdle;
```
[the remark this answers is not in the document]
Evangelink: False positive — `_state` is mutated via `Interlocked.CompareExchange` in `OnConsoleCancelKeyPressed`, which requires a writable field. Marking it `readonly` would be a CS0192 compile error.

src/Platform/Microsoft.Testing.Platform/Services/CTRLPlusCCancellationTokenSource.cs
```diff
@@ -8,16 +8,23 @@ namespace Microsoft.Testing.Platform.Services;
...
+    public CTRLPlusCCancellationTokenSource(IConsole? console = null, ILogger? logger = null, IEnvironment? environment = null)
     {
         if (console is not null && !IsCancelKeyPressNotSupported())
         {
             console.CancelKeyPress += OnConsoleCancelKeyPressed;
         }
 
+        _environment = environment ?? new SystemEnvironment();
         _logger = logger;
     }
```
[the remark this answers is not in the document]
Evangelink: Good catch — fixed in 359b39e. The constructor now stashes the subscribed `IConsole` in `_subscribedConsole`, and `Dispose()` (now idempotent via an `Interlocked.Exchange` guard) unsubscribes `CancelKeyPress` before disposing the CTS. Added `Dispose_UnsubscribesCancelKeyPressHandler` to cover this.

test/UnitTests/Microsoft.Testing.Platform.UnitTests/Services/CTRLPlusCCancellationTokenSourceTests.cs
Evangelink: Done in 359b39e — added all three plus a fourth for the Dispose cleanup:

1. **`FirstCtrlC_WhenCancelCallbackThrows_LogsWarningAndSuppressesException`** — registers a throwing callback on the token so `CancellationTokenSource.Cancel()` aggregates into an `AggregateException`; asserts the warning is logged via a new `RecordingLogger`, no exit, and no propagation.
2. **`ConcurrentFirstCtrlC_OnlyTransitionsStateOnce_AndAllPressesCombinedNeverExitMoreThanOnce`** — fires 16 concurrent `CancelKeyPress` via `Parallel.For`, counts how many times the CTS callback actually runs (must be 1) and how many times `Exit()` is called (must be 1).
3. **`Constructor_WithNullConsole_DoesNotCrash_AndCancelStillWorks`** — verifies the `console: null` constructor path works and `Cancel()` cooperatively cancels without ever invoking `Exit()`.
4. (Bonus, from the Copilot bot comment above) **`Dispose_UnsubscribesCancelKeyPressHandler`** — verifies `Dispose()` unsubscribes the handler so a post-dispose Ctrl+C can't poke a disposed CTS.

All 8 tests pass on net8.0, net9.0, and net462.

src/Platform/Microsoft.Testing.Platform/Services/CTRLPlusCCancellationTokenSource.cs
Evangelink: Fixed in 359b39e. You spotted the right root cause but I went with a slightly different remedy than the suggested patch — that patch had the same logical flaw it was trying to fix: making `Cancel()` transition `_state` to `StateCancelling` means an external `Cancel()` would *still* turn the first user Ctrl+C into a force-exit (the first press would see `_state != StateIdle` and fall through to the force-exit branch).

The fix in the latest commit decouples Ctrl+C counting from `_state` and from `_cancellationTokenSource.IsCancellationRequested` entirely:

- `Cancel()` only flips the CTS, never touches `_state`.
- `OnConsoleCancelKeyPressed` uses `_state` as the single source of truth for **user** Ctrl+C presses:
  - 1st press: `CompareExchange(StateIdle → StateCancelling)`, calls `Cancel()` (no-op if already canceled).
  - 2nd press: `CompareExchange(StateCancelling → StateForcing)`, calls `Exit()`.
  - 3rd+: both CAS fail → suppressed.

Net effect: the user always has to press Ctrl+C **twice** to force-exit, regardless of whether external cancellation already fired. The renamed test `CtrlC_AfterExternalCancel_DoesNotForceExit_ButSecondCtrlCDoes` now asserts exactly that.

PR #8584, opened 2026-05-25
Improve diagnostic logging across MTP crash and IPC paths

Evangelink:
## Summary

Surgical improvements to internal logging across high-risk crash and diagnostic paths in `Microsoft.Testing.Platform`. The goal is to make it noticeably easier to diagnose problems and crashes from logs alone, without changing any public API.

## Changes

| File | Improvement |
|------|-------------|
| `Helpers/UnhandledExceptionHandler.cs` | Pattern-match `e.ExceptionObject is Exception`, route the typed exception through `Log(LogLevel.Critical, msg, ex, Formatter)` (structured sinks now get the object, not just a stringified message), and pass it into `FailFast(msg, ex)`. Console output kept full-fidelity for backward-compat with acceptance tests. |
| `Hosts/ServerTestHost.cs` | Same exception-routing fix for `OnCurrentDomainUnhandledException` / `OnTaskSchedulerUnobservedTaskException`. |
| `IPC/NamedPipeClient.cs` | On IPC EOF the client called `_environment.Exit(GenericFailure)` with **zero diagnostic output**. Now writes a clear `Console.Error.WriteLineAsync` diagnostic naming the pipe and the most likely cause before exiting. Wrapped in try/catch so logging failures cannot shadow the original issue. `[Embedded]` type cannot take an `ILogger` without breaking `MTP_MSBUILD_TASKS` source-embedded compilation, so stderr is the pragmatic compromise. |
| `IPC/NamedPipeServer.cs` | Debug log on graceful client disconnect (was a silent `return`). Error log with pipe name, `WasConnected` and `LoopTaskStatus` before throwing on `Dispose` / `DisposeAsync` timeout. |
| `Configurations/JsonConfigurationProvider.cs` | Added null-safe `LogDebugAsync` / `LogErrorAsync` helpers. Logs: previously swallowed `Path.GetFullPath` failure; missing default config file (was silent return); `JsonConfigurationFileParser.Parse` failures with file path before rethrowing. |
| `Configurations/ConfigurationManager.cs` | Trace-level full-config dump wrapped in try/catch so I/O failures cannot break configuration loading. |
| `Hosts/TestHostControllersTestHost.cs` | Debug log when PID lookup `InvalidOperationException` is swallowed. Warning log on non-graceful exit including OS exit code, IPC-reported exit code (or `<not received>`), whether `TestHostCompletedRequest` was received, PID, and `CancellationRequested`. |
| `Requests/TestHostTestFrameworkInvoker.cs` | Session UID logged at Debug. Session warnings / failures previously only flowed to the output device; now also logged at Warning / Error with phase (`CreateTestSession` / `CloseTestSession`) and session UID. |

## Deliberately out of scope

- **`Microsoft.Testing.Extensions.HotReload`** has zero logging today &mdash; sizable instrumentation effort, deferred.
- **`ServerMode/JsonRpc/TcpMessageHandler.cs` / `MessageHandlerFactory.cs`** &mdash; connection loss, malformed headers and deserialization failures not logged. Substantial server-mode surface change, deferred.
- **`NamedPipeClient`** cannot accept an `ILogger` without breaking the `[Embedded]` / `MTP_MSBUILD_TASKS` source-embedded constraint.

## Validation

- `.\build.cmd -projects Microsoft.Testing.Platform.csproj` &mdash; green for `net8.0`, `net9.0`, `netstandard2.0`.
- `dotnet test Microsoft.Testing.Platform.UnitTests --filter "FullyQualifiedName~Configuration|FullyQualifiedName~UnhandledException|FullyQualifiedName~NamedPipe|FullyQualifiedName~ServerTestHost|FullyQualifiedName~TestHostControllers|FullyQualifiedName~TestHostTestFrameworkInvoker"` &mdash; 75/75 pass.
- `dotnet test Microsoft.Testing.Extensions.UnitTests` &mdash; 334 pass, 4 skipped (platform-specific).
- Existing acceptance tests (`UnhandledExceptionPolicyTests.cs`) only assert message prefixes; verified the new structure still matches.

## Notes

- No public API changes. All edits are internal to existing types.
- No `.resx` changes &mdash; MTP log messages are not localized.

### DIFF
--- a/src/Platform/Microsoft.Testing.Platform/Configurations/ConfigurationManager.cs
+++ b/src/Platform/Microsoft.Testing.Platform/Configurations/ConfigurationManager.cs
@@ -63,9 +63,18 @@ internal sealed class ConfigurationManager(IFileSystem fileSystem, ITestApplicat
             ILogger logger = syncFileLoggerProvider.CreateLogger(nameof(ConfigurationManager));
             if (logger.IsEnabled(LogLevel.Trace))
             {
-                using IFileStream configFileStream = _fileSystem.NewFileStream(defaultJsonConfiguration.ConfigurationFile, FileMode.Open, FileAccess.Read);
-                StreamReader streamReader = new(configFileStream.Stream);
-                await logger.LogTraceAsync($"Configuration file ('{defaultJsonConfiguration.ConfigurationFile}') content:\n{await streamReader.ReadToEndAsync().ConfigureAwait(false)}").ConfigureAwait(false);
+                try
+                {
+                    using IFileStream configFileStream = _fileSystem.NewFileStream(defaultJsonConfiguration.ConfigurationFile, FileMode.Open, FileAccess.Read);
+                    StreamReader streamReader = new(configFileStream.Stream);
+                    await logger.LogTraceAsync($"Configuration file ('{defaultJsonConfiguration.ConfigurationFile}') content:\n{await streamReader.ReadToEndAsync().ConfigureAwait(false)}").ConfigureAwait(false);
+                }
+                catch (Exception ex)
+                {
+                    // The trace-level dump is a diagnostic aid only; never let an I/O failure here
+                    // break the configuration pipeline.
+                    await logger.LogTraceAsync($"Failed to dump configuration file ('{defaultJsonConfiguration.ConfigurationFile}') content: {ex.GetType().FullName}: {ex.Message}").ConfigureAwait(false);
+                }
             }
         }
 
--- a/src/Platform/Microsoft.Testing.Platform/Hosts/TestHostControllersTestHost.cs
+++ b/src/Platform/Microsoft.Testing.Platform/Hosts/TestHostControllersTestHost.cs
@@ -251,10 +251,11 @@ internal sealed class TestHostControllersTestHost : CommonHost, IHost, IDisposab
             {
                 testHostProcessId = testHostProcess.Id;
             }
-            catch (InvalidOperationException) when (testHostProcess.HasExited)
+            catch (InvalidOperationException ex) when (testHostProcess.HasExited)
             {
                 // Access PID can throw InvalidOperationException if the process has already exited:
                 // System.InvalidOperationException: No process is associated with this object.
+                await _logger.LogDebugAsync($"Unable to obtain test host PID; process had already exited (ExitCode: {testHostProcess.ExitCode}). {ex.GetType().FullName}: {ex.Message}").ConfigureAwait(false);
             }
 
             testHostProcess.Exited += (_, _) =>
@@ -355,6 +356,14 @@ internal sealed class TestHostControllersTestHost : CommonHost, IHost, IDisposab
             else if (!testHostProcessInformation.HasExitedGracefully ||
                 _testHostExitCodeReceived != testHostProcess.ExitCode)
             {
+                await _logger.LogWarningAsync(
+                    $"Test host did not exit gracefully. " +
+                    $"OS exit code: '{testHostProcess.ExitCode}', " +
+                    $"IPC-reported exit code: '{(_testHostExitCodeReceived.HasValue ? _testHostExitCodeReceived.Value.ToString(CultureInfo.InvariantCulture) : "<not received>")}', " +
+                    $"TestHostCompletedRequest received: '{_testHostCompletedReceived}', " +
+                    $"PID: '{(_testHostPID.HasValue ? _testHostPID.Value.ToString(CultureInfo.InvariantCulture) : "<unknown>")}', " +
+                    $"CancellationRequested: '{cancellationToken.IsCancellationRequested}'.")
+                    .ConfigureAwait(false);
                 await outputDevice.DisplayAsync(this, new ErrorMessageOutputDeviceData(string.Format(CultureInfo.InvariantCulture, PlatformResources.TestProcessDidNotExitGracefullyErrorMessage, testHostProcess.ExitCode)), cancellationToken).ConfigureAwait(false);
                 exitCode = (int)ExitCode.TestHostProcessExitedNonGracefully;
             }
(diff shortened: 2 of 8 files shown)


src/Platform/Microsoft.Testing.Platform/Hosts/TestHostControllersTestHost.cs
```diff
@@ -355,6 +356,14 @@ protected override async Task<int> InternalRunAsync(CancellationToken cancellati
             else if (!testHostProcessInformation.HasExitedGracefully ||
                 _testHostExitCodeReceived != testHostProcess.ExitCode)
             {
+                await _logger.LogWarningAsync(
+                    $"Test host did not exit gracefully. " +
+                    $"OS exit code: '{testHostProcess.ExitCode}', " +
+                    $"IPC-reported exit code: '{(_testHostExitCodeReceived.HasValue ? _testHostExitCodeReceived.Value.ToString(CultureInfo.InvariantCulture) : "<not received>")}', " +
+                    $"TestHostCompletedRequest received: '{_testHostCompletedReceived}', " +
+                    $"PID: '{(_testHostPID.HasValue ? _testHostPID.Value.ToString(CultureInfo.InvariantCulture) : "<unknown>")}', " +
```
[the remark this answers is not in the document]
Evangelink: Fixed in c77526745 by dropping the redundant ternary and using _testHostPID.Value.ToString(CultureInfo.InvariantCulture) directly.

src/Platform/Microsoft.Testing.Platform/IPC/NamedPipeClient.cs
```diff
@@ -181,6 +181,19 @@ public async Task<TResponse> RequestReplyAsync<TRequest, TResponse>(TRequest req
...
+                    // We deliberately use Console.Error (and not stdout) to avoid corrupting any machine-readable output
+                    // that may be flowing through stdout.
+                    try
+                    {
+                        await Console.Error.WriteLineAsync($"[NamedPipeClient] Pipe '{PipeName}' was closed by the server before a response was received. The peer process likely exited or was killed. Terminating with exit code {(int)ExitCode.GenericFailure}.").ConfigureAwait(false);
+                    }
+                    catch
+                    {
+                        // Best-effort diagnostic only; never let logging failures shadow the original problem.
+                    }
```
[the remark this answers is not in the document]
Evangelink: Fixed in c77526745 by replacing the bare catch with a typed/filtered catch covering IOException, ObjectDisposedException, InvalidOperationException, NotSupportedException, ArgumentException, and OperationCanceledException.

src/Platform/Microsoft.Testing.Platform/Hosts/TestHostControllersTestHost.cs
```diff
@@ -355,6 +356,14 @@ protected override async Task<int> InternalRunAsync(CancellationToken cancellati
             else if (!testHostProcessInformation.HasExitedGracefully ||
                 _testHostExitCodeReceived != testHostProcess.ExitCode)
             {
+                await _logger.LogWarningAsync(
```
Evangelink, line 359: Maybe that's easier to read as multi-line log entry?

src/Platform/Microsoft.Testing.Platform/Configurations/ConfigurationManager.cs
```diff
@@ -63,9 +63,18 @@ internal async Task<IConfiguration> BuildAsync(IFileLoggerProvider? syncFileLogg
...
+                {
+                    using IFileStream configFileStream = _fileSystem.NewFileStream(defaultJsonConfiguration.ConfigurationFile, FileMode.Open, FileAccess.Read);
+                    StreamReader streamReader = new(configFileStream.Stream);
+                    await logger.LogTraceAsync($"Configuration file ('{defaultJsonConfiguration.ConfigurationFile}') content:\n{await streamReader.ReadToEndAsync().ConfigureAwait(false)}").ConfigureAwait(false);
+                }
+                catch (Exception ex)
+                {
+                    // The trace-level dump is a diagnostic aid only; never let an I/O failure here
+                    // break the configuration pipeline.
+                    await logger.LogTraceAsync($"Failed to dump configuration file ('{defaultJsonConfiguration.ConfigurationFile}') content: {ex.GetType().FullName}: {ex.Message}").ConfigureAwait(false);
```
Evangelink, line 76: Do we really want to swallow the exception here?

src/Platform/Microsoft.Testing.Platform/Hosts/TestHostControllersTestHost.cs
Evangelink: Done in d97372b32 — switched to a `$"""..."""` raw interpolated string so each diagnostic field (OS exit code, IPC-reported exit code, TestHostCompletedRequest received, PID, CancellationRequested) is on its own line. Much easier to scan in a log file.

src/Platform/Microsoft.Testing.Platform/Configurations/ConfigurationManager.cs
Evangelink: This isn't really a swallow — the catch logs the failure at Trace with the exception type, message, and file path, so anyone running with trace logging still sees what failed.

A few reasons I think the `try`/`catch` is justified here:

1. The config has already been successfully read and parsed on line 40 (`configurationProvider.LoadAsync()`); `defaultJsonConfiguration` is in memory and we're about to return the `AggregatedConfiguration`. The reopen on line 66 is purely a `LogLevel.Trace`-gated raw-bytes dump for diagnostics.
2. A transient I/O hiccup between line 40 and the trace dump (antivirus rescan, network share blip, file rotation) would otherwise tear down `BuildAsync` and kill the whole test session — just because deep-diagnostic logging couldn't reopen a file we already read successfully. Textbook "logging must not break the host".
3. It mirrors the defensive pattern applied in `NamedPipeClient.cs` in this same PR ("logging failures cannot shadow the original issue").

If you'd prefer narrower exception filtering I'm happy to add `when (ex is IOException or UnauthorizedAccessException)`, but for diagnostic-only code I think the broad catch is defensible. Resolving as-is unless you'd like the narrower filter.

nohwnd (approved):

Issue #6914, opened 2025-11-07
Verify compatibility of retry extension with `--maximum-failed-tests`


Youssef1313:
Need to make sure what the behavior is when both are used together, and whether or not we are broken here.

PR #8651, opened 2026-05-28
Add acceptance test for retry + --maximum-failed-tests interaction (#6914)

Evangelink:
Closes part of #6914.

This PR adds acceptance test coverage for the interaction between the retry extension (`--retry-failed-tests`) and the platform's `--maximum-failed-tests` option.

## What the test pins down

When both flags are passed and `--maximum-failed-tests` fires inside the child test host:

1. The child exits with `ExitCode.TestExecutionStoppedForMaxFailedTests` (13). This is because `TestApplicationResult.GetProcessExitCode()` checks `IsMaxFailedTestsTriggered` before `IsTestAdapterTestSessionFailure`, so the 13 override wins over the usual `AtLeastOneTestFailed` (2) override.
2. The retry orchestrator sees an exit code that is neither `Success` nor `AtLeastOneTestFailed` and aborts the retry loop — it does **not** launch a second attempt. This is the correct functional behavior: when the user explicitly asked the run to stop early, retrying would defeat the point.
3. The retry orchestrator emits the generic `Test suite failed with and exit code different that 2 (failed tests). Failure related to an unexpected condition. Exit code '13'` warning. This is misleading because the stop was intentional — but rather than silently fixing it here, the test asserts the current text with a `TODO(#6914)` comment so a follow-up UX improvement is forced to update the assertion intentionally.

## Follow-up

A separate PR could teach `RetryOrchestrator` to recognize exit code 13 and either suppress the warning or emit a max-failed-tests-aware message. Keeping that out of scope here so this change is purely test coverage.

### DIFF
--- a/test/IntegrationTests/Microsoft.Testing.Platform.Acceptance.IntegrationTests/RetryWithMaxFailedTestsTests.cs
+++ b/test/IntegrationTests/Microsoft.Testing.Platform.Acceptance.IntegrationTests/RetryWithMaxFailedTestsTests.cs
@@ -1,7 +1,5 @@
-﻿#pragma warning disable IDE0073 // The file header does not match the required text
-// Copyright (c) Microsoft Corporation. All rights reserved.
-// Licensed under dual-license. See LICENSE.PLATFORMTOOLS.txt file in the project root for full license information.
-#pragma warning restore IDE0073 // The file header does not match the required text
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
 namespace Microsoft.Testing.Platform.Acceptance.IntegrationTests;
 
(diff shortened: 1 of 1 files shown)


test/IntegrationTests/Microsoft.Testing.Platform.Acceptance.IntegrationTests/RetryWithMaxFailedTestsTests.cs
```diff
@@ -0,0 +1,218 @@
+﻿#pragma warning disable IDE0073 // The file header does not match the required text
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under dual-license. See LICENSE.PLATFORMTOOLS.txt file in the project root for full license information.
+#pragma warning restore IDE0073 // The file header does not match the required text
```
Evangelink:
```suggestion
﻿// Copyright (c) Microsoft Corporation. All rights reserved.
// Licensed under the MIT license. See LICENSE file in the project root for full license information.
```

PR #8656, opened 2026-05-28
Skip MSTEST0065 when expected/actual is null literal

### LINKED ISSUE #8655: MSTEST0065 false positive on Assert.AreEqual(null, collection) — MSTEST0037 already covers this case
### Description

[MSTEST0065](https://learn.microsoft.com/dotnet/core/testing/mstest-analyzers/mstest0065) (`AvoidAssertAreEqualOnCollections`) reports a warning when the call is actually a null check against a collection, for example:

```csharp
Dictionary<string, int>? dictionary = GetDictionary();
Assert.AreEqual(null, dictionary);
```

The analyzer fires because the generic type argument inferred for `Assert.AreEqual<T>` is `Dictionary<string, int>` (a type implementing `IEnumerable<T>`), and suggests switching to `CollectionAssert.AreEqual`. But `CollectionAssert.AreEqual` is not what the user wants here — they are asserting that the reference is `null`.

### Why it's a false positive

In this scenario, [MSTEST0037](https://learn.microsoft.com/dotnet/core/testing/mstest-analyzers/mstest0037) (`UseProperAssertMethods`) already triggers and correctly suggests rewriting the call as:

```csharp
Assert.IsNull(dictionary);
```

That is the right fix. MSTEST0065 is noise here (and its suggested replacement, `CollectionAssert.AreEqual(null, dictionary)`, would not improve the code).

### Expected behavior

MSTEST0065 should not trigger when the `expected` (or `actual`) argument is the `null` literal — MSTEST0037 handles that case and produces a better suggestion.

### Repro

```csharp
[TestMethod]
public void Test()
{
    Dictionary<string, int>? dictionary = null;
    Assert.AreEqual(null, dictionary); // MSTEST0065 fires (false positive); MSTEST0037 also fires (correct)
}
```

### Analyzer source

[`AvoidAssertAreEqualOnCollectionsAnalyzer.cs`](https://github.com/microsoft/testfx/blob/main/src/Analyzers/MSTest.Analyzers/AvoidAssertAreEqualOnCollectionsAnalyzer.cs) — the `AnalyzeInvocation` method does not check whether either argument is a `null` literal before reporting.

Evangelink:
Fixes #8655.

### Problem

[MSTEST0065](https://learn.microsoft.com/dotnet/core/testing/mstest-analyzers/mstest0065) (`AvoidAssertAreEqualOnCollections`) fires on calls like:

```csharp
Dictionary<string, int>? dictionary = ...;
Assert.AreEqual(null, dictionary);
```

because the generic type argument inferred for `Assert.AreEqual<T>` is a collection type. The suggested fix — switch to `CollectionAssert.AreEqual` / `Assert.AreSequenceEqual` — is not what the user wants: they are performing a null check, not a collection comparison.

[MSTEST0037](https://learn.microsoft.com/dotnet/core/testing/mstest-analyzers/mstest0037) (`UseProperAssertMethods`) already triggers on the same call and proposes the correct `Assert.IsNull(dictionary)` (or `Assert.IsNotNull` for `AreNotEqual`) replacement. MSTEST0065 reporting it is redundant and misleading.

### Fix

In `AvoidAssertAreEqualOnCollectionsAnalyzer`, skip reporting when either the `expected` / `notExpected` or `actual` argument is the `null` literal. This is detected via `ILiteralOperation { ConstantValue: { HasValue: true, Value: null } }` after walking down conversions, matching the pattern used by other MSTest analyzers (see `UseProperAssertMethodsAnalyzer.NullChecks.cs`).

### Tests

- Replaced the existing `WhenUsingAssertAreEqualOnNullableListNulls_ReportDiagnostic` test (which asserted on the now-suppressed case) with a `...DoNotReportDiagnostic` version explaining the rationale.
- Added regression tests covering `Assert.AreEqual(null, dictionary)`, `Assert.AreEqual(dictionary, null)`, and `Assert.AreNotEqual(null, dictionary)`.

`dotnet test` for `MSTest.Analyzers.UnitTests` passes (1091/1091 tests).

### DIFF
--- a/src/Analyzers/MSTest.Analyzers/AvoidAssertAreEqualOnCollectionsAnalyzer.cs
+++ b/src/Analyzers/MSTest.Analyzers/AvoidAssertAreEqualOnCollectionsAnalyzer.cs
@@ -10,6 +10,7 @@ using Microsoft.CodeAnalysis.Diagnostics;
 using Microsoft.CodeAnalysis.Operations;
 
 using MSTest.Analyzers.Helpers;
+using MSTest.Analyzers.RoslynAnalyzerHelpers;
 
 namespace MSTest.Analyzers;
 
@@ -74,6 +75,17 @@ public sealed class AvoidAssertAreEqualOnCollectionsAnalyzer : DiagnosticAnalyze
             return;
         }
 
+        // When either argument is the null literal, MSTEST0037 (UseProperAssertMethods) already triggers
+        // and suggests the more meaningful Assert.IsNull / Assert.IsNotNull replacement. Reporting
+        // MSTEST0065 in that case is a false positive because CollectionAssert.AreEqual / Assert.AreSequenceEqual
+        // are not the user's intent — they are performing a null check.
+        // The first parameter is "expected" on Assert.AreEqual and "notExpected" on Assert.AreNotEqual.
+        string firstParameterName = targetMethod.Name == "AreEqual" ? "expected" : "notExpected";
+        if (HasNullLiteralArgument(invocation, firstParameterName) || HasNullLiteralArgument(invocation, "actual"))
+        {
+            return;
+        }
+
         string methodName = $"Assert.{targetMethod.Name}";
         string comparedTypeDisplay = comparedType.ToDisplayString(SymbolDisplayFormat.MinimallyQualifiedFormat);
         context.ReportDiagnostic(invocation.CreateDiagnostic(Rule, methodName, comparedTypeDisplay));
@@ -82,6 +94,12 @@ public sealed class AvoidAssertAreEqualOnCollectionsAnalyzer : DiagnosticAnalyze
     private static bool ShouldReport(ITypeSymbol comparedType, INamedTypeSymbol genericEnumerableSymbol)
         => comparedType.SpecialType != SpecialType.System_String
             && ImplementsGenericEnumerable(comparedType, genericEnumerableSymbol);
+
+    private static bool HasNullLiteralArgument(IInvocationOperation invocation, string parameterName)
+    {
+        IArgumentOperation? argument = invocation.Arguments.FirstOrDefault(arg => arg.Parameter?.Name == parameterName);
+        return argument?.Value.WalkDownConversion() is ILiteralOperation { ConstantValue: { HasValue: true, Value: null } };
+    }
 
     private static bool ImplementsGenericEnumerable(ITypeSymbol type, INamedTypeSymbol genericEnumerableSymbol)
     {
--- a/src/Analyzers/MSTest.Analyzers/AvoidAssertAreEqualOnCollectionsAnalyzer.cs
+++ b/src/Analyzers/MSTest.Analyzers/AvoidAssertAreEqualOnCollectionsAnalyzer.cs
@@ -75,10 +75,13 @@ public sealed class AvoidAssertAreEqualOnCollectionsAnalyzer : DiagnosticAnalyze
             return;
         }
 
-        // When either argument is the null literal, MSTEST0037 (UseProperAssertMethods) already triggers
-        // and suggests the more meaningful Assert.IsNull / Assert.IsNotNull replacement. Reporting
-        // MSTEST0065 in that case is a false positive because CollectionAssert.AreEqual / Assert.AreSequenceEqual
-        // are not the user's intent — they are performing a null check.
+        // When either argument is the null literal, the user is performing a null check rather than
+        // a collection equality check, so suggesting CollectionAssert.AreEqual / Assert.AreSequenceEqual
+        // would be misleading.
+        // When null is in the expected/notExpected position, MSTEST0037 (UseProperAssertMethods) already
+        // triggers and proposes the correct Assert.IsNull / Assert.IsNotNull replacement.
+        // When null is in the actual position, MSTEST0037 does not currently fire, but suppressing
+        // MSTEST0065 is still correct because CollectionAssert guidance is not what the user wants.
         // The first parameter is "expected" on Assert.AreEqual and "notExpected" on Assert.AreNotEqual.
         string firstParameterName = targetMethod.Name == "AreEqual" ? "expected" : "notExpected";
         if (HasNullLiteralArgument(invocation, firstParameterName) || HasNullLiteralArgument(invocation, "actual"))
--- a/test/UnitTests/MSTest.Analyzers.UnitTests/AvoidAssertAreEqualOnCollectionsAnalyzerTests.cs
+++ b/test/UnitTests/MSTest.Analyzers.UnitTests/AvoidAssertAreEqualOnCollectionsAnalyzerTests.cs
@@ -322,6 +322,31 @@ public sealed class AvoidAssertAreEqualOnCollectionsAnalyzerTests
     }
 
     [TestMethod]
+    public async Task WhenUsingAssertAreNotEqualWithCollectionExpectedAndNullActual_DoNotReportDiagnostic()
+    {
+        // Same rationale for Assert.AreNotEqual(x, null): the user is performing a null check, not a
+        // collection comparison, so suggesting CollectionAssert.AreNotEqual would be misleading.
+        string code = """
+            #nullable enable
+            using System.Collections.Generic;
+            using Microsoft.VisualStudio.TestTools.UnitTesting;
+
+            [TestClass]
+            public class MyTestClass
+            {
+                [TestMethod]
+                public void MyTestMethod()
+                {
+                    Dictionary<string, int>? dictionary = new();
+                    Assert.AreNotEqual(dictionary, null);
+                }
+            }
+            """;
+
+        await VerifyCS.VerifyAnalyzerAsync(code);
+    }
+
+    [TestMethod]
     public async Task CodeFix_Ordered_ForAreEqual()
     {
         string code = """
--- a/test/UnitTests/MSTest.Analyzers.UnitTests/AvoidAssertAreEqualOnCollectionsAnalyzerTests.cs
+++ b/test/UnitTests/MSTest.Analyzers.UnitTests/AvoidAssertAreEqualOnCollectionsAnalyzerTests.cs
@@ -222,8 +222,12 @@ public sealed class AvoidAssertAreEqualOnCollectionsAnalyzerTests
     }
 
     [TestMethod]
-    public async Task WhenUsingAssertAreEqualOnNullableListNulls_ReportDiagnostic()
+    public async Task WhenUsingAssertAreEqualOnNullableListNulls_DoNotReportDiagnostic()
     {
+        // When an argument is the null literal, MSTEST0037 (UseProperAssertMethods) already
+        // triggers and suggests Assert.IsNull / Assert.IsNotNull, which is the correct fix.
+        // MSTEST0065 must stay quiet to avoid a redundant (and misleading) suggestion to switch
+        // to CollectionAssert.AreEqual / Assert.AreSequenceEqual.
         string code = """
             #nullable enable
             using System.Collections.Generic;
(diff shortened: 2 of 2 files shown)


test/UnitTests/MSTest.Analyzers.UnitTests/AvoidAssertAreEqualOnCollectionsAnalyzerTests.cs
```diff
@@ -235,12 +239,86 @@ public class MyTestClass
...
+            using System.Collections.Generic;
+            using Microsoft.VisualStudio.TestTools.UnitTesting;
+
+            [TestClass]
+            public class MyTestClass
+            {
+                [TestMethod]
+                public void MyTestMethod()
+                {
+                    Dictionary<string, int>? dictionary = new();
```
Evangelink, line 315: **[MODERATE] Test Completeness — Missing symmetric test for `Assert.AreNotEqual(collection, null)`**

Three null-position combinations are tested:
- `AreEqual(null, collection)` ✅
- `AreEqual(collection, null)` ✅
- `AreNotEqual(null, collection)` ✅

The fourth symmetric combination is missing:
- `AreNotEqual(collection, null)` ❌

The production code does check `HasNullLiteralArgument(invocation, "actual")` which covers this case, but without an explicit test the behaviour for this variant is not verified by the test suite.

**Recommendation:** Add a `WhenUsingAssertAreNotEqualWithCollectionExpectedAndNullActual_DoNotReportDiagnostic` test mirroring the `AreEqual` variant.

src/Analyzers/MSTest.Analyzers/AvoidAssertAreEqualOnCollectionsAnalyzer.cs
```diff
@@ -74,6 +75,17 @@ targetMethod.Name is not ("AreEqual" or "AreNotEqual") ||
             return;
         }
 
+        // When either argument is the null literal, MSTEST0037 (UseProperAssertMethods) already triggers
+        // and suggests the more meaningful Assert.IsNull / Assert.IsNotNull replacement. Reporting
+        // MSTEST0065 in that case is a false positive because CollectionAssert.AreEqual / Assert.AreSequenceEqual
+        // are not the user's intent — they are performing a null check.
+        // The first parameter is "expected" on Assert.AreEqual and "notExpected" on Assert.AreNotEqual.
```
Evangelink, line 85: **[MODERATE] Algorithmic Correctness — Comment is inaccurate for the `actual`-null case**

The comment claims _"MSTEST0037 (UseProperAssertMethods) already triggers"_ when **either** argument is the null literal. That is only partially true.

`UseProperAssertMethodsAnalyzer` only inspects the **first** (`expected` / `notExpected`) argument for a null literal (line ~763 of `UseProperAssertMethodsAnalyzer.cs`):
```csharp
else if (expectedArgument is ILiteralOperation { ConstantValue: { HasValue: true, Value: null } })
```
It has no corresponding check for the `actual` parameter. So for `Assert.AreEqual(dictionary, null)`, MSTEST0037 **does not** fire.

The suppression is still semantically correct — suggesting `CollectionAssert.AreEqual(dictionary, null)` would be the wrong fix for a null check — but the stated justification is wrong for the `actual = null` path, which could mislead future maintainers into thinking there is coverage that doesn't exist.

**Recommendation:** Update the comment to distinguish the two cases, e.g.:
```csharp
// When the expected/notExpected argument is null, MSTEST0037 already triggers and suggests
// Assert.IsNull / Assert.IsNotNull. When the actual argument is null, MSTEST0037 does not
// currently fire, but reporting MSTEST0065 would still be wrong — CollectionAssert.AreEqual
// is not a meaningful replacement for a null check on the actual value.
```

src/Analyzers/MSTest.Analyzers/AvoidAssertAreEqualOnCollectionsAnalyzer.cs
```diff
@@ -83,6 +95,12 @@ private static bool ShouldReport(ITypeSymbol comparedType, INamedTypeSymbol gene
         => comparedType.SpecialType != SpecialType.System_String
             && ImplementsGenericEnumerable(comparedType, genericEnumerableSymbol);
 
+    private static bool HasNullLiteralArgument(IInvocationOperation invocation, string parameterName)
+    {
+        IArgumentOperation? argument = invocation.Arguments.FirstOrDefault(arg => arg.Parameter?.Name == parameterName);
+        return argument?.Value.WalkDownConversion() is ILiteralOperation { ConstantValue: { HasValue: true, Value: null } };
```
Evangelink, line 104: **[NIT] Analyzer Quality — `WalkDownConversion` vs direct `ILiteralOperation` check inconsistency**

`HasNullLiteralArgument` unwraps implicit conversions before matching the null literal:
```csharp
argument?.Value.WalkDownConversion() is ILiteralOperation { ConstantValue: { HasValue: true, Value: null } }
```

`UseProperAssertMethodsAnalyzer` (MSTEST0037), which this suppression defers to, checks the raw operation **without** walking down conversions:
```csharp
expectedArgument is ILiteralOperation { ConstantValue: { HasValue: true, Value: null } }
```

This means for `Assert.AreEqual((List<int>?)null, collection)` (explicit cast-to-null):
- MSTEST0065 suppresses (it walks through the cast and sees `ILiteralOperation`)
- MSTEST0037 does **not** fire (it sees `IConversionOperation`, not `ILiteralOperation`)

The net result is that both analyzers are silent. The suppression is still defensible (MSTEST0065 suggesting `CollectionAssert.AreEqual` for an explicit cast-null is wrong), but the asymmetry is worth noting. If MSTEST0037 is ever fixed to handle this via `WalkDownConversion`, the suppression here will work correctly already.

src/Analyzers/MSTest.Analyzers/AvoidAssertAreEqualOnCollectionsAnalyzer.cs
```diff
@@ -74,6 +75,17 @@ targetMethod.Name is not ("AreEqual" or "AreNotEqual") ||
             return;
         }
 
+        // When either argument is the null literal, MSTEST0037 (UseProperAssertMethods) already triggers
```
Evangelink: **Dimension 1 — Behavioral correctness / Comment accuracy**

The comment claims "MSTEST0037 already triggers" when **either** argument is the null literal, but that is only true for the `expected`/`notExpected` position. `UseProperAssertMethodsAnalyzer.AnalyzeAreEqualOrAreNotEqualInvocation` checks `expectedArgument is ILiteralOperation { ConstantValue: { HasValue: true, Value: null } }` — it does **not** inspect the `actual` argument for null.

For `Assert.AreEqual(collection, null)` / `Assert.AreNotEqual(collection, null)`, MSTEST0037 will **not** fire, so the suppression of MSTEST0065 leaves the user with no diagnostic at all (neither the redundant 0065 nor the correct 0037). This does not regress the original issue (MSTEST0065 suggesting `CollectionAssert.AreEqual` is still wrong), but the comment misrepresents the full behaviour.

Suggested correction:

```suggestion
        // When either argument is the null literal, suggesting CollectionAssert.AreEqual /
        // Assert.AreSequenceEqual would be misleading — the user is performing a null check,
        // not a collection equality check.
        // When null is in the expected/notExpected position, MSTEST0037 (UseProperAssertMethods)
        // already triggers and proposes the correct Assert.IsNull / Assert.IsNotNull replacement.
        // When null is in the actual position, no other diagnostic currently covers it, but
        // MSTEST0065 suppression is still correct because CollectionAssert guidance is wrong.
        // The first parameter is "expected" on Assert.AreEqual and "notExpected" on Assert.AreNotEqual.
```

test/UnitTests/MSTest.Analyzers.UnitTests/AvoidAssertAreEqualOnCollectionsAnalyzerTests.cs
```diff
@@ -235,12 +239,86 @@ public class MyTestClass
...
+                    Assert.AreNotEqual(null, dictionary);
+                }
+            }
+            """;
+
+        await VerifyCS.VerifyAnalyzerAsync(code);
```
Evangelink, line 321: **Dimension 13 — Test coverage**

The null-position matrix is incomplete. The PR adds:
- ✅ `AreEqual(null, collection)` — null in expected
- ✅ `AreEqual(collection, null)` — null in actual
- ✅ `AreNotEqual(null, collection)` — null in notExpected

Missing:
- ❌ `AreNotEqual(collection, null)` — null in actual

The production guard `HasNullLiteralArgument(invocation, "actual")` applies to both `AreEqual` and `AreNotEqual`, so the symmetric `AreNotEqual` case should be tested to prevent regression. Please add a `WhenUsingAssertAreNotEqualWithCollectionExpectedAndNullActual_DoNotReportDiagnostic` test analogous to the existing `WhenUsingAssertAreEqualWithCollectionExpectedAndNullActual_DoNotReportDiagnostic`.

Evangelink (commented):
## Review Summary

The fix is correct and well-motivated. The `HasNullLiteralArgument` helper is implemented consistently with the rest of the codebase (`WalkDownConversion` pattern is used identically in `UseProperAssertMethodsAnalyzer.NullChecks.cs`), the parameter name lookup is accurate for both `AreEqual` (`expected`) and `AreNotEqual` (`notExpected`), and the new tests add good coverage for the regression.

Three findings:

| # | Dimension | Verdict |
|---|-----------|---------|
| 1 | Algorithmic Correctness | 🟡 1 MODERATE — comment inaccurate for `actual = null` |
| 13 | Test Completeness | 🟡 1 MODERATE — missing `AreNotEqual(collection, null)` test |
| 18 | Analyzer Quality | 🔵 1 NIT — WalkDownConversion vs. direct check inconsistency vs. MSTEST0037 |

✅ 18/21 dimensions clean (3 remaining N/A for this change scope).

**Key finding:** The production-code comment states _"MSTEST0037 already triggers"_ when **either** argument is null, but MSTEST0037 only checks the `expected`/`notExpected` parameter — it has no coverage for `actual = null`. The suppression for `actual = null` is still semantically correct (MSTEST0065 suggesting `CollectionAssert.AreEqual(collection, null)` would be wrong guidance), but the comment misrepresents the situation and could mislead future maintainers. See inline comment on line 82 for a suggested replacement.

The missing `AreNotEqual(collection, null)` test is straightforward to add and would complete the 2×2 test matrix.

> Generated by [Expert Code Review (on open)](https://github.com/microsoft/testfx/actions/runs/26573795697) for issue #8656 · sonnet46 3.3M

src/Analyzers/MSTest.Analyzers/AvoidAssertAreEqualOnCollectionsAnalyzer.cs
Evangelink: Addressed in 9395c45fa — updated the comment to clarify that MSTEST0037 only covers the `expected`/`notExpected` position, and explicitly call out that suppressing MSTEST0065 for `actual = null` is still correct because `CollectionAssert` guidance is wrong for a null check.

src/Analyzers/MSTest.Analyzers/AvoidAssertAreEqualOnCollectionsAnalyzer.cs
Evangelink: Addressed in 9395c45fa — used a slightly different wording than the suggestion but with the same intent: the updated comment distinguishes the `expected`/`notExpected` case (covered by MSTEST0037) from the `actual` case (not covered, but suppression still correct).

test/UnitTests/MSTest.Analyzers.UnitTests/AvoidAssertAreEqualOnCollectionsAnalyzerTests.cs
Evangelink: Addressed in 9395c45fa — added `WhenUsingAssertAreNotEqualWithCollectionExpectedAndNullActual_DoNotReportDiagnostic` to complete the 2x2 null-position matrix.

test/UnitTests/MSTest.Analyzers.UnitTests/AvoidAssertAreEqualOnCollectionsAnalyzerTests.cs
Evangelink: Addressed in 9395c45fa — added `WhenUsingAssertAreNotEqualWithCollectionExpectedAndNullActual_DoNotReportDiagnostic` mirroring the `AreEqual` variant.

Issue #6349, opened 2025-08-12
Unify Configuration/Options


thomhurst:
If we take inspiration from Microsoft.Extensions, they have a fluent builder syntax that you can control and order at startup.

This means I can have:

Configuration in a JSON file
Other values in environment variables
Other variables passed through as CLI flags
And potential even other sources in the cloud (but let's ignore them as feels a bit overkill for tests)

It would be nice to build this similarly into MTP. And then create an interface to easily retrieve values, without us having to know the implementation of where it comes from etc.

Currently checking various sources is a lot of manual work.

E.g.

`_testConfiguration.Get("TUnit:UseFeatureA");`

And this would iterate through all the providers similar to how ME libraries do and pull out the last populated value.

thomhurst:
Ah brill.

I think the only thing missing (as I can't see it in those files) is command line inclusion.

To me that's just another way to pass configuration.
So like it'd be good if I could set things like hangdump settings in the JSON configuration. It'd behave exactly like it just current, but you can just pull the configuration out of any of these sources rather than solely from the cli args.

thomhurst:
I'd use cli args as the highest priority, and then fallback

Evangelink:
PR #8664 (merged) addresses the core ask from this thread: CLI options can now be set in `testconfig.json` under `commandLineOptions:*`, with CLI keeping highest precedence and JSON as fallback. Hangdump-style options that previously had to be passed on the CLI can now live in the JSON config.

Two minor follow-up nits that are not part of the original ask:

1. **Scalar bool ambiguity** — for arg-bearing options, JSON `"true"` / `"false"` is ambiguous with the zero-arity presence marker. Callers must use the array form (`"foo": ["true"]`) for non-boolean string values that happen to be `"true"` / `"false"`. Nothing enforces this today.
2. **API visibility** — the new helpers are kept `internal`. We need to decide whether they should become public surface.

(Note: `--config-file` and `--diagnostic*` intentionally bypass `IConfiguration` — they're needed before the config system is built. That's by design, not a gap.)

Closing as the core unification is in.

Evangelink:
Filed follow-up: #8829 — flag bootstrap-only options (`--config-file`, `--diagnostic*`) when set in `testconfig.json` instead of silently ignoring them.

Evangelink:
[...]

PR #8664, opened 2026-05-28
Read CLI options from testconfig.json via IConfiguration (#6349)

Evangelink:
## Summary

Resolves part of #6349: let users specify CLI options inside `testconfig.json` by routing `ICommandLineOptions` through `IConfiguration`.

## Approach

1. New `CommandLineConfigurationSource` (`Order = 0`) wraps the parsed CLI into an `IConfigurationProvider` that flattens each option under `commandLineOptions:<name>` (zero-arity) or `commandLineOptions:<name>:<index>` (arg-bearing).
2. The CLI source is registered in `TestHostBuilder.CommonServices` **before** the JSON source, so CLI keeps highest precedence.
3. `AggregatedConfiguration.TryGetCommandLineOptionFromProviders` does a **provider-aware** lookup: it walks providers in registration order, and the first provider with **any** data for that option wins outright (no per-key cross-provider merging). This avoids cases where `IConfiguration[key]` accidentally returns the JSON `:0` while the CLI set the zero-arity bare key, or vice versa.
4. The public extension helpers (`IsCommandLineOptionSet`, `TryGetCommandLineOptionArguments`) delegate to that method when given an `AggregatedConfiguration`, and `CommandLineHandler` becomes a thin facade that does the same. End result: every consumer of `ICommandLineOptions` transparently sees JSON-sourced options.
5. `JsonConfigurationProvider` exposes a typed, schema-validated enumeration of `commandLineOptions:*` entries. `CommandLineOptionsValidator` runs three extra passes over those entries (unknown-option, arity, per-arg validation) so `testconfig.json` typos surface during startup instead of crashing inside option handlers.
6. `DisplayBannerIfEnabledAsync` reads `--no-banner` from the unified `ICommandLineOptions`, so banner suppression honors `testconfig.json` as well as the CLI.

## What works

- `testconfig.json` keys under `commandLineOptions` are surfaced everywhere `ICommandLineOptions` is consumed.
- CLI still wins over JSON for the same option.
- `--results-directory` defined in JSON is honored (was previously ignored - the legacy `GetResultsDirectoryCore` path was bypassing JSON).
- `--no-banner` defined in JSON suppresses the banner.
- Invalid JSON entries (unknown option, wrong arity, bad arg) fail at startup with a clear error referencing `testconfig.json`, instead of crashing later. Concrete cases now caught:
   - `--timeout: true` (was: `IndexOutOfRangeException` in `args[0]`)
   - `--exit-on-process-exit: "abc"` (was: `FormatException` in `int.Parse(args[0])`)
   - typos like `"timeoutt": "30s"` (was: silently ignored)
- Back-compat ctor on `CommandLineHandler` is kept so external code keeps compiling.

## Rough edges still on the table

These are not addressed in this PR:

1. **Bootstrap readers.** `--diagnostic*` and `--config-file` are read off `parseResult` **before** `IConfiguration` is built. Honoring them from JSON requires either a two-phase config build or rewriting those bootstrap paths.
2. **Scalar bool ambiguity.** For arg-bearing options, JSON `"true"` / `"false"` is ambiguous with the zero-arity presence marker. Callers must use the array form (`"foo": ["true"]`) for non-boolean string values that happen to be `"true"` / `"false"`. No code enforces this today.
3. **API visibility.** New helpers are kept `internal` for now. If we ship this we need to decide whether they become public surface.

## Tests

- `CommandLineConfigurationProviderTests` covers the source/provider flattening + the provider-aware invariants (including `ProviderAwareResolution_CliZeroArityShadowsJsonIndexedArgs` end-to-end).
- `CommandLineConfigurationExtensionsTests` covers the helper-level and handler-end-to-end behavior.
- `JsonCommandLineOptionsTests` (new) covers the JSON enumeration schema, the three new validator passes, disabled entries, shadowing, and case-insensitivity.
- Full UT project green on net8.0 / net9.0 / net462.


src/Platform/Microsoft.Testing.Platform/CommandLine/CommandLineHandler.cs
```diff
@@ -214,10 +229,14 @@ async Task DisplayRegisteredToolsInfoAsync(IOutputDevice outputDevice, IReadOnly
...
-    public bool TryGetOptionArgumentList(string optionName, [NotNullWhen(true)] out string[]? arguments) =>
-        ParseResult.TryGetOptionArgumentList(optionName, out arguments);
+        => _configuration is not null
+            ? _configuration.IsCommandLineOptionSet(optionName)
+            : ParseResult.IsOptionSet(optionName);
+
+    public bool TryGetOptionArgumentList(string optionName, [NotNullWhen(true)] out string[]? arguments)
+        => _configuration is not null
+            ? _configuration.TryGetCommandLineOptionArguments(optionName, out arguments)
+            : ParseResult.TryGetOptionArgumentList(optionName, out arguments);
```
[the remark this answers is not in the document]
Evangelink: Addressed in 1eaacbf59.

JsonConfigurationProvider now exposes a typed, schema-validated enumeration of `commandLineOptions:*` entries, and `CommandLineOptionsValidator` runs three extra passes over them: unknown-option, arity, and per-arg validation. Invalid JSON entries such as `timeout: true` (now caught by arity) or `exit-on-process-exit: "abc"` (now caught by the per-arg validator) fail at startup with a clear error referencing `testconfig.json`, instead of crashing inside the option handler. New `JsonCommandLineOptionsTests` covers all three passes plus disabled entries, shadowing, and case-insensitivity.

(I went with the "validate the JSON view" route rather than gating consumption, since giving the validator the same provider-aware view keeps the runtime fast path unchanged.)

test/UnitTests/Microsoft.Testing.Platform.UnitTests/Configuration/JsonCommandLineOptionsTests.cs
```diff
@@ -0,0 +1,511 @@
...
+        Assert.IsFalse(result.IsValid);
+        Assert.Contains("testconfig.json", result.ErrorMessage);
+    }
+
+    [TestMethod]
+    public async Task Validator_DuplicateOptionNamesDifferingByCase_FailsGracefully()
+    {
+        // Two extension providers register the same option with different casing. Pre-fix this
+        // crashed with a raw ArgumentException inside the OrdinalIgnoreCase ToDictionary call;
+        // it must now produce a clean ValidationResult.Invalid via the duplicate-options pass.
```
Evangelink: **[NIT] Documentation Accuracy — comment misrepresents the pre-fix behavior**

> Pre-fix this crashed with a raw `ArgumentException` inside the `OrdinalIgnoreCase ToDictionary` call

That's not quite right. Pre-fix, the `ToDictionary` call at `CommandLineOptionsValidator.cs:60-62` used the default Ordinal comparer, so `"Timeout"` and `"timeout"` were treated as **two distinct keys** — both passed silently into the dictionary with no exception. The pre-fix `ValidateOptionsAreNotDuplicated` also used the default Ordinal dictionary, so it didn't catch them either.

What this test actually proves is the *positive* behavior added by the fix: case-differing duplicates are now caught as duplicates rather than silently producing two separate entries that the rest of the platform treats as one (since lookups are case-insensitive everywhere downstream).

Suggested rewording:

```csharp
// Two extension providers register the same option with different casing. Pre-fix the
// default Ordinal comparison in ValidateOptionsAreNotDuplicated treated them as distinct
// names and both were silently accepted, even though every downstream lookup
// (TryGetCommandLineOptionFromProviders, testconfig.json) is case-insensitive. The
// OrdinalIgnoreCase fix surfaces this as a clean ValidationResult.Invalid up front.
```

Also: consider tightening the assertion. `Assert.IsFalse(result.IsValid)` accepts *any* failure, including unrelated ones a future refactor might introduce. Asserting that `result.ErrorMessage` contains both provider display names (or the duplicate-options resource substring) would lock the test to the intended code path.

src/Platform/Microsoft.Testing.Platform/CommandLine/CommandLineOptionsValidator.cs
```diff
@@ -45,21 +47,26 @@ public static async Task<ValidationResult> ValidateAsync(
...
 
+        // Option names from the platform side are unique (validated above) and lookups against this
+        // dictionary may be triggered from the CLI parse result (case-sensitive) or from JSON
+        // (testconfig.json is case-insensitive everywhere else). Use OrdinalIgnoreCase so a JSON
+        // entry such as "Timeout": "30s" resolves to the same provider/option metadata as
+        // "--timeout 30s" on the command line.
         var providerAndOptionByOptionName = extensionOptionsByProvider.Union(systemOptionsByProvider)
             .SelectMany(tuple => tuple.Value.Select(option => (provider: tuple.Key, option)))
-            .ToDictionary(tuple => tuple.option.Name);
+            .ToDictionary(tuple => tuple.option.Name, StringComparer.OrdinalIgnoreCase);
```
Evangelink, line 62: **[LOW] Algorithmic Correctness — system+system case-differing collision still crashes here**

The round-1 fix (case-insensitive dictionaries at lines 114, 124, 160) covers extension+extension and extension+system collisions cleanly, but **system+system** case-differing collisions are not caught by any pre-pass validator and still fall through to this `ToDictionary` call, which will throw `ArgumentException: An item with the same key has already been added`.

**Reproducer:** two `ICommandLineOptionsProvider`s in `systemCommandLineOptionsProviders` registering `"Timeout"` and `"timeout"`.

Pre-PR this was masked because the default Ordinal comparison treated them as distinct keys; the switch to `OrdinalIgnoreCase` here surfaces the latent collision. In practice system providers are platform-internal and tested, so this is a programmer-error scenario, but it's the one remaining ungraceful path advertised in the PR description.

**Suggested fix:** either extend `ValidateOptionsAreNotDuplicated` to also iterate `systemOptionsByProvider` (preferred for symmetry), or replace this with a `GroupBy(...).ToDictionary(g => g.Key, g => g.First())` that defers the diagnostic to the existing reserved-options pass.

Happy to defer to a follow-up if you consider system-side dedup out of scope for this PR.

test/UnitTests/Microsoft.Testing.Platform.UnitTests/Configuration/JsonCommandLineOptionsTests.cs
```diff
@@ -0,0 +1,511 @@
...
+            CommandLineParseResult.Empty,
+            systemCommandLineOptionsProviders: [],
+            extensionCommandLineOptionsProviders: [providerA, providerB],
+            new Mock<ICommandLineOptions>().Object);
+
+        Assert.IsFalse(result.IsValid);
+    }
+
+    [TestMethod]
+    public async Task Validator_JsonSparseIndexedEntry_DefensiveSchemaRejection()
```
Evangelink: **[NIT] Naming & Conventions — test name doesn't match what's tested**

`Validator_JsonSparseIndexedEntry_DefensiveSchemaRejection` reads as "the validator rejects a sparse indexed entry", but the body builds a single-valued entry (`["only-one"]`) against an `ArgumentArity(2, 2)` option and asserts the **arity** check fires. There's nothing sparse or indexed about the input — no missing index in a multi-element array.

Suggested rename: `Validator_JsonEntryWithTooFewArguments_FailsArityCheck` (or `Validator_JsonArityFailure_EmitsJsonPrefix` if the testconfig.json prefix is the point).

If you actually want a sparse-indexed test, that would need to bypass the enumerator (which enforces contiguous indices, per the comment block) and hand-build a `JsonCommandLineOptionEntry` with `Arguments` that simulate a gap — but the validator doesn't see indices, only `Arguments.Count`, so there's nothing meaningful to assert. The current test is the right behavior to lock down; only the name needs updating.

src/Platform/Microsoft.Testing.Platform/Hosts/TestHostBuilder.CommonServices.cs
```diff
@@ -190,15 +195,32 @@ private async Task<BuildContext> SetupCommonServicesAsync(
 
         serviceProvider.AddService(context.TestFrameworkCapabilities);
 
+        IReadOnlyList<JsonCommandLineOptionEntry> jsonCommandLineOptions;
+        try
+        {
+            jsonCommandLineOptions = context.Configuration.EnumerateJsonCommandLineOptions();
+        }
+        catch (FormatException ex)
```
Evangelink: **[LOW] Algorithmic Correctness — FormatException path ignores `HasTool` (asymmetric with validator failure path)**

The standard validator-failure block 8 lines below this checks `!loggingState.CommandLineParseResult.HasTool` and **skips** the early-exit when the user is invoking a CLI tool (`--info`, `--server`, `--help`, etc.). That's deliberate — tool invocations should run even when CLI options have validation errors.

This catch block does not have the same guard, so a malformed `testconfig.json` will block tools too:

```text
$ dotnet test -- --info           # user is trying to inspect platform info
[Error] The entry 'foo' under section 'commandLineOptions' in testconfig.json ('...') must be...
[exit code: InvalidCommandLine]
```

Tools that don't read `testconfig.json` at all (`--info`, `--version`, `--diagnostic`) get blocked unnecessarily. Tools that do read it will fail with a clearer error anyway when they actually try to consume the malformed JSON.

Suggested change:

```csharp
catch (FormatException ex)
{
    if (!loggingState.CommandLineParseResult.HasTool)
    {
        await DisplayBannerIfEnabledAsync(...).ConfigureAwait(false);
        await context.ProxyOutputDevice.DisplayAsync(...,
            new ErrorMessageOutputDeviceData(ex.Message), ...).ConfigureAwait(false);
        await context.CommandLineHandler.PrintHelpAsync(...).ConfigureAwait(false);

        CompleteBuilderActivity(context.BuilderActivity, nameof(InformativeCommandLineHost));
        context.EarlyHost = new InformativeCommandLineHost((int)ExitCode.InvalidCommandLine, serviceProvider);
        return context;
    }

    // Tool invocations bypass JSON config validation — they will surface their own errors
    // if they actually need the malformed section.
    jsonCommandLineOptions = [];
}
```

Minor secondary nit on the same path: the standard failure prepends a localized "Invalid command-line arguments:" header (via `AddCommandLine` → `CommandLineInvalidArgumentsErrorMessage`). The FormatException catch outputs `ex.Message` raw with no header. The resource string for `JsonConfigurationSectionMustBeAnObjectErrorMessage` is self-describing so it reads fine, but it does mean the two error paths render differently. Not blocking — flagging for awareness.

Evangelink (commented):
## Round-2 review summary

All four round-1 fixes are correctly applied and the build is clean (0 warnings, 0 errors; 1034 pass / 0 fail). Two minor correctness observations and three test-hygiene NITs left inline. Nothing blocking.

### Fix validation

| Fix | Status |
|---|---|
| Case-insensitive duplicate detection (3 dicts + `ToDictionary`) | ✅ Applied correctly. One remaining edge case at validator.cs:62 — see inline. |
| `FormatException` catch around `EnumerateJsonCommandLineOptions` | ✅ Mirrors the standard failure path, no double-banner risk. One asymmetry re `HasTool` — see inline. |
| `[DoesNotReturn]` on the two throw helpers | ✅ Both helpers carry the attribute. |
| Rewritten XML doc on `TryGetCommandLineOptionArguments` | ✅ No more stale "validation on parseResult only" wording. |

### New test adequacy

| Test | Verdict |
|---|---|
| `EnumerateCommandLineOptions_SectionNameCaseInsensitive_Honored` | ✅ Strong assertions (option name + argument value). |
| `Validator_EmptyOptionNameInJson_FailsWithJsonPrefix` | ⚠️ Only checks `"testconfig.json"` substring — wouldn't catch a resource string change. Advisory. |
| `Validator_DuplicateOptionNamesDifferingByCase_FailsGracefully` | ⚠️ Test comment misrepresents pre-fix behavior; assertion only checks `IsValid == false`. See inline. |
| `Validator_JsonSparseIndexedEntry_DefensiveSchemaRejection` | ⚠️ Test name doesn't match what's tested. See inline. |

### Public API hygiene

✅ `PublicAPI.Unshipped.txt` only has `#nullable enable` — all new types (`JsonCommandLineOptionEntry`, `CommandLineConfigurationSource`, `CommandLineConfigurationProvider`) and methods (`EnumerateJsonCommandLineOptions`, `TryGetCommandLineOptionArguments`, `IsCommandLineOptionSet`) are correctly `internal`.
✅ `JsonCommandLineOptionEntry` uses plain `get;` properties — **no `init` accessors** on any new API.
✅ New `internal` `CommandLineHandler` constructor accepting `IConfiguration?` preserves source/binary compatibility (the existing public constructor delegates with `configuration: null`).

### Summary table

| # | Dimension | Verdict |
|---|-----------|---------|
| 1 | Algorithmic Correctness | 🟡 2 LOW (system+system dictionary, FormatException + `HasTool`) |
| 13 | Test Completeness & Coverage | ⚪ 1 NIT (weaker substring assertions on 3 new tests) |
| 16 | Naming & Conventions | ⚪ 1 NIT (`Validator_JsonSparseIndexedEntry_*` name) |
| 17 | Documentation Accuracy | ⚪ 1 NIT (test comment misrepresents pre-fix behavior) |

✅ 17/21 dimensions clean.

Threading & Concurrency, Security & IPC, Public API & Binary Compatibility, Performance & Allocations, Cross-TFM Compatibility, Resource & IDisposable, Defensive Coding at Boundaries, Localization, Test Isolation, Assertion Quality, Flakiness Patterns, Data-Driven Coverage, Code Structure, Analyzer Quality (N/A), IPC Wire Compatibility (N/A), Build Infrastructure, Scope Discipline — all clean.

### Noted for follow-up (out-of-scope for this PR)

- **Env-var injection of `commandLineOptions:*` keys**: now that `IsCommandLineOptionSet`/`TryGetCommandLineOptionArguments` route through `IConfiguration`, an env var spelled `commandLineOptions__timeout=30s` (with `__`→`:` normalization in `EnvironmentVariablesConfigurationProvider`) will silently shadow JSON values and be invisible to `CommandLineOptionsValidator` (which only sees CLI + JSON entries). This is an existing platform concern that the unification only *exposes* — not introduced here — but worth tracking.
- **Behavioral break for extension authors** who intentionally registered case-differing option names (e.g. `"Timeout"` for one provider, `"timeout"` for another). They were previously accepted as distinct (silently mis-treated as one by every downstream case-insensitive lookup); they now fail validation up front. This is the correct fix, but worth a CHANGELOG entry if not already planned.

Evangelink:
## Round 3 review — `bcb797345`

✅ **21/21 dimensions clean — no blocking, medium, or low findings.**

### Round-2 fixes validation

**Fix #1 — `ValidateOptionsAreNotDuplicated` extended to cover system providers: CORRECT and complete.**

I traced every collision shape through the new pass and confirmed no legitimate existing behavior changes:

- **Extension-vs-extension (case-differing).** Previously crashed with `ArgumentException` at the `providerAndOptionByOptionName.ToDictionary(..., StringComparer.OrdinalIgnoreCase)` lookup on line 60-62. Now caught with the friendly "declared by multiple" error. ✅
- **Extension-vs-system (any casing).** Already caught earlier by `ValidateExtensionOptionsDoNotContainReservedOptions` (line 107-152), which uses `OrdinalIgnoreCase` and returns early. The new `Concat`-over-both-dictionaries iteration is reachable in theory for this case but practically pre-empted. ✅
- **System-vs-system (case-differing).** Previously crashed at the same `ToDictionary` lookup. Now caught with the friendly error. ✅ (Pinned by the new `Validator_DuplicateOptionNamesAcrossSystemAndSystem_FailsGracefully` test.)

I considered whether two providers could legitimately register the same option name (e.g. a system provider exposing a public alias also added by an extension). The contract is that each `ICommandLineOptionsProvider.GetCommandLineOptions()` returns options it _owns_, and the pre-existing `ValidateExtensionOptionsDoNotContainReservedOptions` pass already treats extension-vs-system collisions as errors (any casing). So no legitimate scenario is regressed.

Minor cosmetic observation (NIT, not blocking): line 60 still uses LINQ `Union` over the two dictionaries, while the new code at line 165 uses `Concat`. Both produce the same result here (provider keys are reference-equal across the two source dictionaries, and there are no duplicates by reference), but slight stylistic inconsistency. Not worth a follow-up.

---

**Fix #2 — `FormatException` exception filter + rich error display: CORRECT and complete.**

- The `when (!loggingState.CommandLineParseResult.HasTool)` exception filter is standard C# 6+ and works on all four TFMs.
- The silent-degrade-to-empty-list path for tools is **consistent with the existing pipeline convention**: validation errors are already gated by `!HasTool` on line 230, so tools that have a structurally-broken `commandLineOptions` section already wouldn't see validation surface the issue. The `FormatException` fallback mirrors this.
- I verified `HasTool` semantics: `ParseResult.HasTool => ToolName is not null`. `--help` and `--info` are regular options handled via `IsHelpInvoked` / `IsInfoInvoked` on `CommandLineHandler` (lines 240-247), **not** tools. So `--help` and `--info` go through the normal `!HasTool` path and **do** surface FormatException with the rich `InvalidCommandLineArguments` header.
- Only server-mode tools (those that set `ToolName` via the tool entry, e.g. `--server`) silently degrade. Even for these, the rest of `testconfig.json` (e.g. `results-directory`, custom `IConfiguration[key]` lookups) remains accessible because the throw fires only during the typed-schema enumeration of the `commandLineOptions` section — direct `provider.TryGet` calls bypass it.
- The rich-error formatting (`StringBuilder` + `PlatformResources.InvalidCommandLineArguments` header + `"\t- {ex.Message}"` + `TrimEnd()`) matches `CommandLineOptionsValidator` (line 23-28). `Environment.NewLine` is correctly avoided (confirmed via grep). `StringBuilder` resolves via the global `System.Text` using in `Directory.Build.props`.
- `DisplayBannerIfEnabledAsync` call inside the catch (line 205) safely reads `--no-banner` via `commandLineOptions.IsOptionSet`, which routes through `provider.TryGet` on the already-flattened key-value store, not through `EnumerateCommandLineOptions`. No re-throw risk.

---

### Fresh-pass net-new findings

Nothing blocking, medium, or low. Three honest NIT-level observations, all noted-for-follow-up only:

1. **NIT (style)** — `CommandLineOptionsValidator.cs` line 60 uses `Union` where `Concat` would be slightly clearer and marginally cheaper (the new code at line 165 already uses `Concat`).
2. **NIT (test redundancy)** — `Validator_JsonEntryWithTooFewArguments_FailsArityCheck` (line 454) overlaps with `Validator_JsonArityTooFew_Fails` (line 200); only the arity shape differs (`ArgumentArity(2,2)` vs `ExactlyOne`). Defensible as defensive coverage of a non-trivial arity.
3. **NIT (docs precision)** — The comment on `Validator_DuplicateOptionNamesDifferingByCase_FailsGracefully` (lines 410-415) cites `line 62`, which will become stale on future edits to the validator. Consider replacing with a symbolic reference ("the `OrdinalIgnoreCase` `ToDictionary` lookup").

---

### Other dimensions checked clean

- **Public API & init accessors** — All new types/members are `internal`. No `PublicAPI.Unshipped.txt` churn. ✅
- **Cross-TFM** — Exception filters (C# 6+), `StringBuilder`, `CultureInfo`, `OrdinalIgnoreCase` all available on `net462` / `netstandard2.0` / `net8.0` / `net9.0`. ✅
- **Localization** — Two new resx strings (`JsonCommandLineOptionsEntryMustBeScalarOrArrayErrorMessage`, `JsonCommandLineOptionsValidationErrorPrefix`) with `<comment>` metadata. `.xlf` deltas look auto-generated (stub-per-resource pattern, +10 per language file). No manual xlf edits. ✅
- **Threading** — All affected code paths run on the single-threaded host startup pipeline. No new shared mutable state. ✅
- **IPC wire compatibility** — N/A; no serialized type changes. ✅
- **Resource management** — N/A; no new disposables. ✅
- **Defensive coding** — Trust boundary (user `testconfig.json`) correctly guarded; internal invariants (e.g. sparse-indexed defensive guard in `JsonConfigurationProvider`) correctly preserved. ✅
- **Test isolation** — Tests use isolated `TestProvider` instances per `[TestMethod]`; no shared static mutable state. ✅
- **Flakiness** — No `Thread.Sleep`, no wall-clock assertions, no hard-coded ports in new tests. ✅

---

### Verdict

**Ready to merge.** Round-2 fixes are correct, complete, and free of new bugs. The three NIT observations above are stylistic/documentation polish and do not affect users. Items explicitly deferred in the PR description (scalar-bool ambiguity, `--diagnostic*`/`--config-file` bootstrap reads, `IsCommandLineOptionSet` API visibility) remain out of scope for this PR as agreed.

src/Platform/Microsoft.Testing.Platform/Configurations/JsonConfigurationProvider.cs
```diff
@@ -219,11 +423,28 @@ private static bool StartsWithChar(string text, char c)
...
                 ConfigurationFile ?? "<unknown>"));
+
+        [DoesNotReturn]
+        private void ThrowEntryMustBeScalarOrArray(string fullKey, string sectionName)
+            => throw new FormatException(string.Format(
+                CultureInfo.InvariantCulture,
+                PlatformResources.JsonCommandLineOptionsEntryMustBeScalarOrArrayErrorMessage,
+                fullKey,
+                sectionName,
+                ConfigurationFile ?? "<unknown>"));
```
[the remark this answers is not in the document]
Evangelink: Addressed in e82a193c4. ThrowEntryMustBeScalarOrArray now strips the leading <sectionName>: prefix from the flattened key before formatting, so {0} renders as the entry name relative to the section (e.g. oo or oo:0) instead of the redundant commandLineOptions:foo. The resx `<comment>` was also tightened to document this contract. A new Assert.DoesNotContain("'commandLineOptions:foo'", ex.Message) assertion in EnumerateCommandLineOptions_NestedObject_IsRejected pins the new rendering.

src/Platform/Microsoft.Testing.Platform/Resources/PlatformResources.resx
```diff
@@ -158,6 +158,14 @@
     <value>The entry '{0}' under section '{1}' in the testconfig.json file ('{2}') must be a scalar string value. Nested objects and arrays are not supported.</value>
     <comment>{0} is the full dotted/colon-separated key. {1} is the section name. {2} is the path to the testconfig.json file.</comment>
   </data>
+  <data name="JsonCommandLineOptionsEntryMustBeScalarOrArrayErrorMessage" xml:space="preserve">
+    <value>The entry '{0}' under section '{1}' in the testconfig.json file ('{2}') must be either a scalar value (string, number, boolean) or an array of scalar values. Nested objects are not supported.</value>
+    <comment>{0} is the option name. {1} is the section name (commandLineOptions). {2} is the path to the testconfig.json file.</comment>
```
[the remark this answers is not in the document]
Evangelink: Addressed in e82a193c4 alongside the matching JsonConfigurationProvider change: the throw helper now strips the leading section prefix before formatting {0}, so the placeholder receives the entry name relative to the section (e.g. oo or oo:0) rather than the full colon-separated key. The resx `<comment>` is updated to document this precisely.

test/UnitTests/Microsoft.Testing.Platform.UnitTests/Configuration/JsonCommandLineOptionsTests.cs
Evangelink: Comment reworded in e82a193c4 to use the wording you suggested (silent acceptance via default Ordinal comparison; downstream lookups are case-insensitive; OrdinalIgnoreCase fix surfaces the duplicate up front). The assertion was already tightened in the previous round to Assert.Contains("declared by multiple", result.ErrorMessage), which locks the test to the duplicate-options pass.

src/Platform/Microsoft.Testing.Platform/CommandLine/CommandLineOptionsValidator.cs
Evangelink: Already addressed in 75c81794e (Round-2): ValidateOptionsAreNotDuplicated now iterates xtensionOptionsByProvider.Union(systemOptionsByProvider) with an OrdinalIgnoreCase dictionary, so system+system case-differing collisions are caught here and the function returns at line 47 before reaching the ToDictionary call at line 62. Pinned by Validator_DuplicateOptionNamesAcrossSystemAndSystem_FailsGracefully.

test/UnitTests/Microsoft.Testing.Platform.UnitTests/Configuration/JsonCommandLineOptionsTests.cs
Evangelink: Already addressed in 75c81794e (Round-2): renamed to Validator_JsonEntryWithTooFewArguments_FailsArityCheck (current file line 454) per your suggested name.

src/Platform/Microsoft.Testing.Platform/Hosts/TestHostBuilder.CommonServices.cs
Evangelink: Already addressed in 75c81794e (Round-2): the 	ry { EnumerateJsonCommandLineOptions(); } catch (FormatException ex) when (!loggingState.CommandLineParseResult.HasTool) { ... } catch (FormatException) { jsonCommandLineOptions = []; } shape on lines 199-221 of TestHostBuilder.CommonServices.cs now mirrors the standard validator-failure block: non-tool invocations get the rich InvalidCommandLineArguments header + the raw x.Message, while tool invocations silently degrade to an empty list so --info / --server / --version keep working even with a malformed 	estconfig.json.

test/UnitTests/Microsoft.Testing.Platform.UnitTests/Configuration/JsonCommandLineOptionsTests.cs
Evangelink: Addressed:
- Comment rewritten in e82a193c4 to match your suggested wording (pre-fix the default Ordinal comparison in `ValidateOptionsAreNotDuplicated` treated the two casings as distinct names while every downstream lookup is case-insensitive).
- Assertion tightened in 8b3c8ad46 to also pin both provider display names (`Assert.Contains(providerA.DisplayName, …)` / `Assert.Contains(providerB.DisplayName, …)`) on both `Validator_DuplicateOptionNamesDifferingByCase_FailsGracefully` and the symmetric system/system test, so a refactor that changes which pass surfaces the failure will be caught.

src/Platform/Microsoft.Testing.Platform/Resources/PlatformResources.resx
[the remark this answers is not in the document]
Evangelink: Addressed in e82a193c4 — the resx `<comment>` now describes `{0}` as the section-relative entry name (e.g. `timeout` or an indexed key like `timeout:0`) and the code in `ThrowEntryMustBeScalarOrArray` strips the section prefix so the actual format argument matches that documentation.

src/Platform/Microsoft.Testing.Platform/Configurations/JsonConfigurationProvider.cs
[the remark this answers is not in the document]
Evangelink: Addressed in e82a193c4 — `ThrowEntryMustBeScalarOrArray` now strips the section prefix from the JSON key before formatting, so the error message shows the section-relative entry name (e.g. `timeout` or `timeout:0`) matching the `<comment>` on the resx entry.

src/Platform/Microsoft.Testing.Platform/CommandLine/CommandLineOptionsValidator.cs
```diff
@@ -214,13 +233,31 @@ private static ValidationResult ValidateNoUnknownOptions(
...
             ? ValidationResult.Invalid(stringBuilder.ToTrimmedString())
             : ValidationResult.Valid();
     }
 
     private static ValidationResult ValidateOptionsArgumentArity(
         CommandLineParseResult parseResult,
+        IReadOnlyList<JsonCommandLineOptionEntry>? jsonCommandLineOptions,
         Dictionary<string, (ICommandLineOptionsProvider Provider, CommandLineOption Option)> providerAndOptionByOptionName)
     {
         StringBuilder? stringBuilder = null;
```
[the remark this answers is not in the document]
Evangelink: Fixed in 8647fb6f33169a362796ffaca0a6a83980e691db: CommandLineOptionsValidator.ValidateOptionsArgumentArity now groups parsed option names with StringComparer.OrdinalIgnoreCase, and ArgumentArityTests covers duplicate arity across --zeroOrOneArgumentsOption/--ZeroOrOneArgumentsOption.

src/Platform/Microsoft.Testing.Platform/CommandLine/CommandLineOptionsValidator.cs
```diff
@@ -143,11 +152,17 @@ private static ValidationResult ValidateExtensionOptionsDoNotContainReservedOpti
...
+        // ensures we catch case-differing duplicates (e.g. "Timeout" vs "timeout") with a friendly
+        // error rather than a later ArgumentException when building the lookup dictionary.
+        // We cover both extension and system providers so that any pair of case-differing
+        // duplicates — extension/extension, extension/system, or system/system — surfaces as a
+        // ValidationResult.Invalid rather than crashing the platform.
+        var optionNameToProviders = new Dictionary<string, HashSet<ICommandLineOptionsProvider>>(StringComparer.OrdinalIgnoreCase);
+        foreach (KeyValuePair<ICommandLineOptionsProvider, IReadOnlyCollection<CommandLineOption>> kvp in extensionOptionsByProvider.Concat(systemOptionsByProvider))
         {
             ICommandLineOptionsProvider provider = kvp.Key;
             foreach (CommandLineOption option in kvp.Value)
```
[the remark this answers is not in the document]
Evangelink: Fixed in 8647fb6f33169a362796ffaca0a6a83980e691db: PlatformResources.CommandLineOptionIsDeclaredByMultipleProviders now says multiple providers, UpdateXlf regenerated the xlf files, and CommandLineHandlerTests covers duplicate system providers.

PR #8665, opened 2026-05-28
Reject invalid runId in JSON-RPC requests with InvalidParams error

### LINKED ISSUE #8423: [MTP Core Analysis] JSON-RPC request deserialization coerces invalid runIds to Guid.Empty
## Summary
Both JSON-RPC deserializers accept any `runId` string and silently coerce invalid values to `Guid.Empty` instead of rejecting the request.

## Evidence
- `src/Platform/Microsoft.Testing.Platform/ServerMode/JsonRpc/Json/Json.cs:486-511`
  ```csharp
  string runId = json.Bind<string>(jsonElement, JsonRpcStrings.RunId);
  _ = Guid.TryParse(runId, out Guid result);
  ...
  return new DiscoverRequestArgs(
      RunId: result,
      ...);
  ```
  and the same pattern for `RunRequestArgs`.
- `src/Platform/Microsoft.Testing.Platform/ServerMode/JsonRpc/SerializerUtilities.cs:588-610`
  ```csharp
  _ = Guid.TryParse(GetRequiredPropertyFromJson<string>(properties, JsonRpcStrings.RunId), out Guid runId);
  ...
  return new DiscoverRequestArgs(runId, tests, filter);
  ```
  and the same pattern for `RunRequestArgs`.
- `src/Platform/Microsoft.Testing.Platform/Hosts/ServerTestHost.cs:477-531`
  uses `args.RunId` to create the per-request consumer and to emit completion/update messages back to the client.

## Why this is a real issue
A malformed client payload such as `{"runId":"not-a-guid"}` is not rejected as an invalid request. Instead, MTP creates the request with `Guid.Empty`.

That means request correlation continues with a synthetic run id the client never asked for. Updates and completion notifications are then emitted under `Guid.Empty`, which makes protocol debugging harder and can collapse multiple malformed requests onto the same identifier.

## Suggested resolution
Treat an unparseable `runId` as a protocol error and reject the request during deserialization (or at least during request validation) instead of converting it to `Guid.Empty`. Add tests for invalid `runId` values on both serializer implementations.

## Related issues
- #6179
- #6337

Evangelink:
Fixes #8423.

## Problem

When a client sends a JSON-RPC request (`testing/discoverTests` or `testing/runTests`) with a malformed `runId` value (e.g., a non-GUID string), the deserializer silently coerced it to `Guid.Empty` via `Guid.TryParse`. This hid the protocol violation from the client and could cause hard-to-diagnose downstream behavior.

## Fix

- `DiscoverRequestArgs` and `RunRequestArgs` deserializers in both `Json.cs` (System.Text.Json) and `SerializerUtilities.cs` (Jsonite) now throw `MessageFormatException` when `runId` cannot be parsed as a `Guid`.
- The `RpcMessage` deserializer now parses `id` first, then wraps the per-method params deserialization in a `try/catch`. On failure, it embeds an `InvalidRequestParamsArgs(ErrorCode, ErrorMessage)` sentinel in `Params` so the message can flow through to the request handler without crashing the server's message-handling loop.
- `ServerTestHost.HandleRequestCoreAsync` detects the sentinel and throws a new `JsonRpcException` carrying the proper error code; `HandleRequestAsync` catches it and sends a JSON-RPC `InvalidParams` (-32602) error response to the client.

## Tests

- Added `FormatterUtilitiesTests` coverage for invalid `runId` on both args types (per-args deserializer throws; full `RpcMessage` deserializer captures sentinel).
- Added a `ServerTests` integration test that sends a malformed request, asserts an `InvalidParams` error is returned, and verifies the server cleanly handles the subsequent `exit` notification (i.e., the loop is still alive).
- Fixed pre-existing `DiscoveryRequestCanBeCanceled` test that used an invalid `runId` of `Run1` (now uses a valid GUID).

### DIFF
--- a/src/Platform/Microsoft.Testing.Platform/Hosts/ServerTestHost.cs
+++ b/src/Platform/Microsoft.Testing.Platform/Hosts/ServerTestHost.cs
@@ -368,6 +368,11 @@ internal sealed partial class ServerTestHost : CommonHost, IServerTestHost, IDis
                 await SendErrorAsync(reqId: request.Id, errorCode: errorCode, message: errorMessage, data: null, cancellationToken).ConfigureAwait(false);
                 CompleteRequest(ref _clientToServerRequests, request.Id, completion => completion.TrySetCanceled());
             }
+            catch (JsonRpcException e)
+            {
+                await SendErrorAsync(reqId: request.Id, errorCode: e.ErrorCode, message: e.Message, data: null, cancellationToken).ConfigureAwait(false);
+                CompleteRequest(ref _clientToServerRequests, request.Id, completion => completion.SetException(e));
+            }
             catch (Exception e)
             {
                 await SendErrorAsync(reqId: request.Id, errorCode: 0, message: e.ToString(), data: null, cancellationToken).ConfigureAwait(false);
@@ -420,6 +425,11 @@ internal sealed partial class ServerTestHost : CommonHost, IServerTestHost, IDis
 
         switch (message.Method, message.Params)
         {
+            case (_, InvalidRequestParamsArgs invalidParams):
+                {
+                    throw new JsonRpcException(invalidParams.ErrorCode, invalidParams.ErrorMessage);
+                }
+
             case (JsonRpcMethods.Initialize, InitializeRequestArgs args):
                 {
                     _client = new(args.ClientInfo.Name, args.ClientInfo.Version);
--- a/src/Platform/Microsoft.Testing.Platform/Hosts/ServerTestHost.cs
+++ b/src/Platform/Microsoft.Testing.Platform/Hosts/ServerTestHost.cs
@@ -371,7 +371,7 @@ internal sealed partial class ServerTestHost : CommonHost, IServerTestHost, IDis
             catch (JsonRpcException e)
             {
                 await SendErrorAsync(reqId: request.Id, errorCode: e.ErrorCode, message: e.Message, data: null, cancellationToken).ConfigureAwait(false);
-                CompleteRequest(ref _clientToServerRequests, request.Id, completion => completion.SetException(e));
+                CompleteRequest(ref _clientToServerRequests, request.Id, completion => completion.TrySetException(e));
             }
             catch (Exception e)
             {
--- a/src/Platform/Microsoft.Testing.Platform/ServerMode/JsonRpc/Json/Json.cs
+++ b/src/Platform/Microsoft.Testing.Platform/ServerMode/JsonRpc/Json/Json.cs
@@ -425,12 +425,16 @@ internal sealed class Json
                             _ => null,
                         };
                     }
-                    catch (MessageFormatException ex)
+                    catch (Exception ex) when (ex is MessageFormatException or InvalidOperationException or JsonException)
                     {
                         // If params can't be deserialized for a request, capture the failure so
                         // we can later send back a properly coded JSON-RPC error using the request id.
                         // For notifications there's no one to respond to, but we still avoid
                         // crashing the message-handling loop by swallowing into the sentinel.
+                        // We catch the broader set of deserialization-related exceptions because the
+                        // request payload is untrusted client input and the lower-level helpers
+                        // (e.g. JsonElement.GetString() on a non-string element) can throw types
+                        // other than MessageFormatException.
                         @params = new InvalidRequestParamsArgs(ErrorCodes.InvalidParams, ex.Message);
                     }
                 }
@@ -501,7 +505,7 @@ internal sealed class Json
             string runId = json.Bind<string>(jsonElement, JsonRpcStrings.RunId);
             if (!Guid.TryParse(runId, out Guid result))
             {
-                throw new MessageFormatException($"'{JsonRpcStrings.RunId}' field is not a valid Guid");
+                throw new MessageFormatException(JsonRpcStrings.InvalidRunIdErrorMessage);
             }
 
             json.TryArrayBind(jsonElement, out TestNode[]? testNodes, JsonRpcStrings.Tests);
@@ -518,7 +522,7 @@ internal sealed class Json
             string runId = json.Bind<string>(jsonElement, JsonRpcStrings.RunId);
             if (!Guid.TryParse(runId, out Guid result))
             {
-                throw new MessageFormatException($"'{JsonRpcStrings.RunId}' field is not a valid Guid");
+                throw new MessageFormatException(JsonRpcStrings.InvalidRunIdErrorMessage);
             }
 
             json.TryArrayBind(jsonElement, out TestNode[]? testNodes, JsonRpcStrings.Tests);
@@ -643,21 +647,9 @@ internal sealed class Json
     }
 
     internal T Bind<T>(JsonElement element, string? property = null)
-    {
-        if (property is not null)
-        {
-            try
-            {
-                element = element.GetProperty(property);
-            }
-            catch (KeyNotFoundException ex)
-            {
-                throw new KeyNotFoundException($"Key '{property}' was not found in the dictionary.", ex);
-            }
-        }
-
-        return Deserialize<T>(element);
-    }
+        => property is not null && !element.TryGetProperty(property, out element)
+            ? throw new MessageFormatException($"'{property}' field is missing")
+            : Deserialize<T>(element);
 
     internal bool TryBind<T>(JsonElement element, out T? value, string? property = null)
     {
--- a/src/Platform/Microsoft.Testing.Platform/ServerMode/JsonRpc/Json/Json.cs
+++ b/src/Platform/Microsoft.Testing.Platform/ServerMode/JsonRpc/Json/Json.cs
@@ -405,24 +405,37 @@ internal sealed class Json
 
             if (json.TryBind(jsonElement, out string? method, JsonRpcStrings.Method))
             {
+                bool hasId = json.TryBind(jsonElement, out int id, JsonRpcStrings.Id);
+
                 object? @params = null;
                 if (jsonElement.TryGetProperty(JsonRpcStrings.Params, out JsonElement value))
                 {
-                    // Parse the specific methods
-                    @params = method switch
+                    try
                     {
-                        JsonRpcMethods.Initialize => json.Bind<InitializeRequestArgs>(value),
-                        JsonRpcMethods.TestingDiscoverTests => json.Bind<DiscoverRequestArgs>(value),
-                        JsonRpcMethods.TestingRunTests => json.Bind<RunRequestArgs>(value),
-                        JsonRpcMethods.CancelRequest => json.Bind<CancelRequestArgs>(value),
-                        JsonRpcMethods.Exit => json.Bind<ExitRequestArgs>(value),
+                        // Parse the specific methods
+                        @params = method switch
+                        {
+                            JsonRpcMethods.Initialize => json.Bind<InitializeRequestArgs>(value),
+                            JsonRpcMethods.TestingDiscoverTests => json.Bind<DiscoverRequestArgs>(value),
+                            JsonRpcMethods.TestingRunTests => json.Bind<RunRequestArgs>(value),
+                            JsonRpcMethods.CancelRequest => json.Bind<CancelRequestArgs>(value),
+                            JsonRpcMethods.Exit => json.Bind<ExitRequestArgs>(value),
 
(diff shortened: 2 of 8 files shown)


src/Platform/Microsoft.Testing.Platform/ServerMode/JsonRpc/SerializerUtilities.cs
```diff
@@ -482,22 +482,30 @@ static SerializerUtilities()
...
+                        JsonRpcMethods.TestingDiscoverTests => Deserialize<DiscoverRequestArgs>(paramsObj),
+                        JsonRpcMethods.TestingRunTests => Deserialize<RunRequestArgs>(paramsObj),
+                        JsonRpcMethods.CancelRequest => Deserialize<CancelRequestArgs>(paramsObj),
+                        JsonRpcMethods.Exit => Deserialize<ExitRequestArgs>(paramsObj),
+
+                        // Note: Let the server report unknown RPC request back to the client.
+                        _ => null,
+                    };
+                }
+                catch (MessageFormatException ex)
```
Evangelink: **MAJOR — Incomplete exception guard: wrong-typed `runId` escapes `catch (MessageFormatException)`**

`GetRequiredPropertyFromJson<string>` is implemented as:
```csharp
private static T GetRequiredPropertyFromJson<T>(...) =>
    (T)(GetOptionalPropertyFromJson(properties, propertyName)
        ?? throw new MessageFormatException($"'{propertyName}' field is missing"));
```

If a client sends `"runId": 42` (integer instead of string), the `(string)` unboxing cast throws `InvalidCastException`, **not** `MessageFormatException`. This escapes the `catch (MessageFormatException ex)` added at the outer level (line ~501), propagates to `HandleRequestAsync`'s `catch (Exception e)`, and the client receives an error with code `0` and `e.ToString()` (including stack trace) instead of the intended `-32602 InvalidParams`.

The same escape hatch applies to any other wrong-typed field in `InitializeRequestArgs`, `CancelRequestArgs`, etc.

**Recommendation:** Make `GetRequiredPropertyFromJson<T>` use a safe-cast and throw `MessageFormatException` on type mismatch:
```csharp
private static T GetRequiredPropertyFromJson<T>(IDictionary<string, object?> properties, string propertyName)
{
    object? value = GetOptionalPropertyFromJson(properties, propertyName)
        ?? throw new MessageFormatException($"'{propertyName}' field is missing");
    if (value is not T typed)
        throw new MessageFormatException($"'{propertyName}' field has wrong type (expected {typeof(T).Name})");
    return typed;
}
```
This ensures all missing-or-wrong-typed fields flow through the sentinel path correctly.

test/UnitTests/Microsoft.Testing.Platform.UnitTests/ServerMode/ServerTests.cs
```diff
@@ -222,6 +222,74 @@ public async Task DiscoveryRequestCanBeCanceled()
...
+                "method": "initialize",
+                "params": {
+                    "processId": 32,
+                    "clientInfo": { "name": "testingplatform-unittests", "version": "1.0.0" },
+                    "capabilities": {
+                        "testing": {
+                            "debuggerProvider": true
+                        }
+                    }
+                }
```
Evangelink, line 262: **MODERATE — No integration-level test for `testing/runTests` with an invalid `runId`**

The end-to-end path through `HandleRequestAsync` → `HandleRequestCoreAsync` → `catch (JsonRpcException)` is only exercised for `testing/discoverTests`. While both methods share the same codepath today, a test covering `testing/runTests` at the server level would protect against any future divergence (e.g., method-specific middleware or pre-checks).

Consider parameterising this test with `[DataRow("testing/discoverTests")]` / `[DataRow("testing/runTests")]`, mirroring the pattern already used in `DeserializeRpcMessage_InvalidRunId_CapturesInvalidParamsSentinel`.

src/Platform/Microsoft.Testing.Platform/ServerMode/JsonRpc/Json/Json.cs
```diff
@@ -486,7 +499,10 @@ public Json(Dictionary<Type, JsonSerializer>? serializers = null, Dictionary<Typ
         _deserializers[typeof(DiscoverRequestArgs)] = new JsonElementDeserializer<DiscoverRequestArgs>((json, jsonElement) =>
         {
             string runId = json.Bind<string>(jsonElement, JsonRpcStrings.RunId);
-            _ = Guid.TryParse(runId, out Guid result);
+            if (!Guid.TryParse(runId, out Guid result))
+            {
+                throw new MessageFormatException($"'{JsonRpcStrings.RunId}' field is not a valid Guid");
```
Evangelink: **NIT — Duplicate error message string**

The string `$"'{JsonRpcStrings.RunId}' field is not a valid Guid"` appears verbatim in four places: twice in `Json.cs` (for `DiscoverRequestArgs` and `RunRequestArgs`) and twice in `SerializerUtilities.cs`. A future wording change must be applied in multiple places and it's easy to miss one.

Consider extracting a shared constant (e.g., in `JsonRpcStrings` or a small `JsonRpcErrorMessages` helper class) so all four sites share a single definition.

src/Platform/Microsoft.Testing.Platform/ServerMode/JsonRpc/Json/Json.cs
```diff
@@ -486,7 +499,10 @@ public Json(Dictionary<Type, JsonSerializer>? serializers = null, Dictionary<Typ
         _deserializers[typeof(DiscoverRequestArgs)] = new JsonElementDeserializer<DiscoverRequestArgs>((json, jsonElement) =>
         {
             string runId = json.Bind<string>(jsonElement, JsonRpcStrings.RunId);
```
Evangelink, line 505: **MAJOR — Missing `runId` field throws `KeyNotFoundException`, bypassing the `catch (MessageFormatException)` guard**

`json.Bind<string>(jsonElement, JsonRpcStrings.RunId)` calls `jsonElement.GetProperty(property)`, which throws `KeyNotFoundException` (re-thrown as `KeyNotFoundException`) when the property is absent. Since the outer guard only catches `MessageFormatException`, a request payload that omits `runId` entirely:
```json
{ "jsonrpc": "2.0", "id": 1, "method": "testing/discoverTests", "params": {} }
```
will escape the sentinel path, hit `HandleRequestAsync`'s generic `catch (Exception e)`, and return error code `0` with `e.ToString()` rather than `-32602 InvalidParams`.

**Recommendation:** Update the `Bind<T>` helper to translate `KeyNotFoundException` to `MessageFormatException`:
```csharp
internal T Bind<T>(JsonElement element, string? property = null)
{
    if (property is not null)
    {
        if (!element.TryGetProperty(property, out element))
            throw new MessageFormatException($"Required field '{property}' is missing.");
    }
    return Deserialize<T>(element);
}
```
Using `TryGetProperty` instead of `GetProperty`+catch also avoids the exception-for-control-flow anti-pattern.

src/Platform/Microsoft.Testing.Platform/Hosts/ServerTestHost.cs
```diff
@@ -368,6 +368,11 @@ private async Task HandleRequestAsync(RequestMessage request, CancellationToken
                 await SendErrorAsync(reqId: request.Id, errorCode: errorCode, message: errorMessage, data: null, cancellationToken).ConfigureAwait(false);
                 CompleteRequest(ref _clientToServerRequests, request.Id, completion => completion.TrySetCanceled());
             }
+            catch (JsonRpcException e)
+            {
+                await SendErrorAsync(reqId: request.Id, errorCode: e.ErrorCode, message: e.Message, data: null, cancellationToken).ConfigureAwait(false);
+                CompleteRequest(ref _clientToServerRequests, request.Id, completion => completion.SetException(e));
```
Evangelink: **NIT — `SetException` vs `TrySetException` inconsistency**

The new `catch (JsonRpcException e)` block uses `completion.SetException(e)` while `catch (OperationCanceledException e)` uses `completion.TrySetCanceled()`. The pre-existing `catch (Exception e)` (line 379) also uses `SetException`, so this PR is internally consistent with it. However, both are actually safe here because `CompleteRequest` calls `TryRemove` first and only invokes the lambda if removal succeeds — no double-completion is possible.

That said, it is worth considering whether the intent is `TrySetException` for defensive consistency with `TrySetCanceled` and `TrySetResult` used elsewhere in this method (lines 253, 258, 358). A future refactor could change the locking assumptions and silently introduce a throw.

Evangelink (commented):
## Review Summary

The core design is sound: the sentinel + `JsonRpcException` plumbing correctly keeps the message loop alive and sends the right `-32602 InvalidParams` error code for the targeted `runId` validation case. The cleanup of the `"Run1"` test fixture is a necessary correctness fix.

However, the exception guard that was added to route deserialization failures through the sentinel path is **incomplete**: it only catches `MessageFormatException`, but two adjacent code paths throw different exception types that escape the guard:

1. **`SerializerUtilities.cs` — wrong-typed fields** (`"runId": 42`): `GetRequiredPropertyFromJson<string>` does an unsafe `(T)` unboxing cast → `InvalidCastException`, not caught → client receives error code `0` with `e.ToString()` (stack trace leak) instead of `-32602`.
2. **`Json.cs` — missing field** (`runId` absent entirely): `Bind<string>` calls `element.GetProperty(property)` which throws `KeyNotFoundException`, not `MessageFormatException`, same outcome.

Both scenarios involve untrusted client input arriving at the same trust boundary the PR is hardening, so they are in scope to fix here.

The two **MAJOR** inline comments include concrete reproduction payloads and suggested fixes. The remaining comments are a NIT (duplicate string) and a MODERATE (missing `runTests` integration test).

> Generated by [Expert Code Review (on open)](https://github.com/microsoft/testfx/actions/runs/26584592680) for issue #8665 · sonnet46 6.6M

src/Platform/Microsoft.Testing.Platform/ServerMode/JsonRpc/SerializerUtilities.cs
Evangelink: Fixed in 23ea7a5. `GetRequiredPropertyFromJson<T>` now does a safe `is T typed` check and throws `MessageFormatException` (with the expected type name) on type mismatch, and the catch at the `RpcMessage` boundary was broadened to also include `InvalidCastException` as a defense-in-depth measure for any other code paths still doing direct casts on untrusted input.

test/UnitTests/Microsoft.Testing.Platform.UnitTests/ServerMode/ServerTests.cs
Evangelink: Fixed in 23ea7a5. The test is now parameterised with `[DataRow(JsonRpcMethods.TestingDiscoverTests)]` / `[DataRow(JsonRpcMethods.TestingRunTests)]` and renamed to `RequestWithInvalidRunId_ReturnsInvalidParamsError`.

src/Platform/Microsoft.Testing.Platform/ServerMode/JsonRpc/Json/Json.cs
Evangelink: Fixed in 23ea7a5. Extracted to `JsonRpcStrings.InvalidRunIdErrorMessage` and all four sites updated.

src/Platform/Microsoft.Testing.Platform/ServerMode/JsonRpc/Json/Json.cs
Evangelink: Fixed in 23ea7a5. `Bind<T>` now uses `TryGetProperty` and throws `MessageFormatException` directly. I also broadened the `RpcMessage` boundary catch to handle `InvalidOperationException`/`JsonException` (from `GetString()` etc. on wrong-typed primitives) so all untrusted-input deserialization failures route through the sentinel and yield `-32602 InvalidParams`.

src/Platform/Microsoft.Testing.Platform/Hosts/ServerTestHost.cs
Evangelink: Fixed in 23ea7a5. Changed to `TrySetException` for the new `JsonRpcException` catch. I left the pre-existing `catch (Exception)` `SetException` call alone to minimize churn.

PR #8714, opened 2026-05-31
fix: replace duplicate typeof(long) with typeof(ushort) in BaseSerializer.GetSize(T)

Evangelink:
## Summary

The `GetSize<T>()` switch expression in `BaseSerializer` had a duplicate `typeof(long)` arm that was unreachable dead code. The second arm should be `typeof(ushort)` to properly cover all the numeric types used in the IPC serializers.

**Before:**
```csharp
private static int GetSize<T>() => typeof(T) switch
{
    Type type when type == typeof(int) => sizeof(int),
    Type type when type == typeof(long) => sizeof(long),
    Type type when type == typeof(short) => sizeof(short),
    Type type when type == typeof(bool) => sizeof(bool),
    Type type when type == typeof(byte) => sizeof(byte),
    Type type when type == typeof(long) => sizeof(long),  // ← duplicate, unreachable
    _ => 0,
};
```

**After:**
```csharp
private static int GetSize<T>() => typeof(T) switch
{
    Type type when type == typeof(int) => sizeof(int),
    Type type when type == typeof(long) => sizeof(long),
    Type type when type == typeof(short) => sizeof(short),
    Type type when type == typeof(ushort) => sizeof(ushort),
    Type type when type == typeof(bool) => sizeof(bool),
    Type type when type == typeof(byte) => sizeof(byte),
    _ => 0,
};
```

No behavioral change — `WriteSize<ushort>` is not currently called, so this fix adds completeness while removing dead code.

> Generated by [Adhoc QA](https://github.com/microsoft/testfx/actions/runs/26705977955) · sonnet46 3.2M · [◷](https://github.com/search?q=repo%3Amicrosoft%2Ftestfx+%22gh-aw-workflow-id%3A+adhoc-qa%22&type=pullrequests)
>
<details>

```
gh aw add githubnext/agentics/workflows/adhoc-qa.md@main
```
</details>

### DIFF
--- a/src/Platform/Microsoft.Testing.Platform/IPC/Serializers/BaseSerializer.cs
+++ b/src/Platform/Microsoft.Testing.Platform/IPC/Serializers/BaseSerializer.cs
@@ -5,11 +5,8 @@
 using System.Buffers;
 #endif
 
-#if NET
-using Microsoft.Testing.Platform.Helpers;
-#endif
-
 using Microsoft.CodeAnalysis;
+using Microsoft.Testing.Platform.Helpers;
 
 namespace Microsoft.Testing.Platform.IPC.Serializers;
 
@@ -389,7 +386,7 @@ internal abstract class BaseSerializer
         Type type when type == typeof(ushort) => sizeof(ushort),
         Type type when type == typeof(bool) => sizeof(bool),
         Type type when type == typeof(byte) => sizeof(byte),
-        _ => 0,
+        _ => throw ApplicationStateGuard.Unreachable(),
     };
 
     public static bool IsNullOrEmpty<T>(T[]? list) => list is null || list.Length == 0;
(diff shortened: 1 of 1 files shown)


src/Platform/Microsoft.Testing.Platform/IPC/Serializers/BaseSerializer.cs
```diff
@@ -344,9 +386,9 @@ protected static void WriteAtPosition(Stream stream, int value, long position)
         Type type when type == typeof(int) => sizeof(int),
         Type type when type == typeof(long) => sizeof(long),
         Type type when type == typeof(short) => sizeof(short),
+        Type type when type == typeof(ushort) => sizeof(ushort),
         Type type when type == typeof(bool) => sizeof(bool),
         Type type when type == typeof(byte) => sizeof(byte),
-        Type type when type == typeof(long) => sizeof(long),
         _ => 0,
```
Youssef1313: Shouldn't this throw unreachable?

Evangelink: Good catch — fixed in 6e4ca9b1d. Replaced `_ => 0,` with `_ => throw ApplicationStateGuard.Unreachable(),` to make the invariant explicit (`GetSize<T>` is private and only called by `WriteSize<T>` with a known closed set of primitive types).

PR #8719, opened 2026-05-31
Strip --minimum-expected-tests on retry attempts in RetryOrchestrator

### LINKED ISSUE #5639: Revise `--minimum-expected-tests` behavior with Retry extension
This is a hypothesis that I didn't test yet.

But I think if `--minimum-expected-tests` is used with Retry, it could be problematic.

Let's say a test project has 10 tests, and `--minimum-expected-tests 5` is passed. If one test failed in the initial run, the subsequent run will run only the one failed test, causing `--minimum-expected-tests` to potentially fail the run?

We need to revise this scenario and try to support it.

At least if it already works, we will need a test to cover it.

Evangelink:
Fixes #5639.

When `--minimum-expected-tests` was combined with `--retry-failed-tests`, the retry attempt (which only re-runs the previously failed tests) inherited the threshold computed against the full first-attempt test set. The retry's much smaller test count then tripped the policy and the run exited with code 9 (`MinimumExpectedTestsPolicyViolation`), even when the original first attempt satisfied the threshold.

This change makes `RetryOrchestrator` strip `--minimum-expected-tests` from the per-iteration argument list before launching any retry attempt (i.e. when `lastListOfFailedId` is non-empty). The first attempt still honors the option because it runs the full set of tests.

Adds an acceptance test `RetryFailedTests_WithMinimumExpectedTests_StripsThresholdOnRetry` that exercises the scenario from the issue across all target frameworks and asserts the run completes successfully without the policy violation message.

### DIFF
--- a/src/Platform/Microsoft.Testing.Extensions.Retry/RetryOrchestrator.cs
+++ b/src/Platform/Microsoft.Testing.Extensions.Retry/RetryOrchestrator.cs
@@ -202,6 +202,11 @@ internal sealed class RetryOrchestrator : ITestHostExecutionOrchestrator, IOutpu
                 RemoveOption(finalArguments, TreeNodeFilterCommandLineOptionsProvider.TreenodeFilter);
                 RemoveOption(finalArguments, PlatformCommandLineProvider.FilterUidOptionKey);
 
+                // Strip --minimum-expected-tests on retry attempts: a retry only re-runs the previously
+                // failed tests, so propagating the original threshold (computed against the full test
+                // set) would always trip the policy and fail the run. See issue #5639.
+                RemoveOption(finalArguments, PlatformCommandLineProvider.MinimumExpectedTestsOptionKey);
+
                 // The RSP parser (ResponseFileHelper.SplitCommandLine) strips all '"' characters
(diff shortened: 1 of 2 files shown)


Evangelink (commented):

**Summary**

The fix is minimal, correctly scoped, and addresses the root cause:

- `RemoveOption(finalArguments, PlatformCommandLineProvider.MinimumExpectedTestsOptionKey)` is placed inside the `if (lastListOfFailedId is { Length: > 0 })` guard, so the first attempt retains the threshold while every retry attempt—which re-runs only the previously failed subset—has it stripped. That's exactly the right boundary.
- The acceptance test `RetryFailedTests_WithMinimumExpectedTests_StripsThresholdOnRetry` follows existing conventions (DynamicData over `TargetFrameworks.AllForDynamicData`, `Guid.NewGuid()` result directory, env-var-driven asset behavior) and verifies both the happy-path exit code and the absence of the violation message.
- No public API surface was changed; no IPC wire format was altered; no new allocations on hot paths; threading model is unaffected.

> Generated by [Expert Code Review (on open)](https://github.com/microsoft/testfx/actions/runs/26721446271) for issue #8719 · sonnet46 1.6M

PR #8794, opened 2026-06-03
Add ExecutionMode handshake property to MTP dotnet test IPC protocol

### LINKED ISSUE #8793: Add ExecutionMode to MTP handshake so SDK can detect --help / --list-tests passed via non-CLI channels
## Background

In the dotnet test ↔ MTP IPC protocol, the SDK has no reliable way to know what mode the test app is actually running in (run / help / discovery). This causes two user-facing bugs when `--help`, `-?` or `--list-tests` reach the test app through non-CLI channels that bypass `dotnet test` parsing — e.g.:

- `TestingPlatformCommandLineArguments` MSBuild property (or `RunArguments`)
- `launchSettings.json` `commandLineArgs`

### Today's behavior (verified against `main`)

- **`--help` via these channels**: In `TestHostBuilder.Modes.cs` the help-mode branch (around line 48) calls `HelpInvokedAsync` **without** first calling `IsCompatibleProtocolAsync`, so **the SDK never receives a handshake** in that path. The SDK then receives `CommandLineOptionMessages` while `TestOptions.IsHelp == false` and throws `InvalidOperationException`, which routes to `Environment.FailFast` (ugly stack-trace crash). See `dotnet/sdk` `TestApplication.cs` `OnCommandLineOptionMessages` and the surrounding `FailFast` block.
- **`--list-tests` via these channels**: `DiscoveredTestMessages` flow whether MTP is in run mode or list-tests mode (some adapters legitimately send discovery info during execution). The SDK currently has no signal to disambiguate and silently reports "zero tests run".

PR https://github.com/dotnet/sdk/pull/54487 attempted to fix this purely on the SDK side by tokenizing args from the known channels and pattern-matching a denylist of forbidden options. It was closed because that approach:

1. duplicates work MTP already does (parsing its own args),
2. is fragile (hand-rolled shell tokenizer, hand-maintained denylist),
3. can never catch all channels (env vars, future arg sources, custom launchers),
4. **cannot reliably solve the `--list-tests` case** because the message stream is identical between modes.

cc @Youssef1313 @JanKrivanek

### Part 1: Add an `ExecutionMode` property to `HandshakeMessage`

Today (`src/Platform/Microsoft.Testing.Platform/ServerMode/DotnetTest/IPC/Constants.cs`):

```csharp
internal static class HandshakeMessagePropertyNames
{
    internal const byte PID = 0;
    internal const byte Architecture = 1;
    internal const byte Framework = 2;
    internal const byte OS = 3;
    internal const byte SupportedProtocolVersions = 4;
    internal const byte HostType = 5;
    internal const byte ModulePath = 6;
    internal const byte ExecutionId = 7;
    internal const byte InstanceId = 8;
    internal const byte IsIDE = 9;
}
```

Add:

```csharp
    internal const byte ExecutionMode = 10;
```

Plus a small constants class for the values:

```csharp
internal static class HandshakeMessageExecutionModes
{
    internal const string Run = "run";
    internal const string Help = "help";
    internal const string Discover = "discover";  // matches PlatformCommandLineProvider.DiscoverTestsOptionKey = "list-tests"
}
```

The protocol design already states (Constants comment): *"We can add new properties with new ids, but we CANNOT change the existing ids (to support backwards compatibility)."* New optional properties are forwards/backwards compatible by design — older SDKs simply skip the unknown id. So **no `ProtocolConstants.Version` bump should be required**.

### Part 2: Always send the handshake — including from the `--help` path

`TestHostBuilder.Modes.cs` currently skips the handshake when `IsHelpCommand` is true. Move the `IsCompatibleProtocolAsync` call before any push (including `HelpInvokedAsync`) when in server mode, so the SDK always gets a handshake with the `ExecutionMode` it can rely on.

### Part 3: SDK-side check (follow-up in `dotnet/sdk`)

The SDK will compare `handshake.ExecutionMode` against its own expected mode (`TestOptions.IsHelp` / `TestOptions.IsDiscovery`):

| SDK expects | MTP reports | Action |
| --- | --- | --- |
| `Run` | `Run` | proceed |
| `Discover` | `Discover` | proceed |
| `Help` | `Help` | proceed |
| `Run` | `Help` | reject — user passed `--help` via non-CLI channel |
| `Run` | `Discover` | reject — user passed `--list-tests` via non-CLI channel |
| `Discover` | `Help` | reject — user passed `--help` via non-CLI channel |
| `Help` | anything else | reject (defensive) |
| any | (property missing — older MTP) | fall back to today's behavior |

The SDK reports a clean `GracefulException`-style error naming the offending option and telling the user to pass it directly to `dotnet test`.

## Why this is better than the SDK-only fix

- **MTP is the single source of truth** for which mode it's in — it already parses its args.
- **Covers all current and future arg channels for free** — no hand-maintained denylist.
- **Only way to reliably catch `--list-tests`** — the message stream alone can't disambiguate; the handshake property is definitive.
- **Backwards compatible** via the optional-property protocol design.

## Open questions

- Should `InformativeCommandLineHost` flows (`--info`) also send a mode (e.g. `Info`)? Probably yes for symmetry, but lower priority — `--info` doesn't currently cause user-visible breakage through these channels.
- `IsIDE` is sent on the *response* handshake (SDK → MTP). `ExecutionMode` belongs on the *request* (MTP → SDK), like the other identity properties. Worth calling out explicitly in the impl PR.

## Tracking

- Source bug: dotnet/sdk#50654
- Closed SDK-only attempt: dotnet/sdk#54487

Evangelink:
Implements the testfx side of the design proposed in #8793.

## What

Adds a new optional `ExecutionMode` property (byte `10`) to the handshake that the test host sends to `dotnet test`. This lets the SDK distinguish whether the test host is running in normal `run`, `help`, or `discover` mode, and surface mismatches such as `--help` (or `--list-tests`) being injected via `RunArguments` during what the SDK thinks is a normal run.

Also makes the test host always perform the handshake before printing help in server mode. Previously, `--help` (and other help-printing flows) skipped the handshake entirely, which prevented `dotnet test` from validating the host's mode.

## Wire format additions

Pinned by new freeze tests in `ProtocolTests`:

- `HandshakeMessagePropertyNames.ExecutionMode = 10`
- `HandshakeMessageExecutionModes.Run = "run"`
- `HandshakeMessageExecutionModes.Help = "help"`
- `HandshakeMessageExecutionModes.Discover = "discover"`

`ProtocolConstants.Version` is unchanged: the handshake is property-based, so the new key is an additive optional field that older SDKs ignore.

## Backward compatibility

| SDK | MTP | Behavior |
|---|---|---|
| old | new, `--help` | SDK now receives a handshake it previously did not. SDK's existing `_isHelp` flag (set from its own command line parse) already suppresses normal-run side effects on the SDK side, so behavior should be unchanged. |
| old | new, normal run with injected `--help` | SDK still cannot detect the mismatch (it does not read `ExecutionMode`). Same degraded behavior as today. |
| new | new | SDK reads `ExecutionMode` and can fail fast on mismatch. (Implemented in a follow-up dotnet/sdk PR.) |
| new | old | `ExecutionMode` property absent; SDK falls back to current behavior. |

## Tests

- New freeze tests in `ProtocolTests.cs` pin the byte / string wire values.
- The `TestHostBuilder.Modes.cs` help-branch behavior change (always handshake first) is not unit-tested here: there are no existing tests for `DotnetTestConnection` in isolation, and the path will be exercised end-to-end by the dotnet/sdk follow-up PR. Happy to add an integration test if reviewers prefer.

Closes #8793

## Follow-up

A dotnet/sdk PR will:
- mirror the constants in `CliConstants.cs`,
- read `ExecutionMode` in `TestApplicationHandler.OnHandshakeReceived` and validate against the SDK's known mode,
- update `TerminalTestReporter.HandshakeFailure` to drop the `_isHelp` early-return workaround.


Youssef1313 (commented):
Note: there was some workaround somewhere in SDK related to the missing help handshake.

I don't know if the version should be bumped to signify the fact that this has the proper handshake and then conditionally apply the SDK workaround only for old protocol version. Consider validating that. It's not the end of the world if we didn't do it though, I think.

src/Platform/Microsoft.Testing.Platform/Hosts/TestHostBuilder.Modes.cs
```diff
@@ -49,6 +49,18 @@ private async Task<IHost> BuildHostAsync(BuildContext context)
...
+                // Always perform the handshake (even though we are about to print help) so that
+                // dotnet test on the SDK side can validate the test host's execution mode and
+                // detect mismatches such as --help being injected via RunArguments during a
+                // normal run. The handshake is needed for the SDK to know that the test host
+                // is in help mode, so it can ignore any incoming CommandLineOptionMessages.
+                bool isProtocolCompatible = await pushOnlyProtocol.IsCompatibleProtocolAsync("TestHost").ConfigureAwait(false);
+                if (!isProtocolCompatible)
+                {
+                    CompleteBuilderActivity(context.BuilderActivity, nameof(InformativeCommandLineHost));
+                    return new InformativeCommandLineHost((int)ExitCode.IncompatibleProtocolVersion, context.ServiceProvider);
```
Youssef1313, line 61: As this is kinda its own "host" kind. Should we actually be re-using `TestHost` in the call above?

And if we can just pass something like `InformativeHost`, I think that would automatically imply help without the new protocol field.

The protocol field might still be needed for the case of discovery though.

Evangelink:
Thanks for the review @Youssef1313! Pushed `497dbc7c0` with the changes below.

**1. `InformativeHost` HostType (adopted your suggestion)** — Added `HandshakeMessageHostTypes` (`Constants.cs`) with three centralized strings: `TestHost`, `TestHostController`, and a new `InformativeHost`. The help branch in `TestHostBuilder.Modes.cs` now sends `HostType = InformativeHost`, so the SDK's existing `if (hostType == "TestHost")` gates in `TestApplicationHandler.OnHandshakeReceived` (the InstanceId-required check and the `AssemblyRunStarted` / `(Try N) Running tests from ...` line) naturally skip help-mode handshakes — no SDK change required for the immediate problem, and older SDKs silently tolerate the unknown value.

**2. `ExecutionMode` kept as an orthogonal explicit signal** — `HostType` answers *what kind of host* the process is; `ExecutionMode` answers *what it's doing right now*. Discovery still uses `HostType = TestHost` + `ExecutionMode = Discover`, so we'd still need the byte to distinguish run-vs-discover even if we only relied on HostType. Treating the two dimensions independently keeps the protocol clean and future-proof (e.g., a future host type could still run or discover).

**3. `ProtocolConstants.Version` not bumped** — Documented inline in `Constants.cs`: the SDK only supports a single protocol version at a time and bumping would require a coordinated SDK+platform release with hard breakage for any mismatched pair. The change is purely additive (new optional property byte `10`; unknown bytes are ignored), so older SDKs decoding a new platform's handshake just skip the `ExecutionMode` field — no observable behavior change.

**4. Test coverage** — Added a freeze test `HandshakeMessageHostTypes_ValuesAreStable` in `ProtocolTests.cs` and fixed the misleading "compile-time enforced" comments on the two existing freeze tests (they're runtime, not compile-time, asserts). `CommonHostTests.cs` test stub now also uses the new constant.

**Follow-up SDK scope (separate PR / not blocking this one)** — Even with the HostType change, the SDK could be tightened further to explicitly branch on the `ExecutionMode` byte (e.g., skip `RunComplete`/`SessionEndedMessage` bookkeeping when `ExecutionMode == Help`, exit cleanly without expecting a run summary). Happy to open a follow-up issue/PR on `dotnet/sdk` once this merges.

cc @Youssef1313 — let me know if you want me to open that SDK-side follow-up issue.

test/UnitTests/Microsoft.Testing.Platform.UnitTests/IPC/ProtocolTests.cs
```diff
@@ -150,6 +151,58 @@ public void HandshakeMessageWithNullProperties()
...
 
+    // The HandshakeMessagePropertyNames values are part of the wire protocol
+    // shared with dotnet test in the dotnet/sdk repository. Changing any
+    // existing value is a binary-breaking change for older platform <-> SDK
+    // pairings, so it is intentionally pinned by this test.
+    [TestMethod]
+    public void HandshakeMessagePropertyNames_ValuesAreStable()
+    {
+        // Round-trip the constants through the serializer so the MSTest analyzer
+        // does not flag the comparisons as "always true / always failing".
```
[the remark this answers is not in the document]
Evangelink: Good catch — fixed in 497dbc7c0. The misleading "compile-time enforced" / "round-tripped through serializer" wording has been replaced with a comment that accurately explains why the dictionary indirection is used (to bypass the MSTest analyzer that would otherwise flag `Assert.AreEqual` between two compile-time constants as "always true / always failing"). The test itself remains a runtime freeze of the wire-protocol byte values.

src/Platform/Microsoft.Testing.Platform/Hosts/TestHostBuilder.Modes.cs
Evangelink: Great suggestion — adopted in 497dbc7c0. Introduced `HandshakeMessageHostTypes` with three centralized strings (`TestHost`, `TestHostController`, and the new `InformativeHost`), and the help branch in `TestHostBuilder.Modes.cs` now sends `HostType = InformativeHost`. The SDK's existing `if (hostType == "TestHost")` gates in `TestApplicationHandler.OnHandshakeReceived` (the `InstanceId`-required check and the `AssemblyRunStarted` / `(Try N) Running tests from ...` line) therefore naturally skip help-mode handshakes — no SDK change required for the immediate problem, and older SDKs silently tolerate the unknown value.

I kept `ExecutionMode` as an orthogonal explicit signal: `HostType` answers *what kind of host* the process is, while `ExecutionMode` answers *what it's doing right now*. Discovery still uses `HostType = TestHost` + `ExecutionMode = Discover`, so we'd still need the byte to distinguish run-vs-discover even if we only relied on `HostType`. Treating the two dimensions independently keeps the protocol clean and future-proof.

PR #8851, opened 2026-06-05
Ignore auto-generated AGENTS.md in markdownlint

Evangelink:
AGENTS.md is generated by gh copilot-curate (added in #8773) and uses underscore emphasis (_skill_, _agent_) plus an H2 first heading. Editing it by hand would be lost on the next regeneration, so add it to the ignore list alongside the other auto-managed paths.

Currently blocks the `lint` check on all open PRs (#8580, #8586, #8819, #8820, #8823, #4268 ...).

cc @Evangelink

/cc microsoft/testfx-collaborators

PR #8852, opened 2026-06-05
Also ignore copilot-curate.instructions.md from markdownlint

Evangelink:
Follow-up to #8851. This second file under `.github/instructions/` is also auto-generated by `gh copilot-curate` and triggers 96 markdownlint errors (MD041/MD049/MD038) that would be wiped out on the next regeneration. Adding it to the ignores keeps the lint job green without fighting the generator.

This will also unblock lint CI on PRs #8580, #8586, #8819, #8820, #8823.


Evangelink (commented):
> [!NOTE]
> 🤖 **Automated review by GitHub Copilot.** Posted via a maintainer's GitHub token, so it appears under their account — the account owner did **not** write or approve this content personally. Generated by the [Expert Code Review workflow](https://github.com/microsoft/testfx/actions/runs/27002912109). To request a follow-up action, reply by tagging `@copilot` directly.

This is a config-only PR (`non-code` scope). All runtime, API surface, threading, performance, test quality, and IPC dimensions are N/A. The single-line change to `.markdownlint-cli2.jsonc` is syntactically correct JSONC, the new entry is consistent with the `AGENTS.md` entry directly above it (same comment, same rationale), and the approach of listing auto-generated files explicitly rather than glob-ignoring the whole `.github/instructions/` directory is the right call — it keeps legitimately maintained files under lint coverage.

> Generated by [Expert Code Review (on open)](https://github.com/microsoft/testfx/actions/runs/27002912109) for issue #8852 · sonnet46 1.6M

azat-msft (approved):

PR #8857, opened 2026-06-05
Add grade-tests-on-pr agentic workflow

Evangelink:
## Why

Reviewers want quick per-test feedback when a PR adds or modifies tests ("which of these new tests are weak?") without manually skimming the diff. Today this requires a human spot-check.

This adds an **agentic workflow** that posts a single PR comment grading each new/modified test method.

## What

- `.github/workflows/grade-tests-on-pr.md` — the gh-aw workflow source
- `.github/workflows/grade-tests-on-pr.lock.yml` — generated by `gh aw compile --strict` (gh-aw v0.76.1, matches the version used by other locks in this repo)
- `.github/workflows/scripts/grade-tests/extract-changed-tests.py` — stdlib-only Python that identifies test methods (via `[TestMethod]`, `[Fact]`, `[Theory]`, `[Test]`, `[TestCase]`, `[DataTestMethod]` attributes) whose source spans overlap diff hunks; emits TSV

## Behavior

- **Triggers**: `pull_request: [opened, synchronize, ready_for_review]` + `/grade-tests` slash command on existing PRs
- **Skips**: drafts, OneLocBuild PRs (`dotnet-bot` localization)
- **Short-circuits**: when the deterministic extractor finds no added/modified test methods, the agent never runs (no LLM cost, no comment)
- **Comment**: single comment via `safe-outputs: add-comment` with `hide-older-comments: true` for idempotent re-runs on `synchronize`
- **Concurrency**: `cancel-in-progress: true` per PR handles `synchronize` spam
- **Advisory only**: read-only tools, no edits, no failing checks, no push permissions

## Design

Adopted from a rubber-duck critique:

1. **Deterministic extraction in the workflow, not the LLM** — a 200-line Python script (stdlib only) walks each changed `*.cs` file under `test/**`, identifies test methods by attribute, and emits a TSV of `<file>\t<start>\t<end>\t<fq-name>` for methods whose span overlaps a diff hunk. Tested locally against recent testfx commits.
2. **Inline grading rubric** — the workflow body inlines a condensed rubric (3 sub-dimensions, letter+band, deduction rules) so the workflow is self-contained. The canonical, longer rubric lives in `dotnet/skills` as the `grade-tests` skill ([dotnet/skills#727](https://github.com/dotnet/skills/pull/727)) and they stay aligned.
3. **testfx-specific rules** are documented inline:
   - Tests under `test/UnitTests/TestFramework.UnitTests/` use the internal `TestFramework.ForTestingMSTest` framework -- its `Verify(...)` is equivalent to `Assert.IsTrue(...)`
   - Integration tests under `test/IntegrationTests/` may legitimately exceed 30 lines (size cap relaxed to ~120 lines for those)
   - FluentAssertions is preferred style and counts as equivalent to the MSTest API
   - Stylistic concerns (missing `init` accessors, license headers, etc.) are not flagged
4. **Comment caps**: summary first, then a table capped to ~50 rows, with the full list in a `<details>` block.

## Validation

- Workflow compiles cleanly: `gh aw compile --strict .github/workflows/grade-tests-on-pr.md` -> 0 errors, 0 warnings
- Extractor smoke-tested against `git diff` on the last 5 testfx commits -- correctly identifies test methods and emits FQ names + line spans

## Related

- `dotnet/skills#727` adds the canonical `grade-tests` skill


azat-msft (approved):

Evangelink:
The previous extractor used a fixed regex over attribute names: `[TestMethod]`, `[DataTestMethod]`, `[Fact]`, `[Theory]`, `[Test]`, `[TestCase]`. That **silently misses** every method using a derived test attribute, of which testfx has many:

| Pattern | Where |
|---|---|
| `[STATestMethod]` (and `STATestMethodAttribute : TestMethodAttribute`) | `src/TestFramework/TestFramework/Attributes/TestMethod/STATestMethodAttribute.cs` |
| `[UITestMethod]` (WinUI/UWP) | `src/TestFramework/TestFramework.Extensions/Attributes/*UITestMethodAttribute.cs` |
| `[DerivedSTATestMethod]` (two-levels deep) | `test/IntegrationTests/MSTest.Acceptance.IntegrationTests/STATestMethodTests.cs` |
| `[IterativeTestMethod]` | `test/IntegrationTests/TestAssets/FxExtensibilityTestProject/CustomTestExTests.cs` |
| `[MyTestMethod]` | `test/UnitTests/MSTest.SelfRealExamples.UnitTests/...` |

Algorithmically resolving the inheritance chain would require either a Roslyn syntax pass on the whole repo (~3000+ files, .NET SDK on PATH, global.json conflict with testfx's preview-SDK pin) or a brittle text-based alias map that gets fooled by class declarations inside `MSTest.Analyzers.UnitTests` string fixtures. Both add complexity.

## The fix

**Move the "is this a test?" decision to the agent.** The agent already has to read each method body to grade it (assertion strength, structure, anti-patterns) -- deciding whether the method is a test is essentially free at that point, and LLMs are very reliable at the judgment.

The pre-step is now a tiny inline `awk` + `git diff` block that emits:

```
test/UnitTests/Foo.Tests/BarTests.cs    12-25,40-67
test/IntegrationTests/Acceptance.IntegrationTests/QuxTests.cs   5-30
```

The agent walks each file, identifies methods overlapping the changed regions, and decides which ones are tests (with explicit guidance to grep the repo to verify inheritance chains when uncertain).

## Wins

- Catches **all** custom inherited attributes (no regex maintenance burden)
- Drops the standalone Python script (~200 lines → ~50 lines of bash, in-workflow)
- Inline `awk` is auditable in the workflow itself; no separate file to review
- Language-agnostic: same pre-step works for any `.cs` test file under `test/` (and is easy to extend to `.py`/`.java`/etc. for other repos)

## Smoke test

Tested locally against `HEAD~5..HEAD` on this branch -- correctly emits 2 changed test files with the right line ranges.

Issue #8828, opened 2026-06-04
Add JSON / NDJSON test report extension (Microsoft.Testing.Extensions.JsonReport)


Evangelink:
## Summary

Introduce a new MTP extension `Microsoft.Testing.Extensions.JsonReport` that emits a stable, documented NDJSON (newline-delimited JSON) event stream of test execution events, plus a final summary event. Models the existing `Microsoft.Testing.Extensions.TrxReport`, `Microsoft.Testing.Extensions.HtmlReport`, and `Microsoft.Testing.Extensions.AzureDevOpsReport` extensions.

## Motivation

Today MTP offers two machine-readable outputs:

- `--list-tests json` (test **discovery** only, added in #8280)
- TRX XML via `Microsoft.Testing.Extensions.TrxReport` (verbose, schema-heavy, hard to stream)

There is **no** structured output for **test execution results** that an agent or downstream tool can stream-parse incrementally. Modern test runners all expose this:

- **Jest** (`--json`): https://jestjs.io/docs/cli#--json
- **pytest-json-report**: https://pypi.org/project/pytest-json-report/
- **Vitest** (JSON reporter): https://vitest.dev/guide/cli.html#reporters
- **Bun test** (`--reporter=junit` and JSON variants)

For LLM/agent consumption, NDJSON is far superior to TRX:

- One event per line: agents can grep/tail without parsing
- Stable schema: agents can rely on field names
- Streamable: events flush as tests complete, enabling live reactions

## Proposed shape

Each line is a single JSON object. Event types:

```jsonl
{"event":"run_started","schemaVersion":1,"runId":"...","startTimeUtc":"...","framework":{"name":"MSTest","version":"3.x"}}
{"event":"test_started","testId":"...","testUid":"...","displayName":"...","fullyQualifiedName":"...","source":"..."}
{"event":"test_completed","testId":"...","outcome":"passed|failed|skipped|inconclusive","durationMs":12.3,"messages":[...],"errors":[{"type":"...","message":"...","stackTrace":"..."}],"output":{"stdout":"...","stderr":"..."},"location":{"file":"...","line":42}}
{"event":"run_summary","schemaVersion":1,"runId":"...","durationMs":1234,"counts":{"total":42,"passed":40,"failed":1,"skipped":1,"inconclusive":0},"endTimeUtc":"..."}
```

Per-test fields are optional and omitted when empty to reduce file size / token count.

## CLI surface

```text
[...]

Issue #8827, opened 2026-06-04
Emit compact end-of-run failure index for LLM/agent consumption


Evangelink:
## Summary

At the end of a `Microsoft.Testing.Platform` test run, emit a compact, sentinel-delimited footer that lists every failed test as a single line of `Test FQN | first-line error | file:line`. Default-on in LLM-detected environments only (see `LLMEnvironmentDetector`).

## Motivation

When a test run fails, an LLM agent currently has to re-read the entire terminal transcript to figure out what failed and where. The per-failure blocks are paragraph-style (message + expected/actual + stack), suitable for a human scanning visually but expensive in tokens when the agent only needs the structured summary to plan its next step.

A small machine-friendly footer at a stable position is the lowest-cost way to give the agent a "table of contents" of failures.

## Proposed shape

```
##[failures]
TestFx.UnitTests.FooTests.WhenBarThenBaz | Expected 42 but was 41 | src/Foo.cs:128
TestFx.UnitTests.FooTests.WhenBarThenQux | NullReferenceException: Object reference not set... | src/Foo.cs:142
##[/failures]
```

Properties:

- Sentinel-delimited (`##[failures]` ... `##[/failures]`) so an agent can extract the block with a 2-line regex without parsing the rest of the transcript.
- One line per failure. Fields separated by ` | `. Pipe-in-value escaped as `\|`.
- Test FQN comes from the test node display path.
- First-line error: first non-empty line of the failure message, truncated to ~200 chars (configurable).
- `file:line` extracted from the first user-code frame (after the filter from #8825 / related). Empty if not found.

## Behavior

- Default-on when `LLMEnvironmentDetector.IsLLMEnvironment()` returns `true`.
- Default-off otherwise, to keep the human terminal transcript stable.
- For v1 do not add a CLI option, to avoid `--help`/`--info` acceptance churn and the XLF round.
- If desired in v2, add `--failure-summary <auto|on|off>`.

## Implementation notes

- `TerminalTestReporter` already tracks per-assembly state and emits an `AppendTestRunSummary` at the end (see `TerminalTestReporter.Summary.cs`). The failure index should be appended right after that summary block.
- Today `TerminalTestReporter.TestCompleted` renders failed details and discards them. To build the index, capture a bounded list of `(displayName, firstErrorLine, firstUserCodeFileLine)` per failed test as we render it.
[...]

Issue #8826, opened 2026-06-04
Filter MTP/MSTest internal stack frames from terminal failure output


Evangelink:
## Summary

When MTP renders a failed-test stack trace, hide internal MTP/MSTest plumbing frames so the user (and any LLM agent reading the output) only sees frames that come from their own code or the dependencies they care about.

## Motivation

The current `TerminalTestReporter.Formatting.cs:FormatStackTrace` simply line-splits `exception.StackTrace` and prints every frame. Every failure dumps frames like:

```
   at Microsoft.Testing.Platform.Hosts.CommonTestHost.ExecuteRequestAsync(...)
   at Microsoft.Testing.Platform.MSTest.MSTestBridgedTestFramework.ExecuteRequestAsync(...)
   at Microsoft.VisualStudio.TestPlatform.MSTestAdapter.UnitTestRunner.RunSingleTest(...)
   …
```

These frames have **zero diagnostic value** when the user is debugging their own test — the relevant frames are the user's `TestMethod`, the assertion site, and the production code under test. MSTest's `Assert.*` helpers are already annotated with `[StackTraceHidden]`, so those frames are already invisible; the remaining noise is the MTP/adapter plumbing chain.

For LLM consumers this matters most: each hidden frame is ~80-120 tokens that is pure noise. A typical failed test today emits 8-15 frames where only 2-4 are user-relevant.

## Proposed design

Add a frame-filter at the **rendering layer** (not in `ExceptionFlattener`):

- In `TerminalTestReporter.Formatting.cs:FormatStackTrace`, before calling `AppendStackFrame`, run each parsed frame through a predicate that returns `false` for frames whose declaring type matches one of the known internal namespaces.
- Internal namespace prefixes to hide:
  - `Microsoft.Testing.Platform.` (excluding `Microsoft.Testing.Platform.Extensions.*` — extension authors' own code? TBD)
  - `Microsoft.Testing.Extensions.` (TrxReport, HtmlReport, etc. — plumbing only)
  - `Microsoft.VisualStudio.TestPlatform.MSTestAdapter.` (VSTest adapter plumbing)
  - Possibly `Microsoft.VisualStudio.TestTools.UnitTesting.` for any non-`[StackTraceHidden]` member.
- Collapse contiguous hidden frames into a single `...skipped N frames in Microsoft.Testing.Platform.* ...` line so the user can tell something was hidden (and so the line count still indicates real depth).
[...]

Issue #8825, opened 2026-06-04
Honor NO_COLOR environment variable in MTP terminal output


Evangelink:
## Summary

Honor the [`NO_COLOR`](https://no-color.org) environment variable in the Microsoft.Testing.Platform terminal output. Today MTP only suppresses ANSI escape sequences via `--no-ansi`, `--ansi off`, or the `LLMEnvironmentDetector` (see `src/Platform/Microsoft.Testing.Platform/OutputDevice/TerminalOutputDevice.cs:147-184`); it does not look at `NO_COLOR` at all.

## Motivation

- [`NO_COLOR`](https://no-color.org) is the de-facto cross-tool convention for "I am piped/redirected/aren't a fan of color, please don't emit ANSI." Tools that honor it include `git`, `ripgrep`, `bat`, `gh`, `cargo`, the .NET SDK itself, modern Node.js CLIs, and most coding-agent runtimes (Claude Code, Cursor, Codex, Aider).
- Coding agents and CI systems commonly set `NO_COLOR=1` rather than the MTP-specific `--no-ansi` flag. Today they get ANSI noise in failure output, which costs tokens and breaks log scraping.
- The fix is small and matches existing precedence: `--ansi on|off|auto` still wins, then `--no-ansi`, then `NO_COLOR`, then `LLMEnvironmentDetector`, then CI detection.

## Proposed behavior

In `TerminalOutputDevice.cs` (resolution of `AnsiMode`):

```text
explicit --ansi on             -> AnsiMode.ForceAnsi
explicit --ansi off             -> AnsiMode.NoAnsi
--no-ansi (no --ansi override) -> AnsiMode.NoAnsi
NO_COLOR env var is non-empty   -> AnsiMode.NoAnsi  (NEW)
LLMEnvironmentDetector matches  -> AnsiMode.NoAnsi
in CI                           -> AnsiMode.SimpleAnsi
default                         -> AnsiMode.AnsiIfPossible
```

Per the [NO_COLOR spec](https://no-color.org): _"All command-line software which outputs text with ANSI color added should check for the presence of a NO_COLOR environment variable that, when present (regardless of its value), prevents the addition of ANSI color."_

We treat *any non-empty* `NO_COLOR` value as enabling the behavior, matching the spec.

## Notes

[...]

PR #8859, opened 2026-06-05
Use native GitHub Issue Type field for Bug/Feature/Task instead of type/* labels

Evangelink:
## Problem

Newly created issues are almost always tagged with the legacy `type/feature` label (and sometimes `type/bug`) while the native GitHub **Issue Type** field stays empty. Example: #8828 has `type/feature` but `issueType: null`. The native types (Bug, Feature, Task) were enabled on the `microsoft` org but never wired in.

## Changes

* **`.github/copilot-instructions.md`** - new section _"GitHub issue creation guidelines"_:
  * Use the native Issue Type field (Bug / Feature / Task) instead of `type/bug` / `type/feature` labels.
  * Other `type/*` labels (`type/tech-debt`, `type/test-gap`, `type/automation`, `type/regression`, `type/flaky-test`, ...) stay because they have no native equivalent.
  * How to set the Issue Type from issue templates (frontmatter), the web UI, `gh issue create`, the GraphQL `updateIssueIssueType` mutation, and `gh-aw` `safe-outputs.create-issue.type`.
* **`eng/migrate-labels.ps1`**:
  * Removed `type/bug` and `type/feature` from the target taxonomy.
  * Mapped the legacy `bug`, `enhancement`, `feature-request`, `type/bug`, `type/feature` labels to deletion in `$Migration`.
  * Added `$IssueTypeReplacement` and helper functions (`Get-IssueTypeId`, `Get-IssuesWithLabel`, `Set-IssueType`, `Convert-LabelToIssueType`).
  * Added **Phase 3.5** that converts every issue carrying one of those labels to the matching native Issue Type **before** Phase 4 deletes the empty label, so we do not lose categorization.

## Verification

Dry-run (`pwsh -NoProfile -File .\eng\migrate-labels.ps1 -DryRun -Force`) shows Phase 3.5 picks up the 5 open `type/feature` issues (#8793, #8825, #8826, #8827, #8828) and would set them to native `Feature`, then Phase 4 deletes the now-empty `type/bug` and `type/feature` labels. PRs lose the label without any Issue Type change (PRs do not have one).

After this PR merges, run the script once without `-DryRun` to perform the actual conversion + deletion.

### DIFF
--- a/eng/migrate-labels.ps1
+++ b/eng/migrate-labels.ps1
@@ -454,9 +454,12 @@ function Get-IssueTypeId {
     param([string] $Name)
 
     if ($null -eq $script:IssueTypeIdCache) {
-        $owner, $name = $Repo -split '/', 2
+        # NOTE: Use distinct local variable names ($repoOwner/$repoName) to avoid
+        # shadowing the $Name parameter - PowerShell variable names are
+        # case-insensitive, so `$name` and `$Name` are the same variable.
+        $repoOwner, $repoName = $Repo -split '/', 2
         $query = 'query($owner:String!,$name:String!){ repository(owner:$owner,name:$name){ issueTypes(first:50){ nodes{ id name } } } }'
-        $json = & gh api graphql -f query=$query -F owner=$owner -F name=$name
+        $json = & gh api graphql -f query=$query -F owner=$repoOwner -F name=$repoName
         if ($LASTEXITCODE -ne 0) {
             throw "Failed to fetch issue types for $Repo (exit $LASTEXITCODE)"
         }
@@ -478,7 +481,7 @@ function Get-IssueTypeId {
 function Get-IssuesWithLabel {
     param([string] $Label)
 
-    $owner, $name = $Repo -split '/', 2
+    $repoOwner, $repoName = $Repo -split '/', 2
     $query = @'
 query($owner:String!,$name:String!,$label:String!,$cursor:String){
   repository(owner:$owner,name:$name){
@@ -490,21 +493,20 @@ query($owner:String!,$name:String!,$label:String!,$cursor:String){
 }
 '@
 
-    $results = @()
+    $results = [System.Collections.Generic.List[object]]::new()
     $cursor = $null
     do {
-        if ($cursor) {
-            $json = & gh api graphql -f query=$query -F owner=$owner -F name=$name -F label=$Label -F cursor=$cursor
-        } else {
-            $json = & gh api graphql -f query=$query -F owner=$owner -F name=$name -F label=$Label
-        }
+        $ghArgs = @('api', 'graphql', '-f', "query=$query",
+                    '-F', "owner=$repoOwner", '-F', "name=$repoName", '-F', "label=$Label")
+        if ($cursor) { $ghArgs += @('-F', "cursor=$cursor") }
+        $json = & gh @ghArgs
         if ($LASTEXITCODE -ne 0) { throw "Failed to query issues with label '$Label' (exit $LASTEXITCODE)" }
         $page = ($json | ConvertFrom-Json).data.repository.issues
-        foreach ($n in $page.nodes) { $results += $n }
+        if ($page.nodes) { $results.AddRange([object[]]$page.nodes) }
         $cursor = if ($page.pageInfo.hasNextPage) { $page.pageInfo.endCursor } else { $null }
     } while ($cursor)
 
-    return $results
+    return $results.ToArray()
 }
 
 function Set-IssueType {
@@ -701,7 +703,7 @@ foreach ($label in @($IssueTypeReplacement.Keys)) {
         Write-Host "  = '$label' not present - skipping" -ForegroundColor DarkGray
         continue
     }
-    Write-Host "  c converting '$label' -> issue type '$typeName'" -ForegroundColor Magenta
+    Write-Host "  > converting '$label' -> issue type '$typeName'" -ForegroundColor Magenta
     Convert-LabelToIssueType -Label $label -IssueTypeName $typeName
 }
 
--- a/eng/migrate-labels.ps1
+++ b/eng/migrate-labels.ps1
@@ -542,25 +542,46 @@ function Convert-LabelToIssueType {
     $issues = Get-IssuesWithLabel -Label $Label
     Write-Host "    found $($issues.Count) issue(s) with '$Label' (will set issue type '$IssueTypeName' where unset)" -ForegroundColor DarkGray
 
+    $failures = @()
     foreach ($issue in $issues) {
-        if ($issue.issueType -and $issue.issueType.name) {
-            Write-Host "      = #$($issue.number) already has issue type '$($issue.issueType.name)' - keeping it" -ForegroundColor DarkGray
-        } else {
-            Write-Host "      + #$($issue.number) -> issue type '$IssueTypeName'" -ForegroundColor Green
-            Set-IssueType -IssueNodeId $issue.id -IssueTypeName $IssueTypeName
+        try {
+            if ($issue.issueType -and $issue.issueType.name) {
+                Write-Host "      = #$($issue.number) already has issue type '$($issue.issueType.name)' - keeping it" -ForegroundColor DarkGray
+            } else {
+                Write-Host "      + #$($issue.number) -> issue type '$IssueTypeName'" -ForegroundColor Green
+                Set-IssueType -IssueNodeId $issue.id -IssueTypeName $IssueTypeName
+            }
+            Invoke-Gh issue edit $issue.number --repo $Repo --remove-label $Label | Out-Null
         }
-        Invoke-Gh issue edit $issue.number --repo $Repo --remove-label $Label | Out-Null
+        catch {
+            $failures += [pscustomobject]@{ Kind = 'issue'; Number = $issue.number; Error = $_.Exception.Message }
+            Write-Host "      ! failed to convert issue #$($issue.number): $($_.Exception.Message)" -ForegroundColor Red
+        }
     }
 
     # PRs also carry these labels (no Issue Type on PRs - just strip the label).
     $prJson = & gh pr list --repo $Repo --label $Label --state all --limit 5000 --json number
     if ($LASTEXITCODE -ne 0) { throw "Failed to list PRs with label '$Label'" }
     $prs = @($prJson | ConvertFrom-Json | ForEach-Object { $_.number })
+    if ($prs.Count -ge 5000) {
+        throw "Label '$Label' has >=5000 PRs - pagination support required, refusing to proceed."
+    }
     if ($prs.Count -gt 0) {
         Write-Host "    found $($prs.Count) PR(s) with '$Label' - stripping label (no issue type on PRs)" -ForegroundColor DarkGray
         foreach ($n in $prs) {
-            Invoke-Gh pr edit $n --repo $Repo --remove-label $Label | Out-Null
+            try {
+                Invoke-Gh pr edit $n --repo $Repo --remove-label $Label | Out-Null
+            }
+            catch {
+                $failures += [pscustomobject]@{ Kind = 'pr'; Number = $n; Error = $_.Exception.Message }
+                Write-Host "      ! failed to strip label from PR #${n}: $($_.Exception.Message)" -ForegroundColor Red
+            }
         }
+    }
+
+    if ($failures.Count -gt 0) {
+        $summary = ($failures | ForEach-Object { "$($_.Kind) #$($_.Number): $($_.Error)" }) -join "`n  "
+        throw "$($failures.Count) item(s) failed while converting '$Label' to issue type '$IssueTypeName':`n  $summary"
     }
 }
 
--- a/eng/migrate-labels.ps1
+++ b/eng/migrate-labels.ps1
@@ -547,14 +547,7 @@ function Convert-LabelToIssueType {
             Write-Host "      + #$($issue.number) -> issue type '$IssueTypeName'" -ForegroundColor Green
             Set-IssueType -IssueNodeId $issue.id -IssueTypeName $IssueTypeName
         }
-        if (-not $DryRun) {
-            & gh issue edit $issue.number --repo $Repo --remove-label $Label | Out-Null
-            if ($LASTEXITCODE -ne 0) {
-                throw "Failed to remove label '$Label' from issue #$($issue.number) (exit $LASTEXITCODE)"
-            }
-        } else {
-            Write-Host "    [dry-run] would remove label '$Label' from #$($issue.number)" -ForegroundColor DarkGray
-        }
+        Invoke-Gh issue edit $issue.number --repo $Repo --remove-label $Label | Out-Null
     }
 
     # PRs also carry these labels (no Issue Type on PRs - just strip the label).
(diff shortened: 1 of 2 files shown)


eng/migrate-labels.ps1
```diff
@@ -411,6 +444,131 @@ function Merge-LabelInto {
...
+            Write-Host "      = #$($issue.number) already has issue type '$($issue.issueType.name)' - keeping it" -ForegroundColor DarkGray
+        } else {
+            Write-Host "      + #$($issue.number) -> issue type '$IssueTypeName'" -ForegroundColor Green
+            Set-IssueType -IssueNodeId $issue.id -IssueTypeName $IssueTypeName
+        }
+        if (-not $DryRun) {
+            & gh issue edit $issue.number --repo $Repo --remove-label $Label | Out-Null
+            if ($LASTEXITCODE -ne 0) {
+                throw "Failed to remove label '$Label' from issue #$($issue.number) (exit $LASTEXITCODE)"
+            }
```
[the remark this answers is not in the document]
Evangelink: Good catch - fixed in bd66b36. The issue 'remove-label' call in Convert-LabelToIssueType now goes through Invoke-Gh like the rest of the script (and like the PR-edit path right below it), so dry-run logging and exit-code handling are consistent.

eng/migrate-labels.ps1
```diff
@@ -411,6 +444,131 @@ function Merge-LabelInto {
...
+            throw "Failed to fetch issue types for $Repo (exit $LASTEXITCODE)"
+        }
+        $data = $json | ConvertFrom-Json
+        $script:IssueTypeIdCache = @{}
+        foreach ($t in $data.data.repository.issueTypes.nodes) {
+            $script:IssueTypeIdCache[$t.name] = $t.id
+        }
+    }
+
+    if (-not $script:IssueTypeIdCache.ContainsKey($Name)) {
```
Evangelink: **Algorithmic Correctness — ISSUE**
**SEVERITY: BLOCKING**
**LINES: 457–473**
**SCENARIO:** Any non-dry-run execution; specifically the *first* call to `Get-IssueTypeId` (e.g. when the first issue without an existing type is processed).

**FINDING:** PowerShell variable names are case-insensitive: `$name` and `$Name` are the same variable. Line 457 destructures:
```powershell
$owner, $name = $Repo -split '/', 2   # overwrites $Name with "testfx"
```
This overwrites the `$Name` parameter (which holds the desired type name, e.g. `"Bug"`) with the repo's short name (e.g. `"testfx"`). Lines 470–473 then look up `$Name` in the cache — but `$Name` now holds `"testfx"` instead of `"Bug"`, so `ContainsKey` returns false and the function always throws on its first invocation:
> `Issue type 'testfx' is not configured on microsoft/testfx. Available: Bug, Feature, Task`

Subsequent calls (cache already warm, `if` block skipped) work correctly, but the script never reaches them.

**RECOMMENDATION:** Use a distinct local variable name that cannot collide with the parameter:
```powershell
$repoOwner, $repoName = $Repo -split '/', 2
$json = & gh api graphql -f query=$query -F owner=$repoOwner -F name=$repoName
```
Apply the same rename in `Get-IssuesWithLabel` line 481 (`$owner, $name`) for consistency, even though there is no parameter clash there (the parameter is `$Label`).

Evangelink (commented):
**Algorithmic Correctness — ISSUE (1 BLOCKING)**

One bug found; everything else (dry-run guards, pagination, `$existing` lifetime) is correct.

| | |
|---|---|
| **Severity** | BLOCKING |
| **File** | `eng/migrate-labels.ps1` |
| **Lines** | 457–473 |

**`$name` variable shadows the `$Name` parameter in `Get-IssueTypeId`**

PowerShell variable names are **case-insensitive**: `$name` and `$Name` refer to the same variable. On the first call (cache is `$null`), line 457 executes:

```powershell
$owner, $name = $Repo -split '/', 2   # silently overwrites $Name
```

This replaces the caller-supplied type name (e.g. `"Bug"`) with the repository's short name (e.g. `"testfx"`). Lines 470–473 then consult `$Name` for the cache lookup — using `"testfx"` instead of `"Bug"` — and always throw:

> `Issue type 'testfx' is not configured on microsoft/testfx. Available: Bug, Feature, Task`

Subsequent calls (warm cache, `if` block skipped) are fine, but the script never reaches them.

**Fix:** use a non-clashing local name, e.g.:
```powershell
$repoOwner, $repoName = $Repo -split '/', 2
$json = & gh api graphql -f query=$query -F owner=$repoOwner -F name=$repoName
```
Apply the same rename at line 481 in `Get-IssuesWithLabel` for consistency (no parameter clash there, but it avoids confusion).

> Generated by [Expert Code Review (on open)](https://github.com/microsoft/testfx/actions/runs/27018956189) for issue #8859 · sonnet46 6.8M

eng/migrate-labels.ps1
```diff
@@ -411,6 +444,131 @@ function Merge-LabelInto {
...
+# Issue-type cache (one query per repo). Maps type name (e.g. 'Bug') to its
+# global node ID for the GraphQL updateIssueIssueType mutation.
+# ---------------------------------------------------------------------------
+$script:IssueTypeIdCache = $null
+
+function Get-IssueTypeId {
+    param([string] $Name)
+
+    if ($null -eq $script:IssueTypeIdCache) {
+        $owner, $name = $Repo -split '/', 2
```
Evangelink: **🔴 BLOCKING — Parameter clobbered by case-insensitive variable shadowing**

PowerShell variable names are case-insensitive, so `$name` and `$Name` are the **same** variable. The destructuring assignment on this line:

```powershell
$owner, $name = $Repo -split '/', 2
```

overwrites the `$Name` function parameter (e.g. `"Bug"`) with the repository short-name component (e.g. `"testfx"`). Lines 470–473 then call `ContainsKey($Name)` and `$script:IssueTypeIdCache[$Name]` searching for `"testfx"` instead of `"Bug"`, so **every first call throws**:

> Issue type 'testfx' is not configured on microsoft/testfx. Available: Bug, Feature, Task

Subsequent calls skip the `if`-block (cache is warm) and work fine — but the script aborts on the very first invocation.

**Fix:**
```powershell
$repoOwner, $repoName = $Repo -split '/', 2
$json = & gh api graphql -f query=$query -F owner=$repoOwner -F name=$repoName
```

eng/migrate-labels.ps1
```diff
@@ -411,6 +444,131 @@ function Merge-LabelInto {
...
+}
+'@
+
+    $results = @()
+    $cursor = $null
+    do {
+        if ($cursor) {
+            $json = & gh api graphql -f query=$query -F owner=$owner -F name=$name -F label=$Label -F cursor=$cursor
+        } else {
+            $json = & gh api graphql -f query=$query -F owner=$owner -F name=$name -F label=$Label
```
Evangelink: **MODERATE — DRY violation: two call sites differ only by `$cursor`**

The two `gh api graphql` branches (lines 497 and 499) are identical except for the optional `-F cursor=$cursor` argument. Any future change to the argument list must be applied twice. Consolidate:

```powershell
$ghArgs = @('api', 'graphql', '-f', "query=$query",
            '-F', "owner=$owner", '-F', "name=$name", '-F', "label=$Label")
if ($cursor) { $ghArgs += '-F', "cursor=$cursor" }
$json = & gh `@ghArgs`
```

eng/migrate-labels.ps1
```diff
@@ -531,6 +689,29 @@ foreach ($name in @($Targets.Keys)) {
...
+Write-Host "=== Phase 3.5: convert deprecated labels to native Issue Types ===" -ForegroundColor Cyan
+if (-not $DryRun) { $existing = Get-AllLabels }
+foreach ($label in @($IssueTypeReplacement.Keys)) {
+    $typeName = $IssueTypeReplacement[$label]
+    $labelObj = Get-LabelByName -Labels $existing -Name $label
+    if (-not $labelObj) {
+        Write-Host "  = '$label' not present - skipping" -ForegroundColor DarkGray
+        continue
+    }
+    Write-Host "  c converting '$label' -> issue type '$typeName'" -ForegroundColor Magenta
```
Evangelink: **NIT — Log prefix `"  c "` breaks the established symbol-based legend**

Every other per-item log prefix in the script is a single punctuation symbol (`= `, `+ `, `- `, `! `). Using the letter `c` here diverges from that visual grammar and is harder to grep. Consider `>` (action/transform in progress) or simply reuse `+` since the message text already carries the verb:

```powershell
Write-Host "  > converting '$label' -> issue type '$typeName'" -ForegroundColor Magenta
```

eng/migrate-labels.ps1
```diff
@@ -411,6 +444,131 @@ function Merge-LabelInto {
...
+            Set-IssueType -IssueNodeId $issue.id -IssueTypeName $IssueTypeName
+        }
+        if (-not $DryRun) {
+            & gh issue edit $issue.number --repo $Repo --remove-label $Label | Out-Null
+            if ($LASTEXITCODE -ne 0) {
+                throw "Failed to remove label '$Label' from issue #$($issue.number) (exit $LASTEXITCODE)"
+            }
+        } else {
+            Write-Host "    [dry-run] would remove label '$Label' from #$($issue.number)" -ForegroundColor DarkGray
+        }
```
Evangelink: **MODERATE — Inline dry-run guard duplicates what `Invoke-Gh` already provides**

Issue label removal (lines 550–557) hand-rolls its own `$DryRun` check and `$LASTEXITCODE` assert using the raw `& gh` operator, while PR label removal 10 lines below (line 567) delegates cleanly to `Invoke-Gh`. The existing `Invoke-Gh issue edit` path already exists elsewhere in the script (e.g. line 415). Replace the entire `if (-not $DryRun) { ... } else { ... }` block with:

```powershell
Invoke-Gh issue edit $issue.number --repo $Repo --remove-label $Label | Out-Null
```

`Invoke-Gh` encapsulates the dry-run log message, the actual execution, and the exit-code check.

eng/migrate-labels.ps1
```diff
@@ -411,6 +444,131 @@ function Merge-LabelInto {
...
+    $cursor = $null
+    do {
+        if ($cursor) {
+            $json = & gh api graphql -f query=$query -F owner=$owner -F name=$name -F label=$Label -F cursor=$cursor
+        } else {
+            $json = & gh api graphql -f query=$query -F owner=$owner -F name=$name -F label=$Label
+        }
+        if ($LASTEXITCODE -ne 0) { throw "Failed to query issues with label '$Label' (exit $LASTEXITCODE)" }
+        $page = ($json | ConvertFrom-Json).data.repository.issues
+        foreach ($n in $page.nodes) { $results += $n }
```
Evangelink: **MODERATE — O(n2) array concatenation inside pagination loop**

`$results += $n` reallocates and copies the entire array on every element. For a repo with N issues across P pages of 100, this performs O(N2) copies. Accumulate whole-page arrays instead (O(P) reallocations):

```powershell
foreach ($n in $page.nodes) { $results += $n }
# → replace with:
$results += @($page.nodes)
```

Or use a `List<object>` for O(1) appends:
```powershell
$results = [System.Collections.Generic.List[object]]::new()
# ...
$results.AddRange($page.nodes)
```

eng/migrate-labels.ps1
Evangelink: Fixed in 7fe6287c0 - renamed the destructured locals to $repoOwner/$repoName in Get-IssueTypeId so they no longer shadow the $Name parameter. Verified end-to-end via -DryRun -Force: the issue-type cache now warms correctly and 'type/feature' -> 'Feature' resolves on the first call.

eng/migrate-labels.ps1
Evangelink: Fixed in 7fe6287c0 (same change as the line 470 thread) - the destructured locals in Get-IssueTypeId are now $repoOwner/$repoName, so the $Name parameter is preserved.

eng/migrate-labels.ps1
Evangelink: Fixed in 7fe6287c0 - the two graphql call sites in Get-IssuesWithLabel are now collapsed into a single $ghArgs splat with the cursor appended conditionally.

eng/migrate-labels.ps1
Evangelink: Fixed in 7fe6287c0 - swapped 'c ' for '> ' so the phase-3 conversion line matches the symbol-based legend used elsewhere.

eng/migrate-labels.ps1
Evangelink: Fixed in 7fe6287c0 - Get-IssuesWithLabel now accumulates into a System.Collections.Generic.List[object] (AddRange per page) and returns ToArray(), eliminating the O(n^2) array copy in the pagination loop.

eng/migrate-labels.ps1
Evangelink: Already addressed by bd66b36 (which replaced the inline `& gh` + `\0` block with a single `Invoke-Gh issue edit ... --remove-label` call, mirroring the PR-edit path right below it).

eng/migrate-labels.ps1
```diff
@@ -411,6 +444,126 @@ function Merge-LabelInto {
...
+
+    foreach ($issue in $issues) {
+        if ($issue.issueType -and $issue.issueType.name) {
+            Write-Host "      = #$($issue.number) already has issue type '$($issue.issueType.name)' - keeping it" -ForegroundColor DarkGray
+        } else {
+            Write-Host "      + #$($issue.number) -> issue type '$IssueTypeName'" -ForegroundColor Green
+            Set-IssueType -IssueNodeId $issue.id -IssueTypeName $IssueTypeName
+        }
+        Invoke-Gh issue edit $issue.number --repo $Repo --remove-label $Label | Out-Null
+    }
```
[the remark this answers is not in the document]
Evangelink: Fixed in a50be31f6 - Convert-LabelToIssueType now mirrors Merge-LabelInto's failure-aggregation pattern: each issue and each PR edit is wrapped in try/catch, failures are collected into a per-call list with a red `!` log line, the loop keeps going through the rest of the items, and a single aggregated summary throw fires at the end. Phase 3 still aborts overall on any failure, so Phase 4 deletions never run against a half-migrated state.

eng/migrate-labels.ps1
```diff
@@ -411,6 +444,126 @@ function Merge-LabelInto {
...
+
+    # PRs also carry these labels (no Issue Type on PRs - just strip the label).
+    $prJson = & gh pr list --repo $Repo --label $Label --state all --limit 5000 --json number
+    if ($LASTEXITCODE -ne 0) { throw "Failed to list PRs with label '$Label'" }
+    $prs = @($prJson | ConvertFrom-Json | ForEach-Object { $_.number })
+    if ($prs.Count -gt 0) {
+        Write-Host "    found $($prs.Count) PR(s) with '$Label' - stripping label (no issue type on PRs)" -ForegroundColor DarkGray
```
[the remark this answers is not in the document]
Evangelink: Fixed in a50be31f6 - added the same `>=5000` truncation guard as Merge-LabelInto on the PR-list call, so the script throws instead of silently dropping the tail. (The issue side already paginates via GraphQL cursors in Get-IssuesWithLabel, so no change needed there.)

azat-msft (approved):
I'm not proficient with powershell scripts by any means, but I reviewed the logic anyways and it seems correct to me. Also, dry run output exactly what it should based on the verification steps, so I approve.

PR #8861, opened 2026-06-05
[perf-improver] perf: fast path in HumanReadableDurationFormatter.Render for sub-hour durations

Evangelink:
🤖 *This is an automated contribution from [Perf Improver](https://github.com/microsoft/testfx/actions/runs/27020184760).*

## Goal and Rationale

`HumanReadableDurationFormatter.Render` is called multiple times per terminal progress-frame render tick — once per visible test-worker line for the "duration unchanged" fast-check, and once more inside `AppendTestWorkerProgress`/`AppendTestWorkerDetail` when the line needs a full re-render. For a 4-assembly run refreshing at ~5 fps, this adds up to **~40+ calls per second** over the lifetime of the test run.

Each call currently allocates:
1. A `new StringBuilder()`
2. One or two intermediate strings via `GetFormattedPart` (e.g. `"5s"`, `"59s"`, `" 05s"`)
3. The final `stringBuilder.ToString()` result

That is **3–4 heap allocations per call** — all for a tiny string like `"(5s)"` or `"(2m 30s)"`.

## Approach

On .NET 8+, use `string.Create(IFormatProvider, Span<char>, ref DefaultInterpolatedStringHandler)` with a `stackalloc` buffer. This overload uses the span as a scratch buffer and produces the final heap string in a single allocation — no `StringBuilder`, no intermediate `GetFormattedPart` strings.

The fast path activates when:
- `Days == 0 && Hours == 0` (covers virtually all test runs)
- `showMilliseconds == false` (the default for all progress-frame callers)

Both conditions are true for every caller in `AnsiTerminalTestProgressFrame` and `SimpleTerminalBase`.

The slow path (days, hours, or `showMilliseconds=true`) is unchanged; it is rarely reached and is not in the render hot path.

## Performance Evidence

| Scenario | Before | After |
|---|---|---|
| Allocations per `Render` call (typical: < 1 min) | 3–4 (StringBuilder + 1–2 GetFormattedPart strings + result) | **1** (result only) |
| Allocations per `Render` call (> 1 hour) | 3–4 | 3–4 (unchanged, slow path) |
| `string.Create` heap allocations | n/a | 1 (final string only; scratch buffer is on stack) |

**Methodology**: code inspection + allocation analysis. `HumanReadableDurationFormatter.Render` is called ~5× per render frame; at 5 fps over a 5-minute run that is ~7 500 calls, saving ~15 000–22 500 small string allocations.

The change is `#if NET8_0_OR_GREATER`-guarded, so netstandard2.0 behaviour is completely unchanged.

## Trade-offs

- The nested ternary (`wrapInParentheses ? (minutes == 0 ? ... : ...) : (minutes == 0 ? ... : ...)`) is slightly dense but self-contained. The logic is simple and the four resulting strings are easy to validate visually.
- No behaviour change for `showMilliseconds=true`, durations with hours, or durations with days.
- `netstandard2.0` uses the existing slow path as before.

## Test Status

- `Microsoft.Testing.Platform.UnitTests` (net8.0): **1086 passed, 0 failed, 3 skipped** ✅
- Build (all TFMs: net8.0, net9.0, netstandard2.0): **0 warnings, 0 errors** ✅

## Reproducibility

```bash
./build.sh
artifacts/bin/Microsoft.Testing.Platform.UnitTests/Debug/net8.0/Microsoft.Testing.Platform.UnitTests
```

> Generated by [Perf Improver](https://github.com/microsoft/testfx/actions/runs/27020184760)

> Generated by [Perf Improver](https://github.com/microsoft/testfx/actions/runs/27020184760) · sonnet46 8.8M · [◷](https://github.com/search?q=repo%3Amicrosoft%2Ftestfx+%22gh-aw-workflow-id%3A+perf-improver%22&type=pullrequests)
>
<details>

```
gh aw add githubnext/agentics/workflows/perf-improver.md@main
```
</details>

### DIFF
--- a/src/Platform/Microsoft.Testing.Platform/OutputDevice/Terminal/HumanReadableDurationFormatter.cs
+++ b/src/Platform/Microsoft.Testing.Platform/OutputDevice/Terminal/HumanReadableDurationFormatter.cs
@@ -63,21 +63,18 @@ internal static class HumanReadableDurationFormatter
         }
 
 #if NET8_0_OR_GREATER
-        // Fast path for the common case: duration < 1 hour with no milliseconds.
-        // string.Create with a stackalloc buffer avoids both the StringBuilder allocation and
-        // the intermediate strings that GetFormattedPart would otherwise produce per component.
-        // All progress-frame callers use the defaults (wrapInParentheses=true, showMilliseconds=false).
+        // Fast path for the common progress-frame case: duration < 1 hour with no milliseconds.
         if (!showMilliseconds && duration.Value.Days == 0 && duration.Value.Hours == 0)
         {
             int seconds = duration.Value.Seconds;
             int minutes = duration.Value.Minutes;
             return wrapInParentheses
                 ? (minutes == 0
-                    ? string.Create(CultureInfo.InvariantCulture, stackalloc char[8], $"({seconds}s)")
-                    : string.Create(CultureInfo.InvariantCulture, stackalloc char[12], $"({minutes}m {seconds:D2}s)"))
+                    ? $"({seconds}s)"
+                    : $"({minutes}m {seconds:D2}s)")
                 : (minutes == 0
-                    ? string.Create(CultureInfo.InvariantCulture, stackalloc char[6], $"{seconds}s")
-                    : string.Create(CultureInfo.InvariantCulture, stackalloc char[10], $"{minutes}m {seconds:D2}s"));
+                    ? $"{seconds}s"
+                    : $"{minutes}m {seconds:D2}s");
         }
 #endif
 
--- a/src/Platform/Microsoft.Testing.Platform/OutputDevice/Terminal/HumanReadableDurationFormatter.cs
+++ b/src/Platform/Microsoft.Testing.Platform/OutputDevice/Terminal/HumanReadableDurationFormatter.cs
@@ -62,6 +62,25 @@ internal static class HumanReadableDurationFormatter
             return string.Empty;
         }
 
+#if NET8_0_OR_GREATER
+        // Fast path for the common case: duration < 1 hour with no milliseconds.
+        // string.Create with a stackalloc buffer avoids both the StringBuilder allocation and
+        // the intermediate strings that GetFormattedPart would otherwise produce per component.
+        // All progress-frame callers use the defaults (wrapInParentheses=true, showMilliseconds=false).
+        if (!showMilliseconds && duration.Value.Days == 0 && duration.Value.Hours == 0)
+        {
+            int seconds = duration.Value.Seconds;
+            int minutes = duration.Value.Minutes;
+            return wrapInParentheses
+                ? (minutes == 0
+                    ? string.Create(CultureInfo.InvariantCulture, stackalloc char[8], $"({seconds}s)")
+                    : string.Create(CultureInfo.InvariantCulture, stackalloc char[12], $"({minutes}m {seconds:D2}s)"))
+                : (minutes == 0
+                    ? string.Create(CultureInfo.InvariantCulture, stackalloc char[6], $"{seconds}s")
+                    : string.Create(CultureInfo.InvariantCulture, stackalloc char[10], $"{minutes}m {seconds:D2}s"));
+        }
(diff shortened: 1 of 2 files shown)


azat-msft (approved):

src/Platform/Microsoft.Testing.Platform/OutputDevice/Terminal/HumanReadableDurationFormatter.cs
```diff
@@ -62,6 +62,25 @@ public static string Render(TimeSpan? duration, bool wrapInParentheses = true, b
...
+        // string.Create with a stackalloc buffer avoids both the StringBuilder allocation and
+        // the intermediate strings that GetFormattedPart would otherwise produce per component.
+        // All progress-frame callers use the defaults (wrapInParentheses=true, showMilliseconds=false).
+        if (!showMilliseconds && duration.Value.Days == 0 && duration.Value.Hours == 0)
+        {
+            int seconds = duration.Value.Seconds;
+            int minutes = duration.Value.Minutes;
+            return wrapInParentheses
+                ? (minutes == 0
+                    ? string.Create(CultureInfo.InvariantCulture, stackalloc char[8], $"({seconds}s)")
```
Youssef1313: I don't think this is going to be significantly faster compared to string interpolation. This also seem to add complications without any proven gains.

Evangelink: Good catch — simplified this in 1748e4171 by removing the string.Create/stackalloc path and using straightforward interpolation for the sub-hour fast path.

Evangelink: Good catch — simplified the fast path in 8532ea5, kept negative sub-hour durations on the slow path for parity, and added HumanReadableDurationFormatterTests coverage.

src/Platform/Microsoft.Testing.Platform/OutputDevice/Terminal/HumanReadableDurationFormatter.cs
```diff
@@ -62,6 +62,23 @@ public static string Render(TimeSpan? duration, bool wrapInParentheses = true, b
...
+
+            return (wrapInParentheses, minutes) switch
+            {
+                (true, 0) => $"({seconds}s)",
+                (true, _) => $"({minutes}m {seconds:D2}s)",
+                (false, 0) => $"{seconds}s",
+                _ => $"{minutes}m {seconds:D2}s",
+            };
+        }
+#endif
```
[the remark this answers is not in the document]
Evangelink: Good catch — fixed in 1b25ddc. Each fast-path arm is now wrapped with `string.Create(CultureInfo.InvariantCulture, $"...")` so the NET8 fast path matches the slow path's invariant numeric formatting in locales with non-ASCII digits.

test/UnitTests/Microsoft.Testing.Platform.UnitTests/OutputDevice/Terminal/HumanReadableDurationFormatterTests.cs
```diff
@@ -0,0 +1,35 @@
...
+    [TestMethod]
+    public void Render_WhenDurationIsNull_ReturnsEmptyString()
+        => Assert.AreEqual(string.Empty, HumanReadableDurationFormatter.Render(null));
+
+    [TestMethod]
+    [DataRow(0, true, "(0s)")]
+    [DataRow(5, true, "(5s)")]
+    [DataRow(65, true, "(1m 05s)")]
+    [DataRow(3599, true, "(59m 59s)")]
+    [DataRow(65, false, "1m 05s")]
```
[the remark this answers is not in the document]
Evangelink: Good catch — fixed in 1b25ddc. Added DataRows for `(false, 0)` with pure-second durations (`0s`, `5s`, `59s`) and the parentheses-less `59m 59s` boundary so the previously-uncovered arm now has direct coverage.

PR #8921, opened 2026-06-08
Restore internal TestRequestExecutionTimeInfo for binary compat with older MSBuild extension

Evangelink:
## Summary

PR #8514 deleted the internal `TestRequestExecutionTimeInfo` struct from `Microsoft.Testing.Platform`. Older shipped versions of `Microsoft.Testing.Extensions.MSBuild` (packaged as `Microsoft.Testing.Platform.MSBuild` <= 2.2.x) still reference the type from their `DataTypesConsumed` static array, so any test app that ends up with the **new** `Microsoft.Testing.Platform.dll` and the **old** `Microsoft.Testing.Extensions.MSBuild.dll` side-by-side crashes immediately on startup with:

```
System.TypeLoadException: Could not load type 'Microsoft.Testing.Platform.Extensions.Messages.TestRequestExecutionTimeInfo'
    from assembly 'Microsoft.Testing.Platform, Version=2.3.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a'.
   at Microsoft.Testing.Extensions.MSBuild.MSBuildConsumer..ctor(IServiceProvider, ICommandLineOptions)
```

### Repro from issue #10202

The user-supplied repro pairs `xunit.v3.core.mtp-v2` 4.0.0-pre.108 (transitively pins `Microsoft.Testing.Platform.MSBuild` 2.2.2) with `Microsoft.Testing.Extensions.TrxReport` 2.3.0-preview (brings in the new `Microsoft.Testing.Platform` 2.3.0). NuGet keeps the older `Microsoft.Testing.Extensions.MSBuild.dll` while upgrading the platform, triggering the crash.

## Fix

Restore the internal `TestRequestExecutionTimeInfo` struct as a binary-compat shim. The current platform never publishes it, so the consumer's switch arm that would handle it is now dead code in old extension binaries — but the type *exists*, so the static field initializer no longer trips the type loader. A doc comment on the type explains why it's kept around.

## Regression test

Extends the existing `ForwardCompatibilityTests` test asset to also pin `Microsoft.Testing.Platform.MSBuild` to the previous version (2.2.1), set `GenerateTestingPlatformEntryPoint=false` (preserving the asset's custom `Main`), and call `builder.AddMSBuild()`. I verified the test:

- **fails** with the exact `TypeLoadException` above when the fix file is removed
- **passes** with the fix restored
- doesn't regress the rest of `Microsoft.Testing.Platform.UnitTests` (1104/1104 pass)

### DIFF
--- a/src/Platform/Microsoft.Testing.Platform/Messages/TestRequestExecutionTimeInfo.cs
+++ b/src/Platform/Microsoft.Testing.Platform/Messages/TestRequestExecutionTimeInfo.cs
@@ -22,7 +22,8 @@ namespace Microsoft.Testing.Platform.Extensions.Messages;
 //
 // This type is intentionally marked [Obsolete] so we remember to drop it in the next major
 // release (where breaking binary compatibility with the 2.2.x extension line is acceptable).
-[Obsolete("Kept only for binary compatibility with Microsoft.Testing.Platform.MSBuild <= 2.2.x. Do not use; remove in the next major release.")]
+// Tracked by https://github.com/microsoft/testfx/issues/8925.
+[Obsolete("Kept only for binary compatibility with Microsoft.Testing.Platform.MSBuild <= 2.2.x. Do not use; remove in the next major release. See https://github.com/microsoft/testfx/issues/8925.")]
 internal readonly struct TestRequestExecutionTimeInfo(TimingInfo timingInfo) : IData
 {
     public string DisplayName => nameof(TestRequestExecutionTimeInfo);
--- a/src/Platform/Microsoft.Testing.Platform/Messages/TestRequestExecutionTimeInfo.cs
+++ b/src/Platform/Microsoft.Testing.Platform/Messages/TestRequestExecutionTimeInfo.cs
@@ -19,6 +19,10 @@ namespace Microsoft.Testing.Platform.Extensions.Messages;
 // Microsoft.Testing.Extensions.MSBuild) can still resolve the type at load time. The type
 // is never published by the current platform, so the consumer's switch arm that handles it
 // simply never fires.
+//
+// This type is intentionally marked [Obsolete] so we remember to drop it in the next major
+// release (where breaking binary compatibility with the 2.2.x extension line is acceptable).
+[Obsolete("Kept only for binary compatibility with Microsoft.Testing.Platform.MSBuild <= 2.2.x. Do not use; remove in the next major release.")]
 internal readonly struct TestRequestExecutionTimeInfo(TimingInfo timingInfo) : IData
 {
     public string DisplayName => nameof(TestRequestExecutionTimeInfo);
--- a/src/Platform/Microsoft.Testing.Platform/Messages/TestRequestExecutionTimeInfo.cs
+++ b/src/Platform/Microsoft.Testing.Platform/Messages/TestRequestExecutionTimeInfo.cs
@@ -1,0 +1,42 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+namespace Microsoft.Testing.Platform.Extensions.Messages;
+
+// This internal type was removed in https://github.com/microsoft/testfx/pull/8514 but
+// is referenced by older shipped versions of Microsoft.Testing.Extensions.MSBuild
+// (packaged as Microsoft.Testing.Platform.MSBuild <= 2.2.x). When such an older extension
+// is loaded against this newer Microsoft.Testing.Platform assembly — a common scenario
+// when another package (e.g. xunit.v3.core.mtp-v2) pins an older Microsoft.Testing.Platform.MSBuild
+// version while a newer package (e.g. Microsoft.Testing.Extensions.TrxReport) brings in this
+// newer Microsoft.Testing.Platform — the old MSBuildConsumer's static initializer fails with:
+//   System.TypeLoadException: Could not load type
+//   'Microsoft.Testing.Platform.Extensions.Messages.TestRequestExecutionTimeInfo'
+//   from assembly 'Microsoft.Testing.Platform, Version=2.3.0.0, ...'
+// To preserve binary compatibility with previously shipped extensions, the type definition
+// must continue to exist in this assembly even though no code in this version publishes it.
+// The internal extension projects which still need the InternalsVisibleTo grant (e.g.
+// Microsoft.Testing.Extensions.MSBuild) can still resolve the type at load time. The type
+// is never published by the current platform, so the consumer's switch arm that handles it
+// simply never fires.
+internal readonly struct TestRequestExecutionTimeInfo(TimingInfo timingInfo) : IData
+{
+    public string DisplayName => nameof(TestRequestExecutionTimeInfo);
+
+    public string? Description => "Information about the test execution times.";
+
+    public TimingInfo TimingInfo { get; } = timingInfo;
+
+    public override string ToString()
+    {
+        StringBuilder builder = new StringBuilder("TestRequestExecutionTimeInfo { DisplayName = ")
+            .Append(DisplayName)
+            .Append(", Description = ")
+            .Append(Description)
+            .Append(", TimingInfo = ")
+            .Append(TimingInfo.ToString())
+            .Append(" }");
+
+        return builder.ToString();
+    }
+}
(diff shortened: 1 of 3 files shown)


src/Platform/Microsoft.Testing.Platform/Messages/TestRequestExecutionTimeInfo.cs
```diff
@@ -0,0 +1,42 @@
...
+// when another package (e.g. xunit.v3.core.mtp-v2) pins an older Microsoft.Testing.Platform.MSBuild
+// version while a newer package (e.g. Microsoft.Testing.Extensions.TrxReport) brings in this
+// newer Microsoft.Testing.Platform — the old MSBuildConsumer's static initializer fails with:
+//   System.TypeLoadException: Could not load type
+//   'Microsoft.Testing.Platform.Extensions.Messages.TestRequestExecutionTimeInfo'
+//   from assembly 'Microsoft.Testing.Platform, Version=2.3.0.0, ...'
+// To preserve binary compatibility with previously shipped extensions, the type definition
+// must continue to exist in this assembly even though no code in this version publishes it.
+// The internal extension projects which still need the InternalsVisibleTo grant (e.g.
+// Microsoft.Testing.Extensions.MSBuild) can still resolve the type at load time. The type
```
Evangelink, line 19: **[NIT] Documentation Accuracy**

The sentence conflates two distinct mechanisms:

- **Compile-time**: `InternalsVisibleTo` grants source-code projects (like the current `Microsoft.Testing.Extensions.MSBuild` source) the ability to *reference* this `internal` type.
- **Runtime**: The CLR resolves a `TypeRef` by namespace + type name in the target assembly — `InternalsVisibleTo` plays no role at load time. Old shipped binaries can load the type simply because it *exists* in the assembly, regardless of the access modifier.

Suggested rewrite:
```csharp
// Source builds of extension projects that have an InternalsVisibleTo grant
// (e.g. Microsoft.Testing.Extensions.MSBuild) can still reference this type at compile time.
// Old shipped binaries can load it at runtime because the CLR resolves TypeRefs by
// name/namespace in the referenced assembly, independently of access modifiers.
```

**Recommendation:** Clarify that the InternalsVisibleTo grant governs compile-time access by source projects, while runtime resolution by old binaries depends only on type existence in the assembly.

Evangelink (commented):
> [!NOTE]
> 🤖 **Automated review by GitHub Copilot.** Posted via a maintainer's GitHub token, so it appears under their account — the account owner did **not** write or approve this content personally. Generated by the [Expert Code Review workflow](https://github.com/microsoft/testfx/actions/runs/27126562928). To request a follow-up action, reply by tagging `@copilot` directly.

| # | Dimension | Verdict |
|---|-----------|---------|

✅ 21/22 dimensions clean (4 N/A: Analyzer Quality, Data-Driven Coverage, IPC Wire Compatibility, PowerShell Scripting).

- [ ] Documentation — lines 18-19: sentence conflates `InternalsVisibleTo` (compile-time) with runtime CLR type resolution; see inline comment

---

**Detailed notes per applicable dimension:**

**1 — Algorithmic Correctness:** The fix is correct. Restoring the type in the right namespace (`Microsoft.Testing.Platform.Extensions.Messages`) is all the CLR needs to satisfy the `TypeRef` baked into old `MSBuildConsumer` binaries. The implementation (`IData`, `TimingInfo` constructor, `ToString`) faithfully matches what the old platform exported.

**4 — Public API & Binary Compatibility:** The type is `internal`, so no `PublicAPI.Unshipped.txt` update is needed and no public API surface is added. Binary compat is achieved at exactly the right level: the CLR resolves `typeof(TestRequestExecutionTimeInfo)` in the static field initializer of the old `MSBuildConsumer` by name/namespace lookup alone; struct field layout is irrelevant for this use-case.

**13 — Test Completeness:** The regression test correctly exercises the type-load path: the test asset pins `Microsoft.Testing.Platform.MSBuild` to `2.2.1`, calls `builder.AddMSBuild()` (which loads the old `MSBuildConsumer` type and triggers its static initializer), and asserts exit code 0. A `TypeLoadException` on startup would produce a non-zero exit code, so the assertion is sufficient. Note: the issue repro used version 2.2.2, but 2.2.1 has the same `TestRequestExecutionTimeInfo` reference and exercises the identical code path.

**17 — Documentation Accuracy (NIT):** See inline comment. The existing comment block is mostly clear; only lines 18-19 conflate compile-time `InternalsVisibleTo` access with the runtime CLR type-resolution mechanism. Worth clarifying for future maintainers who may otherwise think the `InternalsVisibleTo` grant is what makes old binaries resolve the type at runtime. Also consider a plain comment (no TODO required unless tracked) noting the approximate version range after which this shim could be removed, e.g. *"Can be removed once `Microsoft.Testing.Platform.MSBuild <= 2.2.x` is no longer in active use."*

> Generated by [Expert Code Review (on open)](https://github.com/microsoft/testfx/actions/runs/27126562928) for issue #8921 · sonnet46 4.1M

fhnaseer (approved):

Youssef1313 (changes requested):
This doesn't sound correct to me.

If new MTP is consumed along side with older MSBuild extension, and the new MTP is never publishing the message, then that completely breaks all implementation assumptions of MSBuild extension and will effectively be broken.

Evangelink:
> This doesn't sound correct to me.
> 
> If new MTP is consumed along side with older MSBuild extension, and the new MTP is never publishing the message, then that completely breaks all implementation assumptions of MSBuild extension and will effectively be broken.

That does seem correct to me. The old MSBuild extension declares the type in the list of consumed type but the type cannot be found (as it's coming from MTP but doesn't exist in new version of MTP):
image

Youssef1313:
Answered offline.

Evangelink:
Follow-up on @Evangelink's observation: just restoring the type fixes the type-load crash but leaves the old `Microsoft.Testing.Platform.MSBuild` (≤ 2.2.x) extension functionally broken — its `MSBuildConsumer` was relying on receiving `TestRequestExecutionTimeInfo` at end-of-session to trigger `HandleSummaryAsync`, which is what sends the `TestRunSummaryRequest` over the IPC pipe to the parent MSBuild process. Without the publish, the old extension loads cleanly but the MSBuild build log never shows the test run summary line.

Pushed 053ba72 which restores the publisher side that was removed in #8514:

- Re-implements `IDataProducer` on `TestHostTestFrameworkInvoker`
- Re-adds `TestRequestExecutionTimeInfo` to `DataTypesProduced`
- Re-emits the message at the end of `ExecuteAsync` (start/end `DateTimeOffset` + `Stopwatch`)
- Wrapped in `#pragma warning disable CS0618` since the type is now `[Obsolete]`

This is safe for current consumers: the new in-tree `MSBuildConsumer` no longer lists `TestRequestExecutionTimeInfo` in its `DataTypesConsumed`, so `AsynchronousMessageBus.PublishAsync` silently drops the message (early-returns when `_dataTypeConsumers.TryGetValue` finds no entry — verified in `Messages/AsynchronousMessageBus.cs`). Only the old shipped consumer picks it up.

Verified locally:
- `./build.cmd -c Debug -pack` → succeeded with 0 warnings.
- `NewerPlatform_WithPreviousExtensions_ShouldExecuteTests` (the test added in this PR) still passes.
- The full `MSBuildTests_Test.InvokeTestingPlatform_Target_Should_Execute_Tests_Without_Showing_Error_Detail_SingleTfm` matrix (12 cases) still passes — the existing assertion `Failed: …, Passed: …, Skipped: …, Total: …, Duration: …` still matches, confirming the new in-tree consumer's `OnTestSessionFinishingAsync`-based summary path is unaffected.

Not added: an automated MSBuild-driven test that pins `Microsoft.Testing.Platform.MSBuild` to `2.2.1` while running on the new platform and asserts the summary line still flows from the *old* consumer. The infrastructure to do that exists in `MSBuildTests.Test.cs` (it would need a SourceCode variant with the package pinned to `$PreviousExtensionVersion$`), but it's a meaningful chunk of test asset work. Happy to add it in this PR or in a follow-up — let me know your preference.

Evangelink:
Quick follow-up to make the obsolete intent harder to lose track of:

- Confirmed the type is already `[Obsolete]` (from c915e821b) and reviewed each remaining usage of it; all are inside `#pragma warning disable CS0618` blocks with comments explaining the binary-compat reason.
- Opened #8925 as the tracking task for the next major to drop the type, the `IDataProducer` plumbing on `TestHostTestFrameworkInvoker`, and revisit the 2.2.1 pin in `ForwardCompatibilityTests`.
- 5a7fc7f1b links #8925 from the `[Obsolete]` message, the type's leading comment, and both CS0618 pragmas on the publisher so the next major's cleanup is grep-able.

PR #8897, opened 2026-06-07
Fix CodeQL pipeline MSB4216 by switching to dotnet MSBuild engine

Evangelink:
## Problem

The scheduled CodeQL pipeline (`microsoft-testfx-codeql-pr`, definition 1407) has been failing every night since 2026-06-05 with:

```
artifacts\toolset\11.0.0-beta.26305.3\InstallDotNetCore.targets(17,5):
error MSB4216: Could not run the ""InstallDotNetCore"" task because MSBuild could not create or connect to a task host
with runtime ""NET"" and architecture ""x86"". [...] the required executable
""D:\a\_work\1\s\.dotnet\sdk\11.0.100-preview.5.26227.104\MSBuild.exe"" exists and can be run.
```

Latest failure: [build 2994041](https://dnceng.visualstudio.com/internal/_build/results?buildId=2994041).

## Root cause

This is the same class of failure addressed in #8355 ([dotnet/msbuild#13739](https://github.com/dotnet/msbuild/issues/13739)) — when an x86 Full Framework MSBuild host runs an Arcade `UsingTask` with `Runtime=""NET""`, MSBuild tries to spin up an x86 .NET task host, which the .NET SDK does not ship.

What changed: the `windows.vs2026.amd64` agent image was updated from `2026.0522.011033` → `2026.0529.211646` between 2026-06-04 (last green build, #2991767) and 2026-06-05 (first failure, #2992621). The Arcade SDK bump in #8810 (26302.1 → 26303.5) that landed at the same time is a red herring — `BuildTasks.props` is byte-identical between the two Arcade versions, and the `Architecture=""*""` attribute is already declared upstream.

The previous workaround in `eng/Tools.props` still applies `Architecture=""*""` to `InstallDotNetCore`, but the new VS MSBuild on this image apparently ignores it (or has a different task-host launch path that defeats it).

## Fix

Pass `-msbuildEngine dotnet` to `cibuild.cmd` in the CodeQL pipeline only. This makes the build use the .NET SDK MSBuild (which is x64) instead of the default VS x86 MSBuild, side-stepping the x86 task-host code path entirely.

- `azure-pipelines.yml` (PR/CI) and `azure-pipelines-official.yml` run on `windows.vs2026preview.scout.amd64` — a different image — and continue to work, so they stay on the VS MSBuild default.
- `eng/Tools.props` (the prior workaround) is left alone so local dev builds launched from an x86 VS Developer Command Prompt still benefit from it.

## Validation

[...]

PR #8928, opened 2026-06-08
Fix CodeQL pipeline pack failure on uap10.0.16299 outputs

Evangelink:
## Problem

The CodeQL pipeline (`azure-pipelines-codeql.yml`) is failing during the build step with:

```
NuGet.Build.Tasks.Pack.targets(222,5): error : Could not find a part of the path 'D:\a\_work\1\s\artifacts\bin\TestFramework.Extensions\Release\uap10.0.16299'. [src\TestFramework\TestFramework.Extensions\TestFramework.Extensions.csproj]
NuGet.Build.Tasks.Pack.targets(222,5): error : Could not find a part of the path 'D:\a\_work\1\s\artifacts\bin\MSTest.TestAdapter\Release\uap10.0.16299\buildTransitive\uwp'. [src\Adapter\MSTest.TestAdapter\MSTest.TestAdapter.csproj]
```

## Root cause

PR #8897 switched the CodeQL pipeline to `-msbuildEngine dotnet` to work around the MSB4216 `InstallDotNetCore` failure that the bundled VS MSBuild started producing on the `windows.vs2026.amd64` image (since image version `2026.0529.211646`).

However, `cibuild.cmd` implicitly runs `-restore -build -test -sign -pack -publish -ci`. The `dotnet` MSBuild engine (running on .NET runtime) cannot fully build the UWP/uap10.0.16299 targets used by `TestFramework.Extensions` and `MSTest.TestAdapter` (those require .NET Framework MSBuild + `MSBuild.Sdk.Extras` + the UWP managed SDK feature). The build for non-UWP TFMs succeeds, but the expected `uap10.0.16299` bin output directories never get created, so the subsequent `-pack` step crashes in the nuspec file lookup.

## Fix

CodeQL only needs source compilation; it does not need NuGet packages, signing, or publish artifacts. Replace the `cibuild.cmd` invocation with a direct call to root `Build.cmd` (which runs `-installWindowsSdk -restore -build` under the hood), and pass `-ci` explicitly to preserve the CI semantics that `cibuild.cmd` previously brought in. This avoids the pack step entirely, which is the right thing to do for CodeQL anyway.

The explicit `Install Windows SDK` PowerShell task earlier in the pipeline is preserved; `Build.cmd`'s own `-installWindowsSdk` call will just no-op because the SDK is already present.


Evangelink:

azat-msft (approved):

PR #8929, opened 2026-06-08
Fix grade-tests-on-pr workflow: split slash_command and pull_request triggers

Evangelink:
## Problem

The `Grade Tests on PR` agentic workflow added in #8857/#8895 runs on every PR but the agent job is always skipped, so no grading comments ever post. Two bugs in the combined workflow cause this:

### Bug 1 — Slash and PR triggers cannot coexist

gh-aw computes activation as `is_team_member && command_position_ok`. The `command_position_ok` check is injected whenever a workflow declares any `slash_command` trigger, and it runs on **every** triggering event with no event-type guard. On a `pull_request` event the command-position step parses the PR title as the "command text", finds no leading `/grade-tests`, and reports:

> None of the commands `[/grade-tests]` matched the first word (found: `'##'`). Workflow will be skipped.

So `activated=false` on every PR event. See e.g. run [27135369416](https://github.com/microsoft/testfx/actions/runs/27135369416/job/76757075547): the pre-extraction step correctly found 5 changed test files, then the agent silently never ran.

### Bug 2 — Custom step outputs are not surfaced as job outputs

The activation gate also required `needs.pre_activation.outputs.has_changed_tests == 'true'`. But the compiled `pre_activation.outputs:` block only declares the gh-aw built-ins (`activated`, `extract_result`, `matched_command`, `setup-*`, ...) — there is no facility to surface arbitrary `$GITHUB_OUTPUT` writes from `on.steps:` as job outputs. So that condition was always `'' == 'true'` → false.

## Fix

Mirror the existing `review.agent.md` / `review-on-open.agent.md` pattern by splitting the workflow:

| File | Trigger | Role |
| --- | --- | --- |
| `shared/grade-tests-shared.md` | (imported) | Tools, permissions, network, safe-outputs, deterministic resolve+checkout+extract steps, and the full agent prompt body |
| `grade-tests-on-pr.agent.md` | `pull_request` (opened/reopened/synchronize/ready_for_review) with `paths: test/**` | Workflow-level `if:` filters drafts and OneLocBuild PRs |
| `grade-tests.agent.md` | `slash_command: grade-tests` on PR comments | Roles: admin/maintainer/write |

The deterministic pre-extraction (resolve PR base/head → checkout PR head with full history → emit TSV of changed test file regions) now lives in the agent job's top-level `steps:` rather than `pre_activation.steps:`. This sidesteps Bug 2 entirely — step outputs are referenced via `${{ steps.*.outputs.* }}` template substitution **in the same job**, which works, instead of `${{ needs.pre_activation.outputs.* }}` cross-job references, which silently break for custom outputs.

The agent prompt is unchanged in spirit; only the fallback wording was nudged so the agent handles an empty TSV (`file_count == 0`) as the *no test methods found* case rather than relying on the pre-step gate.

## Verification

`gh aw compile --strict` succeeds for both new workflows. The compiled activation gates now read:

- `grade-tests-on-pr.agent.lock.yml`: `activated: ${{ steps.check_membership.outputs.is_team_member == 'true' }}` — no `command_position_ok` requirement for PR events ✅
- `grade-tests.agent.lock.yml`: `activated: ${{ ... is_team_member == 'true' && ... command_position_ok == 'true' }}` — correctly kept for the slash command ✅

End-to-end behaviour (whether the agent actually posts grading comments) can only be confirmed after merge + a real PR touching `test/**`.

## Notes

- This PR intentionally does not include the lockfile drift from other unrelated workflows that local `gh aw compile` regenerated; the `agentics-maintenance.yml` job handles those.
- The catalog in `.github/workflows/README.md` is updated with the new entries (build & test diagnostics) and the new shared component.


Evangelink (commented):
> [!NOTE]
> 🤖 **Automated review by GitHub Copilot.** Posted via a maintainer's GitHub token, so it appears under their account — the account owner did **not** write or approve this content personally. Generated by the [Expert Code Review workflow](https://github.com/microsoft/testfx/actions/runs/27139715543). To request a follow-up action, reply by tagging `@copilot` directly.

✅ **22/22 dimensions clean — no findings.**

---

### Analysis notes

**Bug 1 (Slash + PR triggers cannot coexist)** — Correctly fixed. The old single workflow declared both `on: pull_request` and `on: slash_command: ...`, which caused gh-aw's activation gate to inject a `command_position_ok` check on every event type. On every `pull_request` invocation, the check parsed the PR title as command text, found no `/grade-tests` prefix, and set `activated=false`. Splitting into `grade-tests-on-pr.agent.md` (PR-only trigger) and `grade-tests.agent.md` (slash-command-only) eliminates the cross-contamination — each workflow's activation logic now matches its trigger type.

**Bug 2 (Custom step outputs never surfaced as job outputs)** — Correctly fixed. The old activation condition gated on `needs.pre_activation.outputs.has_changed_tests == 'true'`, but `has_changed_tests` was computed inside the agent job's `extract` step, never in `pre_activation`. That output was always `''`, so the comparison was always false. The new PR-trigger workflow replaces the gate with a `paths: test/**` trigger-level filter, which is evaluated by the GitHub Actions scheduler before any job runs. The `extract` step still emits `has_changed_tests` as an output, but it is now consumed inside the agent prompt (to decide whether to post grades or a fallback "nothing to grade" comment), not in a broken activation guard.

**Shared-component pattern** — `shared/grade-tests-shared.md` correctly mirrors the existing `shared/review-shared.md` pattern: tools, permissions, network, deterministic pre-steps, and agent prompt body all live in the shared file; each consumer contributes only its own trigger configuration (`on:`), `safe-outputs.noop`, and `concurrency` block. The `safe-outputs.noop` duplication in each consumer is explained and justified — it is a known workaround for a current gh-aw import-merge limitation around `report-as-issue: false` propagation.

**Concurrency groups** — The groups correctly use `pull_request.number` for the PR-trigger workflow and `issue.number` for the slash-command workflow. Since GitHub assigns equal numbers to PRs and issues, both resolve to the same string for a given PR, so a manual `/grade-tests` invocation and a concurrent push-triggered run will correctly compete in the same cancellation group.

**Security** — Fork PRs are blocked by `github.event.pull_request.head.repo.id == github.repository_id` in both the `pre_activation` condition and the outer `activation` guard. Role checking (`GH_AW_REQUIRED_ROLES: "admin,maintainer,write"`) is applied in the compiled lock's `check_membership` step. Bash scripts use `set -euo pipefail`; all external values passed to git/jq are environment variables, not shell-interpolated strings. Strict mode (`strict: true`) is preserved in the lock metadata, and every `uses:` reference is pinned to a full SHA.

> Generated by [Expert Code Review (on open)](https://github.com/microsoft/testfx/actions/runs/27139715543) for issue #8929 · sonnet46 4.2M

azat-msft (approved):

PR #8930, opened 2026-06-08
Document that AddOpenTelemetryProvider does not register default instrumentation or exporters

### LINKED ISSUE #8409: [MTP Diagnostic Analysis] AddOpenTelemetryProvider defaults to a no-op despite documented default behavior
## Summary
`AddOpenTelemetryProvider()` documents that null delegates use default tracing/metrics behavior, but the implementation just builds empty `TracerProvider` / `MeterProvider` instances. No Microsoft Testing Platform instrumentation or exporter is added unless the caller wires everything manually.

## Evidence
- `Q:\src\testfx\src\Platform\Microsoft.Testing.Extensions.OpenTelemetry\OpenTelemetryProviderExtensions.cs:20-29`
  - `If no delegates are supplied, default OpenTelemetry settings are used.`
  - `public static void AddOpenTelemetryProvider(..., Action<TracerProviderBuilder>? withTracing = null, Action<MeterProviderBuilder>? withMetrics = null)`
- `Q:\src\testfx\src\Platform\Microsoft.Testing.Extensions.OpenTelemetry\OpenTelemetryProvider.cs:19-25`
  - `TracerProviderBuilder tracerProviderBuilder = Sdk.CreateTracerProviderBuilder();`
  - `withTracing?.Invoke(tracerProviderBuilder);`
  - `_tracerProvider = tracerProviderBuilder.Build();`
  - `MeterProviderBuilder meterProviderBuilder = Sdk.CreateMeterProviderBuilder();`
  - `withMetrics?.Invoke(meterProviderBuilder);`
  - `_meterProvider = meterProviderBuilder.Build();`
- `Q:\src\testfx\src\Platform\Microsoft.Testing.Extensions.OpenTelemetry\OpenTelemetryProviderExtensions.cs:45-59`
  - `AddTestingPlatformInstrumentation(...) => builder.AddSource(OpenTelemetryPlatformService.ActivitySourceName);`
  - `AddTestingPlatformInstrumentation(...) => builder.AddMeter(OpenTelemetryPlatformService.MeterName);`
- `Q:\src\testfx\samples\public\MTPOTel\Program.cs:31-42`
  - sample has to call `tracing.AddTestingPlatformInstrumentation();`
  - sample has to call `metrics.AddTestingPlatformInstrumentation();`
  - sample has to add exporters explicitly.

## Why this is a real issue
`builder.AddOpenTelemetryProvider();` looks valid from the API and XML docs, but it produces no test traces or metrics. The package README also says the extension "instruments test execution with OpenTelemetry-compatible traces and metrics", which is not true for the zero-configuration path.

## Suggested resolution
Either:
1. register Microsoft Testing Platform instrumentation by default (and document that exporters are still caller-configured), or
2. fail fast / warn when neither delegate configures instrumentation/exporters, and update the XML docs/README so the API contract matches reality.

## Related issues
- None found

Evangelink:
Fixes #8409.

## Problem

`AddOpenTelemetryProvider` documented that omitting `withTracing` / `withMetrics` would use "default OpenTelemetry settings", but it actually built empty `TracerProvider` / `MeterProvider` instances. The PACKAGE.md claim that the extension "instruments test execution with OpenTelemetry-compatible traces and metrics" was therefore not true for the zero-configuration path — `builder.AddOpenTelemetryProvider();` produced no MTP telemetry.

## Approach

Per the issue, option #2: keep runtime behavior under the caller's control and make the API contract honest. Auto-registering MTP instrumentation inside the provider was considered but rejected to preserve the (real, if niche) scenario of using this extension purely to manage the lifecycle of an OTel pipeline for application-only telemetry; the caller still needs a `Program.cs` to wire exporters either way, so the convenience gain was small.

- `OpenTelemetryProviderExtensions.cs`: XML docs on `AddOpenTelemetryProvider` now state explicitly that no instrumentation or exporter is registered by default and that the caller must wire both via the delegates, including `AddTestingPlatformInstrumentation()` to subscribe to MTP. `AddTestingPlatformInstrumentation` docs describe both use sites (inside the delegates and on a `TracerProviderBuilder` / `MeterProviderBuilder` configured outside the helper).
- `PACKAGE.md`: drops the misleading "instruments test execution" framing and spells out the wiring required to actually collect MTP telemetry.
- `samples/public/MTPOTel/Program.cs`: adds a short comment summarising the contract.

No public API surface change, no runtime behavior change — docs and sample only.

## Verification

Built the extension cleanly (Debug, all TFMs) — 0 warnings, 0 errors.

### DIFF
--- a/samples/public/MTPOTel/Program.cs
+++ b/samples/public/MTPOTel/Program.cs
@@ -28,7 +28,9 @@ public class Program
             _ => new TestFrameworkCapabilities(),
             (capabilities, serviceProvider) => new SimpleTestFramework(serviceProvider));
 
-        // Enable OpenTelemetry with console exporter
+        // Enable OpenTelemetry with console exporter. AddOpenTelemetryProvider does not register any instrumentation
+        // or exporter by default, so the delegates must call AddTestingPlatformInstrumentation() to subscribe to the
+        // Microsoft Testing Platform source/meter and wire whatever exporter(s) you want.
         testApplicationBuilder.AddOpenTelemetryProvider(
             tracing =>
             {
--- a/src/Platform/Microsoft.Testing.Extensions.OpenTelemetry/OpenTelemetryProviderExtensions.cs
+++ b/src/Platform/Microsoft.Testing.Extensions.OpenTelemetry/OpenTelemetryProviderExtensions.cs
@@ -17,15 +17,24 @@ namespace Microsoft.Testing.Extensions;
 public static class OpenTelemetryProviderExtensions
 {
     /// <summary>
-    /// Adds OpenTelemetry tracing and metrics providers to the application builder, allowing customization of tracing
-    /// and metrics configuration.
+    /// Registers OpenTelemetry tracing and metrics providers whose lifetime is managed by the Microsoft Testing Platform.
     /// </summary>
-    /// <remarks>This method enables distributed tracing and metrics collection for the application. To
-    /// customize telemetry behavior, provide configuration delegates for tracing and/or metrics. If no delegates are
-    /// supplied, default OpenTelemetry settings are used.</remarks>
+    /// <remarks>The providers are created with empty configuration. Callers are responsible for wiring everything
+    /// the providers should observe and export, including:
+    /// <list type="bullet">
+    /// <item><description>The Microsoft Testing Platform instrumentation, via
+    /// <see cref="AddTestingPlatformInstrumentation(TracerProviderBuilder)"/> and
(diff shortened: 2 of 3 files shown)


Evangelink:
Filed #8931 to track the Roslyn analyzer follow-up (warn when `Microsoft.Testing.Extensions.OpenTelemetry` is referenced but `AddOpenTelemetryProvider` is never invoked). Keeping this PR scoped to the docs/contract fix per the discussion.

PR #8993, opened 2026-06-10
Fix Invoke_TestTarget_With_Incompatible_Arch on Linux

Evangelink:
The Linux pipeline on `main` is failing on `Invoke_TestTarget_With_Incompatible_Arch`:

```
Assertion failed. Expected string to contain the specified substring.
Expression 'AssertOutputContains' failed for member 'Invoke_TestTarget_With_Incompatible_Arch' at line 188
```

## What changed

On Linux, launching an arm64 ELF apphost from a linux-x64 host no longer surfaces as a `Win32Exception` from MSBuild's `ToolTask`. As a result, the `error MSB6003: ... could not be run. System.ComponentModel.Win32Exception` message that the test expected on line 188 is never emitted on Linux.

What we see now in CI is the failure being reported by `InvokeTestingPlatformTask.HandleTaskExecutionErrors` as the captured tool output instead:

```
.../MSBuild Tests.dll : error run failed: Tests failed: .../MSBuild Tests: 1: Syntax error: word unexpected (expecting `)`)
.../MSBuild Tests.dll : error run failed:
.../MSBuild Tests.dll : error run failed: === COMMAND LINE ===
.../MSBuild Tests.dll : error run failed: /path/MSBuild Tests --internal-msbuild-node /tmp/...
```

(The shell-parse error comes from `/bin/sh` trying to interpret the arm64 ELF binary as a shell script.)

Windows and macOS still go through the `MSB6003 + Win32Exception` path and continue to pass.

## The fix

Rework the OS-specific assertions in `Invoke_TestTarget_With_Incompatible_Arch`:

- For every OS, first assert non-zero exit and that the run banner mentions the incompatible TFM/architecture (`[net10.0|arm64]`) so we still verify the build/launch actually reached the architecture-mismatched apphost.
- Keep the existing `MSB6003` + `An error occurred trying to start process` assertions and the platform-specific OS error strings for Windows and macOS.
- For Linux, assert only the `error run failed: Tests failed:` line emitted by `InvokeTestingPlatformTask` when the tool exits non-zero (paired with the arch-banner check above, which is what guarantees we're actually exercising the incompatible-arch code path).

Fixes the green Linux pipeline on main.

### DIFF
--- a/test/IntegrationTests/Microsoft.Testing.Platform.Acceptance.IntegrationTests/MSBuildTests.Test.cs
+++ b/test/IntegrationTests/Microsoft.Testing.Platform.Acceptance.IntegrationTests/MSBuildTests.Test.cs
@@ -180,30 +180,40 @@ public class MSBuildTests_Test : AcceptanceTestBase<NopAssetFixture>
             failIfReturnValueIsNotZero: false,
             cancellationToken: TestContext.CancellationToken);
 
-        // On Windows, we run the exe directly.
-        // On other OSes, we run with dotnet exec.
-        // This yields two different outputs, pointing to the same issue.
-        string executableName = OperatingSystem.IsWindows() ? "MSBuild Tests.exe" : "MSBuild Tests";
+        // The build should fail and the failure should mention the incompatible target architecture
+        // so we know the build/launch reached the apphost rather than failing for an unrelated reason.
+        result.AssertExitCodeIsNot(0);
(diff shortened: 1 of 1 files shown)


Evangelink:
Pushed a follow-up: the Linux failure mode for launching an incompatible-architecture apphost turns out to be non-deterministic across machines/runs. The original `main` Release Linux run produced the `error run failed: Tests failed:` shape, but Debug Linux on this PR produced the historical `MSB6003 + Exec format error` shape instead. The test now accepts either.

Issue #1837, opened 2023-11-27
Add support for NativeAOT


nohwnd:
_Originally posted by @dotMorten in:_ <https://github.com/nohwnd/mstest-runner-examples/issues/1>

## Summary

One of my biggest griefs with MSTest is I can't test if my stuff works right in an AoT build.
Try building your project with `dotnet publish -c Release -p PublishAot=true -r win-x64` and run it, and you'll get loads of build warnings, and a runtime crash.

Now I understand why it currently doesn't work (too much reflection and datacontract serializers), but solving this would be huge for ensuring AoT compatibility.

It would likely require a little code generator for each `TestMethod` attribute instead of relying on reflection discovery, and using the new System.Text.Json serializers.

nohwnd:
The new MSTest runner which lives in https://github.com/microsoft/testfx/tree/main/src/Platform/Microsoft.Testing.Platform is internally fully NativeAOT compatible, and can already do this. It inserts itself into the executable you building, which makes it work nicely with `dotnet publish` and it's various options. Including trimming and PublishAOT. 

What is missing is a test framework that can do such thing. As you describe MSTest relies internally on reflection. But we have an alternative implementation which is now shipped in Microsoft.Testing.Framework, and we want to eventually use the same approach for mstest.

As an output you will get an application that can run tests, and has no additional dependencies.

I've added an example here, of how you can achieve nativeaot test running with the existing pieces right now. For actual integration of this into MSTest we don't have a public plan yet.

https://github.com/nohwnd/mstest-runner-examples/blob/main/NativeAOTProject1/README.md

Evangelink:
This ticket will be used as epic so removing it from 3.3 miletsone.

nohwnd:
Here is super early preview of the nativeAot for MSTest. 

We've created a whole new engine and a source generator that are published as MSTest.Engine and MSTest.SourceGeneration nuget packages on our preview feed (https://aka.ms/mstest/preview) under our Microsoft Testing Platform Tools license.

Both the engine and source generator are in alpha, so please don't consider any of the public APIs final.

Here is an example of project that can run tests in native aot. Putting it together it quite manual, but we are working on improving this. 

## How to use? 

Create a new mstest or console project. Set `<OutputType>exe</OutputType>` and `<PublishAot>true</PublishAot>`. Reference the listed packages, and add test application setup to `Main` method.

Run `dotnet publish -r win-x64` to publish the application. 

Run the application from the publish folder. 

[...]

PR #9005, opened 2026-06-10
AotReflection: strip nullable annotation from emitted typeof(...) (CS8639)

Evangelink:
Part of #1837. Depends on #9004 (introduces the test project used here).

## Why

While building the AotReflection unit-test project in #9004, the new compile-clean snapshot test surfaced a real bug in the PoC: the generated source produced `CS8639: The typeof operator cannot be used on a nullable reference type`.

## Root cause

[`TestClassModelBuilder.FullyQualifiedFormat`](../tree/main/src/Analyzers/MSTest.AotReflection.SourceGeneration/Generators/TestClassModelBuilder.cs) was built with `SymbolDisplayMiscellaneousOptions.IncludeNullableReferenceTypeModifier`. The resulting FQN string was then fed into **two** very different emission sites:

- **Casts** like `(global::Sample.TestContext?)value!` — where the `?` is harmless (the `value!` already suppresses the nullability warning).
- **`typeof(...)`** expressions — where `?` on a **reference type** is invalid C# (CS8639).

## Fix

Drop `IncludeNullableReferenceTypeModifier` from `FullyQualifiedFormat`. This:

- Strips `?` from reference-type FQNs everywhere (no behavioural impact on casts).
- Preserves `?` on **nullable value types** (`int?`, `DateTime?`) because those are rendered via `UseSpecialTypes` / generics options, which are unaffected.

The property-setter snapshot assertion is updated to match (`(global::Sample.TestContext)value!` instead of `(global::Sample.TestContext?)value!`).

## Regression test

New `Generator_StripsNullableAnnotation_FromTypeofExpressions` test:

- Uses a class with a `TestContext?` property, a `string?` parameter, and an `int?` parameter.
- Compiles the generated registry with the Roslyn C# compiler.
- Asserts text shape: `typeof(global::Sample.TestContext)`, `typeof(string)`, `typeof(int?)`; and the absence of `typeof(global::Sample.TestContext?)` / `typeof(string?)`.
- Asserts there are **zero compile errors** in the emitted code.

## Validation

`dotnet build` + `dotnet run` for `MSTest.AotReflection.SourceGeneration.UnitTests`: `14/14` passing, `0` warnings.


Evangelink:

PR #9029, opened 2026-06-11
Fix telemetry tests for new default diagnostic file name shape

Evangelink:
PR #8971 changed the default Microsoft.Testing.Platform diagnostic log filename prefix from `log` to `<asm>_<tfm>_<arch>` (see [TestApplication.cs](https://github.com/microsoft/testfx/blob/main/src/Platform/Microsoft.Testing.Platform/Builder/TestApplication.cs)), and updated `DiagnosticTests.cs` accordingly with a new `BuildDefaultDiagnosticFilePathPattern` helper. However, three telemetry test files were missed and still asserted against the old `log_*.diag` filename, which is why main is currently red ([build 2996915](https://dnceng.visualstudio.com/internal/_build/results?buildId=2996915&view=results)).

Failing tests fixed:

- `Telemetry_ByDefault_TelemetryIsEnabled`
- `Telemetry_WhenOptingOutTelemetry_WithEnvironmentVariable_TelemetryIsDisabled`
- `Telemetry_WhenOptingOutTelemetry_With_DOTNET_CLI_EnvironmentVariable_TelemetryIsDisabled`
- `Telemetry_WhenEnableTelemetryIsFalse_WithTestApplicationOptions_TelemetryIsDisabled`
- `MTP_RunTests_SendsTelemetryWithSettingsAndAttributes`
- `MTP_DiscoverTests_SendsTelemetryEvent`
- `MTP_WhenTelemetryDisabled_DoesNotSendMSTestEvent`

Each of the three telemetry test files now gets the same `BuildDefaultDiagnosticFilePathPattern` helper that mirrors the one already present in `DiagnosticTests.cs`, so we no longer hard-code the diagnostic filename shape across multiple call sites.

### DIFF
--- a/test/IntegrationTests/MSTest.Acceptance.IntegrationTests/TelemetryTests.cs
+++ b/test/IntegrationTests/MSTest.Acceptance.IntegrationTests/TelemetryTests.cs
@@ -26,7 +26,7 @@ public sealed class TelemetryTests : AcceptanceTestBase<TelemetryTests.TestAsset
     public async Task MTP_RunTests_SendsTelemetryWithSettingsAndAttributes(string tfm)
     {
         string diagPath = Path.Combine(AssetFixture.MTPProjectPath, "bin", "Release", tfm, TestResultsFolderName);
-        string diagPathPattern = Path.Combine(diagPath, @"log_.*.diag").Replace(@"\", @"\\");
+        string diagPathPattern = BuildDefaultDiagnosticFilePathPattern(diagPath, tfm);
 
         var testHost = TestHost.LocateFrom(AssetFixture.MTPProjectPath, MTPAssetName, tfm);
         TestHostResult testHostResult = await testHost.ExecuteAsync(
@@ -66,7 +66,7 @@ public sealed class TelemetryTests : AcceptanceTestBase<TelemetryTests.TestAsset
     public async Task MTP_DiscoverTests_SendsTelemetryEvent(string tfm)
     {
         string diagPath = Path.Combine(AssetFixture.MTPProjectPath, "bin", "Release", tfm, TestResultsFolderName);
-        string diagPathPattern = Path.Combine(diagPath, @"log_.*.diag").Replace(@"\", @"\\");
+        string diagPathPattern = BuildDefaultDiagnosticFilePathPattern(diagPath, tfm);
 
         var testHost = TestHost.LocateFrom(AssetFixture.MTPProjectPath, MTPAssetName, tfm);
         TestHostResult testHostResult = await testHost.ExecuteAsync(
@@ -92,7 +92,7 @@ public sealed class TelemetryTests : AcceptanceTestBase<TelemetryTests.TestAsset
     public async Task MTP_WhenTelemetryDisabled_DoesNotSendMSTestEvent(string tfm)
     {
         string diagPath = Path.Combine(AssetFixture.MTPProjectPath, "bin", "Release", tfm, TestResultsFolderName);
-        string diagPathPattern = Path.Combine(diagPath, @"log_.*.diag").Replace(@"\", @"\\");
+        string diagPathPattern = BuildDefaultDiagnosticFilePathPattern(diagPath, tfm);
 
         var testHost = TestHost.LocateFrom(AssetFixture.MTPProjectPath, MTPAssetName, tfm);
         TestHostResult testHostResult = await testHost.ExecuteAsync(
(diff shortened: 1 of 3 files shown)


Evangelink:
The telemetry test changes in this PR are working as intended — they are verified passing in the latest CI build [1459171](https://dev.azure.com/dnceng-public/public/_build/results?buildId=1459171):

- `MTP_RunTests_SendsTelemetryWithSettingsAndAttributes`, `MTP_DiscoverTests_SendsTelemetryEvent`, `Telemetry_WhenEnableTelemetryIsFalse_WithTestApplicationOptions_TelemetryIsDisabled`, and the other targeted telemetry tests all pass after the change.

The remaining red on the public-PR build is **2 unrelated, pre-existing test failures**:

- `JUnitReport_WithMSTestRetryAttribute_EmitsFinalOutcomePerTest`
- `JUnitReport_WithMTPRetryFailedTestsExtension_EmitsOneReportPerAttempt`

Both fail in `ClassInitialize` with:

```
error NU1102: Unable to find package Microsoft.Testing.Platform with version (>= 2.3.0-preview.26308.7)
  - Found 55 version(s) in dotnet-public [ Nearest version: 2.2.3 ]
  - Found 1 version(s) in local-shipping [ Nearest version: 2.3.0-ci ]
```

I confirmed these same two failures exist on `main` (internal build [2996915](https://dev.azure.com/dnceng/internal/_build/results?buildId=2996915) at `SHA 741048f7`), so they are not introduced by this PR.

Root cause (for tracking — not in scope here): `MSTestJUnitReportMTPRetry.csproj` declares `<PackageReference Include="Microsoft.Testing.Extensions.Retry" Version="$MicrosoftTestingPlatformVersion$" />` (= `2.3.0-ci` on PR builds). NuGet resolves that "best match" by reaching into the `dotnet-tools` feed and pulling a higher `2.3.0-preview.X` of `Microsoft.Testing.Extensions.Retry` whose transitive MTP dependency (`>= 2.3.0-preview.26308.7`) cannot be satisfied by `local-shipping` (which only has `2.3.0-ci`). Worth filing a follow-up issue.

This PR should be safe to merge despite the unrelated red.

Evangelink:
Coming back to address @Evangelink's pushback ("we should have exact match and we should have local folder as nuget winning priority") — those are both **valid critiques**, and my earlier "dotnet-tools transitive" theory was wrong. Here is the actual root cause:

### Real failure chain

1. `MSTestJUnitReportMTPRetry.csproj` declares:
   ```xml
   <PackageReference Include="MSTest" Version="$MSTestVersion$" />
   ```
   → resolves to local-built `MSTest 4.3.0-ci` from `local-shipping`.
2. `MSTest 4.3.0-ci`'s **packed nuspec** hardcodes:
   ```xml
   <dependency id="Microsoft.Testing.Extensions.CodeCoverage" version="18.9.0-preview.26309.4" />
   ```
   (The `18.9.0-preview.26309.4` value is darc-pinned in [`eng/Versions.props`](https://github.com/microsoft/testfx/blob/main/eng/Versions.props#L11).)
3. `Microsoft.Testing.Extensions.CodeCoverage 18.9.0-preview.26309.4` only lives on the **`dotnet-tools`** feed (verified via flat2 index). Its own nuspec declares:
   ```xml
   <dependency id="Microsoft.Testing.Platform" version="2.3.0-preview.26308.7" />
   ```
4. The test-asset `NuGet.config` generated by [`TestAsset.GetNuGetConfig`](https://github.com/microsoft/testfx/blob/main/test/Utilities/Microsoft.Testing.TestInfrastructure/TestAsset.cs#L108-L166) lists **6 feeds** and does **not** include `dotnet-tools`:
   - `local-shipping` (has `MTP 2.3.0-ci`)
   - `local-nonshipping`, `local-tmp-packages` (empty for MTP)
   - `dotnet-public` (has `MTP <= 2.2.3` only)
   - `dotnet10`, `dotnet11` (empty)
5. NuGet can't satisfy `MTP >= 2.3.0-preview.26308.7` in any of those 6 feeds (`2.3.0-ci` sorts **below** `2.3.0-preview.X` in SemVer 2.0 — `c` < `p`) → **NU1102**.

### Why @Evangelink's two critiques are correct

| Critique | Validity | Notes |
|---|---|---|
| **Exact-match pinning** (`Version="[X]"` instead of `Version="X"`) | ✅ Valid hardening | NuGet treats bare `Version="X"` as `[X, )` (open-ended minimum). Every test-asset PackageReference uses the open form, so a higher-preview drift from any feed is technically allowed. Exact pins on direct refs alone would NOT fix this bug (constraint comes via transitive CodeCoverage), but it's the right belt-and-suspenders posture. |
| **Local-folder winning priority** via `packageSourceMapping` | ✅ Valid critique | Current mapping is `<package pattern="*"/>` for every feed — no priority. Per-package patterns mapping `Microsoft.Testing.*` and `MSTest*` to **only** `local-shipping`/`local-nonshipping` would force testfx-family packages to local. Would have surfaced this issue earlier with a clearer error (NU1102 on `CodeCoverage 18.9.0-preview.26309.4` from local feeds, not on transitive MTP). |

### What actually needs to change

Both critiques are correct hardening, but the underlying mechanical fix is that **`MSTest 4.3.0-ci`'s packed nuspec demands a `CodeCoverage` prerelease only available on `dotnet-tools`**, and the test-asset `NuGet.config` doesn't include that feed. ~20 test assets use `<PackageReference Include="MSTest" ...>` and are all theoretically affected by this combination.

Possible fixes (mix and match):

1. **Add `dotnet-tools` to `TestAsset.GetNuGetConfig`** — smallest mechanical change; mirrors the repo's top-level `NuGet.config`.
2. **Exact-match pinning in test-asset csprojs** (`Version="[$X$]"`) — broader hardening for direct refs.
3. **Per-package `packageSourceMapping`** in test-asset `NuGet.config` — `Microsoft.Testing.*` → local feeds + `dotnet-tools`; everything else → `dotnet-public`/`dotnet10`/`dotnet11`. Strongest isolation.
4. **Downgrade `CodeCoverage` dep in `MSTest`'s packed nuspec to nearest stable (`18.8.0`)** — cleanest at the source but requires coordinating with the VSTest `CodeCoverage` release train.

Opening a tracking issue with these options.

PR #9041, opened 2026-06-11
Apply DebuggerDisableUserUnhandledExceptionsAttribute to Assert.Throws helpers

### LINKED ISSUE #9016: Investigate using DebuggerDisableUserUnhandledExceptionsAttribute on throw asserts
Often, when running my tests that expect an exception, the debugger pauses on tests that expects the exception. This might be dependent on debugger settings in IDE.

The ask here is to try to reproduce that, see if DebuggerDisableUserUnhandledExceptionsAttribute fixes it, and decide if the change should be taken.

Evangelink:
Fixes #9016

## Why

When running with Just My Code enabled, Visual Studio raises a *user-unhandled exception* on the expected exception thrown inside an `Assert.Throws*` call, because the catch lives in MSTest framework code. This makes the debugger pause on tests that are working as intended -- a long-standing pain point for tests that exercise error paths.

## What

.NET 9 introduced [`DebuggerDisableUserUnhandledExceptionsAttribute`](https://learn.microsoft.com/dotnet/api/system.diagnostics.debuggerdisableuserunhandledexceptionsattribute) specifically for this scenario: a supporting debugger (VS 17.12+) will not break on user-unhandled exceptions caught by an attributed method. xUnit applies it the same way.

This PR:

- Adds a polyfill for the attribute in `src/Polyfills/` (no-op on TFMs < `net9.0`), mirroring the existing `StackTraceHiddenAttribute` polyfill pattern.
- Applies `[DebuggerDisableUserUnhandledExceptions]` to the two private helpers that own the `try`/`catch` on user code:
  - `Assert.IsThrowsFailing<TException>` (sync)
  - `Assert.IsThrowsAsyncFailingAsync<TException>` (async)

All `Assert.Throws[Exactly][Async]` overloads -- including the interpolated string handler structs `AssertNonStrictThrowsInterpolatedStringHandler` and `AssertThrowsExactlyInterpolatedStringHandler` -- route through these two helpers, so every entry point is covered without touching the public API.

## Scope / safety

- **Strictly limited to `Assert.Throws*`.** General test-method execution is not annotated, so users still get the expected debugger break on unexpected exceptions thrown by their tests.
- **No public API change** (both helpers are `private static`).
- **No behavior change on older runtimes/debuggers.** The attribute is inert when the runtime or debugger does not support `Debugger.BreakForUserUnhandledException`, so this is safe across all supported TFMs (`netstandard2.0;net462;net8.0;net9.0`).
- **Users who really want the debugger to break on the inner exception** can still use VS' "Break when thrown" exception settings -- those are unaffected.

## Validation

- `./build.cmd -c Debug -projects src/TestFramework/TestFramework/TestFramework.csproj` -> succeeds for `net462`, `netstandard2.0`, `net8.0`, `net9.0` with 0 warnings / 0 errors.
- `./build.cmd -c Debug -test -projects test/UnitTests/TestFramework.UnitTests/TestFramework.UnitTests.csproj` -> all tests pass on `net48`, `net8.0`, `net9.0`, and `net8.0-windows10.0.18362.0`.
- Final end-to-end validation under the VS debugger (with Just My Code on) should be performed by a reviewer to confirm the suppression behavior is observed in practice -- the runtime-level metadata is the only thing this PR can guarantee.


Youssef1313 (approved):

Youssef1313:
LGTM, assuming it was manually validated.

Evangelink (commented):
> [!NOTE]
> 🤖 **Automated review by GitHub Copilot.** Posted via a maintainer's GitHub token, so it appears under their account — the account owner did **not** write or approve this content personally. Generated by the [Expert Code Review workflow](https://github.com/microsoft/testfx/actions/runs/27350055711). To request a follow-up action, reply by tagging `@copilot` directly.

## Summary

The implementation is technically sound. The polyfill follows the exact established pattern of `StackTraceHiddenAttribute.cs`, the TFM guard is precise, and the attribute is placed on the only two methods in the file that contain `catch` blocks over user-supplied action delegates. No public API surface is introduced, no `PublicAPI.Unshipped.txt` update is needed, and backward compatibility is preserved across all supported TFMs.

---

## 22-Dimension Verdict Table

| # | Dimension | Result | Notes |
|---|---|---|---|
| 1 | Algorithmic Correctness | ✅ Pass | Attribute placed on both catch-containing helpers; all entry points funnel through them |
| 2 | Threading & Concurrency | ✅ N/A | No shared mutable state introduced |
| 3 | Exception Handling | ✅ Pass | Attribute correctly marks the methods that own the `try/catch` on user delegates |
| 4 | Public API Design | ✅ Pass | Both annotated methods are `private static`; no `PublicAPI.Unshipped.txt` update needed |
| 5 | Resource Management | ✅ N/A | No new disposable resources |
| 6 | Performance | ✅ Pass | Attribute is pure metadata; zero runtime overhead |
| 7 | Security | ✅ N/A | No trust boundary changes |
| 8 | Localization | ✅ N/A | No user-facing strings |
| 9 | Cross-TFM Correctness | ✅ Pass | `!NET9_0_OR_GREATER` guard compiles polyfill inline for `net462`, `netstandard2.0`, and `net8.0`; forwards to BCL on `net9.0+` |
| 10 | Test Coverage | ⚠️ Inherent | Debugger-interaction behavior cannot be unit-tested; PR author notes end-to-end validation under VS is required |
| 11 | Code Style & Conventions | ✅ Pass | Pattern is identical to `StackTraceHiddenAttribute.cs`; `// <auto-generated />`, `#pragma warning disable`, `[Embedded]` all present |
| 12 | Documentation | ✅ N/A | Private infrastructure; no XML doc required |
| 13 | Error Messages | ✅ N/A | No new error messages |
| 14 | Dependency Management | ✅ Pass | No new package dependencies; `Microsoft.CodeAnalysis` reference is used by every polyfill in the directory |
| 15 | IPC Contract Stability | ✅ N/A | MSTest framework, not MTP wire format |
| 16 | Backward Compatibility | ✅ Pass | Attribute is inert metadata on pre-.NET 9 runtimes; no behavioral change for unsupported debuggers |
| 17 | Scope Discipline | ✅ Pass | Exactly the two methods that own the user-code `catch` blocks, nothing more |
| 18 | Architecture & Design | ✅ Pass | Polyfill lives in `src/Polyfills/` and is glob-included by `TestFramework.csproj`; no layer violations |
| 19 | Invariant & Guard Preservation | ✅ Pass | No weakening of existing guards |
| 20 | Build & Infrastructure | ✅ Pass | `TestFramework.csproj` already includes `$(RepoRoot)src/Polyfills/**/*.cs`; no project-file change needed |
| 21 | Polyfill & Compatibility Shim Correctness | ✅ Pass | `[AttributeUsage(AttributeTargets.Method)]` matches the BCL declaration; `TypeForwardedTo` correctly redirects on NET9+; `[Embedded]` prevents symbol collision when multiple packages carry the same polyfill |
| 22 | Debugger Integration | ✅ Pass | Applied to the innermost methods containing the `catch` over user delegates; attribute is recognized by VS 17.12+ state-machine frames for the async overload |

---

## Key Findings

**Polyfill guard (Dim 9 / 21):** Using `!NET9_0_OR_GREATER` (not `!NETCOREAPP`) is intentional and correct — unlike `StackTraceHiddenAttribute` which arrived in .NET 6, this attribute first appeared in .NET 9. The polyfill is therefore compiled inline for `net8.0` as well; the attribute is silently ignored by pre-NET9 runtimes, which is the documented and expected behavior.

**Catch-site coverage (Dim 1 / 22):** There are exactly two `catch (Exception ex)` blocks over user-supplied delegates in `Assert.ThrowsException.cs` (lines 545 and 567). Both are covered. All public `Throws*`/`ThrowsExactly*`/`ThrowsAsync*` overloads — including the interpolated-string handler constructors — route through one of these two helpers.

**Out-of-scope observation (not blocking):** `AssertionValueRenderer.RenderObject` (`Assertions/AssertionValueRenderer.cs:43`) also catches exceptions from user-supplied `ToString()` overrides. This is a defensive rendering guard, semantically different from `Assert.Throws*`. If the team later decides to suppress the debugger break there too, it would be a natural companion change but is out of scope for this PR.

---

## Suggested Follow-up (non-blocking)

- Open a tracking issue for `AssertionValueRenderer.RenderObject` (and the `Assert.That` expression-evaluation catch blocks) if the team wants to evaluate applying `[DebuggerDisableUserUnhandledExceptions]` more broadly to defensive catches inside the framework.

> 🤖 **Automated content by GitHub Copilot.** Posted via a maintainer's GitHub token, so it appears under their account — the account owner did **not** write or approve this content personally. Generated by the [Expert Code Review (on PR ready)](https://github.com/microsoft/testfx/actions/runs/27350055711/agentic_workflow) workflow. · 421.9 AIC · ⌖ 13 AIC · [◷]()

src/Polyfills/DebuggerDisableUserUnhandledExceptionsAttribute.cs
```diff
@@ -0,0 +1,15 @@
+﻿// <auto-generated />
+
+#pragma warning disable
+
+#if !NET9_0_OR_GREATER
```
Evangelink, line 5: The guard `!NET9_0_OR_GREATER` is the right choice here. Unlike `StackTraceHiddenAttribute` (available since .NET 6 and therefore gated by `!NETCOREAPP`), `DebuggerDisableUserUnhandledExceptionsAttribute` was only added in .NET 9, so the polyfill must be compiled inline for `net462`, `netstandard2.0`, **and `net8.0`**. The attribute will be inert in metadata on pre-.NET 9 runtimes, which is the expected and documented behavior.

src/TestFramework/TestFramework/Assertions/Assert.ThrowsException.cs
```diff
@@ -534,6 +534,7 @@ private static async Task<TException> ThrowsExceptionAsync<TException>(Func<Task
         return null!;
     }
 
+    [DebuggerDisableUserUnhandledExceptions]
```
Evangelink, line 537: Coverage audit — these are the **only two `catch` blocks for user-supplied delegates** in `Assert.ThrowsException.cs`, so placement is complete and correct. For the record:

- All sync public overloads (`Throws`, `ThrowsExactly`, and the two `AssertNonStrictThrowsInterpolatedStringHandler` / `AssertThrowsExactlyInterpolatedStringHandler` constructors) funnel through `IsThrowsFailing` ✓  
- All async public overloads (`ThrowsAsync`, `ThrowsExactlyAsync`) funnel through `IsThrowsAsyncFailingAsync` ✓

One out-of-scope observation worth a follow-up issue: `AssertionValueRenderer.RenderObject` also catches exceptions thrown by user-provided `ToString()` implementations. That method is in a different semantic category (defensive rendering, not assertion semantics), but if the team ever wants to extend the suppression there, it would be a natural companion change.

src/Polyfills/DebuggerDisableUserUnhandledExceptionsAttribute.cs
Evangelink: Confirmed — no code change needed; leaving this note here as design rationale for future maintainers.

src/TestFramework/TestFramework/Assertions/Assert.ThrowsException.cs
Evangelink: Confirmed — coverage audit complete, no code change needed.

Evangelink:
Manually validated.

Youssef1313 (approved):

YuliiaKovalova (approved):

Issue #9069, opened 2026-06-12
Assert.Throws<T>() / Assert.ThrowsExactly<T>() returns null when assertion fails inside AssertScope, violating non-nullable return contract


wi5nuu:
<html>
<body>
<html><head></head><body><h2>Description</h2><p>When any <code inline="">Assert.Throws&lt;T&gt;()</code> or <code inline="">Assert.ThrowsExactly&lt;T&gt;()</code> overload is called inside an <code inline="">AssertScope</code> and the assertion fails (either because no exception is thrown or because the wrong exception type is thrown), the method returns <code inline="">null</code> instead of throwing.</p><p>This violates the non-nullable <code inline="">TException</code> return contract and can lead to a <code inline="">NullReferenceException</code> in user code that accesses the returned exception before the enclosing scope is disposed.</p><p>The root cause is that <code inline="">ReportAssertFailed()</code> detects an active <code inline="">AssertScope</code> and records the failure through <code inline="">scope.AddError()</code> rather than throwing immediately. Execution then continues, and the caller reaches a <code inline="">return null!</code> statement that exists only to satisfy the compiler's requirement for a return value.</p><p>As a result, the assertion failure is deferred until scope disposal, while a <code inline="">null</code> value becomes observable to user code.</p><p>This affects all overloads across:</p><ul><li><p><code inline="">Assert.Throws&lt;T&gt;()</code></p></li><li><p><code inline="">Assert.ThrowsExactly&lt;T&gt;()</code></p></li><li><p><code inline="">Assert.ThrowsAsync&lt;T&gt;()</code></p></li><li><p><code inline="">Assert.ThrowsExactlyAsync&lt;T&gt;()</code></p></li></ul><p>including both interpolated-string-handler and explicit-message variants.</p><h2>Version Used</h2><p>Latest <code inline="">main</code> branch (commit current at the time of analysis).</p><p>The issue affects all versions that include the <code inline="">AssertScope</code> feature (introduced in MSTest 3.x).</p><h2>Steps To Reproduce</h2><ol><li><p>Create a new MSTest 3.x project targeting .NET 8 or later.</p></li><li><p>Add the following test:</p></li></ol><pre><code class="language-csharp">[TestMethod]
public void Throws_ReturnsNull_WhenAssertionFailsInsideScope()
{
    using var scope = new AssertScope();

    InvalidOperationException ex =
        Assert.Throws&lt;InvalidOperationException&gt;(() =&gt; { });

[...]

PR #9186, opened 2026-06-16
Document Throws/ContainsSingle null return inside Assert.Scope in RFC 011

Evangelink:
Follow-up to discussion on #9069.

RFC 011 documented that value-returning assertions return `null`/`default` when they fail inside an `Assert.Scope()`, but only named `ContainsSingle`. The `Assert.Throws`/`ThrowsExactly`/`ThrowsAsync`/`ThrowsExactlyAsync` family has the identical behavior (`return null!`) yet wasn't mentioned anywhere in the RFC, which is what made the behavior in #9069 look like a bug.

This PR:
- Adds the `Throws*` family alongside `ContainsSingle` in the value-returning assertion notes (Option 2 list and the chosen Option 3 `all assertions participate in soft collection` list).
- Adds a new **Value-returning assertions return `null`/`default` in scoped mode** subsection with a `Throws` example showing the `NullReferenceException`-as-secondary-symptom behavior.

No behavior change — documentation only. The corresponding user-facing docs are in dotnet/docs#54409.

### DIFF
--- a/docs/RFCs/011-Soft-Assertions-Nullability-Design.md
+++ b/docs/RFCs/011-Soft-Assertions-Nullability-Design.md
@@ -83,6 +83,7 @@ Categorize assertions into tiers based on whether their post-conditions narrow t
 - `Fail` — semantically means "unconditional failure"; annotated `[DoesNotReturn]` on public API
 - `Inconclusive` — semantically means "unconditional inconclusive"; annotated `[DoesNotReturn]` on public API and throws `AssertInconclusiveException` (not `AssertFailedException`)
 - `ContainsSingle` — returns the matched element; returning `default` in soft mode would give callers a bogus `null`/`default(T)` causing downstream errors
+- `Throws<T>` / `ThrowsExactly<T>` / `ThrowsAsync<T>` / `ThrowsExactlyAsync<T>` — return the caught exception; returning `null` in soft mode would give callers a bogus `null` causing downstream errors (same shape as `ContainsSingle`)
 
 **Tier 2 — Soft, but annotations removed:** Assertions that had conditional `[DoesNotReturnIf]` annotations. The annotation is removed so the compiler no longer assumes the condition is guaranteed. The assertions become soft (collected within a scope).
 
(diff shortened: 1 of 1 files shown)


0101 (approved):

docs/RFCs/011-Soft-Assertions-Nullability-Design.md
```diff
@@ -182,6 +183,28 @@ All assertions participate in soft failure collection, with the following except
...
 
+#### Value-returning assertions return `null`/`default` in scoped mode
+
+Some assertions return a value on success:
+
+- `Assert.ContainsSingle(...)` returns the single matched element.
+- `Assert.Throws<T>(...)`, `Assert.ThrowsExactly<T>(...)`, `Assert.ThrowsAsync<T>(...)`, and `Assert.ThrowsExactlyAsync<T>(...)` return the caught exception.
+
+These are soft within a scope, like every other assertion (except `Fail`/`Inconclusive`). When such an assertion *fails* inside a scope, the failure is collected and execution continues — so the method returns `null`/`default(T)` rather than throwing. The non-null return contract (and any nullable annotation) is intentionally not enforced in scoped mode, exactly as described above for all other postconditions.
```
Evangelink, line 193: **[NIT] Documentation Accuracy**

The parenthetical "(except `Fail`/`Inconclusive`)" omits the second class of hard exceptions already documented in the bullet points immediately above: null precondition checks (`CheckParameterNotNull`) inside Assert APIs also throw directly rather than collecting. A reader encountering this sentence in isolation could reasonably conclude that only `Fail`/`Inconclusive` are exceptions to the soft-collection rule.

**Recommendation:** Expand to keep the statement self-contained:

```
These are soft within a scope, like every other assertion (except `Fail`/`Inconclusive` and null precondition checks).
```

Alternatively, add a parenthetical cross-reference: "(see exceptions listed above)" to signal explicitly that the full list is already documented.

Evangelink (commented):
> [!NOTE]
> 🤖 **Automated review by GitHub Copilot.** Posted via a maintainer's GitHub token, so it appears under their account — the account owner did **not** write or approve this content personally. Generated by the [Expert Code Review workflow](https://github.com/microsoft/testfx/actions/runs/27624089630). To request a follow-up action, reply by tagging `@copilot` directly.

| # | Dimension | Verdict |
|---|-----------|---------|

✅ 21/22 dimensions clean.

- [ ] Documentation Accuracy — "(except `Fail`/`Inconclusive`)" parenthetical omits null precondition checks as a second hard-exception class (inline comment on line 193)

---

### Review notes (documentation-only PR)

Dimensions 1–16, 18–22 are **N/A** — this PR changes only `docs/RFCs/011-Soft-Assertions-Nullability-Design.md` with no code, tests, build scripts, or serialization formats.

**Dimension 17 — Documentation Accuracy:** The three changes are accurate and consistent with the implementation.

1. **Tier 1 bullet (Option 2, rejected) — line 87:** Correctly adds the `Throws*` family alongside `ContainsSingle` to the "returning `null` causes downstream errors" rationale in the rejected option's Tier 1 list. The phrasing "returning `null`" (rather than "`null`/`default(T)`") is intentionally tighter than the ContainsSingle entry — `T` is constrained to `Exception` (a reference type) so `default(T)` is always `null`. Accurate.

2. **Pros bullet (Option 3, chosen) — line 109:** Correctly adds all four async/sync variants to the enumerated list of soft assertions. Preserves the existing sort order (ContainsSingle → Throws family → IsTrue/IsFalse). Accurate.

3. **New subsection "Value-returning assertions return `null`/`default` in scoped mode" — lines 186–209:** The behaviour description and code example are accurate and confirmed by the implementation in `Assert.ThrowsException.Core.cs`, which does `return null!` after `ReportAssertFailed` returns (soft mode). The `using` form is appropriate for the synchronous `Throws<T>` example; the async variants (`ThrowsAsync<T>`, `ThrowsExactlyAsync<T>`) are named in the text. The guidance "move the call outside the scope, or restructure so nothing depends on the return value until after the scope is disposed" is both correct and actionable — if the assertion succeeds (exception thrown), the scope collects no failure and Dispose does not throw, so post-scope code can safely use the return value; if it fails, Dispose throws before post-scope code runs, providing the intended protection.

**One NIT** filed inline (see above): the "(except `Fail`/`Inconclusive`)" shorthand omits the also-documented `CheckParameterNotNull` precondition checks. This is a completeness suggestion, not a correctness problem — the full exception list is already given in the preceding bullet points.

> 🤖 **Automated content by GitHub Copilot.** Posted via a maintainer's GitHub token, so it appears under their account — the account owner did **not** write or approve this content personally. Generated by the [Expert Code Review (on PR ready)](https://github.com/microsoft/testfx/actions/runs/27624089630/agentic_workflow) workflow. · 489.3 AIC · ⌖ 12.4 AIC · [◷]()

PR #9192, opened 2026-06-16
Stream discovered tests to the SDK for --list-tests json under --server (#8661)

### LINKED ISSUE #8661: `--list-tests json` produces no JSON output when running in `--server` mode
### Summary

In server mode (`--server dotnettestcli --dotnet-test-pipe <pipe>`), `--list-tests json` emits no JSON output. Both the buffering step and the emission step in `TerminalOutputDevice` short-circuit when `_isServerMode` is true.

### Background

[microsoft/testfx#8280](https://github.com/microsoft/testfx/pull/8280) added the `--list-tests json` argument and JSON discovery output via `TerminalOutputDevice`. The .NET SDK (`dotnet test`) always launches MTP processes with `--server dotnettestcli ...` so the SDK can read handshake/discovery messages over a named pipe. As a result, the JSON output path is never triggered when using `dotnet test --list-tests json`.

### Where

[`src/Platform/Microsoft.Testing.Platform/OutputDevice/Terminal/TerminalOutputDevice.cs`](https://github.com/microsoft/testfx/blob/main/src/Platform/Microsoft.Testing.Platform/OutputDevice/Terminal/TerminalOutputDevice.cs):

- `ConsumeAsync` (around L534):
```csharp
if (_isServerMode)
{
    return Task.CompletedTask;
}
```
`_discoveredTestsForJson` is never populated when running under `--server`.

- `DisplayAfterSessionEndRunAsync` (around L419):
```csharp
if (_isServerMode)
{
    return;
}
```
The JSON document is never written when running under `--server`.

### Repro

```
dotnet test --list-tests json
```
against any MTP test project: no JSON is emitted.

### Suggested direction

Allow the JSON buffering + emission paths to run in server mode as well (gated on `--list-tests json`), or expose the JSON document over the dotnet-test-pipe protocol so the SDK can render it. Some care is needed to avoid double emission when both controller and test host roles run — today the JSON path is restricted to `_processRole == TestProcessRole.TestHost` (line ~442), which should continue to apply.

### Related

- dotnet/sdk#49754 — request to expose `--list-tests json` from `dotnet test`.
- dotnet/sdk#54490 — SDK CLI surface to forward `--list-tests {text|json}` (pending this fix to produce JSON output).

Evangelink:
Fixes #8661.

## Problem
In server mode (`--server dotnettestcli --dotnet-test-pipe <pipe>`) — the mode the .NET SDK always uses for `dotnet test` — `--list-tests json` emitted no JSON output, because `TerminalOutputDevice` short-circuits on `_isServerMode`.

## Approach (per review feedback from @Youssef1313)
Rather than having the test host serialize the JSON and write it to stdout under `--server`, the test host **streams the discovered tests to the SDK over the dotnet-test pipe**, and the SDK is responsible for producing the JSON — combining the discovered tests from every test app into a single document. This is more reliable, keeps the output ownership in one place, and naturally handles multi-project runs.

## Changes (testfx side)
- `TerminalOutputDevice`: restored the plain `if (_isServerMode) return` guards so the terminal device stays silent under `--server`. The standalone `app --list-tests json` path (no SDK involved) is unchanged and still emits the JSON document itself.
- `DotnetTestDataConsumer`: always streams the full discovery details (file location, method identifier, traits) over the pipe, not only for IDEs, so the SDK has the complete object to build the JSON.
- Acceptance test (`DotnetTestPipeListTestsJsonTests`): now asserts the host keeps stdout/stderr clean under `--server dotnettestcli` and that the discovered tests are streamed as `DiscoveredTestMessages` frames over the pipe (decoded via the black-box `DotnetTestPipeProtocol` reader).

## Follow-up (SDK side)
The matching change on dotnet/sdk (build the `--list-tests json` document from the streamed `DiscoveredTestMessages`) is the consuming half and will be done separately. Related: dotnet/sdk#54490.

## Verification
- All 4 `DotnetTestPipe` acceptance tests pass.
- 21 `TerminalOutputDevice` / `DiscoveredTest` platform unit tests pass.
- Debug build + `-pack` succeed with 0 warnings.

### DIFF
--- a/src/Platform/Microsoft.Testing.Platform/OutputDevice/TerminalOutputDevice.cs
+++ b/src/Platform/Microsoft.Testing.Platform/OutputDevice/TerminalOutputDevice.cs
@@ -455,7 +455,11 @@ internal sealed partial class TerminalOutputDevice : IHotReloadPlatformOutputDev
 
     public async Task DisplayAfterSessionEndRunAsync(CancellationToken cancellationToken)
     {
-        if (_isServerMode)
+        // In --list-tests json mode we must still emit the JSON document, even under --server
+        // (e.g. when launched by `dotnet test` with `--server dotnettestcli`). The JSON is written
+        // to stdout below, which the dotnet-test SDK captures and forwards. Only the genuine server
+        // run (no JSON discovery) short-circuits here, because its output flows through the pipe.
+        if (_isServerMode && !_isListTestsJson)
         {
             return;
         }
@@ -588,7 +592,12 @@ internal sealed partial class TerminalOutputDevice : IHotReloadPlatformOutputDev
     {
         RoslynDebug.Assert(_terminalTestReporter is not null);
         cancellationToken.ThrowIfCancellationRequested();
-        if (_isServerMode)
+
+        // In --list-tests json mode we still need to buffer discovered tests, even under --server
+        // (e.g. when launched by `dotnet test` with `--server dotnettestcli`), so the JSON document
+        // can be emitted at the end of the session. Only the genuine server run (no JSON discovery)
+        // short-circuits here, because its data flows through the pipe instead of the terminal.
+        if (_isServerMode && !_isListTestsJson)
         {
             return Task.CompletedTask;
         }
--- a/src/Platform/Microsoft.Testing.Platform/OutputDevice/TerminalOutputDevice.cs
+++ b/src/Platform/Microsoft.Testing.Platform/OutputDevice/TerminalOutputDevice.cs
@@ -455,11 +455,11 @@ internal sealed partial class TerminalOutputDevice : IHotReloadPlatformOutputDev
 
     public async Task DisplayAfterSessionEndRunAsync(CancellationToken cancellationToken)
     {
-        // In --list-tests json mode we must still emit the JSON document, even under --server
-        // (e.g. when launched by `dotnet test` with `--server dotnettestcli`). The JSON is written
-        // to stdout below, which the dotnet-test SDK captures and forwards. Only the genuine server
-        // run (no JSON discovery) short-circuits here, because its output flows through the pipe.
-        if (_isServerMode && !_isListTestsJson)
+        // Under --server (e.g. `dotnet test` with `--server dotnettestcli`) the terminal device stays
+        // silent: discovered tests are streamed to the SDK through the dotnet-test pipe (see
+        // DotnetTestDataConsumer), and the SDK owns rendering — including building the --list-tests json
+        // document by combining the discovered tests from every test app into a single output.
+        if (_isServerMode)
         {
             return;
         }
@@ -593,11 +593,10 @@ internal sealed partial class TerminalOutputDevice : IHotReloadPlatformOutputDev
         RoslynDebug.Assert(_terminalTestReporter is not null);
         cancellationToken.ThrowIfCancellationRequested();
 
-        // In --list-tests json mode we still need to buffer discovered tests, even under --server
-        // (e.g. when launched by `dotnet test` with `--server dotnettestcli`), so the JSON document
-        // can be emitted at the end of the session. Only the genuine server run (no JSON discovery)
-        // short-circuits here, because its data flows through the pipe instead of the terminal.
-        if (_isServerMode && !_isListTestsJson)
+        // Under --server (e.g. `dotnet test` with `--server dotnettestcli`) the terminal device does not
+        // buffer or render anything: data flows to the SDK through the dotnet-test pipe instead, and the
+        // SDK is responsible for producing the output (including the --list-tests json document).
+        if (_isServerMode)
         {
             return Task.CompletedTask;
         }
--- a/src/Platform/Microsoft.Testing.Platform/ServerMode/DotnetTest/IPC/DotnetTestDataConsumer.cs
+++ b/src/Platform/Microsoft.Testing.Platform/ServerMode/DotnetTest/IPC/DotnetTestDataConsumer.cs
@@ -54,15 +54,13 @@ internal sealed class DotnetTestDataConsumer : IPushOnlyProtocolConsumer
                 switch (testNodeDetails.State)
                 {
                     case TestStates.Discovered:
-                        TestFileLocationProperty? testFileLocationProperty = null;
-                        TestMethodIdentifierProperty? testMethodIdentifierProperty = null;
-                        TestMetadataProperty[] traits = [];
-                        if (_dotnetTestConnection.IsIDE)
-                        {
-                            testFileLocationProperty = testNodeUpdateMessage.TestNode.Properties.SingleOrDefault<TestFileLocationProperty>();
-                            testMethodIdentifierProperty = testNodeUpdateMessage.TestNode.Properties.SingleOrDefault<TestMethodIdentifierProperty>();
-                            traits = testNodeUpdateMessage.TestNode.Properties.OfType<TestMetadataProperty>();
-                        }
+                        // Always stream the full discovery details (file location, method identifier,
+                        // traits) to the SDK — not only for IDEs. `dotnet test --list-tests json` is
+                        // produced on the SDK side by combining the discovered tests from every test
+                        // app into a single document, so the SDK needs the complete object here.
+                        TestFileLocationProperty? testFileLocationProperty = testNodeUpdateMessage.TestNode.Properties.SingleOrDefault<TestFileLocationProperty>();
+                        TestMethodIdentifierProperty? testMethodIdentifierProperty = testNodeUpdateMessage.TestNode.Properties.SingleOrDefault<TestMethodIdentifierProperty>();
+                        TestMetadataProperty[] traits = testNodeUpdateMessage.TestNode.Properties.OfType<TestMetadataProperty>();
 
                         DiscoveredTestMessages discoveredTestMessages = new(
                             _executionId,
--- a/src/Platform/Microsoft.Testing.Platform/ServerMode/DotnetTest/IPC/DotnetTestDataConsumer.cs
+++ b/src/Platform/Microsoft.Testing.Platform/ServerMode/DotnetTest/IPC/DotnetTestDataConsumer.cs
@@ -54,13 +54,20 @@ internal sealed class DotnetTestDataConsumer : IPushOnlyProtocolConsumer
                 switch (testNodeDetails.State)
                 {
                     case TestStates.Discovered:
-                        // Always stream the full discovery details (file location, method identifier,
-                        // traits) to the SDK — not only for IDEs. `dotnet test --list-tests json` is
-                        // produced on the SDK side by combining the discovered tests from every test
-                        // app into a single document, so the SDK needs the complete object here.
-                        TestFileLocationProperty? testFileLocationProperty = testNodeUpdateMessage.TestNode.Properties.SingleOrDefault<TestFileLocationProperty>();
-                        TestMethodIdentifierProperty? testMethodIdentifierProperty = testNodeUpdateMessage.TestNode.Properties.SingleOrDefault<TestMethodIdentifierProperty>();
-                        TestMetadataProperty[] traits = testNodeUpdateMessage.TestNode.Properties.OfType<TestMetadataProperty>();
+                        // Only stream the full discovery details (file location, method identifier,
+                        // traits) when the consumer asked for them. We reuse the existing IsIDE flag
+                        // for that — despite the name, it is the handshake signal a consumer sets when
+                        // it wants the complete discovery object (e.g. an IDE, or the SDK when running
+                        // `dotnet test --list-tests json`). Plain runs keep the payload minimal.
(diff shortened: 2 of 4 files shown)


Evangelink:
### 🧪 Test quality grade — PR #9192

<table>
  <thead>
  </thead>
  <tbody>
    <tr>
      <td>new</td>
      <td><code>DotnetTestPipeListTestsJsonTests.<br>DotnetTestPipe_<br>ListTestsJson_<br>EmitsJsonDocumentToStdoutUnderServerMode</code></td>
      <td>A</td>
      <td>90–100</td>
      <td>Six meaningful assertions (exit code, stdout JSON shape, empty stderr, schema version, both discovered test names); clear AAA structure; no anti-patterns.</td>
    </tr>
  </tbody>
</table>

`/grade-tests`.</sub>



> 🤖 **Automated content by GitHub Copilot.** Posted via a maintainer's GitHub token, so it appears under their account — the account owner did **not** write or approve this content personally. Generated by the [Grade Tests on PR (on open / sync)](https://github.com/microsoft/testfx/actions/runs/27644748580/agentic_workflow) workflow. · 140.4 AIC · ⌖ 13.1 AIC · [◷]( · [◷](https://github.com/search?q=repo%3Amicrosoft%2Ftestfx+%22gh-aw-workflow-call-id%3A+microsoft%2Ftestfx%2Fgrade-tests-on-pr.agent%22&type=issues))

Evangelink (commented):
> [!NOTE]
> 🤖 **Automated review by GitHub Copilot.** Posted via a maintainer's GitHub token, so it appears under their account — the account owner did **not** write or approve this content personally. Generated by the [Expert Code Review workflow](https://github.com/microsoft/testfx/actions/runs/27644748529). To request a follow-up action, reply by tagging `@copilot` directly.

---

**Review notes (informational, not blocking):**

The fix is a minimal, surgical change: two `if (_isServerMode)` early-return guards promoted to `if (_isServerMode && !_isListTestsJson)` so the JSON-discovery path remains active under `--server dotnettestcli`. The logic is consistent across all flag combinations:

| `_isServerMode` | `_isListTestsJson` | Old `ConsumeAsync`/`DisplayAfterSessionEndRunAsync` | New behavior |
|---|---|---|---|
| `false` | `false` | pass-through (unchanged) | pass-through (unchanged) ✓ |
| `false` | `true` | pass-through (unchanged) | pass-through (unchanged) ✓ |
| `true` | `false` | early-return (unchanged) | early-return (unchanged) ✓ |
| `true` | `true` | **early-return (bug)** | pass-through → JSON buffered and emitted ✓ |

**Dimension-by-dimension:**

1. **Algorithmic Correctness** — ✅ All four `(_isServerMode, _isListTestsJson)` cases produce the correct result. `DisplayBeforeSessionStartAsync` still short-circuits on `_isServerMode || _isListTestsJson`, so `_terminalTestReporter.TestExecutionStarted/AssemblyRunStarted` are not called in server+json mode — but the JSON emit path in `DisplayAfterSessionEndRunInternalAsync` calls only `_console.WriteLine(DiscoveredTestsJsonSerializer.Serialize(...))` inside the monitor lock, never touching the reporter, so this is safe.
2. **Threading & Concurrency** — ✅ `_discoveredTestsForJson` writes (single-producer from the message bus pump) are still unguarded per the existing design comment. The read inside `DisplayAfterSessionEndRunInternalAsync` is still guarded by `_asyncMonitor`. No new shared-state exposure.
3. **Security** — ✅ JSON emitted to stdout (captured by the SDK); no path traversal, no leaked secrets, no argument injection.
4. **Public API** — ✅ `TerminalOutputDevice` is `internal sealed`. No `PublicAPI.Unshipped.txt` changes needed.
5. **Performance** — ✅ The hot path (`_isServerMode=true && _isListTestsJson=false`) still hits the early-return immediately; the added `!_isListTestsJson` boolean test is negligible overhead.
6. **Cross-TFM** — ✅ No new API usage; test asset targets `$(NetCurrent)`, consistent with other DotnetTestPipe tests.
7. **Resource Management** — ✅ `JsonDocument` is properly disposed with `using var`. `NamedPipeServerStream` in `FakeDotnetTestSdk` is wrapped in `using`.
8. **Defensive Coding** — ✅ No defensive coding added for internal invariants. `RoslynDebug.Assert(_terminalTestReporter is not null)` preserved.
9. **Localization** — ✅ Only code comments added, no user-facing strings.
10. **Test Isolation** — ✅ Each test invocation uses a `Guid.NewGuid()` pipe name; no shared file-system paths; no static mutation.
11. **Assertion Quality** — ✅ Test covers: exit code (equality), stdout shape (boolean+string), stderr absence (negative), JSON schema version (equality), and both test display names (collection). Five distinct assertion categories for one test method.
12. **Flakiness** — ✅ Pipe server created before process starts; cancellation via `TestContext.CancellationToken`; no `Thread.Sleep`.
13. **Test Completeness** — ✅ Existing `DotnetTestPipe_ChildEmitsNoStdoutOrStderrForPassingRun_Baseline` guards the regression (server-only mode still emits no stdout). New test covers the fixed path.
14. **Data-Driven** — N/A (acceptance test, single scenario appropriate).
15. **Code Structure** — ✅ Guard-clause idiom; condition `_isServerMode && !_isListTestsJson` is idiomatic and mirrors the intent.
16. **Naming** — ✅ `DotnetTestPipeListTestsJsonTests`, `DotnetTestPipe_ListTestsJson_EmitsJsonDocumentToStdoutUnderServerMode`, `DiscoveredTestsJsonSchemaVersion` all follow codebase conventions.
17. **Documentation** — ✅ Comments explain *why* (not just what); `DiscoveredTestsJsonSchemaVersion` constant has an accurate doc comment explaining the intentional duplication and the coupling it creates.
18. **Analyzer Quality** — N/A (no `src/Analyzers/` changes).
19. **IPC Wire Compatibility** — ✅ No wire-format changes; `DotnetTestPipeProtocol` is untouched.
20. **Build Infrastructure** — ✅ New test file picked up by SDK wildcard glob; no new package references.
21. **Scope Discipline** — ✅ Two files, one concern, cross-references issue #8661.
22. **PowerShell Hygiene** — N/A (no `.ps1` changes).

> 🤖 **Automated content by GitHub Copilot.** Posted via a maintainer's GitHub token, so it appears under their account — the account owner did **not** write or approve this content personally. Generated by the [Expert Code Review (on PR ready)](https://github.com/microsoft/testfx/actions/runs/27644748529/agentic_workflow) workflow. · 614.5 AIC · ⌖ 12.5 AIC · [◷]()

src/Platform/Microsoft.Testing.Platform/OutputDevice/TerminalOutputDevice.cs
```diff
@@ -472,7 +455,11 @@ public async Task DisplayAfterHotReloadSessionEndAsync(CancellationToken cancell
 
     public async Task DisplayAfterSessionEndRunAsync(CancellationToken cancellationToken)
     {
-        if (_isServerMode)
+        // In --list-tests json mode we must still emit the JSON document, even under --server
+        // (e.g. when launched by `dotnet test` with `--server dotnettestcli`). The JSON is written
+        // to stdout below, which the dotnet-test SDK captures and forwards. Only the genuine server
+        // run (no JSON discovery) short-circuits here, because its output flows through the pipe.
+        if (_isServerMode && !_isListTestsJson)
```
Youssef1313: I think it's much more reliable to have this done on protocol level or SDK.

At the end, I think SDK should be responsible for printing the output.

SDK should also combine the tests from all test apps into a single json object.

Evangelink: Agreed, this makes total sense. I pushed a commit that pivots to this approach:

- `TerminalOutputDevice` no longer renders anything under `--server` (restored the plain `if (_isServerMode) return` guards). The standalone `app --list-tests json` path — where there is no SDK — is unchanged.
- `DotnetTestDataConsumer` now always streams the full discovery details (file location, method identifier, traits) over the dotnet-test pipe, not just for IDEs, so the SDK has the complete object.
- The SDK becomes responsible for printing the output and combining the tests from all test apps into a single JSON document.

The acceptance test now asserts the host keeps stdout/stderr clean under `--server dotnettestcli` and that the discovered tests are streamed as `DiscoveredTestMessages` frames over the pipe. The matching SDK-side change (building the JSON from those messages) will be a follow-up on dotnet/sdk.

src/Platform/Microsoft.Testing.Platform/ServerMode/DotnetTest/IPC/DotnetTestDataConsumer.cs
```diff
@@ -54,15 +54,13 @@ public async Task ConsumeAsync(IDataProducer dataProducer, IData value, Cancella
...
-                        TestFileLocationProperty? testFileLocationProperty = null;
-                        TestMethodIdentifierProperty? testMethodIdentifierProperty = null;
-                        TestMetadataProperty[] traits = [];
-                        if (_dotnetTestConnection.IsIDE)
-                        {
-                            testFileLocationProperty = testNodeUpdateMessage.TestNode.Properties.SingleOrDefault<TestFileLocationProperty>();
-                            testMethodIdentifierProperty = testNodeUpdateMessage.TestNode.Properties.SingleOrDefault<TestMethodIdentifierProperty>();
-                            traits = testNodeUpdateMessage.TestNode.Properties.OfType<TestMetadataProperty>();
-                        }
+                        // Always stream the full discovery details (file location, method identifier,
```
Youssef1313: Instead of "always", could we have SDK asking for that info only when running with `--list-tests json`?

Youssef1313: We already have that concept, it's named "IsIDE" but that's more of an implementation detail and unfortunate naming (though I wouldn't rename this)

Evangelink: Good point — reverted the unconditional streaming. The full discovery details (file location, method identifier, traits) are again gated, and I reused the existing `IsIDE` flag as the "consumer wants the complete discovery object" signal (the SDK would set it when running `--list-tests json`). Plain runs keep the `DiscoveredTestMessages` payload minimal (Uid + DisplayName). Fixed in 480a169d7.

Youssef1313 (commented):
Looks like no product change is needed here and the only changes needed will be on SDK side only.

test/IntegrationTests/Microsoft.Testing.Platform.Acceptance.IntegrationTests/DotnetTestPipe/DotnetTestPipeListTestsJsonTests.cs
```diff
@@ -0,0 +1,141 @@
...
+    public TestContext TestContext { get; set; } = null!;
+
+    [TestMethod]
+    public async Task DotnetTestPipe_ListTestsJson_StreamsDiscoveredTestsOverPipeAndKeepsStdoutClean()
+    {
+        var testHost = TestInfrastructure.TestHost.LocateFrom(
+            AssetFixture.TargetAssetPath, AssetName, TargetFrameworks.NetCurrent);
+
+        FakeDotnetTestSdkResult result = await FakeDotnetTestSdk.RunAsync(
+            testHost, extraArguments: "--list-tests json", cancellationToken: TestContext.CancellationToken);
```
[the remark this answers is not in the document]
Evangelink: Good point — the test now runs with isIde: true (commit 58bec7b15), so the host advertises IsIDE and streams the full discovery object, matching the intended dotnet test --list-tests json scenario.

test/IntegrationTests/Microsoft.Testing.Platform.Acceptance.IntegrationTests/DotnetTestPipe/DotnetTestPipeListTestsJsonTests.cs
```diff
@@ -0,0 +1,141 @@
...
+            foreach (string displayName in DotnetTestPipeProtocol.DecodeDiscoveredTestDisplayNames(frame.Body))
+            {
+                discoveredDisplayNames.Add(displayName);
+            }
+        }
+
+        Assert.IsTrue(sawDiscoveredFrame, "Expected at least one DiscoveredTestMessages frame over the dotnet-test pipe.");
+        Assert.Contains("Test1", discoveredDisplayNames);
+        Assert.Contains("Test2", discoveredDisplayNames);
+    }
```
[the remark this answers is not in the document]
Evangelink: Fixed in 58bec7b15. The test asset's test nodes now publish `TestFileLocationProperty`, `TestMethodIdentifierProperty` and `TestMetadataProperty` (traits), and the harness decodes the full `DiscoveredTestMessage` (via the new `DecodeDiscoveredTests`) and asserts file path / namespace / type / method name / traits arrive over the pipe — so a regression where those fields stop flowing now fails the test. Verified end-to-end with a passing run.

0101 (approved):

PR #9193, opened 2026-06-16
Apply DebuggerDisableUserUnhandledExceptionsAttribute to Assert.Throws helpers

### LINKED ISSUE #9016: Investigate using DebuggerDisableUserUnhandledExceptionsAttribute on throw asserts

Evangelink:
Fixes #9016

When running with **Just My Code** enabled, Visual Studio raises a *user-unhandled exception* on the expected exception thrown inside an `Assert.Throws*` call, because the catch lives in MSTest framework code. This makes the debugger pause on tests that are working as intended.

.NET 9 introduced `DebuggerDisableUserUnhandledExceptionsAttribute` specifically for this scenario: a supporting debugger (VS 17.12+) will not break on user-unhandled exceptions caught by an attributed method. xUnit applies it the same way.

### Changes
* Adds a polyfill for the attribute (no-op on TFMs < net9.0), mirroring the `StackTraceHiddenAttribute` polyfill pattern.
* Applies `[DebuggerDisableUserUnhandledExceptions]` to the two private helpers that own the try/catch on user code:
  - `Assert.IsThrowsFailing<TException>`
  - `Assert.IsThrowsAsyncFailingAsync<TException>`

All `Assert.Throws[Exactly][Async]` overloads (and the interpolated string handler structs) route through these two helpers, so all entry points are covered without touching the public API.

Scope is deliberately limited to `Assert.Throws*` -- general test method execution is left alone so users still get the expected debugger break on unexpected exceptions from their tests.


Evangelink:
Closing as duplicate — this work was already merged via #9041 (the DebuggerDisableUserUnhandledExceptions attribute is in main, with the polyfill now under src/Polyfills/ and the helpers in Assert.ThrowsException.Core.cs after a later refactor). This PR was opened from a stale local branch that still held the pre-squash commit.

PR #9209, opened 2026-06-17
Reference System.Memory 4.6.3 in the .NETFramework adapter

Evangelink:
On `net462`, `MSTestAdapter.PlatformServices` references `System.Memory` (the linked XxHash code uses `Span`/`MemoryMarshal`). That reference is currently `4.0.1.2`, pulled in transitively as `System.Memory 4.5.5` via `Microsoft.TestPlatform.ObjectModel` -> `System.Reflection.Metadata 8.0.0` -> `System.Collections.Immutable 8.0.0`.

Hosts that unify `System.Memory` to `4.0.5.0` (e.g. Visual Studio) and run the VSTest adapter under a test host that does not apply the test source's binding redirects (the VSTest test host loads its own config) cannot satisfy the `4.0.1.2` reference: the deployed `4.0.5.0` is rejected by exact-version assembly resolution and discovery fails with `FileNotFoundException` for `System.Memory 4.0.1.2`.

This pins `System.Memory` to `4.6.3` (assembly `4.0.5.0`) and references it explicitly so the `net462` adapter assemblies bind `4.0.5.0`, matching what those hosts deploy. The reference is conditioned on `net462`/`netstandard2.0`/`uwp`; modern .NET (net8.0+) uses the in-box `Span` and is unaffected.

Verified by rebuilding `net462`: `MSTestAdapter.PlatformServices.dll` now references `System.Memory 4.0.5.0` (was `4.0.1.2`).

### DIFF
--- a/Directory.Packages.props
+++ b/Directory.Packages.props
@@ -35,6 +35,12 @@
          Tracking context: https://github.com/microsoft/testfx/issues/9091 (Task 5). -->
     <MicrosoftTestingInternalFrameworkVersion>1.5.0-preview.24577.4</MicrosoftTestingInternalFrameworkVersion>
     <SystemThreadingTasksExtensionsVersion>4.5.4</SystemThreadingTasksExtensionsVersion>
+    <!-- System.Memory flows in transitively (via System.Reflection.Metadata / System.Collections.Immutable)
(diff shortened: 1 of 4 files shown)


0101 (approved):
Automated safety check passed: no dangerous changes detected (no secrets, exfiltration, destructive ops, or suspicious dependencies). Approving as requested. Note: this is a quick safety sanity check, not a full code review.

PR #9218, opened 2026-06-17
Share dotnet test wire contract (ObjectFieldIds + Constants) as source

Evangelink:
## What

Establish a **source-shared** single source of truth for the `dotnet test` named-pipe protocol **wire contract**.

`ObjectFieldIds.cs` (serializer-id + field-id registry) and `Constants.cs` (handshake property names, execution modes, session-event types, test states, protocol version) are duplicated **by hand** in `dotnet/sdk` — where its `ObjectFieldIds` literally carries a *"must be kept aligned with the testfx repo"* warning. Any drift is a silent wire-protocol break.

These two files are **zero-dependency** (plain `internal` consts, no `[Embedded]`, no `Extensions.Messages`), so they can be **compiled into another assembly via a shared `.props`** instead of copied.

## Changes

- **`DotnetTestProtocolContract.props`** — shared-source manifest that `<Compile>`-includes `ObjectFieldIds.cs` + `Constants.cs`. Microsoft.Testing.Platform does **not** import it (it already globs these files); it is for external consumers.
- **`Microsoft.Testing.Platform.DotnetTestProtocolContract.UnitTests`** — a standalone consumer that does **not** reference the platform's protocol types. It compiles the shared source into its **own assembly** and pins every serializer id (incl. reserved id `4`), handshake property name, execution mode, session-event type, test state and the protocol version. This is the proof that the contract is self-contained and consumable with one source of truth.
- Registered the new project in `TestFx.slnx`.
- Filled gaps in the in-repo contract tests (`ProtocolTests.cs`): round-trip `CommandLineOptionMessages` / `FileArtifactMessages` / `TestSessionEvent` and `SerializerIds_AreStable` pinning the full id registry.

## Why only the contract (not models/serializers) — yet

While prototyping a fuller share, I verified hard couplings that block the models/serializers today:
- The `[Embedded]` transport (`NamedPipe*`, `IRequest`/`IResponse`/`INamedPipeSerializer`) is **per-assembly, not runtime-unified** (confirmed via an `InvalidCastException`) and is shared platform-wide.
- `DiscoveredTestMessage` couples to `TestMetadataProperty` (`Microsoft.Testing.Platform.Extensions.Messages`).
- testfx serializers use a different class shape (`NamedPipeSerializer<T>`) than the SDK copy (`BaseSerializer` + `INamedPipeSerializer`).

[...]

PR #9231, opened 2026-06-18
Add source NuGet package for the dotnet test wire contract (follow-up to #9218)

Evangelink:
## What

Follow-up #1 to #9218 (which established the shared `dotnet test` wire-contract **source** via `DotnetTestProtocolContract.props`).

This wires that source-share into a real **source-only NuGet package**, `Microsoft.Testing.Platform.DotnetTestProtocol`, so `dotnet/sdk`'s `dotnet test` implementation can **compile** `ObjectFieldIds.cs` + `Constants.cs` from a single source of truth instead of hand-copying them (where any drift is a silent wire-protocol break).

## Changes

- **New package project** `src/Platform/Microsoft.Testing.Platform.DotnetTestProtocol/`:
  - Ships the two zero-dependency contract files as `contentFiles/cs/any/...` with `BuildAction=Compile`, so the consumer compiles them into its **own** assembly (the `internal` types are then visible with no `InternalsVisibleTo`).
  - `IncludeBuildOutput=false` ⇒ pure source package, no `lib/`.
  - `PACKAGE.md` documents scope and that it is a `dotnet test` implementation detail, not for direct end-user use.
- **Refactor `DotnetTestProtocolContract.props`** to declare the canonical file list **once** as `@(DotnetTestProtocolContractSource)`. Both the source-share `<Compile>` (used by the standalone contract test project) and the package's `contentFiles` derive from that item, so the NuGet and the in-repo source can never drift.
- **Registered** the project in `TestFx.slnx`.

## Verification

- `build.cmd -pack` produces `Microsoft.Testing.Platform.DotnetTestProtocol.2.3.0-dev.nupkg` (0 warnings, 0 errors). The `.nuspec` carries both files under `<contentFiles ... buildAction="Compile" />`.
- An **external consumer** project (no reference to `Microsoft.Testing.Platform`) restores the produced package and compiles, accessing `Microsoft.Testing.Platform.IPC.CommandLineOptionMessagesFieldsId` and `TestStates` from the shared source — **0 warnings, 0 errors**.
- The existing standalone contract test project (`Microsoft.Testing.Platform.DotnetTestProtocolContract.UnitTests`) still builds across `net8.0` / `net9.0` / `net462` after the `.props` refactor.

## Open question for reviewers

The package is currently `IsShipping=false` — i.e. a **transport/source** package flowed cross-repo via dependency flow, **not** published to nuget.org. This seemed the idiomatic choice for an internal-only source contract consumed solely by `dotnet/sdk`. If you'd rather publish it publicly, it's a one-line flip (`IsShipping=true` + drop `DevelopmentDependency`). Happy to change.

## Follow-ups (still open from #9218)

2. Decouple `DiscoveredTestMessage` from `TestMetadataProperty` and converge the serializer shape, then extend this package to the message **models + serializers**.
3. Switch `dotnet/sdk` to consume this package and delete its hand-copied `ObjectFieldIds`.

### DIFF
--- a/src/Platform/Microsoft.Testing.Platform.DotnetTestProtocol/Microsoft.Testing.Platform.DotnetTestProtocol.csproj
+++ b/src/Platform/Microsoft.Testing.Platform.DotnetTestProtocol/Microsoft.Testing.Platform.DotnetTestProtocol.csproj
@@ -1,6 +1,14 @@
 <Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
+    <!--
+      netstandard2.0 is deliberate, NOT net8/9/10: this is a source-only package. The contract files are shipped as
+      contentFiles/cs/any and are compiled into the CONSUMER's compilation under the consumer's own TFM (e.g. net11
+      in dotnet/sdk), so this TargetFramework never reaches consumers. It only governs the throwaway lib build we use
+      as a "the shared source compiles standalone" guard - and netstandard2.0 makes that guard maximally portable
+      (and needs no runtime packs). Bump it only if a future addition (models/serializers) genuinely needs a newer
+      API; even then it would not change what consumers compile against.
+    -->
     <TargetFramework>netstandard2.0</TargetFramework>
 
     <!--
--- a/src/Platform/Microsoft.Testing.Platform.DotnetTestProtocol/Microsoft.Testing.Platform.DotnetTestProtocol.csproj
+++ b/src/Platform/Microsoft.Testing.Platform.DotnetTestProtocol/Microsoft.Testing.Platform.DotnetTestProtocol.csproj
@@ -1,0 +1,49 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFramework>netstandard2.0</TargetFramework>
+
+    <!--
+      Source-only package: it ships the 'dotnet test' wire-contract source files (ObjectFieldIds.cs + Constants.cs)
+      as contentFiles so a consumer (the dotnet/sdk 'dotnet test' implementation) compiles them into its OWN
+      assembly. This keeps a single source of truth for the protocol ids/constants instead of hand-copying them.
+
+      The project still compiles the two files (into a lib that is intentionally NOT packed - IncludeBuildOutput is
+      false) so we get a build-time guarantee that the shared contract compiles standalone.
+    -->
+    <IsPackable>true</IsPackable>
+    <IncludeBuildOutput>false</IncludeBuildOutput>
+    <SuppressDependenciesWhenPacking>true</SuppressDependenciesWhenPacking>
+    <DevelopmentDependency>true</DevelopmentDependency>
+
+    <!--
+      Transport/source package: consumed cross-repo (dotnet/sdk) via dependency flow, not meant to be installed by
+      end users from nuget.org. Flip to 'true' (and remove DevelopmentDependency) if public publication is desired.
+    -->
+    <IsShipping>false</IsShipping>
+
+    <!-- NU5128: source-only package has no lib/ assembly, only contentFiles. -->
+    <NoWarn>$(NoWarn);NU5128</NoWarn>
+
+    <PackageDescription>
+      <![CDATA[Microsoft Testing is a set of platform, framework and protocol intended to make it possible to run any test on any target or device.
+
+This package shares - as compiled source - the 'dotnet test' named-pipe wire contract (serializer/field ids and handshake/session/state constants) so that the protocol has a single source of truth across the microsoft/testfx and dotnet/sdk repositories. It is an implementation detail of 'dotnet test' integration and is not intended for direct end-user consumption.]]>
+    </PackageDescription>
+  </PropertyGroup>
+
+  <!--
+    Pulls in @(DotnetTestProtocolContractSource) (the canonical contract file list) and, by default, compiles those
+    files. The package layout below packs the SAME item as contentFiles, so the NuGet and the in-repo source-share
+    can never drift.
+  -->
+  <Import Project="$(RepoRoot)src\Platform\Microsoft.Testing.Platform\ServerMode\DotnetTest\DotnetTestProtocolContract.props" />
+
+  <ItemGroup>
+    <None Include="@(DotnetTestProtocolContractSource)"
+          Pack="true"
+          BuildAction="Compile"
+          PackagePath="contentFiles/cs/any/DotnetTestProtocol/%(Filename)%(Extension)" />
+  </ItemGroup>
+
+</Project>
--- a/src/Platform/Microsoft.Testing.Platform.DotnetTestProtocol/Microsoft.Testing.Platform.DotnetTestProtocol.csproj
+++ b/src/Platform/Microsoft.Testing.Platform.DotnetTestProtocol/Microsoft.Testing.Platform.DotnetTestProtocol.csproj
@@ -4,10 +4,9 @@
     <!--
       netstandard2.0 is deliberate, NOT net8/9/10: this is a source-only package. The contract files are shipped as
       contentFiles/cs/any and are compiled into the CONSUMER's compilation under the consumer's own TFM (e.g. net11
-      in dotnet/sdk), so this TargetFramework never reaches consumers. It only governs the throwaway lib build we use
-      as a "the shared source compiles standalone" guard - and netstandard2.0 makes that guard maximally portable
-      (and needs no runtime packs). Bump it only if a future addition (models/serializers) genuinely needs a newer
-      API; even then it would not change what consumers compile against.
+      in dotnet/sdk), so this TargetFramework never reaches consumers. This project does not compile the contract
+      files itself (see DotnetTestProtocolContractCompile below), so the TFM only governs the empty placeholder
+      assembly; netstandard2.0 keeps it the lightest and needs no runtime packs.
     -->
     <TargetFramework>netstandard2.0</TargetFramework>
 
@@ -16,17 +15,24 @@
       as contentFiles so a consumer (the dotnet/sdk 'dotnet test' implementation) compiles them into its OWN
       assembly. This keeps a single source of truth for the protocol ids/constants instead of hand-copying them.
 
-      The project still compiles the two files (into a lib that is intentionally NOT packed - IncludeBuildOutput is
-      false) so we get a build-time guarantee that the shared contract compiles standalone.
+      DotnetTestProtocolContractCompile=false makes the imported .props skip its <Compile> items: this project only
+      *packs* the files, it does not compile them. The guarantee that the shared contract compiles standalone is
+      provided by the dedicated consumer test project
+      (test/UnitTests/Microsoft.Testing.Platform.DotnetTestProtocolContract.UnitTests), not by this package build.
     -->
     <IsPackable>true</IsPackable>
     <IncludeBuildOutput>false</IncludeBuildOutput>
     <SuppressDependenciesWhenPacking>true</SuppressDependenciesWhenPacking>
+    <DotnetTestProtocolContractCompile>false</DotnetTestProtocolContractCompile>
+    <EnableDefaultCompileItems>false</EnableDefaultCompileItems>
+
+    <!-- Build-only / source contribution: it must not flow as a transitive runtime dependency to consumers. -->
     <DevelopmentDependency>true</DevelopmentDependency>
 
     <!--
-      Transport/source package: consumed cross-repo (dotnet/sdk) via dependency flow, not meant to be installed by
-      end users from nuget.org. Flip to 'true' (and remove DevelopmentDependency) if public publication is desired.
+      IsShipping=false keeps this package OUT of nuget.org and in the internal (NonShipping) feed used for cross-repo
+      dependency flow (dotnet/sdk). NOTE: the exact internal-feed shipping mechanism still needs confirmation with
+      the Arcade team (see PR discussion).
     -->
     <IsShipping>false</IsShipping>
 
@@ -41,9 +47,9 @@ This package shares - as compiled source - the 'dotnet test' named-pipe wire con
   </PropertyGroup>
 
   <!--
-    Pulls in @(DotnetTestProtocolContractSource) (the canonical contract file list) and, by default, compiles those
-    files. The package layout below packs the SAME item as contentFiles, so the NuGet and the in-repo source-share
-    can never drift.
+    Pulls in @(DotnetTestProtocolContractSource) - the canonical contract file list. We set
+    DotnetTestProtocolContractCompile=false above, so the .props does NOT add <Compile> items here; this project only
+    packs that same item as contentFiles, so the NuGet and the in-repo source-share can never drift.
   -->
   <Import Project="$(RepoRoot)src\Platform\Microsoft.Testing.Platform\ServerMode\DotnetTest\DotnetTestProtocolContract.props" />
 
--- a/src/Platform/Microsoft.Testing.Platform/ServerMode/DotnetTest/DotnetTestProtocolContract.props
+++ b/src/Platform/Microsoft.Testing.Platform/ServerMode/DotnetTest/DotnetTestProtocolContract.props
@@ -25,18 +25,18 @@
       it (it compiles these files via its default source globbing); importing it there would double-include them.
 
         The single canonical list of contract files lives in the @(DotnetTestProtocolContractSource) item below.
-        - Source consumers (e.g. the standalone contract test project) get them as <Compile> items (the default).
-        - The Microsoft.Testing.Platform.DotnetTestProtocol package project sets
-          DotnetTestProtocolContractCompile=false and instead packs @(DotnetTestProtocolContractSource) as
-          contentFiles, so the NuGet and the source-share never drift.
+      - Source consumers (e.g. the standalone contract test project) get them as <Compile> items (the default).
+      - The Microsoft.Testing.Platform.DotnetTestProtocol package project sets
+        DotnetTestProtocolContractCompile=false and instead packs @(DotnetTestProtocolContractSource) as
+        contentFiles, so the NuGet and the source-share never drift.
   -->
   <ItemGroup>
-      <DotnetTestProtocolContractSource Include="$(MSBuildThisFileDirectory)IPC\ObjectFieldIds.cs" />
-      <DotnetTestProtocolContractSource Include="$(MSBuildThisFileDirectory)IPC\Constants.cs" />
+    <DotnetTestProtocolContractSource Include="$(MSBuildThisFileDirectory)IPC\ObjectFieldIds.cs" />
+    <DotnetTestProtocolContractSource Include="$(MSBuildThisFileDirectory)IPC\Constants.cs" />
   </ItemGroup>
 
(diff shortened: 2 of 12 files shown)


Youssef1313 (commented):
I would prefer to ensure that all types that ship in a source-only package are marked with `[Embedded]`

src/Platform/Microsoft.Testing.Platform.DotnetTestProtocol/Microsoft.Testing.Platform.DotnetTestProtocol.csproj
```diff
@@ -0,0 +1,49 @@
...
+      false) so we get a build-time guarantee that the shared contract compiles standalone.
+    -->
+    <IsPackable>true</IsPackable>
+    <IncludeBuildOutput>false</IncludeBuildOutput>
+    <SuppressDependenciesWhenPacking>true</SuppressDependenciesWhenPacking>
+    <DevelopmentDependency>true</DevelopmentDependency>
+
+    <!--
+      Transport/source package: consumed cross-repo (dotnet/sdk) via dependency flow, not meant to be installed by
+      end users from nuget.org. Flip to 'true' (and remove DevelopmentDependency) if public publication is desired.
```
Youssef1313: DevelopmentDependency has nothing to do with public publication.

Please ensure with Arcade folks what's the right way to ship only in internal feeds. I also think it's better to have "Internal" in the package name.

src/Platform/Microsoft.Testing.Platform.DotnetTestProtocol/Microsoft.Testing.Platform.DotnetTestProtocol.csproj
Youssef1313, line 1: Should this project have `<DotnetTestProtocolContractCompile>false</DotnetTestProtocolContractCompile>`

src/Platform/Microsoft.Testing.Platform.DotnetTestProtocol/Microsoft.Testing.Platform.DotnetTestProtocol.csproj
```diff
@@ -0,0 +1,57 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <!--
+      netstandard2.0 is deliberate, NOT net8/9/10: this is a source-only package. The contract files are shipped as
+      contentFiles/cs/any and are compiled into the CONSUMER's compilation under the consumer's own TFM (e.g. net11
+      in dotnet/sdk), so this TargetFramework never reaches consumers. It only governs the throwaway lib build we use
+      as a "the shared source compiles standalone" guard - and netstandard2.0 makes that guard maximally portable
+      (and needs no runtime packs). Bump it only if a future addition (models/serializers) genuinely needs a newer
+      API; even then it would not change what consumers compile against.
```
Youssef1313:
```suggestion
      (and needs no runtime packs).
```

src/Platform/Microsoft.Testing.Platform.DotnetTestProtocol/Microsoft.Testing.Platform.DotnetTestProtocol.csproj
```diff
@@ -0,0 +1,57 @@
...
+    -->
+    <TargetFramework>netstandard2.0</TargetFramework>
+
+    <!--
+      Source-only package: it ships the 'dotnet test' wire-contract source files (ObjectFieldIds.cs + Constants.cs)
+      as contentFiles so a consumer (the dotnet/sdk 'dotnet test' implementation) compiles them into its OWN
+      assembly. This keeps a single source of truth for the protocol ids/constants instead of hand-copying them.
+
+      The project still compiles the two files (into a lib that is intentionally NOT packed - IncludeBuildOutput is
+      false) so we get a build-time guarantee that the shared contract compiles standalone.
```
Youssef1313: To me, that should be covered by separate integration tests.

src/Platform/Microsoft.Testing.Platform/ServerMode/DotnetTest/DotnetTestProtocolContract.props
```diff
@@ -23,10 +23,20 @@
     USAGE
       Import this file from any project that needs the contract. Microsoft.Testing.Platform itself does NOT import
       it (it compiles these files via its default source globbing); importing it there would double-include them.
+
+        The single canonical list of contract files lives in the @(DotnetTestProtocolContractSource) item below.
+        - Source consumers (e.g. the standalone contract test project) get them as <Compile> items (the default).
+        - The Microsoft.Testing.Platform.DotnetTestProtocol package project sets
+          DotnetTestProtocolContractCompile=false and instead packs @(DotnetTestProtocolContractSource) as
```
Evangelink: **[NIT] Documentation Accuracy**

This comment is inaccurate. The package project (`Microsoft.Testing.Platform.DotnetTestProtocol.csproj`) does **not** set `DotnetTestProtocolContractCompile=false` — it never sets that property at all. As a result, the default condition in the props evaluates to `true` and `<Compile>` items ARE added alongside the `<None Pack="true">` items. This is intentional (the `.csproj` explicitly says "we get a build-time guarantee that the shared contract compiles standalone"), but the comment misrepresents the mechanism.

**Recommendation:** Correct the comment to describe what actually happens:

```xml
- The Microsoft.Testing.Platform.DotnetTestProtocol package project leaves compilation
  enabled (as a standalone-compile guard) and also packs @(DotnetTestProtocolContractSource)
  as contentFiles via <None Pack="true"> items, so the NuGet and the source-share never drift.
  Set DotnetTestProtocolContractCompile=false to suppress the <Compile> items for a project
  that only wants to pack without compiling.
```

src/Platform/Microsoft.Testing.Platform/ServerMode/DotnetTest/DotnetTestProtocolContract.props
```diff
@@ -23,10 +23,20 @@
...
+          contentFiles, so the NuGet and the source-share never drift.
   -->
   <ItemGroup>
-    <Compile Include="$(MSBuildThisFileDirectory)IPC\ObjectFieldIds.cs" Link="DotnetTestProtocol\ObjectFieldIds.cs" />
-    <Compile Include="$(MSBuildThisFileDirectory)IPC\Constants.cs" Link="DotnetTestProtocol\Constants.cs" />
+      <DotnetTestProtocolContractSource Include="$(MSBuildThisFileDirectory)IPC\ObjectFieldIds.cs" />
```
Evangelink: **[NIT] Naming & Conventions — Indentation**

All three new item elements (`<DotnetTestProtocolContractSource>` ×2, `<Compile>`) use 6-space indentation inside their `<ItemGroup>`, but `.editorconfig` specifies `indent_size = 2` for `.props` files. An item at level 2 (inside `<ItemGroup>` which is level 1) should be indented 4 spaces (2 + 2). The original `<Compile>` items that this PR replaces used the correct 4-space indentation.

**Recommendation:** Change `      <DotnetTestProtocolContractSource` → `    <DotnetTestProtocolContractSource` (6 → 4 spaces) for all three new items.

Evangelink (commented):
> [!NOTE]
> 🤖 **Automated review by GitHub Copilot.** Posted via a maintainer's GitHub token, so it appears under their account — the account owner did **not** write or approve this content personally. Generated by the [Expert Code Review workflow](https://github.com/microsoft/testfx/actions/runs/27746514363). To request a follow-up action, reply by tagging `@copilot` directly.

| # | Dimension | Verdict |
|---|-----------|---------|
| 16 | Naming & Conventions | 🟢 1 NIT |

✅ 20/22 dimensions clean (Threading, Security, Public API, Performance, Cross-TFM, Resource Mgmt, Defensive Coding, Localization, Test Isolation, Assertion Quality, Flakiness, Test Completeness, Data-Driven Coverage, Code Structure, Analyzer Quality, IPC Wire Compatibility, Build Infrastructure, Scope Discipline, PowerShell Hygiene not applicable or clean).

**MSBuild Authoring (supplemental):** Same indentation issue as dimension 16 above — no additional findings.

- [ ] **Naming & Conventions / MSBuild Authoring** — Items in both new `<ItemGroup>` blocks in `DotnetTestProtocolContract.props` use 6-space indentation; `.editorconfig` requires `indent_size = 2` for `.props` files, so level-2 items should be 4 spaces.
- [ ] **Documentation Accuracy** — Comment at line 29–31 of `DotnetTestProtocolContract.props` claims the package project "sets `DotnetTestProtocolContractCompile=false`", but the `.csproj` never sets that property. The actual behavior is: `<Compile>` items are active (intentional compilation guard) **and** `<None Pack="true">` items pack the same files as `contentFiles`. The comment should describe the hybrid mechanism or be corrected to match reality.

**Notes (not blocking):**
- The `PACKAGE.md` README is automatically wired up by `Directory.Build.targets` (which sets `PackageReadmeFile` when the file exists) — no explicit `<PackageReadmeFile>` property is needed in the `.csproj`. ✅
- The new project is registered in `TestFx.slnx` but not in `Microsoft.Testing.Platform.slnf`. If platform developers regularly work on or rebuild this package, consider adding it to the filtered solution; if it's a transport-only artifact consumed exclusively by the SDK flow, omitting it is fine.
- The `<IsPackable>true</IsPackable>` in the `.csproj` is redundant (already set by `src/Platform/Directory.Build.props`) but harmless.

> 🤖 **Automated content by GitHub Copilot.** Posted via a maintainer's GitHub token, so it appears under their account — the account owner did **not** write or approve this content personally. Generated by the [Expert Code Review (on PR ready)](https://github.com/microsoft/testfx/actions/runs/27746514363/agentic_workflow) workflow. · 895.5 AIC · ⌖ 12.5 AIC · [◷]()

src/Platform/Microsoft.Testing.Platform/ServerMode/DotnetTest/DotnetTestProtocolContract.props
```diff
@@ -23,10 +23,20 @@
     USAGE
       Import this file from any project that needs the contract. Microsoft.Testing.Platform itself does NOT import
       it (it compiles these files via its default source globbing); importing it there would double-include them.
+
+        The single canonical list of contract files lives in the @(DotnetTestProtocolContractSource) item below.
+        - Source consumers (e.g. the standalone contract test project) get them as <Compile> items (the default).
+        - The Microsoft.Testing.Platform.DotnetTestProtocol package project sets
+          DotnetTestProtocolContractCompile=false and instead packs @(DotnetTestProtocolContractSource) as
+          contentFiles, so the NuGet and the source-share never drift.
```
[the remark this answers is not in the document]
Evangelink: Resolved in d9ab77b5f: the package project now actually sets `DotnetTestProtocolContractCompile=false` (it is pack-only), so this comment now matches reality.

src/Platform/Microsoft.Testing.Platform/ServerMode/DotnetTest/DotnetTestProtocolContract.props
Evangelink: Fixed in d9ab77b5f by making the code match the comment: the package now sets `DotnetTestProtocolContractCompile=false` and only packs the files (the `<Compile>` items are suppressed). The standalone-compile guarantee is provided by the dedicated consumer test project.

src/Platform/Microsoft.Testing.Platform/ServerMode/DotnetTest/DotnetTestProtocolContract.props
Evangelink: Fixed in d9ab77b5f — all three new item elements are now 4-space indented per `.editorconfig` `indent_size=2`.

src/Platform/Microsoft.Testing.Platform.DotnetTestProtocol/Microsoft.Testing.Platform.DotnetTestProtocol.csproj
Evangelink: Yes — done in d9ab77b5f. The project now sets `DotnetTestProtocolContractCompile=false` and is pack-only (no compile).

src/Platform/Microsoft.Testing.Platform.DotnetTestProtocol/Microsoft.Testing.Platform.DotnetTestProtocol.csproj
Evangelink: Agreed — done in d9ab77b5f. The package no longer compiles the files; the compile guarantee is provided by the dedicated `Microsoft.Testing.Platform.DotnetTestProtocolContract.UnitTests` consumer project (which imports the props with the default `Compile=true`).

src/Platform/Microsoft.Testing.Platform.DotnetTestProtocol/Microsoft.Testing.Platform.DotnetTestProtocol.csproj
Evangelink: Incorporated in d9ab77b5f — reworked that comment; the netstandard rationale now ends with `needs no runtime packs` (the lib is no longer built as a guard, so the longer rationale was dropped).

src/Platform/Microsoft.Testing.Platform.DotnetTestProtocol/Microsoft.Testing.Platform.DotnetTestProtocol.csproj
Evangelink: Fixed the wording in d9ab77b5f: I no longer tie `DevelopmentDependency` to publication — it is documented as a build-only/non-transitive marker, and `IsShipping=false` is documented as what keeps the package off nuget.org (internal NonShipping feed). On the exact internal-feed mechanism and an `Internal` package name: agreed these need Arcade confirmation — I left a TODO-style note in the csproj and would rather confirm the desired final name + feed approach with you/Arcade before renaming, to avoid churn. Want me to go ahead with e.g. `Microsoft.Testing.Platform.DotnetTestProtocol.Internal` now?

Evangelink:
> I would prefer to ensure that all types that ship in a source-only package are marked with `[Embedded]`

Good point — but I'd like to align on the approach before doing it, because these two files aren't *only* shipped as source here:

- `ObjectFieldIds.cs` and `Constants.cs` are also compiled **directly** into `Microsoft.Testing.Platform` (via default globbing) and into the `Microsoft.Testing.Platform.MSBuild` package. Marking their types `[Embedded]` would change them in those assemblies too (embedded ⇒ compiler-generated + hidden across assembly boundaries), which affects the platform's own internal usage and its `InternalsVisibleTo` unit tests (`ProtocolTests`).
- `[Embedded]` requires `Microsoft.CodeAnalysis.EmbeddedAttribute` to be available in **every** compilation that includes the source — including the `dotnet/sdk` consuming compilation. So we'd also need to ship/define that attribute as part of this source package.
- The types are already `internal`, so they don't leak into the consumer's public API today; `[Embedded]` would additionally make them invisible cross-assembly and dedupe duplicate copies.

So this is feasible but cross-cutting. I see two options:

1. **Ship an `EmbeddedAttribute` in the source package** and apply `[Embedded]` to the contract types — then reconcile the platform + MSBuild-package compilations (which compile the same files) so they don't break.
2. Keep the types plain `internal` (they're non-public in the consumer regardless) and revisit `[Embedded]` if/when we extend the package to the message **models/serializers**.

Which would you prefer? I lean toward (2) for this contract-only PR and tackling `[Embedded]` as part of the models/serializers follow-up where it matters more, but happy to do (1) here if you'd rather. Not changing it yet pending your call.

Evangelink:
Follow-up on `[Embedded]` — I actually tried it (mark all ~24 contract types `[Embedded]` + add the `EmbeddedAttribute` polyfill to the standalone consumer), and it **breaks the build**, which nicely pinpoints the real constraint:

```
VoidResponseSerializer.cs(12,31): error CS0103: The name 'VoidResponseFieldsId' does not exist in the current context
  [Microsoft.Testing.Extensions.MSBuild / HangDump / Retry / TrxReport / ...]
```

Why: several extensions compile `IPC/Serializers/VoidResponseSerializer.cs` **as shared source** (it's itself `[Embedded]`), but they do **not** compile `ObjectFieldIds.cs` — they resolve `VoidResponseFieldsId` from `Microsoft.Testing.Platform` via **`InternalsVisibleTo`**. Marking `VoidResponseFieldsId` (and siblings) `[Embedded]` hides them across the assembly boundary, so every extension that source-shares `VoidResponseSerializer` can no longer see the field-id → `CS0103`.

So today the design is inconsistent: the *serializer* is source-shared + `[Embedded]`, but the *field-ids it uses* are IVT-shared plain `internal`. To make the field-ids/constants `[Embedded]`, they'd also have to be **source-shared** into every assembly that uses them (HangDump, Retry, TrxReport, Extensions.MSBuild, …), not handed out via IVT — i.e. a broader restructure of how these IPC bits are shared, well beyond this contract-package PR.

**Proposal:** keep the contract types plain `internal` in this PR (they're non-public in any consumer regardless), and tackle `[Embedded]` as part of a dedicated change that converts the field-id/constants sharing from IVT to source across the platform + extensions (ideally together with the models/serializers follow-up). I've reverted the `[Embedded]` experiment. WDYT?

Evangelink:
Update: implemented `[Embedded]` after all (commit 4e28ad6e5) — @Evangelink pointed out the obvious fix to my earlier "this would need a big restructure" hand-wringing: the consumers just need to **source-link** the ids instead of getting them via `InternalsVisibleTo`.

What I did:
- Marked all ~24 contract types in `ObjectFieldIds.cs` + `Constants.cs` `[Embedded]`.
- Source-linked `ObjectFieldIds.cs` into the 4 extensions that share `VoidResponseSerializer` (`HangDump`, `Retry`, `TrxReport`, `Microsoft.Testing.Extensions.MSBuild`) — they needed `VoidResponseFieldsId` and were getting it via IVT.
- Source-linked `ObjectFieldIds.cs` + `Constants.cs` into `Microsoft.Testing.Platform.UnitTests` (ProtocolTests references the ids/consts).
- The standalone contract test project now provides the `EmbeddedAttribute` polyfill, mirroring how an external consumer (dotnet/sdk, which already defines it) supplies it.

**Full repo build is green (0 warnings, 0 errors).**

Note: this source-links the whole dotnet-test `ObjectFieldIds.cs` into the 4 extensions even though they only use the 3 base transport ids (`VoidResponse`/`TestHostCompleted`/`TestHostProcessPID`). A cleaner variant would relocate those 3 base ids into the base `IPC/` folder (already source-linked everywhere) and keep the dotnet-test contract separate — happy to do that follow-up if you prefer.

Evangelink:
### 🧪 Test quality grade — PR #9231

No new or modified test methods were identified in the changed regions
of this PR. Nothing to grade.

<sub>Re-run with `/grade-tests`.</sub>



> 🤖 **Automated content by GitHub Copilot.** Posted via a maintainer's GitHub token, so it appears under their account — the account owner did **not** write or approve this content personally. Generated by the [Grade Tests on PR (on open / sync)](https://github.com/microsoft/testfx/actions/runs/27764468396/agentic_workflow) workflow. · 388.7 AIC · ⌖ 12.5 AIC · [◷]( · [◷](https://github.com/search?q=repo%3Amicrosoft%2Ftestfx+%22gh-aw-workflow-call-id%3A+microsoft%2Ftestfx%2Fgrade-tests-on-pr.agent%22&type=issues))

0101 (approved):

PR #9266, opened 2026-06-19
Share terminal multi-assembly summary counts + port SDK oracle test (3/n)

Evangelink:
## What

Slice **3/n** of the test-first terminal-reporter superset port. Adds the per-assembly listing to the **final** run summary: when `ShowAssembly` is set and more than one assembly ran, each assembly is listed under the run-level verdict with its result + compact counts. Additive (in-process leaves `ShowAssembly` off -> byte-identical).

Ports the SDK's `TestExecutionCompleted_WithMultipleAssemblies_PrintsPerAssemblyCountsInSummary` test (dotnet/sdk#52128), asserting `[+2/x1/?0]` and `[+5/x0/?2]`.

Stacked on **#9265**.

## Verification
- Platform clean net8.0/net9.0/netstandard2.0 (0 warnings).
- Full unit suite green: **1201, 0 failed**.

> Review/merge #9263 -> #9265 first.

PR #9267, opened 2026-06-19
Share terminal retry counting model + port SDK oracle test (4/n)

Evangelink:
## What

Slice **4/n** (the high-risk one) of the test-first terminal-reporter superset port: adopt the SDK's **instanceId-based retry counting model** so retried assemblies render the `/r{N}` segment and the summary reflects the latest attempt's results.

Stacked on **#9266**. This completes the four reporter behaviors the `dotnet/sdk` orchestrator relies on (handshake, per-assembly counts, multi-assembly summary, retry).

## How

- `TestProgressState` now derives `PassedTests`/`FailedTests`/`SkippedTests` from per-test reported results keyed by `(testNodeUid, attempt)`, and `TotalTests` is computed. A new instance id (registered via `NotifyHandshake` on `AssemblyRunStarted`) is a **retry attempt** that supersedes the previous attempt's tally and bumps `RetriedFailedTests`. Adds `ReportPassingTest`/`ReportSkippedTest`/`ReportFailedTest`, `NotifyHandshake`, `TryCount`.
- `TestCompleted` threads the **instance id** — the in-process host uses its single fixed execution id as the instance id (one attempt), the orchestrator overload passes the real one — and routes counting through the `Report*` methods.
- `AssemblyRunStarted` registers the attempt via `NotifyHandshake`.
- Updated the `SilenceDrivenHeartbeatRenderer` test helper to seed counts via `Report*` (the counters are no longer directly settable).

## Oracle test ported

`AssemblyRunCompleted_WhenTestsWereRetried_ShowsRetriedCount` — ported from the SDK's `TerminalTestReporterTests`: attempt 1 fails `flaky-1`, attempt 2 (new instance id, same execution id) passes it, so the final tally is `[+1/x0/?0/r1]`.

## Why it's safe for the in-process host

The in-process path reports each test once under a single fixed instance id, so the retry model collapses to the same single-attempt counts. The ~89 byte-exact in-process tests are the proof.

## Verification

- Platform clean on **net8.0 / net9.0 / netstandard2.0** (0 warnings incl. IDE/SA).
- Full `Microsoft.Testing.Platform.UnitTests` suite green: **1202 total, 0 failed** (incl. all 4 ported SDK oracle tests + the byte-exact in-process tests).
- Standalone `TerminalReporterContract` consumer compiles.

> Stacked PR — review/merge #9263 → #9265 → #9266 first.


src/Platform/Microsoft.Testing.Platform/OutputDevice/Terminal/TestProgressState.cs
```diff
@@ -1,14 +1,24 @@
...
 [Embedded]
 internal sealed class TestProgressState
 {
+    // Tracks the per-test-node tally and the attempt it belongs to, so retries (which re-report the same test node
+    // uid under a new instance id) replace rather than double-count the earlier attempt's result.
+    private readonly Dictionary<string, TestNodeInfoEntry> _testUidToResults = [];
+
+    // Ordered list of instance ids seen for this assembly. Each new instance id is a retry attempt. In most runs
+    // there is exactly one (no retry).
+    private readonly List<string> _orderedInstanceIds = [];
```
Evangelink, line 27: **[MODERATE] Threading & Concurrency**

`_testUidToResults` (a plain `Dictionary<string, TestNodeInfoEntry>`) and `_orderedInstanceIds` (a plain `List<string>`) — plus the `PassedTests`/`SkippedTests`/`FailedTests`/`RetriedFailedTests`/`TryCount` fields — have no synchronization. The review dimension rule flags any `Dictionary` or `List` modified from more than one thread without locking.

**Why it's safe today (but worth documenting):**
- `_assemblies` on `TerminalTestReporter` is `ConcurrentDictionary`, so retrieving a `TestProgressState` for a given `executionId` is thread-safe.
- In the in-process host the message pipeline has a single consumer, so `TestCompleted` for one assembly is always serialized.
- In the orchestrator each child-process IPC channel is also processed by a single consumer; multiple assemblies have distinct `executionId`s and therefore distinct `TestProgressState` instances.

**Risk:**
If a future orchestrator optimization ever processes multiple result batches from the same child process in parallel — or if anyone calls `ReportPassingTest`/`ReportFailedTest` for the same `executionId` from two threads — the dictionary and the plain-int counters would race.

**Recommendation:** Add a single-line XML doc comment (or an inline comment on the class) explicitly stating the single-threaded-access contract, e.g.:

```csharp
/// <remarks>
/// Not thread-safe. All methods must be called from the same thread that owns
/// the containing <see cref="TerminalTestReporter"/> pipeline consumer.
/// </remarks>
```

This documents the invariant, makes the assumption auditable, and makes it easy to detect if the threading model ever changes.

test/UnitTests/Microsoft.Testing.Platform.UnitTests/OutputDevice/Terminal/TerminalTestReporterTests.cs
```diff
@@ -1305,6 +1305,41 @@ public void TestExecutionCompleted_WithMultipleAssemblies_PrintsPerAssemblyCount
...
+        ReportOrchestratorTest(terminalReporter, assembly, executionId, instanceId: "inst-1", testUid: "flaky-1", TestOutcome.Fail);
+
+        // Attempt 2: a new instance id triggers a retry; the failing test now passes.
+        terminalReporter.AssemblyRunStarted(assembly, "net9.0", "x64", executionId, instanceId: "inst-2");
+        ReportOrchestratorTest(terminalReporter, assembly, executionId, instanceId: "inst-2", testUid: "flaky-1", TestOutcome.Passed);
+
+        terminalReporter.AssemblyRunCompleted(executionId, exitCode: 0, outputData: null, errorData: null);
+
+        string assemblyLine = GetAssemblySummaryLine(stringBuilderConsole.Output, assembly);
+        Assert.Contains("[+1/x0/?0/r1]", assemblyLine);
```
Evangelink: **[MODERATE] Test Completeness**

Two gaps worth filling in this test:

**1. `TryCount` is never asserted.**
`TryCount` is the key new observable (`> 1` means retries happened). Adding a check here verifies the new property at the boundary where it transitions from 1 to 2:

```csharp
// After both AssemblyRunStarted calls:
// (retrieve the TestProgressState via a suitable accessor or test hook)
// Assert.AreEqual(2, assemblyRun.TryCount);
```

Even asserting indirectly (via the `/r{N}` segment) doesn't test `TryCount` itself. Consider exposing it via a small test-only accessor or verifying in a `TestProgressState` unit test.

**2. Only 2 attempts are tested; the 3-attempt accumulation is not.**
`RetriedFailedTests` accumulates `value.Failed` on each retry transition. For 3 attempts (fail→fail→pass) the expected final state is `[+1/x0/?0/r2]`, which exercises the code path where `RetriedFailedTests` increments twice for the same UID. Without this, a regression that resets `RetriedFailedTests` instead of accumulating it would stay undetected.

Suggested additional test method:

```csharp
[TestMethod]
public void AssemblyRunCompleted_WhenTestsWereRetriedTwice_AccumulatesRetriedCount()
{
    // Attempt 1: fail; attempt 2: fail; attempt 3: pass → [+1/x0/?0/r2]
    ...
    terminalReporter.AssemblyRunStarted(assembly, "net9.0", "x64", executionId, instanceId: "inst-1");
    ReportOrchestratorTest(terminalReporter, assembly, executionId, "inst-1", "flaky-1", TestOutcome.Fail);
    terminalReporter.AssemblyRunStarted(assembly, "net9.0", "x64", executionId, instanceId: "inst-2");
    ReportOrchestratorTest(terminalReporter, assembly, executionId, "inst-2", "flaky-1", TestOutcome.Fail);
    terminalReporter.AssemblyRunStarted(assembly, "net9.0", "x64", executionId, instanceId: "inst-3");
    ReportOrchestratorTest(terminalReporter, assembly, executionId, "inst-3", "flaky-1", TestOutcome.Passed);
    ...
    Assert.Contains("[+1/x0/?0/r2]", assemblyLine);
}
```

src/Platform/Microsoft.Testing.Platform/OutputDevice/Terminal/TerminalTestReporter.TestCompletion.cs
```diff
@@ -25,6 +25,8 @@ internal void TestCompleted(
         FlatException[] flatExceptions = ExceptionFlattener.Flatten(errorMessage, exception);
         TestCompleted(
             executionId,
+            // In-process host: a single attempt, so the instance id is the (fixed) execution id.
+            instanceId: executionId,
```
Evangelink, line 29: **[NIT] Documentation — implicit contract between in-process host and `NotifyHandshake`**

The private `TestCompleted` overload hard-codes `instanceId: executionId` for the in-process host, which only works correctly if the in-process host also called `AssemblyRunStarted` with `instanceId == executionId` — otherwise `GetAttemptNumberFromInstanceId` throws `Unreachable`.

This implicit pairing is clear here but is not documented at the `AssemblyRunStarted` call site. A short comment such as:

```csharp
// In-process host MUST pass instanceId == executionId here; the in-process TestCompleted
// overload forwards executionId as the instanceId (single fixed attempt).
```

would make the contract explicit and prevent callers from accidentally passing a different value.

Evangelink (commented):
> [!NOTE]
> 🤖 **Automated review by GitHub Copilot.** Posted via a maintainer's GitHub token, so it appears under their account — the account owner did **not** write or approve this content personally. Generated by the [Expert Code Review workflow](https://github.com/microsoft/testfx/actions/runs/27842731096). To request a follow-up action, reply by tagging `@copilot` directly.

## PR #9267 — Share terminal retry counting model + port SDK oracle test (4/n)

| # | Dimension | Verdict |
|---|-----------|---------|
| 2 | Threading & Concurrency | 🟡 1 MODERATE |
| 13 | Test Completeness | 🟡 1 MODERATE |

✅ 19/22 dimensions clean.

- [ ] **Threading** — `_testUidToResults` (`Dictionary`) and `_orderedInstanceIds` (`List`) in `TestProgressState` are non-thread-safe but undocumented as single-threaded-only. The current architecture (single pipeline consumer per assembly) prevents actual races; the ask is a doc comment to make the invariant auditable and prevent future regressions.
- [ ] **Test Completeness** — the retry oracle test covers 2 attempts; the 3-attempt (fail→fail→pass) accumulation of `RetriedFailedTests` is untested. `TryCount` is also never directly asserted.
- [ ] **NIT** — `instanceId: executionId` in the in-process `TestCompleted` overload relies on the caller passing the same value to `AssemblyRunStarted`; that pairing is undocumented.

---

### Algorithmic correctness (✅)

The retry-deduplication state machine in `ReportGenericTestResult` is correct. Traced all four transitions:

| Previous attempt | Current attempt | `RetriedFailedTests` change | Counter change |
|---|---|---|---|
| (none) | any | +0 | +1 in new category |
| Fail | Pass | +prev.Failed (+1) | −1 fail, +1 pass |
| Fail | Fail | +prev.Failed (+1) | net 0 (old subtracted, new added) |
| Pass | Pass | +0 | net 0 |

`TotalTests` stays invariant across retries (correct: same test, new attempt). `RetriedFailedTests` accumulates only `value.Failed` so a pass→pass "retry" (degenerate case) correctly contributes 0.

### `TotalTests` refactor (✅)

Removing `asm.TotalTests++` from `TestDiscovered` in `TerminalTestReporter.Summary.cs` is safe: in discovery mode `TotalTests` is now `DiscoveredTests` (incremented via `DiscoveredTests++`); in execution mode it is `PassedTests + SkippedTests + FailedTests`. Both are correct.

### Performance (✅)

`_orderedInstanceIds.IndexOf` is O(N) where N = retry count (almost always ≤ 3). `_testUidToResults` is O(1) lookup. The static lambdas in `ReportPassingTest`/`ReportSkippedTest`/`ReportFailedTest` are compiler-cached (no per-call delegate allocation). No concerns.

### Public API surface (✅)

All new and modified types are `internal sealed`; no `PublicAPI.Unshipped.txt` changes needed.

test/UnitTests/Microsoft.Testing.Platform.UnitTests/OutputDevice/Terminal/TerminalTestReporterTests.cs
Evangelink: Both gaps are now covered:
- **TryCount asserted** — TestProgressState_NotifyHandshake_CountsEachNewInstanceIdAsAnAttempt asserts the 0→1→2→3 transitions directly (incl. the re-seen-instance no-op).
- **3-attempt accumulation** — added AssemblyRunCompleted_WhenTestsWereRetriedTwice_AccumulatesRetriedCount (fail→fail→pass), asserting the rendered [+1/x0/?0/r2] segment so a regression that resets instead of accumulates RetriedFailedTests is caught (c0d649f46).

src/Platform/Microsoft.Testing.Platform/OutputDevice/Terminal/TestProgressState.cs
Evangelink: Documented — TestProgressState now carries a THREADING comment stating it is intentionally not internally synchronized and that each instance is only mutated by the single pipeline consumer that owns its \xecutionId\ (do not call concurrently for the same assembly without adding synchronization).

src/Platform/Microsoft.Testing.Platform/OutputDevice/Terminal/TerminalTestReporter.TestCompletion.cs
Evangelink: Documented at the call site — the in-process \AssemblyRunStarted\ call in \TerminalOutputDevice.SessionLifecycle.cs\ passes \InProcessExecutionId\ for both executionId and instanceId, with a comment explaining that the in-process \TestCompleted\ overload forwards executionId as the instanceId so the attempt-number lookup only resolves if the same id was registered here.

PR #9291, opened 2026-06-20
Optimize terminal retry attempt lookup to O(1) (#9267 follow-up)

Evangelink:
Follow-up to a review comment on the now-merged #9267: [discussion r3445729371](https://github.com/microsoft/testfx/pull/9267#discussion_r3445729371).

## What

`TestProgressState.GetAttemptNumberFromInstanceId` ran a linear `List<string>.IndexOf` for **every** reported test result (`ReportGenericTestResult` calls it per completion) — avoidable overhead in a hot path for large runs.

## How

Replaced the ordered `List<string>` of instance ids with a `Dictionary<string,int>` mapping each instance id to its 1-based attempt number, maintained in `NotifyHandshake`. Both the handshake registration and the per-result attempt lookup are now O(1).

Behavior is unchanged: the highest attempt number always equals `TryCount`, so the "not the most recent attempt" guard and the 1-based attempt numbers are preserved.

## Verification

- `Microsoft.Testing.Platform` builds clean on net9.0 (0 warnings / 0 errors).
- Full `Microsoft.Testing.Platform.UnitTests` suite green: **1216 total, 0 failed** — including the retry oracle tests (`AssemblyRunCompleted_WhenTestsWereRetried_ShowsRetriedCount`, `...RetriedTwice_AccumulatesRetriedCount`) and `TestProgressState_NotifyHandshake_CountsEachNewInstanceIdAsAnAttempt`.

### DIFF
--- a/src/Platform/Microsoft.Testing.Platform/OutputDevice/Terminal/TestProgressState.cs
+++ b/src/Platform/Microsoft.Testing.Platform/OutputDevice/Terminal/TestProgressState.cs
@@ -14,7 +14,7 @@ internal sealed class TestProgressState
     // THREADING: this type is intentionally not internally synchronized. Each TestProgressState instance is owned by
     // a single executionId; the reporter looks it up from a ConcurrentDictionary (_assemblies), but the Microsoft
     // Testing Platform message pipeline delivers events for a given executionId on a single consumer, so the mutating
-    // members below (the dictionary/list and the Passed/Skipped/Failed/Retried/TryCount counters) are only ever
+    // members below (the dictionaries and the Passed/Skipped/Failed/Retried/TryCount counters) are only ever
     // touched by one thread at a time for a given instance. Do not call these members concurrently for the same
     // assembly without adding synchronization.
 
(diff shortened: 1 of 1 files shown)


Evangelink (commented):
> [!NOTE]
> 🤖 **Automated review by GitHub Copilot.** Posted via a maintainer's GitHub token, so it appears under their account — the account owner did **not** write or approve this content personally. Generated by the [Expert Code Review workflow](https://github.com/microsoft/testfx/actions/runs/27867910417). To request a follow-up action, reply by tagging `@copilot` directly.

**Summary:** The refactoring is correct and behavior-preserving. I traced through every call path in both the old `List<string>` and the new `Dictionary<string, int>` implementations and confirmed they are semantically identical:

- `NotifyHandshake`: the guard `index != _orderedInstanceIds.Count - 1` (old) and `attemptNumber != TryCount` (new) both detect the same condition — a handshake replay for a stale (non-current) attempt — and correctly throw `ApplicationStateGuard.Unreachable()` in both cases.
- `GetAttemptNumberFromInstanceId`: old returned `IndexOf + 1` (0-based → 1-based); new returns the stored value directly, which was set to `TryCount` at registration time (already 1-based). Equivalent.
- The threading contract is unchanged: `TestProgressState` is still intentionally unsynchronized, relying on the MTP message pipeline's single-consumer-per-executionId guarantee. The `Dictionary` has the same thread-safety requirements as the replaced `List`, so no regression.
- `ApplicationStateGuard.Unreachable()` is preserved in both call sites — no invariant guards were weakened.
- No public API surface touched; no `.resx`/`.xlf` impact; no cross-TFM concern (`Dictionary<TKey,TValue>` is available on all targets).

0101 (approved):

PR #9328, opened 2026-06-22
Share StackTraceHelper regex init across Platform/MSBuild (#9317)

### LINKED ISSUE #9317: [duplicate-code] Duplicate Code: StackTraceHelper regex initialization block duplicated across Platform and MSBuild projects with diverged patter
[Content truncated due to length]
*Analysis of commit `0aae71eb9f4386ea7c50c59572401cc1f060a3ef`*

**Assignee**: `@copilot`

## Summary

The localized-string lookup and regex-construction block inside `StackTraceHelper` is copy-pasted between the main `Microsoft.Testing.Platform` project and the `Microsoft.Testing.Platform.MSBuild` tasks project. The two copies cannot share a binary reference due to build layering, so the logic was duplicated. The regex patterns have **already diverged** between the two copies, meaning a frame that matches in one project's parsing may not match in the other's.

### Pattern: StackTraceHelper — regex initialization block (~35 identical lines) + diverged regex

- **Severity**: High  
- **Occurrences**: 2
- **Locations**:
  - `src/Platform/Microsoft.Testing.Platform/Helpers/StackTraceHelper.cs` (lines 17–61)
  - `src/Platform/Microsoft.Testing.Platform.MSBuild/Tasks/StackTraceHelper.cs` (lines 63–100)

**Duplicated initialization block** (near-verbatim in both files):
```csharp
string atResourceName = "Word_At";
string inResourceName = "StackTrace_InFileLineNumber";
string? atString = null;
string? inString = null;

try
{
    MethodInfo? getResourceStringMethod = typeof(Environment).GetMethod(
        "GetResourceString",
        BindingFlags.Static | BindingFlags.NonPublic, null, [typeof(string)], null);
    if (getResourceStringMethod is not null)
    {
        atString = (string?)getResourceStringMethod.Invoke(null, [atResourceName]);
        inString = (string?)getResourceStringMethod.Invoke(null, [inResourceName]);
    }
}
catch { /* If we fail, populate the defaults below. */ }

atString = atString == null || atString == atResourceName ? "at" : atString;
inString = inString == null || inString == inResourceName ? "in {0}:line {1}" : inString;

string inPattern = string.Format(CultureInfo.InvariantCulture, inString, "(?<file>.+)", @"(?<line>\d+)");
```

**Diverged regex patterns** (critical difference — lines 60 vs 99):
```csharp
// Microsoft.Testing.Platform version (src/Platform/Microsoft.Testing.Platform/Helpers/StackTraceHelper.cs:60)
// Two-branch pattern to also capture frames without file location:
s_regex = new Regex(@$"^   {atString} ((?<code>.+) {inPattern}|(?<code1>.+))$",
    RegexOptions.Compiled | RegexOptions.ExplicitCapture);  // NO timeout

// MSBuild Tasks version (src/Platform/Microsoft.Testing.Platform.MSBuild/Tasks/StackTraceHelper.cs:99)
// Single-branch pattern — only matches frames WITH file location:
s_regex = new Regex(@$"^   {atString} (?<code>.+) {inPattern}$",
    RegexOptions.Compiled, matchTimeout: TimeSpan.FromSeconds(1));  // HAS timeout
```

The MSBuild copy also implements `TryFindLocationFromStackFrame` and `TryGetStackFrameLocation` (~45 lines total, lines 10–60) as standalone methods that consume the regex, while the Platform version exposes `GetFrameRegex()` as a public method and leaves frame-level parsing to callers. This is an additional divergence in the public API shape.

## Impact Analysis

- **Maintainability**: Any improvement to the regex (e.g., handling VB.NET-localized stack frames, handling new runtime `at` keyword translations) requires updating both copies independently.
- **Bug Risk**: The MSBuild copy uses a single-branch regex that **only matches frames with file/line info**, while the MTP copy uses a two-branch pattern that also captures frames without location. A bug fix applied to one pattern will silently not apply to the other. The MSBuild copy also has a 1-second timeout not present in the MTP copy.
- **Code Bloat**: ~80 lines of duplicated initialization and frame-parsing logic across two projects.

## Refactoring Recommendations

1. **Extract the initialization block into a shared source file**
   - The `src/Platform/SharedExtensionHelpers/` project already exists as a shared internal library. Move the regex initialization logic to a helper there (e.g., `StackTraceRegexHelper.cs`).
   - Both `Microsoft.Testing.Platform` and `Microsoft.Testing.Platform.MSBuild` already depend on this shared project (or can reference it).
   - Estimated effort: 2–3 hours

2. **Align the regex patterns**
   - Decide which regex pattern is correct and apply it to both:
     - Two-branch `(code+location | code-only)` — more permissive, matches all frames
     - Single-branch `code+location` — MSBuild Tasks behavior, only matches frames with file info
   - Align the timeout policy: either both use no timeout (if the regex is provably linear) or both use a consistent timeout value.

3. **Align the frame-parsing API shape**
   - The Platform version exposes `GetFrameRegex()` and lets callers parse. The MSBuild version encapsulates parsing in `TryFindLocationFromStackFrame` / `TryGetStackFrameLocation`. Consider consolidating around one API shape in the shared helper.

## Implementation Checklist

- [ ] Compare the two regex patterns and decide on the canonical form
- [ ] Align the timeout policy across both copies
- [ ] Extract the ~35-line initialization block into `src/Platform/SharedExtensionHelpers/`
- [ ] Update both `StackTraceHelper` files to call the shared helper
- [ ] Run acceptance tests that exercise stack trace parsing (`HelpInfoTests`, error message tests)

## Analysis Metadata

- **Analyzed Files**: 2 (`StackTraceHelper.cs` in Platform and MSBuild Tasks)
- **Detection Method**: File name deduplication + semantic comparison + diff analysis
- **Commit**: `0aae71eb9f4386ea7c50c59572401cc1f060a3ef`
- **Analysis Date**: 2026-06-22

> 🤖 **Automated content by GitHub Copilot.** Posted via a maintainer's GitHub token, so it appears under their account — the account owner did **not** write or approve this content personally. Generated by the [Duplicate Code Detector](https://github.com/microsoft/testfx/actions/runs/27932903917/agentic_workflow) workflow. · 1.4K AIC · ⌖ 38.7 AIC · ⊞ 43.4K · [◷]( · [◷](https://github.com/search?q=repo%3Amicrosoft%2Ftestfx+is%3Aissue+%22gh-aw-workflow-call-id%3A+microsoft%2Ftestfx%2Fduplicate-code-detector%22&type=issues))
>
<details>

```
gh aw add githubnext/agentics/workflows/duplicate-code-detector.md@main
```
</details>

> - [x] expires  on Jun 24, 2026, 6:16 AM UTC

Evangelink:
Fixes #9317. Extracts the localized stack-frame resource lookup and regex pattern construction into shared source linked by Platform and MSBuild. This is behavior-preserving for each consumer: Platform keeps its two-branch pattern and no-timeout policy, while MSBuild keeps its location-only pattern and original 1-second timeout policy.


src/Platform/SharedExtensionHelpers/StackTraceRegexHelper.cs
```diff
@@ -0,0 +1,58 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+namespace Microsoft.Testing.Platform.Helpers;
+
+internal static class StackTraceRegexHelper
+{
+    internal const int MatchTimeoutMilliseconds = 1_000;
+
+    internal static TimeSpan MatchTimeout => TimeSpan.FromMilliseconds(MatchTimeoutMilliseconds);
```
Evangelink: **MINOR — Computed property allocates a new `TimeSpan` on every access**

`MatchTimeout` is a property that calls `TimeSpan.FromMilliseconds(...)` each time it is read. Callers read it once per lazy-init, so the practical impact is negligible, but by convention in this codebase shared-constant values like this should be `static readonly` fields:

```csharp
internal static readonly TimeSpan MatchTimeout = TimeSpan.FromMilliseconds(MatchTimeoutMilliseconds);
```

This avoids a heap allocation (boxing is not a concern for `TimeSpan` since it is a value type on the stack, but a property call still implies a method invocation) and makes the intent — "this is a fixed value computed once" — clearer.

src/Platform/Microsoft.Testing.Platform/Helpers/StackTraceHelper.cs
```diff
@@ -6,58 +6,25 @@ namespace Microsoft.Testing.Platform.Helpers;
 internal static partial class StackTraceHelper
 {
 #if NET7_0_OR_GREATER
-    // Specifying no timeout, the regex is linear. And the timeout does not measure the regex only, but measures also any
-    // thread suspends, so the regex gets blamed incorrectly.
-    [GeneratedRegex(@"^   at ((?<code>.+) in (?<file>.+):line (?<line>\d+)|(?<code1>.+))$", RegexOptions.ExplicitCapture)]
+    // Keep the location-less branch because terminal stack rendering preserves frames that do not include source locations.
+    [GeneratedRegex(@"^   at ((?<code>.+) in (?<file>.+):line (?<line>\d+)|(?<code1>.+))$", RegexOptions.ExplicitCapture, StackTraceRegexHelper.MatchTimeoutMilliseconds)]
```
Evangelink: **MINOR — `[GeneratedRegex]` hardcodes English; the `#else` path attempts localization**

The NET7+ source-generated path uses a compile-time-constant pattern that hard-codes `"at"` and `"in {0}:line {1}"`. The `#else` path passes through `StackTraceRegexHelper.CreateFrameRegexPattern(matchFramesWithoutLocation: true)`, which calls `GetLocalizedStackFrameRegexParts()` and attempts to retrieve the locale-specific strings via reflection.

This inconsistency was already present before this PR, but now that the localization logic lives in the shared helper it is more visible: the helper exists but is *not* used by the path that runs on all modern .NET targets. Consider adding a comment such as:

```csharp
// NOTE: GeneratedRegex requires a compile-time-constant pattern, so localization
// via GetLocalizedStackFrameRegexParts() is intentionally not used on this path.
// Pre-NET7 targets fall back to the #else branch which does attempt localization.
```

This makes the design decision explicit for future maintainers.

src/Platform/SharedExtensionHelpers/StackTraceRegexHelper.cs
```diff
@@ -0,0 +1,58 @@
...
+
+    internal static string CreateFrameRegexPattern(bool matchFramesWithoutLocation)
+    {
+        (string atString, string inPattern) = GetLocalizedStackFrameRegexParts();
+
+        return matchFramesWithoutLocation
+            ? @$"^   {atString} ((?<code>.+) {inPattern}|(?<code1>.+))$"
```
Evangelink: **MINOR — `atString` and `inPattern` are not `Regex.Escape`d before interpolation**

`atString` comes from the runtime's localised resource (e.g. `"at"`, `"bei"`, `"à"`) and `inPattern` is constructed from a localised template. Both are interpolated directly into the regex pattern string.

If a locale ever produces a string that contains a regex metacharacter (`.`, `+`, `(`, `[`, etc.) the generated pattern will silently misbehave rather than throw. The probability is very low for the specific resources used, and this was the existing behaviour before this PR. Still, now that the logic is centralised in one place this would be the right spot to add:

```csharp
var escapedAt = Regex.Escape(atString);
return matchFramesWithoutLocation
    ? @$"^   {escapedAt} ((?<code>.+) {inPattern}|(?<code1>.+))$"
    : @$"^   {escapedAt} (?<code>.+) {inPattern}$";
```

`inPattern` contains `(?<file>.+)` and `(?<line>\d+)` that were deliberately constructed as regex fragments, so those placeholders must *not* be escaped — only the `atString` portion should be.

src/Platform/SharedExtensionHelpers/StackTraceRegexHelper.cs
```diff
@@ -0,0 +1,58 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+namespace Microsoft.Testing.Platform.Helpers;
+
+internal static class StackTraceRegexHelper
+{
+    internal const int MatchTimeoutMilliseconds = 1_000;
```
Evangelink, line 13: **MAJOR — Removing the no-timeout design without replacing its rationale**

The code deleted from `Microsoft.Testing.Platform/Helpers/StackTraceHelper.cs` carried an explicit warning:

> "Specifying no timeout, the regex is linear. And the timeout does not measure the regex only, but measures also any thread suspends, so the regex gets blamed incorrectly."

That concern is real and still applies here. On .NET, `Regex.Match` measures *wall-clock elapsed time*, not CPU time. A thread executing a match can be preempted during a GC pause, a thread-pool saturation event, or a debugger stop, and then resume to find it has already exceeded the budget — even though the regex itself had barely started. Under those conditions all three callers will silently lose information:

| Caller | Fallback behaviour on timeout |
|---|---|
| `TerminalTestReporter.AppendStackFrame` | Raw, unstyled line appended |
| `AzureDevOpsReporter.GetStackFrameLocation` | Returns `null` → no location in the ADO result |
| `MSBuild StackTraceHelper.TryGetStackFrameLocation` | Returns `false` → no file/line annotation on the MSBuild error |

These failures are invisible to users and hard to correlate with GC pressure.

**Suggested mitigations (pick one):**

1. **Restore the no-timeout design** with the existing comment re-stated in `StackTraceRegexHelper`, and accept the theoretical (but, per the original analysis, practically unreachable because the pattern is linear) ReDoS risk.
2. **Keep the timeout but document it**: add an XML/block comment on `MatchTimeoutMilliseconds` explaining the GC-pause risk and why 1 s was chosen as an acceptable balance, so future maintainers don't remove the catch blocks thinking "this never fires."
3. **Validate the pattern is actually linear** using the .NET source-generator backtracking complexity analyser (`Regex.IsMatch` on NET8+ can report complexity), and if confirmed linear, remove the timeout with the rationale comment.

At minimum, please add back some form of the original rationale comment so the intent is not lost.

Evangelink (commented):
> [!NOTE]
> 🤖 **Automated review by GitHub Copilot.** Posted via a maintainer's GitHub token, so it appears under their account — the account owner did **not** write or approve this content personally. Generated by the [Expert Code Review workflow](https://github.com/microsoft/testfx/actions/runs/27942815565). To request a follow-up action, reply by tagging `@copilot` directly.

## Summary

This PR centralises duplicated stack-frame regex construction into a new `StackTraceRegexHelper` shared file, adds `RegexMatchTimeoutException` handling at all three call sites, and wires a 1-second timeout into the regex. The refactoring itself is clean and well-scoped. The main concern is the **behavioural change introduced by the timeout** — the original code deliberately omitted a timeout with a documented rationale (wall-clock timeout includes GC pauses / thread suspensions) and that rationale was removed rather than updated.

No BLOCKING issues found. Two MAJOR concerns.

---

## 22-Dimension Verdict

| # | Dimension | Severity | Status | Notes |
|---|-----------|----------|--------|-------|
| 1 | Algorithmic Correctness | MAJOR | ✅ PASS | Logic is preserved. Pattern semantics unchanged. Timeout fallbacks are all graceful. |
| 2 | Threading & Concurrency | BLOCKING | ✅ PASS (pre-existing) | `s_regex` lazy-init in both `InitializeRegex()` paths is an unlocked double-check; a benign race since reference writes are atomic and `Regex` construction is idempotent. Pre-dates this PR. |
| 3 | Memory & Performance | MAJOR | ⚠️ MINOR | `MatchTimeout` is a computed property (new `TimeSpan` struct on every call). In practice called once per lazy-init, so impact is negligible — but `static readonly TimeSpan` is the correct idiom here. See inline comment. |
| 4 | API Design & Backward Compatibility | BLOCKING | ✅ PASS | All new types are `internal`. No public API surface changes. No `PublicAPI.Unshipped.txt` entry needed. |
| 5 | Error Handling & Resilience | MAJOR | ✅ PASS | `RegexMatchTimeoutException` is caught at all three call sites with appropriate, graceful fallbacks. |
| 6 | Cross-TFM Correctness | BLOCKING | ✅ PASS | New file compiles on all targets. `#if NET7_0_OR_GREATER` guard is preserved. File-link approach (compile-time sharing) works correctly across both consuming projects. |
| 7 | Test Quality | MAJOR | ⚠️ CONCERN | The three new `catch (RegexMatchTimeoutException)` branches have zero test coverage. Existing tests in `TerminalTestReporterTests` exercise the happy path but not the timeout fallback. A test for `AppendStackFrame` with a mock `ITerminal` where the regex fires the timeout (e.g. by patching `s_regex` or by injecting a slow pattern) would confirm the fallback output is correct. Similarly no unit tests exist for `StackTraceRegexHelper.CreateFrameRegexPattern`. |
| 8 | Localization & Globalization | MAJOR | ⚠️ MINOR (pre-existing) | `[GeneratedRegex]` (NET7+) hardcodes English `"at"` and `"in"`. Pre-NET7 `#else` path calls `CreateFrameRegexPattern()` which attempts localization. Inconsistency pre-exists; see inline comment for documentation suggestion. |
| 9 | Security | BLOCKING | ✅ PASS | The 1-second timeout mitigates potential ReDoS on adversarial stack trace lines. `RS0030` suppression is correctly scoped. |
| 10 | Code Style & Conventions | MINOR | ⚠️ MINOR | `MatchTimeout` should be `static readonly TimeSpan` (see inline). All other conventions followed: `is null`/`is not null`, file-scoped namespace, `ExplicitCapture`. |
| 11 | Documentation & Comments | MINOR | ⚠️ MAJOR | The original comment explaining *why* there was no timeout was removed. That rationale (wall-clock timeout penalises GC-paused threads, not slow regexes) still applies. See inline comment on `StackTraceRegexHelper.cs:8`. |
| 12 | Dependency Management | MAJOR | ✅ PASS | No new NuGet dependencies. File-link pattern is established in this codebase. |
| 13 | Build System | MINOR | ✅ PASS | `<Compile Include>` correctly added to both `Microsoft.Testing.Platform.csproj` and `Microsoft.Testing.Platform.MSBuild.csproj`. |
| 14 | IPC Contract Stability | BLOCKING | N/A | No IPC protocol or wire-format changes. |
| 15 | Public API Surface | BLOCKING | ✅ PASS | `StackTraceRegexHelper` is `internal static`. No `PublicAPI.Unshipped.txt` changes needed. |
| 16 | Analyzer Rules Compliance | MAJOR | ✅ PASS | `#pragma warning disable/restore RS0030` correctly scoped around the reflection call. `ExplicitCapture` used. |
| 17 | Behavioral Regression Risk | MAJOR | ⚠️ MAJOR | Adding a 1-second wall-clock timeout is a behavioural change. Under GC pressure or thread-pool saturation the timeout can fire for a well-formed stack frame, causing silent loss of diagnostic information across all three callers (AzureDevOps test location, terminal syntax highlighting, MSBuild file/line annotation). The original author explicitly avoided a timeout for exactly this reason. See inline comment on `StackTraceRegexHelper.cs:8` for concrete mitigations. |
| 18 | Scope Discipline | MINOR | ✅ PASS | PR is tightly scoped: centralise regex helper + add timeout guards. |
| 19 | Naming Conventions | MINOR | ✅ PASS | `StackTraceRegexHelper`, `MatchTimeoutMilliseconds`, `CreateFrameRegexPattern`, `matchFramesWithoutLocation` all follow codebase conventions. |
| 20 | Null Safety | MAJOR | ✅ PASS | `[MemberNotNull(nameof(s_regex))]` preserved. `is null`/`is not null` pattern used throughout the diff. |
| 21 | Regex Safety | MAJOR | ✅ PASS | Timeout added defensively. `atString` interpolated without `Regex.Escape` (pre-existing; see inline comment on `StackTraceRegexHelper.cs:17`). `ExplicitCapture` reduces overhead. |
| 22 | Configuration & Feature Flags | MINOR | ⚠️ MINOR | The 1-second timeout is a hardcoded constant with no way to adjust it at runtime. Note that the Azure DevOps user-filter timeout (`StackFrameFilterMatchTimeoutMs = 500 ms`) is half as long and intended for user-supplied, potentially adversarial patterns. Having the internal, carefully-crafted pattern carry a *larger* timeout than user patterns is slightly counter-intuitive; documenting the rationale would help. |

---

## Actionable items

| Priority | Item |
|----------|------|
| **MAJOR** | Restore (or replace) the rationale comment for the timeout decision — see inline on `StackTraceRegexHelper.cs:8`. |
| **MAJOR** | Add unit tests for the `RegexMatchTimeoutException` fallback paths in `AppendStackFrame`, `GetStackFrameLocation`, and `TryGetStackFrameLocation`. |
| MINOR | Change `MatchTimeout` from a computed property to `static readonly TimeSpan` — see inline on `StackTraceRegexHelper.cs:10`. |
| MINOR | Document the NET7+ / pre-NET7 localization split — see inline on `StackTraceHelper.cs:10`. |
| MINOR | Consider `Regex.Escape(atString)` when building the pattern — see inline on `StackTraceRegexHelper.cs:17`. |

src/Platform/SharedExtensionHelpers/StackTraceRegexHelper.cs
```diff
@@ -0,0 +1,58 @@
...
+                atString = (string?)getResourceStringMethod.Invoke(null, [atResourceName]);
+
+                // <value>in {0}:line {1}</value>
+                inString = (string?)getResourceStringMethod.Invoke(null, [inResourceName]);
+            }
+        }
+        catch
+        {
+            // If we fail, populate the defaults below.
+        }
```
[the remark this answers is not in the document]
Evangelink: Done in 322094e5. Replaced the generic `catch` with a typed `catch (Exception ex) when (ex is AmbiguousMatchException or TargetInvocationException or TargetParameterCountException or MemberAccessException or InvalidOperationException or NotSupportedException)`, keeping the same fall-back-to-defaults behavior.

src/Platform/SharedExtensionHelpers/StackTraceRegexHelper.cs
Evangelink: Done in 322094e5. `MatchTimeout` is now a `static readonly` field.

src/Platform/Microsoft.Testing.Platform/Helpers/StackTraceHelper.cs
Evangelink: Done in 322094e5. Added a NOTE comment above the `[GeneratedRegex]` explaining that the source-generated path requires a compile-time-constant pattern, so the localized helper is intentionally not used there, and that pre-NET7 targets fall back to the localized `#else` branch.

src/Platform/SharedExtensionHelpers/StackTraceRegexHelper.cs
Evangelink: Done in 322094e5. `atString` is now `Regex.Escape`d before interpolation; `inPattern` is left untouched since it is built from intentional regex fragments.

src/Platform/SharedExtensionHelpers/StackTraceRegexHelper.cs
Evangelink: Addressed in 322094e5 (and the earlier c677d6a2 preserving the Platform no-timeout behavior). The Platform caller keeps the original no-timeout design with its rationale comment restored; only the MSBuild caller applies the 1s timeout (its original behavior). I documented `MatchTimeoutMilliseconds` per option 2: it now explains the wall-clock/GC-pause caveat, that the timeout is MSBuild-only, why 1s was chosen, and warns against removing the `RegexMatchTimeoutException` catch assuming it never fires.

0101 (approved):

Evangelink:
## 🔴 Build Failure Analysis

### Summary

The build fails with **4 errors**, all caused by a single root cause: `TerminalResources.cs.xlf` is out-of-date with `TerminalResources.resx`.

```
'OutputDevice/Terminal/xlf/TerminalResources.cs.xlf' is out-of-date with
'OutputDevice/Terminal/TerminalResources.resx'.
Run `msbuild /t:UpdateXlf` to update .xlf files.
```

### Root Cause

`TerminalResources.resx` (in `src/Platform/Microsoft.Testing.Platform/OutputDevice/Terminal/`) was recently updated on `main` with **8 new resource entries**, but the corresponding XLF localization files were never regenerated. When this PR merged `origin/main` into its branch (the merge commit `cc6b07a`), it picked up both the new `.resx` entries **and** the stale XLF state.

**Missing entries in `TerminalResources.cs.xlf`** (and all other language XLFs):

| Resource key | English value |
|---|---|
| `DiscoveredTestsInAssembly` | `Discovered {0} tests in assembly` |
| `DiscoveredTestsSummary` | `Discovered {0} tests in {1} assemblies.` |
| `DiscoveredTestsSummarySingular` | `Discovered {0} tests.` |
| `DiscoveringTestsFrom` | `Discovering tests from` |
| `Error` | `error` |
| `Retried` | `retried` |
| `RunningTestsFrom` | `Running tests from` |
| `Try` | `try {0}` |

**RESX entries:** 59 · **XLF trans-units:** 47 → **12 missing** (8 new keys + 4 from image/metadata nodes in the XML that aren't resource strings).

### Affected Projects

The `Microsoft.Testing.Platform.csproj` build failed, which cascaded failures into:
- `Microsoft.Testing.Extensions.AzureDevOpsReport.csproj`
- `Microsoft.Testing.Extensions.AzureFoundry.csproj`
- `Microsoft.Testing.Extensions.CrashDump.csproj`
- `Microsoft.Testing.Extensions.TrxReport.Abstractions.csproj`

### Fix

This is **not caused by the changes in this PR** (the PR only touches `StackTrace*` files). The stale XLF came from a recent main-branch commit. However, since the PR's head is a merge commit that already contains the main-branch state, you need to regenerate the XLF as part of this PR before it can merge cleanly.

Run the following on your branch, then commit the changed XLF files:

```bash
dotnet msbuild src/Platform/Microsoft.Testing.Platform/Microsoft.Testing.Platform.csproj /t:UpdateXlf
```

This will regenerate all 14 language XLF files under `src/Platform/Microsoft.Testing.Platform/OutputDevice/Terminal/xlf/`. As per the repo's localization guidelines, **do not edit the XLF files manually** — always use `UpdateXlf`.

Once the XLF files are regenerated and committed, the build should pass.



> 🤖 **Automated content by GitHub Copilot.** Posted via a maintainer's GitHub token, so it appears under their account — the account owner did **not** write or approve this content personally. Generated by the [Build Failure Analysis](https://github.com/microsoft/testfx/actions/runs/27950747577/agentic_workflow) workflow. · 447.9 AIC · ⌖ 12.4 AIC · ⊞ 46.9K · [◷]( · [◷](https://github.com/search?q=repo%3Amicrosoft%2Ftestfx+%22gh-aw-workflow-call-id%3A+microsoft%2Ftestfx%2Fbuild-failure-analysis%22&type=issues))

PR #9371, opened 2026-06-23
Work around MSB4216 NET task host drive-casing failure in CodeQL build

AlesProkop:
## Problem

The CodeQL pipeline (and other .NET Framework MSBuild host scenarios on certain hosted CI agents) fails with **MSB4216** when Arcade's `InstallDotNetCore` task launches the .NET task host:

```
error MSB4216: Could not run the "InstallDotNetCore" task because MSBuild could not
create or connect to a task host with runtime "NET" and architecture "x86".
```

Root cause (see dotnet/msbuild#14026): the task host handshake **salt** is a case-sensitive hash of the SDK tools directory. The .NET Framework host derives it from `$(NetCoreSdkRoot)` (drive-letter casing propagated verbatim from the environment, e.g. ADO's `D:\a\_work\1\s`), while the .NET task host child resolves its own path via `Environment.ProcessPath`, which on some agents reports the volume's canonical **lowercase** drive (`d:`). The salts then differ and the handshake fails. The failure is agent-specific, which is why it reproduces intermittently.

## Fix (temporary)

An `InitialTargets` target in `eng/Versions.props` rewrites **only the drive-letter casing** of `$(NetCoreSdkRoot)` to match the casing the child task host will use. It learns that casing by launching the SDK's own `dotnet` host (`dotnet.exe` — the runtime host, which always runs and lives on the same volume as the SDK) and reading `GetModuleFileNameW(NULL)`, the exact API behind the child's `Environment.ProcessPath`. The detected drive letter is adopted only when it is the **same letter** differing solely in case, so it can never change the actual drive — only its casing. Windows-only, cached, and a safe no-op on agents that already work.

Knobs:
- `DisableNetCoreSdkRootDriveCasingWorkaround=true` — skip entirely.
- `NetCoreSdkRootDriveCasingOverride=lower|upper` — force the casing without probing (emergency/testing).

## Validation

Reproduced and validated on the actual failing agent type via the `microsoft-testfx` CodeQL pipeline: the probe resolved `d:`, the target rewrote `NetCoreSdkRoot` `D:` → `d:`, and **MSB4216 was eliminated** (two independent runs that both landed on a "bad" agent went green; an `Override=lower` run also confirmed the salt reads the live, target-mutated property).

## Temporary / follow-up

This is a stopgap. The permanent fixes are:
- host-side: **dotnet/arcade#17039** (builds on @ViktorHofer's dotnet/arcade#17002) — once it flows into this repo's Arcade, this inline workaround can be removed;
- child-side: **dotnet/msbuild#14027** (handshake salt widening) flowing into the .NET SDK.

Remove this once both are in.

### DIFF
--- a/eng/Versions.props
+++ b/eng/Versions.props
@@ -1,4 +1,4 @@
-<Project>
+<Project InitialTargets="NormalizeNetCoreSdkRootCasing">
   <PropertyGroup Label="Version settings">
     <!-- MSTest version -->
     <VersionPrefix>4.3.0</VersionPrefix>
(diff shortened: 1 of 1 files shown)


AlesProkop:
@microsoft-github-policy-service agree company="Microsoft"

Evangelink:
Closing for now, I'll wait for the arcade fix to flow.

Issue #9404, opened 2026-06-24
Add a testconfig.json key reference page in docs/


Evangelink:
Follow-up to #9381 (Task 5).

No page in `docs/` lists all valid `mstest:*` (and `platformOptions:*`) testconfig.json keys with their types, defaults, and XML RunSettings equivalents. The new [`docs/testconfig.schema.json`](https://github.com/microsoft/testfx/blob/main/docs/testconfig.schema.json) provides machine-readable validation but not human-facing reference documentation.

### Proposal
Add `docs/testconfig-reference.md` with:
- A table of all `mstest:*` keys with type, default, and XML RunSettings equivalent (source of truth: `SetSettingsFromConfig` + `MSTestAdapterSettings.ToSettings` + `RunConfigurationSettings.SetRunConfigurationSettingsFromConfig`).
- A note on the architectural split: behavioral keys → `MSTestSettings`; deployment / assembly resolution / apartment state → `MSTestAdapterSettings` / `RunConfigurationSettings`.
- A note on `platformOptions:*` keys (`PlatformConfigurationConstants`).
- A canonical full-featured example JSON, with a `$schema` reference pointing at `docs/testconfig.schema.json`.

Cross-link from `docs/testconfig.schema.md`.

Set the issue **Type** field to `Task`.

Issue #9403, opened 2026-06-24
Clarify and document 0-value behavior for `timeout:*` testconfig.json keys


Evangelink:
Follow-up to #9381 (Task 4).

`ParseIntegerSetting` (`MSTestSettings.Configuration.cs`) requires `result > 0`, so setting e.g. `"mstest:timeout:test": 0` silently has no effect and only emits the generic `InvalidValue` warning. The XML parser uses the same `> 0` guard, so behavior is consistent between formats — the gap is documentation and error-message quality.

### Options
- **Option A** — Accept `0` as "no timeout / clear the value": change the guard to `result >= 0` (in both JSON and XML parsers for parity) and define the semantics.
- **Option B** — Keep the guard but improve the warning to say something like: "0 is not a valid timeout; omit the key to use the default (no timeout)."

Document the chosen behavior in the testconfig.json reference (see #9381 Task 5) and reflect any constraint changes in [`docs/testconfig.schema.json`](https://github.com/microsoft/testfx/blob/main/docs/testconfig.schema.json) (`timeout.*` currently has `minimum: 1`).

### Key files
- `src/Adapter/MSTestAdapter.PlatformServices/MSTestSettings.Configuration.cs`
- `docs/testconfig.schema.json`

Set the issue **Type** field to `Task`.

Issue #9402, opened 2026-06-24
Move `orderTestsByNameInClass` under `mstest:execution:` in testconfig.json for consistency


Evangelink:
Follow-up to #9381 (Task 2).

`orderTestsByNameInClass` is parsed from the flat `mstest:orderTestsByNameInClass` key (`MSTestSettings.Configuration.cs`, `SetSettingsFromConfig`) while all other behavioral controls (`randomizeTestOrder`, `mapInconclusiveToFailed`, `mapNotRunnableToFailed`, `treatDiscoveryWarningsAsErrors`, `considerEmptyDataSourceAsInconclusive`, `launchDebuggerOnAssertionFailure`) live under `mstest:execution:`.

### Proposal
- Read `orderTestsByNameInClass` from `mstest:execution:orderTestsByNameInClass`.
- Keep a backward-compatible read from the old flat `mstest:orderTestsByNameInClass` key, emitting a deprecation warning when it is used.
- Update the canonical `testconfig.json` example in `ConfigurationSettingsTests.cs`.
- Update [`docs/testconfig.schema.json`](https://github.com/microsoft/testfx/blob/main/docs/testconfig.schema.json) to move the key under `execution` (and, if we keep back-compat, allow the deprecated flat key too).

### Key files
- `src/Adapter/MSTestAdapter.PlatformServices/MSTestSettings.Configuration.cs`
- `test/IntegrationTests/MSTest.Acceptance.IntegrationTests/ConfigurationSettingsTests.cs`
- `docs/testconfig.schema.json`

Note: set the issue **Type** field to `Task` (do not add the deprecated `type/task` label).

Issue #9401, opened 2026-06-24
Add `TimeoutFromConfigTests` acceptance tests covering the JSON testconfig.json timeout path


Evangelink:
Follow-up to #9381 (Task 3).

`TimeoutFromRunSettingsTests` (`test/IntegrationTests/MSTest.Acceptance.IntegrationTests/`) covers all 8 fixture-type timeouts via XML `.runsettings` only. The `mstest:timeout:*` JSON config path is untested end-to-end.

### Proposal
Create a parallel `TimeoutFromConfigTests` that:
1. Generates a `testconfig.json` with the target `mstest:timeout:*` key set to a short deadline (e.g. 300 ms).
2. Runs the test host and asserts each fixture method is timed out with the correct message.
3. Covers the same 8 method kinds: `assemblyInitialize`, `assemblyCleanup`, `classInitialize`, base-class `classInitialize`, `classCleanup`, base-class `classCleanup`, `testInitialize`, `testCleanup`.

### Notes
- Acceptance tests run with assembly-level method parallelization; mark classes that share a single mutable generated asset `[DoNotParallelize]`.
- Do not hard-code `\(\d+ms\)` for rendered durations — use `AcceptanceAssert.DurationPattern`.
- Requires `.\build.cmd -pack` before running.

Set the issue **Type** field to `Task`.

PR #9405, opened 2026-06-24
Add testconfig.json JSON schema for SchemaStore publication

Evangelink:
## What

Adds a draft-07 JSON Schema for `testconfig.json`, the configuration file consumed by Microsoft.Testing.Platform (MTP) and MSTest, plus a maintainer doc.

Implements **Task 1** of #9381 (publish the MSTest testconfig.json schema to SchemaStore).

### Files
- `docs/testconfig.schema.json` — the schema. Covers the full `mstest:*` surface (`timeout`, `execution`, `parallelism`, `output`, `deployment`, `assemblyResolution`, `orderTestsByNameInClass`), the `platformOptions:*` host options, and the `environmentVariables` section. Enums, ranges and per-key descriptions are derived from the actual parsers (`MSTestSettings.Configuration.cs`, `MSTestAdapterSettings.cs`, `RunConfigurationSettings.cs`, `PlatformConfigurationConstants.cs`).
- `docs/testconfig.schema.md` — how to reference the schema, the SchemaStore submission steps, how versioning works (rolling `main` URL; no in-file version), and a "keep in sync" pointer to the parser source-of-truth files.

### Validation
The schema was validated against the canonical `testconfig.json` example used in the repo's acceptance tests, and confirmed to reject bad enums, sub-minimum timeouts, unknown keys, and missing required `path` in `assemblyResolution`.

## SchemaStore registration

The companion SchemaStore catalog PR is open: **SchemaStore/schemastore#5842**. It is a self-hosted/external registration pointing at `https://raw.githubusercontent.com/microsoft/testfx/main/docs/testconfig.schema.json`, so it only goes green once this PR merges to `main` and the URL resolves.

## Follow-ups

Tasks 2–5 from #9381 are tracked separately: #9402, #9401, #9403, #9404.

### DIFF
--- a/docs/testconfig.schema.json
+++ b/docs/testconfig.schema.json
@@ -1,0 +1,264 @@
+{
+  "$schema": "http://json-schema.org/draft-07/schema#",
+  "$id": "https://raw.githubusercontent.com/microsoft/testfx/main/docs/testconfig.schema.json",
+  "title": "MSTest / Microsoft.Testing.Platform testconfig.json",
+  "description": "Configuration file consumed by Microsoft.Testing.Platform (MTP) and MSTest. The file is named 'testconfig.json' (or '<AssemblyName>.testconfig.json' once it is copied next to the test executable). See https://learn.microsoft.com/dotnet/core/testing/unit-testing-mstest-configure.",
+  "type": "object",
+  "properties": {
+    "$schema": {
+      "type": "string",
+      "description": "Reference to this JSON schema, enabling editor completion and validation."
+    },
+    "mstest": {
+      "$ref": "#/definitions/mstest"
+    },
+    "platformOptions": {
+      "$ref": "#/definitions/platformOptions"
+    },
+    "environmentVariables": {
+      "type": "object",
+      "description": "Environment variables to apply to the test host process. Mirrors the <EnvironmentVariables> element of legacy .runsettings. Declaring at least one entry opts the run into the test host controller process model. Do not store secrets here.",
+      "additionalProperties": {
+        "type": ["string", "number", "boolean", "null"]
+      }
+    }
+  },
+  "additionalProperties": true,
+  "definitions": {
+    "mstest": {
+      "type": "object",
+      "description": "MSTest-specific settings. This is the JSON equivalent of the <MSTest> (or <MSTestV2>) section of a .runsettings file.",
+      "additionalProperties": false,
+      "properties": {
+        "timeout": {
+          "type": "object",
+          "description": "Timeouts (in milliseconds) for tests and fixture methods.",
+          "additionalProperties": false,
+          "properties": {
+            "test": {
+              "type": "integer",
+              "minimum": 1,
+              "description": "Default timeout, in milliseconds, applied to every test method."
+            },
+            "assemblyInitialize": {
+              "type": "integer",
+              "minimum": 1,
+              "description": "Timeout, in milliseconds, for [AssemblyInitialize] methods."
+            },
+            "assemblyCleanup": {
+              "type": "integer",
+              "minimum": 1,
+              "description": "Timeout, in milliseconds, for [AssemblyCleanup] methods."
+            },
+            "classInitialize": {
+              "type": "integer",
+              "minimum": 1,
+              "description": "Timeout, in milliseconds, for [ClassInitialize] methods."
+            },
+            "classCleanup": {
+              "type": "integer",
+              "minimum": 1,
+              "description": "Timeout, in milliseconds, for [ClassCleanup] methods."
+            },
+            "testInitialize": {
+              "type": "integer",
+              "minimum": 1,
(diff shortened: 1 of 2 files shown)


0101 (approved):

docs/testconfig.schema.json
```diff
@@ -0,0 +1,264 @@
...
+              "enum": ["STA", "MTA"]
+            },
+            "disableAppDomain": {
+              "type": "boolean",
+              "description": "When true, disables AppDomain creation (.NET Framework only). Defaults to false."
+            },
+            "collectSourceInformation": {
+              "type": "boolean",
+              "description": "When true, source file and line information is collected during discovery. Defaults to true."
+            }
```
Evangelink, line 125: **MAJOR — `collectSourceInformation` is not implemented in the JSON config path**

`collectSourceInformation` under `mstest.execution` is documented here as a valid key, but none of the JSON configuration parsers actually read it:

- `MSTestSettings.SetSettingsFromConfig` (the 18-setting switch in `MSTestSettings.Configuration.cs`) does **not** call `ParseBooleanSetting` for `execution:collectSourceInformation`.
- `RunConfigurationSettings.SetRunConfigurationSettingsFromConfig` only reads `mstest:execution:executionApartmentState`.
- `MSTestAdapterSettings.ToSettings(IConfiguration)` only reads the `deployment:*` and `assemblyResolution` keys.

A grep of the full codebase confirms all occurrences of `CollectSourceInformation` are in XML runsettings paths or XML path comments (part of `<RunConfiguration>`, not `<MSTest>`), never in a `configuration["mstest:execution:collectSourceInformation"]` lookup.

A user who adds `"collectSourceInformation": false` to their `testconfig.json` will get no effect — the setting will be silently ignored.

**Recommendation**: Remove this property from the schema until it is wired up in the JSON config path, or add an `"x-not-yet-supported": true` / note in the `description` that this key is only honoured when using `.runsettings`.

docs/testconfig.schema.json
```diff
@@ -0,0 +1,264 @@
...
+            "additionalProperties": false,
+            "required": ["path"],
+            "properties": {
+              "path": {
+                "type": "string",
+                "description": "Directory to probe. Environment variables (for example %HOMEDRIVE%) and relative paths are supported."
+              },
+              "includeSubDirectories": {
+                "type": ["boolean", "string"],
+                "description": "When true, subdirectories of 'path' are also probed. Defaults to false."
```
Evangelink, line 193: **NIT — `["boolean", "string"]` union leaks an IConfiguration implementation detail**

`IConfiguration` always surfaces all values as strings internally, so from the configuration layer's perspective `true` and `"true"` are indistinguishable. However, from a JSON author's perspective the correct way to express this is a plain JSON boolean:

```json
{ "path": "C:\\foo", "includeSubDirectories": true }
```

Exposing `"type": ["boolean", "string"]` implicitly tells tooling that `"includeSubDirectories": "true"` is also a valid document, which makes schema-driven editors accept `"yes"` or `"1"` without complaint (those are strings that would silently fail `bool.TryParse`).

**Recommendation**: Use `"type": "boolean"` only.

Evangelink (commented):
> [!NOTE]
> 🤖 **Automated review by GitHub Copilot.** Posted via a maintainer's GitHub token, so it appears under their account — the account owner did **not** write or approve this content personally. Generated by the [Expert Code Review workflow](https://github.com/microsoft/testfx/actions/runs/28121753221). To request a follow-up action, reply by tagging `@copilot` directly.

## Review summary

All enum values cross-checked against the actual parser code — `parallelism.scope`, `launchDebuggerOnAssertionFailure`, and `executionApartmentState` are all correct. The `platformOptions` key paths all match `PlatformConfigurationConstants`. The overall schema structure, `$schema` dialect, `$id` URL, and the `.md` documentation are solid.

One factual correctness issue needs to be fixed before merge:

### 🔴 MAJOR — `collectSourceInformation` not wired up in JSON config

`mstest.execution.collectSourceInformation` is documented as a valid testconfig.json key, but no JSON configuration parser in the codebase reads `mstest:execution:collectSourceInformation`. A grep across the entire repo shows every occurrence of `CollectSourceInformation` lives in XML runsettings paths or XML-path code comments. The three JSON parsers invoked for `testconfig.json` — `MSTestSettings.SetSettingsFromConfig`, `RunConfigurationSettings.SetRunConfigurationSettingsFromConfig`, and `MSTestAdapterSettings.ToSettings(IConfiguration)` — do not call any `ParseBooleanSetting`/`TryParse` on that key. Users who set this in their `testconfig.json` will get a silent no-op.

Fix: remove the property from the schema (or clearly document it is `.runsettings`-only) until the JSON-path implementation is in place. Inline comment is left on line 125.

### 🟡 NIT — `includeSubDirectories` type union

`"type": ["boolean", "string"]` leaks how `IConfiguration` works internally. JSON authors should write `true`/`false`; `"type": "boolean"` is the correct schema type. Inline comment is left on line 193.

PR #9455, opened 2026-06-26
Exclude inner-build-only MTP logic from the cross-targeting outer build

### LINKED ISSUE #8043: Core MTP is incorrectly packaged
Lots of the stuff done in props/targets should **only** be relevant for inner builds and shouldn't be done for outer builds.

Currently, everything is done in outer builds as well as inner builds as we do the same thing for build, buildTransitive, and buildMultitargeting.

- build and buildTransitive are for inner builds
- buildMultitargeting is for outer builds.

Stuff that shouldn't happen in outer builds:

- `_MTPAddArguments` target
- Adding `Microsoft.Testing.Platform.Application` assembly metadata.
- Setting `IsTestingPlatformApplication`.

Things that I'm not sure about and will need validation by Test Explorer team:

- The `ProjectCapability` items.
    - TestingPlatformServer.ExitOnProcessExitCapability
    - TestingPlatformServer.UseListTestsOptionForDiscoveryCapability
    - TestingPlatformServer
    - TestContainer

They probably don't harm and we can keep them happening in outer builds if that's what Test Explorer would expect. This is also opening some questions, what if TestingPlatformServer or TestContainer capability is disabled for one TFM in a multi-targeted project. How does Test Explorer behave in Visual Studio today?

Evangelink:
## Summary

The core `Microsoft.Testing.Platform` package funnels `build/`, `buildTransitive/`, and `buildMultiTargeting/` into a single `buildMultiTargeting/Microsoft.Testing.Platform.targets` file, so inner-build-only logic also ran in the cross-targeting (outer) build of multi-targeted projects (#8043).

This PR guards the two operations that are **clearly** inner-build-only and safe to exclude from the outer build, using the standard outer-build discriminator `'$(TargetFramework)' != ''` (the same idiom already used in `Microsoft.Testing.Platform.MSBuild.targets`):

- **`AssemblyMetadata "Microsoft.Testing.Platform.Application"`** — the outer build produces no assembly, so this was inert there; inner builds still emit it unchanged.
- **`_MTPAddArguments` target** — it hooks `ComputeRunArguments`, which only ever runs in an inner build (`dotnet run` cannot drive a cross-targeting outer build).

## Intentionally left unchanged

The remaining items called out in #8043 are **not** touched here because removing them from the outer build is potentially breaking and needs Test Explorer team validation:

- The `ProjectCapability` items (`TestingPlatformServer`, `TestContainer`, `TestingPlatformServer.ExitOnProcessExitCapability`, `TestingPlatformServer.UseListTestsOptionForDiscoveryCapability`) — VS Test Explorer likely reads these at the project/outer level for multi-TFM projects, so keeping them is the non-breaking choice.
- `IsTestingPlatformApplication` — it is the gate the outer-build capability declarations depend on, so it must remain in the outer build while those capabilities do.

## Why a guard rather than a folder restructure

The package packs `build/` and `buildTransitive/` per-TFM (e.g. `build/net8.0/`) while `buildMultiTargeting/` has no TFM subfolder. That makes sibling relative-path imports between the per-TFM folders fragile (a consumer's `$(TargetFramework)` doesn't always match the resolved asset folder). The `'$(TargetFramework)' != ''` guard is functionally equivalent for these two items and robust.

## Notes

- No tests assert on these two items.
- The runtime reflection path (`IPlatformInformation`) reads `Microsoft.Testing.Platform.Application.BuildTimeUTC`, which is set in the `.csproj` and is untouched.

Partially addresses #8043.

Fixes: #8043

### DIFF
--- a/src/Platform/Microsoft.Testing.Platform/buildMultiTargeting/Microsoft.Testing.Platform.targets
+++ b/src/Platform/Microsoft.Testing.Platform/buildMultiTargeting/Microsoft.Testing.Platform.targets
@@ -17,8 +17,9 @@
   </ItemGroup>
 
   <!-- Add assembly attribute metadata for reflection usage. -->
-  <!-- This is an inner-build-only concern: the cross-targeting (outer) build of a multi-targeted -->
-  <!-- project produces no assembly, so '$(TargetFramework)' is empty there and we must not emit it. -->
+  <!-- This is an inner-build-only concern: in the cross-targeting (outer) build of a multi-targeted -->
+  <!-- project, '$(TargetFramework)' is empty and no assembly is produced, so AssemblyMetadata items -->
+  <!-- are inert there and must not be emitted. -->
   <ItemGroup>
     <AssemblyMetadata Include="Microsoft.Testing.Platform.Application" Value="true" Condition=" '$(TargetFramework)' != '' AND '$(IsTestingPlatformApplication)' == 'true' " />
   </ItemGroup>
--- a/src/Platform/Microsoft.Testing.Platform/buildMultiTargeting/Microsoft.Testing.Platform.targets
+++ b/src/Platform/Microsoft.Testing.Platform/buildMultiTargeting/Microsoft.Testing.Platform.targets
@@ -16,12 +16,16 @@
     <ProjectCapability Include="TestContainer" />
   </ItemGroup>
 
-  <!-- Add assembly attribute metadata for reflection usage -->
+  <!-- Add assembly attribute metadata for reflection usage. -->
+  <!-- This is an inner-build-only concern: the cross-targeting (outer) build of a multi-targeted -->
+  <!-- project produces no assembly, so '$(TargetFramework)' is empty there and we must not emit it. -->
   <ItemGroup>
-    <AssemblyMetadata Include="Microsoft.Testing.Platform.Application" Value="true" Condition=" '$(IsTestingPlatformApplication)' == 'true' " />
+    <AssemblyMetadata Include="Microsoft.Testing.Platform.Application" Value="true" Condition=" '$(TargetFramework)' != '' AND '$(IsTestingPlatformApplication)' == 'true' " />
   </ItemGroup>
 
-  <Target Name="_MTPAddArguments" BeforeTargets="ComputeRunArguments">
+  <!-- 'dotnet run' only ever drives an inner build, so this target is an inner-build-only concern. -->
+  <!-- Guarding on a non-empty '$(TargetFramework)' keeps it out of the cross-targeting (outer) build. -->
+  <Target Name="_MTPAddArguments" BeforeTargets="ComputeRunArguments" Condition=" '$(TargetFramework)' != '' ">
     <PropertyGroup>
       <RunArguments Condition=" '$(TestingPlatformCommandLineArguments)' != '' ">$(RunArguments) $(TestingPlatformCommandLineArguments)</RunArguments>
     </PropertyGroup>


src/Platform/Microsoft.Testing.Platform/buildMultiTargeting/Microsoft.Testing.Platform.targets
```diff
@@ -16,12 +16,16 @@
     <ProjectCapability Include="TestContainer" />
   </ItemGroup>
 
-  <!-- Add assembly attribute metadata for reflection usage -->
+  <!-- Add assembly attribute metadata for reflection usage. -->
+  <!-- This is an inner-build-only concern: the cross-targeting (outer) build of a multi-targeted -->
+  <!-- project produces no assembly, so '$(TargetFramework)' is empty there and we must not emit it. -->
```
Evangelink: **[NIT] Documentation Accuracy**

The causal chain in this comment is inverted. As written, "produces no assembly, **so** '$(TargetFramework)' is empty" implies that the absence of an assembly is what causes `$(TargetFramework)` to be empty — but it's the other way around: the outer build is *defined* by `$(TargetFramework)` being empty, and that is also what causes no assembly to be produced. Both are consequences of being the outer build, not one causing the other.

Suggested rewording:

```xml
<!-- This is an inner-build-only concern: in the cross-targeting (outer) build of a multi-targeted -->
<!-- project, '$(TargetFramework)' is empty and no assembly is produced, so AssemblyMetadata items  -->
<!-- are inert there and must not be emitted.                                                        -->
```

This phrasing correctly presents both facts as properties of the outer build rather than implying causality between them.

Evangelink (commented):
> [!NOTE]
> 🤖 **Automated review by GitHub Copilot.** Posted via a maintainer's GitHub token, so it appears under their account — the account owner did **not** write or approve this content personally. Generated by the [Expert Code Review workflow](https://github.com/microsoft/testfx/actions/runs/28235454300). To request a follow-up action, reply by tagging `@copilot` directly.

| # | Dimension | Verdict |
|---|-----------|---------|

✅ 21/22 dimensions clean.

- [ ] Documentation — comment's causal chain in the `AssemblyMetadata` block is inverted (see inline)

---

<details>
<summary>Full dimension verdicts</summary>

| # | Dimension | Verdict |
|---|-----------|---------|
| 1 | Algorithmic Correctness | ✅ LGTM |
| 2 | Threading & Concurrency | ➖ N/A |
| 3 | Security & IPC Contract Safety | ➖ N/A |
| 4 | Public API & Binary Compatibility | ➖ N/A |
| 5 | Performance & Allocations | ➖ N/A |
| 6 | Cross-TFM Compatibility | ➖ N/A |
| 7 | Resource & IDisposable Management | ➖ N/A |
| 8 | Defensive Coding at Boundaries | ➖ N/A |
| 9 | Localization & Resources | ➖ N/A |
| 10 | Test Isolation | ➖ N/A |
| 11 | Assertion Quality | ➖ N/A |
| 12 | Flakiness Patterns | ➖ N/A |
| 13 | Test Completeness & Coverage | ✅ LGTM (acknowledged by author; changes are inert in outer build) |
| 14 | Data-Driven Test Coverage | ➖ N/A |
| 15 | Code Structure & Simplification | ✅ LGTM |
| 16 | Naming & Conventions | ✅ LGTM |
| 17 | Documentation Accuracy | 🟢 1 NIT — see inline |
| 18 | Analyzer & Code Fix Quality | ➖ N/A |
| 19 | IPC Wire Compatibility | ➖ N/A |
| 20 | Build Infrastructure & Dependencies | ✅ LGTM |
| 21 | Scope & PR Discipline | ✅ LGTM |
| 22 | PowerShell Scripting Hygiene | ➖ N/A |
| — | MSBuild Authoring (supplemental) | ✅ LGTM |

</details>

---

**MSBuild correctness notes (supplemental):**

- `'$(TargetFramework)' != ''` is the correct and canonical inner-build discriminator. It matches the idiom already used in `Microsoft.Testing.Platform.MSBuild.targets` (confirmed in source). ✅
- The `build/` and `buildTransitive/` targets both redirect to `buildMultiTargeting/`, so the guard also applies when the file is imported for single-TFM projects — but in those projects `$(TargetFramework)` is always set (to the single TFM), so the guard evaluates to `true` and nothing changes there. ✅
- Putting the `Condition` on the `<Target>` element (rather than gating via `BeforeTargets`) is the correct approach for suppressing a target in a specific build phase; it avoids adding a dangling hook that would be silently skipped. ✅
- The `AssemblyMetadata` condition extends the pre-existing item-level condition (rather than moving it to the `<ItemGroup>`), which is consistent with how the item was already authored. ✅

src/Platform/Microsoft.Testing.Platform/buildMultiTargeting/Microsoft.Testing.Platform.targets
Evangelink: Good catch — applied your suggested wording in 1a3ea87de so the comment presents both facts as properties of the outer build rather than implying causality.

Youssef1313 (commented):
Have you considered dropping the whole buildMultiTargeting from MTP core package and only rely on build/buildTransitive for everything? Are there any problems you can observe doing it that way?

0101 (approved):

PR #9456, opened 2026-06-26
[DO NOT MERGE] Survey: gauge source-gen support across acceptance assets

Evangelink:
## ⚠️ Throwaway measurement branch — DO NOT MERGE

This is an **exploration** to answer one question before we roll out source generation across the acceptance suite: **how many existing acceptance assets cannot build under source generation today, and why?**

It builds on the foundation merged in #9370.

## How it works

- A global pipeline variable `MSTEST_ACCEPTANCE_SOURCEGEN_SURVEY=1` turns on **survey mode** in `TestAssetFixtureBase`.
- In survey mode, **every** fixture (opt-out, ignoring `SourceGenMetadataModes`) attempts to build its asset under **`MetadataMode.SourceGeneration`** and **records** the outcome via `SourceGenSurvey` instead of asserting — so a single CI run measures the whole suite without turning it red.
- Each outcome is one tagged console line (streamed to the CI test log because `TestingPlatformCaptureOutput=false`) plus a backup file under `artifacts/log/sourcegen-survey`:
  ```
  ##SGSURVEY## os=linux mode=SourceGeneration result=FAIL asset=FrameworkOnlyTests reason=error CS0234: ...ReflectionMetadataHook...
  ##SGSURVEY## os=linux mode=SourceGeneration result=PASS asset=TestInconclusive reason=
  ```
- To keep already-converted classes green, survey builds the **union** of `{SourceGeneration}` and each fixture's declared modes (non-fatally), so `InconclusiveTests` / `TestRunParametersTests` still find their `bin/<sub>` outputs.

## Scope / caveats

- **SourceGeneration mode only** for this first pass, to bound CI cost. AOT (`MSTest.AotReflection.SourceGeneration`) is a planned second pass (its gaps are a superset).
- Survey runs only in the **Debug** test legs (Win/Linux/Mac), which are the ones that run acceptance assets.
- This roughly **doubles asset-build work** in those legs, so they will be noticeably slower and **may approach job timeouts**. That's acceptable for a one-off measurement — console survey lines stream live, so even a leg that times out yields partial data. If a leg consistently times out, we'll narrow scope (e.g. one OS, or `MSTest.Acceptance` only).

## What to do with the results

Once CI runs, I'll grep `##SGSURVEY##` out of the three Debug test logs, dedupe by asset, and produce a per-OS PASS/FAIL summary bucketed by failure reason (no adapter ref, generics, inherited `[TestClass]`, cross-assembly reflection, VSTest-host, …). That gives us the concrete exclusion list and the size of the rollout for the real (opt-in) follow-up PRs.

## Local validation

39/39 pilot tests pass; survey records `FrameworkOnly=FAIL (CS0234 ReflectionMetadataHook)`, `TestInconclusive=PASS`, `TestRunParameters=PASS`.

<sub>Co-authored-by: Copilot</sub>

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -33,6 +33,13 @@ variables:
   # Cannot use key:value syntax in root defined variables
   - name: _TeamName
     value: MSTest
+  # THROWAWAY SURVEY BRANCH — do not merge.
+  # Turns on the source-gen survey in TestAssetFixtureBase: every acceptance fixture attempts to build
+  # its asset under MetadataMode.SourceGeneration (opt-out) and RECORDS the outcome via SourceGenSurvey
+  # (tagged "##SGSURVEY##" console lines) instead of asserting, so a single CI run gauges how many
+  # acceptance assets cannot support source generation today. Remove before merging.
+  - name: MSTEST_ACCEPTANCE_SOURCEGEN_SURVEY
+    value: 1
   - name: _RunAsInternal
     value: False
   - name: _RunAsPublic
--- a/test/Utilities/Microsoft.Testing.TestInfrastructure/SourceGenSurvey.cs
+++ b/test/Utilities/Microsoft.Testing.TestInfrastructure/SourceGenSurvey.cs
@@ -1,0 +1,95 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+namespace Microsoft.Testing.TestInfrastructure;
+
+/// <summary>
+/// THROWAWAY SURVEY HELPER — do not merge.
+///
+/// <para>
+/// Records, per acceptance asset, whether it can be built under a source-generation
+/// <see cref="MetadataMode"/>. <see cref="TestAssetFixtureBase"/> drives this in survey mode by
+/// attempting the source-gen build for every fixture (opt-out) and reporting the outcome here
+/// instead of failing the test. The goal is to gauge how many existing acceptance assets cannot
+/// support source generation today, and why.
+/// </para>
+///
+/// <para>
+/// Each outcome is emitted as a single tagged console line (so it surfaces in the CI test-step log,
+/// which streams test-host stdout when <c>TestingPlatformCaptureOutput=false</c>) and appended to a
+/// report file under <c>artifacts/log/sourcegen-survey</c> as a backup.
+/// </para>
+/// </summary>
+public static class SourceGenSurvey
+{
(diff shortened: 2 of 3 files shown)


Evangelink:
## Survey results (build 1482936, all Debug legs green)

Extracted `##SGSURVEY##` from the published `Logs_Build_*_Debug/sourcegen-survey/` backup files (the fixture `Console.WriteLine` is captured by MSTest during `ClassInitialize`, so it does not reach the test-step console log — the backup file is the source of truth).

| OS | distinct assets | PASS | FAIL |
|----|----|----|----|
| Linux | 92 | 91 | 1 |
| macOS | 92 | 91 | 1 |
| Windows | 108 | 107 | 1 |
| **Union** | **108** | **107** | **1** |

**The only build-gate blocker is `FrameworkOnlyTests`** on every OS:
> `CS0234: The type or namespace name 'ReflectionMetadataHook' does not exist ...` — the asset references only `MSTest.TestFramework`, not the adapter that carries the hook, so the generated registration cannot compile. This is expected and legitimately unsupportable.

### ⚠️ Important caveat: this measures the BUILD gate only
A successful build does **not** mean the asset runs correctly under source generation. Known runtime gaps all compile fine and therefore land in the PASS bucket here:
- VSTest-host assets (never use the MTP metadata hook),
- inherited `[TestClass]`, generic test methods, cross-assembly reflection,
- NativeAOT / trimming behavior.

So the real "cannot support source-gen" set is larger than 1 — but it is gated at **runtime**, which this build survey does not exercise.

### Takeaways for the rollout
1. **Build compatibility is essentially a non-issue** — we do not need a large "won't compile" exclusion list; only `FrameworkOnlyTests` must be excluded.
2. The meaningful rollout cost is **runtime** validation: each class must be parameterized by `MetadataMode` and threaded through `TestHost.LocateFrom` to actually run in source-gen mode, which is what surfaces the runtime gaps.
3. Suggested **survey v2** (cheap, high-signal): after building each source-gen variant, also *run it once* and assert tests were discovered / exit code is sane. That catches the dominant runtime gap ("0 tests discovered" for VSTest/inherited/generics) without a full per-class conversion.

Cost was manageable (Debug legs 14–34 min, no timeouts), so the same harness can drive the AOT pass and the runtime survey.

Evangelink:
Survey complete — purpose served. Build-gate result: 107/108 acceptance assets build cleanly under SourceGeneration on all OSes; the only blocker is FrameworkOnlyTests. Acted on in #9462 (opt-out default, FrameworkOnlyTests excluded). Closing this throwaway measurement branch; not merging.

PR #9502, opened 2026-06-29
Remove duplicated reflection helpers in ReflectHelper

Evangelink:
## Summary

Addresses part of #9499 (Code Duplication Analysis).

`ReflectHelper.GetTestCategories` and `ReflectHelper.GetTestPropertiesAsTraits` (static methods) duplicated the algorithm bodies of the `IReflectionOperations` extension methods of the same name in `ReflectionHelper`. The static `ReflectHelper.GetCustomAttributesCached` helper already delegates to `PlatformServiceProvider.Instance.ReflectionOperations`, so the two static methods are now thin delegates to the shared extension methods — removing ~70 lines of true duplication with zero behavior change.

The `ReflectionHelper` extension methods are the "live" implementations (used by `TypeEnumerator`); `ReflectHelper.GetTestCategories` (static) had no production callers and `ReflectHelper.GetTestPropertiesAsTraits` (static) is only referenced from a unit test, which keeps working via the delegate.

## Scope notes

The dedup-analysis issue lists 21 candidates. The remaining items were intentionally left as-is:
- **Jsonite folder** (4 high-priority items): vendored third-party BSD-licensed code (Alexandre Mutel, 2016) — must not be refactored.
- **`ReflectHelper` vs `ReflectionOperations`**: intentional architectural split (singleton facade vs service); merging would break the `ReflectHelper.Instance` surface.
- **`SynchronizationContextPreservingAsyncTaskMethodBuilder`**: generic/non-generic struct pair required by the compile-time async-builder pattern; cannot be unified.
- **Assert-condition analyzers / other analyzers**: every analyzer in the project repeats the `Initialize` boilerplate by convention; extracting a base for just two would be locally inconsistent.
- **Report engines**: the duplication is constructor forwarding across two assemblies, which can't be DRYed via inheritance.

## Validation

- `MSTestAdapter.PlatformServices` builds clean (0 warnings, 0 errors).
- `ReflectionOperationsTests` (47) and `TestPropertyAttributeTests` (3) pass.

Co-authored-by: Copilot <223556219+Copilot@users.noreply.github.com>

### DIFF
--- a/src/Adapter/MSTestAdapter.PlatformServices/Helpers/ReflectHelper.cs
+++ b/src/Adapter/MSTestAdapter.PlatformServices/Helpers/ReflectHelper.cs
@@ -6,6 +6,7 @@ using System.Security;
 #endif
 
 using Microsoft.VisualStudio.TestPlatform.MSTestAdapter.PlatformServices;
+using Microsoft.VisualStudio.TestPlatform.MSTestAdapter.PlatformServices.Helpers;
 using Microsoft.VisualStudio.TestPlatform.ObjectModel;
 using Microsoft.VisualStudio.TestTools.UnitTesting;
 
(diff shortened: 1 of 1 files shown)


Evangelink (commented):
> [!NOTE]
> 🤖 **Automated review by GitHub Copilot.** Posted via a maintainer's GitHub token, so it appears under their account — the account owner did **not** write or approve this content personally. Generated by the [Expert Code Review workflow](https://github.com/microsoft/testfx/actions/runs/28372765612). To request a follow-up action, reply by tagging `@copilot` directly.

---

### Behavioral equivalence — verified ✔

**`GetTestCategories`:** The old body called `ReflectHelper.GetCustomAttributesCached(method)` / `(owningType)` / `(owningType.Assembly)`, which is a static shim that already delegated to `PlatformServiceProvider.Instance.ReflectionOperations.GetCustomAttributesCached(...)`. The new one-liner delegates directly to `PlatformServiceProvider.Instance.ReflectionOperations.GetTestCategories(categoryAttributeProvider, owningType)`, which is the extension method in `ReflectionHelper` — that method calls `reflectionOperations.GetCustomAttributesCached(...)` on the same instance, hitting the same `ConcurrentDictionary` cache. Attribute order (method → type → assembly) and the lazy-`List<string>` / spread-to-array return pattern are identical.

**`GetTestPropertiesAsTraits`:** The old body used `GetCustomAttributesCached(testPropertyProvider)` for the method and `testPropertyProvider.ReflectedType is { } testClass ? GetCustomAttributesCached(testClass) : []` for the class. The new one-liner delegates to `ReflectionHelper.GetTestPropertiesAsTraits`, which uses the identical null-guarded pattern via `reflectionOperations.GetCustomAttributesCached`. The two-pass count-then-fill algorithm and the empty-array fast path are preserved verbatim.

The PR's zero-behavior-change claim is correct.

0101 (approved):

PR #9503, opened 2026-06-29
Fix [ClassCleanup] leak when ITestFilter drops the last test of an initialized class

Evangelink:
## Summary

Targets `dev/amauryleve/custom-test-filter` (#8896). Fixes a `[ClassCleanup]` leak in the new `ITestFilter` drop path, reported by an internal team using per-worker test sharding (each worker enumerates the full set per class, runs its subset, and `Drop`s the rest).

## The bug

`UnitTestRunner.FinishFilteredOutTestAsync` assumed a filtered-out test never loaded its type and therefore skipped class cleanup, only calling `MarkClassComplete`. That assumption is false when a **sibling** test of the same class already ran in this worker: `[ClassInitialize]` executed and `[ClassCleanup]` is still owed.

Because `ClassCleanupManager` counts down over the full (pre-filter) test set, `isLastTestInClass` is independent of run-vs-dropped. So whenever a class's **last-in-order** test is dropped — while earlier tests of that class ran — the worker leaks the class cleanup. `ForceCleanup` is not a safety net (it only fires on `IsGracefulStopRequested`).

The analogous assembly-cleanup asymmetry was already guarded by `_assemblyInitializeWasExecuted`, which is why only class cleanup leaked.

## The fix

When the dropped test is the last in its class, run `[ClassCleanup]` if the class was initialized in this worker. Detection reuses the existing `_lastRunnableTestByClass` map, which is populated only for classes that both have an executable cleanup method **and** ran a non-filtered test. The `TestClassInfo` is already cached in the `TypeCache`, so resolving it loads **no new type** — preserving the feature's "a `Drop` pays zero type-load / init cost" guarantee. A fully-dropped class (type never loaded) still correctly skips both init and cleanup.

## Validation

- Added acceptance test `TestFilterClassCleanupTests` reproducing the leak: a filter drops the last-in-order test of an initialized class and the only test of a fully-dropped class.
  - Against the **unfixed** adapter the test fails — output shows `ClassInitialize` + the run test but no `ClassCleanup`, exactly the reported symptom.
  - Against the **fixed** adapter it passes; the fully-dropped class runs neither init nor cleanup.
- `MSTestAdapter.PlatformServices.UnitTests` green across net462/net48/net8.0/net9.0/net8.0-windows.

Co-authored-by: Copilot <223556219+Copilot@users.noreply.github.com>

### DIFF
--- a/src/Adapter/MSTestAdapter.PlatformServices/Execution/UnitTestRunner.cs
+++ b/src/Adapter/MSTestAdapter.PlatformServices/Execution/UnitTestRunner.cs
@@ -584,9 +584,11 @@ internal sealed class UnitTestRunner
     }
 
     /// <summary>
-    /// Handles the bookkeeping (class-cleanup countdown, end-of-assembly cleanup) for a test that
-    /// was filtered out by a <see cref="ITestFilter"/>. Mirrors the tail of <see cref="RunSingleTestAsync"/>
-    /// without ever requiring <c>testMethodInfo</c>, since the filter ran before any type was loaded.
+    /// Handles the bookkeeping (class-cleanup countdown, class cleanup, end-of-assembly cleanup) for a
+    /// test that was filtered out by a <see cref="ITestFilter"/>. Mirrors the tail of
+    /// <see cref="RunSingleTestAsync"/>. The filtered-out test never loaded its own type, but if a
+    /// sibling test of the same class already ran in this worker the class was initialized and still
+    /// owes its <c>[ClassCleanup]</c>, so it is executed here when this is the last test of the class.
     /// </summary>
     private async Task<TestResult[]> FinishFilteredOutTestAsync(
         TestMethod testMethod,
@@ -598,8 +600,53 @@ internal sealed class UnitTestRunner
         _classCleanupManager.MarkTestComplete(testMethod, out bool isLastTestInClass);
         if (isLastTestInClass)
         {
-            // No class cleanup possible: we never loaded the test type, so there's nothing to
-            // execute. Still mark the class as complete so end-of-assembly cleanup is gated correctly.
+            // The class-cleanup countdown spans the full (pre-filter) set of tests, so the "last test
+            // in class" can land on a filtered-out test. The filtered-out test itself never loaded the
+            // type, but a SIBLING test of the same class may have run earlier in this worker — which
+            // means [ClassInitialize] already executed and [ClassCleanup] is still owed. We must run it
+            // here; otherwise the class leaks its cleanup whenever its last-in-order test is dropped.
+            //
+            // _lastRunnableTestByClass is populated only for classes that both have an executable
+            // cleanup method AND ran at least one non-filtered test in this worker, so its presence is
+            // exactly the signal that the type is already loaded and cleanup is pending. Resolving the
+            // test method info therefore hits the TypeCache and never loads a new type.
+            if (_lastRunnableTestByClass.TryGetValue(testMethod.FullClassName, out UnitTestElement? lastRunnableTest))
+            {
+                TestMethodInfo? testMethodInfo = _typeCache.GetTestMethodInfo(lastRunnableTest.TestMethod);
+                if (testMethodInfo is not null)
+                {
+                    ITestContext testContextForClassCleanup = PlatformServiceProvider.Instance.GetTestContext(testMethod: null, testMethod.FullClassName, testContextProperties, messageLogger, testContextForTestExecution.Context.CurrentTestOutcome);
+                    try
+                    {
+                        // Flow properties set during AssemblyInitialize and ClassInitialize so the
+                        // ClassCleanup method observes them, mirroring the run path in RunSingleTestAsync.
+                        var classCleanupImpl = (TestContextImplementation)testContextForClassCleanup.Context;
+                        classCleanupImpl.MergeProperties(testMethodInfo.Parent.Parent.PostAssemblyInitProperties);
+                        classCleanupImpl.MergeProperties(testMethodInfo.Parent.PostClassInitProperties);
+
+                        // Note: filterResult is empty for a dropped test, so any TestContext output
+                        // written by a *successful* [ClassCleanup] is not attached to a result here
+                        // (RunClassCleanupAsync only flushes TestContext output onto an existing result).
+                        // Console output is unaffected — it goes to process stdout, which the test host
+                        // still surfaces. A *failing* cleanup produces its own result, handled below.
+                        TestResult? cleanupResult = await testMethodInfo.Parent.RunClassCleanupAsync(testContextForClassCleanup, filterResult).ConfigureAwait(false);
+                        if (cleanupResult is not null)
+                        {
+                            // The current test was filtered out (no result of its own), so a class
+                            // cleanup failure must be attached to the last real test that ran in the class.
+                            cleanupResult.AssociatedUnitTestElement = lastRunnableTest;
+                            filterResult = [.. filterResult, cleanupResult];
+                        }
+                    }
+                    finally
+                    {
+                        (testContextForClassCleanup as IDisposable)?.Dispose();
+                    }
+                }
+            }
+
+            // Mark the class as complete so end-of-assembly cleanup is gated correctly. Done after the
+            // class cleanup above so assembly cleanup never runs before this class is fully torn down.
             _classCleanupManager.MarkClassComplete(testMethod.FullClassName);
         }
 
--- a/test/IntegrationTests/MSTest.Acceptance.IntegrationTests/TestFilterClassCleanupTests.cs
+++ b/test/IntegrationTests/MSTest.Acceptance.IntegrationTests/TestFilterClassCleanupTests.cs
@@ -1,0 +1,211 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using Microsoft.Testing.Platform.Acceptance.IntegrationTests;
+using Microsoft.Testing.Platform.Acceptance.IntegrationTests.Helpers;
+using Microsoft.Testing.Platform.Helpers;
+
+namespace MSTest.Acceptance.IntegrationTests;
+
+[TestClass]
+public sealed class TestFilterClassCleanupTests : AcceptanceTestBase<TestFilterClassCleanupTests.TestAssetFixture>
+{
+    public TestContext TestContext { get; set; } = null!;
+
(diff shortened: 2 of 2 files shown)


Evangelink:
### 🧪 Test quality grade — PR #9503



<table>
  <thead>
  </thead>
  <tbody>
    <tr>
      <td>new</td>
      <td><code>TestFilterClassCleanupTests.<br>ClassCleanup_<br>RunsWhenLastTestOfInitializedClassIsDropped</code></td>
      <td>A</td>
      <td>90–100</td>
      <td>Rich assertion set covers both the positive fix (ClassCleanup runs for the initialized class) and the guard against over-correction (ClassCleanup does not run for the fully-dropped class); clear AAA structure with informative comments.</td>
    </tr>
  </tbody>
</table>

`/grade-tests`.</sub>



> 🤖 **Automated content by GitHub Copilot.** Posted via a maintainer's GitHub token, so it appears under their account — the account owner did **not** write or approve this content personally. Generated by the [Grade Tests on PR (on open / sync)](https://github.com/microsoft/testfx/actions/runs/28373587839/agentic_workflow) workflow. · 179.3 AIC · ⌖ 12.7 AIC · ⊞ 43.7K · [◷]( · [◷](https://github.com/search?q=repo%3Amicrosoft%2Ftestfx+%22gh-aw-workflow-call-id%3A+microsoft%2Ftestfx%2Fgrade-tests-on-pr.agent%22&type=issues))

0101 (approved):

test/IntegrationTests/MSTest.Acceptance.IntegrationTests/TestFilterClassCleanupTests.cs
```diff
@@ -0,0 +1,164 @@
...
+    public TestContext TestContext { get; set; } = null!;
+
+    // Regression test for the [ClassCleanup] leak when an ITestFilter drops the *last-in-order* test of
+    // a class that was already initialized by an earlier (non-dropped) test in the same worker. This is
+    // the exact scenario reported for the per-worker sharding pattern (each worker enumerates the full
+    // set per class, runs its subset, and Drops the rest). Before the fix, ClassCleanupManager counted
+    // the dropped test as the last test of the class, so the drop path skipped class cleanup even though
+    // the type had been loaded and [ClassInitialize] had run.
+    [TestMethod]
+    public async Task ClassCleanup_RunsWhenLastTestOfInitializedClassIsDropped()
```
Evangelink, line 22: **[MODERATE] Test Completeness & Coverage**

The single test only covers the happy path (cleanup runs and succeeds). The newly-introduced branch at `UnitTestRunner.cs:580` — where `cleanupResult is not null` because `[ClassCleanup]` itself throws — is entirely untested. The result-attachment logic (`cleanupResult.AssociatedUnitTestElement = lastRunnableTest`) and the array-spread (`filterResult = [.. filterResult, cleanupResult]`) on the following lines are therefore unvalidated in the drop path.

Consider adding a second scenario:
```csharp
[TestMethod]
public async Task ClassCleanup_FailureIsAttributedToLastRealTest_WhenLastTestIsDropped()
{
    // Filter drops the last-in-order test; [ClassCleanup] throws.
    // Expected: ExitCode.Failure, and the failure result is attributed to Test_A_Run,
    // not to the non-existent dropped test.
}
```
This would catch a regression if `AssociatedUnitTestElement` is ever misrouted or if the `filterResult` spread is accidentally skipped.

src/Adapter/MSTestAdapter.PlatformServices/Execution/UnitTestRunner.cs
```diff
@@ -550,8 +552,48 @@ private async Task<TestResult[]> FinishFilteredOutTestAsync(
...
+                    ITestContext testContextForClassCleanup = PlatformServiceProvider.Instance.GetTestContext(testMethod: null, testMethod.FullClassName, testContextProperties, messageLogger, testContextForTestExecution.Context.CurrentTestOutcome);
+                    try
+                    {
+                        // Flow properties set during AssemblyInitialize and ClassInitialize so the
+                        // ClassCleanup method observes them, mirroring the run path in RunSingleTestAsync.
+                        var classCleanupImpl = (TestContextImplementation)testContextForClassCleanup.Context;
+                        classCleanupImpl.MergeProperties(testMethodInfo.Parent.Parent.PostAssemblyInitProperties);
+                        classCleanupImpl.MergeProperties(testMethodInfo.Parent.PostClassInitProperties);
+
+                        TestResult? cleanupResult = await testMethodInfo.Parent.RunClassCleanupAsync(testContextForClassCleanup, filterResult).ConfigureAwait(false);
```
Evangelink, line 632: **[NIT] Documentation Accuracy**

`filterResult` is `[]` here — the dropped test produced no result of its own. `RunClassCleanupAsync`'s inner `DoRunAsync` flushes `TestContext.WriteLine` output by appending to `results[results.Length - 1]`, so when `results` is empty, any text written to `TestContext` by a *successful* `[ClassCleanup]` is silently discarded.

`Console.WriteLine` output is unaffected (goes to process stdout, which the test host surfaces through `testHostResult`). This is an acceptable limitation in the drop path, but a short comment would save future readers from searching for where the output went:

```csharp
// Note: filterResult is empty for a dropped test, so TestContext output written by a
// successful [ClassCleanup] is not attached to any result here. Console output is still
// captured via process stdout.
TestResult? cleanupResult = await testMethodInfo.Parent.RunClassCleanupAsync(
    testContextForClassCleanup, filterResult).ConfigureAwait(false);
```

test/IntegrationTests/MSTest.Acceptance.IntegrationTests/TestFilterClassCleanupTests.cs
```diff
@@ -0,0 +1,164 @@
...
+    [ClassCleanup]
+    public static void ClassCleanup()
+        => Console.WriteLine("PartiallyDroppedClass.ClassCleanup");
+
+    [TestMethod]
+    public void Test_A_Run()
+        => Console.WriteLine("PartiallyDroppedClass.Test_A_Run");
+
+    [TestMethod]
+    public void Test_Z_Dropped()
```
Evangelink, line 184: **[NIT] Naming & Conventions**

The names `Test_A_Run` and `Test_Z_Dropped` carry an implicit ordering assumption: MSTest's default alphabetical ordering guarantees that `A` executes before `Z` is attempted, ensuring `[ClassInitialize]` fires (driven by `Test_A_Run`) before `Test_Z_Dropped` is evaluated by the filter. If a future editor renames or reorders these methods, the scenario silently breaks (both tests run, neither is dropped last, `[ClassCleanup]` fires through the regular path, and the bug is no longer exercised).

A brief comment makes the dependency visible:

```csharp
// Method names are intentional: alphabetical ordering ensures Test_A_Run executes first
// (triggering [ClassInitialize] for the class), then Test_Z_Dropped is the last-in-order
// test and is dropped by the filter — exactly the scenario the fix addresses.
[TestMethod]
public void Test_Z_Dropped()
    => Console.WriteLine("PartiallyDroppedClass.Test_Z_Dropped");
```

Evangelink (commented):
> [!NOTE]
> 🤖 **Automated review by GitHub Copilot.** Posted via a maintainer's GitHub token, so it appears under their account — the account owner did **not** write or approve this content personally. Generated by the [Expert Code Review workflow](https://github.com/microsoft/testfx/actions/runs/28373587813). To request a follow-up action, reply by tagging `@copilot` directly.

| # | Dimension | Verdict |
|---|-----------|---------|
| 13 | Test Completeness & Coverage | 🟡 1 MODERATE |
| 16 | Naming & Conventions | 💬 1 NIT |
| 17 | Documentation Accuracy | 💬 1 NIT |

✅ 19/22 dimensions clean.

- [ ] Test Completeness — the `cleanupResult is not null` branch (cleanup fails in the drop path) has no acceptance-test coverage; `AssociatedUnitTestElement` attribution is unvalidated
- [ ] Naming — the `Test_A_Run` / `Test_Z_Dropped` alphabetical-ordering assumption is undocumented
- [ ] Documentation — successful `[ClassCleanup]` output is silently discarded when `filterResult` is empty; no comment explains this

---

**Overall assessment:** The fix is algorithmically correct and well-reasoned. The `_lastRunnableTestByClass` sentinel is the right signal: it's populated only when both conditions are true — the class has an executable cleanup method **and** at least one non-filtered test ran in this worker — so a fully-dropped class is still correctly skipped. The `ExecuteClassCleanupAsync` semaphore and `IsClassCleanupExecuted` flag in `TestClassInfo` prevent any risk of double execution even under concurrent test scheduling. Thread safety is maintained via `ConcurrentDictionary.TryGetValue`. The `ConfigureAwait(false)` and try/finally disposal patterns mirror the surrounding code correctly. The three items above are polish suggestions; none blocks the merge.

test/IntegrationTests/MSTest.Acceptance.IntegrationTests/TestFilterClassCleanupTests.cs
Evangelink: Done. Added a second acceptance test `ClassCleanup_FailureIsAttributedToLastRealTest_WhenLastTestIsDropped` that drops the last-in-order test and makes `[ClassCleanup]` throw (gated by a `THROW_CLEANUP` env var). It asserts `ExitCode.AtLeastOneTestFailed`, that the cleanup exception surfaces (proving the `cleanupResult is not null` branch + the `filterResult` spread run in the drop path), and that the failure is not attributed to the dropped test. Verified it fails on the unfixed adapter.

src/Adapter/MSTestAdapter.PlatformServices/Execution/UnitTestRunner.cs
Evangelink: Done. Added a comment at the `RunClassCleanupAsync` call explaining that `filterResult` is empty for a dropped test, so a successful `[ClassCleanup]`'s `TestContext` output is not attached to any result here, while Console output still surfaces via process stdout.

test/IntegrationTests/MSTest.Acceptance.IntegrationTests/TestFilterClassCleanupTests.cs
Evangelink: Done, with a correction worth flagging: the adapter's `OrderTestsByNameInClass` actually defaults to `false`, so alphabetical ordering was not really guaranteed — the test was relying on undefined discovery order. I enabled `<OrderTestsByNameInClass>true</OrderTestsByNameInClass>` in the asset runsettings to make the A-before-Z ordering a real guarantee, and added a comment on the methods documenting that the names + ordering setting are what exercise the bug.

Issue #9139, opened 2026-06-15
Silence-driven progress heartbeat renderer for SimpleAnsi/NoAnsi terminal modes


Evangelink:
## Summary

Implement the silence-driven progress heartbeat renderer for the `SimpleAnsi` and `NoAnsi` terminal modes. This is the core implementation of #9125 and the natural successor to #9138.

## Background

Today, `TerminalOutputDevice.cs:212-216` force-disables progress for `SimpleAnsi`/`NoAnsi`, leaving CI / piped / file-redirected runs with zero progress signal between the banner and the summary. Previously the platform emitted a per-DLL summary every 3 seconds, which was rejected as spam (#6753). The fix is event-driven, not time-driven.

## Design

The terminal reporter picks a renderer based on capability:

| AnsiMode | Renderer |
|---|---|
| `AnsiIfPossible` / `ForceAnsi` | existing cursor-redraw renderer (unchanged) |
| `SimpleAnsi` / `NoAnsi` | **NEW** silence-driven heartbeat renderer |

Three independent emission rules. Each produces single lines, no cursor tricks, no duplicate timestamps (CIs already prepend their own):

### E1 — silence heartbeat

Emit one summary line **only** when no test completion event has occurred for N seconds (default 30s). A healthy fast suite emits zero heartbeat lines.

`running … 312/3200 completed, 2 failed | active: SlowTest.Foo`

### E2 — slow-test surfacing

When a single test exceeds a per-test threshold (default 60s), emit one line with exponential backoff (60s → 2m → 4m → 8m …):

`[slow] still running after 60s: SlowTest.IntegrationFoo (MyAcceptance, net9.0)`

E2 fires in **all** modes (including cursor) because it leaves a durable scrollback line; cursor frames are transient.

### E3 — failures inline

Verify failures continue to be printed at the moment of failure for `SimpleAnsi`/`NoAnsi`. Already true in cursor mode; mostly a verification item.

## Gating

- Renderer is selected when `--progress` (#9138) resolves to `on` AND `AnsiMode` is `SimpleAnsi`/`NoAnsi`.
- `--progress off` (or legacy `--no-progress`) suppresses everything.
- Test-host controller / `--list-tests` / server mode continue to suppress.

## Knobs (env vars only, no new CLI flags)

- `MTP_PROGRESS_SILENCE_SECONDS` (default `30`; `0` disables E1)
[...]

Issue #9125, opened 2026-06-14
Add CI-friendly textual progress heartbeat to MTP terminal reporter (raw log progress signal)


Evangelink:
## Summary

When running `dotnet test` (MTP) in AzDO (and any non-ANSI CI), users currently have a binary choice in the **raw console log**:

| Mode | Per-test "passed" lines | Progress in raw log |
|---|---|---|
| `--output normal` (default) | ❌ none (good) | ❌ silent until summary (bad) |
| `--output detailed` | ✅ one per test (noisy) | ❌ still no time-based progress |

`--report-azdo-progress` only updates the AzDO **timeline** view via `##vso[task.logdetail …]` commands; it does **not** print anything human-readable to the raw build log, which is what you stare at when investigating a CI run.

We currently work around this in our own pipelines with `--output detailed`, which produces hundreds of lines like:

```
passed UnsupportedRunSettingsEntriesAreFlagged ("net10.0") (3s 768ms)
  from D:\a\_work\1\s\artifacts\bin\MSTest.Acceptance.IntegrationTests\Debug\net11.0\MSTest.Acceptance.IntegrationTests.dll (net11.0|x64)
passed UnsupportedRunSettingsEntriesAreFlagged_Localization ("fr-FR","it-IT",null,"fr-FR") (3s 756ms)
…
```

That's a lot of noise (failures get buried) and it still does not convey **time-based** progress — only completion order.

## Motivation

The maintainer experience in CI is:

- "Did the test run hang?" requires scrolling 10k lines of passed-test names, or staring at the timeline view in a separate panel.
- Dropping `--output detailed` gives a clean log but no signal at all between the banner and the summary — for a 15‑minute test run that feels broken.
- The cursor-redraw progress (`SimpleAnsi`/`AnsiIfPossible`) is intentionally disabled in CI because raw log files can't display cursor movement, so today there is no progress at all in raw logs.

## Proposed feature: CI-friendly textual progress heartbeat

Emit a **single-line plain-text status line periodically** (default ~30s) in `SimpleAnsi` and `NoAnsi` modes. Each line is appended to the log (no cursor magic), so it's readable in AzDO/GitHub Actions/Jenkins raw logs.

Example shape (subject to bikeshed):

```
[00:30] running … 124/? completed, 2 failed (active: MyClass.MyTest, MyOther.OtherTest)
[01:00] running … 312/? completed, 2 failed (active: SlowTest.Foo)
[...]

PR #9541, opened 2026-07-01
Add Microsoft.Testing.Extensions.GitHubActionsReport (groups, annotations, step summary, slow-test notices)

### LINKED ISSUE #9142: New extension: Microsoft.Testing.Extensions.GitHubActionsReport (groups, annotations, step summary)
## Summary

Add a new `Microsoft.Testing.Extensions.GitHubActionsReport` extension package (sibling of the existing `Microsoft.Testing.Extensions.AzureDevOpsReport`) that emits GitHub Actions-native workflow commands so test runs on GitHub Actions produce a first-class experience:

1. **Per-assembly log groups** — `::group::` / `::endgroup::` so the runner UI collapses them by default.
2. **Failure annotations** — `::error file=…,line=…,title=Test failed::FQN` so failures appear in the PR **Files changed** gutter and the workflow Annotations tab.
3. **Job summary** — write a markdown summary to `\` rolling up totals, failures, slowest tests.
4. **Slow-test notices** — `::notice::` for tests crossing the slow-test threshold from #9139.

Hooks into the `IProgressEnricher` surface introduced by #9139.

## Motivation

- GitHub Actions is now one of the most common CI hosts for .NET projects. There is currently no first-party MTP integration.
- `\` is a uniquely powerful GH surface — a markdown blob that appears on the workflow run summary page. Test runners that use it provide a massively better PR review experience than those that don't.
- Failure annotations on the diff gutter let reviewers jump straight to the failing assert in code review without scrolling logs.

### Log groups
On `OnAssemblyStart(asm)` → `::group::Tests: {asm.Name} ({asm.TargetFramework})`.
On `OnAssemblyEnd(asm)` → `::endgroup::`.

### Failure annotations
On `OnFailure(testResult)`:

`::error file={file},line={line},col={col},title=Test failed: {fqn}::{firstLineOfMessage}`

Multiline error messages: prefer `::error` followed by the message body via `%0A` URL-escaped newlines, or fall back to a single-line snippet.

### Slow-test notices
On `OnSlowTestEmit(test, currentDuration)`:

`::notice file={file},line={line}::{fqn} still running after {currentDuration}s`

### Job summary
Append a markdown block to the file pointed to by `\` (per [Microsoft docs](https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#adding-a-job-summary)):

```markdown
## ✅ Test Run Summary

| Assembly | Passed | Failed | Skipped | Duration |
|---|---:|---:|---:|---:|
| MSTest.UnitTests (net9.0) | 845 | 0 | 2 | 1m 12s |
| MSTest.Acceptance.IntegrationTests (net9.0) | 312 | 1 | 0 | 4m 37s |

### ❌ Failures (1)
- `MyOther.OtherTest` — Expected: 42 but was: 41 ([MyOther.cs:88](…))

### ⏱ Slowest tests
- `SlowTest.IntegrationFoo` — 4m 12s
- `SlowTest.Foo` — 2m 03s
```

Concurrency-safe append (per [GH docs](https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions)) since multiple matrix legs may write concurrently — though each leg gets its own file, so this is mostly a non-issue.

## Knobs

- `--report-gh-groups` — on/off, default `on`.
- `--report-gh-annotations` — on/off, default `on`.
- `--report-gh-step-summary` — on/off, default `on`.
- `--report-gh-slow-test-notices` — on/off, default `on`.

Auto-detection: extension activates iff `GITHUB_ACTIONS=true`; otherwise no-ops silently. Discoverable via a single master switch — TBD whether `--report-gh` (similar to `--report-azdo-progress`) or auto-on when the env is detected. Recommendation: auto-on when env detected, with each individual knob still respected.

## Touchpoints

- **New package**: `src/Platform/Microsoft.Testing.Extensions.GitHubActionsReport/` (mirror the `AzureDevOpsReport` folder layout).
- `GitHubActionsReporter.cs` — implements `IProgressEnricher`.
- `GitHubActionsReportCommandLineProvider.cs` — the four options above.
- `GitHubActionsEscaper.cs` — escape special chars in titles/messages per [GH workflow commands docs](https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#example-of-a-debug-message).
- Resource strings + `dotnet msbuild .../GitHubActionsReport.csproj /t:UpdateXlf`.
- Help/info acceptance test expectations (alphabetically sorted):
  - `test/IntegrationTests/Microsoft.Testing.Platform.Acceptance.IntegrationTests/HelpInfoAllExtensionsTests.cs`
  - `test/IntegrationTests/Microsoft.Testing.Platform.Acceptance.IntegrationTests/MSBuild.KnownExtensionRegistration.cs`
- Unit tests in `test/UnitTests/Microsoft.Testing.Extensions.UnitTests/` — verify escape behaviour, group emission, step-summary markdown shape.

## Open questions

1. Auto-on (when `GITHUB_ACTIONS=true`) vs explicit `--report-gh-*` opt-in master switch. AzDO precedent is explicit opt-in via `--report-azdo-progress`; symmetric choice would be `--report-gh` master switch.
2. Should `\` be byte-capped to GitHub's 1 MiB limit? (Yes, with truncation note.)
3. Should we emit `::warning::` for skipped tests or only for failures? (Probably only on opt-in.)
4. How to surface the file/line for failure annotations when running on Linux runners but the test assembly was built on Windows (path-separator translation).

## Out of scope

- Other CI hosts (GitLab, Jenkins, Buildkite, TeamCity) — each is its own package.
- Uploading test result artifacts (use the existing `actions/upload-artifact` action + the platform's existing TRX/JUnit/HTML reports).
- PR review comments (would need `pull-requests: write` token, separate concern).

## Related

- #9125 — parent design discussion (see "Building blocks" S1 grouping + B7 annotations).
- #9139 — silence-driven heartbeat + `IProgressEnricher` hook (prerequisite).
- `Microsoft.Testing.Extensions.AzureDevOpsReport` — sibling extension; aim for symmetric command-line shape where it makes sense.

azat-msft:
## Summary

Fixes #9142.

Adds a new `Microsoft.Testing.Extensions.GitHubActionsReport` extension package (sibling of the existing `Microsoft.Testing.Extensions.AzureDevOpsReport`) that emits GitHub Actions-native workflow commands so test runs on GitHub Actions produce a first-class experience. This PR implements **all four** requirements from the issue.

## What's included

All four building blocks from #9142:

1. **Per-assembly log groups** — emits `::group::` / `::endgroup::` so each test assembly's output is collapsed by default in the runner UI (`GitHubActionsReporter.cs`).
2. **Failure annotations** — emits an `::error file=…,line=…,title=…::` command for each failing test, so failures appear in the workflow Annotations tab and, when the source location can be resolved, on the pull request's **Files changed** diff gutter (`GitHubActionsAnnotationReporter.cs`).
3. **Job summary** — appends a markdown roll-up (per-assembly totals, failures, slowest tests) to the file pointed to by `GITHUB_STEP_SUMMARY`, which GitHub renders on the workflow run summary page (`GitHubActionsSummaryReporter.cs`).
4. **Slow-test notices** — emits a `::notice` command for any test still running past a threshold, default 60s (`GitHubActionsSlowTestReporter.cs`).

Supporting pieces:

- `GitHubActionsEscaper.cs` — escapes special characters per GitHub workflow-command rules.
- `GitHubActionsCommandLineProvider.cs` / `GitHubActionsCommandLineOptions.cs` — the master switch plus per-feature toggles.
- `GitHubActionsRepositoryRoot.cs` / `TestNodeIdentity.cs` / `TargetFrameworkMonikerHelper.cs` — source-location and identity helpers used for annotations.
- Extensions entry point + MSBuild builder hook, resources (`.resx` + regenerated `.xlf` for all locales), `PublicAPI`, `BannedSymbols`, `PACKAGE.md`, build props.
- Wired into `TestFx.slnx`, Platform `InternalsVisibleTo`, and the unit-test project.

## Behaviour

- Activates when `GITHUB_ACTIONS=true` and `report-gh` is on.
- Each feature is individually toggleable:

| Option | Description | Default |
|---|---|---|
| `--report-gh` | Master switch to enable the extension outside GitHub Actions | off |
| `--report-gh-groups on\|off` | Per-assembly log groups | on |
| `--report-gh-annotations on\|off` | Failure annotations | on |
| `--report-gh-step-summary on\|off` | Markdown job summary | on |
| `--report-gh-slow-test-notices on\|off` | Slow-test notices | on |
| `--report-gh-slow-test-threshold <seconds>` | Seconds before a slow-test notice is emitted | 60 |

## Tests

New unit tests in `test/UnitTests/Microsoft.Testing.Extensions.UnitTests/`:

- `GitHubActionsEscaperTests.cs`
- `GitHubActionsCommandLineProviderTests.cs`
- `GitHubActionsReporterTests.cs` (group emission)
- `GitHubActionsAnnotationReporterTests.cs`
- `GitHubActionsSummaryReporterTests.cs`
- `GitHubActionsSlowTestReporterTests.cs`

## Related

- #9142 — feature request (this PR).
- #9125 — parent design discussion.
- #9139 — progress-enricher hook.
- `Microsoft.Testing.Extensions.AzureDevOpsReport` — sibling extension.


src/Platform/Microsoft.Testing.Extensions.GitHubActionsReport/GitHubActionsAnnotationReporter.cs
```diff
@@ -0,0 +1,252 @@
...
+    // Fully-qualified type prefixes for MSTest assertion implementations. A stack frame whose
+    // 'code' starts with any of these is treated as framework internals and skipped when looking
+    // for the user's call site to annotate.
+    private static readonly string[] AssertionImplementationCodePrefixes =
+    [
+        "Microsoft.VisualStudio.TestTools.UnitTesting.Assert.",
+        "Microsoft.VisualStudio.TestTools.UnitTesting.AssertExtensions.",
+        "Microsoft.VisualStudio.TestTools.UnitTesting.CollectionAssert.",
+        "Microsoft.VisualStudio.TestTools.UnitTesting.StringAssert.",
+    ];
```
azat-msft: This does not cover framework-specific prefixes. Adding those would introduce a never-ending maintainability burden. Copilot suggests adding the resolution based on `TestNode`'s own `TestFileLocationProperty`  ( `FilePath`  +  `LineSpan`).
However, the azure reporter extension has similar logic. If this is re-worked here, also be better re-worked in the azure extension

test/UnitTests/Microsoft.Testing.Extensions.UnitTests/GitHubActionsSummaryReporterTests.cs
```diff
@@ -0,0 +1,70 @@
...
+            new("Slow", "T.Slow", GitHubActionsSummaryReporter.TerminalKind.Passed, TimeSpan.FromSeconds(65)),
+        ];
+
+        string markdown = GitHubActionsSummaryReporter.BuildMarkdown(records, "T", "net9.0");
+
+        Assert.Contains("### ⏱ Slowest tests", markdown);
+        int slowIndex = markdown.IndexOf("- `T.Slow` — 1m 05s", StringComparison.Ordinal);
+        int fastIndex = markdown.IndexOf("- `T.Fast` — 10ms", StringComparison.Ordinal);
+        Assert.IsGreaterThanOrEqualTo(0, slowIndex, markdown);
+        Assert.IsGreaterThan(slowIndex, fastIndex, markdown);
```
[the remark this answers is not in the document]
azat-msft: The definition of `IsGreaterThan` is 
```
public static void IsGreaterThan<T>(T lowerBound, T value   ....
```
and the description states 
```
Tests whether the value is greater than the lower bound and throws an exception if it is not.
```
So the current implementation (``fastIndex` is greater than `slowIndex`) is correct`

Evangelink:
### Review remarks

Nice work — escaping, slow-test threading, and localization all look solid. Collecting the remaining points from my pass below (the multi-assembly `dotnet test` output-forwarding topic is intentionally left out — we'll handle that separately). Grouped by severity.

---

#### Major

**1. No help/info acceptance coverage for the new options**
`HelpInfoAllExtensionsTests` generates the `AllExtensionsInfoTest.csproj` asset with a fixed `<ItemGroup>` of extension `PackageReference`s, and `Microsoft.Testing.Extensions.GitHubActionsReport` isn't in that list. So the provider never loads in that host, `--report-gh*` never renders, and the golden `--help`/`--info` expectations keep matching — the tests pass only because the options are absent, i.e. there's zero coverage for them. Fix is two-part: add the package to the asset **and** add the six options to both the `--help` and `--info` expectation blocks (kept alphabetical). (Side note: `VideoRecorder` is in the same situation — has a CLI provider but isn't in the asset — whereas `AzureFoundry`/`PackagedApp`/`Telemetry` are absent but add no options, so those are harmless.)

**2. `GitHubActionsAnnotationReporter.ConsumeAsync` lacks the defensive try/catch its siblings have**
`GitHubActionsSlowTestReporter` and `GitHubActionsSummaryReporter` wrap their per-node work in try/catch (log a warning, keep going). `ConsumeAsync` in the annotation reporter doesn't, and it can throw on user-influenced data:
- `TestNode.Properties.SingleOrDefault<TestNodeStateProperty>()` throws `InvalidOperationException` if a node ever carries more than one state property — prefer `FirstOrDefault`.
- source-location resolution / `IFileSystem.ExistFile(...)` can throw on a malformed path.

An exception here propagates into the platform's data-consumer dispatch instead of degrading to "no annotation for this test". Please mirror the sibling consumers' try/catch + `LogWarning`.

---

#### Moderate

**3. `GITHUB_STEP_SUMMARY` concurrent-writer hazard**
`GitHubActionsSummaryReporter` opens the summary file with `FileMode.Append, FileAccess.Write, FileShare.Read`. When several assemblies run in the same job/step (a multi-project `dotnet test` over one solution), they append to the same file concurrently; `FileShare.Read` denies other writers, so it throws `IOException` → caught → the summary is silently dropped with only a warning. The AzDo sibling avoids this by writing each run to a uniquely-named file. Suggest `FileShare.ReadWrite` + a short bounded retry on `IOException` (or a cross-process mutex), or at minimum document the single-assembly-per-step assumption.

**4. `TargetFrameworkMonikerHelper.cs` duplicates a shared helper instead of linking it**
`src/Platform/SharedExtensionHelpers/TargetFrameworkMonikerHelper.cs` already exists and the AzDo project links it (`<Compile Include="$(RepoRoot)src\Platform\SharedExtensionHelpers\TargetFrameworkMonikerHelper.cs" .../>`). This project copies it instead. Please delete the copy and link the shared file. (`TestNodeIdentity.cs` being copied is consistent with AzDo keeping its own copy — acceptable, though a future shared helper would be nicer.)

**5. Brittle, checkout-coupled annotation test assertions**
`GitHubActionsAnnotationReporterTests` hard-codes `line=19` / `line=40` — the positions of the `throw` statements in the test file itself — and resolves the source path via `GitHubActionsRepositoryRoot.FindGitRoot()` against the real checkout. Any edit to the top of the file shifts those numbers, and a relocated/packaged layout breaks source resolution. Suggest computing the expected line dynamically (`[CallerLineNumber]` or search-for-throw) and injecting a fake `IFileSystem`/git-root so the test doesn't depend on physical repo layout.

**6. Not wired into the MSTest.Sdk `AllMicrosoft` profile**
`AzureDevOpsReport` is auto-referenced under `TestingExtensionsProfile == 'AllMicrosoft'` via `EnableMicrosoftTestingExtensionsAzureDevOpsReport` (`ClassicEngine.targets`, with the matching `VSTest.targets` guard and `NativeAOT.targets` warning). There's no equivalent `EnableMicrosoftTestingExtensionsGitHubActionsReport` wiring, so MSTest.Sdk users on the "all Microsoft extensions" profile get AzDo reporting but not GitHub Actions reporting. If deferring is intended, could you file a follow-up issue; otherwise add the parallel `Enable…`/version/`PackageReference` block in the three targets files.

**7. Group `::endgroup::` ordering guard dropped vs. the AzDo sibling**
`AzureDevOpsLogGroupReporter` deliberately also implements `IDataConsumer` (no-op `ConsumeAsync`) and guards with a `_groupOpened` flag so `::endgroup::` is ordered into the consumer phase and provably wraps all output. `GitHubActionsReporter` implements only `ITestSessionLifetimeHandler`, emits `::endgroup::` unconditionally (no guard, no try/catch in `OnTestSessionStartingAsync`), and relies on registration order. In practice it's probably fine here (annotations emit during `ConsumeAsync` before session-finishing; the summary writes to a file; slow notices stop mid-session), but the sibling's intentional engineering was silently dropped — either mirror the pattern or add a comment explaining why the simpler ordering is sufficient.

---

#### Low / nit

**8.** `GitHubActionsRepositoryRoot.s_cachedGitRoot` is process-static, written without synchronization and never reset. Benign (idempotent, derived from a constant base dir) but it's mutable static state touched directly by unit tests — consider computing per-instance or documenting the intentional process-lifetime cache.

**9.** No `docs/Changelog-Platform.md` entry for the new package — add one when it's ready to ship.

**10.** `GetThresholdSeconds` in the slow-test reporter re-parses/validates the threshold already validated by the command-line provider. Minor DRY — consider passing the parsed value through.

**11.** AzDo links `RootFinder.cs` for git-root discovery; this package hand-rolls `GitHubActionsRepositoryRoot.FindGitRoot()`. If the semantics match, reuse `RootFinder`; if they differ, a one-line comment on why would help.

src/Platform/Microsoft.Testing.Extensions.GitHubActionsReport/GitHubActionsAnnotationReporter.cs
```diff
@@ -0,0 +1,252 @@
...
+    public Task<bool> IsEnabledAsync() => Task.FromResult(_isEnabled);
+
+    public async Task ConsumeAsync(IDataProducer dataProducer, IData value, CancellationToken cancellationToken)
+    {
+        cancellationToken.ThrowIfCancellationRequested();
+
+        if (value is not TestNodeUpdateMessage nodeUpdateMessage)
+        {
+            return;
+        }
```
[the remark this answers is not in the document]
azat-msft: The enabling/disabling the extension is handled by setting `isEnabled` property, no need to check it every time this method is called.

src/Platform/Microsoft.Testing.Extensions.GitHubActionsReport/TestNodeIdentity.cs
```diff
@@ -0,0 +1,57 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using Microsoft.Testing.Platform;
+using Microsoft.Testing.Platform.Extensions.Messages;
+
+namespace Microsoft.Testing.Extensions.GitHubActionsReport;
+
+internal static class TestNodeIdentity
```
Evangelink: This duplicates the shared `SharedExtensionHelpers/TestNodeIdentity.cs` that this same PR extracts for AzureDevOps — the two differ only by namespace and one doc sentence. The `.csproj` already `<Compile Include>`-links four other shared helpers (`SlowTestReporterBase`, `TargetFrameworkMonikerHelper`, …), so linking this one and deleting the local copy keeps a single source of truth. The test project already uses the `ghactions` alias, so the namespace collision is handled.

src/Platform/Microsoft.Testing.Extensions.GitHubActionsReport/GitHubActionsSummaryReporter.cs
```diff
@@ -0,0 +1,357 @@
...
+                    case TimingProperty t:
+                        timing = t;
+                        break;
+                    case SerializableKeyValuePairStringProperty kv when kv.Key == FullyQualifiedNamePropertyKey && fqnValue is null:
+                        fqnValue = kv.Value;
+                        break;
+                }
+            }
+
+            string fullyQualifiedName = fqnValue ?? displayName;
```
Evangelink: Name resolution here (`fqnValue ?? displayName`, where `fqnValue` only comes from the `vstest.TestCase.FullyQualifiedName` key) never consults `TestMethodIdentifierProperty`, unlike the annotation and slow-test reporters which use `TestNodeIdentity.GetTestName`. For natively-run MTP tests (which carry `TestMethodIdentifierProperty` but not the vstest key) the same test will appear as `DisplayName` in the summary but `Namespace.Type.Method` in the annotation. Consider reusing `TestNodeIdentity` so all three surfaces agree (this also removes the third copy of the `vstest.TestCase.FullyQualifiedName` constant declared at line 30).

src/Platform/Microsoft.Testing.Extensions.GitHubActionsReport/GitHubActionsSummaryReporter.cs
```diff
@@ -0,0 +1,357 @@
...
+            lock (_stateLock)
+            {
+                snapshot = [.. _records.Values];
+            }
+
+            string markdown = BuildMarkdown(snapshot, _testApplicationModuleInfo.TryGetAssemblyName() ?? "unknown assembly name", _targetFrameworkMoniker.Value);
+
+            try
+            {
+                using IFileStream stream = _fileSystem.NewFileStream(path!, FileMode.Append, FileAccess.Write, FileShare.Read);
```
Evangelink: `FileShare.Read` denies concurrent writers. When several test assemblies run in parallel in one GitHub job they all append to the same `GITHUB_STEP_SUMMARY` file, so a colliding process throws → caught below → summary silently dropped. `FileShare.ReadWrite` lets the OS serialize the appends (the whole markdown is written in a single `WriteAsync`, so this is a pragmatic fix).

src/Platform/Microsoft.Testing.Extensions.GitHubActionsReport/GitHubActionsSummaryReporter.cs
```diff
@@ -0,0 +1,357 @@
...
+        try
+        {
+            cancellationToken.ThrowIfCancellationRequested();
+
+            if (!_isEnabled || value is not TestNodeUpdateMessage update)
+            {
+                return Task.CompletedTask;
+            }
+
+            TestNodeStateProperty? state = update.TestNode.Properties.SingleOrDefault<TestNodeStateProperty>();
```
Evangelink: `SingleOrDefault<TestNodeStateProperty>()` throws if a malformed node carries more than one state property. The annotation reporter deliberately uses `FirstOrDefault` with a comment for exactly this reason. Both are inside try/catch so it degrades rather than crashes, but aligning on `FirstOrDefault` here (and in `SlowTestReporterBase`) would be more consistent and defensive.

src/Platform/Microsoft.Testing.Extensions.GitHubActionsReport/buildTransitive/Microsoft.Testing.Extensions.GitHubActionsReport.props
```diff
@@ -0,0 +1,3 @@
+﻿<Project>
+  <Import Project="$(MSBuildThisFileDirectory)..\..\buildMultiTargeting\Microsoft.Testing.Extensions.GitHubActionsReport.props" />
```
Evangelink: This imports `buildMultiTargeting` directly, but 14 of 16 Platform extensions (including the AzureDevOpsReport sibling) route `buildTransitive/<TFM>` through `build/<TFM>` via `..\..\build\$([System.IO.Path]::GetFileName($([System.IO.Path]::GetDirectoryName($(MSBuildThisFileDirectory)))))\<Name>.props`. Functionally equivalent, but worth matching the sibling for consistency (only VideoRecorder currently deviates the same way).

src/Platform/Microsoft.Testing.Extensions.GitHubActionsReport/GitHubActionsFeature.cs
```diff
@@ -0,0 +1,31 @@
...
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using Microsoft.Testing.Platform.CommandLine;
+using Microsoft.Testing.Platform.Helpers;
+
+namespace Microsoft.Testing.Extensions.GitHubActionsReport;
+
+/// <summary>
+/// Shared activation logic for the GitHub Actions reporters. The extension auto-activates when running
+/// on GitHub Actions (<c>GITHUB_ACTIONS=true</c>) or when the <c>--report-gh</c> master switch is set;
```
Evangelink: Doc/code mismatch: this says the extension activates on GitHub Actions "**or** when the `--report-gh` master switch is set", but `IsMasterEnabled` is `IsRunningOnGitHubActions(environment) && commandLine.IsOptionSet(...)` — i.e. **and** (matching AzureDevOps). The code is correct; please fix the wording to "and".

Evangelink (commented):
## Code review

Reviewed the full diff plus a second pass with an expert MSTest/Microsoft.Testing.Platform reviewer. Overall this is a strong, well-tested addition that faithfully mirrors the `AzureDevOpsReport` sibling — no blocking or high-severity bugs. The reporters are thread-safe, the escaper matches `actions/toolkit` exactly, the `::endgroup::`-last ordering trick is sound, and the `[TPEXP]` / `init` / PublicAPI discipline is correct. Findings below are maintainability/consistency oriented. Inline comments are attached to the specific spots.

### Main points

1. **Share `TestNodeIdentity` instead of duplicating it.** This PR extracts `SharedExtensionHelpers/TestNodeIdentity.cs` and wires AzureDevOps to reuse it, but the new project ships a near-identical copy differing only by namespace. The csproj already links four other shared helpers — link this one too and drop the local copy.
2. **Name resolution is inconsistent across the three reporters.** The summary reporter resolves names as `vstest FQN ?? DisplayName` and never consults `TestMethodIdentifierProperty`, so natively-run tests show as `DisplayName` in the job summary but `Namespace.Type.Method` in annotations/slow-test notices. Reusing `TestNodeIdentity.GetTestName` fixes this and removes a third copy of the `vstest.TestCase.FullyQualifiedName` magic string.
3. **`GITHUB_STEP_SUMMARY` opened with `FileShare.Read`** — parallel assemblies in one job collide and the summary is silently dropped (caught → warning). `FileShare.ReadWrite` lets the OS serialize appends.

### Smaller / consistency

- `buildTransitive/*.props` imports `buildMultiTargeting` directly, whereas 14/16 Platform extensions (including AzDO) route through `build/<TFM>`. Functionally equivalent; align for consistency.
- `SingleOrDefault<TestNodeStateProperty>()` (summary + slow-test base) throws on a malformed 2-state node, whereas the annotation reporter deliberately uses `FirstOrDefault` with a comment explaining why. Align on `FirstOrDefault`.
- The `GitHubActionsFeature` XML doc says the extension activates on GH Actions **or** `--report-gh`, but the code requires **both** (`IsRunningOnGitHubActions && IsOptionSet`). Code is correct; fix the doc.
- The group reporter must be registered last (same requirement AzDO documents with a well-known GUID + a prominent warning comment). GH uses an undocumented arbitrary GUID; worth documenting the intent, especially since the `AllMicrosoft` SDK profile enables AzDO and GH together.
- Annotation source-location tests don't cover the deterministic `/_/` remap, `GITHUB_WORKSPACE` root, or assertion-frame skipping (tests throw a raw `Exception`, never `Assert.Fail`).

### Verified good

- Versioning/experimental setup matches the JUnitReport + AzureFoundry precedent (version-prefix in `Directory.Build.props`, `SuppressFinalPackageVersion`, `[TPEXP]` in PublicAPI, no `init` accessors). `--help`/`--info`, known-extension registration, and SDK acceptance expectations are all updated and alphabetized.
- The `SlowTestReporterBase` extraction (moving AzDO's ~217-line reporter body into a shared base) is a clean, correct de-duplication.
- Minor scope note: renaming `GetTargetFrameworkMoniker()` → `GetTargetFrameworkMonikerIncludingPlatform()` and changing the fallback `"unknown"` → `"unknown framework"` affects shipped JUnit/Ctrf report filenames (adds a space when the TFM can't be resolved). No stale callers and no test pins the old literal, so low risk — just flagging the cross-package reach.

Evangelink:
Addressed all review feedback in 8b85b38 (built clean across net8.0/net9.0/netstandard2.0, `Microsoft.Testing.Extensions.UnitTests` green — 597 passed / 0 failed).

- **Share `TestNodeIdentity`** — deleted the project-local copy and `<Compile>`-linked `SharedExtensionHelpers/TestNodeIdentity.cs`.
- **Summary name resolution** — the summary reporter now resolves the name via `TestNodeIdentity.GetTestName` (prefers `TestMethodIdentifierProperty`) so it matches the annotation/slow-test surfaces; removed the duplicated `vstest.TestCase.FullyQualifiedName` constant.
- **`GITHUB_STEP_SUMMARY`** — opened with `FileShare.ReadWrite` so parallel assemblies don't drop the summary.
- **`SingleOrDefault` → `FirstOrDefault`** — applied to the summary reporter and the shared `SlowTestReporterBase`.
- **`buildTransitive` props** — now routes through `build/<TFM>` to match the sibling extensions' convention.
- **Activation doc** — fixed `GitHubActionsFeature` wording ("and", not "or").
- **Group-reporter GUID** — documented that endgroup-last is guaranteed by intra-provider registration order, not by GUID-based ordering.
- **Annotation coverage** — added tests for the deterministic `/_/` build-root remap and assertion-frame skipping.

Evangelink:
Followed up on the CLI-option feedback in f27fbff:

- The on/off knobs (`--report-gh-groups`/`-annotations`/`-step-summary`/`-slow-test-notices`) now validate through the shared `CommandLineOptionArgumentValidator`, so they accept `on`/`off`, `true`/`false`, `enable`/`disable`, and `1`/`0` (reading the disabled state via `IsOffValue`).
- `--report-gh-slow-test-threshold` now parses with the shared `TimeSpanParser` (default unit = seconds), so `60`, `90s`, `2m`, `1.5h` all work. `TimeSpanParser.cs` is Compile-linked the same way HangDump does it (it is an `[Embedded]` type hidden across assembly boundaries).
- Updated the option descriptions / invalid-value messages (mirroring the CrashDump wording), regenerated all `.xlf` locales, updated the `--help`/`--info` acceptance expectations, and added provider tests for the aliases and duration formats.

Evangelink:
Pushed two more commits (3e38093, c51947f) — built clean across net8.0/net9.0/netstandard2.0, `Microsoft.Testing.Extensions.UnitTests` green (610 passed / 0 failed).

**Hardcoded assertion FQNs (annotation reporter:37):** extracted the whole stack-trace source-location resolver — frame parsing, deterministic `/_/` handling, MSTest assertion-frame skipping, repo relativization — into `SharedExtensionHelpers/StackTraceSourceLocationResolver`, so AzureDevOps and GitHub Actions now share **one** copy of the assertion-prefix list instead of two. I also documented that the skip is only needed on **.NET Framework** (MSTest's `Assert`/`CollectionAssert`/`StringAssert` are `[StackTraceHidden]`, which the CLR honors on .NET Core 2.1+, so those frames are already absent) and gated it at runtime via `SkipAssertionFramesForCurrentRuntime`. It's a *runtime* check on purpose: the extensions ship as `netstandard2.0` and that build is what loads under .NET Framework, so a compile-time `#if NETFRAMEWORK` would never apply. Tests force the skip on explicitly, so coverage is unchanged.

**SlowTestReporterBase (:119, :240):** keep the first-seen start time on repeated `InProgress` updates (`TryAdd`) so progress heartbeats can't keep pushing the threshold out and mask a slow test; and skip emitting a notice for a test that completed mid-scan.

**Activation docs (PACKAGE.md:23/:27):** per the offline decision to keep the code's **AND** semantics, PACKAGE.md now states the extension activates when `GITHUB_ACTIONS=true` **and** `--report-gh` is passed (was implying auto-on-CI), and documents the threshold duration format. This matches the `GitHubActionsFeature` doc fix from the earlier commit.

**Summary test ordering (:61):** rewrote the slowest-first assertion with `IsLessThan` + a comment so the intended `slowIndex < fastIndex` relationship is unambiguous.

Already handled in earlier commits (f27fbff/8b85b38): shared `TestNodeIdentity`, summary name resolution via `TestMethodIdentifierProperty`, `FileShare.ReadWrite`, `FirstOrDefault` for the state property, `buildTransitive` routing, and the `GitHubActionsFeature` "and" doc.

Not changed, on purpose: the `ConsumeAsync` feature-flag check (annotation reporter:83) — as noted, `IsEnabledAsync` already gates the consumer; and `TestFramework.Extensions.csproj:107`, which is outside this PR's scope.

src/Platform/Microsoft.Testing.Extensions.GitHubActionsReport/TestNodeIdentity.cs
Evangelink: Done in 8b85b38 — deleted the project-local copy and Compile-linked the shared `SharedExtensionHelpers/TestNodeIdentity.cs`.

src/Platform/Microsoft.Testing.Extensions.GitHubActionsReport/GitHubActionsSummaryReporter.cs
Evangelink: Done in 8b85b38 — the summary reporter now resolves the name via `TestNodeIdentity.GetTestName` (which prefers `TestMethodIdentifierProperty`), so it matches the annotation/slow-test surfaces; the duplicate `vstest.TestCase.FullyQualifiedName` constant is gone.

src/Platform/Microsoft.Testing.Extensions.GitHubActionsReport/GitHubActionsSummaryReporter.cs
Evangelink: Done in 8b85b38 — the job summary is now opened with `FileShare.ReadWrite` so concurrent assemblies in one job don't drop it.

src/Platform/Microsoft.Testing.Extensions.GitHubActionsReport/GitHubActionsSummaryReporter.cs
Evangelink: Done in 8b85b38 — switched to `FirstOrDefault<TestNodeStateProperty>()` here and in the shared `SlowTestReporterBase`.

src/Platform/Microsoft.Testing.Extensions.GitHubActionsReport/buildTransitive/Microsoft.Testing.Extensions.GitHubActionsReport.props
Evangelink: Done in 8b85b38 — `buildTransitive` now routes through `build/<TFM>` like the sibling extensions.

src/Platform/Microsoft.Testing.Extensions.GitHubActionsReport/GitHubActionsFeature.cs
Evangelink: Done in 8b85b38 — fixed the XML doc to say "and" (the code's `IsMasterEnabled` requires both `GITHUB_ACTIONS=true` and `--report-gh`, which is the intended AND semantics).

src/Platform/SharedExtensionHelpers/SlowTestReporterBase.cs
```diff
@@ -0,0 +1,266 @@
...
+                return Task.CompletedTask;
+            }
+
+            string uid = update.TestNode.Uid;
+            TestNodeStateProperty? state = update.TestNode.Properties.FirstOrDefault<TestNodeStateProperty>();
+            if (state is InProgressTestNodeStateProperty)
+            {
+                string testName = GetTestName(update.TestNode);
+                TimeSpan threshold = ResolveThreshold(testName);
+                _inProgress[uid] = new InProgressTest(testName, _clock.UtcNow, threshold);
```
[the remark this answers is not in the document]
Evangelink: Done in 3e38093 — repeated `InProgress` updates now keep the first-seen start time (`TryAdd`), so progress heartbeats can't keep pushing the threshold out and mask a genuinely slow test.

src/Platform/Microsoft.Testing.Extensions.GitHubActionsReport/PACKAGE.md
```diff
@@ -0,0 +1,44 @@
...
+
+This package extends Microsoft.Testing.Platform with:
+
+- **Per-assembly log groups**: emits `::group::` / `::endgroup::` workflow commands so each test assembly's output is collapsed by default in the runner UI
+- **Failure annotations**: emits an `::error` workflow command for each failing test so failures appear in the workflow Annotations tab and, when the source location can be resolved, on the pull request's "Files changed" diff gutter
+- **Job summary**: appends a markdown roll-up (totals, failures, slowest tests) to the file pointed to by `GITHUB_STEP_SUMMARY`, which GitHub renders on the workflow run summary page
+- **Slow-test notices**: emits a `::notice` workflow command for any test still running past a threshold (default 60 seconds)
+- **CI auto-detection**: detects GitHub Actions environments through the `GITHUB_ACTIONS` variable
+
+All features are enabled by default when running on GitHub Actions (or when the `--report-gh` master switch is set), and no-op otherwise. Each feature can be toggled individually:
```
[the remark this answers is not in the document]
Evangelink: Done in c51947f — PACKAGE.md now states the extension activates when `GITHUB_ACTIONS=true` and `--report-gh` is passed (AND), matching the code.

src/Platform/Microsoft.Testing.Extensions.GitHubActionsReport/PACKAGE.md
```diff
@@ -0,0 +1,44 @@
...
+
+| Option | Description | Default |
+|---|---|---|
+| `--report-gh` | Master switch to enable the extension outside GitHub Actions | off |
```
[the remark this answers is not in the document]
Evangelink: Done in c51947f — the option table no longer claims `--report-gh` enables the extension outside GitHub Actions; it now says the switch is required in addition to running on GitHub Actions.

src/Platform/SharedExtensionHelpers/SlowTestReporterBase.cs
```diff
@@ -0,0 +1,266 @@
...
+
+            try
+            {
+                await EmitSlowTestAsync(test.TestName, elapsed, cancellationToken).ConfigureAwait(false);
+            }
+            catch (Exception ex) when (ex is not OperationCanceledException)
+            {
+                LogUnexpectedException(nameof(ScanOnceAsync), ex);
+            }
+        }
```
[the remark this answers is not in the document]
Evangelink: Done in 3e38093 — `ScanOnceAsync` now re-checks the in-progress map and skips emitting a notice for a test that completed (was removed) mid-scan.

src/Platform/SharedExtensionHelpers/TargetFrameworkMonikerHelper.cs
```diff
@@ -7,7 +7,12 @@ namespace Microsoft.Testing.Extensions;
...
-    public static string GetTargetFrameworkMoniker()
+    public static string GetTargetFrameworkMonikerIncludingPlatform()
         => TargetFrameworkParser.GetShortTargetFrameworkIncludingPlatform(Assembly.GetEntryAssembly())
-            ?? "unknown";
+            ?? "unknown framework";
+
+    public static string GetTargetFrameworkMonikerWithRuntimeFallback()
+        => TargetFrameworkParser.GetShortTargetFramework(Assembly.GetEntryAssembly()?.GetCustomAttribute<System.Runtime.Versioning.TargetFrameworkAttribute>()?.FrameworkDisplayName)
+            ?? TargetFrameworkParser.GetShortTargetFramework(System.Runtime.InteropServices.RuntimeInformation.FrameworkDescription)
+            ?? "unknown framework";
```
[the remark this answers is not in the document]
Evangelink: Good catch — done in 50c0d32. Removed `GetTargetFrameworkMonikerWithRuntimeFallback` entirely and switched both GitHub Actions reporters to `GetTargetFrameworkMonikerIncludingPlatform` (same as the AzureDevOps sibling). `GetShortTargetFrameworkIncludingPlatform` already has the runtime fallback built in and preserves the platform component (e.g. `net8.0-windows10.0.18362.0`), so the duplication is gone and platform-specific TFMs are no longer flattened.

src/Platform/Microsoft.Testing.Extensions.GitHubActionsReport/GitHubActionsSlowTestReporter.cs
```diff
@@ -0,0 +1,80 @@
...
+            CultureInfo.InvariantCulture,
+            "::notice title={0}::{1}",
+            GitHubActionsEscaper.EscapeProperty(title),
+            GitHubActionsEscaper.EscapeData(message));
+    }
+
+    // Re-reads the already-provider-validated threshold option (the CLI provider guarantees a parseable
+    // positive duration). Accepts a bare number of seconds or a value with a unit suffix (e.g. '90s', '2m').
+    // This mirrors the sibling AzureDevOpsSlowTestReporter, which likewise reads its history options straight
+    // from ICommandLineOptions rather than threading a parsed value through.
```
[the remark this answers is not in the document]
Evangelink: Done in 50c0d32 — reworded to "the threshold option, which the CLI provider has already validated".

Evangelink (approved):

PR #9565, opened 2026-07-03
Avoid MSB4011 duplicate-import warnings when layering MSTest.Sdk on another base SDK

### LINKED ISSUE #9562: Explore reducing MSB4011 duplicate-import warnings when MSTest.Sdk is combined with another base SDK
### Summary

`MSTest.Sdk` implicitly imports `Microsoft.NET.Sdk` as its base SDK (both `Sdk/Sdk.props.template` and `Sdk/Sdk.targets` do `<Import ... Sdk=""Microsoft.NET.Sdk"" />`). This means it can't be used as the outer `<Project Sdk=""..."">` for projects that need a different base SDK such as `Microsoft.NET.Sdk.Web` (ASP.NET Core integration test projects).

A workaround is to hand-write the SDK imports for both SDKs instead of using the `Sdk` attribute:

```xml
<Project>
  <Import Project=""Sdk.props"" Sdk=""Microsoft.NET.Sdk.Web"" />
  <Import Project=""Sdk.props"" Sdk=""MSTest.Sdk"" />
  <PropertyGroup>
  </PropertyGroup>
  <Import Project=""Sdk.targets"" Sdk=""MSTest.Sdk"" />
  <Import Project=""Sdk.targets"" Sdk=""Microsoft.NET.Sdk.Web"" />
</Project>
```

This **functionally works** — the project restores, builds, references ASP.NET Core (`WebApplication.CreateBuilder()`), and runs under the MSTest runner (`Passed! Failed: 0, Passed: 2`). However, because both `MSTest.Sdk` and `Microsoft.NET.Sdk.Web` each import `Microsoft.NET.Sdk`, the build emits duplicate-import warnings:

```
warning MSB4011: "...\Microsoft.NET.Sdk\Sdk\Sdk.props" cannot be imported again.
It was already imported at "...\Microsoft.NET.Sdk.Web\Targets\Sdk.Server.props".
This is most likely a build authoring error. This subsequent import will be ignored.
```

### Ask

Explore whether we can make the mixed/manual-import scenario clean, e.g.:

- Guard our base-SDK import with a condition that skips it when `Microsoft.NET.Sdk` (or a derived SDK such as `Microsoft.NET.Sdk.Web`) has already been imported (for example, check a well-known property set by the base SDK props), or
- Provide a documented, first-class way to layer `MSTest.Sdk` on top of a different base SDK without the `MSB4011` noise.

### Context

Came up while reviewing dotnet/docs#54554, which documents that `MSTest.Sdk` can't replace a different base SDK and recommends configuring MSTest manually for such projects. It would be nice if combining SDKs were a supported, warning-free path.

Evangelink:
Fixes #9562.

## Problem

`MSTest.Sdk` implicitly imports `Microsoft.NET.Sdk` as its base SDK in both `Sdk.props` and `Sdk.targets`. When `MSTest.Sdk` is layered on top of a *different* base SDK (for example `Microsoft.NET.Sdk.Web` for an ASP.NET Core integration test project) using manual SDK imports, the base `Microsoft.NET.Sdk` gets imported twice and the build emits `MSB4011` duplicate-import warnings:

```
warning MSB4011: "...\Microsoft.NET.Sdk\Sdk\Sdk.props" cannot be imported again.
It was already imported at "...\Microsoft.NET.Sdk.Web\Targets\Sdk.Server.props".
```

## Fix

Guard the `Microsoft.NET.Sdk` import behind a new `_MSTestSdkImportsMicrosoftNETSdk` property. `Microsoft.NET.Sdk` sets `UsingMicrosoftNETSdk=true` very early (before anything else), so `MSTest.Sdk` can detect whether an outer SDK has already imported the base SDK and skip re-importing it. The decision is recorded in `Sdk.props` and reused in `Sdk.targets` so both stay in sync — when layered, the outer SDK owns the base import.

The common `<Project Sdk="MSTest.Sdk">` usage is unchanged: `UsingMicrosoftNETSdk` is unset at that point, so `MSTest.Sdk` still imports the base SDK exactly as before.

## Changes

- `src/Package/MSTest.Sdk/Sdk/Sdk.props.template` — compute `_MSTestSdkImportsMicrosoftNETSdk` and gate the base props import on it.
- `src/Package/MSTest.Sdk/Sdk/Sdk.targets` — gate the base targets import on the same property.
- `src/Package/MSTest.Sdk/README.md` — document the first-class layering pattern (import the base SDK first).
- `test/IntegrationTests/MSTest.Acceptance.IntegrationTests/SdkTests.cs` — new acceptance test `MSTestSdk_LayeredOnTopOfWebSdk_BuildsWithoutDuplicateImportWarnings` that builds a `Microsoft.NET.Sdk.Web` + `MSTest.Sdk` project, asserts it runs, and asserts no `MSB4011` in the output.

## Verification

Reproduced the scenario with the real base SDKs plus a control:

| Scenario | `... cannot be imported again` (MSB4011) |
| --- | --- |
| Mixed Web SDK, **unguarded** (before) | 12 |
| Mixed Web SDK, **guarded** (after) | 0 |
| Normal single-SDK (after) | 0 — base SDK still imported, compiles cleanly |

### DIFF
--- a/src/Package/MSTest.Sdk/README.md
+++ b/src/Package/MSTest.Sdk/README.md
@@ -18,3 +18,30 @@ Instead, the `.targets` files should split the version specification depending o
   matching `PackageVersion` item carrying the `Version` metadata so CPM resolves the version.
 
 This approach works regardless of `CentralPackageVersionOverrideEnabled` and does not produce `NU1009` warnings.
+
+## Layering on top of a different base SDK
+
+`MSTest.Sdk` implicitly imports `Microsoft.NET.Sdk` as its base SDK, so `<Project Sdk="MSTest.Sdk">`
+is enough for the common case. To combine it with a different base SDK (for example
+`Microsoft.NET.Sdk.Web` for an ASP.NET Core integration test project), import both SDKs manually and
+list `Microsoft.NET.Sdk.Web` first so it owns the base `Microsoft.NET.Sdk` import:
+
+```xml
+<Project>
+  <Import Project="Sdk.props" Sdk="Microsoft.NET.Sdk.Web" />
+  <Import Project="Sdk.props" Sdk="MSTest.Sdk" />
+
+  <PropertyGroup>
+    <TargetFramework>net10.0</TargetFramework>
+  </PropertyGroup>
+
+  <Import Project="Sdk.targets" Sdk="MSTest.Sdk" />
+  <Import Project="Sdk.targets" Sdk="Microsoft.NET.Sdk.Web" />
+</Project>
+```
+
+`Sdk.props`/`Sdk.targets` guard their `Microsoft.NET.Sdk` import behind the
+`_MSTestSdkImportsMicrosoftNETSdk` property: `MSTest.Sdk` only imports the base SDK when no other SDK
+has already done so (detected via the `UsingMicrosoftNETSdk` property that `Microsoft.NET.Sdk` sets
+very early). This keeps the manual/mixed layering scenario free of `MSB4011` duplicate-import
+warnings.
--- a/src/Package/MSTest.Sdk/README.md
+++ b/src/Package/MSTest.Sdk/README.md
@@ -40,6 +40,17 @@ list `Microsoft.NET.Sdk.Web` first so it owns the base `Microsoft.NET.Sdk` impor
 </Project>
 ```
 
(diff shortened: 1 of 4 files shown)


Evangelink:
Let's move forward with the merge, we will update if we get some different input from @rainersigwald

0101 (approved):

PR #9726, opened 2026-07-08
[v5] Remove VSTest support from the test adapter and make MTP the default

Evangelink:
## Summary

First slice of the v5 work to run MSTest exclusively on Microsoft.Testing.Platform (MTP). This targets the long-lived `dev/v5` branch.

This removes the VSTest adapter extension surface (`ITestExecutor`, `ITestDiscoverer`, registration attributes and the public VSTest entry points) from `MSTest.TestAdapter`. MSTest now runs only as an MTP application.

This does not remove compatibility with VSTest clients. Starting with VSTest 18.10, `vstest.console` detects an MTP application and hosts it over the MTP protocol. This PR pins `Microsoft.TestPlatform` and `Microsoft.NET.Test.Sdk` to `18.10.0-preview-26365-01`.

The acceptance tests cover the two entry points separately: normal cross-platform `dotnet test` with the MTP runner selected, and direct packaged VSTest 18.10 hosting. The packaged-client tests use the net462 console on Windows and `Microsoft.TestPlatform.CLI` through the repo dotnet muxer on Linux and macOS.

## Changes

- Remove `ITestExecutor`, `ITestDiscoverer`, VSTest extension registration and the unused `EqtTraceLogger` bootstrap.
- Make the adapter set `IsTestingPlatformApplication=true` by default and report leftover `EnableMSTestRunner` values as an informational migration message.
- Keep the internal discovery and execution engines used by `MSTestBridgedTestFramework`.
- Update adapter unit tests to use the internal MTP discovery and execution APIs.
- Keep cross-platform `dotnet test` coverage for the normal MTP runner path.
- Run SDK, Aspire, Playwright and telemetry scenarios through packaged VSTest 18.10 on Windows, Linux and macOS. Keep the `net462` AppDomain scenario Windows-only.
- Remove `Microsoft.NET.Test.Sdk` from the v5 `MSTest` meta-package. MTP build integration already flows through `MSTest.TestAdapter`, and VSTest 18.10 can host the produced MTP application directly. The package validation rejects reintroducing the classic Test SDK dependency.
- Keep both browser-wasm assets on the supported `MSTest` meta-package path. Removing the obsolete Test SDK dependency prevents `Microsoft.CodeCoverage` from copying its incompatible `System.Memory.dll` into the browser AppBundle.
- Remove the raw C++/CLI DLL VSTest test. A DLL without an MTP entry point is not an MTP application and cannot use the removed adapter surface. `CppCliMtpTests` remains the supported C++/CLI scenario and builds a managed C++/CLI executable that hosts MTP directly.
- Suppress the new obsolete `IFrameworkHandle.EnableShutdownAfterTestRun` warning in the VSTest bridge while preserving the interface implementation.

## Verification

- `build.cmd -pack`
- Packed `MSTest` nuspec has no `Microsoft.NET.Test.Sdk`/TestHost dependency
- 20 focused acceptance cases covering MTP-selected `dotnet test`, packaged VSTest run/discovery, SDK, Aspire, Playwright, telemetry and AppDomain scenarios
- Managed `Microsoft.TestPlatform.CLI` discovery against an MTP application
- `BrowserWasmExecution_RunsTestsUnderNode`
- `BrowserWasmExecution_RunSettingsEnvironmentVariables_FailsWithClearDiagnostic`
- Representative analyzer and assembly-resolution tests using the `MSTest` meta-package

## Follow-ups

- Remove `EnableMSTestRunner` and VSTest coexistence from the remaining MTP MSBuild targets.
- Remove `UseVSTest` and `Sdk/VSTest` from `MSTest.Sdk`.
- Remove the remaining VSTestBridge object-model coupling from the adapter.
- Remove the `isMTP` flag threaded through the adapter execution pipeline.
- Update samples and documentation for the final v5 shape.

🤖


src/Adapter/MSTest.TestAdapter/Properties/AssemblyInfo.cs
```diff
@@ -1,7 +1,2 @@
-﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Copyright (c) Microsoft Corporation. All rights reserved.
```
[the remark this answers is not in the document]
Evangelink: Good catch — fixed. AssemblyInfo.cs is re-saved as UTF-8 with BOM (first bytes EF BB BF) to match .editorconfig's charset = utf-8-bom.

src/Adapter/MSTest.TestAdapter/buildTransitive/common/MSTest.TestAdapter.targets
```diff
@@ -6,12 +6,14 @@
...
 
-  <!-- Handle the coexistence between testing platform and Microsoft.NET.Test.Sdk  -->
+  <!--
+    MSTest v5 always runs on Microsoft.Testing.Platform (MTP). The legacy VSTest opt-in property
+    'EnableMSTestRunner' has been removed: the adapter is always a testing platform application.
+  -->
   <PropertyGroup>
-    <EnableMSTestRunner Condition=" '$(EnableMSTestRunner)' == '' ">false</EnableMSTestRunner>
-    <IsTestingPlatformApplication>$(EnableMSTestRunner)</IsTestingPlatformApplication>
+    <IsTestingPlatformApplication Condition=" '$(IsTestingPlatformApplication)' == '' ">true</IsTestingPlatformApplication>
```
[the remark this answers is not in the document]
Evangelink: Good call — I went ahead and added the transitional warning in this slice rather than deferring it. Setting `EnableMSTestRunner` (any value) now emits a build warning telling users it's obsolete in v5 and to remove it, suppressible via `<SuppressMSTestRunnerObsoleteWarning>true</SuppressMSTestRunnerObsoleteWarning>`. Verified it fires when set, is silent when unset, and honors the suppression switch.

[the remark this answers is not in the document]
Evangelink: Update: I initially added this as a build warning, but that turned out too aggressive — a `<Warning>` is promoted to an error under `MSBuildTreatWarningsAsErrors=true` (which lots of our test-asset builds enable), so it hard-failed every project that still sets the harmless `EnableMSTestRunner` property. I've downgraded it to a high-importance `<Message>` so users still get the migration guidance without breaking warnings-as-errors builds (suppression switch is now `SuppressMSTestRunnerObsoleteMessage`). Verified it prints under `MSBuildTreatWarningsAsErrors=true` with exit code 0.

test/IntegrationTests/MSTest.Acceptance.IntegrationTests/SdkTests.cs
```diff
@@ -62,28 +67,36 @@ public void TestMethod1()
...
             AssetName,
-            SingleTestSourceCodeVSTest
+            SingleTestSourceCode
             .PatchCodeWithReplace("$MSTestVersion$", MSTestVersion)
             .PatchCodeWithReplace("$TargetFramework$", multiTfm)
-            .PatchCodeWithReplace("$ExtraProperties$", "<UseVSTest>true</UseVSTest>"));
+            .PatchCodeWithReplace("$ExtraProperties$", "<TestWithPackagedVSTest>true</TestWithPackagedVSTest>"));
 
-        DotnetMuxerResult compilationResult = await DotnetCli.RunAsync($"test -c {buildConfiguration} {testAsset.TargetAssetPath}", workingDirectory: testAsset.TargetAssetPath, cancellationToken: TestContext.CancellationToken);
+        DotnetMuxerResult compilationResult = await DotnetCli.RunAsync($"build -c {buildConfiguration} {testAsset.TargetAssetPath}", workingDirectory: testAsset.TargetAssetPath, cancellationToken: TestContext.CancellationToken);
```
Evangelink, line 91: @nohwnd is it expected that we cannot run with `dotnet test`?

test/IntegrationTests/MSTest.Acceptance.IntegrationTests/PlaywrightSdkTests.cs
```diff
@@ -37,32 +38,35 @@ public async Task EnablePlaywrightProperty_WhenUsingRunner_AllowsToRunPlaywright
     }
 
     [TestMethod]
+    [OSCondition(OperatingSystems.Windows)]
```
Evangelink: Calling vstest console should work x-plat, isn't it?

test/IntegrationTests/MSTest.Acceptance.IntegrationTests/TelemetryTests.cs
```diff
@@ -318,6 +316,7 @@ public void TestWithTimeout()
     <PackageReference Include="Microsoft.NET.Test.Sdk" Version="$MicrosoftNETTestSdkVersion$" />
     <PackageReference Include="MSTest.TestAdapter" Version="$MSTestVersion$" />
     <PackageReference Include="MSTest.TestFramework" Version="$MSTestVersion$" />
+    <PackageDownload Include="Microsoft.TestPlatform" Version="[$MicrosoftNETTestSdkVersion$]" />
```
nohwnd, line 313: exact version here is probably overkill?

test/IntegrationTests/MSTest.Acceptance.IntegrationTests/SdkTests.cs
nohwnd: No, it should work. I kept this separate from the direct packaged-client test and renamed the SDK-routing case to `RunTests_With_DotnetTest_UsesMTP`. The generated asset now has its own MTP `global.json` and runs `dotnet test` on every OS.

🤖

test/IntegrationTests/MSTest.Acceptance.IntegrationTests/PlaywrightSdkTests.cs
nohwnd: Yes, the Windows restriction was wrong. The helper now runs the packaged net462 `vstest.console.exe` on Windows and `Microsoft.TestPlatform.CLI`'s `vstest.console.dll` through the repo `DotnetMuxer` on Linux and macOS. I removed the OS condition from SDK, Aspire, Playwright and telemetry tests. AppDomain and C++/CLI stay Windows-only because their inputs are platform-specific.

🤖

test/IntegrationTests/Microsoft.Testing.Platform.Acceptance.IntegrationTests/BrowserWasmExecutionTests.cs
```diff
@@ -91,7 +91,8 @@ bundle ships ICU by default.
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="MSTest" Version="$MSTestVersion$" />
+    <PackageReference Include="MSTest.TestAdapter" Version="$MSTestVersion$" />
+    <PackageReference Include="MSTest.TestFramework" Version="$MSTestVersion$" />
```
nohwnd: also no? step backwards?

test/IntegrationTests/Microsoft.Testing.Platform.Acceptance.IntegrationTests/BrowserWasmExecutionTests.cs
```diff
@@ -180,7 +181,8 @@ public void FailingTest()
...
```
nohwnd: also no? step backwards?

test/IntegrationTests/MSTest.Acceptance.IntegrationTests/TelemetryTests.cs
nohwnd: The brackets are required here. NuGet `PackageDownload` only supports exact versions, written as `[version]`. Pinning is also intentional because this test runs the exact VSTest client version under test, rather than whichever compatible version restore might select.

🤖

test/IntegrationTests/Microsoft.Testing.Platform.Acceptance.IntegrationTests/BrowserWasmExecutionTests.cs
nohwnd: Agree, separate adapter/framework references were a step backwards. I restored `MSTest` here and fixed the package instead: v5 no longer depends on `Microsoft.NET.Test.Sdk`, which was pulling `Microsoft.CodeCoverage` build assets and copying its `System.Memory.dll` into the browser bundle. This test now passes through the supported meta-package path.

🤖

test/IntegrationTests/Microsoft.Testing.Platform.Acceptance.IntegrationTests/BrowserWasmExecutionTests.cs
nohwnd: Same fix here. The runsettings browser asset uses `MSTest` again, and the v5 meta-package drops the obsolete `Microsoft.NET.Test.Sdk` dependency at the source. This keeps the acceptance test on the real user path without putting the incompatible code coverage `System.Memory.dll` into the AppBundle.

🤖

PR #9730, opened 2026-07-08
Add [self, nextMajor) upper bound on extensions' Microsoft.Testing.Platform dependency

Evangelink:
## Summary

The Microsoft.Testing.Platform extensions (`src/Platform/Microsoft.Testing.Extensions.*`, plus `Microsoft.Testing.Platform.MSBuild`) reference `Microsoft.Testing.Platform` via a `ProjectReference`. When packed, NuGet emits that as an **open-ended, minimum-version** dependency — e.g. `Microsoft.Testing.Platform version="2.4.0"`, which means `[2.4.0, )`.

That open upper bound lets an extension built against Testing Platform **2.x** silently resolve a future, potentially breaking **3.0.0**. Since the extensions and the platform are versioned and ship in lockstep, this PR caps the packed dependency at the next major as a safety net:

```
Microsoft.Testing.Platform version="[2.4.0-…, 3.0.0)"
```

## Changes

- **`src/Platform/Directory.Build.props`** — introduces `TestingPlatformNextMajorVersion`, computed as *(major of `TestingPlatformVersionPrefix`) + 1* `.0.0` (i.e. `3.0.0` for the current 2.x line).
- **`src/Platform/Directory.Build.targets`** (new) — a pack-time target (`AfterTargets="_GetProjectReferenceVersions"`) that rewrites the packed `Microsoft.Testing.Platform` dependency's version to `[self, nextMajor)`. It:
  - only touches the `Microsoft.Testing.Platform` entry (sibling deps such as `TrxReport.Abstractions`, `Telemetry`, `Platform.AI` are left open-ended);
  - preserves other dependency metadata (e.g. `exclude="Build,Analyzers"`);
  - is idempotent (skips entries that are already a range);
  - imports the repo-root `Directory.Build.targets` via `GetPathOfFileAbove` so nothing is lost by introducing this file.

## Verification

Packed several projects locally and inspected the generated `.nuspec` dependencies:

| Package | `Microsoft.Testing.Platform` dep |
| --- | --- |
| `…Extensions.Retry` | `[2.4.0-dev, 3.0.0)` ✅ |
| `…Extensions.CrashDump` | `[2.4.0-dev, 3.0.0)` (sibling `TrxReport.Abstractions` stays `2.4.0-dev`) ✅ |
| `…Extensions.VSTestBridge` | `[2.4.0-dev, 3.0.0)` (siblings `Telemetry`, `TrxReport.Abstractions` stay open) ✅ |
| `Microsoft.Testing.Platform.MSBuild` | `[2.4.0-dev, 3.0.0)` ✅ |
| `Microsoft.Testing.Platform` | no self-dependency, unaffected ✅ |
| `…Extensions.AzureFoundry` | references `Platform.AI` only — unaffected (out of scope) ✅ |

Scope note: this targets the direct `Microsoft.Testing.Platform` dependency specifically. `Microsoft.Testing.Platform.AI` (referenced only by AzureFoundry) is intentionally not bounded here.

Co-authored-by: Copilot App <223556219+Copilot@users.noreply.github.com>

### DIFF
--- a/src/Platform/Directory.Build.props
+++ b/src/Platform/Directory.Build.props
@@ -6,6 +6,12 @@
   <PropertyGroup>
     <!-- Platform and extensions use TestingPlatform version, not MSTest version. -->
     <VersionPrefix>$(TestingPlatformVersionPrefix)</VersionPrefix>
+
+    <!-- Exclusive upper bound used as a safety net on the in-repo Microsoft.Testing.Platform package
+         dependency of the extensions (see Directory.Build.targets): the next major after the current
+         Testing Platform version. For a 2.x platform this evaluates to 3.0.0, producing a [self, 3.0.0)
+         dependency range so an extension built against 2.x never silently resolves a breaking next major. -->
+    <TestingPlatformNextMajorVersion Condition=" '$(TestingPlatformNextMajorVersion)' == '' ">$([MSBuild]::Add($(TestingPlatformVersionPrefix.Split('.')[0]), 1)).0.0</TestingPlatformNextMajorVersion>
   </PropertyGroup>
 
   <!-- Build config -->
--- a/src/Platform/Directory.Build.targets
+++ b/src/Platform/Directory.Build.targets
@@ -1,0 +1,29 @@
+<Project>
+
+  <Import Project="$([MSBuild]::GetPathOfFileAbove('Directory.Build.targets', '$(MSBuildThisFileDirectory)../'))" />
+
+  <!--
+    Add a [self, nextMajor) upper bound on the in-repo Microsoft.Testing.Platform package dependency that
+    the extensions emit through their ProjectReference to Microsoft.Testing.Platform.csproj.
+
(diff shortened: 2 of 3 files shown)


Evangelink:
Addressed the two review suggestions in 2bf3ac195:

- **String.Copy nit** — swapped the obsolete System.String::Copy idiom for System.String::new in the pack-time target condition.
- **Pack-verification hardening** — added PackageDependencyUpperBoundTests, an acceptance test that opens the produced .nuspec files and asserts every packed Microsoft.Testing.Platform dependency is a closed [min, nextMajor) range (upper bound read from ng/Versions.props). This fails if the NuGet-internal _GetProjectReferenceVersions / _ProjectReferencesWithVersions targets are ever renamed and the bound silently reverts to open-ended. Verified it passes on the real packages and fails on a bare-version package (3 violations, one per TFM group).

0101 (approved):

PR #9771, opened 2026-07-09
Fix IndexOutOfRangeException in VSTestBridge BuildFilter UID escaping

azat-msft:
## Summary

`ContextAdapterBase.BuildFilter` reconstructs a VSTest `TestCaseFilter` string from a `TestNodeUidListFilter` — the by-UID test selection that IDEs (e.g. Visual Studio Test Explorer) send over the Microsoft.Testing.Platform server protocol. Its hand-rolled escaper had an index bug: the "is this character already escaped?" guard used the **node index `i`** instead of the **character index `k`**:

```csharp
if (i - 1 < 0 || currentTestNodeUid.Value[k - 1] != '\\')
```

It reads `Value[k - 1]` but tests `i - 1 < 0`. So for any node **after the first** (`i > 0`) whose **first** character (`k == 0`) is a filter operator (`\ ( ) & | = ! ~`), it evaluates `Value[-1]` and throws `IndexOutOfRangeException`, aborting the whole run. The same index confusion could also under-escape an operator that follows a literal backslash, splitting a `FullyQualifiedName` value at what should be an escaped operator.

## Fix

Replace the hand-rolled loop with VSTest's canonical `FilterHelper.Escape`, which escapes every filter operator unconditionally — exactly what a raw (un-escaped) test-node UID needs — removing the index bug entirely.

## Tests

Adds `RunContextAdapterFilterTests` covering `BuildFilter` output through the public `RunContextAdapter.GetTestCaseFilter(...)`:

- `FullyQualifiedName=` vs `Id=<guid>` selection (and mixed), multi-node `|` (OR) joining.
- Runsettings `TestCaseFilter` and command-line `--filter`, and their `&` combination with a UID selection.
- Special-character escaping for all operators (each operator escaped individually).
- Index-dimension coverage: a **single** node (`i == 0`) and a **later** node (`i > 0`) whose name starts with a filter operator both escape and no longer throw.
- Backslash-then-operator escapes **both** characters.
- Empty `TestNodeUidListFilter([])` behavior pinned.
- Round-trip `MatchTestCase` tests that assert the built filter actually matches the exact test case and rejects near-misses, guarding the full escape → parse → match path.

All 20 tests pass (net8.0, net9.0, net462).


test/UnitTests/Microsoft.Testing.Extensions.VSTestBridge.UnitTests/ObjectModel/RunContextAdapterFilterTests.cs
```diff
@@ -0,0 +1,213 @@
...
+        // the character index (k) when looking back at the previous character. For any node after the
+        // first (i > 0) whose first character (k == 0) is a filter operator, this evaluated
+        // Value[k - 1] == Value[-1] and threw IndexOutOfRangeException.
+        RunContextAdapter adapter = CreateAdapter(EmptyRunSettings, CreateUidFilter("A.B.Test", "(weird"));
+
+        Assert.AreEqual("(FullyQualifiedName=A.B.Test|FullyQualifiedName=\\(weird)", GetFilterValue(adapter));
+    }
+
+    [TestMethod]
+    public void GetTestCaseFilter_WithBackslashFollowedBySpecialCharacter_EscapesBoth()
```
Evangelink, line 208: Nit (non-blocking): this method bundles two distinct scenarios — `A.B|C` (operator only) and `A.B\|C` (literal backslash + operator, the case the old guard under-escaped). Consider splitting into two `[TestMethod]`s or a `[DataRow]`-driven test so a failure points at the exact case. Both expected strings are correct as written.

test/UnitTests/Microsoft.Testing.Extensions.VSTestBridge.UnitTests/ObjectModel/RunContextAdapterFilterTests.cs
```diff
@@ -0,0 +1,213 @@
...
+    </RunConfiguration>
+</RunSettings>
+""";
+        RunContextAdapter adapter = CreateAdapter(runSettings, new NopFilter(), commandLineFilter: "Priority=1");
+
+        Assert.AreEqual("(Category=Fast) & (Priority=1)", GetFilterValue(adapter));
+    }
+
+    [TestMethod]
+    public void GetTestCaseFilter_WithSecondNodeStartingWithSpecialCharacter_DoesNotThrowAndEscapes()
```
Evangelink, line 186: Optional: a single-node (`i == 0`) UID whose first char is an operator (e.g. `"(weird"`) never threw under the old code because `i - 1 < 0` short-circuited. Adding that case alongside this one would make the coverage exhaustive over the index dimension. Genuinely optional — content escaping is already well covered.

Evangelink (commented):
Reviewed in depth (manual pass + a second expert-review pass) and built/ran the test project locally — **all 14 tests pass on net8.0, net9.0 and net462, 0 warnings**.

**Verdict: LGTM.** The root cause is correctly diagnosed: the old escaper guard bounds-checked the node index `i` but indexed with the char index `k` (`Value[k - 1]` while testing `i - 1 < 0`), so any node after the first whose first char is a filter operator read `Value[-1]` and threw `IndexOutOfRangeException`; it also under-escaped an operator following a literal backslash. Replacing it with `FilterHelper.Escape` (VSTest canonical escaper, already reachable via the imported `...Common.Filtering` namespace) is the right fix — MTP sends raw, unescaped UIDs, so escaping every operator unconditionally is exactly correct, and the `Escape` -> `FilterExpressionWrapper`/`Condition` tokenizer -> `Unescape` round-trip is lossless. I independently decoded every expected string in the tests against the actual output of `Escape` and they all match, including the subtle `A.B\\\|C` triple-backslash case.

Only a few optional, non-blocking follow-ups below.

- **(Low)** An empty UID list (`new TestNodeUidListFilter([])`) is constructible and yields the filter string `"()"` (pre-existing behavior, not introduced here). Since this PR adds the definitive filter-building suite, a test pinning that outcome would be nice to have.

No public API added; assertion style and conventions are compliant. Thanks for the thorough tests!

test/UnitTests/Microsoft.Testing.Extensions.VSTestBridge.UnitTests/ObjectModel/RunContextAdapterFilterTests.cs
azat-msft: Done — split into two focused [TestMethod]s: GetTestCaseFilter_WithOperator_EscapesOperator (bare A.B|C) and GetTestCaseFilter_WithBackslashFollowedBySpecialCharacter_EscapesBoth (A.B\|C), so a failure now points at the exact case.

test/UnitTests/Microsoft.Testing.Extensions.VSTestBridge.UnitTests/ObjectModel/RunContextAdapterFilterTests.cs
azat-msft: Added — `GetTestCaseFilter_WithSingleNodeStartingWithSpecialCharacter_DoesNotThrowAndEscapes` covers the single-node (`i == 0`) `"(weird"` case, so coverage is now exhaustive over the index dimension.

azat-msft:
Addressed the follow-ups from your review:

- Split the operator vs. backslash+operator case into two focused `[TestMethod]`s (`GetTestCaseFilter_WithOperator_EscapesOperator` and `GetTestCaseFilter_WithBackslashFollowedBySpecialCharacter_EscapesBoth`).
- Added the single-node `"(weird"` case (`GetTestCaseFilter_WithSingleNodeStartingWithSpecialCharacter_DoesNotThrowAndEscapes`) so the index dimension is exhaustively covered.
- Added `GetTestCaseFilter_WithEmptyNodeList_ThrowsOnEmptyGroup` for the empty `TestNodeUidListFilter([])`. Note: while `BuildFilter` does produce the `()` string, going through the public `GetTestCaseFilter` the VSTest parser rejects the empty parenthesis and throws `TestPlatformFormatException` (with `FilterValue == "()"`), so the test pins that actual observable behavior.

All tests pass locally on net8.0 (net9.0 SDK is not installed on my box, but the change is test-only).

test/UnitTests/Microsoft.Testing.Extensions.VSTestBridge.UnitTests/ObjectModel/RunContextAdapterFilterTests.cs
```diff
@@ -0,0 +1,293 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using Microsoft.Testing.Extensions.VSTestBridge.CommandLine;
+using Microsoft.Testing.Extensions.VSTestBridge.ObjectModel;
+using Microsoft.Testing.Platform.CommandLine;
+using Microsoft.Testing.Platform.Requests;
```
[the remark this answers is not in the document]
Evangelink: Fixed in 22d9268. `TestNodeUid` lives in `Microsoft.Testing.Platform.Extensions.Messages`, which `CreateUidFilter` uses via `new TestNodeUid(uid)` — so removing that using broke the build with `CS0246` on every TFM (net8.0/net9.0/net462), as CI confirmed (build 1501720). Re-added the directive in alphabetical order; the project now builds warning-clean and all 20 tests pass. Good catch — this correctly reverts the earlier (incorrect) unused-using removal.

test/UnitTests/Microsoft.Testing.Extensions.VSTestBridge.UnitTests/ObjectModel/RunContextAdapterFilterTests.cs
```diff
@@ -0,0 +1,293 @@
...
+namespace Microsoft.Testing.Extensions.VSTestBridge.UnitTests.ObjectModel;
+
+/// <summary>
+/// Tests for the filter building logic in <see cref="ContextAdapterBase"/> (exercised through
+/// <see cref="RunContextAdapter"/>), most notably the private <c>BuildFilter</c> method that turns a
+/// <see cref="TestNodeUidListFilter"/> coming from the server protocol into a VSTest
+/// <c>TestCaseFilter</c> expression.
+/// </summary>
+[TestClass]
+public sealed class RunContextAdapterFilterTests
```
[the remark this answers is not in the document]
Evangelink: Updated the PR description: the ``## Tests`` section now says 20 tests (matching the 20 ``[TestMethod]`` in the suite) and lists the added cases (single-node index coverage, empty-list, and the round-trip ``MatchTestCase`` tests). Thanks for flagging.

Evangelink (approved):

PR #9780, opened 2026-07-09
Align CTRF report writer overwrite behavior and add public IArtifactNamingService

Evangelink:
## Summary

Two related changes to MTP report file naming.

### 1. Align CTRF report writer to overwrite + warn (matches TRX/HTML/JUnit)

`CtrfReportEngine` was the only report extension that, for a **default-generated** file name, used `FileMode.CreateNew` plus a `_1`/`_2` disambiguating-suffix retry loop instead of overwriting. This change aligns it with the shared rule already used by the TRX, HTML and JUnit report extensions: **always overwrite (`FileMode.Create`) and emit a warning when the file already existed**, giving users a single, predictable rule.

- `CtrfReportEngine.FileWriter.cs`: replaced `WriteWithRetryAsync` (and the `SplitCtrfExtension` helper) with a small `WriteAsync` that mirrors `HtmlReportEngine`.
- `CtrfReportEngine.cs`: `GenerateReportCoreAsync` now calls `WriteAsync` and no longer threads the `fileNameExplicitlyProvided` flag.
- `CtrfReportEngineTests.cs`: default-name test now expects `FileMode.Create`; `GenerateReportAsync_AppendsDisambiguatingSuffix_When_DefaultFileExists` was replaced by `GenerateReportAsync_OverwritesAndWarns_When_DefaultFileExists`, and the IOException test was renamed to `..._When_WriteFails` — both mirroring the `HtmlReportEngineTests` equivalents. (27/27 passing.)

The `CtrfReportFileExistsAndWillBeOverwritten` resource already existed, so no resx/xlf changes were needed.

### 2. New public, injectable `IArtifactNamingService` in MTP core

Exposes file-name templating as a public, injectable core service so it can be consumed outside the existing linked-source (`[Embedded]` helper) pattern.

- `IArtifactNamingService` (public) — single member `ResolveFileName(string template)`; expands the standard placeholders (`{pname}`, `{pid}`, `{asm}`, `{tfm}`, `{arch}`, `{time}`) and sanitizes the leaf file name while preserving any directory portion.
- `ArtifactNamingService` (internal) — wraps the existing `ArtifactNamingHelper` templating with process/name/time resolution from `ITestApplicationModuleInfo`/`IEnvironment`/`IClock`.
- `ArtifactFileNameSanitizer` (internal) — a core-internal copy of the invalid-char/reserved-name sanitization from the extension-side `ReportFileNameSanitizer` (which has shipped InternalAPI entries and is intentionally left in place; the two are marked for future consolidation).
- Registered as a common service in `TestHostBuilder.CommonServices.cs` and resolvable via a new public `ServiceProviderExtensions.GetArtifactNamingService(...)` accessor.
- Public/internal API tracking files updated; new `ArtifactNamingServiceTests` (4/4 passing) cover placeholder expansion, sanitization, directory preservation, and unknown-placeholder passthrough.

## Verification

Built `Microsoft.Testing.Platform.UnitTests` and `Microsoft.Testing.Extensions.UnitTests` and ran the targeted filters:
- `ArtifactNamingService` — 4/4 passing
- `CtrfReportEngineTests` — 27/27 passing

## ⚠️ Heads-up: pre-existing, unrelated baseline break

There is a pre-existing repo-wide build break introduced by #9774 that is **not** addressed here and needs a separate fix: `IPC/Serializers/BaseSerializer.cs` gained two `protected static` methods (`ReadFields`, `WriteListPayload`) but no matching `InternalAPI.Unshipped.txt` entries were added, so analyzer **RS0051** fails as an error in every project that links that source (Microsoft.Testing.Platform and the HangDump / MSBuild / Retry / TrxReport extensions). A transient local workaround was applied only to verify this PR and then reverted, so it is intentionally not part of this change.

### DIFF
--- a/src/Platform/Microsoft.Testing.Extensions.CtrfReport/CtrfReportEngine.FileWriter.cs
+++ b/src/Platform/Microsoft.Testing.Extensions.CtrfReport/CtrfReportEngine.FileWriter.cs
@@ -7,72 +7,19 @@ namespace Microsoft.Testing.Extensions.CtrfReport;
 
 internal sealed partial class CtrfReportEngine
 {
-    private async Task<(string FileName, string? Warning)> WriteWithRetryAsync(string finalPath, byte[] bytes, bool fileNameExplicitlyProvided)
+    private async Task<(string FileName, string? Warning)> WriteAsync(string finalPath, byte[] bytes)
     {
-        // Explicit file names: use FileMode.Create (overwrite). Default-generated file
-        // names: use FileMode.CreateNew but retry with disambiguating suffixes when the
-        // file already exists, so concurrent runs (or two runs within the same second
-        // sharing the result directory) don't fail with IOException.
-        if (fileNameExplicitlyProvided)
-        {
-            bool willOverwrite = _fileSystem.ExistFile(finalPath);
-            await WriteBytesAsync(finalPath, FileMode.Create, bytes).ConfigureAwait(false);
-            return (
-                finalPath,
-                willOverwrite
-                    ? string.Format(CultureInfo.InvariantCulture, ExtensionResources.CtrfReportFileExistsAndWillBeOverwritten, finalPath)
-                    : null);
-        }
-
-        DateTimeOffset firstTry = _clock.UtcNow;
-        string directory = Path.GetDirectoryName(finalPath) ?? string.Empty;
-        string fileName = Path.GetFileName(finalPath);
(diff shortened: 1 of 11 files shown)


Evangelink:
Split into two independent PRs per maintainer request: #9782 (Align CTRF report writer overwrite behavior with TRX/HTML/JUnit) and #9783 (Add public injectable IArtifactNamingService to Microsoft.Testing.Platform). Closing this combined PR in favor of those two.

PR #10061, opened 2026-07-19
docs: Rename testCases to tests in RunTestsParams interface

richardwerkman:
When developing the MTP testrunner for Stryker.NET I noticed a mistake in the docs. The real param for test filtering is named `tests`. This can be seen [here](https://github.com/microsoft/testfx/blob/fd744761254d82f054ebba7f8479af01f3822c5a/samples/Playground/ServerMode/v1.0.0/RunRequest.cs#L9). I've verified and test filtering only works when using `tests`, not `testCases` like specified in the docs.


Evangelink:
Thanks for the contribution @richardwerkman

Evangelink (approved):

Issue #10016, opened 2026-07-16
Crash-resilient TRX: an always-loadable padded incremental writer


nohwnd:
## Summary

We write the `.trx` straight to disk as a file that is always correct, even when the test host crashes. The file starts life already in a **failed** state, with its variable regions pre-padded with spaces: enough to hold ~10k tests out of the box (~6.6 MB on disk, no discovery count needed). After every test we write the result into those regions and update the counters in place; the counters live in space-padded cells too, sized so we never reflow the whole document on an ordinary write.

Because XML ignores whitespace between elements, that padding is invisible to any reader. The file is a valid, parsable TRX at every point, with no repair tool. The big, unbounded part, test output, is appended at the end of the file, so it grows for free and never needs a pad.

We also reserve a few in-progress slots near the top of the results and write a test into one when it starts, blanking it when it finishes. So if the host dies inside a test, that slot survives and names the test that killed the run, something today's TRX can never show, because a test that crashes the process never produces a result row.

When a run outgrows its padding we atomically reflow into a fresh file and swap it in (rare by design; a 200 MB-or-smaller copy, done off the test's own thread). On a clean exit we do one last cosmetic reflow to strip the spaces, so the shipped file is byte-identical to today's output. Nothing a reader needs rides on that final pass: finalization only buys tidiness, never correctness.

## Background and Motivation

Today, if the test host crashes mid-run in the default in-process mode, the `.trx` is never written: the report is built once, in memory, at session end. The existing streaming/crash-recovery feature (#8263) only helps the out-of-process `--crashdump` case, and even then it writes results to a proprietary binary sidecar that a surviving controller has to translate back into a `.trx`. On an in-process crash it produces nothing (the code itself warns "if the test host crashes during this run, the TRX will not contain partial results"). Building the whole report in memory is also an out-of-memory risk on very large runs (the same class of problem as vstest#2255).

The goal: produce a complete, loadable `.trx` at every reasonable crash point, in both in-process (the default) and out-of-process (`--crashdump`) modes, with minimal per-test overhead. The on-disk file is always a valid TRX: no binary intermediate, no translation step, no surviving process required.

"Reasonable crash" means process death: `FailFast`, unhandled native crash/segfault, `Environment.Exit`, external kill. It explicitly excludes OS crash and power loss (that needs `fsync`, which we do not pay).

### Core invariant

1. **Always closed.** At every flush boundary the file on disk is well-formed XML: all tags closed, loadable without any repair tool.
[...]

PR #10106, opened 2026-07-21
Prototype crash-resilient TRX writes

nohwnd:
Related #10016

Wanted to implement both options and see what actually survives interrupted writes. Neither implementation is wired into production.

The padded file cannot provide the guarantee from the issue. Across 5,656 injected operation failures and byte cuts:

- 4,545 were malformed
- 719 were parseable but inconsistent
- 22 were repairable
- 370 were truthful

The append-only journal recovered a complete prefix in all 246 interruption cases, including restart and append after a torn final record. Publishing complete streamed snapshots through sibling-file replacement kept the destination old or new and truthful in all 3,519 observations.

For 10k synthetic results, journal plus snapshots did 10,004 flushes and wrote 18.3 MB. The padded writer did 89,959 flushes and wrote 24.6 MB. Replay keeps one record, no result collection, and no `XDocument`.

## Changes

- Add an isolated padded incremental writer with fixed pads, counters, running slots, reflow, and compaction.
- Add an append-only journal with torn-tail recovery and streamed snapshot publication.
- Add deterministic operation and partial-write injection, semantic TRX classification, concurrent-reader barriers, and real-file-system tests.
- Cover CRLF preservation, UTF-8-safe truncation, long metadata, large output, summary data, startup overwrite, replacement sharing modes, and 10k-result diagnostics.

## Verification

- 148 focused tests passed across `net462`, `net472`, `net8.0`, and `net9.0`.
- 3,432 extension unit tests passed, with 20 existing skips.
- 20 Release diagnostics passed.
- TRX acceptance passed: 40 MTP, 6 MSTest TRX, and 6 MSTest output tests.
- Debug and Release builds, full repository build, and package production passed.

This does not claim XSD, Visual Studio, Azure DevOps, third-party reader, network filesystem, antivirus, physical power-loss, or actual attachment-copying compatibility. Atomic overwrite intentionally fails closed on .NET Framework where the same replacement guarantee is not available.

LMK if this should stay as evidence only, or if I should follow with the production journal and snapshot path.

🤖


nohwnd:
Closing the investigation after recording the findings and recommendation in #10016. The prototype disproved the padded writer guarantee and supports journal plus atomic snapshots instead. This branch remains available as evidence, but the prototype is intentionally not wired into production and is not intended for merge.

🤖

PR #10297, opened 2026-07-28
Fix Content-Length byte/char mismatch in the server-mode TCP transport

azat-msft:
## The bug

`TcpMessageHandler` declares `Content-Length` in **UTF-8 bytes** on the write path, but the read path consumed that number as **decoded characters**.

**Write path** (correct, unchanged):

```csharp
int byteCount = Encoding.UTF8.GetByteCount(messageStr);
await _writer.WriteLineAsync($"Content-Length: {byteCount}");
...
await _writer.BaseStream.WriteAsync(rentedBytes.AsMemory(0, byteCount), cancellationToken);
```

**Read path before:**

```csharp
char[] commandCharsBuffer = ArrayPool<char>.Shared.Rent(commandSize);
Memory<char> memoryBuffer = new(commandCharsBuffer, 0, commandSize);
await _reader.ReadBlockAsync(memoryBuffer, cancellationToken);   // characters, not bytes
return _formatter.Deserialize<RpcMessage>(memoryBuffer);
```

`_reader` is a `StreamReader`, so `ReadBlockAsync` yields **decoded characters**. The `#else` (Jsonite, net462/netstandard2.0) branch had the identical defect with `char[] commandChars = new char[commandSize]`.

**Read path after** — consume exactly `Content-Length` bytes, then UTF-8-decode, symmetric with the write path:

```csharp
byte[] bodyBuffer = ArrayPool<byte>.Shared.Rent(commandSize);
if (!await ReadExactlyAsync(bodyBuffer, commandSize, cancellationToken)) { return null; }
int charCount = Encoding.UTF8.GetCharCount(bodyBuffer, 0, commandSize);
char[] charsBuffer = ArrayPool<char>.Shared.Rent(charCount);
Encoding.UTF8.GetChars(bodyBuffer, 0, commandSize, charsBuffer, 0);
return _formatter.Deserialize<RpcMessage>(new ReadOnlyMemory<char>(charsBuffer, 0, charCount));
```

For pure ASCII, bytes == characters, so nothing broke and the defect stayed latent. For **any** multi-byte UTF-8 content — a localized test name, an exception message, or captured standard output in German, Japanese, Czech, Cyrillic, or an emoji — the two counts disagree. The reader under-reads the frame, leaves the body's tail in the stream, and the framing **desynchronizes permanently**: the leftover bytes are parsed as the *next* frame's headers.

**The symptom only appears from the second frame onward.** The first frame still deserializes fine (its JSON is merely truncated at a point the parser may or may not reject); it is the following read that picks up mid-message garbage. Any regression test therefore has to send at least two messages.

Both compilation branches were affected — `#if NETCOREAPP` (System.Text.Json, net8+) and `#else` (Jsonite, net462/netstandard2.0) — and both are fixed.

## This affects the live server, not just the new client package

`src/Platform/Microsoft.Testing.Platform/ServerMode/JsonRpc/TcpMessageHandler.cs` is the shipping MTP server-mode transport. The source-only client package added in #10085 compiles the same file as source, which is how the bug surfaced, but the defect is in the server transport itself and has always been there. Anything speaking server mode against a test host that emits non-ASCII test names or output could hit it.

## The `StreamReader` buffering hazard

Reading the body as raw bytes from `_reader.BaseStream` while the headers still went through `StreamReader` would have been a *worse* bug: `StreamReader` fills its own internal buffer, so by the time the header line is returned it has typically already swallowed part of the body, and `BaseStream` would resume past it.

The fix therefore removes `StreamReader` from the read path entirely. Headers and body are both read through a single byte-level buffer owned by the handler (`_readBuffer` / `ReadHeaderLineAsync` / `ReadExactlyAsync` / `FillReadBufferAsync`), so header and body reads sit on the same side of the buffering boundary and nothing can be stranded. This mirrors what vstest's hand-written client (the now-deleted `MtpServerConnection`, replaced in microsoft/vstest#16300) already did correctly: read the header line at the byte level, then read exactly `count` bytes and `Encoding.UTF8.GetString` them.

One compatibility detail worth calling out: the old `StreamReader` had byte-order-mark detection enabled by default and silently swallowed a leading UTF-8 preamble. Some peers write their headers through a preamble-emitting encoder (`new StreamWriter(stream, Encoding.UTF8)` — the platform's own `ServerTests` do exactly this), so the new reader keeps that tolerance explicitly via `TrimPreamble`, and there is a test for it. Cancellation semantics and the `SocketError.ConnectionReset` graceful-disconnect handling are unchanged.

## Tests

The regression tests live in `test/UnitTests/Microsoft.Testing.Platform.UnitTests/ServerMode/TcpMessageHandlerTests.cs`, merged into the existing test class. They stand up a pair of `TcpMessageHandler`s joined over loopback TCP, so the bytes on the wire are exactly what a real server and client exchange. That project multi-targets net462 alongside net8.0/net9.0, so every test runs against **both** the `#if NETCOREAPP` (System.Text.Json) and `#else` (Jsonite) compilation branches.

Covering the byte/char defect itself:

- `ReadAsync_RawUtf8FramesFromPeer_DoesNotDesynchronizeSubsequentFrames` — two frames from a conformant peer that emits raw UTF-8, the first with multi-byte content. Fails before the fix on **every** TFM.
- `ReadAsync_BodyLargerThanReadBuffer_SpansMultipleRefillsWithoutDesynchronizing` — a body far larger than the read buffer, so it spans several refills; a second frame follows to prove the reader stopped on the exact byte boundary. Fails before the fix.
- `ReadAsync_HandlerWrittenFramesWithNonAscii_DoesNotDesynchronizeSubsequentFrames` — the same round trip through `TcpMessageHandler` on both ends, using real `client/log` and `telemetry/update` notifications. See the caveat below: this one only exercises the defect on net462.
- `ReadAsync_AsciiFrames_RoundTripInOrder` — guards the ASCII fast path against over-reading.

Covering the paths this change introduces:

- `ReadAsync_HeaderLineLongerThanInitialBuffer_GrowsAndParsesFrame` — a 5 KB header line, driving the `_headerLineBuffer` growth path through several doublings.
- `ReadAsync_PeerDisconnectsMidBody_ReturnsNull` — a peer that announces a body then hangs up mid-way, which must surface as a graceful disconnect rather than a hang or a throw. Fails before the fix.
- `ReadAsync_LeadingByteOrderMark_IsSkipped` — pins the BOM tolerance the old `StreamReader` provided implicitly.

### Coverage caveat worth knowing

`ReadAsync_HandlerWrittenFramesWithNonAscii_...` **cannot detect this bug on .NET.** `Utf8JsonWriter` uses `JavaScriptEncoder.Default`, which escapes every non-ASCII character to `\uXXXX`, so on net8.0/net9.0 the body reaches the wire as pure ASCII, byte count equals char count, and the pre-fix reader passes it unchanged. It only exercises the regression on net462, where Jsonite emits BMP characters unescaped. The docstring says so outright.

The cross-TFM coverage comes from the two tests that frame raw UTF-8 bytes themselves and so bypass formatter escaping entirely.

This is also **why the bug stayed latent**: MTP-to-MTP traffic is accidentally ASCII on the wire, so the defect only bites when the peer emits raw UTF-8 — which is exactly what an external client such as vstest's does.

### Failure output before the fix

net8.0:

```
failed ReadAsync_RawUtf8FramesFromPeer_DoesNotDesynchronizeSubsequentFrames (100ms)
  System.Text.Json.JsonReaderException: 'C' is invalid after a single JSON value.
  Expected end of data. LineNumber: 0 | BytePositionInLine: 90.
```

The `'C'` is the `C` of the *next* frame's `Content-Length` header being read as part of the previous body — the desynchronization made visible.

net462, where the non-ASCII tests hang until the read times out:

```
failed ReadAsync_RawUtf8FramesFromPeer_DoesNotDesynchronizeSubsequentFrames (30s 080ms)
  Timed out reading a frame; the transport is most likely desynchronized.
failed ReadAsync_HandlerWrittenFramesWithNonAscii_DoesNotDesynchronizeSubsequentFrames (30s 080ms)
  Timed out reading a frame; the transport is most likely desynchronized.
```

Verified by swapping `TcpMessageHandler.cs` back to its parent revision, rebuilding, and re-running — so these are genuine failing-before/passing-after regression tests, not tests fitted to the new implementation.

### Suite results (before → after)

Run locally in Release on Windows.

| Suite | Before | After |
| --- | --- | --- |
| `Microsoft.Testing.Platform.UnitTests` (net8.0) | 1440/1440 | **1447/1447** |
| `Microsoft.Testing.Platform.UnitTests` (net9.0) | 1440/1440 | **1447/1447** |
| `Microsoft.Testing.Platform.UnitTests` (net462) | 1416/1416 | **1423/1423** |
| ↳ `ServerTests` (live server over TCP) | 5/5 | 5/5 |
| ↳ serialization tests | 78/78 | 78/78 |
| `Microsoft.Testing.Platform.ServerClient.UnitTests` (net8.0) | 24/24 | 24/24 |
| `Microsoft.Testing.Platform.ServerClient.UnitTests` (net462) | 24/24 | 24/24 |
| `MtpServerClientSourcePackage*` (contract + hostile consumer) | 9/9 | 9/9 |
| `MtpServerClientAcceptanceTests` (real generated MTP app) | 3/3 | 3/3 |
| MTP acceptance `Server*` | 14/14 | 14/14 |

The platform suites are up 7 (the 4 transport tests plus 3 new buffer-path tests), and `ServerClient.UnitTests` is unchanged at 24.

The `ServerTests` row is the interesting one: it is the platform's own server-mode suite driving a real `TestApplication` over TCP, and it is what caught the BOM regression in an intermediate version of this change.

The fix is additionally validated outside this repo: a locally-built pack of the source package was consumed by the vstest retarget work, where CrossPlatEngine builds clean on net462/netstandard2.0/net8.0 and the MTP suites pass 140/140 and 16/16 — including the raw-UTF-8 peer traffic that exposes this bug.

## Relationship to #10085

The `src/` change is **independent of #10085** and applies to `main` as-is. The regression tests live in `Microsoft.Testing.Platform.UnitTests`, which exists on `main` and already multi-targets net462, so nothing about this fix requires the source-package work.

Verified rather than assumed — rebasing all three commits onto current `main`:

- rebases cleanly, zero conflicts;
- `TcpMessageHandlerTests.cs` is byte-identical on `main` and on this PR's base;
- every symbol the new tests use exists on `main` in the same files;
- the only difference in `TcpMessageHandler.cs` between `main` and this base is an unrelated 5-line `Substring`-vs-range-operator change from #10085, which none of these commits touch.

The base is left as `nohwnd-mtp-client-source-package` so it reads as an increment on that work, but it can be retargeted to `main` at any time if the shipping-server fix should land independently.

## Provenance

This was raised by an automated reviewer on #10085 and acknowledged there by the author as a real latent bug, to be fixed in a dedicated platform PR rather than inside the packaging change. This is that PR, stacked on `nohwnd-mtp-client-source-package` so it reads as an increment on that work.


src/Platform/Microsoft.Testing.Platform/ServerMode/JsonRpc/TcpMessageHandler.cs
```diff
@@ -156,6 +191,106 @@ private async Task<int> ReadHeadersAsync(CancellationToken cancellationToken)
...
+    /// <summary>
+    /// Reads a single CRLF-terminated header line from the shared byte buffer. Headers are ASCII by protocol,
+    /// but they are decoded as UTF-8 (a superset) so a non-conformant peer cannot corrupt the framing.
+    /// Returns <see langword="null"/> at end of stream.
+    /// </summary>
+    private async Task<string?> ReadHeaderLineAsync(CancellationToken cancellationToken)
+    {
+        // Headers are short; a small growable buffer avoids allocating per byte. Plain arrays rather than
+        // ArrayPool so this compiles on netstandard2.0/net462 (System.Buffers is out of framework there).
+        byte[] lineBuffer = new byte[HeaderLineInitialCapacity];
```
Evangelink: Nit, non-blocking: this allocates a fresh 256-byte array **per header line**, so roughly 3 arrays (~768 B) per frame on top of the 3 strings. The previous `StreamReader.ReadLineAsync` reused its internal buffer and only allocated the resulting string, so this is a small new steady-state allocation on the hottest read path — server mode emits a `testing/testUpdates/tests` notification per test.

Since the reader is already single-threaded by construction (`_readBufferOffset`/`_readBufferCount` are instance fields with the same assumption, and `MtpJsonRpcConnection` drives exactly one read loop), promoting `lineBuffer` to an instance field seeded at `HeaderLineInitialCapacity` and grown in place is exactly as safe and removes the per-line allocation entirely.

src/Platform/Microsoft.Testing.Platform/ServerMode/JsonRpc/TcpMessageHandler.cs
```diff
@@ -66,22 +84,45 @@ public TcpMessageHandler(
                     return null;
                 }
 
+                // Content-Length counts UTF-8 bytes, so consume exactly that many bytes and decode
+                // afterwards. Reading characters here would under-read every multi-byte frame.
 #if NETCOREAPP
-                char[] commandCharsBuffer = ArrayPool<char>.Shared.Rent(commandSize);
+                byte[] bodyBuffer = ArrayPool<byte>.Shared.Rent(commandSize);
```
Evangelink, line 104: Observation only, pre-existing and **not** made worse here — flagging it so it's a conscious decision rather than an oversight.

`commandSize` comes straight off the wire and is still trusted unbounded: a peer declaring `Content-Length: 2147483647` drives `Rent(commandSize)` here and `new byte[commandSize]` on the `#else` branch. Note this change actually *halves* the worst case, since the old code rented that many `char`s.

If you ever want a sanity cap, this is the natural place for it. Happy to leave it out of a scoped bug fix.

test/UnitTests/Microsoft.Testing.Platform.ServerClient.UnitTests/TcpMessageHandlerFramingTests.cs
```diff
@@ -0,0 +1,202 @@
...
+        Assert.AreEqual(NonAsciiText, ((IDictionary<string, object?>)first.Params!)["text"]?.ToString());
+        Assert.AreEqual("testing/second", second.Method);
+    }
+
+    /// <summary>
+    /// Guards the ASCII fast path: a plain frame followed by another must still round-trip, so the byte-exact
+    /// read does not over-consume when bytes and characters happen to coincide.
+    /// </summary>
+    [TestMethod]
+    public async Task ReadAsync_AsciiFrames_RoundTripInOrder()
```
Evangelink: Coverage gap worth closing, non-blocking: the four tests are all well-chosen for the byte/char defect itself, but every frame here is comfortably under `ReadBufferSize` (4096) and every header line under `HeaderLineInitialCapacity` (256). That means the two *new* loops added by this PR only ever execute a single iteration:

- the `while (written < count)` refill loop in `ReadExactlyAsync` — never exercised across a buffer boundary, even though a test-node update carrying a stack trace or captured stdout routinely exceeds 4 KB in production;
- the `lineBuffer` doubling path in `ReadHeaderLineAsync` — never exercised at all.

Also untested: `ReadExactlyAsync` returning `false` (peer disconnects mid-body), which must surface as `ReadAsync` returning `null` rather than throwing or hanging.

I wrote throwaway probes for all three against this branch and they pass, so there is no defect here — but the code paths are load-bearing and currently unguarded against future regression. Three small additions would cover them: a body of ~20 KB, a header line of ~5 KB, and a frame truncated mid-body followed by a close.

Evangelink (approved):
## Summary

Reviewed in depth and **validated locally** — this is a real, correctly-diagnosed bug and the fix is sound. Approving.

### What I verified (not just read)

I checked out the head commit into a worktree, built both TFM legs, and ran the suites:

| Check | Result |
| --- | --- |
| `Microsoft.Testing.Platform.ServerClient.UnitTests` (net8.0) | 28/28 pass |
| `Microsoft.Testing.Platform.ServerClient.UnitTests` (net462) | 28/28 pass |
| `Microsoft.Testing.Platform.UnitTests` (net8.0, incl. `ServerTests`) | 1440/1440 pass |
| `Microsoft.Testing.Platform.UnitTests` (net462) | 1416/1416 pass |
| Build both legs | clean, 0 warnings |

**The regression tests genuinely fail without the fix.** I reverted only `TcpMessageHandler.cs` to its parent revision, rebuilt, and re-ran — `ReadAsync_RawUtf8FramesFromPeer_DoesNotDesynchronizeSubsequentFrames` and `ReadAsync_LeadingByteOrderMark_IsSkipped` both fail on net8.0 with exactly the `'C' is invalid after a single JSON value` reader exception the description quotes (the `C` of the *next* frame's `Content-Length`). That is a properly failing-before/passing-after regression test, not a test written to match the new implementation.

**The bug is confirmed present on `main`,** not just on the stacked base branch: `main`'s `WriteRequestAsync` declares `Encoding.UTF8.GetByteCount(messageStr)` while `ReadAsync` consumes that count through `StreamReader.ReadBlockAsync` (characters) on both compilation branches. So this really is a shipping server-transport defect, and the framing desynchronizes permanently once any multi-byte body goes over the wire.

### Correctness review of the new read path

The reasoning about `StreamReader` is right and worth keeping in the comment: reading headers through a `StreamReader` and the body from `BaseStream` would have been strictly worse, because the reader's internal buffer would already have swallowed part of the body. Collapsing both onto one byte-level buffer is the correct shape.

Specific points I checked and found sound:

- `FillReadBufferAsync` is only ever called when `_readBufferOffset == _readBufferCount`, so the unconditional `_readBufferOffset = 0` cannot discard unconsumed bytes.
- Offset/count are assigned only *after* the `await`, so a throwing/cancelled read leaves the buffer state intact.
- `GetCharCount`/`GetChars` use the same `Encoding.UTF8` replacement fallback, so the counts cannot disagree on malformed input.
- Both pooled buffers are returned on every exit path, including the mid-frame-disconnect `return null`.
- `Dispose()` swapping `_reader.Dispose()` for `_readStream.Dispose()` is behavior-preserving: `StreamReader` owned the stream (`leaveOpen: false`), and both real call sites (`MessageHandlerFactory` and `MtpServerProcess`) pass the *same* `NetworkStream` for both directions.
- Cancellation semantics are genuinely unchanged: the removed `NET7_0_OR_GREATER`/`NET6_0_OR_GREATER` ladder existed only for `StreamReader.ReadLineAsync`, which forwards the token to the same `Stream.ReadAsync` the new code calls directly.

To back the parts the PR's own tests do not reach, I wrote six throwaway probes against the fixed handler — a 20 KB body spanning many `FillReadBufferAsync` calls, a 5 KB header line forcing several `lineBuffer` doublings, a frame dribbled one byte at a time so header and body straddle refills at arbitrary offsets, bare-LF headers, and a peer that disconnects mid-body. All six pass: the buffer-boundary logic is correct. (Deleted afterwards — see the coverage note inline.)

### Findings

Nothing blocking. Three non-blocking notes inline: a per-frame allocation on the hot read path, a test-coverage gap for the multi-iteration buffer paths, and a pre-existing unbounded `Content-Length` that this change makes strictly better rather than worse.

### One process question

The `src/` half of this fix is independent of #10085 and applies to `main` as-is, but the PR is stacked on `nohwnd-mtp-client-source-package` because the regression test lives in the new `ServerClient.UnitTests` project. That couples a shipping-server bug fix to the packaging work. Worth deciding explicitly whether you want the transport fix to be able to merge to `main` on its own (e.g. with the test placed in `Microsoft.Testing.Platform.UnitTests` instead) — entirely your call, and not a reason to hold this up if #10085 is landing shortly.

src/Platform/Microsoft.Testing.Platform/ServerMode/JsonRpc/TcpMessageHandler.cs
azat-msft: Done in `a12f08a08`. `lineBuffer` is now the instance field `_headerLineBuffer`, seeded at `HeaderLineInitialCapacity` and grown in place, so the per-line allocation is gone entirely.

Your safety argument holds and I made it explicit in the comment: it is safe for exactly the same reason `_readBufferOffset`/`_readBufferCount` already are, namely that reads are single-threaded by construction and driven by one read loop.

One consequence worth naming rather than leaving implicit: the buffer now grows to the longest header line seen on the connection and stays there. That is bounded by the same trust boundary as `Content-Length` (a loopback channel to a test host this process launched), and it is strictly better than reallocating per line, but I noted it in the comment so it is a recorded tradeoff rather than an accident. The new `ReadAsync_HeaderLineLongerThanInitialBuffer_GrowsAndParsesFrame` test exercises the growth path through several doublings.

azat-msft: Done in `a12f08a08`. `lineBuffer` is now the instance field `_headerLineBuffer`, seeded at `HeaderLineInitialCapacity` and grown in place, so the per-line allocation is gone entirely.

Your safety argument holds and I made it explicit in the comment: it is safe for exactly the same reason `_readBufferOffset`/`_readBufferCount` already are, namely that reads are single-threaded by construction and driven by one read loop.

One consequence worth naming rather than leaving implicit: the buffer now grows to the longest header line seen on the connection and stays there. That is bounded by the same trust boundary as `Content-Length` (a loopback channel to a test host this process launched), and it is strictly better than reallocating per line, but I noted it in the comment so it is a recorded tradeoff rather than an accident. The new `ReadAsync_HeaderLineLongerThanInitialBuffer_GrowsAndParsesFrame` test exercises the growth path through several doublings.

src/Platform/Microsoft.Testing.Platform/ServerMode/JsonRpc/TcpMessageHandler.cs
azat-msft: Agreed, and thanks for flagging it as a decision rather than silently leaving it. I have kept the behaviour unchanged and recorded the reasoning in a comment in `a12f08a08`, so the next reader does not have to re-derive it:

Server mode is a loopback channel to a test host this process launched, not an untrusted network peer, and legitimate bodies are unbounded in principle — a test node update can carry an arbitrarily large stack trace or captured stdout. Any cap would therefore be a guess that risks rejecting valid traffic, which is a worse failure mode than the allocation it avoids. Your point that this change *halves* the worst case is now in the comment too, since it is the part that makes leaving it alone defensible in a fix whose whole purpose is the byte/char distinction.

If a cap is ever wanted it belongs here, and it should come with a protocol-level error response rather than an allocation failure — out of scope for this PR.

test/UnitTests/Microsoft.Testing.Platform.ServerClient.UnitTests/TcpMessageHandlerFramingTests.cs
azat-msft: Good catch, and you were right that these are load-bearing paths. All three are covered in `a12f08a08`:

| Test | Path covered | Fails pre-fix? |
| --- | --- | --- |
| `ReadAsync_BodyLargerThanReadBuffer_SpansMultipleRefillsWithoutDesynchronizing` | `ReadExactlyAsync` refill loop across buffer boundaries (~16 KB body vs 4 KB buffer) | Yes |
| `ReadAsync_HeaderLineLongerThanInitialBuffer_GrowsAndParsesFrame` | `_headerLineBuffer` doubling (5 KB header vs 256 B seed) | No — guards a path this PR introduced |
| `ReadAsync_PeerDisconnectsMidBody_ReturnsNull` | `ReadExactlyAsync` returning `false` | Yes |

Two deliberate choices in how they assert:

The large-body test carries its non-ASCII payload in the **method name** and asserts the whole string round-trips, so it compares every byte of the oversized frame rather than merely checking that a frame arrived. It also sends a second frame afterwards, so it proves the reader stopped on the exact byte boundary rather than just that a big read completed.

The mid-body-disconnect test asserts `ReadAsync` returns `null` specifically — not that it throws, and not merely that it completes — because the contract that matters is that a truncated frame is indistinguishable from a graceful disconnect to the read loop.

I confirmed the first and third genuinely fail against the pre-fix transport rather than being written to match the new implementation: I swapped `TcpMessageHandler.cs` back to its parent revision, rebuilt, and re-ran. On net8.0 the large-body test fails and the mid-body test fails; on net462 both fail. Restoring the fix returns all 8 to green on net8.0, net9.0 and net462.

azat-msft:
Thanks for the depth here — reverting `TcpMessageHandler.cs` to its parent to confirm the tests genuinely fail, and writing throwaway probes for the buffer-boundary paths, is exactly the review this deserved given it is live server transport code.

All three inline findings are addressed in `a12f08a08`, and I answered your process question below with evidence rather than an opinion.

## Changes since your review

**Per-frame allocation** — `lineBuffer` is now the instance field `_headerLineBuffer`, grown in place. I also recorded the one consequence worth naming: it grows to the longest header line seen on the connection and stays there, bounded by the same trust boundary as `Content-Length`.

**Unbounded `Content-Length`** — behaviour unchanged, reasoning now recorded in a comment so it reads as a decision. Your observation that this change halves the previous worst case is part of that rationale.

**Coverage gaps** — three tests added, covering the `ReadExactlyAsync` refill loop across buffer boundaries, the `_headerLineBuffer` doubling path, and `ReadExactlyAsync` returning `false` on a mid-body disconnect. I verified the first and third fail against the pre-fix transport on both net8.0 and net462, so they are real regression guards rather than tests fitted to the implementation.

## The process question — the fix is now genuinely independent of #10085

You were right that this was worth deciding explicitly, and acting on it turned out to be cheap. **The tests have moved from `ServerClient.UnitTests` into `Microsoft.Testing.Platform.UnitTests`**, merged into the existing `TcpMessageHandlerTests`. That project already multi-targets net462 alongside net8.0/net9.0, so both the `#if NETCOREAPP` (System.Text.Json) and `#else` (Jsonite) compilation branches are still covered — the reason the tests were in the new project in the first place was convenience, not necessity.

I verified the decoupling rather than assuming it. Rebasing all three commits onto current `main` (`e08d1635b`):

- rebases **cleanly, zero conflicts**;
- `TcpMessageHandlerTests.cs` is **byte-identical** on `main` and on this PR's base, so the tests land in the same place either way;
- every symbol the new tests use (`JsonRpcMethods.ClientLog`/`TelemetryUpdate`, `LogEventArgs`, `ServerLogMessage`, `TelemetryEventArgs`, `FormatterUtilities`, `TcpMessageHandler`) exists on `main` in the same files;
- the **only** difference in `TcpMessageHandler.cs` between `main` and this PR's base is an unrelated 5-line `Substring`-vs-range-operator change from #10085, which none of my commits touch.

So the transport fix can merge to `main` on its own whenever you want it to. **I have deliberately left the base as `nohwnd-mtp-client-source-package`** rather than retargeting unilaterally — you have already approved this shape, and if #10085 is landing shortly the stacking is harmless. But it is now a one-command retarget if you would rather the shipping-server fix went in independently. Entirely your call; say the word and I will rebase.

One thing worth knowing either way: the fix is validated end to end outside this repo. A locally-built pack of the source package was consumed by the vstest retarget work, where CrossPlatEngine builds clean on net462/netstandard2.0/net8.0 and the MTP suites pass 140/140 and 16/16 — including the raw-UTF-8 peer traffic that is precisely what exposes this bug.

## Test results after the changes

| Suite | Before this PR | After |
| --- | --- | --- |
| `Microsoft.Testing.Platform.UnitTests` net8.0 | 1440/1440 | **1447/1447** |
| `Microsoft.Testing.Platform.UnitTests` net9.0 | 1440/1440 | **1447/1447** |
| `Microsoft.Testing.Platform.UnitTests` net462 | 1416/1416 | **1423/1423** |
| `ServerClient.UnitTests` net8.0 | 24/24 | 24/24 |
| `ServerClient.UnitTests` net462 | 24/24 | 24/24 |
| `MtpServerClientSourcePackage*` (contract + hostile consumer) | 9/9 | 9/9 |
| `MtpServerClientAcceptanceTests` | 3/3 | 3/3 |
| MTP acceptance `Server*` | 14/14 | 14/14 |

The platform suites are up 7 (4 tests moved in from `ServerClient.UnitTests`, 3 new), and `ServerClient.UnitTests` is back to its original 24 — the arithmetic confirms nothing was lost in the move.

## One coverage caveat I made explicit

While re-reviewing the committed change I confirmed something the docstring had only hinted at, and stated it plainly in `58bba249f`: `ReadAsync_HandlerWrittenFramesWithNonAscii_...` **cannot detect this bug on .NET**. `Utf8JsonWriter` uses `JavaScriptEncoder.Default`, which escapes every non-ASCII character to `\uXXXX`, so on net8.0/net9.0 the body reaches the wire as pure ASCII, byte count equals char count, and the pre-fix reader would have passed it unchanged. It only exercises the regression on net462, where Jsonite emits BMP characters unescaped.

The cross-TFM coverage comes from `ReadAsync_RawUtf8FramesFromPeer_...` and `ReadAsync_BodyLargerThanReadBuffer_...`, which frame raw UTF-8 bytes themselves and so bypass formatter escaping on every TFM. This is also the reason the bug stayed latent for so long: MTP-to-MTP traffic is accidentally ASCII on the wire, and it only bites when the peer emits raw UTF-8 — which is exactly what an external client such as vstest's does. The docstring now says all of this outright so nobody over-trusts that test as cross-TFM coverage.

Evangelink (commented):
## New review pass

Re-reviewed after `a12f08a0` / `58bba249`. **The diagnosis and the fix are correct**, and I verified it rather than reading it. Nothing below blocks; the first two are the ones I'd actually act on.

### What I ran

Built `Microsoft.Testing.Platform.UnitTests` (net8.0/net9.0/net462) from the head commit:

| Check | Result |
| --- | --- |
| `TcpMessageHandlerTests` — net8.0 | 8/8 pass |
| `TcpMessageHandlerTests` — net462 | 8/8 pass |
| Full `Microsoft.Testing.Platform.UnitTests` — net8.0 | 1932/1932 pass |
| Full `Microsoft.Testing.Platform.UnitTests` — net462 | 1891/1891 pass |

Then I reverted **only** `TcpMessageHandler.cs` to `dbb946d4~1`, rebuilt, and re-ran, to confirm the tests are genuine failing-before regression tests rather than fitted to the new implementation:

```
failed ReadAsync_RawUtf8FramesFromPeer_DoesNotDesynchronizeSubsequentFrames
  System.Text.Json.JsonReaderException: 'C' is invalid after a single JSON value.
failed ReadAsync_LeadingByteOrderMark_IsSkipped
  System.Text.Json.JsonReaderException: 'C' is invalid after a single JSON value.
failed ReadAsync_BodyLargerThanReadBuffer_SpansMultipleRefillsWithoutDesynchronizing
  Timed out reading a frame; the transport is most likely desynchronized.
failed ReadAsync_PeerDisconnectsMidBody_ReturnsNull
  Timed out reading a frame; the transport is most likely desynchronized.

total: 8, failed: 4, succeeded: 4
```

4 of the 8 fail pre-fix on net8.0 alone. Claim confirmed.

### Reader correctness

Points I specifically convinced myself of, since this is framing code where a subtle error desynchronizes a live connection:

- Headers and body share **one** `_readBuffer` with one `_readBufferOffset`/`_readBufferCount` pair, and both `ReadHeaderLineAsync` and `ReadExactlyAsync` drain it before refilling. That's the invariant the whole fix rests on and it holds — no bytes can be stranded on the wrong side of a buffering boundary, which is exactly the trap that made the naive "headers via `StreamReader`, body via `BaseStream`" fix wrong.
- Header lines are accumulated as **bytes** and decoded once at the terminator, so a multi-byte character — or the BOM itself — split across a refill is handled correctly. Easy to get wrong with a per-chunk decode; this doesn't.
- `TcpMessageHandler` is the only `IMessageHandler` implementation in the tree, so there's no sibling transport carrying the same defect.
- The token reaching `ReadAsync` is `_testApplicationCancellationTokenSource.CancellationToken` (`PassiveNode`), i.e. connection-lifetime. So the net462 `WithCancellationAsync` pattern abandoning a read that still owns `_readBuffer` is a teardown-only concern, not a mid-connection corruption risk. Worth having checked given the buffer is now shared instance state rather than a per-call local.
- `Dispose` swapping `_reader.Dispose()` for `_readStream.Dispose()` is equivalent — `StreamReader.Dispose` disposed the underlying stream anyway.

---

### 1. The .NET path now transcodes bytes → chars → bytes

`Json.Deserialize<T>(ReadOnlyMemory<char> utf8Json)` calls `JsonDocument.Parse(ReadOnlyMemory<char>)`, which immediately transcodes back to UTF-8 into its own pooled array:

```csharp
// System.Text.Json
public static JsonDocument Parse(ReadOnlyMemory<char> json, JsonDocumentOptions options = default)
{
    ReadOnlySpan<char> jsonChars = json.Span;
    int expectedByteCount = JsonReaderHelper.GetUtf8ByteCount(jsonChars);
    byte[] utf8Bytes = ArrayPool<byte>.Shared.Rent(expectedByteCount);
    ...
```

So on `#if NETCOREAPP` the new code does `GetCharCount` + `Rent<char>` + `GetChars`, and STJ then undoes all three — while the raw UTF-8 bytes we started from were already sitting in `bodyBuffer`. `JsonDocument.Parse(ReadOnlyMemory<byte> utf8Json)` is the overload that wants exactly what we have.

An `IMessageFormatter.Deserialize<T>(ReadOnlyMemory<byte>)` overload (NETCOREAPP only — net462/Jsonite keeps `Encoding.UTF8.GetString`) removes one transcode and one pooled rent per frame on the hottest read path, which is the same path this PR is otherwise careful to keep allocation-free.

Worth doing partly for a non-performance reason: `Deserialize<T>(ReadOnlyMemory<char> **utf8Json**)` and `MessageFormatter.Deserialize<T>(ReadOnlyMemory<char> **serializedUtf8Content**)` are a `char`-typed parameter with a `utf8`-shaped name. That naming is the byte/char ambiguity this PR exists to fix, still present one layer down. Fine as a follow-up, but it belongs on the record next to this fix rather than rediscovered later.

### 2. Retarget to `main`

Base is `nohwnd-mtp-client-source-package` (#10085), still open. This fixes the **shipping** server-mode transport, and the description already establishes it rebases onto `main` cleanly with the tests landing in a project that exists there. I'd retarget now rather than leave a live-transport fix gated on an unrelated packaging PR.

### 3. Negative `Content-Length` crashes the read loop (verified)

The new comment says `commandSize` is deliberately trusted *unbounded*, and I agree with that reasoning for large values. But negative values aren't just unbounded — they throw. I probed it:

```
Content-Length: -5

PROBE-RESULT: threw System.ArgumentOutOfRangeException:
  minimumLength ('-5') must be a non-negative value. (Parameter 'minimumLength')
```

`int.TryParse` happily yields `-5`, the `is -1` guard doesn't catch it, and `ArrayPool<byte>.Shared.Rent(-5)` throws past the `SocketException`/`IOException` filter — so a malformed header takes down the read loop with an unrelated exception type instead of the graceful `null` disconnect every other malformed-frame path produces. (net462 gets `OverflowException` from `new byte[-5]`.)

Pre-existing, and I wouldn't ask for it normally. But the PR now carries a comment asserting this input was consciously reasoned about, and `if (commandSize < 0) { return null; }` is one line that makes the comment true.

### 4. The trust-boundary argument is weaker for `_headerLineBuffer` than for the body

The comment justifies unbounded header-line growth by pointing at `Content-Length`'s trust boundary. The two aren't equivalent: bodies are legitimately unbounded (a stack trace, captured stdout), which is what makes refusing to cap them correct. Header lines have no legitimate large case, and a peer that simply never sends `\n` doubles the buffer forever. If a cap is ever wanted, headers are both cheaper and more defensible than the body — worth saying so rather than grouping them.

### 5. Undocumented line-terminator delta

`StreamReader.ReadLineAsync` treated a bare `\r` as a terminator; `ReadHeaderLineAsync` breaks only on `\n`. Protocol-irrelevant, and arguably stricter-is-better — but the comment reads as if it enumerates the cases:

```csharp
// Tolerate both CRLF and a bare LF.
```

Adding "a bare CR is not a terminator, unlike StreamReader" closes it.

### Test nits

- `ConnectedHandlers.CreateAsync` leaks `listener` and `clientSocket` if `ConnectAsync`/`AcceptTcpClientAsync` throws.
- `WriteRawFrame` uses blocking `stream.Write`, and every frame is written before any read starts. `ReadAsync_BodyLargerThanReadBuffer_...` pushes ~27 KB that way and relies on socket buffers absorbing it. Fine today, but the margin is invisible in the test, so growing the payload later turns a clear assertion failure into a CI hang. Starting the read first — or a comment naming the constraint — makes the dependency explicit.
- `ReadWithTimeoutAsync` leaves an uncancelled 30 s `Task.Delay` per call.
- `ReadAsync_LeadingByteOrderMark_IsSkipped` carries `NonAsciiMethod`, so pre-fix it fails for the byte/char reason rather than isolating BOM handling. I checked it *does* still catch a BOM-only regression (the `\uFEFF` prefix makes the first header unparseable → `contentSize` stays -1 → `ReadAsync` returns null → the cast yields null → NRE on `.Method`), so the coverage is real. The failure mode just won't point at the BOM.

---

Good change — the byte-level rewrite is the right shape rather than a patch over `ReadBlockAsync`, and the pre-fix failure evidence holds up under independent reproduction. My only real asks are the retarget (#2) and, if you want it in scope, the one-line negative guard (#3).

PR #10310, opened 2026-07-29
Quote const in analyzer localization lock metadata

nohwnd:
Quotes `'const'` in the resource lock comment and regenerates the analyzer XLF files.

🤖


Evangelink (approved):

Evangelink:
Thanks @nohwnd

Issue #10258, opened 2026-07-27
Local `gh aw compile` silently corrupts two pinned actions in generated `.lock.yml` files


Evangelink:
Compiling **any** agentic workflow in this repo with a locally-installed `gh aw` extension (v0.83.1) rewrites two security-relevant action pins away from the repo-aligned values in every lock file it touches. Because the corruption lands in generated files that reviewers rarely read line-by-line, it can ride along unnoticed in an unrelated PR.

Surfaced while working on #10252, but it is **not specific to that change** — it affects every agentic workflow and every contributor compiling locally.

## The two corruptions

**1. `actions/checkout` downgraded v7.0.1 → v7.0.0**

| | SHA | Version | Occurrences in committed locks |
|---|---|---|---|
| Aligned (expected) | `3d3c42e5aac5ba805825da76410c181273ba90b1` | v7.0.1 | 168 |
| Emitted by local CLI | `9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0` | v7.0.0 | 0 |

The repo deliberately aligned every checkout pin to v7.0.1 in #10174. A local compile partially undoes that.

**2. `github/gh-aw-actions/setup` un-pinned from an immutable SHA to a mutable tag**

```diff
- uses: github/gh-aw-actions/setup@8bdba8075360648fe6802302a5b4e016361dc6ac # v0.83.1
+ uses: github/gh-aw-actions/setup@v0.83.1
```

This one is the more serious of the two: a version delta still resolves to a fixed commit, whereas a mutable tag can be repointed upstream at any time. Observed when regenerating `.github/workflows/agentic_commands.yml`.

## Mechanism

`.github/aw/actions-lock.json` keys its checkout override **by version string** and holds only an `actions/checkout@v7.0.1` entry. For *framework-injected* steps (patch-context, threat-detection, create-PR) the local CLI build requests checkout at **v7.0.0**, so the lookup key-misses, no override applies, and the CLI's stale built-in SHA wins.

Note that `actions-lock.json` cannot be edited to fix this — the injected version label is baked into the binary, so adding a `v7.0.0` entry would only pin the downgrade rather than prevent it.

## Why the `compiler_version` header does not catch it

[...]

PR #10364, opened 2026-07-31
Gate the build-failure binlog download on the commenter's write access

YuliiaKovalova:
## Problem

The slash command's `fetch-binlog` job has **no `if:` gate at all**.

gh-aw's role check (`roles: [admin, maintainer, write]`) is compiled into the
`pre_activation` job, and that job is generated with `needs: fetch-binlog`
(`build-failure-analysis-command.lock.yml`). So authorization runs **after**
the binlogs have already been downloaded — ~220 MB compressed / 1.2 GB
extracted for a typical build. `route_slash_command.cjs` in the central
dispatcher performs no permission check either, so **any** commenter on any PR
could trigger the full download.

Inverting the dependency would create a cycle, so the check is repeated as a
pre-gate inside `fetch-binlog`.

Because the command uses `strategy: centralized` it is started via
`workflow_dispatch` and has no `github.event.comment` payload, so the actor
is read from `aw_context.actor` — the same place `item_number` already
comes from. `role_name` and `permission` are both consulted: `role_name`
keeps `maintain`/`triage` distinct, while `permission` still reports push
access for custom org roles. Any API failure returns a document with neither
field and lands in the deny branch.

## Also backported from the dotnet/sdk port (dotnet/sdk#55539)

* **Missing-leg detection.** Failed legs that published no logs artifact are
  passed to the agent as `GH_AW_MISSING_LEGS` so it cannot conclude
  `no build failure` from the legs that happened to upload.
  Matching is **token-based, not prefix-based**: timeline job names and
  artifact names are not spelled alike — `Linux Debug` publishes
  `Logs_Build_Linux_Debug` but `Windows Debug` publishes
  `Logs_Build_Windows_NT_Debug`. A prefix rule would report a missing leg on
  every Windows failure.
* **Download bounds.** At most 40 artifacts and 3 GB compressed in total. The
  existing caps were per-artifact plus a cumulative *uncompressed* budget, so
  many mid-sized archives could still be pulled over the network first.
* **`unzip` exit 11** (`no files matched`) is no longer reported as an
  extraction failure — the leg published logs, they just hold no binlog.
* An artifact with no download URL now warns instead of being silently skipped.

## Validation — all in real runs, nothing assumed

Permission gate, executed in Actions on this branch:

| Path | Result |
|---|---|
| write access | `'YuliiaKovalova' has 'admin' access …; proceeding.` — download ran |
| no write access | `'octocat' does not have write access … (resolved role 'read')` — `Download binlogs => skipped` |

Full chain against **real failed build 1534452** (PR #10358), run end-to-end in
Actions: `Extracted 6 binlog(s) from 6/6 legs`, **no** false missing-leg
warning for `Windows Debug`, `GH_AW_MISSING_LEGS` exported empty, and the
agent read all 6 binlogs via `binlog-mcp` and correctly called `noop`
(*"All 6 build legs … compiled cleanly — non-build failure"*), which is right:
that build failed in integration tests, not compilation.

The new bounds were also exercised against the same real build: forcing the
compressed budget to 100 MB stops at leg 4 and fails closed; forcing the
artifact cap to 2 refuses the build; a build with no `Logs_Build_*` artifacts
emits `binlog-found=false`.

## Note on the `actions/checkout` pin

The regenerated locks keep `actions/checkout` on the **v7.0.1** SHA recorded
in `.github/aw/actions-lock.json`. Compiling on the repo's pinned v0.83.1
toolchain rewrites it to v7.0.0 — that is #10258, and it reproduces on
untouched workflows too (I confirmed by recompiling `link-checker`), so the
pin is realigned rather than regressed, exactly as #10174 did.
`python .github/scripts/check_action_pins.py` passes: *2076 references across
45 files*.

Compiled with `gh aw compile --strict` (v0.83.1): 0 errors, 0 warnings.


.github/workflows/build-failure-analysis.md
```diff
@@ -267,15 +268,67 @@ jobs:
...
+          # Match on tokens rather than a name prefix: timeline job names and
+          # artifact names are not spelled the same way. `Linux Debug` publishes
+          # `Logs_Build_Linux_Debug`, but `Windows Debug` publishes
+          # `Logs_Build_Windows_NT_Debug` — an inserted `NT` that any prefix
+          # rule would miss, reporting a missing leg on every Windows failure.
+          # Requiring each token of the job name to appear as a token of the
+          # artifact name handles both spellings.
+          timeline_json=$(curl -sSL --retry 3 --max-time 60 "${ADO_API}/build/builds/${BUILD_ID}/timeline?api-version=7.1" 2>/dev/null || true)
+          MISSING_LEGS=""
+          if [ -n "${timeline_json}" ]; then
```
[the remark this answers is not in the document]
YuliiaKovalova: Fixed in 0426f61 — this was a real hole, thanks.

The old code only checked `[ -n "$timeline_json" ]`. Three distinct failures all fell through to `MISSING_LEGS=""`, which is byte-for-byte the same result as *"every failed leg published its logs"*:

1. `curl … || true` returning an empty body (network/DNS/timeout),
2. a non-JSON body (proxy HTML error page, truncated response),
3. a well-formed JSON **ADO error document** — valid JSON, no `records` key, so the `jq -r '.records // []'` pipe quietly yields nothing.

The consequence is the one the check exists to prevent: the agent sees an empty missing-leg list, concludes the artifact set is complete, and can report *"no compile error — non-build failure"* from a set whose completeness was never established.

**Fix:** probe the shape before trusting it, and distinguish *"verified complete"* from *"could not verify"*.

```bash
timeline_ok=0
if printf '%s' "${timeline_json}" | jq -e 'type == "object" and has("records")' >/dev/null 2>&1; then
  timeline_ok=1
fi
if [ "${timeline_ok}" -eq 1 ]; then
  …existing leg scan…
fi

if [ "${timeline_ok}" -ne 1 ]; then
  MISSING_LEGS="(unknown - could not read the build timeline)"
  echo "::warning::Could not read the timeline for build ${BUILD_ID}; unable to verify that every failed leg published a logs artifact."
elif [ -n "${MISSING_LEGS}" ]; then
  echo "::warning::Failed leg(s) with no published logs artifact: ${MISSING_LEGS}"
fi
```

`shared/build-failure-analysis-shared.md` now states that a `GH_AW_MISSING_LEGS` value beginning with `(unknown` must be treated exactly like a non-empty list — i.e. the agent must not claim the failure is non-build-related.

**Deliberately advisory, not fail-closed.** A transient ADO timeline hiccup must not suppress analysis of a genuine compile break; the binlogs themselves are already in hand. Downgrading the *confidence* of the conclusion is the proportionate response, and it is the artifact-download path (not this one) that fails closed.

**Verified.** 13-case unit test over the extracted probe — all pass:

| input | `MISSING_LEGS` |
|---|---|
| empty body / HTML page / truncated JSON | `(unknown - could not read the build timeline)` |
| ADO error document / JSON array / `null` / whitespace | `(unknown - could not read the build timeline)` |
| records present, all legs logged | *(empty)* |
| records present, one leg missing | `MacOS Release` |
| `records: []` / `records: null` / no failed jobs / non-Job record | *(empty)* |

The same flaw was present in dotnet/sdk#55539, which this workflow was backported from; fixed there in the same shape (`bc719fc`).

.github/workflows/build-failure-analysis-command.md
```diff
@@ -192,15 +247,67 @@ jobs:
...
+          # Match on tokens rather than a name prefix: timeline job names and
+          # artifact names are not spelled the same way. `Linux Debug` publishes
+          # `Logs_Build_Linux_Debug`, but `Windows Debug` publishes
+          # `Logs_Build_Windows_NT_Debug` — an inserted `NT` that any prefix
+          # rule would miss, reporting a missing leg on every Windows failure.
+          # Requiring each token of the job name to appear as a token of the
+          # artifact name handles both spellings.
+          timeline_json=$(curl -sSL --retry 3 --max-time 60 "${ADO_API}/build/builds/${BUILD_ID}/timeline?api-version=7.1" 2>/dev/null || true)
+          MISSING_LEGS=""
+          if [ -n "${timeline_json}" ]; then
```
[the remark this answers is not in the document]
YuliiaKovalova: Same fix, applied to the auto (check_run) workflow as well — see the reply on the companion thread for the full analysis. Both workflows and the shared prompt were updated in 0426f61, and the probe is covered by a 13-case unit test (all pass).

.github/workflows/build-failure-analysis-command.md
```diff
@@ -192,15 +247,81 @@ jobs:
...
+                for tok in $(printf '%s' "${jobname}" | tr -c 'A-Za-z0-9' ' '); do
+                  case "_${n}_" in
+                    *_"${tok}"_*) ;;
+                    *) all=0; break ;;
+                  esac
+                done
+                [ "${all}" -eq 1 ] && { found=1; break; }
+              done
+              [ "${found}" -eq 0 ] && MISSING_LEGS="${MISSING_LEGS:+${MISSING_LEGS}, }${jobname}"
+            done < <(printf '%s' "${timeline_json}" | jq -r '.records // [] | map(select(.type=="Job" and .result=="failed")) | .[].name' 2>/dev/null)
```
Evangelink: **The token rule fixes the insertion case but not substitution — this will fire on real builds of this pipeline.**

Cross-referencing `azure-pipelines.yml` against `eng/common/templates/job/job.yml:53` (`Logs_Build_$(Agent.Os)_$(_BuildConfig)`):

| Failed job (timeline `.name`) | Published artifact | Match? |
|---|---|---|
| `Windows Debug` | `Logs_Build_Windows_NT_Debug` | ✅ |
| `Linux Debug` | `Logs_Build_Linux_Debug` | ✅ |
| **`MacOS Debug`** (`azure-pipelines.yml:712`) | `Logs_Build_Darwin_Debug` — `Agent.Os` is `Darwin` on macOS | ❌ false positive |
| **`WindowsSamples Debug`** (`:762`, logs named `Logs_Build_$(Agent.JobName)_$(_BuildConfig)` at `:758`) | `Logs_Build_Debug_Debug` | ❌ false positive |
| **`Detect changed paths`** (`:200`, not under the arcade template, publishes no logs) | — | ❌ false positive |
| `Windows Debug` in the `msbuild_cache_seed` stage (`:133`, publishes no logs) | matches the *build* stage's artifact | ❌ silently masked |

macOS is the one that matters: the pipeline documents it as "historically slow and flaky" and non-blocking (`:715-722`), and the residual failure modes it calls out (90-minute timeout, agent loss, publish-task failure) all still mark the job `failed` while arcade's log publish runs on `always()` — so the artifact *is* published, just spelled `Darwin`.

The validation on build 1534452 does not cover this: the macOS legs there **succeeded** (`$(_MacOSNonBlockingTrailer)` swallows command exit codes), so there was no failed macOS job to mis-match, and `WindowsSamples` was gated off by `DetectChanges`. "No false missing-leg warning for `Windows Debug`" only exercises the case the token rule was designed for.

Not cosmetic: `shared/build-failure-analysis-shared.md` now tells the agent that a non-empty `GH_AW_MISSING_LEGS` means it must **not** conclude "non-build failure". Since clean-compile / test-only failures are the common case here, a routine flaky-macOS build turns into a noisy or hedged analysis.

Options: filter the timeline to the `build` stage plus a known job-name allowlist; or map job → artifact explicitly (`Windows`→`Windows_NT`, `MacOS`→`Darwin`) instead of inferring; or at minimum treat a job whose leading token maps to no known OS/leg as "not a build leg" and skip it.

Two smaller points on the same query:
- `result=="failed"` misses `canceled` / `abandoned` legs, which also publish no logs and are a real gap.
- Matching ignores stage scoping entirely, which is what masks the `msbuild_cache_seed` row above.

.github/workflows/build-failure-analysis-command.md
```diff
@@ -226,6 +347,13 @@ jobs:
             if [ "${ZIP_BYTES}" -gt "${MAX_ZIP_BYTES}" ]; then
               echo "::warning::Skipping ${name}: download exceeded ${MAX_ZIP_BYTES} bytes."; continue
             fi
+            # Bound cumulative *compressed* bytes too. The per-artifact and
+            # cumulative-uncompressed caps still allow many mid-sized archives
+            # to be pulled over the network before any of them is inspected.
+            if [ $((TOTAL_ZIP_BYTES + ZIP_BYTES)) -gt "${MAX_TOTAL_ZIP_BYTES}" ]; then
+              echo "::warning::Cumulative compressed download budget ${MAX_TOTAL_ZIP_BYTES} reached at ${name}; stopping."; break
+            fi
+            TOTAL_ZIP_BYTES=$((TOTAL_ZIP_BYTES + ZIP_BYTES))
```
Evangelink: `TOTAL_ZIP_BYTES` is only accumulated **after** the per-artifact `MAX_ZIP_BYTES` skip on line 347 and the empty-download skip on line 344 — but the bytes are already on the wire by then (`curl … | head -c $((MAX_ZIP_BYTES + 1))`, line 342).

So an artifact that trips the 500 MB per-artifact cap costs ~500 MB of network and is never charged to the cumulative budget. With `MAX_ARTIFACTS=40`, a build whose `Logs_Build_*` artifacts are all oversized downloads up to ~20 GB, not 3 GB — exactly the "many archives pulled over the network before any of them is inspected" scenario the comment above cites. `timeout-minutes: 15` on the job caps the real-world damage, but the stated bound does not hold.

Suggest moving the increment up to right after `ZIP_BYTES` is computed (line 343) and doing the budget check after incrementing. Bytes on the wire are bytes on the wire whether or not the archive is later skipped.

.github/workflows/build-failure-analysis-command.md
```diff
@@ -326,6 +464,7 @@ jobs:
             echo "pr-merge-sha=${BUILD_MERGE_SHA}"
             echo "ado-build-id=${BUILD_ID}"
             echo "ado-build-url=${ADO_BUILD_UI}?buildId=${BUILD_ID}"
+            echo "missing-legs=${MISSING_LEGS}"
```
Evangelink: `MISSING_LEGS` is built from ADO timeline job display names, which come from the pipeline YAML **in the PR branch** — fork-controlled for public PR validation. A `displayName: "x\ny"` puts a newline in the value, which turns this single `key=value` line into arbitrary extra output keys. Because `missing-legs` is echoed *last* in this block, an injected `pr-number=` would override the legitimate one just above and could retarget the analysis comment. Line 520 has the same shape for `$GITHUB_ENV`.

This file already uses the safe idiom a few lines below (`GH_AW_BINLOG_LIST<<GH_AW_EOF`), and the loop at line 330 explicitly sanitizes `name` because it is "PR-controlled ADO artifact metadata" — same provenance, same argument.

Either strip CR/LF where `MISSING_LEGS` is assembled:

```bash
MISSING_LEGS=$(printf '%s' "${MISSING_LEGS}" | tr -d '\r\n')
```

or use a delimiter heredoc for both this write and the `$GITHUB_ENV` one on line 520.

.github/workflows/build-failure-analysis-command.md
```diff
@@ -256,8 +384,18 @@ jobs:
...
+            # `unzip` exit 11 means "no files matched" — the artifact carries no
+            # binlog at all. That is not an extraction failure: the leg did
+            # publish its logs, they simply contain no binlog, and the
+            # fail-closed check below already accounts for a leg that staged
+            # nothing. Reporting it as "extraction failed or timed out" sends
+            # the reader chasing a corrupt-archive theory that isn't there. Any
+            # other non-zero exit (corrupt archive, timeout) is a real failure.
+            uz=0
+            timeout 120 unzip -o /tmp/a.zip '*.binlog' -d /tmp/ax >/dev/null 2>&1 || uz=$?
+            if [ "${uz}" -ne 0 ] && [ "${uz}" -ne 11 ]; then
```
Evangelink: With `uz=11` the code falls through to `TOTAL_BYTES=$((TOTAL_BYTES + UNCOMP))` on line 403 — but nothing was written to `/tmp/ax`. That contradicts the comment immediately below it: *"Consume the cumulative budget only once the archive actually extracted … so a skipped leg can't wrongly exhaust the budget and force later legs to be dropped as 'incomplete'."*

Concretely: a 2 GB logs archive containing no binlog now burns half the 4 GB budget and can push a genuinely useful later leg past `MAX_TOTAL_BYTES` → `break` → `staged_legs` mismatch → silent `emit_none`.

Since both paths already end in `staged_legs < ${#names[@]}`, the behavioural outcome of the exit-11 special case is identical either way — the change is purely a log-text improvement, so it would be a shame to pay for it with a budget-accounting regression. Suggest `continue`-ing on exit 11 with its own non-alarming warning ("published logs contain no binlog"), which keeps the better diagnostics *and* the documented budget semantics.

.github/workflows/build-failure-analysis-command.md
```diff
@@ -89,9 +89,64 @@ jobs:
...
+        env:
+          GH_TOKEN: ${{ github.token }}
+          COMMENTER: ${{ github.event.comment.user.login || fromJSON(github.event.inputs.aw_context || github.event.client_payload.aw_context || '{}').actor }}
+        run: |
+          set +e
+          authorized=false
+          if [ -z "${COMMENTER}" ]; then
+            echo "::warning::No actor resolved from the slash-command event / aw_context; skipping the binlog download."
+          else
+            resp=$(gh api "repos/${GITHUB_REPOSITORY}/collaborators/${COMMENTER}/permission" 2>/dev/null)
```
Evangelink, line 142: `PR_NUMBER` is validated as numeric a little further down with the explicit rationale *"so a malformed event/aw_context payload can't reach those URLs with unexpected content"*. `COMMENTER` has the same provenance (`aw_context`) and reaches both an API path here and `::warning::` output on line 142, but gets no equivalent guard.

Two lines makes the gate self-consistent:

```bash
case "${COMMENTER}" in
  *[!A-Za-z0-9-]*) echo "::warning::Refusing implausible actor '${COMMENTER}'."; COMMENTER="" ;;
esac
```

Two nits while here: `maintainer` on line 136 is not a GitHub `role_name` value (`maintain` is) — harmless and consistent with gh-aw's `GH_AW_REQUIRED_ROLES`, but worth a word so nobody "fixes" `maintain` away later. And line 142's `resolved role '${role:-none}'` prints `none` when only `permission` resolved; `${role:-${perm:-none}}` would match the success message on line 140.

.github/workflows/build-failure-analysis-command.md
```diff
@@ -192,15 +247,81 @@ jobs:
...
-          # all legs so many individually-small artifacts can't collectively
-          # exhaust the runner's disk.
+          # extraction time, AND enforce cumulative budgets across all legs so
+          # many individually-small artifacts can't collectively exhaust the
+          # runner's disk or its network time.
           MAX_ZIP_BYTES=524288000       # 500 MB compressed per artifact
           MAX_UNZIP_BYTES=2147483648    # 2 GB uncompressed per artifact
           MAX_TOTAL_BYTES=4294967296    # 4 GB uncompressed across all artifacts
+          MAX_TOTAL_ZIP_BYTES=3221225472 # 3 GB compressed downloaded in total
+          MAX_ARTIFACTS=40              # legs to process (testfx currently has 7)
```
Evangelink: Nit: the pipeline can publish **8** `Logs_Build_*` artifacts — Windows/Linux/MacOS × Debug/Release, plus `WindowsSamples` × 2 (`azure-pipelines.yml:752-762`). The PR description says 6 (that build had `WindowsSamples` gated off by `DetectChanges`). Since the count is path-dependent, maybe say so rather than pinning a number that is wrong in both directions.

Evangelink (commented):
Reviewed the whole change (both `.md` sources, both regenerated locks, and the shared prompt) at `0426f61`.

**The headline change — the write-access pre-gate — is sound and correctly implemented.** It defaults to deny, every failure path (empty actor, `gh` error, 404, unparseable JSON) lands in the deny branch, it is ordered after gh-aw's injected `Configure GH_HOST` step so `gh api` targets the right host, and the actor is passed via `env:` rather than interpolated into `run:`. I also checked the pin claim independently: no `uses:` line changes in either lock, and `python .github/scripts/check_action_pins.py` passes on this branch (*2076 references across 45 files*). The duplicated blocks in the command and non-command workflows are byte-identical — no drift. The `timeline_ok` probe added in the latest push is the right call: an unreadable timeline reporting as "nothing missing" was the sharpest edge of the missing-leg feature.

My remaining concerns are all in the **backported** part, not the permission gate. Two I'd want addressed before merge; the rest are smaller. Details inline; below is what does not map to a diff line.

---

### `GH_AW_MISSING_LEGS` is missing from the agent playbook's variable table

`shared/build-failure-analysis-shared.md` step 1 adds `GH_AW_MISSING_LEGS` to the list the agent must read, then tells it to `cat .github/agents/build-failure-analyst.agent.md`. That file's table (lines 25-33) is introduced with *"The caller also sets the environment variables below. You must read all of them before doing anything else."* — and the new variable is not in it. Worth a row so the playbook stays authoritative, especially now that it carries a sentinel value (`(unknown …`) whose semantics are documented only in the shared prompt.

### Prompt question: silent `noop` when legs are missing

When `GH_AW_MISSING_LEGS` is non-empty **and** every visible leg compiled cleanly, the guidance is "name them in your `noop` reason" — so the incomplete-picture warning never reaches a human on the PR at all. Deliberate, or would a short comment be better in that specific case? Genuine question.

### Residual: only the *role* half of `pre_activation` is replicated

The pre-gate checks the role but not the command **position**, so a write-access user writing `blah /analyze-build-failure` still triggers the full download and is only refused afterwards by `check_command_position`. Fine trade-off, but a sentence in the comment block would stop a future reader from reading it as an oversight.

### Structural alternative (take it or leave it)

A tiny `authorize` job that `fetch-binlog` declares `needs:` on would make the gate a first-class node in the job graph — visible in the Actions UI, skippable as a whole, no runner spun for the fetch — without recreating the `pre_activation` cycle. Functionally equivalent to the inline step; purely a legibility call.

### Duplication

~75 lines are now copy-pasted verbatim across `build-failure-analysis.md` and `build-failure-analysis-command.md`. They are identical today, but this is the third block to be duplicated, and the latest push had to make the same `timeline_ok` edit twice. If gh-aw can share a `steps:` fragment the way `imports:` already shares the prompt, this seems like the moment.

.github/workflows/build-failure-analysis-command.md
YuliiaKovalova: You're right, and it's worse than the `MacOS`/`Darwin` case you spotted -- I replayed the rule over ~100 recent builds and it produced **only** false positives, never a true one.

The root cause is that the artifact name never comes from the timeline job name. Arcade builds it from pipeline variables:

- `eng/common/templates/job/job.yml` -> `Logs_Build_$(Agent.Os)_$(_BuildConfig)`
- `Agent.Os` is `Windows_NT` / `Linux` / `Darwin`, while the jobs are called `Windows` / `Linux` / `MacOS`

So any spelling I infer is a guess. Measured on real builds:

| case | old rule |
|---|---|
| every build with a failed macOS leg (1529490, 1529415, 1529404, 1529398, 1529394, 1529386, 1529380) | reports `MacOS Debug, MacOS Release` missing while `Logs_Build_Darwin_*` **were** published |
| 10 of 40 recent builds | report `Windows $(_BuildConfig) Release` -- the `msbuild_cache_seed` stage, whose display name contains the literal unexpanded variable |
| `Detect changed paths`, `Monitor Helix Jobs` (SDK) | publish nothing by design -> reported missing |
| true positives | **0** |

Rather than patch the matching, I dropped name inference entirely and asked the timeline the question directly. Arcade's publish step is its own `Task` record parented to the job, so:

- failed/canceled/abandoned job **with** a `Publish logs` task that never succeeded -> genuinely missing
- failed job with **no** such task -> not a logs-publishing leg, skipped (this is what removes `Detect changed paths`, `msbuild_cache_seed` and `Monitor Helix Jobs`)

Matching is on an alphanumeric-normalized name because the same task is called `Publish logs` here and `🟣 Publish Logs` in dotnet/sdk -- exact equality silently matched zero there.

Also added `canceled` and `abandoned` to the result filter as you noted; those legs finish without logs too.

Replayed the compiled `.lock.yml` (not the source) against real ADO: the new rule returns empty on every one of the builds listed above, and 16 unit cases cover emoji task names, the injection and length bounds, and malformed timelines. Same change is in dotnet/sdk#55539.

Residual gap, documented: a job abandoned before any task record materializes has no `Publish logs` record and is skipped.

.github/workflows/build-failure-analysis-command.md
YuliiaKovalova: Confirmed, and I reproduced your numbers. `curl ... | head -c $((MAX_ZIP_BYTES + 1))` puts the bytes on the wire before the size is known, so charging `TOTAL_ZIP_BYTES` after the per-artifact skip meant an oversized artifact cost a full `MAX_ZIP_BYTES` of transfer and was accounted as zero.

Reordered to: compute `ZIP_BYTES` -> charge it -> check the total budget (`break`) -> then the zero-size and per-artifact skips. The bound is now `MAX_TOTAL_ZIP_BYTES` + at most one artifact.

Simulated with 40 artifacts at 501 MiB each: **20,000 MiB before, 3,500 MiB after**. A healthy build (6 x 80 MB) is byte-identical to before.

.github/workflows/build-failure-analysis-command.md
YuliiaKovalova: Good catch -- `MISSING_LEGS` is ADO job display names, which come from YAML on the PR branch, so on a fork PR it is attacker-controlled. A newline would have injected additional keys into `$GITHUB_OUTPUT`/`$GITHUB_ENV`.

Sanitized at assembly rather than at write time, so every consumer gets the clean value: `gsub("[[:cntrl:]]"; " ")` inside jq, then `tr -d '\r\n'` and `cut -c1-400`. The value is also written **first** in the block so it cannot override `pr-number` or any later key even if something slipped through.

(Note for anyone copying this: writing `[\u0000-\u001f]` inside a bash single-quoted jq program does *not* escape -- Oniguruma sees the literal characters and reads it as the range `u0`-`u`, which strips most of the ASCII alphabet. `[[:cntrl:]]` is correct and UTF-8 safe.)

Covered by unit cases with embedded `\n`, `\r` and a 900-char name.

.github/workflows/build-failure-analysis-command.md
YuliiaKovalova: Right -- exit 11 means `unzip` matched nothing, so `UNCOMP` is whatever was already on disk and adding it to `TOTAL_BYTES` charges the budget for an extraction that never happened. A run of large binlog-free archives could push a genuinely useful later leg out of budget.

Now `continue`s on 11 with a non-alarming message before any accounting.

Worth noting this is load-bearing in dotnet/sdk: on `release/*` builds the artifact layout has no `_Logs_Attempt` suffix, so the candidate set is every artifact on the build and non-log ones (e.g. `BuildConfiguration`) legitimately hit exit 11. Treating that as a failure fails the run closed on every servicing PR.

.github/workflows/build-failure-analysis-command.md
YuliiaKovalova: All three taken.

- `COMMENTER` comes from the same `aw_context` provenance as `PR_NUMBER` but had none of its validation. Added the same `case` charset guard (GitHub logins are `[A-Za-z0-9-]`); anything else is rejected before it reaches the API call, and an unresolvable actor denies.
- Message now reports `${role:-${perm:-none}}` so a response carrying only `permission` is shown instead of printing `none` for a value we did have.
- `maintainer` vs `maintain`: both are intentional. `role_name` returns `maintain`, and `permission` has historically returned `maintainer` on some responses -- the check is the union of the two fields, so both spellings are listed. Added a comment saying so; it read like a typo otherwise.

.github/workflows/build-failure-analysis-command.md
YuliiaKovalova: Fixed -- the comment now says it's a cap only and that the real artifact count is path-dependent, instead of pinning a number that is wrong for this pipeline and drifts as legs are added.

Evangelink (approved):

PR #10407, opened 2026-08-03
Restrict the binlog pre-gate to `.permission` and separate the concurrency groups

YuliiaKovalova:
Carries over review feedback from dotnet/arcade#17228 (thanks @Evangelink), where this pre-gate shipped first. Two changes, plus a fix for a bug found while validating the first one.

## 1. The pre-gate matched `role_name`, which widened it

The gate accepted the union of `role_name` and `permission` over `admin|maintain|maintainer|write`. The [REST docs for the endpoint](https://docs.github.com/en/rest/collaborators/collaborators#get-repository-permissions-for-a-user) say:

> The `permission` attribute provides the legacy base roles of `admin`, `write`, `read`, and `none`, where the `maintain` role is mapped to `write` and the `triage` role is mapped to `read`. The `role_name` attribute provides the name of the assigned role, **including custom roles**.

So `.permission in {admin, write}` is already exactly "has push access or better" — precisely the set `roles: [admin, maintainer, write]` describes, maintainers included. The old comment's claim that a `.permission`-only check would deny maintainers was wrong. Matching `role_name` only added risk: a custom organization role merely *named* like a privileged one (a custom `maintainer` inheriting read) would clear the gate with no push access.

The gate now reads `.permission` alone. Note this is a slightly *stricter* early-out than gh-aw's authoritative `pre_activation` check, which prefers `role_name` and falls back to `inherited_role` — the safe direction for a pre-gate, since it can only cost a skipped download, never grant access.

**Bug found while validating:** `gh api ... --jq '.permission'` does not fail closed cleanly. On a non-2xx response `gh` prints the error document to **stdout** and `--jq` does not filter it, so `perm` becomes a raw JSON blob that gets echoed into the workflow log. Reading the response first and extracting with `jq` yields `""` for any error shape.

`shell: bash` added to every hand-written `run:` step so the shell is explicit rather than inherited from the runner default.

## 2. The two workflows shared a concurrency group

Both resolved to `build-failure-analysis-<pr>`. Concurrency groups are repository-global and `cancel-in-progress` is true, so an automatic analysis and an on-demand `/analyze-build-failure` for the same PR cancelled each other.

Observed live in a fork: run `30814539759` (command, `item_number=999999`) had `pre_activation`, `activation`, `agent` and `safe_outputs` all cancelled when run `30814558788` (automatic, `pr-number=999999`) started.

The command workflow now uses `build-failure-analysis-cmd-<pr>`. Each still collapses its own repeat invocations for a PR.

## Verification

- `gh aw compile --strict` with **v0.83.4** (matching current `main` after the rebase): 0 errors, 0 warnings
- `fetch-binlog` still byte-identical across both locks (same normalized SHA-256)
- `retention-days` still `"1"`, not `1.0`
- `bash -n` clean on all three generated `run:` blocks per lock
- `actions/checkout` pins are all v7.0.1 — the compile drift from #10258 no longer reproduces on v0.83.4, so no manual realignment was needed; `check_action_pins.py` passes (2177 refs / 47 files)
- the new gate exercised against the live GitHub API: admin → allow; `read` → deny; `github-actions[bot]` → rejected before the call; empty login → rejected; `foo/../../bar` → rejected; nonexistent user → deny (`'none'`)

The same changes are in dotnet/arcade#17240 and dotnet/sdk#55539.


Evangelink (changes requested):
The source workflow changes look correct: the `.permission`-only gate is fail-closed and removes the custom-role-name authorization hole, the `-cmd-` concurrency namespace prevents command and automatic runs from cancelling each other, and the explicit Bash shells are appropriate.

I cannot approve the current head because both generated lock files conflict with current `main` and were compiled with gh-aw v0.83.1, while `main` has moved to v0.83.4. Resolving the conflicts by retaining this branch's lock files would downgrade the compiler/setup pin and several digest-pinned workflow dependencies.

Please rebase onto current `main`, regenerate both lock files with `gh aw compile --strict` using the repository-aligned v0.83.4 toolchain, and rerun the action-pin audit. Once the regenerated locks preserve the source changes and current pins, this should be ready to approve.

Evangelink (approved):
Re-reviewed after the rebase and strict v0.83.4 regeneration. The merge conflicts and stale-toolchain concern are resolved, the generated action and container pins match current `main`, and the intended permission and concurrency changes remain correct.

Issue #10505, opened 2026-08-07
Inherited [TestInitialize] from a base compiled against MSTest v3 does not run under v4


nohwnd:
When a test class compiled against MSTest v4 inherits a `[TestInitialize]` from a base class that was compiled against MSTest v3, the inherited method does not run. There is no build error and no discovery error, the test just runs without the setup.

I am fairly sure this is because v4 renamed the framework assembly from `Microsoft.VisualStudio.TestPlatform.TestFramework` to `MSTest.TestFramework`, so the `[TestInitialize]` on the v3 compiled base is a different type identity than the one the v4 adapter looks for, and the adapter does not recognize it.

## Repro

Two projects, the base compiled against v3, the test against v4.

`BaseLibV3/BaseLibV3.csproj`

```xml
<Project Sdk="Microsoft.NET.Sdk">
  <PropertyGroup>
    <TargetFramework>net8.0</TargetFramework>
  </PropertyGroup>
  <ItemGroup>
    <PackageReference Include="MSTest.TestFramework" Version="3.6.0" />
  </ItemGroup>
</Project>
```

`BaseLibV3/TestBase.cs`

```csharp
using Microsoft.VisualStudio.TestTools.UnitTesting;

namespace Repro
{
    public abstract class TestBase
    {
        protected bool BaseInitializeRan;

        [TestInitialize]
        public void BaseInitialize() => BaseInitializeRan = true;
    }
}
```

`TestsV4/TestsV4.csproj`

```xml
<Project Sdk="Microsoft.NET.Sdk">
  <PropertyGroup>
    <TargetFramework>net8.0</TargetFramework>
  </PropertyGroup>
  <ItemGroup>
    <PackageReference Include="MSTest.TestFramework" Version="4.3.3" />
    <PackageReference Include="MSTest.TestAdapter" Version="4.3.3" />
    <PackageReference Include="Microsoft.NET.Test.Sdk" Version="18.4.0" />
  </ItemGroup>
  <ItemGroup>
    <ProjectReference Include="..\BaseLibV3\BaseLibV3.csproj" />
  </ItemGroup>
</Project>
```

`TestsV4/SampleTests.cs`

```csharp
using Microsoft.VisualStudio.TestTools.UnitTesting;

namespace Repro
{
    [TestClass]
    public class SampleTests : TestBase
    {
        [TestMethod]
        public void InheritedTestInitialize_ShouldHaveRun()
[...]

PR #10508, opened 2026-08-07
Detect inherited fixtures/tests from a different MSTest version (MSTEST0082)

nohwnd:
Fixes the silent v3/v4 inheritance breakage from #10505 by detecting it at build time.

## The problem

When a `[TestClass]` compiled against MSTest v4 inherits a `[TestInitialize]` / `[TestMethod]` / … from a base class compiled against v3, the inherited member does not run and inherited test methods are not discovered — with no build error and no discovery error. The framework assembly was renamed:

| MSTest | framework assembly |
| --- | --- |
| v3 | `Microsoft.VisualStudio.TestPlatform.TestFramework` |
| v4 | `MSTest.TestFramework` |

so the attribute baked into the v3-compiled base is a different CLR type identity than the one the v4 adapter matches, and it is silently ignored.

## The rule

`InheritedMemberFromDifferentMSTestVersionAnalyzer` (MSTEST0082, Usage, Warning). For a test class it walks the base types and warns when an inherited lifecycle or test-method attribute resolves to a framework assembly whose name differs from the one the test project compiles against. Assembly identity is matched by name, so a benign version bump of the same framework is not flagged — only the rename is.

One rule covers the six inherited attributes (`TestInitialize`, `TestCleanup`, `ClassInitialize`, `ClassCleanup`, `TestMethod`, `DataTestMethod`), with the attribute name passed as a message argument. `AssemblyInitialize`/`AssemblyCleanup` are excluded: assembly fixtures are discovered per-assembly, never inherited, so recompiling the base cannot make them run.

It only warns when recompiling the base would actually fix the problem:
- both the framework assembly and each inherited attribute are resolved through the canonical MSTest attribute in the applied attribute's base chain, so custom `[TestClass]` / `[TestMethod]` subclasses are handled — no false positive on a custom test-class attribute, and a custom `[RetryTest : TestMethodAttribute]` from a v3 base is still caught;
- the member must be one MSTest would run/discover: public, correct static-ness, valid signature (parameterless instance fixtures, `TestContext`-shaped class fixtures, `void`/`Task`/`ValueTask` returns), class fixtures only with `InheritanceBehavior.BeforeEachDerivedClass`, and members overridden or hidden with `new` in a more-derived type are suppressed.

It's a `SymbolKind.NamedType` action that early-outs via `IsTestClass` and skips same-assembly base types.

## Tests

15 unit tests. Warns: inherited `[TestInitialize]`, inherited `[TestMethod]`, a custom `[RetryTest : TestMethodAttribute]` subclass, an inheritable `[ClassInitialize]`, and the realistic graph where the old framework is reached only through the base-library metadata (not a direct reference of the test project). Silent: same-version cross-assembly base, intra-assembly inheritance, custom `[TestClass]` with a same-version inherited fixture, private / overridden / `new`-hidden / invalid-signature members, non-inheritable `[ClassInitialize]`, `[AssemblyInitialize]`, and a non-test-class derived type.

🤖


src/Analyzers/MSTest.Analyzers/InheritedMemberFromDifferentMSTestVersionAnalyzer.cs
```diff
@@ -0,0 +1,156 @@
...
+
+        // The test project compiles against exactly one MSTest framework assembly. Anchor on the assembly that
+        // provides the [TestClass] attribute actually applied to this (source-defined) test class, because source
+        // binding is deterministic even when a differently-versioned framework is also referenced transitively.
+        IAssemblySymbol? referenceAssembly = null;
+        foreach (AttributeData attribute in classSymbol.GetAttributes())
+        {
+            if (attribute.AttributeClass.Inherits(testClassAttributeSymbol))
+            {
+                referenceAssembly = attribute.AttributeClass!.ContainingAssembly;
```
Evangelink: This uses the concrete applied test-class attribute's assembly as the "framework" assembly, which creates a false positive for MSTest's supported custom `[TestClass]` extensibility.

A minimal same-version repro is:

```csharp
public sealed class MyTestClassAttribute : TestClassAttribute { }

public abstract class TestBase
{
    [TestInitialize]
    public void Initialize() { }
}

[MyTestClass]
public sealed class Tests : TestBase { }
```

`IsTestClass` correctly recognizes `MyTestClassAttribute` because it derives from the resolved `TestClassAttribute`. The same `Inherits` check on line 81 then succeeds, but line 83 records `MyTestClassAttribute.ContainingAssembly`—the test/helper assembly—as `referenceAssembly`. The inherited `[TestInitialize]` comes from `MSTest.TestFramework`, so the assembly-name comparison reports MSTEST0082 even though every MSTest type is from the same version. This also reproduces when the base and derived classes are in the same assembly.

Custom test-class attributes are an intentional framework feature: `TestClassAttribute` is non-sealed and exposes virtual `GetTestMethodAttribute`, and existing analyzer tests cover derived test-class attributes (for example `PreferTestInitializeOverConstructorAnalyzerTests.WhenDerivedTestClassAttributeAndExplicitParameterlessCtor_Diagnostic`).

Please anchor the comparison on the canonical `TestClassAttribute` in the applied attribute's base chain (or simply `testClassAttributeSymbol.ContainingAssembly` when it is unambiguous), rather than the concrete custom attribute's assembly. A regression test should use a custom `[TestClass]` plus a same-framework inherited `[TestInitialize]` and expect no diagnostic; placing the custom attribute in a separate helper project would cover the most realistic shape.

src/Analyzers/MSTest.Analyzers/InheritedMemberFromDifferentMSTestVersionAnalyzer.cs
```diff
@@ -0,0 +1,156 @@
...
+        if (attributeClass is null
+            || attributeClass.ContainingNamespace is not { } containingNamespace
+            || !string.Equals(containingNamespace.ToDisplayString(), MSTestNamespace, StringComparison.Ordinal))
+        {
+            return null;
+        }
+
+        // Attributes whose effect is inheritance-sensitive and matched by exact type identity at run time. When the
+        // declaring type comes from a foreign framework assembly these are silently ignored on the derived class.
+        return attributeClass.Name switch
```
Evangelink: This exact concrete-name/namespace match misses inherited custom test-method attributes compiled against the old framework, even though they have the same silent discovery failure that MSTEST0082 is intended to expose.

For example, a v3 base library can define:

```csharp
public sealed class RetryTestAttribute : TestMethodAttribute { }

public abstract class TestBase
{
    [RetryTest]
    public void InheritedTest() { }
}
```

When a v4 `[TestClass]` derives from `TestBase`, the v4 adapter checks whether the runtime attribute is the v4 `TestMethodAttribute` (`AttributeQueryHelper.IsAttributeDefined<TAttribute>` uses `attribute is TAttribute`). A `RetryTestAttribute` whose base is the v3 `TestMethodAttribute` fails that CLR identity check, so the inherited test is silently undiscovered. The analyzer also misses it: the concrete attribute is named `RetryTestAttribute` (and commonly uses a consumer namespace), so this helper returns `null` without examining its v3 `TestMethodAttribute` ancestor.

Custom `TestMethodAttribute` subclasses are supported and exercised throughout this repository, including `MSTest.SourceGeneration.UnitTests.ReflectionMetadataGeneratorTests.Generator_DiscoversMethodsViaCustomTestMethodAttributeSubclass` and the `LegacyExtensibilityTests` acceptance coverage.

Please walk the applied attribute's base-type chain and identify the canonical well-known MSTest attribute there; compare that ancestor's `ContainingAssembly` with the current framework. Add a regression test where the simulated legacy project defines both a v3-like `TestMethodAttribute` and a custom subclass, applies the subclass to an inherited test, and expects MSTEST0082. If custom attributes are intentionally outside this proof-of-concept's scope, document the limitation explicitly so the rule does not imply complete coverage of inherited test methods.

Evangelink (commented):
Additional review notes after validating the analyzer against MSTest's custom-attribute extensibility. I am submitting these as comments only—not requesting changes—so the PR is not rejected or blocked while the author is away.

src/Analyzers/MSTest.Analyzers/InheritedMemberFromDifferentMSTestVersionAnalyzer.cs
nohwnd: Good catch, fixed. The reference assembly is now resolved from the canonical `TestClassAttribute` in the applied attribute's base chain (falling back to the well-known symbol), so a custom `[TestClass]` subclass resolves to the framework assembly instead of the user assembly. Added a regression test with a custom `[TestClass]` plus a same-framework inherited `[TestInitialize]` that expects no diagnostic (`WhenCustomTestClassAttributeAndSameVersionInheritedTestInitialize_NoDiagnostic`). Done in 3998b0c.

🤖

src/Analyzers/MSTest.Analyzers/InheritedMemberFromDifferentMSTestVersionAnalyzer.cs
nohwnd: Done. The member scan now walks the applied attribute's base chain to the canonical MSTest attribute, so a v3-based `[RetryTest : TestMethodAttribute]` is matched through its `TestMethodAttribute` ancestor and compared on that ancestor's assembly. Added `WhenInheritedCustomTestMethodSubclassComesFromDifferentFrameworkAssembly_Diagnostic`, where the simulated legacy project defines `TestMethodAttribute` plus a `RetryTest` subclass applied to an inherited method, expecting MSTEST0082. Done in 3998b0c.

🤖

nohwnd:
Addressed the automated review as well (3998b0c):

- Dropped `[AssemblyInitialize]`/`[AssemblyCleanup]` — assembly fixtures are discovered per-assembly, never inherited, so recompiling the base cannot make them run.
- `[ClassInitialize]`/`[ClassCleanup]` now warn only when the attribute specifies `InheritanceBehavior.BeforeEachDerivedClass`; the default `None` does not flow to derived classes even in the same version.
- Filtered to the effective inherited-member set: public instance test methods and instance fixtures (overridden base lifecycle methods are suppressed), and public static class fixtures.
- Added a test for the unresolved/missing framework graph — the framework is compiled separately and omitted from the consumer references, so the base attribute is reached as a missing type through the base library metadata.
- The rule description now lists `[DataTestMethod]`, and same-assembly base types are skipped.

13 tests, all green.

🤖

nohwnd:
Second pass (8555a7f), addressing the follow-up review:

- **Full per-kind signature validation** before reporting: parameterless instance fixtures, `TestContext`-shaped class fixtures, `void`/`Task`/`ValueTask` returns, correct static-ness. A member with a mismatched attribute but an invalid signature (which would not run even after recompiling) is no longer flagged.
- **`new`-hidden members** are now suppressed as well as overrides — MSTest suppresses an inherited member whenever a more-derived type declares one with the same name.
- Fixed the `[ClassInitialize]` positive test to use a valid signature (a `TestContext` parameter) so it demonstrates a real scenario rather than an already-invalid member; added tests for a `new`-hidden member and an invalid-signature fixture.
- `GetFrameworkAssembly` and `HasBeforeEachDerivedClassBehavior` converted to LINQ per the code-quality analyzer.
- PR description corrected to six inherited attributes (assembly fixtures excluded).

15 tests, all green; analyzer builds clean (0 warnings).

🤖

src/Analyzers/MSTest.Analyzers/InheritedMemberFromDifferentMSTestVersionAnalyzer.cs
```diff
@@ -0,0 +1,268 @@
...
+    private static bool HasClassFixtureParameterShape(IMethodSymbol method)
+        => method.Parameters.IsEmpty
+            || (method.Parameters.Length == 1
+                && method.Parameters[0].Type.ContainingNamespace is { } containingNamespace
+                && string.Equals(containingNamespace.ToDisplayString(), MSTestNamespace, StringComparison.Ordinal)
+                && string.Equals(method.Parameters[0].Type.Name, "TestContext", StringComparison.Ordinal));
+
+    private static bool HasBeforeEachDerivedClassBehavior(AttributeData attribute)
+        => attribute.ConstructorArguments.Any(argument =>
+            argument.Type is INamedTypeSymbol argumentType
```
Evangelink, line 312: The new "unreferenced framework" test still does not exercise the class-fixture path that depends on decoded constructor arguments. `AddLegacyFrameworkAndBaseProjects` uses Roslyn project/compilation references, which preserve enough symbol information for the base library's attribute; a real emitted base-library PE whose legacy framework dependency is absent from the consumer behaves differently.

A reproducer needs three compilation steps:

1. Emit a legacy framework PE containing `InheritanceBehavior` and `ClassInitializeAttribute(InheritanceBehavior)`.
2. Emit a base-library PE against it with `[ClassInitialize(InheritanceBehavior.BeforeEachDerivedClass)]`.
3. Analyze a consumer that references the base-library PE and the current MSTest framework, but **not** the legacy framework PE.

In that third compilation Roslyn represents the legacy attribute through missing/error metadata: the assembly/type name remains available, but the attribute constructor cannot be bound and `ConstructorArguments` is empty. This predicate therefore returns `false`, and MSTEST0082 misses the inherited class fixture even though the encoded attribute selected `BeforeEachDerivedClass`. The existing unresolved test uses `[TestInitialize]`, which does not read constructor arguments, so it cannot catch this asymmetry.

Please add an emitted-PE regression test for both `ClassInitialize` and `ClassCleanup`. There is no safe one-line fallback here: treating an undecodable argument as `BeforeEachDerivedClass` would falsely flag the parameterless/default `InheritanceBehavior.None` case. Reliable complete detection would require decoding the custom-attribute blob from the defining metadata; otherwise the rule should explicitly document that inherited class fixtures are only detected when the legacy attribute constructor resolves, and avoid describing the project-reference test as equivalent to missing PE metadata.

Evangelink (commented):
Fresh review of the latest head. The current automated review already captures the method-hiding, return-type, and ClassInitialize parameter-shape mismatches, so I am not duplicating those inline. This additional comment covers the remaining unresolved-metadata case. Comment-only review; no rejection or request-changes state.

Evangelink (commented):
## Fresh code + MSTest expert review (`8555a7ff`)

I ran two independent reviews of the complete latest diff. They converged on the same four open semantic gaps and found no additional correctness, concurrency, localization, API, or performance issues. The PR head has not changed since the previous review, so I am consolidating the result here rather than duplicating the existing inline threads.

1. **Missing legacy PE metadata can hide inherited class fixtures.** `HasBeforeEachDerivedClassBehavior` depends on decoded `AttributeData.ConstructorArguments`. In the real three-PE graph (legacy framework PE → base-library PE → consumer omitting the legacy framework), Roslyn retains the missing attribute identity but cannot bind its constructor, leaving the arguments empty. An explicit `BeforeEachDerivedClass` class initialize/cleanup therefore escapes MSTEST0082. The current project-reference test covers only `[TestInitialize]` and preserves dependency symbols. Add an emitted-PE test with `MetadataReference.CreateFromImage`; complete support requires metadata-blob decoding, otherwise document/limit this scope. Existing thread: `InheritedMemberFromDifferentMSTestVersionAnalyzer.cs:228`.

2. **Any same-name derived method suppresses a real warning.** `IsHiddenOrOverriddenInDerivedType` treats every derived method with the same name as a replacement. The adapter is kind-specific: an invalid/private `Initialize(int)` does not suppress base `Initialize()`, different-signature test overloads remain discoverable, and class fixtures have no ordinary name-based suppression. Make hiding signature- and member-kind-aware and add C#/VB coverage for each case. Existing current-head automated finding: line 244.

3. **Invalid return types pass analyzer validation.** Name-only matching accepts `Task<T>` and `ValueTask<T>`, while `ReturnsVoid` also accepts `async void`/VB `Async Sub`. The adapter requires exact non-generic `Task`/`ValueTask` and rejects asynchronous void. Resolve the task symbols at compilation start, compare with `SymbolEqualityComparer.Default`, and require `!method.IsAsync` for void. Existing current-head automated finding: line 217.

4. **`ClassInitialize` and `ClassCleanup` parameter rules are conflated.** The shared `ClassFixture` branch allows zero or one `TestContext`, but initialize requires exactly one while cleanup permits zero or one. Preserve distinct member kinds and add a no-diagnostic test for parameterless legacy `ClassInitialize`. Existing current-head automated finding: line 207.

Previously reported custom `[TestClass]`, custom `TestMethodAttribute`, assembly-fixture, `[DataTestMethod]`, private/override, and PR-scope issues are fixed. CI is green, but it does not exercise the four cases above.

This is a comment-only review: no rejection or request-changes state.

src/Analyzers/MSTest.Analyzers/InheritedMemberFromDifferentMSTestVersionAnalyzer.cs
nohwnd: Confirmed and fixed in 236dc34. You're right that the project reference preserves the referenced symbols, so it isn't equivalent to an absent PE. I added real emitted-PE regression tests — compile the legacy framework and the base library to PEs, then link only the base library into the consumer: the instance-fixture case is still detected with the legacy framework absent, while `ClassInitialize` and `ClassCleanup` are not, exactly the empty-`ConstructorArguments` asymmetry you described. Since decoding the attribute blob isn't available to an analyzer, I documented the limitation on the analyzer (type remarks + `HasBeforeEachDerivedClassBehavior`) and corrected the misleading project-reference test name/comment so it no longer claims equivalence with missing PE metadata.

> [!NOTE]
> Auto-resolved by the [GitHub Copilot app](https://github.com/github/app).

nohwnd:
All four gaps from your Aug 9 review are now addressed (`110fd39`, building on `236dc34`):

1. **Missing legacy PE metadata / class fixtures** — added emitted-PE tests (`MetadataReference.CreateFromImage`): instance fixtures are still detected with the legacy framework absent; `ClassInitialize`/`ClassCleanup` are (documented as) not detected because `ConstructorArguments` is empty. The type remarks and `HasBeforeEachDerivedClassBehavior` note the limitation, and the project-reference test name/comment no longer claim missing-PE equivalence.
2. **Hiding** — `IsHiddenOrOverriddenInDerivedType` is now member-kind- and signature-aware: an instance fixture is suppressed only by a same-name derived method that is itself a valid fixture; a test method only by a same-signature method; class fixtures accumulate and are never name-suppressed. Overrides still always suppress. Added a C# test per case plus two Visual Basic tests for parity.
3. **Return types** — `Task`/`ValueTask` are resolved at compilation start and compared with `SymbolEqualityComparer`, so `Task<T>`/`ValueTask<T>` are rejected, and `void` requires `!method.IsAsync`.
4. **`ClassInitialize` / `ClassCleanup` params** — split into distinct member kinds: `ClassInitialize` requires exactly one `TestContext`, `ClassCleanup` permits zero or one; a parameterless legacy `[ClassInitialize]` no longer warns, and there's a no-diagnostic test for it.

28 tests cover the rule; full analyzer suite green.

🤖

test/UnitTests/MSTest.Analyzers.UnitTests/InheritedMemberFromDifferentMSTestVersionAnalyzerTests.cs
```diff
@@ -0,0 +1,910 @@
...
+    }
+
+    // The emitted assemblies must be compiled against the same reference assemblies the analyzer test uses for the
+    // consumer, otherwise the base library's core-type references (e.g. System.Object) resolve to a different corlib
+    // identity and the consumer fails with CS0012.
+    private static readonly MetadataReference[] FrameworkReferences = ResolveFrameworkReferences();
+
+    private static MetadataReference[] ResolveFrameworkReferences()
+    {
+        string nuGetConfig = Path.Combine(RootFinder.Find(), "NuGet.config");
```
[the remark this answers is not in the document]
nohwnd: `Path.Combine` is correct here: the second argument is the compile-time constant `"NuGet.config"`, which is never rooted, so it cannot drop the first argument. `Path.Join` isn't available on .NET Framework 4.7.2, which this test project targets (`net472;net8.0`) — switching to it fails the net472 build with `CS1501: No overload for method 'Join' takes 2 arguments`. This also matches the existing `CSharpCodeFixVerifier`/`VisualBasicCodeFixVerifier` test helpers, which use the same `Path.Combine(RootFinder.Find(), "NuGet.config")` pattern. Keeping `Path.Combine`.

> [!NOTE]
> Auto-resolved by the [GitHub Copilot app](https://github.com/github/app).

src/Analyzers/MSTest.Analyzers/InheritedMemberFromDifferentMSTestVersionAnalyzer.cs
```diff
@@ -0,0 +1,424 @@
...
+    // mirroring MSTEST0003's rule (e.g. 'T' must appear as 'T', 'T[]', or 'List&lt;T&gt;' in some parameter).
+    private static bool AllGenericTypeParametersAreInferable(IMethodSymbol method)
+    {
+        foreach (ITypeParameterSymbol typeParameter in method.TypeParameters)
+        {
+            if (!method.Parameters.Any(parameter => IsOrHasTypeParameter(parameter.Type, typeParameter)))
+            {
+                return false;
+            }
+        }
```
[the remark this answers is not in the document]
nohwnd: Done — rewrote `AllGenericTypeParametersAreInferable` as a single `TypeParameters.All(... Parameters.Any(...))` expression, so the filtering is explicit LINQ with the same short-circuit semantics.

> [!NOTE]
> Auto-resolved by the [GitHub Copilot app](https://github.com/github/app).

src/Analyzers/MSTest.Analyzers/InheritedMemberFromDifferentMSTestVersionAnalyzer.cs
```diff
@@ -0,0 +1,424 @@
...
+
+        if (type is INamedTypeSymbol namedType)
+        {
+            foreach (ITypeSymbol typeArgument in namedType.TypeArguments)
+            {
+                if (IsOrHasTypeParameter(typeArgument, typeParameter))
+                {
+                    return true;
+                }
+            }
```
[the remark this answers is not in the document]
nohwnd: Done — `IsOrHasTypeParameter` now returns `namedType.TypeArguments.Any(typeArgument => IsOrHasTypeParameter(...))` instead of the manual `foreach`, keeping behavior identical.

> [!NOTE]
> Auto-resolved by the [GitHub Copilot app](https://github.com/github/app).

src/Analyzers/MSTest.Analyzers/InheritedMemberFromDifferentMSTestVersionAnalyzer.cs
```diff
@@ -0,0 +1,393 @@
...
+
+        for (int index = 0; index < left.Parameters.Length; index++)
+        {
+            IParameterSymbol leftParameter = left.Parameters[index];
+            IParameterSymbol rightParameter = right.Parameters[index];
+
+            // ref/out/in all share the same by-ref CLR signature, so compare by-value versus by-ref rather than the
+            // exact RefKind (a derived 'Run(out int)' still hides a base 'Run(ref int)').
+            if ((leftParameter.RefKind == RefKind.None) != (rightParameter.RefKind == RefKind.None)
+                || !SymbolEqualityComparer.Default.Equals(leftParameter.Type, rightParameter.Type))
```
Evangelink: This signature comparison omits return type and static/instance calling convention, so it suppresses MSTEST0082 for base test methods that CLR reflection still exposes.

For example:

```csharp
// Legacy base assembly
public abstract class TestBase
{
    [TestMethod]
    public void Run() { }
}

// Current consumer assembly
[TestClass]
public sealed class Tests : TestBase
{
    public new int Run() => 0;
    // The same problem occurs with: public new static void Run() { }
}
```

`HaveSameSignature` returns true in both cases because arity and parameters match. However, a direct probe using the adapter's effective flags (`Public | NonPublic | Instance | Static`) returns **both** methods:

```text
DerivedReturn: Int32 Run()  and  BaseReturn: Void Run()
DerivedStatic: static Void Run()  and  BaseStatic: instance Void Run()
```

That matches `TypeEnumerator`: it enumerates all runtime methods, rejects the derived `int`/static method in `TestMethodValidator`, and still discovers the inherited base `[TestMethod]`. Because the base attribute has the old CLR identity, that inherited test is silently lost under v4—the exact case MSTEST0082 should report—but this helper treats the unrelated derived method as hiding it and suppresses the diagnostic.

Please make the `TestMethod` replacement check mirror discovery: the derived method must itself be a discoverable public instance test-method shape, and its CLR signature comparison must include the return type (as `MethodInfo.ToString()` does in duplicate detection) plus the static/instance convention. Add diagnostic regression cases for a derived `new int Run()` and `new static void Run()` over an inherited legacy `void Run()`.

Evangelink (commented):
Fresh review of the latest head (`53457772`) using independent code and MSTest-expert passes plus a direct CLR reflection probe. The existing current-head automated review already records the protected nested-class visibility issue, so this review adds only the new signature-hiding defect below. Comment-only; no rejection or request-changes state.

src/Analyzers/MSTest.Analyzers/InheritedMemberFromDifferentMSTestVersionAnalyzer.cs
nohwnd: Good catch — fixed. You're right that `HaveSameSignature` ignored return type and static/instance convention, so it suppressed the diagnostic for base test methods that discovery still exposes.

I made the test-method replacement check mirror `TypeEnumerator`: the derived method now has to be a discoverable current-version test method itself (public, instance, void/Task/ValueTask, inferable generics, and a current-framework `[TestMethod]`/`[DataTestMethod]`), and the signature comparison now includes the return type and the static/instance convention. So a derived `new int Run()`, `new static void Run()`, or a `new void Run()` without a recognized `[TestMethod]` no longer suppresses — the inherited legacy test is reported, exactly as reflection would still discover it.

Added diagnostic regression tests for all three shadowing shapes (different return type, static, and no-`[TestMethod]`), and corrected the same-signature no-diagnostic hiding test to use a genuine current-version `[TestMethod]` that really does hide the base. Done in 18228f3e.

> [!NOTE]
> Auto-resolved by the [GitHub Copilot app](https://github.com/github/app).

JanKrivanek (approved):
**Code looks good to go** at head `824705c9`; no major correctness, API, or test-coverage concerns found.

src/Analyzers/MSTest.Analyzers/InheritedMemberFromDifferentMSTestVersionAnalyzer.cs
```diff
@@ -0,0 +1,480 @@
...
+    // not by name, so 'Run&lt;T&gt;(T)' and 'Run&lt;U&gt;(U)' share a signature. Recurses through arrays and constructed
+    // generic types and falls back to symbol identity for everything else.
+    private static bool AreSignatureTypesEquivalent(ITypeSymbol left, ITypeSymbol right)
+    {
+        if (left is ITypeParameterSymbol leftTypeParameter && right is ITypeParameterSymbol rightTypeParameter)
+        {
+            return leftTypeParameter.TypeParameterKind == rightTypeParameter.TypeParameterKind
+                && (leftTypeParameter.TypeParameterKind != TypeParameterKind.Method
+                    ? SymbolEqualityComparer.Default.Equals(leftTypeParameter, rightTypeParameter)
+                    : leftTypeParameter.Ordinal == rightTypeParameter.Ordinal);
```
Evangelink: The ordinal normalization here does not match the adapter's actual de-duplication key for generic methods. `TypeEnumerator` first keys tests by `MethodInfo.ToString()`, and that string preserves **generic parameter names**.

A direct reflection probe for:

```csharp
public class Base
{
    [TestMethod]
    public void Run<T>(T value) { }
}

public class Derived : Base
{
    [TestMethod]
    public new void Run<U>(U value) { }
}
```

returns both methods:

```text
Void Run[U](U)
Void Run[T](T)
```

Those keys differ, so `foundDuplicateTests` stays false and `TypeEnumerator` returns both tests after the legacy base is recompiled. The analyzer instead treats `T` and `U` as equivalent by ordinal, considers the derived method to hide the base method, and suppresses MSTEST0082—missing a real discovery change.

Please compare method type parameters the same way `MethodInfo.ToString()` does for this purpose: names must match (while recursively preserving the existing array/constructed-type handling). Change `WhenInheritedGenericTestMethodShadowedByRenamedTypeParameter_NoDiagnostic` to expect the diagnostic, and add a companion same-name `Run<T>(T)` hider that remains no-diagnostic.

src/Analyzers/MSTest.Analyzers/InheritedMemberFromDifferentMSTestVersionAnalyzer.cs
```diff
@@ -0,0 +1,480 @@
...
+        {
+            foreach (ISymbol member in type.GetMembers(baseMethod.Name))
+            {
+                if (member is not IMethodSymbol derivedMethod)
+                {
+                    continue;
+                }
+
+                // An override always replaces the inherited member.
+                if (Overrides(derivedMethod, baseMethod))
```
Evangelink, line 347: The unconditional override suppression misses custom MSTest-derived attributes that explicitly enable inheritance.

Standard `TestMethodAttribute` and lifecycle attributes use `Inherited = false`, so an attribute-less override normally does replace the attributed base member. But supported custom attributes can opt back in:

```csharp
[AttributeUsage(AttributeTargets.Method, Inherited = true)]
public sealed class RetryTestAttribute : TestMethodAttribute { }

public abstract class Base
{
    [RetryTest]
    public virtual void Run() { }
}

[TestClass]
public sealed class Derived : Base
{
    public override void Run() { }
}
```

`ReflectionOperations.GetCustomAttributes` uses `inherit: true`; a direct probe confirms the custom attribute appears on `Derived.Run`. When `RetryTestAttribute` derives from the v3 framework type, the v4 adapter ignores that inherited instance. Recompiling the base/custom attribute against v4 makes the same override discoverable. MSTEST0082 should therefore report this migration break, but this early return suppresses it before member-kind logic.

Please make override suppression account for the applied attribute's effective `AttributeUsageAttribute.Inherited` value. Keep suppressing standard/default non-inherited attributes and overrides that carry their own current-version test attribute; report when the legacy custom attribute is inheritable and is the only attribute that would make the override discoverable. Add C# and VB tests for `Inherited=true`, `Inherited=false`, and a current-attributed override.

Evangelink (commented):
Fresh code and MSTest-expert review of head `824705c9`, reconciled with direct reflection probes. The previous visibility/signature blockers were addressed, but two narrow false negatives remain. This is a comment-only review; no rejection or request-changes state.

src/Analyzers/MSTest.Analyzers/InheritedMemberFromDifferentMSTestVersionAnalyzer.cs
nohwnd: Fixed, and verified with a reflection probe first — you're right on both counts. `MethodInfo.ToString()` keeps the type-parameter name (`Void Run[U](U)` vs `Void Run[T](T)`), so I was normalizing by ordinal in the wrong direction: a renamed `Run<U>(U)` does NOT hide `Run<T>(T)`.

`AreSignatureTypesEquivalent` now compares method type parameters by name (still recursing through arrays/constructed types). Flipped `WhenInheritedGenericTestMethodShadowedByRenamedTypeParameter` to expect the diagnostic, and added `WhenInheritedGenericTestMethodShadowedBySameTypeParameterName_NoDiagnostic` as the same-name companion. Done in 183e9c38.

> [!NOTE]
> Auto-resolved by the [GitHub Copilot app](https://github.com/github/app).

src/Analyzers/MSTest.Analyzers/InheritedMemberFromDifferentMSTestVersionAnalyzer.cs
nohwnd: Fixed — and confirmed `ReflectionOperations.GetCustomAttributes` uses `inherit: true` and that all the standard attributes are `Inherited=false`, plus a probe showing an override inherits an `Inherited=true` custom attribute but not an `Inherited=false` one.

Override suppression is no longer unconditional: it now reads the applied attribute's effective `AttributeUsage.Inherited` (walking the attribute's base chain, defaulting to true). It keeps suppressing non-inherited attributes (the standard case) and overrides that carry their own current-version attribute, and reports when an inheritable legacy custom attribute is the only thing that would make the override discoverable. Added C# and VB tests for `Inherited=true` (diagnostic), `Inherited=false` (no diagnostic), and a current-attributed override (no diagnostic), and set the existing override test's attribute to `Inherited=false` to match the real framework. Done in 183e9c38.

> [!NOTE]
> Auto-resolved by the [GitHub Copilot app](https://github.com/github/app).

src/Analyzers/MSTest.Analyzers/InheritedMemberFromDifferentMSTestVersionAnalyzer.cs
```diff
@@ -0,0 +1,527 @@
...
+                continue;
+            }
+
+            foreach (KeyValuePair<string, TypedConstant> namedArgument in usage.NamedArguments)
+            {
+                if (string.Equals(namedArgument.Key, "Inherited", StringComparison.Ordinal) && namedArgument.Value.Value is bool inherited)
+                {
+                    return inherited;
+                }
+            }
```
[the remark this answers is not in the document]
nohwnd: Done — `IsAppliedAttributeInheritable` now filters `usage.NamedArguments.Where(namedArgument => namedArgument.Key == "Inherited")` and just extracts the boolean inside the loop, so the filtering is explicit. Behavior is unchanged. Done in 23b5cf78.

> [!NOTE]
> Auto-resolved by the [GitHub Copilot app](https://github.com/github/app).

PR #10626, opened 2026-08-18
Fix apphost DOTNET_ROOT path assertion

nohwnd:
Compare the derived Windows path without drive-letter case sensitivity. `DOTNET_HOST_PATH` can use a lowercase drive letter while `RootFinder` returns uppercase, even though both resolve to the same directory.

Verified: the affected `net11.0|x64` acceptance test passes locally.

🤖


Evangelink (approved):
Approving assymiung there was a scenario that triggered the issue and fix

PR #10638, opened 2026-08-18
Skip the binlog steps when there are no binlogs

nohwnd:
`artifacts/tmp/<config>/testsuite` is only created once the acceptance tests actually run, so it is still missing whenever a build or test failure happens before that point. `CopyFiles@2` and `PublishBuildArtifacts` both hard-fail on a missing folder ("Not found SourceFolder" / "Not found PathtoPublish"), so every failed build got two extra red steps on top of the step that actually broke:

- `Copy binlogs` — fails on the missing `testsuite` folder. Arcade pre-creates `artifacts/tmp/<config>` (`Create-Directory $TempDir` in `eng/common/tools.ps1`) but nothing creates the `testsuite` subfolder until an acceptance test does.
- `Publish integration tests binlogs` — then fails because the copy never created the staging folder.

Neither tells you anything about the failure, and in the official build they sit right next to the real one.

Now a detection step looks for binlogs first and both steps skip when there is nothing to collect, so a failed build only shows the step that broke. When binlogs do exist nothing changes — same artifact, same layout. It also fixes the case where a run that otherwise passed got failed by these two steps because no binlogs happened to be produced.

The same two steps exist in the Windows and Linux/macOS PR jobs with the same failure mode, so they get the same treatment (`pwsh` on Windows, `bash` on the others). The Windows application-model template already handled this by pre-creating its staging directory, so it is untouched.

I checked the detection logic against the three cases — folder missing, folder present but no binlogs, nested binlog present — for both the PowerShell and the bash version, and verified the bash one is safe under `set -e`.

🤖

### DIFF
--- a/azure-pipelines-official.yml
+++ b/azure-pipelines-official.yml
@@ -165,6 +165,28 @@ extends:
                   # mapping is required; the history service no-ops if the token or endpoints are unavailable.
                   SYSTEM_ACCESSTOKEN: $(System.AccessToken)
 
+              # artifacts/tmp/<config>/testsuite is created by the acceptance tests as they run, so it does not
+              # exist yet when the build or the test step fails before reaching them. Both binlog steps below
+              # hard-fail on a missing folder ('Not found SourceFolder' / 'Not found PathtoPublish'), so every
+              # failed build used to report two extra red steps that have nothing to do with the real failure.
+              # Detect the binlogs up front and let the two steps skip when there is nothing to collect, so a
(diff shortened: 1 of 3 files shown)


nohwnd:
Closing as a duplicate. #10613 does the same thing and merged in 26bed22b00f232bf3653948f5174dd4521acda90, so main already probes for the binlogs and skips both steps when the integration suite produced none. This branch conflicts with that and adds nothing on top.

🤖

PR #10641, opened 2026-08-18
Require an execute bit before selecting a Unix apphost

nohwnd:
When launching a managed `.dll`, the server-mode client prefers a sibling apphost and decides with `File.Exists`, which says yes to a file that cannot be started. A test payload built on a Windows agent and executed on a Linux machine arrives with its extensionless apphost stripped of POSIX permission bits, because a zip written on Windows records none. `Process.Start` then throws `Permission denied` and aborts the run rather than falling back.

`BuildLaunch` now gates the choice on `IsUsableApphost`, which additionally requires an execute bit on Unix, so an unusable candidate falls back to `dotnet <dll>`. `File.GetUnixFileMode` is .NET 7+, so the check is fenced on `NET7_0_OR_GREATER` and net462, netstandard2.0, net5.0 and net6.0 consumers keep the existence-only check. That stays correct because the apphost path probe is already OS-aware and never offers a Windows `.exe` on Unix.

The fence is the target framework rather than the package's modern-.NET symbol, which records which JSON slice a consumer compiles and is defined only for net8.0+. NuGet serves the net5.0 slice to net5.0, net6.0 and net7.0 alike, so fencing on it would leave a Linux net7.0 consumer on the existence-only path even though it has the API. An anti-drift test asserts the call sits inside a `NET7_0_OR_GREATER` block in every packed slice.

This restores the second half of the fix in microsoft/vstest#16336, which was lost when vstest deleted its own client in favour of this package in microsoft/vstest#16300.

Reproduced end to end before fixing: `dotnet publish -r linux-x64` on Windows emits an extensionless apphost beside the `.dll`; a Windows-written zip stores `ExternalAttributes = 0`; extracting on Linux yields mode 0644; `Process.Start` throws `Permission denied`. The four alternatives were all checked and none holds — the check is absent from `main`, the layout is reachable, `Process.Start` does not degrade gracefully, and the `.dll` branch is live for consumers even though testfx's own callers pass an apphost.

Verified: `build.cmd -pack` passes; the new tests pass on Windows net8.0 and net462 and on Linux net8.0, and reverting either guard makes its test fail.

Follow-up in microsoft/vstest, once a package with this ships: bump `MicrosoftTestingPlatformServerModeClientSourcesVersion` in `eng/Versions.props` and add an acceptance test for the layout.

🤖


src/Platform/Microsoft.Testing.Platform.ServerMode.Client.Sources/Client/MtpServerProcess.cs
```diff
@@ -335,12 +337,51 @@ private static string GetAppHostPath(string managedAssembly)
...
+        if (!File.Exists(apphost))
+        {
+            return false;
+        }
+
+#if NETCOREAPP
+        // File.GetUnixFileMode is .NET 7+. Consumers compiling the netstandard2.0 slice (net462, and
+        // net5.0-net7.0 which select that slice) fall back to the existence check above. That stays correct
+        // because GetAppHostPath already refuses to consider a Windows ".exe" on Unix, so the only case left
+        // uncovered there is an extensionless file that lost its execute bit.
```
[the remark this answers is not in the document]
nohwnd: Good catch, you're right. NuGet serves the net5.0 slice to net5.0, net6.0 and net7.0, and `MTP_CLIENT_USE_MODERN_DOTNET` is only defined for net8.0+, so a Linux net7.0 consumer kept the existence-only check even though it has `File.GetUnixFileMode`.

Fenced on `#if NET7_0_OR_GREATER` instead. The pack transform only rewrites the `NETCOREAPP` token, so a TFM symbol survives it and the consumer's own SDK defines it. `MtpClientModernPolyfills.cs` already does the same thing with `NET5_0 || NET6_0 || NET7_0`.

Checked the produced package: all three slices now have the call inside the `NET7_0_OR_GREATER` block. Added an anti-drift test for it, and putting the old fence back makes it fail on all three.

🤖

> [!NOTE]

src/Platform/Microsoft.Testing.Platform.ServerMode.Client.Sources/Client/MtpServerProcess.cs
```diff
@@ -335,12 +337,55 @@ private static string GetAppHostPath(string managedAssembly)
...
+        // and is defined only for net8.0+, while NuGet serves the net5.0 slice to net5.0, net6.0 and net7.0
+        // consumers alike. Fencing on it would drop a Linux net7.0 consumer back to the existence-only path
+        // even though it has the API. net462, netstandard2.0, net5.0 and net6.0 consumers genuinely lack the
+        // API and keep the existence check above, which stays correct because GetAppHostPath already refuses
+        // to consider a Windows ".exe" on Unix, so the only case left uncovered there is an extensionless
+        // file that lost its execute bit.
+        if (!OperatingSystem.IsWindows())
+        {
+            const UnixFileMode ExecuteBits = UnixFileMode.UserExecute | UnixFileMode.GroupExecute | UnixFileMode.OtherExecute;
+            return (File.GetUnixFileMode(apphost) & ExecuteBits) != 0;
```
Evangelink, line 404: Checking whether *any* execute bit is set does not prove the current process can execute the file. POSIX selects the owner, group, or other class that applies to the caller; for example, a caller-owned file with only `GroupExecute` set passes this mask even though the owner cannot execute it, and launch still fails with `EACCES`. Could we check effective executable access or retry with `dotnet` when the apphost start is denied? A test using an execute bit from a non-applicable class would guard this.

src/Platform/Microsoft.Testing.Platform.ServerMode.Client.Sources/Client/MtpServerProcess.cs
```diff
@@ -335,12 +337,55 @@ private static string GetAppHostPath(string managedAssembly)
...
+        {
+            return false;
+        }
+
+#if NET7_0_OR_GREATER
+        // Fenced on the target framework because File.GetUnixFileMode is .NET 7+. Deliberately not the
+        // package's modern-.NET compilation symbol: that one records which JSON slice a consumer compiles
+        // and is defined only for net8.0+, while NuGet serves the net5.0 slice to net5.0, net6.0 and net7.0
+        // consumers alike. Fencing on it would drop a Linux net7.0 consumer back to the existence-only path
+        // even though it has the API. net462, netstandard2.0, net5.0 and net6.0 consumers genuinely lack the
```
Evangelink: This leaves net5.0 and net6.0 consumers with the original failure mode. On Unix, an extensionless apphost extracted without execute permission still passes `File.Exists`, so these TFMs select it and `Process.Start` fails with `EACCES` instead of falling back to `dotnet`. Could we use a downlevel-compatible permission probe, or retry through the muxer when starting the apphost fails due to permissions? A packaged net6.0 test would cover the affected slice.

src/Platform/Microsoft.Testing.Platform.ServerMode.Client.Sources/Client/MtpServerProcess.cs
Evangelink: Good catch — fixed in a43021a. Apphost startup now retries through `dotnet` only on Unix `EACCES`, and `StartAsyncOnUnixWhenOnlyNonApplicableExecuteBitIsSetRetriesThroughDotnet` covers an owner file with only `GroupExecute` set.

src/Platform/Microsoft.Testing.Platform.ServerMode.Client.Sources/Client/MtpServerProcess.cs
Evangelink: Fixed in a43021a. The Unix `EACCES` retry is available on every consumer TFM, and `PackagedNet6Consumer_WhenSiblingApphostCannotExecute_RetriesThroughDotnet` exercises the packed net6.0 slice end to end.

Evangelink (approved):

PR #10896, opened 2026-08-31
Add combinatorial test data support

AArnott:
Ports the current exhaustive combinatorial test-data functionality from `AArnott/Xunit.Combinatorial` into `MSTest.TestFramework`, allowing MSTest users to generate Cartesian products from inferred or explicitly supplied parameter values.

- Adds explicit, inferred, range, and random parameter-value providers.
- Intentionally omits pairwise generation, permutations, test-case exclusions, member-backed dynamic values, class-data activation, and the upstream fluent builder.
- Isolates the APIs in `Microsoft.VisualStudio.TestTools.UnitTesting.Combinatorial` for compatibility with the existing `Combinatorial.MSTest` package.
- Adds focused API coverage, executable MSTest examples, localization, and NativeAOT coverage.
- Records that Andrew Arnott contributed this implementation under the repository's MIT License.

Related issue: N/A


src/TestFramework/TestFramework/Internal/CombinatorialValuesUtilities.cs
```diff
@@ -0,0 +1,73 @@
...
+    internal static IEnumerable<object?> GetValuesFor(ParameterInfo parameter)
+    {
+        if (parameter is null)
+        {
+            throw new ArgumentNullException(nameof(parameter));
+        }
+
+        ICombinatorialValuesProvider? valuesSource = parameter.GetCustomAttributes()
+            .OfType<ICombinatorialValuesProvider>()
+            .SingleOrDefault();
```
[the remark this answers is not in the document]
AArnott: Removed `AllowMultiple = true` from `CombinatorialMemberDataAttribute`, so each parameter has exactly one value provider and the existing single-provider resolution remains well-defined. Added attribute-usage coverage for the restriction.

> [!NOTE]

src/TestFramework/TestFramework/Attributes/DataSource/CombinatorialRangeAttribute.cs
```diff
@@ -0,0 +1,126 @@
...
+        if (count < 1)
+        {
+            throw new ArgumentOutOfRangeException(nameof(count));
+        }
+
+        Values = new object[count];
+        for (int i = 0; i < count; i++)
+        {
+            Values[i] = from + i;
+        }
```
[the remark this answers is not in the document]
AArnott: Added widened boundary validation to the signed count overload and widened the stepped-range count and element arithmetic. Overflowing ranges now throw rather than wrapping, with regression coverage for both forms.

> [!NOTE]

src/TestFramework/TestFramework/Attributes/DataSource/CombinatorialRangeAttribute.cs
```diff
@@ -0,0 +1,126 @@
...
+        if (count < 1)
+        {
+            throw new ArgumentOutOfRangeException(nameof(count));
+        }
+
+        Values = new object[count];
+        for (uint i = 0; i < count; i++)
+        {
+            Values[i] = from + i;
+        }
```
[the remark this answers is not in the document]
AArnott: Added widened validation to the unsigned count overload and replaced wraparound-dependent stepped loops with a bounded count plus widened offset arithmetic. Boundary and overshoot cases now have regression coverage.

> [!NOTE]

src/TestFramework/TestFramework/Attributes/DataSource/CombinatorialMemberDataAttribute.cs
```diff
@@ -0,0 +1,214 @@
...
+
+    /// <summary>
+    /// Gets the member name.
+    /// </summary>
+    public string MemberName { get; }
+
+    /// <summary>
+    /// Gets or sets the type from which to retrieve the member. The test class is used by default.
+    /// </summary>
+    public Type? MemberType { get; set; }
```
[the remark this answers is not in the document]
AArnott: Annotated `MemberType` with the same `DynamicDataOperations.RequiredMemberTypes` preservation contract used by `DynamicDataAttribute`, so reflected fields, properties, and methods survive trimming.

> [!NOTE]

src/TestFramework/TestFramework/Attributes/DataSource/CombinatorialClassDataAttribute.cs
```diff
@@ -0,0 +1,55 @@
...
+    /// </summary>
+    /// <param name="valuesSourceType">The type that provides values.</param>
+    /// <param name="arguments">Arguments to pass to the constructor of <paramref name="valuesSourceType"/>.</param>
+    public CombinatorialClassDataAttribute(Type valuesSourceType, params object[]? arguments)
+        => _values = GetValues(valuesSourceType, arguments);
+
+    /// <inheritdoc />
+    public object?[] GetValues(ParameterInfo parameter) => _values;
+
+    private static object?[] GetValues(Type valuesSourceType, object[]? arguments)
```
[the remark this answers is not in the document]
AArnott: Annotated both the public `valuesSourceType` constructor parameter and the private helper parameter with `PublicConstructors`, preserving the activation contract through `Activator.CreateInstance`.

> [!NOTE]

src/TestFramework/TestFramework/Attributes/DataSource/CombinatorialClassDataAttribute.cs
```diff
@@ -0,0 +1,55 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+namespace Microsoft.VisualStudio.TestTools.UnitTesting;
+
+/// <summary>
+/// Specifies a class that provides values for a parameter on a combinatorial test method.
+/// </summary>
+[CLSCompliant(false)]
+public class CombinatorialClassDataAttribute : Attribute, ICombinatorialValuesProvider
```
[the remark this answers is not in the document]
AArnott: Added `[AttributeUsage(AttributeTargets.Parameter)]` to `CombinatorialClassDataAttribute` and covered its target and multiplicity through reflection-based tests.

> [!NOTE]

src/TestFramework/TestFramework/Attributes/DataSource/CombinatorialMemberDataAttribute.cs
```diff
@@ -0,0 +1,214 @@
...
+        {
+            methodInfo = reflectionType.GetRuntimeMethods()
+                .FirstOrDefault(method => method.Name == MemberName && ParameterTypesCompatible(method.GetParameters(), Arguments));
+            if (methodInfo is not null)
+            {
+                break;
+            }
+        }
+
+        if (methodInfo is null || !methodInfo.IsStatic)
```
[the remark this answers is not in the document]
AArnott: Method selection now requires public static visibility inside the search predicate, so an instance overload cannot shadow a valid static overload. Added a regression test with matching instance and static overloads.

> [!NOTE]

src/TestFramework/TestFramework/Attributes/DataSource/CombinatorialRandomDataAttribute.cs
```diff
@@ -0,0 +1,102 @@
...
+                values[index++] = value;
+            }
+            else
+            {
+                collisionCount++;
+            }
+
+            if (collisionCount > collisionChecker.Count * 5 && collisionCount > 1000)
+            {
+                throw new InvalidOperationException("Too many collisions occurred while generating unique random values.");
```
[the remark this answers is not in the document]
AArnott: Replaced collision retries and the cutoff with Floyd's without-replacement sampling algorithm. Dense ranges, including selecting all 1,001 available values, now complete deterministically with unique results and are covered by regression tests.

> [!NOTE]

src/TestFramework/TestFramework/Attributes/DataSource/CombinatorialRandomDataAttribute.cs
```diff
@@ -0,0 +1,102 @@
...
+            throw new InvalidOperationException(string.Format(CultureInfo.CurrentCulture, "{0} must be positive.", nameof(Count)));
+        }
+
+        if (Minimum > Maximum)
+        {
+            throw new InvalidOperationException(string.Format(CultureInfo.CurrentCulture, "{0} must not be greater than {1}.", nameof(Minimum), nameof(Maximum)));
+        }
+
+        int maxPossibleValues = Maximum - Minimum + 1;
+        if (Count > maxPossibleValues)
```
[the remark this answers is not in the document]
AArnott: Range cardinality is now computed in `long`, and inclusive sampling uses a bounded 32-bit rejection sampler rather than `Maximum + 1`. This supports the full signed integer domain and `int.MaxValue`, both covered by tests.

> [!NOTE]

src/TestFramework/TestFramework/Attributes/DataSource/CombinatorialMemberDataAttribute.cs
```diff
@@ -0,0 +1,219 @@
...
+            if (propertyInfo is not null)
+            {
+                break;
+            }
+        }
+
+        if (propertyInfo?.GetMethod is not { IsPublic: true, IsStatic: true })
+        {
+            return null;
+        }
```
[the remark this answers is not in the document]
AArnott: Changed property, method, and field lookup to inspect eligible declared members at each inheritance level and continue past ineligible hidden members. The same update adds deterministic most-specific overload selection and nullable-argument support, with regression coverage for each case.

> [!NOTE]

src/TestFramework/TestFramework/Attributes/DataSource/CombinatorialTestCaseGenerator.cs
```diff
@@ -0,0 +1,121 @@
...
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+namespace Microsoft.VisualStudio.TestTools.UnitTesting;
+
+/// <summary>
+/// Determines whether a generated test case, represented by one selected value index per dimension, is allowed.
+/// </summary>
+/// <param name="indices">The selected zero-based value index for each dimension.</param>
+/// <returns><see langword="true"/> if the test case is allowed; otherwise, <see langword="false"/>.</returns>
+public delegate bool CombinatorialIndexPredicate(int[] indices);
```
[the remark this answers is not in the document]
AArnott: Kept the array delegate for `netstandard2.0` compatibility but now passes a defensive copy to every generator predicate, preventing mutation of the traversal buffer. Added a regression test whose predicate deliberately overwrites its input and verifies all emitted rows remain correct.

> [!NOTE]

src/TestFramework/TestFramework/Attributes/DataSource/CombinatorialTheoryDataBuilder.cs
```diff
@@ -0,0 +1,264 @@
...
+    /// <summary>
+    /// Adds complete hand-authored rows for the base columns.
+    /// </summary>
+    /// <param name="rows">The rows to add. Every row must have the same width.</param>
+    /// <returns>This builder.</returns>
+    public CombinatorialTheoryDataBuilder AddRows(params object?[][] rows)
+    {
+        if (rows.Length == 0)
+        {
+            throw new ArgumentException("At least one row is required.", nameof(rows));
```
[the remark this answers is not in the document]
AArnott: Moved every builder diagnostic into `FrameworkMessages.resx` and regenerated all FrameworkMessages XLF files with `UpdateXlf`. I also covered the rest of the new combinatorial APIs so this localization pattern is consistent across the feature.

> [!NOTE]

src/TestFramework/TestFramework/Attributes/DataSource/CombinatorialClassDataAttribute.cs
```diff
@@ -0,0 +1,64 @@
...
+    /// <param name="arguments">Arguments to pass to the constructor of <paramref name="valuesSourceType"/>.</param>
+    public CombinatorialClassDataAttribute(
+        [DynamicallyAccessedMembers(DynamicallyAccessedMemberTypes.PublicConstructors)] Type valuesSourceType,
+        params object[]? arguments)
+    {
+        _valuesSourceType = valuesSourceType ?? throw new ArgumentNullException(nameof(valuesSourceType));
+        if (!typeof(IEnumerable<object[]>).IsAssignableFrom(valuesSourceType))
+        {
+            throw new InvalidOperationException(
+                $"The values source {valuesSourceType} must be assignable to {typeof(IEnumerable<object[]>)}.");
```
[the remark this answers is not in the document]
AArnott: Moved both class-provider diagnostics into `FrameworkMessages.resx` and regenerated all FrameworkMessages XLF files with `UpdateXlf`. The same change localizes the member, range, random, exclusion, generator, and provider-resolution diagnostics as well.

> [!NOTE]

src/TestFramework/TestFramework/Attributes/DataSource/CombinatorialMemberDataAttribute.cs
```diff
@@ -0,0 +1,300 @@
...
+
+        for (int i = 0; i < parameters.Length; i++)
+        {
+            if (arguments[i] is object argument)
+            {
+                if (!parameters[i].ParameterType.GetTypeInfo().IsAssignableFrom(argument.GetType().GetTypeInfo()))
+                {
+                    return false;
+                }
```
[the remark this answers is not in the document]
AArnott: Updated member-method compatibility so a boxed `T` is accepted for a `Nullable<T>` parameter, and added a regression using a non-null `int?` argument. This is pushed in `e3a10151b`.

> [!NOTE]

src/TestFramework/TestFramework/Attributes/DataSource/CombinatorialMemberDataAttribute.cs
```diff
@@ -0,0 +1,303 @@
...
+            Type otherType = otherParameters[i].ParameterType;
+            if (candidateType == otherType)
+            {
+                continue;
+            }
+
+            if (!otherType.GetTypeInfo().IsAssignableFrom(candidateType.GetTypeInfo()))
+            {
+                return false;
+            }
```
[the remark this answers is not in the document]
AArnott: Consolidated nullable compatibility into one helper used by candidate collection, overload specificity, and member element-type validation. Added regressions for `T` versus `Nullable<T>` overload selection and `IEnumerable<T>` feeding a `Nullable<T>` parameter; `4d822f3cd` also corrects the reviewed resource wording and regenerates the XLF files.

> [!NOTE]

src/TestFramework/TestFramework/Attributes/DataSource/CombinatorialClassDataAttribute.cs
```diff
@@ -0,0 +1,68 @@
...
+        {
+            throw new InvalidOperationException(
+                string.Format(
+                    CultureInfo.CurrentCulture,
+                    FrameworkMessages.CombinatorialClassDataSourceTypeInvalid,
+                    valuesSourceType,
+                    typeof(IEnumerable<object[]>)));
+        }
+
+        _arguments = arguments is null ? null : [.. arguments];
```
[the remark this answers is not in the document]
AArnott: Normalized an explicit null `params` array to one null constructor argument, made the element annotations nullable, and updated the public API baseline. Added an attribute-metadata regression using `[CombinatorialClassData(..., null)]`; the same commit also applies this behavior to `ExcludeTestCaseAttribute` with runner-level coverage.

> [!NOTE]

src/TestFramework/TestFramework/Attributes/DataSource/CombinatorialMemberDataAttribute.cs
```diff
@@ -0,0 +1,306 @@
...
+    /// <summary>
+    /// Initializes a new instance of the <see cref="CombinatorialMemberDataAttribute"/> class.
+    /// </summary>
+    /// <param name="memberName">The name of the public static member that provides values.</param>
+    /// <param name="arguments">Arguments for a method member. They are ignored for fields and properties.</param>
+    public CombinatorialMemberDataAttribute(string memberName, params object?[]? arguments)
+    {
+        MemberName = memberName ?? throw new ArgumentNullException(nameof(memberName));
+        Arguments = arguments;
+    }
```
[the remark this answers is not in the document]
AArnott: Normalized an explicit null `params` array to one null member-method argument and made `Arguments` consistently non-null after construction. Added an attribute-metadata regression using `[CombinatorialMemberData(..., null)]` in `b98fd29ba`.

> [!NOTE]

Evangelink (commented):
The larger file count is mostly explained by testfx requirements such as localization, API baselines, tests, diagnostics, and NativeAOT coverage. I recommend retaining those and the more robust range/random implementations, while reducing unrelated public surface and addressing the inline correctness and compatibility issues before merge.

src/TestFramework/TestFramework/Internal/CombinatorialValuesUtilities.cs
```diff
@@ -0,0 +1,93 @@
...
+    }
+
+    internal static object? GetValueForTestCase(ParameterInfo parameter, ICombinatorialValuesProvider? valuesSource, object?[] candidateValues, int candidateIndex)
+    {
+        if (valuesSource is not (CombinatorialMemberDataAttribute or CombinatorialClassDataAttribute))
+        {
+            return candidateValues[candidateIndex];
+        }
+
+        object?[] freshValues = valuesSource.GetValues(parameter);
```
Evangelink: This re-invokes member/class data once per generated row and parameter, after the provider was already evaluated to build `candidateValues`. Discovery and execution may each enumerate the data again, so expensive or stateful providers can run many times and return different ordering while still passing the length check. Please materialize each provider once and select directly from `candidateValues`; fresh-per-row values would need to be an explicit, documented behavior.

src/TestFramework/TestFramework/Attributes/DataSource/CombinatorialTheoryDataBuilder.cs
```diff
@@ -0,0 +1,271 @@
...
+/// Determines whether a generated combinatorial data row is allowed.
+/// </summary>
+/// <param name="values">The values in the completed data row.</param>
+/// <returns><see langword="true"/> if the row is allowed; otherwise, <see langword="false"/>.</returns>
+public delegate bool CombinatorialTheoryDataPredicate(object?[] values);
+
+/// <summary>
+/// Builds test data from hand-authored base rows and generated columns.
+/// </summary>
+public sealed class CombinatorialTheoryDataBuilder
```
Evangelink: Could we remove this builder from this PR? It is not used by `CombinatorialDataAttribute` or any other production path, introduces a sizable standalone public API, and overlaps with `DynamicData`/`TestDataRow<T>`. The “TheoryData” terminology is also foreign to MSTest. If there is demand for this helper, it would be easier to review as a separate API proposal.

src/TestFramework/TestFramework/Attributes/DataSource/CombinatorialTestCaseGenerator.cs
```diff
@@ -0,0 +1,121 @@
...
+
+    /// <summary>
+    /// Generates every permutation of the supplied values.
+    /// </summary>
+    /// <typeparam name="T">The type of value to permute.</typeparam>
+    /// <param name="values">The values to permute.</param>
+    /// <returns>Every positional permutation of <paramref name="values"/>.</returns>
+    /// <remarks>Input positions are distinct, so equal input values may produce equal output rows.</remarks>
+    public static T[][] GeneratePermutations<T>(T[] values)
+    {
```
Evangelink: This public permutation API has no production caller in this change; only its dedicated tests exercise it. Please remove `GeneratePermutations<T>` from this PR and make the remaining generator/predicate internal, since combination generation is currently an implementation detail of `CombinatorialDataAttribute`. We should avoid permanently exposing a general-purpose combinatorics API without a product use case.

src/TestFramework/TestFramework/Attributes/DataSource/ExcludeTestCaseAttribute.cs
```diff
@@ -0,0 +1,160 @@
...
+                {
+                    if (EqualityComparer<object?>.Default.Equals(candidateValues[parameterIndex][valueIndex], excludedValue))
+                    {
+                        matches[valueIndex] = true;
+                        anyMatches = true;
+                    }
+                }
+
+                if (!anyMatches)
+                {
```
Evangelink: Silently discarding an exclusion when one value matches no candidate makes typos and boxed-type mismatches look successful while the unwanted test case still runs. Please fail with an actionable diagnostic instead of returning `null`, so an unusable `[ExcludeTestCase]` cannot become a silent no-op.

src/TestFramework/TestFramework/Attributes/DataSource/CombinatorialMemberDataAttribute.cs
```diff
@@ -0,0 +1,306 @@
...
+                    string.Join(", ", Arguments.Select(p => p?.GetType().FullName ?? FrameworkMessages.Common_NullInMessages)))
+                : string.Format(CultureInfo.CurrentCulture, FrameworkMessages.CombinatorialMemberNotFound, MemberName, type.FullName);
+            throw new ArgumentException(message);
+        }
+
+        IEnumerable values = accessor() as IEnumerable
+            ?? throw new ArgumentException(
+                string.Format(CultureInfo.CurrentCulture, FrameworkMessages.CombinatorialMemberReturnedNull, MemberName, type.FullName));
+
+        return values is IEnumerable<object[]> rows
```
Evangelink: Flattening `IEnumerable<object[]>` here gives the same shape the opposite meaning from `DynamicData` and `ClassData`, where each array is a row of arguments. That semantic difference in the same namespace will be surprising and difficult to change after release. Could these providers require `IEnumerable<T>` candidate values instead, without special-case flattening?

src/TestFramework/TestFramework/Attributes/DataSource/CombinatorialDataAttribute.cs
```diff
@@ -0,0 +1,56 @@
...
+
+using Microsoft.VisualStudio.TestTools.UnitTesting.Internal;
+
+namespace Microsoft.VisualStudio.TestTools.UnitTesting;
+
+/// <summary>
+/// Provides every possible combination of values for the parameters of a test method.
+/// </summary>
+[AttributeUsage(AttributeTargets.Method, AllowMultiple = true, Inherited = true)]
+public class CombinatorialDataAttribute : Attribute, ITestDataSource
```
Evangelink: These public type names collide with the existing `Combinatorial.MSTest` package (`CombinatorialDataAttribute`, `CombinatorialValuesAttribute`, `CombinatorialRangeAttribute`, etc.). Because MSTest injects the `Microsoft.VisualStudio.TestTools.UnitTesting` namespace through `buildTransitive`, current package users can get ambiguous references immediately after upgrading. The acceptance-test `using` removals in this PR demonstrate the collision. We need an explicit compatibility/migration strategy before shipping these names in-box.

src/TestFramework/TestFramework/Attributes/DataSource/CombinatorialMemberDataAttribute.cs
AArnott: Removed the `IEnumerable<object[]>` flattening special case. Each returned element is now one candidate, and a regression verifies that each `object[]` remains intact for an array-valued parameter.

> [!NOTE]

src/TestFramework/TestFramework/Attributes/DataSource/ExcludeTestCaseAttribute.cs
AArnott: An explicit exclusion value that matches no candidate now throws an actionable localized `ArgumentException` identifying the value and parameter position. Added focused regression coverage and regenerated all XLF files.

> [!NOTE]

src/TestFramework/TestFramework/Attributes/DataSource/CombinatorialTheoryDataBuilder.cs
AArnott: Removed `CombinatorialTheoryDataBuilder` entirely, including its tests, diagnostics, translations, and public API entries.

> [!NOTE]

src/TestFramework/TestFramework/Attributes/DataSource/CombinatorialDataAttribute.cs
AArnott: Moved the built-in public APIs to `Microsoft.VisualStudio.TestTools.UnitTesting.Combinatorial`. Existing consumers of `Combinatorial.MSTest` therefore keep resolving its root types without ambiguity, while migration to the built-in implementation is explicit via the nested namespace.

> [!NOTE]

src/TestFramework/TestFramework/Internal/CombinatorialValuesUtilities.cs
AArnott: Updated in 9b437d93b. Candidate values are now materialized once per parameter and every generated row indexes those cached arrays; focused member/class regressions verify construction and enumeration each occur once.

> [!NOTE]

src/TestFramework/TestFramework/Attributes/DataSource/CombinatorialTestCaseGenerator.cs
AArnott: Removed permutation support and made the remaining Cartesian generator and predicate internal implementation details. Their public API entries are gone and the focused combination tests remain.

> [!NOTE]

src/TestFramework/TestFramework/Attributes/DataSource/ExcludeTestCaseAttribute.cs
```diff
@@ -0,0 +1,162 @@
...
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+namespace Microsoft.VisualStudio.TestTools.UnitTesting.Combinatorial;
+
+/// <summary>
+/// Suppresses generation of a specific test case from a combinatorial test.
+/// </summary>
+[AttributeUsage(AttributeTargets.Method, AllowMultiple = true, Inherited = true)]
+[CLSCompliant(false)]
+public class ExcludeTestCaseAttribute : Attribute
```
Evangelink: Before committing to this API shape, could we consider handling constrained combinations through `DynamicData` instead? `ExcludeTestCase` only expresses equality patterns, splits one data source across multiple method attributes, and requires `AnyDataValue` as a `typeof(...)` sentinel. A reusable Cartesian-product helper would let users apply ordinary LINQ predicates, including relational constraints, while staying within MSTest's established data-source model.

src/TestFramework/TestFramework/Attributes/DataSource/CombinatorialTestCaseGenerator.cs
```diff
@@ -0,0 +1,79 @@
...
+/// Determines whether a generated test case, represented by one selected value index per dimension, is allowed.
+/// </summary>
+/// <param name="indices">The selected zero-based value index for each dimension.</param>
+/// <returns><see langword="true"/> if the test case is allowed; otherwise, <see langword="false"/>.</returns>
+internal delegate bool CombinatorialIndexPredicate(int[] indices);
+
+/// <summary>
+/// Generates exhaustive combinations.
+/// </summary>
+internal static class CombinatorialTestCaseGenerator
```
Evangelink: Could we keep this index-based implementation internal but expose a small value-based helper, for example `CombinatorialDataGenerator.Generate(params object?[][] dimensions)` returning `IEnumerable<object?[]>`? Users could feed it through `[DynamicData]` and compose `.Where(...)`, projections, or `TestDataRow<T>` metadata themselves. That would provide an MSTest-style escape hatch for constrained combinations without the specialized `ExcludeTestCase`/`AnyDataValue` public API or the broader theory-data builder.

Evangelink (commented):
A possible simplification is to keep the attribute-based path focused on straightforward Cartesian generation, while offering a small public generator for advanced composition through the existing `DynamicData` model.

Evangelink (commented):
Two related API-shape suggestions: remove the class-instantiation data source and align named dynamic parameter values with MSTest's existing `DynamicData` terminology.

src/TestFramework/TestFramework/Attributes/DataSource/CombinatorialClassDataAttribute.cs
```diff
@@ -0,0 +1,68 @@
...
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+namespace Microsoft.VisualStudio.TestTools.UnitTesting.Combinatorial;
+
+/// <summary>
+/// Specifies a class that provides values for a parameter on a combinatorial test method.
+/// </summary>
+[AttributeUsage(AttributeTargets.Parameter)]
+[CLSCompliant(false)]
+public class CombinatorialClassDataAttribute : Attribute, ICombinatorialValuesProvider
```
Evangelink: Could we remove `CombinatorialClassDataAttribute` from the initial API? MSTest normally obtains reusable data from named static members rather than constructing an arbitrary enumerable class during discovery. This also overlaps with the external-type support on the member-based provider while adding activation, lifetime, trimming, repeat-enumeration, and unusual `IEnumerable<object[]>` flattening semantics. Inline values/ranges, a named dynamic-values source, and `DynamicData` plus a public Cartesian-product helper would cover the scenarios more consistently.

src/TestFramework/TestFramework/Attributes/DataSource/CombinatorialMemberDataAttribute.cs
```diff
@@ -0,0 +1,288 @@
...
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+namespace Microsoft.VisualStudio.TestTools.UnitTesting.Combinatorial;
+
+/// <summary>
+/// Specifies a member that provides values for a parameter on a combinatorial test method.
+/// </summary>
+[AttributeUsage(AttributeTargets.Parameter)]
+[CLSCompliant(false)]
+public class CombinatorialMemberDataAttribute : Attribute, ICombinatorialValuesProvider
```
Evangelink: This scenario has value, but `MemberData` is not MSTest terminology and suggests xUnit row semantics. Could this be narrowed and named `CombinatorialDynamicValuesAttribute` instead? That pairs naturally with `CombinatorialValuesAttribute`, follows MSTest's established `DynamicData` naming, and makes clear that the named static member supplies candidate values for one parameter rather than complete rows. Ideally the source would simply return `IEnumerable<T>` and avoid class-data or row-flattening semantics.

src/TestFramework/TestFramework/Attributes/DataSource/CombinatorialClassDataAttribute.cs
AArnott: Removed `CombinatorialClassDataAttribute` from the initial API, along with its activation diagnostics, tests, and API entries. Named reusable candidates now flow exclusively through the static-member provider.

> [!NOTE]

src/TestFramework/TestFramework/Internal/CombinatorialValuesUtilities.cs
```diff
@@ -0,0 +1,76 @@
...
+            yield return false;
+        }
+        else if (dataType == typeof(int))
+        {
+            yield return 0;
+            yield return 1;
+        }
+        else if (dataType.GetTypeInfo().IsEnum)
+        {
+            foreach (string name in Enum.GetNames(dataType))
```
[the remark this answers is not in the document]
AArnott: Fixed in b3e411b61. Source generation now emits `DynamicDependency(PublicFields)` for enum and nullable-enum test-method parameter types; focused generator coverage and NativeAOT acceptance exercise an inferred custom enum.

> [!NOTE]

src/TestFramework/TestFramework/Attributes/DataSource/CombinatorialTestCaseGenerator.cs
AArnott: Added public `CombinatorialDataGenerator.Generate(params object?[][])` while retaining the index generator internally. The executable example now demonstrates `DynamicData` plus a relational LINQ constraint.

> [!NOTE]

src/TestFramework/TestFramework/Attributes/DataSource/CombinatorialMemberDataAttribute.cs
AArnott: Renamed and narrowed the API to `CombinatorialDynamicValuesAttribute`, including source, tests, examples, diagnostics, translations, NativeAOT coverage, and API baselines.

> [!NOTE]

src/TestFramework/TestFramework/Attributes/DataSource/CombinatorialTestCaseGenerator.cs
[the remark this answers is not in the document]
AArnott: Updated the PR description to reflect the intentionally narrowed scope: exhaustive combinations only, with no permutations or fluent builder. It now also documents the DynamicData generator and other current API decisions.

> [!NOTE]

src/TestFramework/TestFramework/Attributes/DataSource/ExcludeTestCaseAttribute.cs
AArnott: Implemented this shape in b3e411b61. `ExcludeTestCaseAttribute` and `AnyDataValue` are removed; constrained combinations now use the public Cartesian generator through `DynamicData` and ordinary LINQ predicates.

> [!NOTE]

src/TestFramework/TestFramework/Attributes/DataSource/CombinatorialTestCaseGenerator.cs
Evangelink: After comparing this again with Youssef's implementation, I think my suggestion to add this public helper was premature. The core attribute can generate its Cartesian product directly, while advanced users can already return complete rows through `DynamicData`. We should not add permanent, weakly typed `object?[][]` API solely to compensate for removing other speculative features. Sorry for the churn; I recommend removing `CombinatorialDataGenerator` unless we have concrete independent demand for it.

src/TestFramework/TestFramework/Attributes/DataSource/CombinatorialDataGenerator.cs
```diff
@@ -0,0 +1,44 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+namespace Microsoft.VisualStudio.TestTools.UnitTesting.Combinatorial;
+
+/// <summary>
+/// Generates exhaustive combinations of values for use with data-driven tests.
+/// </summary>
+public static class CombinatorialDataGenerator
```
Evangelink: I reconsidered the earlier suggestion that led to this type. Youssef's implementation does not need a separate generator because Cartesian generation is simply part of `CombinatorialDataAttribute`. Unless there is a concrete use case independent of this feature, this public `object?[][]` API adds permanent weakly typed surface area without being necessary for combinatorial tests. Please remove it for v1; the internal generator may remain for implementation clarity or be folded into the attribute once exclusions and the builder are gone.

src/Analyzers/MSTest.SourceGeneration/Generators/TestClassModelBuilder.cs
```diff
@@ -308,7 +308,19 @@ private static EquatableArray<TestParameterModel> BuildParameters(IMethodSymbol
         for (int i = 0; i < method.Parameters.Length; i++)
         {
             IParameterSymbol p = method.Parameters[i];
-            parameters[i] = new TestParameterModel(p.Type.ToDisplayString(SymbolDisplayFormats.FullyQualified), p.Name);
+            INamedTypeSymbol? namedType = p.Type as INamedTypeSymbol;
+            if (namedType?.OriginalDefinition.SpecialType == SpecialType.System_Nullable_T)
+            {
+                namedType = namedType.TypeArguments[0] as INamedTypeSymbol;
+            }
```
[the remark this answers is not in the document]
AArnott: Added direct nullable-enum coverage in 6289730b2. The generator regression now uses `CustomValue?`, and the NativeAOT asset verifies all three inferred cases (`null`, `First`, and `Second`).

> [!NOTE]

src/TestFramework/TestFramework/Attributes/DataSource/CombinatorialDataGenerator.cs
AArnott: Removed the public `CombinatorialDataGenerator` API and its dedicated tests/example. `CombinatorialDataAttribute` now maps the internal Cartesian index generator directly onto its materialized parameter values.

> [!NOTE]

src/TestFramework/TestFramework/Attributes/DataSource/CombinatorialDynamicValuesAttribute.cs
```diff
@@ -0,0 +1,288 @@
...
+
+        if (enumeratedType.IsGenericType && enumeratedType.GetGenericTypeDefinition() == typeof(IEnumerable<>))
+        {
+            throw new ArgumentException(
+                string.Format(
+                    CultureInfo.CurrentCulture,
+                    FrameworkMessages.CombinatorialDynamicValuesNestedEnumerableUnsupported,
+                    MemberName,
+                    declaringType.FullName,
+                    enumeratedType.GenericTypeArguments[0].Name));
```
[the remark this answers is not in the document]
AArnott: Removed the blanket nested-enumerable rejection so the normal element-to-parameter assignability check determines validity. Added coverage showing `IEnumerable<IEnumerable<int>>` supplies individual candidates to an `IEnumerable<int>` parameter while remaining incompatible with an `int` parameter, and removed the obsolete localized diagnostic.

> [!NOTE]

Evangelink (commented):
A few remaining simplifications would make the initial API smaller and align its attribute behavior with existing MSTest data sources.

src/TestFramework/TestFramework/Attributes/DataSource/CombinatorialDataAttribute.cs
```diff
@@ -0,0 +1,52 @@
...
+
+using Microsoft.VisualStudio.TestTools.UnitTesting.Internal;
+
+namespace Microsoft.VisualStudio.TestTools.UnitTesting.Combinatorial;
+
+/// <summary>
+/// Provides every possible combination of values for the parameters of a test method.
+/// </summary>
+[AttributeUsage(AttributeTargets.Method, AllowMultiple = true, Inherited = true)]
+public class CombinatorialDataAttribute : Attribute, ITestDataSource
```
Evangelink: Before this ships, could we make the inheritance and multiplicity contract deliberate? A second `[CombinatorialData]` instance can only generate the same cases again, and `Inherited = true` differs from both `DataRowAttribute` and `DynamicDataAttribute`. Please use `AllowMultiple = false, Inherited = false` and seal this type. The other new public attribute classes should also be sealed: none exposes virtual members or another designed inheritance point, and this cannot be tightened after release.

src/TestFramework/TestFramework/Attributes/DataSource/CombinatorialTestCaseGenerator.cs
```diff
@@ -0,0 +1,68 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+namespace Microsoft.VisualStudio.TestTools.UnitTesting;
+
+/// <summary>
+/// Generates exhaustive combinations.
+/// </summary>
+internal static class CombinatorialTestCaseGenerator
```
Evangelink: Now that exclusions, the fluent builder, permutations, and the public generator are gone, this helper has only one production caller. Its null/negative-size validation is unreachable because `CombinatorialDataAttribute` derives every dimension from an array length, and the current approach eagerly allocates all `int[]` selections before allocating the actual data rows. Could we fold a small lazy Cartesian-product loop into `CombinatorialDataAttribute` and remove this type, its focused tests, internal API entries, and the `CombinatorialNegativeDimensionSize` resource?

src/TestFramework/TestFramework/Attributes/DataSource/CombinatorialRangeAttribute.cs
```diff
@@ -0,0 +1,145 @@
...
+        {
+            ulong offset = (ulong)i * step;
+            Values[i] = (uint)(ascending ? from + offset : from - offset);
+        }
+    }
+
+    /// <summary>
+    /// Gets the values that should be passed to this parameter.
+    /// </summary>
+    public object[] Values { get; }
```
Evangelink: Does this need to be public API? Youssef's implementation keeps the generated array private, and callers can mutate the live array returned here. `GetValues` is the provider contract, so I suggest keeping the storage private and removing `Values` from the public baseline.

src/TestFramework/TestFramework/Attributes/DataSource/CombinatorialRandomDataAttribute.cs
```diff
@@ -0,0 +1,120 @@
...
+
+    /// <summary>
+    /// Gets or sets the seed used for random number generation.
+    /// </summary>
+    public int Seed { get; set; } = NoSeed;
+
+    /// <summary>
+    /// Gets the generated values.
+    /// </summary>
+    public object[] Values => _values.Value;
```
Evangelink: Please remove this public `Values` property and keep generation behind `GetValues`. Besides exposing mutable internal state, this getter can throw for an invalid `Count`/range and `Lazy<T>` then caches that exception. Validation belongs in the provider operation rather than a property getter; Youssef's implementation likewise exposes only `GetValues` and keeps its cache private.

src/TestFramework/TestFramework/Attributes/DataSource/CombinatorialDataAttribute.cs
AArnott: Made `CombinatorialDataAttribute` sealed with `AllowMultiple = false` and `Inherited = false`. The four parameter-provider attributes are sealed as well, with unit coverage for the contract.

> [!NOTE]

src/TestFramework/TestFramework/Attributes/DataSource/CombinatorialRandomDataAttribute.cs
AArnott: Removed the public random `Values` property and its API baseline entry. `GetValues` now validates the current configuration before accessing the private lazy cache, so invalid configuration is reported by the provider operation without caching the validation exception.

> [!NOTE]

src/TestFramework/TestFramework/Attributes/DataSource/CombinatorialRangeAttribute.cs
AArnott: Removed the public `CombinatorialRangeAttribute.Values` property and its API baseline entry. Generated values are now held in a private field and exposed only through `GetValues`.

> [!NOTE]

src/TestFramework/TestFramework/Attributes/DataSource/CombinatorialTestCaseGenerator.cs
AArnott: Folded a lazy odometer-style Cartesian loop into `CombinatorialDataAttribute`, preserving stable row order without eagerly allocating index arrays. Removed the internal generator, its focused tests/API entries, and the now-unreachable localized diagnostic.

> [!NOTE]

src/TestFramework/TestFramework/Attributes/DataSource/CombinatorialDataAttribute.cs
```diff
@@ -0,0 +1,79 @@
...
+            yield break;
+        }
+
+        int[] indices = new int[values.Length];
+        while (true)
+        {
+            object?[] row = new object?[values.Length];
+            for (int i = 0; i < row.Length; i++)
+            {
+                row[i] = values[i][indices[i]];
```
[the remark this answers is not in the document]
AArnott: Keeping the one-time provider materialization is deliberate and follows the earlier maintainer review on this PR; re-invoking a stateful or expensive provider per Cartesian row would reverse that accepted contract and can itself produce unstable data. MSTest data-source values are passed through as supplied, as with other data sources, and there is no general safe way to clone arbitrary mutable candidates, so callers that need isolated mutable objects should return distinct candidates or immutable values.

> [!NOTE]

src/TestFramework/TestFramework/Attributes/DataSource/CombinatorialDynamicValuesAttribute.cs
```diff
@@ -0,0 +1,277 @@
...
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+namespace Microsoft.VisualStudio.TestTools.UnitTesting.Combinatorial;
+
+/// <summary>
+/// Specifies a member that provides values for a parameter on a combinatorial test method.
+/// </summary>
+[AttributeUsage(AttributeTargets.Parameter)]
+[CLSCompliant(false)]
+public sealed class CombinatorialDynamicValuesAttribute : Attribute, ICombinatorialValuesProvider
```
Evangelink: After reviewing the reduced v1 surface, I suggest deferring `CombinatorialDynamicValuesAttribute` entirely. It is useful, but it contributes roughly 277 lines of bespoke reflection and overload-resolution logic, has no counterpart in the reference implementation, and duplicates concepts already handled by `DynamicDataOperations`. Inline values, ranges, random values, inferred values, and `DynamicData` cover a coherent first release without it.

A later proposal could first extract a shared internal raw-member resolver from `DynamicDataOperations`, then let `DynamicDataAttribute` interpret the result as rows and a dynamic-values provider interpret it as `IEnumerable<T>`. That would give both APIs consistent member lookup, overload, trimming, and source-generation behavior instead of committing this parallel implementation now.

AArnott: Removed `CombinatorialDynamicValuesAttribute` from the v1 surface along with its reflection implementation, API entries, diagnostics, tests, and examples. Updated NativeAOT coverage and the PR description; member-backed values can be reconsidered later on top of a shared `DynamicDataOperations` resolver.

> [!NOTE]

Evangelink (approved):

PR #10912, opened 2026-09-01
Stop repeated acceptance cache failures

Evangelink:
Intermittent Azure Pipeline Cache authorization failures were making every generated acceptance fixture wait about five minutes before its normal fallback build. Those repeated delays caused unrelated tests to appear slow and eventually exhausted the Windows Release job timeout; increasing the timeout only masked the retry storm.

Add a cache-root-scoped circuit breaker shared across acceptance test processes. The first fixture probes the cache while other fixtures wait. If that probe fails, its diagnostics are preserved and all remaining fixtures bypass the cache and use the established `dotnet build` fallback immediately. Successful probes continue enabling normal concurrent cache use, while any later cache failure trips the breaker. The original 120-minute job timeout remains unchanged so genuine hangs stay visible. File-lock contention is recognized using the precise Windows sharing/lock HRESULTs and Unix `EWOULDBLOCK` errno values.

Validation:
- `build.cmd -pack` succeeds.
- All 26 `TestAssetFixtureBaseTests` pass.
- The four circuit-breaker tests passed in 20 consecutive runs (80/80).
- The first CI run confirmed both Windows jobs complete within the original timeout; its unrelated Linux failure exposed the Unix lock-contention mapping now covered by the same tests.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -434,11 +434,10 @@ stages:
         # XML-documentation-only PRs use it only for compiler and package validation. Debug remains
         # skippable for samples-only and XML-doc-only PRs.
         condition: and(succeededOrFailed(), ne(dependencies.DetectChanges.outputs['detect.infrastructureOnly'], 'true'), or(eq(variables._BuildConfig, 'Release'), and(ne(dependencies.DetectChanges.outputs['detect.xmlDocsOnly'], 'true'), ne(dependencies.DetectChanges.outputs['detect.hasProductChanges'], 'false'))))
-        # A completely cold acceptance cache builds every generated fixture. Cold or pool-congested
-        # Release runs can spend more than 135 minutes in the test step after setup and still need
-        # time to finish the suite and publish diagnostics. Keep completion headroom without changing
-        # the test command's existing failure semantics.
-        timeoutInMinutes: 180
+        # A completely cold acceptance cache builds every generated fixture. Keep enough headroom
+        # for that bootstrap path while retaining a bounded timeout that still detects pathologically
+        # slow or hung agents. Warm PR runs remain substantially faster.
+        timeoutInMinutes: 150
         pool:
           name: NetCore-Public
           demands: ImageOverride -equals windows.vs2026preview.scout.amd64.open
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -435,9 +435,9 @@ stages:
         # skippable for samples-only and XML-doc-only PRs.
         condition: and(succeededOrFailed(), ne(dependencies.DetectChanges.outputs['detect.infrastructureOnly'], 'true'), or(eq(variables._BuildConfig, 'Release'), and(ne(dependencies.DetectChanges.outputs['detect.xmlDocsOnly'], 'true'), ne(dependencies.DetectChanges.outputs['detect.hasProductChanges'], 'false'))))
         # A completely cold acceptance cache builds every generated fixture. Keep enough headroom
-        # for that bootstrap path while retaining a bounded timeout that still detects pathologically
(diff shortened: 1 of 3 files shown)


0101 (approved):

azure-pipelines.yml
```diff
@@ -434,10 +434,11 @@ stages:
...
         condition: and(succeededOrFailed(), ne(dependencies.DetectChanges.outputs['detect.infrastructureOnly'], 'true'), or(eq(variables._BuildConfig, 'Release'), and(ne(dependencies.DetectChanges.outputs['detect.xmlDocsOnly'], 'true'), ne(dependencies.DetectChanges.outputs['detect.hasProductChanges'], 'false'))))
-        # A completely cold acceptance cache builds every generated fixture. Keep enough headroom
-        # for that bootstrap path; warm PR runs remain substantially faster and the test command
-        # retains its existing failure semantics.
-        timeoutInMinutes: 120
+        # A completely cold acceptance cache builds every generated fixture. Cold or pool-congested
+        # Release runs can spend more than 135 minutes in the test step after setup and still need
+        # time to finish the suite and publish diagnostics. Keep completion headroom without changing
+        # the test command's existing failure semantics.
+        timeoutInMinutes: 180
```
Youssef1313: Feels to me like a potential real hang.

2 hours is already 4x the typical time needed to run the job.

Evangelink: I checked the full Release log and this was broad agent slowness rather than one hung test. The runner kept progressing through different acceptance classes up to cancellation (including new tests starting at 14:55), no test reached the 16-minute slow-test report, and it produced results for 29,779 passed / 47 skipped / 0 failed tests; the existing 15-minute per-test hang-dump threshold remains unchanged. The extra job headroom therefore does not disable hang detection, but lets these degraded-agent runs finish and publish their diagnostics.

> [!NOTE]

Youssef1313: That's too suspicious to randomly get 4x slowdown. There must be something wrong going on IMO.

Youssef1313:
> [slow] still running after 8m: MSTest.Acceptance.IntegrationTests.DeadlockTests.DeadlockCaseClassInitWaitingOnPreviousTestMethod ("net462")

https://dev.azure.com/dnceng-public/public/_build/results?buildId=1577317&view=logs&j=7c8326b9-0a5f-532a-e6de-db8515c72d9a&t=a2e3061e-1c43-5369-d1c5-db1b5bad61d7

This is extremely suspicious.

azure-pipelines.yml
Evangelink: You were right. Comparing failed and healthy logs showed each generated fixture was spending about five minutes in an unauthorized Azure Pipeline Cache request before its fallback build; the repeated failures produced the apparent suite-wide slowdown. I replaced the timeout change with a cache-root-scoped circuit breaker that preserves the first failure diagnostic, then bypasses cache attempts for the rest of the job, and restored the original 120-minute timeout.

> [!NOTE]

Evangelink:
> [slow] still running after 8m: MSTest.Acceptance.IntegrationTests.DeadlockTests.DeadlockCaseClassInitWaitingOnPreviousTestMethod ("net462")

Agreed that this is suspicious. The name is the outer acceptance test, and its shared `TestAssetFixtureBase` builds the generated `DeadlockTests` asset before the inner deadlock regression executes (the inner test itself uses two one-second waits). In the pre-fix run, the generated asset paid the same roughly five-minute unauthorized cache delay for both its Reflection and SourceGen builds, which explains the eight-plus-minute report. The new circuit breaker preserves the first cache failure, then sends remaining fixture builds directly to `dotnet build`; the original 120-minute job timeout is restored.

PR #11090, opened 2026-09-07
Resolve reference assemblies through the repo NuGet.config in the OS platform analyzer test

nohwnd:
`WhenUnsupportedPlatformHasMessage_AddsExcludeCondition` sets `ReferenceAssemblies` in the object initializer. That replaces the value the shared verifier assigns in its constructor, and drops the repository `NuGet.config` path along with it, so reference-assembly resolution falls back to the Roslyn testing library's own default NuGet configuration and reaches `api.nuget.org`.

`ReferenceAssemblies.Net.Net100` is the reason it shows up here: nothing in this repository targets .NET 10, so `Microsoft.NETCore.App.Ref` 10.x is never restored by the ordinary build and resolution has to reach a feed.

This only fails in the internal pipeline, which is network-isolated — see build [20260907.7 (3068841)](https://dev.azure.com/dnceng/internal/_build/results?buildId=3068841&view=results), where the whole failure is this one test on net8.0 (Linux and Windows) and net472 (Windows). The public gate cannot catch this class of failure because it is not isolated, which is why [microsoft/testfx#11044](https://github.com/microsoft/testfx/pull/11044) went in green.

Chaining `WithNuGetConfigFilePath` onto the override matches what `UseArchitectureConditionAttributeInsteadOfRuntimeCheckAnalyzerTests` and `InheritedMemberFromDifferentMSTestVersionAnalyzerTests` already do.

Verified: with a cleared package folder, `Microsoft.NETCore.App.Ref` 10.0.1 is served by `https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet-public/nuget/v3/index.json`, and `MSTest.Analyzers.UnitTests` passes on both net472 and net8.0.

🤖


Evangelink (approved):
Verified this is the right fix. I independently investigated build [3068920](https://dev.azure.com/dnceng/internal/_build/results?buildId=3068920&view=results) and reached the same root cause: `WhenUnsupportedPlatformHasMessage_AddsExcludeCondition` sets `ReferenceAssemblies` in the object initializer, which drops the `NuGet.config` path the shared verifier constructor assigns, so resolution falls back to `api.nuget.org` — blocked by the network-isolated internal pool.

The fix matches the existing convention: `UseArchitectureConditionAttributeInsteadOfRuntimeCheckAnalyzerTests` and `InheritedMemberFromDifferentMSTestVersionAnalyzerTests` already chain `.WithNuGetConfigFilePath(Path.Combine(RootFinder.Find(), "NuGet.config"))` the same way when overriding `ReferenceAssemblies` in the initializer.

The current CI failures on this PR (`Build` task, `xlf/Resources.cs.xlf is out-of-date with Resources.resx`) are an unrelated pre-existing issue — same failure I saw independently on build 3068877 — and this PR doesn't touch any `.resx`/`.xlf` files, so it's not caused by this change.

PR #11184, opened 2026-09-10
Fix Unix pipe timeout test failure by handling ConnectionReset exception

Evangelink:
## Description
This PR fixes a test failure in the Linux Release build for PR #11180 (backport of #11152 to rel/4.4).

## Problem
The test `ServerDisposal_WhenClientConnectsWithoutSendingRequest_DoesNotHang` in `TestHostControllerCancellationTests.cs` was failing on Linux with .NET 9.0 with the following error:
```
System.IO.IOException: Connection reset by peer ---→ System.Net.Sockets.SocketException: Connection reset by peer
```

## Root Cause
On Unix systems (Linux), when a pipe is forcefully closed, the system reports it as a "connection reset" error instead of returning EOF (0 bytes). The test was not handling this expected behavior.

## Solution
Updated the test to wrap the read operation in a try-catch block that catches `IOException` with the `ConnectionReset` socket error code. This allows the test to pass on both:
- **Windows**: Returns 0 bytes when pipe is closed gracefully
- **Unix**: Throws ConnectionReset exception when pipe is closed abruptly

## Changes
- Updated `ServerDisposal_WhenClientConnectsWithoutSendingRequest_DoesNotHang()` method in `test/UnitTests/Microsoft.Testing.Platform.UnitTests/Hosts/TestHostControllerCancellationTests.cs`
- Added try-catch block to handle Unix-specific ConnectionReset exception
- Added timeout verification using `Task.WhenAny` to ensure disposal completes without hanging

## Related PRs
- Fixes pipeline errors in #11180 (backport of #11152 to rel/4.4)
- This fix mirrors the changes already present in the main branch


Evangelink:
Closing in favor of pushing the fix directly to the existing backport PR #11180.

